PR #13869, opened 2024-11-21
fix #13655 - restore dot notation navigation

### LINKED ISSUE #13655: Grails 7: Accessing config key 'grails.sitemesh.default.layout' through dot notation is deprecated
### Issue description

There is an open discussion on which way to proceed.

Sample App: https://github.com/jamesfredley/grails-website-test/tree/7.0.0-SNAPSHOT

Start with:
./gradlew bootRun
or
./grailsw run-app

```
2024-09-19T08:56:44.757-04:00  WARN 13688 --- [  restartedMain] org.grails.config.NavigableMap           : Accessing config key 'grails.sitemesh.default.layout' through dot notation is deprecated, and it will be removed in a future release. Use 'config.getProperty(key, targetClass)' instead.
```

jdaugherty:


codeconsole:
@jdaugherty  why were all those `public` access modifiers there previously? Was it a Java class that got converted to Groovy?

jdaugherty:
> @jdaugherty why were all those `public` access modifiers there previously? Was it a Java class that got converted to Groovy?

That was my assumption, btw: this is still a draft pull request.  I need to make sure the functional tests are passing before I can activate this.

codeconsole:
> > @jdaugherty why were all those `public` access modifiers there previously? Was it a Java class that got converted to Groovy?
> 
> That was my assumption, btw: this is still a draft pull request. I need to make sure the functional tests are passing before I can activate this.

if that is the case, is going back to Java an option since this class is used so much?

jdaugherty:
> > > @jdaugherty why were all those `public` access modifiers there previously? Was it a Java class that got converted to Groovy?
> > 
> > 
> > That was my assumption, btw: this is still a draft pull request. I need to make sure the functional tests are passing before I can activate this.
> 
> if that is the case, is going back to Java an option since this class is used so much?

I'll look into it.

grails-bootstrap/src/main/groovy/org/grails/config/NavigableMap.groovy
```diff
@@ -488,4 +489,172 @@ class NavigableMap implements Map<String, Object>, Cloneable {
     boolean equals(Object obj) {
```
jdaugherty, line 489: @paulk-asert This class is tagged with both `@EqualsAndHashCode` and it then implements it's own equals / hashcode.  Do you know if the groovy transformation will have any effect?

jdaugherty:
I went ahead and marked this for review so we can restore this function for the milestone.  I'll work on the java conversion separately from this PR and will leave the ticket open.

jamesfredley (approved):

codeconsole (approved):

PR #13877, opened 2024-11-23
Add minimum Java version so the grails forge cli can pull that property instead of defaulting based on its own version

jdaugherty:


jdaugherty:
I have an associated 7.0.0 branch here:  jdaugherty:feature/7.0.0/pomJavaVersion

jdaugherty:
Closing this, we're going to default the version in the cli instead.

codeconsole:
@jdaugherty couldn't you have just pulled this from spring?  The spring-bom is imported using the grails-gradle-plugin and I am pretty sure spring has a parent pom with this same information

jdaugherty:
Probably but we don't really need to make it that complicated because either way we have to update the JdkVersion enum in the cli project and there was already a placeholder for the default

codeconsole:
actually, that wouldn't have worked because Spring went from 8 > 17

PR #13905, opened 2024-12-01
Mark dispatch action as deprecated due to grails-gsp#551

jdaugherty:


codeconsole:
I think we should log a warning when it is invoked to make people aware of the deprecation. What do you think?

jdaugherty:
> I think we should log a warning when it is invoked to make people aware of the deprecation. What do you think?

That's a great idea.  I've updated the PR to add a warning if the action is not null and log warnings are enabled.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingInfo.java
```diff
@@ -212,6 +215,13 @@ public String getId() {
...
     }
 
+    /**
+     * @deprecated
+     * This method will be removed in a future grails version since the associated g:submitAction is being removed.
+     * Grails will no longer support redirecting to a different action name by adding a parameter with the prefix
+     * '_action'
+     */
+    @Deprecated(since = "7.0.0")
     private String checkDispatchAction(HttpServletRequest request) {
```
matrei, line 225: Is it necessary to deprecate a private method?

jdaugherty: I guess we could use a todo instead.  I would prefer the annotation because it lets us know when it was deprecated so we know when to remove it.  Also, it lets us know via compiler checks

codeconsole: I think that is useful.   It seems like there is a history of not removing things as planned.  Annotating makes it clearer.

codeconsole: would it be beneficial to say use g: formActionSubmit instead in the warning message?

jdaugherty: It's possible someone is relying on this behavior to redirect for a reason we don't know about.  To my knowledge, the only "framework" reason is that tag ... so I guess.  Do you have a proposed wording?

matrei: Should we also add `forRemoval = true` attribute?

jdaugherty: I added the attribute, @codeconsole do you have a suggestion on the warning message?

codeconsole: How about just add it at the end?  The tag is the only place this is used, right?

```java
"Dispatch Action [%s] detected; Dispatch Actions will be removed in a future version of Grails. Use g: formActionSubmit instead."
```

jdaugherty:
@codeconsole I updated the message.

codeconsole (approved):

PR #13908, opened 2024-12-05
Don't set layout if none is specified

codeconsole:
Currently if no layout is specified, __none__ is set.  Since there is no concept of a none layout in Sitemesh 3, it attempts to find a layout named __none__.

Before 7 is final, more time will need to be spent perfecting layout resolution.   There is not clear separation of concerns.

The following question needs to be addressed prior to RC1

Should layout resolution be specific to controllers?  currently (and prior) this trait is applied to Controllers, RestControllers, and Interceptors.  I am not sure if RestControllers or Interceptors should be so tightly coupled to SiteMesh.

Perhaps all this layout logic should be removed and done directly inside Controller.

### DIFF
--- a/grails-plugin-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRenderer.groovy
+++ b/grails-plugin-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRenderer.groovy
@@ -139,7 +139,7 @@ trait ResponseRenderer extends WebAttributes {
         else {
             renderMarkupInternal webRequest, closure, response
         }
-        setLayout webRequest.currentRequest, false, layoutArg
+        setLayout webRequest.currentRequest, layoutArg
     }
 
     private void renderJsonInternal(HttpServletResponse response, @DelegatesTo(value = StreamingJsonBuilder.StreamingJsonDelegate.class, strategy = Closure.DELEGATE_FIRST) Closure callable) {
@@ -163,7 +163,7 @@ trait ResponseRenderer extends WebAttributes {
         applyContentType response, argMap, body
         handleStatusArgument argMap, webRequest, response
         render body
-        setLayout webRequest.currentRequest, false, layoutArg
+        setLayout webRequest.currentRequest, layoutArg
     }
 
     /**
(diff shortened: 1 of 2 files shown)


codeconsole:
https://github.com/codeconsole/grails-plugin-sitemesh3/issues/4

grails-plugin-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRenderer.groovy
```diff
@@ -533,13 +533,12 @@ trait ResponseRenderer extends WebAttributes {
     }
 
     private void applySiteMeshLayout(HttpServletRequest request, boolean renderView, String explicitSiteMeshLayout) {
```
matrei: Rename method to `setLayout` and possibly make `static`?
Param `renderView` is unused both before and after this change. Remove?
Rename param `explicitSiteMeshLayout` to `layout`?

grails-plugin-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRenderer.groovy
matrei, line 1: Rename all occurrences in this file of variable name `explicitSiteMeshLayout` outside of `applySiteMeshLayout()` to `layoutArg`?

matrei (approved):
Excellent! This resolves the `render(text: '...')` problem and three more test are successful in `grails-functional-tests`.

I have made some comments that I think will make the code clearer and also make it less coupled with a specific layout engine.

jdaugherty (approved):

matrei:
Also the `WebUtils.NONE_LAYOUT` is now unused, Delete?

jeffscottbrown:
Instead of this…

```groovy
private void setLayout(HttpServletRequest request, String layout) {
    if (layout == null || request.getAttribute(WebUtils.LAYOUT_ATTRIBUTE) != null) {
        // layout has been set already
        return
    }
    request.setAttribute WebUtils.LAYOUT_ATTRIBUTE, layout
}
```

Could it just be this…

```groovy
private void setLayout(HttpServletRequest request, String layout) {
    // if layout is provided and the layout attribute hasn’t already been set, set it
    if (layout != null && request.getAttribute(WebUtils.LAYOUT_ATTRIBUTE) == null) {
        request.setAttribute WebUtils.LAYOUT_ATTRIBUTE, layout
    }
}
```

codeconsole:
> Instead of this…

true, I was working backwards.  I guess I was trying to preserve as much of the original code as possible.

jeffscottbrown:
> I guess I was trying to preserve as much of the original code as possible.

My suggestion works against that goal.  Please ignore.

PR #13911, opened 2024-12-05
Remove no longer needed return statement

codeconsole:


codeconsole:
https://github.com/grails/grails-core/pull/13908

Issue #13840, opened 2024-11-15
Grails 7: grails-core: update domain inheritance tests


jamesfredley:
### Issue description

Now that domain inheritance is restored with Groovy 4.0.24

https://github.com/search?q=org%3Agrails%20groovy-5106&type=code

jdaugherty:
FYI: https://issues.apache.org/jira/browse/GROOVY-10140 is what broke the DomainConstraintGettersSpec.  I'm still researching.  I have confirmed that if you put transient on both a getter & setter in groovy 3 , then the metaclass will show that property as transient.  This support appears removed in Groovy 4.

jdaugherty:
Example groovy code that works in groovy 3 showing the property is transient: 

		import java.lang.reflect.* 

		class Test {
			transient String somePropertyName

			transient String getSomeOtherProperty() { null }
			transient void setSomeOtherProperty(String ignored) { }
		}

		def testing = new Test()
		def metaClass = testing.metaClass
		System.out.println("Found Properties:")
		System.out.println(metaClass.getProperties().collect{ it.name } )

		System.out.println("Some Other Property Modifiers")

		def values = metaClass.getMetaProperty('someOtherProperty').modifiers
		System.out.println(Modifier.isTransient(values))

jdaugherty:
Upon further research, the following tests were straight out ignored in 6.2.x: 

	ensure only public non-static bool properties with getter and setter are constrained properties
	ensure constrained bool method properties are only public ones with both getter and setter
	ensure only public non-static inherited bool properties with getter and setter are constrained properties
	ensure constrained inherited bool method properties are only public ones with both getter and setter
	ensure only public non-static bool properties from trait with getter and setter are constrained properties
	ensure constrained bool method properties from trait are only public ones with both getter and setter
	ensure trait transient properties and methods are not validated
	ensure trait transient methods and properties are not constrained

jdaugherty:
Concerning the `transient` keyword on a method in the `DomainConstraintGettersSpec`, since it doesn't work in 6.2.x on properties and did work on the method form, with the method form removed, these methods can be removed from the example classes.

jdaugherty:
The test `DomainConstraintGettersSpec` used the `is{PropertyName}` syntax instead of `get{PropertyName}`.  Switching to the `get` syntax fixes the previously ignored non-trait tests.  

[...]

PR #13915, opened 2024-12-10
#13840 - Add back @Entity for GORM Inheritance

jdaugherty:
This fixes all of the gorm inheritance tests (#13840) except the transient property ones.  In groovy 4 you can't have this type of code any more: 

     transient String getTransientStringProperty{ null } 

You can have: 

     transient String transientStringProperty

But gorm is still constraining these properties so it needs further research on the grails-data-mapping project.  I hope to investigate this one.

### DIFF
--- a/grails-test-suite-persistence/src/test/groovy/grails/web/databinding/GrailsWebDataBinderSpec.groovy
+++ b/grails-test-suite-persistence/src/test/groovy/grails/web/databinding/GrailsWebDataBinderSpec.groovy
@@ -1804,7 +1804,7 @@ class Widget {
     }
 }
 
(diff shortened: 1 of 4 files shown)


jdaugherty:
I have held off on checking in `DomainConstraintGettersSpec` until I can get to the root of the constrained property issue.

jdaugherty:
I've gone ahead and fixed `DomainConstraintGettersSpec` , see ticket comments for research & to preserve these findings historically.

matrei (approved):

jdaugherty:
FYI The 'is' property syntax is only valid in the java bean spec for primitive booleans.  It is not valid for Boolean.class.  This is why the tests were failing in 6.x because hibernate did nit consider them a property.

Issue #13690, opened 2024-09-26
Grails 7: Finish Micronaut to Spring Conversion


codeconsole:
Full list of failing tests:  https://github.com/grails/grails-core/pull/13702/files

There are parts of grails-core are not 100% working. These are identified by failing tests which need to be fixed.

 
[DataBindingConfigurationSpec.groovy](https://github.com/grails/grails-core/blob/7.0.x/grails-plugin-databinding/src/test/groovy/org/grails/plugins/databinding/DataBindingConfigurationSpec.groovy) Ordering

RespondMethodSpec
ContentFormatControllerTests
CommandObjectNoDataSpec
WithFormatContentTypeSpec
GrailsWebDataBinderConfigurationSpec
CommandObjectsSpec

```
|    +--- io.micronaut:micronaut-inject:4.5.3
|    |    +--- org.slf4j:slf4j-api:2.0.13 -> 2.0.16
|    |    +--- jakarta.inject:jakarta.inject-api:2.0.1
|    |    +--- jakarta.annotation:jakarta.annotation-api:2.1.1 -> 3.0.0
|    |    \--- io.micronaut:micronaut-core:4.5.3
|    |         \--- org.slf4j:slf4j-api:2.0.13 -> 2.0.16
|    +--- jakarta.inject:jakarta.inject-api:2.0.1
|    +--- io.micronaut.spring:micronaut-spring-context:5.7.0
|    |    +--- io.micronaut.cache:micronaut-cache-core:4.3.0
|    |    |    +--- io.projectreactor:reactor-core:3.5.11 -> 3.6.9
|    |    |    |    \--- org.reactivestreams:reactive-streams:1.0.4
|    |    |    +--- io.micronaut:micronaut-inject:4.4.0 -> 4.5.3 (*)
|    |    |    +--- io.micronaut:micronaut-aop:4.4.0 -> 4.5.3
|    |    |    |    +--- org.slf4j:slf4j-api:2.0.13 -> 2.0.16
|    |    |    |    +--- io.micronaut:micronaut-inject:4.5.3 (*)
|    |    |    |    \--- io.micronaut:micronaut-core:4.5.3 (*)
|    |    |    +--- io.micronaut:micronaut-http:4.4.0 -> 4.5.3
|    |    |    |    +--- io.projectreactor:reactor-core:3.5.11 -> 3.6.9 (*)
|    |    |    |    +--- org.slf4j:slf4j-api:2.0.13 -> 2.0.16
[...]

PR #13917, opened 2024-12-11
#13690 - ensure the data binder is initialized lazily

jdaugherty:


codeconsole (changes requested):
Please add comment why it needs to be lazy and at least 1 reference to 1 ValueConverter that is not getting initialized when it is not lazy.

codeconsole (approved):

PR #13930, opened 2024-12-20
Spring Boot 3.4.1

codeconsole:
Tested with `grails-gsp` and `grails-spring-security-core` plugin
Ran grails-functional-tests
Ran example app
Ran production app


codeconsole:
I need to merge this to get the production app running in a production environment.   I think it should go out in M1.  If there are any strong objections, we can revert, but considering there are 80 bug fixes in this I advise moving forward. 

I will be deploying this into a live environment shortly.

PR #13946, opened 2024-12-27
back to snapshots of dependencies

jamesfredley:
Required to resolve https://github.com/grails/grails-core/actions/runs/12485931976/job/34926927726 which was coming from `grails-bootstrap:7.0.0-M1`

Plus add grails-bom and grails-plugin-codecs dependencies


jamesfredley:
needed https://github.com/grails/grails-testing-support/pull/460

matrei (approved):

PR #13948, opened 2024-12-31
handle properties version lookup in grails-bom

### LINKED ISSUE #13760: Distribution Fails to Run
### Expected Behavior

```console
.\gradlew assemble
unzip .\build\distribrutions\grails-7.0.0-SNAPSHOT.zip
.\build\distributions\grails-7.0.0-SNAPSHOT\bin\grails -version
| Grails Version: 7.0.0-SNAPSHOT
| JVM Version: 17.0.12
```

### Actual Behaviour

```console
.\gradlew assemble
unzip .\build\distribrutions\grails-7.0.0-SNAPSHOT.zip
.\build\distributions\grails-7.0.0-SNAPSHOT\bin\grails -version

Exception: java.lang.NoSuchMethodError thrown from the UncaughtExceptionHandler in thread "main"
```

### Steps To Reproduce

_No response_

### Environment Information

- Operating system: Windows 11
- JDK Version: OpenJDK 64-Bit Server VM Temurin-17.0.12+7 (build 17.0.12+7, mixed mode, sharing)

### Example Application

_No response_

### Version

7.0.0-SNAPSHOT

jamesfredley:
test with .`/gradlew assemble`
unzip build/distributions/grails-7.0.0-SNAPSHOT.zip
bin/grails 

Handles properties version lookup in grails-bom

https://repo.grails.org/ui/repos/tree/PomView/libs-snapshots-local/org/grails/grails-bom/7.0.0-SNAPSHOT/grails-bom-7.0.0-20241229.020850-191.pom

${profiles-angular.version} was coming from grails-bom and `org.grails.cli.boot.GrailsDependencyVersions` needed to be updated to handle version lookup

```
<properties>
     <profiles-angular.version>10.0.1</profiles-angular.versio>
</properties>
<dependencyManagement>
     <dependencies>
          <dependency>
               <groupId>org.grails.profiles</groupId>
               <artifactId>angular</artifactId>
               <version>${profiles-angular.version}</version>
          </dependency>
     </dependencies>
</dependencyManagement>
```

### DIFF
--- a/grails-shell/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
+++ b/grails-shell/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 original authors
+ * Copyright 2014-2024 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -96,12 +96,9 @@ class GrailsDependencyVersions implements DependencyManagement {
      *
      * @return the version with lookup from properties when required
      */
-    String versionLookup(String version){
-        if (version?.startsWith('${') && version?.endsWith('}')) {
-            return versionProperties.get(version.substring(2, version.length()-1))
-        }
-
-        return version
+    String versionLookup(String version) {
+        version?.startsWith('${') && version?.endsWith('}') ?
+                versionProperties[version[2..-2]] : version
     }
 
     protected void addDependency(String group, String artifactId, String version) {
--- a/grails-shell/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
+++ b/grails-shell/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
@@ -68,10 +68,40 @@ class GrailsDependencyVersions implements DependencyManagement {
 
     @CompileDynamic
     void addDependencyManagement(GPathResult pom) {
-        pom.dependencyManagement.dependencies.dependency.each { dep ->
-            addDependency(dep.groupId.text(), dep.artifactId.text(), dep.version.text())
-        }
         versionProperties = pom.properties.'*'.collectEntries { [(it.name()): it.text()] }
+        pom.dependencyManagement.dependencies.dependency.each { dep ->
+            addDependency(dep.groupId.text(), dep.artifactId.text(), versionLookup(dep.version.text()))
+        }
+    }
+
+    /**
+     * Handles properties version lookup in grails-bom
+     *
+     *   <properties>
+     *    <ant.version>1.10.15</ant.version>
+     *   </properties>
+     *
+     *   <dependencyManagement>
+     *    <dependencies>
+     *     <dependency>
+     *      <groupId>org.apache.ant</groupId>
+     *      <artifactId>ant</artifactId>
+     *      <version>${ant.version}</version>
+     *     </dependency>
+     *    </dependencies>
+     *   </dependencyManagement>
+     *
+     * @param version
+     *            either the version or the version to lookup
+     *
+     * @return the version with lookup from properties when required
+     */
+    String versionLookup(String version){
+        if (version?.startsWith('${') && version?.endsWith('}')) {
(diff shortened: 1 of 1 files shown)


jdaugherty (approved):

codeconsole (dismissed):
Since it is groovy, how about?

```groovy
    String versionLookup(String version){
    	version?.startsWith('${') && version.endsWith('}') ?
        	versionProperties[version[2..-2]] : version
    }
```

matrei (approved):
Thank you @jamesfredley! 🚀  

You could also update the license header year of `GrailsDependencyVersions`.

gsartori:
> Since it is groovy, how about?
> 
> ```groovy
>     String versionLookup(String version){
>     	version?.startsWith('${') && version.endsWith('}') ?
>         	versionProperties[version[2..-2]] : version
>     }
> ```

The Groovy feature of not having a mandatory `return` is nice if the whole team agrees on that. Since we want to attract Java developers (at least this topic is in the air) I would keep `return`. Actually, @codeconsole, given our `def/var` fight I am surprised you  are proposing this :)

This would be my version:

```groovy
String versionLookup(String version) {
    Boolean isVariable = version?.startsWith('${') && version.endsWith('}')
    return isVariable
            ? versionProperties[version[2..-2]]
            : version
}
```

gsartori (approved):

matrei:
> This would be my version:
> 
> ```groovy
> String versionLookup(String version) {
>     Boolean isVariable = version?.startsWith('${') && version.endsWith('}')
>     return isVariable
>             ? versionProperties[version[2..-2]]
>             : version
> }
> ```

I like this, except the capital B on Boolean.

Also, there is a null-check on the `version`. Is that necessary? Is null a valid value for `version`?

gsartori:
`Boolean` vs `boolean` (or any Object vs primitive type in the JVM) is an old story. I prefer to keep the code "readble and coherent" VS "I believe it will run farter". The cost of maintaining the code is immensly bigger than the benefit of a quark-improvement in performances.

The difference is nano-seconds, and not really relevant in 98% of use cases. I know we are talking religions here, but give a look here: https://stackoverflow.com/questions/6911563/java-performance-true-vs-boolean-true/6911627#6911627

matrei:
> `Boolean` vs `boolean` (or any Object vs primitive type in the JVM) is an old story. I prefer to keep the code "readble and coherent" VS "I believe it will run farter". The cost of maintaining the code is immensly bigger than the benefit of a quark-improvement in performances.
> 
> The difference is nano-seconds, and not really relevant in 98% of use cases. I know we are talking religions here, but give a look here: https://stackoverflow.com/questions/6911563/java-performance-true-vs-boolean-true/6911627#6911627

Why is `Boolean` more readable than `boolean`?
I did not suggest it because "I believe it will run faster".
I suggested it, because I believe it is the correct type.

gsartori:
> Why is `Boolean` more readable than `boolean`? 

Writing/Reading code requires mental energy (dopaminergic circuit mainly). Each decision we take requires mental energy. When we run out of it, we start makining mistakes. Mental energy (fuel) and mistakes (crashes) are "costs" from a Business point of view.

Always using type wrappers instead of deciding when to use a primitive `type` or an `Object` is one less decision to take. You just use a `Type`. Is less mental energy -> less cost.

On a personal note: from a style point of view it is more coherent, we are developing Object Oriented code, not a Procedural "C" like code.

matrei:
>Writing/Reading code requires mental energy (dopaminergic circuit mainly)

My mental energy loss is great when I see boxing/unboxing primitive types for no reason. I have to ask myself: why was this done?

>Always using type wrappers instead of deciding when to use a primitive type or an Object is one less decision to take. You just use a Type. Is less mental energy -> less cost.

We'll have to agree to disagree on this one 😄

gsartori:
> My mental energy loss is great when I see boxing/unboxing primitive types for no reason. I have to ask myself: why was this done?

That's because we haven't agreend on the coding style yet. Having "standards" is another argument to reduce costs and increment quality. Using only `Types` whould mean having less variance both in function and in style.

Occam with its Occam's Razor would have agreed with me: we would have less variables, or variance in this case, to accompish the same task.

jamesfredley:
@codeconsole Thanks, I've updated it to the groovier syntax.  

@matrei I have updated to license date.

Issue #13552, opened 2024-07-17
Restructure Grails-Core repository


sdelamo:
see: https://github.com/orgs/grails/discussions/13516

repo: https://github.com/grails/grails-async

jdaugherty:
Per discussion we have agreed to move the following into grails-core:
1. grails-async 
2. grails-testing-support for non-gorm based testing

jdaugherty:
I was able to import grails-async's project history into the grails-core repository (locally) with the following commands:

    git remote add async git@github.com:grails/grails-async.git
    git fetch async
    git merge -s ours --no-commit --allow-unrelated-histories async/6.0.x
    git read-tree --prefix=grails-async/ -u async/6.0.x
    git commit -m "Imported grails-async"

But when I started testing includeBuild approach to keep the build isolated, I ran into these issues:
* https://github.com/gradle/gradle/issues/2534 - which is easy enough to work around by importing properties in the child project's settings.gradle
* https://github.com/gradle/gradle/issues/22521 - which prevents using includeBuild for our projects.

This means we'll likely have to just use include in the grails projects.  

At a high level, here's what needs to happen for any migrations:
1. Clean up the project to use the same dependency management as grails core (async uses version catalogs while core does not)
2. Remove buildSrc in the project to facilitate easier moves
3. Clean up the `gh-pages` branch in the child project
4. Import the code history into the default branch
5. Import the pages branch into the gh-pages branch
6. Remove old `.github` folders in child project
7. Move child project's gradle projects into root
8. Update any includes / references

jdaugherty:
`grails-async` will move to the following locations: 
1. grails-core in the `grails-events` , `grails-async` , and `grails-test-examples` folders
https://github.com/grails/grails-core/pull/13955

2. grails-doc in the async section the grails doc guide
https://github.com/grails/grails-doc/pull/947

jdaugherty:
Project Restructure Plan
* Note:
     * ✅ Scaffolding & fields have been decoupled over the past week so they are no longer be required for views
     * ✅ grails-async has been decoupled from views, etc and can now be moved to core
     * ✅ grails-views was decoupled from the gradle plugin
* Pre-reqs:
     * ✅ grails-testing-support needs to be split into 2 separate repos; one for gorm and one for grails-core.  this removes the grails-data-mapping dependency which means grails views could be moved into core
* Grails Core should have:
     * ✅ grails-async (move to core) - https://github.com/grails/grails-core/pull/13955
[...]

PR #13955, opened 2025-01-07
#13552 Merge grails-async into grails-core

jdaugherty:
* All of the tests pass locally.
* Docs will be moved to grails-doc repo (separate PR)
* Old async github configuration (workflows, dependabot, issue templates, etc) is removed after importing
* I preserved the artifact names & project info for old async artifacts
* async was split into `grails-async` , `grails-events`, & `grails-test-examples` folders
* I found a bug where the test projects were in the bom, they are now removed
* fixed async & events in the bom to use the grails version
* any part that was questionable has this comment prefix `TODO - async move`

This PR contains the entire history of grails-async, which was imported under /grails-async.  It then modifies the original async structure to be like this: 

        grails-async
        ├── core
        │   ├── build
        │   └── src
        ├── gpars
        │   ├── build
        │   └── src
        ├── plugin
        │   ├── build
        │   ├── grails-app
        │   └── src
        ├── rxjava
        │   ├── build
        │   └── src
        └── rxjava2
            ├── build
            └── src
        grails-events
        ├── compat
        │   ├── build
        │   └── src
        ├── core
        │   ├── build
        │   └── src
        ├── gpars
        │   ├── build
        │   └── src
        ├── plugin
        │   ├── build
        │   ├── grails-app
        │   └── src
        ├── rxjava
        │   ├── build
        │   └── src
        ├── rxjava2
        │   ├── build
        │   └── src
        ├── spring
        │   ├── build
        │   └── src
        └── transform
            ├── build
            └── src
        grails-test-examples
        └── async-events-pubsub-demo
            ├── grails-app
            └── src

To make this easier to review, only review the commits after "Imported grails-async" ... any commit prior is just imported for-like under `/grails-async`.  Commits after that are moving the project, splitting it, etc.


jdaugherty:
Please do not merge this until the associate doc pull request is ready.

build.gradle
```diff
@@ -166,13 +166,23 @@ apply from: "gradle/idea.gradle"
...
     version = grailsVersion
-    group = "org.grails"
 
     ext.isTestSuite = subproject.name.startsWith("grails-test-suite")
+    ext.isExample = subproject.name.contains("grails-test-examples")
+    ext.isPlugin = subproject.name in ['grails-async-plugin', 'grails-events-plugin']
     ext.isCiBuild = subproject.hasProperty("isCiBuild") || System.getenv().get("CI") as Boolean
+    ext.isAsyncProject = subproject.name.startsWith('grails-async') || subproject.name.startsWith('grails-events')
+
+    if(isPlugin) {
```
matrei: Space after if?

build.gradle
```diff
@@ -166,13 +166,23 @@ apply from: "gradle/idea.gradle"
 subprojects { subproject ->
 
     version = grailsVersion
-    group = "org.grails"
 
     ext.isTestSuite = subproject.name.startsWith("grails-test-suite")
+    ext.isExample = subproject.name.contains("grails-test-examples")
```
matrei: Single quotes?

build.gradle
```diff
@@ -166,13 +166,23 @@ apply from: "gradle/idea.gradle"
...
+    ext.isPlugin = subproject.name in ['grails-async-plugin', 'grails-events-plugin']
     ext.isCiBuild = subproject.hasProperty("isCiBuild") || System.getenv().get("CI") as Boolean
+    ext.isAsyncProject = subproject.name.startsWith('grails-async') || subproject.name.startsWith('grails-events')
+
+    if(isPlugin) {
+        group = 'org.grails.plugins'
+    }
+    else {
+        group = "org.grails"
```
matrei: Single quotes?

jdaugherty:
https://github.com/grails/grails-doc/pull/947 is the doc PR.  For merge order: we should merge this one first, and then that one (and kick off the doc build again).

build.gradle
```diff
@@ -166,13 +166,23 @@ apply from: "gradle/idea.gradle"
...
+    }
+
     ext.pomInfo = {
-        delegate.name 'Grails® framework'
-        delegate.description 'Grails Web Application Framework'
+        delegate.name isAsyncProject ? 'Grails Async Framework' : 'Grails® framework'
+        delegate.description isAsyncProject ? 'Grails Async Libraries' : 'Grails Web Application Framework'
```
matrei: I'm not a fan of having conditional logic to handle different sub-projects in the root build file. It eventually leads to a big bowl of spaghetti that few understands or dares to touch. I prefer more of a "composition over inheritance" and "separation of concerns"-approach. Which in this particualar case for example would be: `grails-async/core/build.gradle`  sets `ext.projectTitle = 'Grails Async Framework'` and includes `gradle/publish-config.gradle`, which uses `delegate.name project.findProperty('projectTitle') ?: 'Grails® framework'`.

I know it has been this way, and we don't necessarily need to change it in this PR, but as we get bigger builds I think we should consider moving to this approach.

build.gradle
```diff
@@ -189,14 +199,16 @@ subprojects { subproject ->
             delegate.developerConnection 'scm:git@github.com:grails/grails-core.git'
         }
 
+        List developers = isAsyncProject ? ['Graeme Rocher', 'Mattias Reichel'] : ['Graeme Rocher']
```
matrei: You can remove my name here. It is not warranted.

codeconsole:
After reviewing a few things, I think bringing this in may a bad idea:

1.  grails-async and events aren't even used by a starter grails app or in most grails applications?
2. the preferred approach going forward should probably be using `@Async`? 

if 1 and 2 are true, this code really doesn't belong in this repo and really should be on an independent release cycle.

matrei:
> 1. grails-async and events aren't even used by a starter grails app or in most grails applications.

`grails-spring-security-core` depends on them.

> 2. the preferred approach going forward should probably be using `@Async`?

This may or may not be true. I think `grails-async` enables a very groovy programming style.

build.gradle
```diff
@@ -248,36 +260,36 @@ subprojects { subproject ->
...
 
-                        developers {
-                            developer {
-                                id = 'graemerocher'
-                                name = 'Graeme Rocher'
-                                email = 'grails-build@users.noreply.github.com'
-                            }
-                        }
+                        // TODO - async move - Other grails projects are excluding dependency management, historically grails-core did not
+                        // now that async is moved in, it's no longer excluding it
```
matrei: I think we should exclude the `dependencyManagement` tag in every POM except for `grails-bom` (which is not using this code).

gradle/assemble.gradle
```diff
@@ -92,7 +92,8 @@ tasks.register('sourcesJars', Sync) {
     from rootProject.subprojects.findAll { subproject ->
         !subproject.name.startsWith('grails-test-suite') &&
         !subproject.name.startsWith('grails-dependencies') &&
-        !subproject.name.startsWith('grails-bom')
+        !subproject.name.startsWith('grails-bom') &&
+        !subproject.name.contains("grails-test-examples")
```
matrei: Single quotes?

grails-async/core/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-async/core/build.gradle
```diff
@@ -1,13 +1,12 @@
 plugins {
+    id 'groovy'
     id 'java-library'
```
matrei: Both these are already applied to this project in the root `build.gradle`. But I think that is a bad practice, so I am fine with keeping these here.

grails-async/gpars/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-async/plugin/build.gradle
```diff
@@ -1,26 +1,25 @@
...
-    id 'org.grails.grails-plugin'
-    id 'maven-publish'
-    id 'signing'
 }
 
+group = 'org.grails.plugins'
+
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-async/plugin/build.gradle
```diff
@@ -1,26 +1,25 @@
...
 
     api project(':grails-async-core')
     api project(':grails-events-compat')
     api project(':grails-events-core')
-    api 'org.grails:grails-core'
-    api 'org.grails:grails-web-common'
+    api project(':grails-core')
```
matrei: Sort alphabetically?

grails-async/rxjava/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-async/rxjava2/build.gradle
```diff
@@ -1,14 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-bom/build.gradle
```diff
@@ -142,6 +139,7 @@ publishing {
 
                 project.parent.subprojects.findResults {
                     if (it.name == 'grails-bom') return null
+                    if (it.name.startsWith("grails-test-suite") || it.name.contains("grails-test-examples")) return null
```
matrei: Single quotes?

grails-events/compat/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'groovy'
+    id 'java-library'
+}
+
+group = 'org.grails'
+
+dependencies {
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/core/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/gpars/build.gradle
```diff
@@ -1,14 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/plugin/build.gradle
```diff
@@ -1,12 +1,12 @@
...
-    id 'org.grails.grails-plugin'
-    id 'maven-publish'
-    id 'signing'
 }
 
+group = 'org.grails.plugins'
+
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/plugin/build.gradle
```diff
@@ -18,17 +18,11 @@ dependencies {
 
     // As this is a plugin these dependencies will be provided by the Grails application
     // This removes circular dependencies
-    compileOnly 'org.grails:grails-bootstrap'
-    compileOnly 'org.grails:grails-core'
+    implementation project(':grails-bootstrap')
+    implementation project(':grails-core')
```
matrei: I still think it's good practice to use `compileOnly` where ever we can to promote decoupling. We know that these two libraries will always be provided by the consuming application.

grails-events/rxjava/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/rxjava/build.gradle
```diff
@@ -1,13 +1,12 @@
...
 
     api project(':grails-events-core')
     api 'io.reactivex:rxjava:1.3.8'
```
matrei: Should this version be moved to `gradle.properties`?

grails-events/gpars/build.gradle
```diff
@@ -1,14 +1,12 @@
...
 
     api project(':grails-events-core')
     api 'org.codehaus.gpars:gpars:1.2.1', {
```
matrei: Should this version be moved to `gradle.properties`?

grails-async/rxjava2/build.gradle
```diff
@@ -1,14 +1,12 @@
...
 
     api project(':grails-async-core')
     api 'io.reactivex.rxjava2:rxjava:2.2.21'
```
matrei: Should this version be moved to `gradle.properties`?

grails-async/rxjava/build.gradle
```diff
@@ -1,13 +1,12 @@
...
 
     api project(':grails-async-core')
     api 'io.reactivex:rxjava:1.3.8'
```
matrei: Should this version be moved to `gradle .properties`?

grails-async/gpars/build.gradle
```diff
@@ -1,13 +1,12 @@
...
 
     api project(':grails-async-core')
     api 'org.codehaus.gpars:gpars:1.2.1', {
```
matrei: Should this version be moved to `gradle.properties`?

gradle/assemble.gradle
```diff
@@ -166,7 +167,7 @@ tasks.register('zipDist', Zip) {
     exclude '**/CVS/**'
 
     into("grails-$grailsVersion") {
-        from(rootProject.layout.projectDirectory) {
+        from(project.layout.projectDirectory) {
```
matrei: As this is an file meant for inclusion, I think we should keep this as `rootProject.layout.projectDirectory` to be explicit about which project it is referring to.

grails-events/rxjava2/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/rxjava2/build.gradle
```diff
@@ -1,13 +1,12 @@
...
 
     api project(':grails-events-core')
     api 'io.reactivex.rxjava2:rxjava:2.2.21'
```
matrei: Should this version be moved to `gradle.properties`?

grails-events/spring/build.gradle
```diff
@@ -1,14 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-events/transform/build.gradle
```diff
@@ -1,13 +1,12 @@
...
     id 'java-library'
-    id 'maven-publish'
-    id 'signing'
 }
 
 group = 'org.grails'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-test-examples/async-events-pubsub-demo/build.gradle
```diff
@@ -7,19 +18,19 @@ version = rootProject.version
 group = 'pubsub.demo'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
```
matrei: Single quotes?

grails-test-examples/async-events-pubsub-demo/build.gradle
```diff
@@ -1,3 +1,14 @@
+buildscript {
+    repositories {
+        gradlePluginPortal()
+        mavenCentral()
```
matrei: Remove `gradlePluginPortal()` and `mavenCentral()`?

`id 'org.grails.grails-web'` needs to be changed to `apply plugin: 'org.grails.grails-web'`, preferably below the setting of group and version.

Dependency `org.grails:grails-web-testing-support` is not resolving correctly. Cannot resolve version for transitive dependency `org.grails:grails-test`. This is probably a misconfiguration in `grails-testing-support` (no BOM applied?)

The tests seems to fail with a problem over competing LoggerFactory implementations, need investigating.

settings.gradle
```diff
@@ -76,5 +76,42 @@ include (
...
+project(':grails-events-spring').projectDir = file('grails-events/spring')
+project(':grails-events-rxjava').projectDir = file('grails-events/rxjava')
+project(':grails-events-rxjava2').projectDir = file('grails-events/rxjava2')
+project(':grails-events-plugin').projectDir = file('grails-events/plugin')
+
+project(':grails-async-core').projectDir = file('grails-async/core')
+project(':grails-async-gpars').projectDir = file('grails-async/gpars')
+project(':grails-async-rxjava').projectDir = file('grails-async/rxjava')
+project(':grails-async-rxjava2').projectDir = file('grails-async/rxjava2')
+project(':grails-async-plugin').projectDir = file('grails-async/plugin')
```
matrei: Also add:
`project(':grails-test-examples-async-events-pubsub-demo').projectDir = file('grails-test-examples/async-events-pubsub-demo')`

grails-test-examples/async-events-pubsub-demo/build.gradle
```diff
@@ -7,19 +18,19 @@ version = rootProject.version
 group = 'pubsub.demo'
 
 dependencies {
-    implementation platform("org.grails:grails-bom:${grailsVersion}")
+    implementation platform(project(":grails-bom"))
 
     implementation project(':grails-plugin-async')
     implementation project(':grails-plugin-events')
```
matrei: Project names are `:grails-async-plugin` and `:grails-events-plugin`.

grails-test-suite-uber/build.gradle
```diff
@@ -25,7 +25,7 @@ dependencies {
                 project(':grails-spring')
 
     testImplementation "org.grails:grails-datastore-gorm-hibernate5"
-    testImplementation "org.grails.plugins:async"
+    testImplementation project(":grails-async-core")
```
matrei: Single quotes?
This dependency is changed from `plugin` to `core`. Is that deliberate?

grails-test-suite-web/build.gradle
```diff
@@ -29,7 +29,7 @@ dependencies {
     testImplementation "org.grails:grails-datastore-gorm-hibernate5"
     testImplementation "org.grails:grails-web-testing-support"
     testImplementation "org.grails.plugins:gsp"
-    testImplementation "org.grails.plugins:async"
+    testImplementation project(":grails-async-core")
```
matrei: Single quotes?
This dependency is changed from `plugin` to `core`. Is that deliberate?

matrei (changes requested):

davydotcom:
👍

davydotcom:
This merge makes sense to me. It very rarely changes outside of grails core. In that context alone I think it's fine to merge into the single mono repo to reduce complexity in release

build.gradle
jdaugherty: Given that you were okay removing the developer reference that was in Async, I reverted this to what it was before.  

I agree with you on composition.  I'll see if I can get this moved.  Thank you.

grails-async/core/build.gradle
jdaugherty: I missed those, I'd like to be consistent with grails-core.  For now, I'll remove it and then we can rework after the PR? Your comments on composition are 100% on point.  I think we need a second pass PR for some of that work though.

grails-test-examples/async-events-pubsub-demo/build.gradle
jamesfredley: We did land on double quotes for grails-forge so that all were the same, mostly to support setting a version that is not provided by grails-bom:

`"io.awspring.cloud:spring-cloud-starter-aws:$springCloudStarterAWSVersion"`

gradle/assemble.gradle
jdaugherty: Thiswas when I was troubleshooting the build directories, and I agree it should be reverted.

grails-test-examples/async-events-pubsub-demo/build.gradle
jdaugherty: We plan to merge the web testing support into this project, so I assume we can fix it then?  I've cleaned up the maven repos & changed the apply.   I'm going to leave the test support issue until we do that merge.

grails-test-suite-uber/build.gradle
jdaugherty: Not deliberate.  Good catch!

grails-test-suite-web/build.gradle
jdaugherty: Not deliberate, good catch!

grails-test-examples/async-events-pubsub-demo/build.gradle
jamesfredley: I think these blocks should be standardized to the following:

```
buildscript {
    repositories {
        maven { url = 'https://repo.grails.org/grails/core' }
        mavenCentral()
    }
    dependencies {
        classpath "org.grails:grails-gradle-plugin:${project['grails-gradle-plugin.version']}"
    }
}
```

based on https://github.com/grails/grails-gradle-plugin/issues/351#issuecomment-2537359841

And `apply plugin:` for our Grails Gradle plugins, since that works with snapshots and releases.  Only releases support id due the details in issue 351.

repo.grails.org is first, since we can see stats on https://repo.grails.org/ui/packages/gav:%2F%2Forg.grails:grails-gradle-plugin?name=grails-gradle-plugin&type=packages

build.gradle
jdaugherty: Per discussion, I made these consistent.  We need to do a second pass on cleaning this up with composition.  I'll leave that to you in a follow-up PR.

jdaugherty:
Per discussion with @jamesfredley  & @matrei , I have changed the bom scope to implementation from api.  We do not want any grails project exporting the bom as an API dependency, because all projects & grails apps should explicitly include the version.  Otherwise, you'll fight version resolution like @codeconsole has commented on in the slack channel.

jdaugherty:
After fixing the projectDir for the example grails application from the async/events, it's failing to build with this error: 

        Caused by: java.lang.NoSuchMethodError: 'org.springframework.asm.Type org.springframework.asm.Type.getMethodType(org.springframework.asm.Type, org.springframework.asm.Type[])'

It's failing when the gradle plugin is applied in the test project.  The gradle plugin depends on the grails-gradle-model, which is in core.  Because of this circular reference, to correctly test, I don't think a full grails project can live inside of core (unless we had a mono repo).  

I'm looking for solutions to this problem.

matrei:
@jdaugherty Add this to `grails-test-examples/async-events-pubsub-demo/build.gradle` and I think it will work:
```groovy

bootRun {
    mainClass = 'pubsub.demo.Application'
}

tasks.withType(Test).configureEach {
    useJUnitPlatform()
}
```

jdaugherty:
> @jdaugherty Add this to `grails-test-examples/async-events-pubsub-demo/build.gradle` and I think it will work:
> 
> ```groovy
> bootRun {
>     mainClass = 'pubsub.demo.Application'
> }
> 
> tasks.withType(Test).configureEach {
>     useJUnitPlatform()
> }
> ```

Adding these did not resolve the error:

                Caused by: java.lang.NoSuchMethodError: 'org.springframework.asm.Type org.springframework.asm.Type.getMethodType(org.springframework.asm.Type, org.springframework.asm.Type[])'
                at org.springframework.boot.loader.tools.MainClassFinder.<clinit>(MainClassFinder.java:61)
                at org.springframework.boot.gradle.plugin.ResolveMainClassName.findMainClass(ResolveMainClassName.java:145)
                at org.springframework.boot.gradle.plugin.ResolveMainClassName.resolveMainClassName(ResolveMainClassName.java:139)
                at org.springframework.boot.gradle.plugin.ResolveMainClassName.resolveAndStoreMainClassName(ResolveMainClassName.java:124)

codeconsole:
> grails-spring-security-core depends on them.

@matrei that's not a reason to merge async into core.  It's probably the opposite. IMO, It is very likely grails-spring-security-core plugin has no future.  Most of what it does works on deprecated functionality scheduled to be removed this year.  

Its current state doesn't even remotely match the existing filter chain.  I had to completely write a new plugin to properly lock down my application.

jdaugherty:
I may have found the cause for the test project:  https://github.com/hierynomus/license-gradle-plugin/issues/161

matrei:
> > grails-spring-security-core depends on them.
> 
> @matrei that's not a reason to merge async into core. It's probably the opposite. IMO, It is very likely grails-spring-security-core plugin has no future. Most of what it does works on deprecated functionality scheduled to be removed this year.
> 
> Its current state doesn't even remotely match the existing filter chain. I had to completely write a new plugin to properly lock down my application.

This was only intended as feedback to your question/statement that "most grails applications don't use" async & events.

build.gradle
```diff
@@ -7,7 +7,9 @@ buildscript {
         classpath "io.github.gradle-nexus:publish-plugin:$gradleNexusPublishPluginVersion"
         classpath "com.netflix.nebula:gradle-extra-configurations-plugin:$gradleExtraConfigurationsPluginVersion"
         classpath "com.bmuschko:gradle-nexus-plugin:$gradleNexusPluginVersion"
-        classpath "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:$gradleLicensePluginVersion"
+        classpath "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:$gradleLicensePluginVersion", {
+            exclude group: 'org.springframework', module: 'spring-core'
```
matrei, line 12: Add comment with link to the issue describing why it is excluded.

matrei (approved):

PR #13962, opened 2025-01-09
Updating events/async to support rxjava3

jdaugherty:
This just copies the existing design of rxjava2 and switches to the latest library.  Tests pass so I'm assuming this is correct.


matrei (approved):
Looking good!

jamesfredley (approved):

PR #13967, opened 2025-01-10
Ensure plugins are published to the same location as they were prior …

jdaugherty:
…to the move

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,6 +176,15 @@ subprojects { subproject ->
     ext.isCiBuild = subproject.hasProperty('isCiBuild') || System.getenv().get('CI') as Boolean
     ext.isAsyncProject = subproject.name.startsWith('grails-async') || subproject.name.startsWith('grails-events')
 
+    // Historically async mapped the core libraries to different artifact ids
+    ext.mappedArtifactIds = [
+            'grails-async-core'       : 'grails-async',
+            'grails-async-plugin'     : 'async',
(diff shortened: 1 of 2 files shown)


jdaugherty:
The grails-async plugin was publishing to a different location for it's plugins:  https://github.com/grails/grails-async/blob/4b45613bc1ae81607f44680087aebc1bb1c18d70/gradle/publishing.gradle#L67

jdaugherty:
This PR also fixes the artifact id mappings in the bom

jamesfredley (approved):

build.gradle
```diff
@@ -247,7 +256,9 @@ subprojects { subproject ->
                             }
                         }
                         if (isGrailsRepoPublish) {
-                            url "https://repo.grails.org/grails/libs-snapshots-local"
+                            url = isPlugin ?
+                                    uri('https://repo.grails.org/grails/plugins3-snapshots-local') :
+                                    uri('https://repo.grails.org/grails/libs-snapshots-local')
                         } else {
                             url 'https://maven.pkg.github.com/grails/grails-core'
```
matrei, line 263: @jdaugherty This should probably set the a `URI` with the `uri(Object path)` method (like the ones above). Also use setter syntax with `=`

jdaugherty: I had copied this from the async project, but will change it as part of the open PR.  thank you!

PR #13968, opened 2025-01-10
Remove grails gradle plugin usage since the plugins are not full grails apps

jdaugherty:
Grails Plugins that do not provide views (need the view compiler) or do not run independently have no need for the grails gradle plugins.  It's a better design to have a separate test project that imports the plugin to show the functionality.  

Another way to put this: the gradle plugin seems to only impact the build of the plugin and not what's exported.  If you only export the code via gradle & test externally it's not needed.  

Going forward, this can lead to a huge amount of cleanup since we aren't including the grails gradle plugin, which is then including grails-core again, and many other side effects.  It eliminates a key circular dependency that can cause unexpected behaviors and greatly simplifies our build files.  

It also helps us decide what plugins to apply.  For example, by not applying the grails gradle plugin, the dependency plugin doesn't get applied which gives us consistent resolution of dependencies.  

Finally, it's overhead we don't need in our build process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -260,7 +260,7 @@ subprojects { subproject ->
                                     uri('https://repo.grails.org/grails/plugins3-snapshots-local') :
                                     uri('https://repo.grails.org/grails/libs-snapshots-local')
                         } else {
-                            url 'https://maven.pkg.github.com/grails/grails-core'
+                            url('https://maven.pkg.github.com/grails/grails-core')
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


jamesfredley (approved):

matrei (approved):
Cool! And as long as the projects are in `grails-core` I don't think we will get much, if any, collateral damage.

jdaugherty:
I went ahead and fixed the ascii docs on this PR as well.  Properties are now goign to generate in the documentation again =)

jdaugherty:
@matrei  I made the uri() change per your other review.  

@jamesfredley I fixed the ascii doc generation.  I'm going to go ahead and merge so I can continue on the gsp updates.  Let me know if you would prefer a different solution.

PR #13971, opened 2025-01-14
Add graalvm versions to grails-bom, will be used in providedRuntime

jamesfredley:
to exclude jars from war and reduce size significantly.  

This is the first part of the following solution to reduce the war file size when using `asset-pipeline` by excluding the graalvm jars from the war using Gradle's War Plugin `providedRuntime` configuration scope.  Asset-pipeline-core requires the graalvm jars and some transitive dependencies during `compile`, `test` and `bootRun`, but not in the war.  Since they are needed during `bootRun` and Asset-pipeline-core is used in the non-grails Asset-pipeline distributions, using the war plugin scope is a direct path to solve this particular situation.  
 https://docs.gradle.org/current/userguide/war_plugin.html#sec:war_dependency_management

https://github.com/jamesfredley/asset-pipeline-small-jar-testing/commit/426cd6192bdb1c47be2c90298324c692a67b4e88#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R68


gradle.properties
```diff
@@ -47,6 +47,7 @@ converters.version=6.0.0-SNAPSHOT
 directory-watcher.version=0.18.0
 flying-saucer-pdf-openpdf.version=9.4.0
 geb-spock.version=7.0
+graal-sdk.version=24.1.1
```
jdaugherty, line 47: Does the sdk pull in any external dependencies?  I worry about including this because it's going to require more frequent updates so if it does, we could introduce conflicts.

jdaugherty:
What about the jar file?  Is the target to only exclude these from the war?

jamesfredley:
https://github.com/grails/grails-forge/issues/345#issuecomment-2565768893

https://github.com/bertramdev/asset-pipeline/issues/328

https://github.com/bertramdev/asset-pipeline/pull/352

jamesfredley:
@jdaugherty between the 3 graalvm dependencies there are are few transitive dependencies.  They are included in 7.0.0-M1 currently, since asset-pipeline-core 5.0.5 provides them.  The version of graalvm-sdk/js specified in this PR is newer then the version provided by asset-pipeline-core.    https://repo1.maven.org/maven2/com/bertramlabs/plugins/asset-pipeline-core/5.0.5/asset-pipeline-core-5.0.5.pom

Yes, the goal is the exclude these 3 and their transitive dependencies from the war, since it increases the base war size from ~85MB to > 140MB.

I also built an example project against asset-pipeline 5.0.4, which only exposes graal-sdk as a compile dependency https://repo1.maven.org/maven2/com/bertramlabs/plugins/asset-pipeline-core/5.0.4/asset-pipeline-core-5.0.4.pom using developmentOnly: https://github.com/jamesfredley/asset-pipeline-small-jar-testing/commit/ae3039054b19ae5fd26a7445f00a27a57102a644

Of those two, I prefer the version using `providedRuntime`, since asset-pipeline-core needs the three graal-sdk/js dependencies for the runtime `bootRun`, but not war execution and `providedRuntime` is the Gradle War Plugin's mechanism for excluding dependencies from the war.  The root issue:  https://github.com/grails/grails-core/issues/13913#issuecomment-2532078876.  Both approaches would benefit from this PR so that the graalvm versions live in grails-bom instead of the end application.

jdaugherty:
What about the forge cli project - won't this cause conflicts?  It uses graal and will likely be on the latest going forward / a different version than the asset pipeline?

jamesfredley:
I don't think forge CLI will be an issue, since it does not run the generated application, it just generates the files.

jdaugherty (approved):

codeconsole:
Shouldn't these versions really be managed by asset pipeline?  Aren't non-grails apps also affected by the same dependency resolution? 

It also seems to be environment specific? I have yet to experience this exception on any of my apps and I have been running asset pipeline `5.0.4`.   Seems overkill to add something for everyone that doesn't affect everyone.

jamesfredley:
@codeconsole If you want to try to eliminate the need for these dependencies on bootRun, https://github.com/grails/grails-core/issues/13913#issuecomment-2532078876 has the list of where to look.  

I am proceeding with the solution which reduces the war size against 5.0.5 for now and moving on to other priorities.

codeconsole:
It seems like this may have been overkill because the exception is only needing 

```
    implementation 'org.graalvm.polyglot:js:24.1.1' // avoid java.lang.ClassNotFoundException: org.graalvm.polyglot.Value when calling js file
    implementation 'org.graalvm.polyglot:polyglot:24.1.1' // avoid java.lang.ClassNotFoundException: org.graalvm.polyglot.Value when 
calling js file
```

https://github.com/bertramdev/asset-pipeline/pull/352#issuecomment-2520788853

PR #13972, opened 2025-01-15
Fixing issue with rxjava2/3 based Promise implementations not handling null

davydotcom:
It can be often the case that a Promise is used for the purposes of asynchronous operations in an isolated thread / persistence interceptor context. These async operations may not even care about a return type because they may be kicking off other error handling behaviors. In the case of the rxjava based implementations, rxjava gets angry when a return is null based and throws several exceptions: 

Example: 
```bash
RxPromiseSpec > Test promise null handling FAILED
    java.lang.NullPointerException: onSuccess called with null. Null values are generally not allowed in 2.x operators and sources.
        at io.reactivex.internal.operators.single.SingleCreate$Emitter.onSuccess(SingleCreate.java:65)
        at org.grails.async.factory.rxjava2.RxPromise$_closure1.doCall(RxPromise.groovy:47)
        at groovy.lang.Closure.call(Closure.java:433)
        at io.reactivex.internal.operators.single.SingleCreate.subscribeActual(SingleCreate.java:39)
        at io.reactivex.Single.subscribe(Single.java:3666)
        at io.reactivex.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
        at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:608)
        at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
        at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:840)

```

This PR resolves this by translating null to a Void object type which allows this to pass through and behave like other Promise implementations.

### DIFF
--- a/grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
+++ b/grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
@@ -56,9 +56,8 @@ class RxPromiseSpec extends Specification {
 
     void 'Test promise null handling'() {
 
-        when: 'a promise map is created'
+        when: 'a null promise result is created'
         def promise = Promises.createPromise {
-            sleep 1000
             return null
         }
         def result = promise.get()
--- a/grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
+++ b/grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
@@ -54,6 +54,19 @@ class RxPromiseSpec extends Specification {
             result == [one: 1, two: 2, four: 4]
     }
 
+    void 'Test promise null handling'() {
+
+        when: 'a promise map is created'
+        def promise = Promises.createPromise {
+            sleep 1000
+            return null
+        }
+        def result = promise.get()
+
+        then: 'result is void'
+        result == Void
+    }
+
     void 'Test promise list handling'() {
 
         when: 'a promise list is created from two promises'
--- a/grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
+++ b/grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
@@ -43,9 +43,8 @@ class RxPromiseSpec extends Specification {
 
     void 'Test promise null handling'() {
 
-        when: 'a promise map is created'
+        when: 'a null promise result is created'
         def promise = Promises.createPromise {
-            sleep 1000
             return null
         }
         def result = promise.get()
--- a/grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
+++ b/grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
@@ -41,6 +41,19 @@ class RxPromiseSpec extends Specification {
 
     }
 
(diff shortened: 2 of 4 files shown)


grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
```diff
@@ -54,6 +54,19 @@ class RxPromiseSpec extends Specification {
             result == [one: 1, two: 2, four: 4]
     }
 
+    void 'Test promise null handling'() {
+
+        when: 'a promise map is created'
+        def promise = Promises.createPromise {
+            sleep 1000
```
matrei: Do we need to sleep here?

grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
```diff
@@ -41,6 +41,19 @@ class RxPromiseSpec extends Specification {
 
     }
 
+    void 'Test promise null handling'() {
+
+        when: 'a promise map is created'
+        def promise = Promises.createPromise {
+            sleep 1000
```
matrei: Do we need to sleep here?

matrei (approved):
Nice! Added some comments on the tests.

grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
jdaugherty: In the past the sleeps have been used to ensure a delayed response.

grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
jdaugherty: In the past the sleeps have been used to ensure a delayed response.

jdaugherty (approved):
Should we remove rxjava1 since we have 2 + 3 now?

grails-async/rxjava3/src/test/groovy/org/grails/async/factory/rxjava3/RxPromiseSpec.groovy
davydotcom: let me fix that i copy pasted

davydotcom:
> Should we remove rxjava1 since we have 2 + 3 now?

rxjava1 is well beyond end of life, I believe. I'm open for deprecation and maybe future removal or just removal.

grails-async/rxjava2/src/test/groovy/org/grails/async/factory/rxjava2/RxPromiseSpec.groovy
davydotcom: resolved

matrei: I think only if you want to force a timout or need an event to not publish to fast.

PR #13973, opened 2025-01-15
Add `grails-web-taglib` to test dependencies

matrei:
This is a temporary fix to get the tests working while we sort out the right dependency scopes in `grails-gsp`.

### DIFF
--- a/grails-test-examples/async-events-pubsub-demo/build.gradle
+++ b/grails-test-examples/async-events-pubsub-demo/build.gradle
@@ -47,6 +47,8 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-logging'
     implementation 'org.springframework.boot:spring-boot-starter-tomcat'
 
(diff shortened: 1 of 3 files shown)


jdaugherty (changes requested):
Why not just set the scopes back & the imports back in the gsp project?  This work around will only help core, and not the other projects that are broken downstream.

matrei:
> Why not just set the scopes back & the imports back in the gsp project? This work around will only help core, and not the other projects that are broken downstream.

Yes, that is perhaps a better option :+1:

jdaugherty:
I did this here: https://github.com/grails/grails-gsp/pull/563/files .  We should still fix this the "right" way once we have time.  But this restores the dependency order as before (minus the `api "org.apache.groovy:groovy:$groovyVersion"` dependency and it keeps your test dependencies since those work)

matrei:
Used https://github.com/grails/grails-gsp/pull/563 instead.

PR #13990, opened 2025-01-30
Switch graalvm version from 24.1.1 to 23.0.7.  

### LINKED ISSUE #13983: 7.0.0-SNAPSHOT BUG - pom files are incorrectly being resolved in classLoader
### Bug

.pom files are now incorrectly being resolved in the classLoader and this is breaking resource resolution as there should only be jars

```
Failed to load manifest entries from jar file '/Users/user/.gradle/caches/modules-2/files-2.1/org.graalvm.js/js-community/24.1.1/3e874f113d724138c39287df69e5cfb7e5fcf700/js-community-24.1.1.pom': java.util.zip.ZipException: zip END header not found

java.util.zip.ZipException: zip END header not found
	at java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1649)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1657)
```

```groovy
new PathMatchingResourcePatternResolver(grailsApplication.classLoader).getResources("classpath*:messages_*.properties")
```

Simple Example App (1 line of code + gradle.properties change from forge create-app)
https://github.com/codeconsole/grails-resource-bug

This was broken AFTER the release of M1 as M1 works fine. 

### Version

7.0.0-SNAPSHOT

jamesfredley:
23.0.7 does not add the pom dependency.

https://repo1.maven.org/maven2/org/graalvm/js/js/23.0.7/js-23.0.7.pom

Resolves https://github.com/grails/grails-core/issues/13983


codeconsole:
Although this will fix the issue, I think this might mask a larger problem on why the error is occurring in my other project that is using 5.0.4 and does not have graalvm as a dependency.

PR #13995, opened 2025-02-04
Update and consolidate version for 10 active profiles

jamesfredley:
The 10 active profiles have been consolidated into https://github.com/grails/grails-profiles and share the same version now

removes `grails-bom/profiles.properties` and sets version for all 10 active profiles in `gradle.properties`


matrei (approved):
Nice to get these synced up! LGTM 👍

Issue #13557, opened 2024-07-17
Move grails-functional-test into grails-core


sdelamo:
discussion: https://github.com/orgs/grails/discussions/13516
repo: https://github.com/grails/grails-functional-tests

PR #14007, opened 2025-02-09
#13557 - Move functional tests to grails-core

jdaugherty:
#13557 - Moves the functional tests to grails-core under the grails-test-examples project.  

While merging this I found the grails-dependencies project was incorrectly updated in Grails 4 as part of the gradle scope updates.  Maven scopes do not include a scope of 'api'.  I've gone ahead and converted this to a gradle project that publishes the correct pom.  This allows test projects to use the grails-dependencies project directly as well.

You can set the gradle property 'skipFunctionalTests' to not run this locally.  You can also use the gradle property 'onlyFunctionalTests' to only run the functional tests.


jdaugherty:
This PR restores the functionality of the `grails-dependencies` project.  I'm skeptical it's been used in several years because of how it was broken, but this is really the equivalent to the spring starter 'org.springframework.boot:spring-boot-starter-web' ... where it adds reasonable dependencies by default so that you can define a grails application with just: 

     dependencies {
          implementation "org.grails:grails-bom:${grailsVersion}"
          implementation 'org.grails:grails-dependencies'
     }

I suspect the original code was pom  based so that it wouldn't create a circular dependency.  We've shown that we can pull those versions from the bom so I think this is a less of a concern now and being gradle based gives us checking so this doesn't break in the future.  

We may want to rename this project in the longer term to reflect that it's really a starter.  I also think we should expand on it with everything but non-datamapping dependencies and what we typically release as part of the `grails-core` major release workflow.

jamesfredley (approved):
The following commits look good:

https://github.com/grails/grails-core/pull/14007/commits/e37c73266881052829deb115ff26c51a893f05fe

https://github.com/grails/grails-core/pull/14007/commits/d42660ac2b7e265260401047f96f0d939119c498

https://github.com/grails/grails-core/pull/14007/commits/cc03c8358bf90d61dabbc06dedc07280b6f78fb2

https://github.com/grails/grails-core/pull/14007/commits/651bf5a2a8823e8d3f1ada2c8dae43a3ee68f9dd

matrei (approved):
Nicely solved! ⭐

PR #14043, opened 2025-02-28
Change closure parameter name to parentName so it is distinct from class field name

jamesfredley:
Change closure parameter `name` to `parentName` so it is distinct from class field `name`

Exception now communicates the correct relationship between child and parent profiles
It was displaying the parent profile name twice


jamesfredley:
Helps with debugging https://github.com/grails/grails-gradle-plugin/issues/411

jdaugherty (approved):

PR #14075, opened 2025-03-19
Matrei/rename proposal

matrei:
Following my discussion with @jdaugherty, I've created a new branch with my renaming ideas.


RENAME.md
```diff
@@ -0,0 +1,136 @@
...
+| org.grails                         | tck-base                                        | org.apache.grails          | grails-data-mapping-tck-base                |                        |                          | grails-data-mapping    |
+| org.grails                         | tck-domains                                     | org.apache.grails          | grails-data-mapping-tck-domains             |                        |                          | grails-data-mapping    |
+| org.grails                         | tck-tests                                       | org.apache.grails          | grails-data-mapping-tck-tests               |                        |                          | grails-data-mapping    |
+| org.grails                         | grails-datastore-gorm-hibernate6                | org.apache.grails          | grails-datastore-gorm-hibernate6            |                        |                          | grails-data-hibernate6 |
```
jdaugherty: I think in your example, new artifactId should be 'grails-data-mapping-hibernate6'?

RENAME.md
```diff
@@ -0,0 +1,136 @@
...
+| org.grails.plugins                 | hibernate6                                      | org.apache.grails          | grails-data-hibernate6                      |                        |                          | grails-data-hibernate6 |
+| org.grails.plugins                 | cache                                           | org.apache.grails          | grails-cache                                |                        |                          | grails-cache           |
+| org.grails.plugins                 | geb                                             | org.apache.grails          | grails-geb                                  |                        |                          | grails-geb             |
+| org.grails.plugins                 | spring-security-acl                             | org.apache.grails          | grails-spring-security-acl                  |                        |                          | grails-spring-security |
```
jdaugherty: Should we drop "spring" from the security names?  It seems redundant since we only run on SpringBoot now.

matrei: Yes, we could drop it, but as there are other security frameworks (for example: https://shiro.apache.org/), I thought it could be beneficial to keep it, to make clear what this is based on. I'm not sure how you mean with "since we only run SpringBoot now"?

jamesfredley:
https://github.com/apache/grails-core/pull/14077 some additional changes on top of these

RENAME.md
jamesfredley: I lean slightly towards `spring-security-` for clarity.

jdaugherty: I'd vote keep the spring security then.  Thank you for listening.

matrei:
Replaced by https://github.com/apache/grails-core/pull/14080

PR #14080, opened 2025-03-20
artifact rename proposal

jdaugherty:
High level feedback on the proposal: 
* `org.apache.grails` will be used for any plugin traditionally included by an end-user in their grails or spring app.  
* group ids for implementation details or dependencies of a grails plugin will use a subpackage.  
    * i.e. the codecs plugin would have a group of `org.apache.grails` and the codecs-api that the plugin includes would have a group of `org.apache.grails.codecs`
* testing artifacts for end-user grails applications will have a special rule to be prefixed with `grails-testing-` to ensure people are not including them in a regular application.
* gradle plugins will be consolidated under the repo https://github.com/apache/grails-gradle-plugin. There will be one artifact published that can be included in a build script / source to import all of the gradle plugins.  That artifact will be: `org.apache.grails:grails-gradle-plugins` 
* because profiles are not an artifact distributed to an application, they will not be prefixed with a "grails-"
* artifacts should be prefixed with `grails-` (unless a profile)
* json => gson to reflect the view technology and to avoid confusion with other json implementations
* `views` is dropped from the artifact id to be consistent with view technologies (gsp, gson, markup)
* several artifacts are renamed for clarity, for example `grails-console` -> `grails-gradle-console` 
* we have appended '-cli' where a cli is being provided
* for plural vs singular on package names / artifacts / groups we have decided on plural usage, while single in class usage.  This matches the existing code base for projects like events


jamesfredley (approved):
Reviewed: https://github.com/jdaugherty/grails-core/blob/3b4da92e339b7b688011e4a8ffc3c2aaea723680/RENAME.md?plain=1 iteration

I am still a little concerned with the artifacts being security-spring and the project name being spring-security.  Matching the upstream project seems like the best idea https://github.com/spring-projects/spring-security.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | grails-async-rxjava2                            | org.apache.grails.async                         | grails-async-rxjava2                                          |                        |                        | grails-core            |
+| org.grails                        | grails-async-rxjava3                            | org.apache.grails.async                         | grails-async-rxjava3                                          |                        |                        | grails-core            |
+| org.grails                        | grails-bom                                      | org.apache.grails                               | grails-bom                                                    |                        |                        | grails-core            |
+| org.grails                        | grails-bootstrap                                | org.apache.grails.bootstrap                     | grails-bootstrap                                              |                        |                        | grails-core            |
```
rainboyan, line 36: Why using nested group here? only one in bootstrap?

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails.plugins                | cache                                           | org.apache.grails                               | grails-cache                                                  |                        |                        | grails-cache           |
+| org.grails                        | grails-plugin-codecs                            | org.apache.grails                               | grails-codecs                                                 |                        |                        | grails-core            |
+| org.grails                        | grails-codecs                                   | org.apache.grails.codecs                        | grails-codecs-core                                            |                        |                        | grails-core            |
+| org.grails                        | grails-console                                  | org.apache.grails                               | grails-gradle-console                                         |                        |                        | grails-core            |
```
rainboyan: grails-console provides the console and shell extended Groovy Console and Shell, can be used for Gradle task, but also can be used in Shell command.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails.plugins                | converters                                      | org.apache.grails                               | grails-converters                                             |                        |                        | grails-core            |
+| org.grails                        | grails-core                                     | org.apache.grails                               | grails-core                                                   |                        |                        | grails-core            |
+| org.grails.plugins                | hibernate5                                      | org.apache.grails                               | grails-data-hibernate5                                        |                        |                        | grails-data-mapping    |
+| org.grails.plugins                | database-migration                              | org.apache.grails                               | grails-data-hibernate5-dbmigration                            |                        |                        | grails-data-mapping    |
```
rainboyan, line 45: Database migration is a more general concept, I think it can be implemented in the future on NoSQL databases such as MongoDB. From a user perspective, I want to be as consistent as possible in the development experience.
Currently, only Hibernate/JDBC is supported, but what about the future?

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | gorm-hibernate5-spring-boot                     | org.apache.grails                               | grails-data-hibernate5-spring-boot                            |                        |                        | grails-data-mapping    |
```
rainboyan, line 46: I'm thinking about why not create a grails-spring-boot repository to include all the modules that support Spring Boot, including Hibernate, MongoDB, gsp, and other views.

RENAME.md
jdaugherty: bootstrap isn't used in an end user grails app - we said only libraries that are would use 'org.apache.grails'.

Do you have a better name suggestion than 'org.apache.grails.bootstrap'?

RENAME.md
jdaugherty: Aren't both launched by gradle though?

RENAME.md
jdaugherty: We currently use liquidbase for the hibernate 5 migrations.  We are certainly open to more but people would need to contribute those. I assume the name is ok for you?  Its clear its hibernate5 this way.

RENAME.md
jdaugherty: There actually used to be one and the library went unmaintained.  The largest reason we have consolidated libraries to keep them maintained as spring, etc are upgraded. This PR is only for the group or artifact names though.  We can revisit the repo names on a separate thread.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | tck-base                                        | org.apache.grails.data                          | grails-data-mapping-tck-base                                  |                        |                        | grails-data-mapping    |
+| org.grails                        | tck-domains                                     | org.apache.grails.data                          | grails-data-mapping-tck-domains                               |                        |                        | grails-data-mapping    |
+| org.grails                        | tck-tests                                       | org.apache.grails.data                          | grails-data-mapping-tck-tests                                 |                        |                        | grails-data-mapping    |
+| org.grails                        | grails-gorm-testing-support                     | org.apache.grails.testing                       | grails-data-mapping-testing-support                           |                        |                        | grails-data-mapping    |
```
rainboyan: I don't think it's a good idea to use different groupIds in the same repository.

rainboyan: I don't think it's a good idea to use different groupIds in the same repository.

jdaugherty: Because of the circular dependencies, I don't think we can realistically adhere to keeping the same package for an entire repo.  We have intentionally brought code together in 7 because we found so much of it wasn't being tested.  We may even be forced to bring more repositories together because of the requirements of the ASF release process.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | grails-shell                                    | org.apache.grails                               | grails-shell-cli                                              |                        |                        | grails-core            |
+| org.grails                        | grails-plugin-sitemesh3                         | org.apache.grails                               | grails-sitemesh3                                              |                        |                        | grails-views           |
+| org.grails                        | grails-spring                                   | org.apache.grails                               | grails-spring                                                 |                        |                        | grails-core            |
+| org.grails.plugins                | spring-security-acl                             | org.apache.grails                               | grails-security-spring-acl                                    |                        |                        | grails-spring-security |
```
rainboyan: How about changing repository to: grails-security, 
and using group org.apache.grails.security?
-- org.apache.grails.security.spring (package name)
---- grails-security-spring-acl
---- grails-security-spring-cas
-- org.apache.grails.security.shiro (package name)
---- grails-security-shiro

RENAME.md
rainboyan: Actually, I have rewrote Gradle shell and console tasks to Shell Command, in order to streamline the GrailsGradlePlugin, while also keeping the use of grails commands consistent.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | grails-plugin-domain-class                      | org.apache.grails                               | grails-domain-class                                           |                        |                        | grails-core            |
+| org.grails                        | grails-encoder                                  | org.apache.grails.view                          | grails-encoder                                                |                        |                        | grails-core            |
+| org.grails.plugins                | events                                          | org.apache.grails                               | grails-events                                                 |                        |                        | grails-core            |
+| org.grails                        | grails-events                                   | org.apache.grails.event                         | grails-events-core                                            |                        |                        | grails-core            |
```
rainboyan: I think using org.apache.grails.events is better, and the artifacts also have -events prefix.

RENAME.md
matrei: Some other suggestions:
- `org.apache.grails:grails-groovy-console`
- `org.apache.grails:grails-swing-console`

(For anyone who wonders why renaming at all, it is to differentiate it from the `grails-web-console`.)

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails                        | grails-plugin-mimetypes                         | org.apache.grails                               | grails-mimetypes                                              |                        |                        | grails-core            |
+| org.grails.plugins                | quartz                                          | org.apache.grails                               | grails-quartz                                                 |                        |                        | grails-quartz          |
+| org.grails.plugins                | grails-redis                                    | org.apache.grails                               | grails-redis                                                  |                        |                        | grails-redis           |
+| org.grails                        | grails-plugin-rest                              | org.apache.grails                               | grails-rest-responder                                         |                        |                        | grails-core            |
```
rainboyan: `RestResponder` is a trait. I think just call grails-rest, or grails-resources because `@Resource` annotation and transformation to domain class. I remember that we have resources plugin before. That can be a misunderstanding,  I know that Rails developers often use this as such.

RENAME.md
rainboyan: Maybe merging the web-console into grails-console, just like Spring Boot's H2Console.

matrei: I see now that the `shell` task, `grails.ui.shell.GrailsShell`, is also part of the `grails-console` library, so maybe my previous suggestions won't work.

> Maybe merging the web-console into grails-console, just like Spring Boot's H2Console.

That is an interesting idea. Currently the `grails-web-console` is located here http://github.com/grails-plugins/grails-web-console and not moved to Apache.

Maybe we need a new more general name that would capture what all these features have in common (comparison Laraval Tinker), for example:
- `grails-anvil`
- `grails-bench`
- `grails-playground`
- `grails-lab`
- `grails-smith`
- `grails-whiz`
- `grails-workshop`

rainboyan: Yes, I've also considered the Database Console and Admin Console. I've implemented the Admin Console, and Django has this as well. Rails has the `dbconsole` command.

RENAME.md
matrei: Rest Responder is actually the name of the plugin: [RestResponderGrailsPlugin](https://github.com/apache/grails-core/blob/7.0.x/grails-plugin-rest/src/main/groovy/org/grails/plugins/web/rest/plugin/RestResponderGrailsPlugin.groovy)

Maybe one of these would be more descriptive:
- `org.apache.grails:grails-rest-transforms`
- `org.apache.grails:grails-rest-support`

RENAME.md
matrei: For me "Admin" connotates to a security role or to changing settings. Database Console would collide with the H2 Database Console.

RENAME.md
matrei: > I don't think it's a good idea to use different groupIds in the same repository.

@rainboyan Why don't you think it's a good idea? (We will use this practice extensively in the Grails Core repository.)

RENAME.md
jdaugherty: lets change it back to just console then?  

I am against merging it with the web console because I have had security auditors request it be turned off - we actually turn it off by excluding the dependency for any production based build.  This was a requirement to ensure limited / non-admin access as part of SOX compliance.

RENAME.md
jdaugherty: rest support could confuse it with the spring security rest plugin, transform would be a better name

RENAME.md
matrei: > bootstrap isn't used in an end user grails app

It is however part of `grails-dependencies`? Does that indicate that it is an end user grails app dependency?

jdaugherty: Its used to setup the minimum dependencies- and probably shouldn't be a part of grails dependencies.

RENAME.md
matrei: Are you saying you need `grails-console` as a dependency in production apps? If not, would not excluding `grails-console` also exclude the `grails-web-console` feature, if they were merged?

RENAME.md
matrei: > rest support could confuse it with the spring security rest plugin

Hmm :thinking: , why would `grails-rest-support` confuse it with the spring security rest plugin (more than any other name including rest)?

RENAME.md
matrei: On the other hand `grails-core` is allowed in the `org.apache.grails` groupId.
Could `org.apache.grails.core` be a groupId for both `grails-core` and `grails-bootstrap`?
Or could both live in `org.apache.grails`?

jdaugherty: grails-core is included in end apps and we said we would use just 'org.apache.grails' for those cases

RENAME.md
jdaugherty: The agreement was to remove Plugin / starter conventions and replace with a single root group id "org.apache.grails"

RENAME.md
jdaugherty: I'm saying they are separate functions and we should not combine the gradle projects.  You would be including an admin function by default, which isn't considered a secure practice.

RENAME.md
jdaugherty: Because when I think of rest, I think of the oauth workflow that exists in the spring-security-rest plugin. 

`grails-rest-transformers` prevents that confusion in my mind, and since it was one of the suggested items seems like a reasonable choice =)

RENAME.md
rainboyan: I haven't seen this practice in other open source projects before. I don't think this solves the circular dependency problem.

jdaugherty: I'm saying that the circular dependency problem is slowly being solved by collapsing the repos.  Meaning, there will be several gradle projects in the same repo going forward.  For example, grails-core.

jdaugherty: Spring uses an artifact prefixes to distinguish the purpose, but since we're removing `Plugin` from the name we agreed on using the group id instead of using prefixes.

RENAME.md
rainboyan: Maybe I missed the previous discussion, but is this convention a well-known rule, or is it just a common understanding among Grails developers? This makes it difficult for me to understand.

rainboyan: Maybe I missed the previous discussion, but is this convention a well-known rule, or is it just a common understanding among Grails developers? This makes it difficult for me to understand.

jdaugherty: In the discussion for this PR on the mailing list, we made this proposal so we could move forward with removing `plugin` from the names.

RENAME.md
```diff
@@ -0,0 +1,150 @@
...
+| org.grails.grails-publish         | org.grails.grails-publish.gradle.plugin         | org.apache.grails.gradle.grails-publish         | org.apache.grails.gradle.grails-publish.gradle.plugin         | grails-publish         | grails-publish         | grails-gradle-plugin   |
+| org.grails.plugins.views-json     | org.grails.plugins.views-json.gradle.plugin     | org.apache.grails.gradle.grails-json            | org.apache.grails.gradle.grails-gson.gradle.plugin            | views-json             | grails-gson            | grails-gradle-plugin   |
+| org.grails.plugins.views-markup   | org.grails.plugins.views-markup.gradle.plugin   | org.apache.grails.gradle.grails-markup          | org.apache.grails.gradle.grails-markup.gradle.plugin          | views-markup           | grails-markup          | grails-gradle-plugin   |
+| org.grails.grails-web             | org.grails.grails-web.gradle.plugin             | org.apache.grails.gradle.grails-web             | org.apache.grails.gradle.grails-web.gradle.plugin             | grails-web             | grails-web             | grails-gradle-plugin   |
```
rainboyan, line 156: Does grails-web plugin still extend from grails-app? So what's the difference between them? Should use both of them in a project?

RENAME.md
jamesfredley: For the conversation around console consolidation, we should discuss that agilely after the groupid/artifactid rename, which allows us to get snapshots working at ASF and is a high priority.

https://github.com/apache/grails-core/issues/14082

RENAME.md
rainboyan: If I must have a choice between using `plugin` suffix and using the root groupId `org.apache.grails` to distinguish between plugins and libraries, I will choose the first one. In Grace, I have always used `plugin` as the archiveClassifier of the plugin jars.
For example, fields-6.0-plugin.jar, cache-6.0-plugin.jar, So, it's easy for us to know that these are plugins.

RENAME.md
jamesfredley: Having done the upgrade work on grails-data-hibernate5-dbmigration for Grails 7 it became clear that hibernate6 will be  a distinct plugin, grails-data-hibernate6-dbmigration, and hibernate7 will likely be also.  

If https://github.com/liquibase/liquibase-mongodb was used to support mongodb I think that plugin should be named grails-data-mongodb-dbmigration.

If we do go down this path it is possible some some common code could be pulled from the plugins and shared.  Let's put this into an issue on grails-data-mapping, if we want to proceed.  I do not believe this impacts the immediate need for a decision on groupid/artifactid so we can get snapshots working at ASF.

RENAME.md
jamesfredley: With the mass consolidate of repositories prior to the move to ASF and potential for further consolidation, which was necessary to reduce release time, there will need to be different groupIds in the same repository.

For this specific instance if `org.apache.grails.testing` was changed to `org.apache.grails.data.testing`, does that address the concern?

RENAME.md
matrei: OK, thanks for the clarification. I think `grails-rest-transforms` works. We do have a `grails-events-transform` (with singular form) also. Should we try and align their names?

RENAME.md
matrei: Yes, `org.grails.grails-web` extends the base `GrailsGradlePlugin` (in this document renamed from `org.grails.grails-core` to `org.grails.grails-app` and you don't need to apply both. (I don't think I have ever seen the `org.grails.grails-core` Gradle Plugin applied in a Grails project.)

However, I think there are discussions to restructure this so that the Gradle Plugins do not extend.

RENAME.md
jdaugherty: I've updated the PR for this.

RENAME.md
jdaugherty: I've reverted the name change & James has opened a ticket to address in the future.

RENAME.md
jdaugherty: What's ironic is the only thing that grails-web provides you is the gradle task 'urlMappingsReport'.  Everything else is on the base.

RENAME.md
rainboyan: What do you think about these two libraries, I think both are easy to understand than the above one.

org.apache.grails.testing:grails-testing-data-mapping-support 

vs 

org.apache.grails.data:grails-data-mapping-testing-support

RENAME.md
rainboyan: I think we should move some task configurations down to GrailsWebGradlePlugin, such as Console, Shell, ApplicationCommands, runCommand, runScript.
These are not needed for Plugin projects.

RENAME.md
jdaugherty: We did not realize this until after we refactored all of the build files, but one of the largest problems was including test dependencies on the production classpath of many grails apps.  One of the reasons I pushed to have a `.testing` namespace was to ensure that doesn't happen going forward. 

In your example, we have several of these testing support libraries: 
grails-views-gson-testing-support
grails-testing-support
grails-web-testing-support
grails-datamapping-testing-support

Shouldn't we follow the pattern on all of these then?  Would this work for you? 
org.apache.grails.testing:grails-testing-support-views-gson
org.apache.grails.testing:grails-testing-support
org.apache.grails.testing:grails-testing-support-web
org.apache.grails.testing:grails-testing-support-datamapping

jdaugherty: I spoke to @matrei and I'm willing to relax the `.testing` group for consistency if we prefix like this.  I'm going to go ahead and update this pr.  I'll shorten "testing-support" to "testing" though.

jdaugherty: This is now updated, @rainboyan can you please confirm this is acceptable?

RENAME.md
jdaugherty: Some plugins are self-contained grails apps.  I'll start a separate thread on the gradle restructure once we get 7 out the door?  https://github.com/apache/grails-gradle-plugin/issues/389 is likely going to be the ticket to document that work.

RENAME.md
rainboyan: Now we have two, but it is impossible to distinguish from the name

- org.apache.grails.testing:grails-test
- org.apache.grails:grails-testing

jdaugherty: .testing is the nested group, and is not meant to be included externally.  It is an api dependency of grails-testing.  There is a common strategy in the grails code base where the plugin & the code is separated into separate projects.  Sometimes we add the `-core` to the end in these cases.  I'll update grails-test to be grails-test-core to make it more clear.

jdaugherty: The PR is updated with grails-testing-core to follow the pattern we've used else where.

RENAME.md
```diff
@@ -0,0 +1,172 @@
...
+2. Any group that is further nested is an implementation detail of Grails.  While users can choose to override them, they are typically not meant to be included in a Grails app directly.
+3. Testing artifacts for Grails applications / plugins, will have the prefix `grails-testing-` to remind developers these artifacts should not generally be included in the `implementation` scope of their application.
+4. Gradle plugins will be consolidated under a single [repository](https://github.com/apache/grails-gradle-plugin).
+5. As part of the gradle plugin consolidation, we adopt consistent names and remove `views-` from the views gradle plugins. Instead they will match the existing gradle plugin naming pattern - i.e. `grails-gsp`, `grails-markup`, etc.
+5. There will be one artifact published that can be included in either a `buildSrc/build.gradle` or a `buildScript` to import all Grails gradle plugins.  That plugin will be `org.apache.grails:grails-gradle-plugins`.
+6. Grails artifacts should be prefixed with `grails-` with the exception of profiles.  
+7. Profiles will not use a `grails-` prefix because they are distributed with an application.
```
matrei: Should probably be "... are not distributed ..."

RENAME.md
rainboyan: bootstrap will be included in the end apps, and most of other modules depend on it.

RENAME.md
rainboyan: +1

RENAME.md
rainboyan: Agree with @jamesfredley

RENAME.md
jdaugherty: by used, I mean it won't be directly included by the end user.  It will be exposed by the api dependency.

RENAME.md
rainboyan: +1

RENAME.md
rainboyan: +1

rainboyan: > Spring uses an artifact prefixes to distinguish the purpose, but since we're removing Plugin from the name we agreed on using the group id instead of using prefixes.

Until a better solution comes out, I still recommend not changing.

RENAME.md
rainboyan: Self contained plugins are still applied SpringBootPlugin, but bootRun/bootJar tasks are disabled default.

matrei:
This is what a default generated web-app (from the current state of `grails-forge-cli`) dependencies blocks will look like with this proposal. I think it looks pretty good. Note the outlier `org.apache.grails.web:grails-web-boot`:
```groovy
buildscript {
    dependencies {
        classpath "com.bertramlabs.plugins:asset-pipeline-gradle"
        classpath platform("org.apache.grails:grails-bom:$grailsVersion")
        classpath "org.apache.grails:grails-gradle-plugins"
        classpath "org.apache.grails:grails-data-hibernate5"
    }
}

apply plugin: "asset-pipeline"
apply plugin: "org.apache.grails.gradle.grails-web"
apply plugin: "org.apache.grails.gradle.grails-gsp"

dependencies {
    profile "org.apache.grails.profiles:web"
    developmentOnly "org.webjars.npm:bootstrap"
    developmentOnly "org.webjars.npm:bootstrap-icons"
    developmentOnly "org.webjars.npm:jquery"
    implementation platform("org.apache.grails:grails-bom:$grailsVersion")
    implementation "org.apache.grails:grails-core"
    implementation "org.apache.grails:grails-logging"
    implementation "org.apache.grails:grails-databinding"
    implementation "org.apache.grails:grails-i18n"
    implementation "org.apache.grails:grails-interceptors"
    implementation "org.apache.grails:grails-rest-transforms"
    implementation "org.apache.grails:grails-services"
    implementation "org.apache.grails:grails-url-mappings"
    implementation "org.apache.grails.web:grails-web-boot"
    implementation "org.apache.grails:grails-gsp"
    implementation "org.apache.grails:grails-data-hibernate5"
    implementation "org.apache.grails:grails-scaffolding"
    implementation "org.springframework.boot:spring-boot-autoconfigure"
    implementation "org.springframework.boot:spring-boot-starter"
    implementation "org.springframework.boot:spring-boot-starter-actuator"
    implementation "org.springframework.boot:spring-boot-starter-logging"
    implementation "org.springframework.boot:spring-boot-starter-tomcat"
    implementation "org.springframework.boot:spring-boot-starter-validation"
    console "org.apache.grails:grails-console"
    runtimeOnly "com.bertramlabs.plugins:asset-pipeline-grails"
    runtimeOnly "com.h2database:h2"
    runtimeOnly "com.zaxxer:HikariCP"
    runtimeOnly "org.fusesource.jansi:jansi"
    integrationTestImplementation testFixtures("org.apache.grails:grails-geb")
    testImplementation "org.apache.grails:grails-testing-support-gorm"
    testImplementation "org.apache.grails:grails-testing-support-web"
    testImplementation "org.spockframework:spock-core"
}
```

PR #14084, opened 2025-03-23
Artifact rename proposal

rainboyan:
# Grails Repackaging

As part of the move to the Apache Software Foundation, the grails artifacts must adhere to the apache namespace.

## Package Rules

The Grails developers have decided on several rules regarding packaging. These rules are enumerated here:

1. In general, after the migration of Grails to Apache, the group names are changed from `org.grails` `org.grails.plugins` to `org.apache.grails` `org.apache.grails.plugins`. For those who are still using Grails 5/6, migrating to 7.0 is very easy. There are also some sub-projects, there are multiple implementations, so we use subgroups to distinguish more clearly. For example, grails-async uses `org.apache.grails.async`, grails-events uses `org.apache.grails.events`, grails-data uses `org.apache.grails.data`, grails-views uses `org.apache.grails.views`, and grails-spring-security uses `org.apache.grails.spring`. ~~For Grails applications, the group used should always be `org.apache.grails`.~~
2. Grails plugins are divided into core plugins and community plugins, the core plugins are mainly concentrated in the `grails-core` repository, and some community plugins include grails-cache, grails-fields, grails-geb, and grails-scaffolding. Core plugins are often prefixed with `grails-plugin-`, such as `grails-plugin-codecs`, but community plugins are generally not specified, and some use `-plugin` as a suffix, but plugin can be added as `archiveClassifier` to the artifact when packaging, for example, `grails-cache` will be packaged as `grace-cache-7.0.0-plugin.jar`. In the `org.apache.grails.plugins` group, currently, the development team maintains several plugins, including `grails-cache` `grails-fields` `grails-geb` `grails-quartz` `grails-sitemesh3` `grails-redis`. ~~Any group that is further nested is an implementation detail of Grails. While users can choose to override them, they are typically not meant to be included in a Grails app directly.~~
3. `grails-test` will be changed to `grails-testing-api`, it includes the origin mock test apis and some core api from `grails-testing-suppport`, but `grails-testing-suppport` will be an aggregate that will depend on the other modules testing-support, including `grails-web-testing-support` `grails-gorm-testing-support`, `grails-views-testing-support`, so you only need to add `grails-testing-support` to your project when using it. ~~Testing artifacts for Grails applications / plugins, will have the prefix `grails-testing-support-` to remind developers these artifacts should not typically be included in the `implementation` scope of their application.~~
4. Gradle plugins will be consolidated under a single [repository](https://github.com/apache/grails-gradle-plugin). Gradle plugin `grails-web` will be changed to `grails-app`, just as we have `grails-plugin` for Grails plugins, and we use `create-app` command to create a webapp, use `create-plugin` to create plugin project.
5. As part of the gradle plugin consolidation, we adopt consistent names and remove `views-` from the views gradle plugins. Instead they will match the existing gradle plugin naming pattern - i.e. `grails-gsp`, `grails-markup`, etc.
5. There will be one artifact published that can be included in either a `buildSrc/build.gradle` or a `buildScript` to import all Grails gradle plugins. That plugin will be `org.apache.grails:grails-gradle-plugin`.
6. Grails artifacts should be prefixed with `grails-` with the exception of profiles.
7. Profiles will not use a `grails-` prefix because they are not distributed with an application.
8. Because there are many modern JSON implementations, and Grails may choose to implement them in the future, we will refer to our existing `.gson` implementation as `gson` instead of `json`.
9. We have appended `-cli` where a CLI is being provided, such as `grails-forge-cli`, but it is only used to generate new Grails project, not distributed with an application. `grails-shell` is still the main cli for developers to use to generate Grails domains, controllers, services, views, in the future, the `create-*` commands will be replace by `generate` and `destroy` commands from the new Code Generation API, you also can privide your own generators in your plugins.
10. Several artifacts have been renamed for clarity. In general we use the suffix `-core` to signify a core of the framework~~project~~, such as Aysnc and Events framework, they have many different implements, you can choose which one to use in your app. ~~and drop the suffix on the package that the end user will use~~.
11. Package names, artifact ids, and group ids will use the plural form. Implementing classes will use the singular form.
    * For example: `org.apache.grails.events:grails-events` with the `EventBusPlugin`


rainboyan:
Replaced by #14080

PR #14092, opened 2025-03-28
Add support for HTTP 307/308 redirect status codes

### LINKED ISSUE #11625: [Feature Request]Making it possible to redirect with 307 and 308 status
In some cases, I want to backward compatibility with old API. So I should use redirect to redirect old URL to the new one.

And I think the http code [307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) and [308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) would be better.

But grails `redirect` is only support 301 and 302 for now: https://docs.grails.org/latest/ref/Controllers/redirect.html

So I think it's better to make possible to redirect with 307 and 308.

dakshmehta007:
Added support for HTTP 307/308 status codes in the ResponseRedirector by:

Fixes #11625 

1. Extending the redirect logic to check for a `tempRedirect` flag in the statusConfig
2. Using appropriate status codes based on permanent + tempRedirect flags:
   - permanent: false, tempRedirect: false -> 302 (default)
   - permanent: true, tempRedirect: false -> 301
   - permanent: false, tempRedirect: true -> 307
   - permanent: true, tempRedirect: true -> 308

3. Added test cases to verify both temporary (307) and permanent (308) redirect behavior

## Usage##
```groovy
**For 307 Temporary Redirect**
request.setAttribute('statusConfig', [tempRedirect: true])
redirect(action: 'foo')

**For 308 Permanent Redirect**
request.setAttribute('statusConfig', [tempRedirect: true])
redirect(action: 'foo', permanent: true)

### DIFF
--- a/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
+++ b/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
@@ -54,13 +54,11 @@ class RedirectController {
     }
 
     def toActionTemporaryRedirect() {
-        request.setAttribute('statusConfig', [tempRedirect: true])
-        redirect(action:'foo')
+        redirect(action:'foo', temporary:true)
     }
 
     def toActionPermanentRedirect() {
-        request.setAttribute('statusConfig', [tempRedirect: true])
-        redirect(action:'foo', permanent: true)
+        redirect(action:'foo', permanent:true, temporary:true)
     }
 
     def toRoot() {
--- a/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
+++ b/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
@@ -53,6 +53,16 @@ class RedirectController {
         redirect(action:"${prefix}oo")
     }
 
+    def toActionTemporaryRedirect() {
+        request.setAttribute('statusConfig', [tempRedirect: true])
+        redirect(action:'foo')
+    }
+
+    def toActionPermanentRedirect() {
+        request.setAttribute('statusConfig', [tempRedirect: true])
+        redirect(action:'foo', permanent: true)
+    }
+
     def toRoot() {
         redirect(controller:'default')
     }
--- a/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
+++ b/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
@@ -61,6 +61,14 @@ class RedirectController {
         redirect(action:'foo', permanent:true, temporary:true)
     }
 
+    def toActionMovedTemporary() {
+        redirect(action:'foo', moved:true)
+    }
+
+    def toActionMovedPermanent() {
+        redirect(action:'foo', permanent:true, moved:true)
+    }
+
     def toRoot() {
         redirect(controller:'default')
     }
--- a/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
+++ b/grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy
@@ -54,19 +54,19 @@ class RedirectController {
     }
 
     def toActionTemporaryRedirect() {
-        redirect(action:'foo', temporary:true)
+        redirect(action:'foo', moved: false)
     }
 
     def toActionPermanentRedirect() {
-        redirect(action:'foo', permanent:true, temporary:true)
+        redirect(action:'foo', permanent:true, moved:false)
     }
 
     def toActionMovedTemporary() {
-        redirect(action:'foo', moved:true)
+        redirect(action:'foo')
     }
 
     def toActionMovedPermanent() {
-        redirect(action:'foo', permanent:true, moved:true)
+        redirect(action:'foo', permanent:true)
     }
 
     def toRoot() {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/mvc/RedirectController.groovy [hunk 1]
```diff
@@ -53,6 +53,16 @@ class RedirectController {
         redirect(action:"${prefix}oo")
     }
 
+    def toActionTemporaryRedirect() {
+        request.setAttribute('statusConfig', [tempRedirect: true])
```
jdaugherty: The existing code seems to add a dedicated parameter for these purposes.  Can you please update your PR to do so?  i.e. ARGUMENT_PERMANENT - there should be one for ARGUMENT_TEMPORARY

dakshmehta007: Sure.

dakshmehta007:
@jdaugherty  I've updated the implementation as suggested to use a dedicated argument instead of request attributes. Now the implementation changes:

- Adds a new `ARGUMENT_TEMPORARY` constant following the pattern of `ARGUMENT_PERMANENT`
- Uses arguments instead of request attributes for controlling redirect behavior
- Updates the usage to be more consistent with existing patterns: groovy

// For 307 Temporary Redirect
redirect(action:'foo', temporary:true)

// For 308 Permanent Redirect 
redirect(action:'foo', permanent:true, temporary:true)

jdaugherty:
@dakshmehta007 Can you please also submit a pull request to update the documentation?  (https://github.com/apache/grails-doc/blob/7.0.x/src/en/ref/Controllers/redirect.adoc)

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -39,6 +39,7 @@ class ResponseRedirector {
 
     public static final String ARGUMENT_PERMANENT = "permanent"
     public static final String ARGUMENT_ABSOLUTE = "absolute"
+    public static final String ARGUMENT_TEMPORARY = "temporary" // Add new constant
```
jdaugherty: Drop the comment to be consistent with the current code

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -104,13 +105,13 @@ class ResponseRedirector {
                 namedParameters.put(LinkGenerator.ATTRIBUTE_PARAMS, configuredParams + webRequest.originalParams)
             }
         }
-        redirectResponse(linkGenerator.getServerBaseURL(), linkGenerator.link(namedParameters), request, response, permanent, absolute)
+        redirectResponse(linkGenerator.getServerBaseURL(), linkGenerator.link(namedParameters), request, response, permanent, absolute, arguments)
```
jdaugherty: Lets match the existing style here and not pass the map?

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -126,7 +127,15 @@ class ResponseRedirector {
         }
 
         String redirectUrl = useJessionId ? response.encodeRedirectURL(redirectURI) : redirectURI
-        int status = permanent ? HttpServletResponse.SC_MOVED_PERMANENTLY : HttpServletResponse.SC_MOVED_TEMPORARILY
+
+        // Update to use arguments instead of request attributes
+        boolean temporary = Boolean.TRUE == arguments.get(ARGUMENT_TEMPORARY)
```
jdaugherty: This may be a string value (like permanent).  Can you please parse this in the prior method to be consistent and handle the same edge cases as permanent.

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -126,7 +127,15 @@ class ResponseRedirector {
...
+        int status
+        if (permanent) {
+            status = temporary ? HttpServletResponse.SC_PERMANENT_REDIRECT : HttpServletResponse.SC_MOVED_PERMANENTLY
```
jdaugherty: Instead of the name temporary, wouldn't 'moved' be a better name?  Where permanent decides between temporary/ permanent and moved decides between the MOVED value or not.

jdaugherty (changes requested):

dakshmehta007:
@jdaugherty, here is the pull request you requested (https://github.com/apache/grails-doc/pull/966) in the grails-doc repository.

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -39,6 +39,7 @@ class ResponseRedirector {
 
     public static final String ARGUMENT_PERMANENT = "permanent"
     public static final String ARGUMENT_ABSOLUTE = "absolute"
+    public static final String ARGUMENT_MOVED = "moved"  // Renamed from ARGUMENT_TEMPORARY
```
jdaugherty: can you remove the comment?

grails-web-url-mappings/src/main/groovy/grails/web/mapping/ResponseRedirector.groovy
```diff
@@ -110,37 +111,32 @@ class ResponseRedirector {
     /*
      * Redirects the response the the given URI
      */
-    private void redirectResponse(String serverBaseURL, String actualUri, HttpServletRequest request, HttpServletResponse response, boolean permanent, boolean absolute) {
-        if(log.isDebugEnabled()) {
```
jdaugherty, line 114: There appear to be far more changes to the original code than I expected on this subsequent change.  

Some examples:
* not calling the link generator / change of hte uri
* removal of logging
* header changes 

Can you explain why you rewrote this?

jdaugherty (changes requested):

jdaugherty:
I've updated this PR so that the tests pass, there are tests for each, and the original code is preserved where possible.

PR #14065, opened 2025-03-12
Optimize the performance of the Shell CLI by reducing the length of t…

rainboyan:
…he CLASSPATH

* Excludes runtime dependencies of grails-shell
* Only include dependencies of grails-shell in the final distribution, no sources, javadocs, ivy files

Fixes gh-14029

PR #14093, opened 2025-03-30
Restructure build

matrei:
Make the build composition based.


gradle/assemble-config.gradle
```diff
@@ -0,0 +1,19 @@
...
+    inputs.files(tasks.named('javadocJar').map { it.outputs.files })
+    outputs.dir(distInstallDir)
+    from layout.buildDirectory.dir('libs')
+    into distInstallDir
+}
+
+tasks.withType(PublishToMavenLocal).configureEach {
+    dependsOn('installToHomeDist')
+    doLast {
+        copy {
```
jdaugherty, line 13: From what I can tell, `homeDistDir` is set to `dist` in the root project directory.

So the same concern I have with installToHomeDist, I have here.  What ensures the changed dependencies are actually repopulated and old jars aren't left in that directory?  I don't see a clean-up of this directory.

gradle/docs-config.gradle
```diff
@@ -0,0 +1,19 @@
+apply from: rootProject.layout.projectDirectory.file('gradle/docs-dependencies.gradle')
+
+ext {
+    includeInApiDocs = true
+}
+
+tasks.named('groovydoc', Groovydoc) {
+    classpath = configurations.documentation
+    groovyClasspath = configurations.documentation
+    access = GroovydocAccess.PRIVATE
```
jdaugherty: We have been changing these to protected in other projects, can we change it here?

gradle/docs-config.gradle
```diff
@@ -0,0 +1,19 @@
+apply from: rootProject.layout.projectDirectory.file('gradle/docs-dependencies.gradle')
+
+ext {
+    includeInApiDocs = true
+}
+
+tasks.named('groovydoc', Groovydoc) {
```
jdaugherty, line 7: grailsPublish ties groovydoc and javadoc as long as you have this: 

	java {
	    withSourcesJar()
	    withJavadocJar()
	}

Why is this needed here?

gradle/java-config.gradle
```diff
@@ -0,0 +1,19 @@
+compileJava.options.release = javaVersion.toInteger()
+
+tasks.withType(GroovyCompile).configureEach {
+    groovyOptions.encoding = 'UTF-8'
+    options.encoding = 'UTF-8'
+    options.fork = true
+    options.forkOptions.jvmArgs = ['-Xms128M', '-Xmx1G']
+}
+
+tasks.withType(Jar).configureEach {
```
jdaugherty, line 16: Why keep the attribute addition when we build in github actions and no other 'core' project does this?

gradle/unit-test.gradle
```diff
@@ -1,129 +0,0 @@
-// todo Unify test tasks into one multithreaded execution unit with a custom fork frequency
-// todo Add test progress listener with dot notation.
-
-configurations {
```
jdaugherty, line 4: We probably should create a ticket to adopt jacoco and setup a github actions to publish code coverage %

grails-dependencies/src/main/resources/publish-fix
```diff
@@ -0,0 +1 @@
+Grails Gradle Plugin needs something in a sourceSet to be able to publish
```
jdaugherty, line 1: This is fine for this review, but I would like to fix this separately:

What do you think about fixing this by adding a 'bom' property to grails publish?  It would remove this requirement and be a way to convey this project only publishes dependencies.  I am open to other names for the property if you can think of one.

grails-docs/build.gradle
```diff
@@ -1,10 +1,23 @@
...
     // Required to keep Gradle classes off the test compile classpath.
     gradleConf.extendsFrom compileClasspath
 }
 
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     gradleConf gradleApi()
+    api "org.apache.groovy:groovy:$groovyVersion"
```
jdaugherty: Will this work?  This is a gradle task so it must use the groovy version of gradle.

grails-logging/build.gradle
```diff
@@ -1,3 +1,41 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-core")
```
jdaugherty, line 14: single quotes?

grails-plugin-codecs/build.gradle
```diff
@@ -1,4 +1,50 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web"), project(":grails-encoder")
```
jdaugherty, line 14: I have never seen this syntax; can this be split across 2 lines?

Single quotes?

grails-plugin-databinding/build.gradle
```diff
@@ -1,5 +1,18 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
+dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(':grails-core'), project(':grails-web')
```
jdaugherty, line 14: Can we please split these across multiple lines?

grails-plugin-i18n/build.gradle
```diff
@@ -1,4 +1,50 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web")
```
jdaugherty, line 14: single quotes?

grails-plugin-url-mappings/build.gradle
```diff
@@ -1,4 +1,50 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web"), project(":grails-plugin-controllers")
```
jdaugherty, line 14: split & single quotes?

grails-plugin-validation/build.gradle
```diff
@@ -1,4 +1,50 @@
...
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-core"),
         project(':grails-web')
```
jdaugherty, line 15: can we declare the grails-web dependency as a separate api dependency?

grails-spring/build.gradle
```diff
@@ -1,7 +1,44 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api "org.springframework:spring-tx"
```
jdaugherty, line 14: single quotes?

grails-test-suite-base/build.gradle
```diff
@@ -1,4 +1,16 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(':grails-bootstrap'),
```
jdaugherty, line 14: can we split this up into separate api statements?

grails-test-suite-base/build.gradle
```diff
@@ -7,4 +19,35 @@ dependencies {
         project(':grails-plugin-domain-class')
 
     api project(":grails-plugin-converters")
```
jdaugherty, line 21: single quotes?

grails-test-suite-persistence/build.gradle
```diff
@@ -1,3 +1,12 @@
...
+    id 'groovy'
+    id 'java-library'
+    id 'project-report'
+}
+
+version = projectVersion
+group = 'org.grails'
+
 configurations.testCompileClasspath {
     exclude module: "grails-plugin-testing"
```
jdaugherty, line 11: single quotes?

grails-test-suite-persistence/build.gradle
```diff
@@ -1,3 +1,12 @@
...
 }
```
jdaugherty, line 12: Can we split the dependencies into their own lines and scopes?

grails-test-suite-uber/build.gradle
```diff
@@ -1,39 +1,60 @@
...
+    testRuntimeOnly 'com.h2database:h2'
+    testRuntimeOnly 'org.apache.tomcat:tomcat-jdbc'
+    testRuntimeOnly 'org.aspectj:aspectjrt'
+    testRuntimeOnly 'org.aspectj:aspectjweaver'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly 'org.springframework:spring-aspects'
+
+
+    // Testing
+    testImplementation('org.spockframework:spock-core') { transitive = false }
```
jdaugherty, line 55: Do you know why we include the transitive other places but exclude it here?

grails-web-mvc/build.gradle
```diff
@@ -1,4 +1,50 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web-common"),
```
jdaugherty, line 14: Can we split these?

grails-web-url-mappings/build.gradle
```diff
@@ -1,7 +1,20 @@
...
+}
+
+version = projectVersion
+group = 'org.grails'
+
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web-common")
```
jdaugherty, line 14: single quotes?

grails-web/build.gradle
```diff
@@ -1,8 +1,54 @@
...
+
+    implementation platform(project(':grails-bom'))
+
     api project(":grails-web-common")
     api project(":grails-web-databinding")
     api project(":grails-web-url-mappings")
     api project(":grails-web-mvc")
+    api "org.apache.groovy:groovy:$groovyVersion"
 
     api "org.grails:grails-web-gsp"
```
jdaugherty, line 20: single quotes?

jdaugherty, line 20: side topic: isn't this in the grails views project?  This seems like a reason to bring views into core

build.gradle
```diff
@@ -36,569 +10,66 @@ ext {
...
-
-    // This added to prevent remote cache miss, because project JAR include manifest file with Build-By and Created-By properties which might be different for CI vs Local.
+    // This is added to prevent a remote cache misses, because the project JAR include a manifest
+    // file with Built-By and Created-By properties which might be different for CI vs Local.
     normalization {
         runtimeClasspath {
             metaInf {
-                ignoreAttribute("Built-By")
-                ignoreAttribute("Created-By")
+                ignoreAttribute('Built-By')
```
jdaugherty, line 55: Let's assume we remove the properties,  then would this still be required?  Can we remove this + remove the jar properties?

gradle/assemble-config.gradle
```diff
@@ -0,0 +1,19 @@
+tasks.register('installToHomeDist', Copy) {
```
jdaugherty, line 1: From what I can tell, `distInstallDir` is set to `build/dist-tmp` in the root project.  

This is included in each project, so that means each project will copy it's jars into this directory.  That means we can't use the `Sync` task to "cleanup" what's not supposed to belong.  That's unfortunate: if the dependencies change, how do we ensure the old dependency isn't still copied to build/dist-tmp?  I don't see any code that forces a clean when these tasks are run across projects.

gradle/assemble-config.gradle
```diff
@@ -0,0 +1,19 @@
...
+    inputs.files(tasks.named('jar').map { it.outputs.files })
```
jdaugherty, line 2: Isn't what determines the input to this task what creates the libs?  Thus, it's just the configuration that's needed? Shouldn't we let gradle determine the files by the configuration instead of explicitly listing them all?

gradle/assemble-root-config.gradle
```diff
@@ -5,40 +5,23 @@ import org.grails.gradle.GrailsBuildPlugin
...
-            def javadocArtifacts = javadocFor(configuration).resolvedConfiguration.lenientConfiguration.artifacts.groupBy {
-                it.moduleVersion.id
-            }
-            def pomArtifacts = pomFor(configuration).resolvedConfiguration.lenientConfiguration.artifacts.groupBy {
-                it.moduleVersion.id
-            }
-
-            for (artifact in configuration.resolvedConfiguration.resolvedArtifacts) {
+        project(':grails-shell').configurations.cli.with {
+            exclude(group: 'org.codehaus.groovy')
```
jdaugherty: Shouldn't this exclude `org.apache.groovy` too?

gradle/assemble-root-config.gradle
```diff
@@ -57,16 +40,20 @@ tasks.register('populateDependencies', Sync) {
 // Can't use sync task here because this directory contains other things as well.
 tasks.register('sourcesJars', Sync) {
     into homeSrcDir
-    from rootProject.subprojects.findAll { subproject ->
-        !subproject.name.startsWith('grails-test-suite') &&
-        !subproject.name.startsWith('grails-dependencies') &&
-        !subproject.name.startsWith('grails-bom') &&
-        !subproject.name.contains('grails-test-examples')
+    from subprojects.findAll {
+        !it.name.startsWith('grails-test-suite') &&
```
jdaugherty: What are your thoughts about adding a property to the root that captures the list of projects that are published but not a test / bom / dependency set?  It seems like this logic is still used after the refactor so having it in one place would make this easier.

It seems this set of projects is used more than once in this file alone.

gradle/assemble-root-config.gradle
```diff
@@ -116,7 +103,7 @@ class GrailsCreateStartScripts extends CreateStartScripts {
 }
 
 tasks.register('install') { task ->
-    dependsOn 'populateDependencies', 'grailsCreateStartScripts'
+    dependsOn 'syncCliDependencies', 'grailsCreateStartScripts'
     subprojects { Project project ->
         if(!project.name.startsWith('grails-test-suite') && !project.name.startsWith('grails-test-examples')) {
             task.dependsOn("$project.name:publishToMavenLocal")
```
jdaugherty: Shouldn't this be the same set of projects from the line#44 comment?

gradle/assemble-root-config.gradle
```diff
@@ -126,7 +113,8 @@ tasks.register('install') { task ->
 //task install(dependsOn: [populateDependencies, grailsCreateStartScripts] + subprojects.findAll { !it.name.startsWith('grails-test-suite') }
 //                                                                                                        *.collect { Project p -> p.tasks.withType(PublishToMavenLocal)})
```
jdaugherty: Can we removed the commented code?

gradle/functional-test-config.gradle
```diff
@@ -0,0 +1,39 @@
+def pluginsGroupProjects = rootProject.subprojects
+        .findAll { it.group == 'org.grails.plugins' }
+        .collect { it.name }
+
+configurations.configureEach {
```
jdaugherty, line 5: Can we leave the original comment here (slightly adjusted)?  

        // Test projects will often include dependencies from grails-core, this will ensure any dependencies included
        // will be substituted with projects in this repository instead of pulling upstream

gradle/functional-test-config.gradle
```diff
@@ -0,0 +1,39 @@
...
+    resolutionStrategy {
+        dependencySubstitution {
+            for (def possibleProject : rootProject.subprojects) {
+                if (!possibleProject.name.startsWith('grails-test-suite') && !possibleProject.name.contains('grails-test-examples')) {
+                    def artifactId = mappedArtifactIds[possibleProject.name] ?: possibleProject.name
+                    def substitutedArtifact = "${possibleProject.name in pluginsGroupProjects ? 'org.grails.plugins' : 'org.grails'}:$artifactId"
```
jdaugherty: Now that we have the projects specifying `pomArtifactId` and the group is defined inside of the project, do we even need this mapping?  Can't this now be `${possibleProject.group}:${possibleProject.pomArtifactId ?: possibleProject.name}`?

gradle/functional-test-config.gradle
```diff
@@ -0,0 +1,39 @@
+def pluginsGroupProjects = rootProject.subprojects
+        .findAll { it.group == 'org.grails.plugins' }
+        .collect { it.name }
+
+configurations.configureEach {
+    resolutionStrategy {
+        dependencySubstitution {
+            for (def possibleProject : rootProject.subprojects) {
+                if (!possibleProject.name.startsWith('grails-test-suite') && !possibleProject.name.contains('grails-test-examples')) {
+                    def artifactId = mappedArtifactIds[possibleProject.name] ?: possibleProject.name
```
jdaugherty: Is this mapping still needed?  Each project will specify it directly now, yes?

grails-bom/build.gradle
```diff
@@ -15,6 +19,9 @@ def versions = new Properties()
 versions.load(new StringReader(new File("$projectDir.parentFile/gradle.properties").text))
 
 ext {
```
jdaugherty, line 21: Can we update the task registrations in this file to be lazy?

jdaugherty (changes requested):
1. as discussed, we should leave the gradle plugins (gsp, views, etc) in the functional tests. I realize that the compiled gsps aren't used by the tests because we run in dev mode, but it still exercises that plugin in a basic manner.  Eventually, we'll want to add production mode tests to make use of these for end-to-end coverage.

2. Can you please confirm you've published to maven local before this change & saved off the artifacts.  Then, published after this change and diffed the directory structure / names / files?  This will ensure we haven't broken anything.

3. Can you please confirm you've made sure the same # of tests run before and after this change? 

See other comments, but overall this is an amazing job.  Thank you so much!

jdaugherty:
Did you also test grails-doc and make sure if it has this code that it can still generate?

gradle/java-config.gradle
matrei: I'm don't know why it is added, so I did not want to remove it. Maybe we can remove it?

gradle/assemble-root-config.gradle
matrei: Yes

gradle/functional-test-config.gradle
matrei: Oh, sorry, I did not mean to remove that. I'll add it back.

gradle/functional-test-config.gradle
matrei: For some reason `grails-testing-support` and `grails-web-boot` do not show the right `group`.
Added some logging:
```console
Substituting org.grails:grails-spring with project grails-spring
Substituting org.grails:grails-test with project grails-test
Substituting grails.core.ROOT:grails-testing-support with project grails-testing-support
Substituting org.grails:grails-web with project grails-web
Substituting grails.core.ROOT:grails-web-boot with project grails-web-boot
Substituting org.grails:grails-web-common with project grails-web-common
```

So, as it is now, it only works as they are in the default group.

gradle/java-config.gradle
jdaugherty: lets remove it; its likely left over pre-github actions change

gradle/functional-test-config.gradle
jdaugherty: If you add the substitution code in a ` project.afterEvaluate { } ` block, does it work?

matrei: > If you add the substitution code in a project.afterEvaluate { } block, does it work?

No, that does not seem to work.

gradle/java-config.gradle
matrei: I do not understand the connection between adding these attributes to the jar files and using GitHub actions.
Are you saying that by using GitHub actions, this is not necessary?

jdaugherty: Yes, I am saying that historically grails was built on travis and before that on someone’s machine so it was important to know who built it.  Now it’s not needed because we have a standard CI.

gradle/docs-config.gradle
matrei: When removing this and adding `java.withJavadocJar()` the Grails Gradle Plugin seems to package both javadoc and groovydoc so that when unzipping the javadoc jar it has duplicates of the files, so you cannot extract it. And in the `build/docs` directory there is a `groovydoc` and a `javadoc` dir.

Shouldn't we only run `groovydoc` and package that into the `javadoc` jar, like I have done in this project now?

gradle/java-config.gradle
matrei: OK, thanks for clearing that up for me, I'll remove it.

gradle/docs-config.gradle
jdaugherty: This is intentional.  if you unzip with a commandline client it will overwrite the files and it will be as you expect.  We opened a bug ticket on gradle to make it replace instead of add duplicates.

I believe the reason we don’t run just the groovydoc package is some projects do include java only in a java source.  

As for the directories, our intention was to build the jar.  I need to look at the directories to see whats happening.

gradle/java-config.gradle
matrei: It seems removing the jar attributes broke `GrailsUtilTest` which is testing `GrailsUtil.getGrailsVersion()` which seems to be fetching the version from the jar attributes.

jdaugherty: is this there for the shell then?  or is that test even relevant now?

matrei: > is this there for the shell then? or is that test even relevant now?

`GrailsUtil.getGrailsVersion()` is used in 19 places in the `grails-core` projects.

gradle/assemble-config.gradle
matrei: These distribution directories are deleted by a `clean` task in the root build file.

gradle/docs-config.gradle
matrei: In the `grails-core` repo, there is not one single `src/main/java` source set.
Which projects are we using `src/main/java` in?

I don't think the solution is optimal at the moment. We are effectively generating and packaging both groovydoc and javadoc only to throw away the groovydoc when extracting the jar.

Should we be advocating groovy usage, and package generated groovydoc in the doc jar and only use the `src/main/groovy` source set?

gradle/unit-test.gradle
matrei: #14096

gradle/docs-config.gradle
matrei: > Can you please confirm you've published to maven local before this change & saved off the artifacts. Then, published after this change and diffed the directory structure / names / files? This will ensure we haven't broken anything.

Confirmed (after some fixes, thank you!)

matrei:
> Can you please confirm you've made sure the same # of tests run before and after this change?

Confirmed `2298 tests (28 ignored)` (after a fix, thank you!)

matrei:
> Can you please confirm you've published to maven local before this change & saved off the artifacts. Then, published after this change and diffed the directory structure / names / files? This will ensure we haven't broken anything.

Confirmed (after some fixes, thank you!)

grails-dependencies/src/main/resources/publish-fix
matrei: Yes, that would make sense.

grails-docs/build.gradle
matrei: IDK. I just refactored this dependency to be added in the subproject build file instead of in the root build file.
The `grails-docs` was not in the `compiledByGradleGroovyVersion` collection.
https://github.com/apache/grails-core/blob/f673a05c5e1e7e61830608564c0497554eaefbf4/build.gradle#L63
https://github.com/apache/grails-core/blob/f673a05c5e1e7e61830608564c0497554eaefbf4/build.gradle#L401-L410
https://github.com/apache/grails-core/blob/f673a05c5e1e7e61830608564c0497554eaefbf4/build.gradle#L598-L604

grails-logging/build.gradle
matrei: :smile: I had that one coming! However, this line has not been touched in this PR.

I have NOT gone through the dependencies sections in this PR, as I know the PR is blocking progress and I want to get it done.

Clarifying and cleaning up dependencies can be done in incremental steps later.
There is also work left to clean up the rest of the subproject build files.

grails-plugin-codecs/build.gradle
matrei: See previous response

grails-plugin-databinding/build.gradle
matrei: See previous response

grails-plugin-i18n/build.gradle
matrei: See previous response

grails-plugin-url-mappings/build.gradle
matrei: See previous response

grails-plugin-validation/build.gradle
matrei: See previous response

grails-spring/build.gradle
matrei: See previous response

grails-test-suite-base/build.gradle
matrei: See previous response

grails-test-suite-base/build.gradle
matrei: See previous response

grails-test-suite-persistence/build.gradle
matrei: See previous response

grails-test-suite-persistence/build.gradle
matrei: See previous response

grails-test-suite-uber/build.gradle
matrei: No, see previous response

grails-web-mvc/build.gradle
matrei: See previous response

grails-web-url-mappings/build.gradle
matrei: See previous response

grails-web/build.gradle
matrei: See previous response

grails-logging/build.gradle
jdaugherty: Makes sense.  Thank you.

grails-dependencies/src/main/resources/publish-fix
jdaugherty: https://github.com/apache/grails-gradle-plugin/issues/429

gradle/java-config.gradle
jdaugherty: So it looks like Environment.groovy is what uses this.  That's fine, but can we apply this only to the bootstrap class then?  It looks for the manifest file in the same location as the Environment.class, which is only in the bootstrap project.

grails-web/build.gradle
matrei: `grails-web-gsp` is in fact not used in `grails-web` and can be removed (unless some downstream project depends on `grails-web` providing it, which we know is a bad idea).

It seems `grails-plugin-mimetypes` is using `org.grails.gsp.ResourceAwareTemplateEngine` and it's sole dependency is `grails-web`.

gradle/docs-config.gradle
jdaugherty: I'm only suggesting we fix this in the gradle plugin in a separate pull and switching to this for now to be consistent across the projects.

grails-docs/build.gradle
jdaugherty: I think we have found a bug =)  

Can we please change this to be the groovy version that's used by gradle?

grails-web/build.gradle
jdaugherty: Looks like ResourceAwareTemplateEngine is used in grails-web-common too.  I suspect we need to revisit the project structure of gsp.  It seems like over time several classes have been pulled up the project chain.  I'll take a look at this after the apache publishing is working.

jdaugherty: https://github.com/apache/grails-views/issues/726

build.gradle
jdaugherty: I think we can remove this if we only apply it to bootstrap.  It's ok to miss the cache on that library.

gradle/assemble-root-config.gradle
jdaugherty: Since these are compiled with the groovy version of gradle, this is ok for now.

gradle/docs-config.gradle
matrei: OK, I'll switch to letting the Gradle Plugin handle this.

matrei:
> Did you also test grails-doc and make sure if it has this code that it can still generate?

I don't know what you mean by this.

gradle/functional-test-config.gradle
jdaugherty: Making the functional tests dependent on the framework projects seems to fix resolving the group:  

		Set<Project> nonFunctionalProjects = rootProject.subprojects.findAll { !it.name.startsWith('grails-test-examples') }
		nonFunctionalProjects.each {
		    project.evaluationDependsOn(it.path)
		}

jdaugherty:
> > Did you also test grails-doc and make sure if it has this code that it can still generate?
> 
> I don't know what you mean by this.

https://github.com/apache/grails-doc can either checkout or point to a grails home.  It will then build the documentation to be published.  You should be able to checkout that project, point it to this branch, and then run the doc task to generate documentation.  If it doesn't work, we'll need to adjust the gradle files ...

gradle/docs-config.gradle
jdaugherty: https://github.com/apache/grails-gradle-plugin/issues/430

gradle/java-config.gradle
matrei: > but can we apply this only to the bootstrap class then?

Do you mean add the version to the jar manifest only in the `grails-bootstrap` project?

matrei:
> > > Did you also test grails-doc and make sure if it has this code that it can still generate?
> > 
> > 
> > I don't know what you mean by this.
> 
> https://github.com/apache/grails-doc can either checkout or point to a grails home. It will then build the documentation to be published. You should be able to checkout that project, point it to this branch, and then run the doc task to generate documentation. If it doesn't work, we'll need to adjust the gradle files ...

Aha, I see what you mean. It does seem to work. `grails-doc` even includes Groovy 4 on the build classpath!!

https://github.com/apache/grails-doc/blob/174051c0d071bc889d213f46820d2aab9b02eb4a/buildSrc/build.gradle#L16-L20

gradle/java-config.gradle
matrei: After moving the attributes to only the `grails-bootstrap` jar I'm getting test errors all over the place.
Seems to be originating from:
https://github.com/apache/grails-data-mapping/blob/00fbc39170146c6ae844c66c7e2056a4e7565ba6/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersion.groovy#L147

grails-docs/build.gradle
jamesfredley: I think we got lucky on this one and the Groovy 4 compiled classes ran fine in Groovy 3, but I do agree with changing any code executed during the Gradle build to use `$GroovySystem.version` as a precaution.

matrei: Changed!

build.gradle
matrei: We should keep this and the JAR attributes in this PR to ensure functionality is preserved. We can revisit once Snapshots are up and running.

gradle/assemble-config.gradle
matrei: Won't the `inputs` of the `installToHomeDist` task make sure any changes to the output jars are detected and that they are copied to/overwritten in `distInstallDir`?

gradle/java-config.gradle
matrei: We should keep the JAR attributes in this PR to ensure functionality is preserved. We can revisit once Snapshots are up and running.

gradle/assemble-config.gradle
matrei: I don't think using only the dependency configuration will do it. There are other factors (like changed sources) that can change the `jar` task output.

gradle/assemble-root-config.gradle
matrei: Done!

gradle/assemble-root-config.gradle
matrei: Refactored

build.gradle
```diff
@@ -35,570 +9,76 @@ ext {
...
-                def ossUser = System.getenv("SONATYPE_USERNAME") ?: project.hasProperty("sonatypeOssUsername") ? project.sonatypeOssUsername : ''
-                def ossPass = System.getenv("SONATYPE_PASSWORD") ?: project.hasProperty("sonatypeOssPassword") ? project.sonatypeOssPassword : ''
-                def ossStagingProfileId = System.getenv("SONATYPE_STAGING_PROFILE_ID") ?: project.hasProperty("sonatypeOssStagingProfileId") ? project.sonatypeOssStagingProfileId : ''
-                nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
-                username = ossUser
-                password = ossPass
-                stagingProfileId = ossStagingProfileId
-            }
-        }
+    mappedArtifactIds = [
```
jdaugherty: Should we build this list based on the pomArtifactId? It looks like this is only used in the grails bom.  Maybe move the logic there?

build.gradle
```diff
@@ -35,570 +9,76 @@ ext {
...
-                    }
-
-                    if(isAsyncProject) {
-                        versionMapping {
-                            usage('java-api') { fromResolutionOf('runtimeClasspath') }
-                            usage('java-runtime') { fromResolutionResult() }
-                        }
-                    }
+                if (details.requested.group == 'org.spockframework') {
+                    details.useVersion(findProperty('spock.version') as String)
```
jdaugherty: Since we have a bom, is this required?

build.gradle
jdaugherty: https://github.com/apache/grails-core/issues/14098

gradle/docs-config.gradle
```diff
@@ -0,0 +1,14 @@
...
```
jdaugherty, line 7: Shouldn't this be .configure { }?

gradle/java-config.gradle
jdaugherty: I agree we should address after this PR.  We should consolidate to one, but it's out of scope of this PR.

https://github.com/apache/grails-core/issues/14098

grails-core/src/main/groovy/grails/core/DefaultGrailsApplication.java
```diff
@@ -627,8 +627,8 @@ public ArtefactInfo getArtefactInfo(String artefactType) {
      * @param methodName The name of the method
      * @param args       The arguments to the method
      * @return The return value of the method
-     * TODO Need to add matches for add<Artefact>Class(java.lang.Class) and add<Artefact>Class(GrailsClass)
      */
+    // TODO Need to add matches for add<Artefact>Class(java.lang.Class) and add<Artefact>Class(GrailsClass)
```
jdaugherty, line 631: By moving this comment out of the javadoc, won't this mean it won't show externally in the api documentation?

grails-docs/build.gradle
```diff
@@ -1,24 +1,51 @@
...
 
 dependencies {
+
+    implementation platform(project(':grails-bom'))
+
     gradleConf gradleApi()
+
+    // grails-docs classes are used in Gradle builds,
+    // so we must compile with Groovy 3 until Gradle upgrades to Groovy 4.
+    compileOnly "org.codehaus.groovy:groovy:$GroovySystem.version"
```
jdaugherty, line 25: thank you!

grails-test-examples/hyphenated/build.gradle
```diff
@@ -11,10 +11,6 @@ version "0.1"
 group "hyphenated"
 
 apply plugin: "org.grails.grails-web"
-apply plugin: "org.grails.grails-gsp"
```
jdaugherty, line 14: Looks like this project has gsps, can we keep the gsp plugin?

grails-test-examples/issue-11102/build.gradle
```diff
@@ -11,18 +11,8 @@ buildscript {
 version "0.1"
 group "issue11102"
 
-apply plugin: "war"
 apply plugin: "asset-pipeline"
 apply plugin: "org.grails.grails-web"
-apply plugin: "org.grails.grails-gsp"
```
jdaugherty, line 17: Looks like the project still has gsps, can we keep this?

grails-test-examples/issue-views-182/build.gradle
```diff
@@ -4,18 +4,13 @@ buildscript {
...
         classpath "org.grails:grails-gradle-plugin:${project['grails-gradle-plugin.version']}"
-        classpath "org.grails.plugins:views-gradle:${project['plugins-views-gradle.version']}"
     }
 }
 
 version "0.1"
 group "issueviews182"
 
 apply plugin:"org.grails.grails-web"
-apply plugin:"org.grails.plugins.views-json"
```
jdaugherty, line 15: There are `gson` views in this project, can we please keep this?

grails-test-examples/micronaut/build.gradle
```diff
@@ -4,24 +4,16 @@ buildscript {
...
 }
 
 version "0.1"
 group "micronaut"
 
-apply plugin:"eclipse"
-apply plugin:"idea"
-apply plugin:"war"
 apply plugin:"org.grails.grails-web"
-apply plugin:"org.grails.grails-gsp"
```
jdaugherty, line 19: Can we keep gsp?

jdaugherty (approved):
Can we please be sure to add back the remaining gsp / gson plugins where I marked?

build.gradle
matrei: Thanks for reminding me!

build.gradle
matrei: Good catch!

gradle/docs-config.gradle
matrei: I don't think that should not be necessary when using `tasks.named()`.

grails-core/src/main/groovy/grails/core/DefaultGrailsApplication.java
matrei: Yes, that is true. I did not recognize it as part of the javadoc, but more of a code comment. Do you think it should be part of public javadoc?

grails-test-examples/hyphenated/build.gradle
matrei: The demo33 project has gsp compilation check during build (with explaining comments).
I thought it would be enough to have that smoke test in one project.

grails-test-examples/issue-11102/build.gradle
matrei: See previous comment.

grails-test-examples/issue-views-182/build.gradle
matrei: The demo33 project has json-views compilation check during build (with explaining comments).
I thought it would be enough to have that smoke test in one project.

grails-test-examples/micronaut/build.gradle
matrei: See previous comment.

matrei:
> Can we please be sure to add back the remaining gsp / gson plugins where I marked?

We could, but I think we are only wasting build time. The demo33 project smoke tests the Gradle plugins.

jdaugherty:
> > Can we please be sure to add back the remaining gsp / gson plugins where I marked?
> 
> We could, but I think we are only wasting build time. The demo33 project smoke tests the Gradle plugins.

We don't know how these projects would diverge over time so it's safe to add them to all of them.  Specifically, micronaut is likely going to be different than the base grails app so it's probably critical to have it there at the least.

grails-core/build.gradle
```diff
@@ -1,10 +1,23 @@
...
+
+    implementation platform(project(':grails-bom'))
+
     compileOnly "org.aspectj:aspectjrt", "org.aspectj:aspectjweaver"
     api "jakarta.inject:jakarta.inject-api"
     api "jakarta.persistence:jakarta.persistence-api"
     api "jakarta.annotation:jakarta.annotation-api"
 
     implementation "com.github.ben-manes.caffeine:caffeine"
+    api "org.apache.groovy:groovy:$groovyVersion"
```
jamesfredley, line 20: These surprised me, but they were in the POMs and did not cause version issues so we are fine.

jamesfredley:
I tested grails-shell CLI zip and these changes add a doc directory with groovydocs (1,815 files) whereas it was 222 files total before.  I think we should skip the docs in the zip distribution.  

Additionally this issue https://github.com/apache/grails-core/pull/14065/files is back on Windows.

grails-core/src/main/groovy/grails/core/DefaultGrailsApplication.java
jdaugherty: I'm not sure, maybe since it was there before it was intentional?  I'd leave it for now.

jdaugherty:
> I tested grails-shell CLI zip and these changes add a doc directory with groovydocs (1,815 files) whereas it was 222 files total before. I think we should skip the docs in the zip distribution.
> 
> Additionally this issue https://github.com/apache/grails-core/pull/14065/files is back on Windows.

I guess we should skip the withJavadoc() for the cli project?

jamesfredley (approved):
I'd like to fix the grails-shell CLI issue on Windows that reappeared before merging this, but would also be OK moving that to another ticket.

grails-core/build.gradle
jdaugherty: the hard coded version is likely why we have to use the override technique we do for groovy.  Since it already did this, I wasn't going to advocate for change. But eventually, we probably should pull that version from the bom instead.

matrei:
> I tested grails-shell CLI zip and these changes add a doc directory with groovydocs (1,815 files) whereas it was 222 files total before. I think we should skip the docs in the zip distribution.

Are you sure this has changed? I checked the previous release ([v7.0.0-M3](https://github.com/apache/grails-core/releases/download/v7.0.0-M3/grails-7.0.0-M3.zip)), and it appears to include the API docs in the doc subdirectory.

jamesfredley (approved):
@matrei The changes for the CLI on windows work perfectly.

I went back and checked and you are correct on the doc directory being present for 7.0.0-M3 and 7.0.0-M1.  The doc directory was not present on 6.2.x distributions and I have created an issue to explore that separately.  https://github.com/apache/grails-core/issues/14100

Great work reorganizing this build!!!

jdaugherty (approved):

PR #14107, opened 2025-04-03
apache coordinate migration for grails-geb

jdaugherty:
This PR will need to be merged after the PR in grails-geb


jdaugherty:
Because I have to keep changing coordinates, and the time it takes to do this, I'm disabling the functional tests to save time.  Will re-enable after views + data mapping are published.

PR #14115, opened 2025-04-04
apply asset-pipeline-gradle plugin

jamesfredley:
when using com.bertramlabs.plugins:asset-pipeline-grails

resolves java.lang.ClassNotFoundException: org.graalvm.polyglot.Context

bertramdev/asset-pipeline#370
bertramdev/asset-pipeline#372
bertramdev/asset-pipeline#374


jdaugherty (approved):

jdaugherty:
Will merge after we re-enable the functional tests from changing the coordinates.

matrei (commented):
The  `asset-pipeline-gradle` plugin is not applied in these applications as there are no assets in them.

jdaugherty:
There are no assets, but because there is a runtime dependency on the asset pipeline, it outputs the java.lang.ClassNotFoundException: org.graalvm.polyglot.Context error over and over.   It's our understanding that applying the plugin will fix that?

matrei:
> it outputs the java.lang.ClassNotFoundException: org.graalvm.polyglot.Context error over and over

That is surprising to me considering there are no assets.

> there is a runtime dependency on the asset pipeline

I don't think there is a runtime dependency and if it is declared in the dependencies block, it can be probably be removed.

matrei:
In some of these apps the `asset` taglib is actually used but the assets are missing.
Maybe we should add the assets in these apps?

jdaugherty:
Aren't the build dependencies applied before the code can know assets exist?  That means we still have to apply them, no?

matrei (approved):

PR #14133, opened 2025-04-14
Merge grails-gradle-plugin repository

jdaugherty:
Per the decision [1], this merges the grails-gradle-plugin repository into this one.

This merge has the following impacts: 

1. Dependabot wasn't updating our versions in the bom (see byte-buddy.version as an example, 1.15.5 is checked in vs 1.17.5 is out).  It's not working because gradle property support is missing[2].  This commit refactors to the suggested approach of using a dependencies.gradle instaed of a gradle.properties to set values.  

2. This change also fixes some of the grails-views artifacts in the bom since they were wrong / duplicated.

3. The grails-bom is now split into `grails-gradle-bom` and `grails-bom`.  the gradle bom is to be used when importing the platform in gradle (it will correctly set the groovy version too).  `grails-bom` is to be used for dependencies of the application.

4. As part of adding the gradle plugins, we can't have a circular dependency to a child project, so this change decouples grails-bootstrap & the gradle related projects.  I've moved up logic from grails-bootstrap into grails-gradle-model and made sure grails-bootstrap exposes grails-gradle-model with an api dependency.  I consulted @davydotcom for the best approach and we decided on this one.

[1] https://lists.apache.org/thread/fgpwg1d8fsskkvhl1b9dgfn0gzxb7f0v
[2] https://github.com/dependabot/dependabot-core/issues/1618


grails-bom/build.gradle
```diff
@@ -12,221 +15,237 @@ javaPlatform {
...
-    }
-}
-
-tasks.register('dependabot') {
-    group = 'Dependency Management'
-    description = 'This task updates the dependabot/build.gradle and gradle.properties files by syncing them with the latest external dependencies listed in the Grails BOM (Bill of Materials). This ensures that Dependabot can monitor these dependencies and automatically create pull requests when newer versions become available.'
-
-    dependsOn 'syncProps', 'dependabotBuild'
-}
+//tasks.register('syncProps') {
```
jamesfredley: Can we remove?

grails-bootstrap/build.gradle
```diff
@@ -12,18 +12,17 @@ group = 'org.apache.grails.bootstrap'
...
 
+    api 'org.apache.grails.gradle:grails-gradle-model', {
+        exclude group: 'org.codehaus.groovy'
+    }
 
-    compileOnly "org.codehaus.groovy:groovy:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-templates:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-xml:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-ant:$GroovySystem.version"
+    compileOnly "org.apache.groovy:groovy"
```
jamesfredley: I think we still want to compile `grails-bootstrap` with with `$GroovySystem.version` since it is used by the gradle plugin when running the build.  

It technically works when compiled with Groovy 4, since the classes are Groovy 3 compatible, but that is lucky.

jamesfredley: This was due to grails-shell dependencies

https://github.com/apache/grails-core/blob/7.0.x/grails-shell-cli/build.gradle#L74

jamesfredley: It was done in these 3:

https://github.com/search?q=repo%3Aapache%2Fgrails-core+%22the+Groovy+version+provided+by+Gradle%22&type=code

grails-gradle/dependencies.gradle
```diff
@@ -0,0 +1,55 @@
+// see https://github.com/dependabot/dependabot-core/issues/1618
```
jamesfredley, line 1: Let's add a comment to each of the dependencies.gradle files referencing the other one since the versions are now split between them.

grails-gradle/gradle/docs-config.gradle
```diff
@@ -0,0 +1,42 @@
+configurations.register('documentation')
+dependencies {
+    add('documentation', platform(project(':grails-gradle-bom')))
+    add('documentation', 'org.fusesource.jansi:jansi')
+    add('documentation', 'jline:jline')
+    add('documentation', 'com.github.javaparser:javaparser-core')
+    add('documentation', "org.codehaus.groovy:groovy")
```
jamesfredley, line 7: single quotes above, double quotes below.   lets standardize these lines.

jamesfredley (approved):

grails-bootstrap/build.gradle
jdaugherty: A couple thoughts: 

1. The gradle plugin no longer uses bootstrap - I've pulled that code into the model.  Otherwise we would have a circular dependency.

2. I think you're right on shell though.  We'll need to compile with groovy 3 only because of Grails shell.

.github/workflows/gradle.yml
```diff
@@ -27,8 +27,35 @@ jobs:
...
+        with:
+          distribution: liberica
+          java-version: ${{ matrix.java }}
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY  }}
+      - name: "🔨 Build project"
+        working-directory: 'grails-gradle'
+        id: build
```
matrei: Do we need this `id`?

.github/workflows/gradle.yml
```diff
@@ -79,7 +106,7 @@ jobs:
         run: ./gradlew build -PonlyFunctionalTests
   publish:
     if: github.event_name == 'push' && needs.skip_check.outputs.found_skip_publish != 'true'
-    needs: [build, functional]
+    needs: [buildGradle, build, functional]
```
matrei: Should `skip_check` also be in this list, as it is referenced in the condition above?

.github/workflows/gradle.yml
```diff
@@ -103,9 +130,9 @@ jobs:
           MAVEN_PUBLISH_USERNAME: ${{ secrets.NEXUS_USER }}
           MAVEN_PUBLISH_PASSWORD: ${{ secrets.NEXUS_PW  }}
         run: >
-          ./gradlew
-          -Dorg.gradle.internal.publish.checksums.insecure=true
-          publish
+          ./gradlew publish &&
+          cd grails-gradle && 
+          ./gradlew publish -Dorg.gradle.internal.publish.checksums.insecure=true
```
matrei, line 134: Do we need the `-Dorg.gradle.internal.publish.checksums.insecure=true` option?

.github/workflows/gradle.yml
```diff
@@ -103,9 +130,9 @@ jobs:
...
   docs:
     if: github.event_name == 'push' && needs.skip_check.outputs.found_skip_publish != 'true'
     needs: publish
```
matrei: Should `skip_check` also be in `needs` as we are referencing it in the above condition?

dependencies.gradle
```diff
@@ -0,0 +1,102 @@
...
+            'liquibase-hibernate5.version' : '4.27.0',
+            'mongodb.version'              : '5.3.1',
+            'rxjava.version'               : '1.3.8',
+            'rxjava2.version'              : '2.2.21',
+            'rxjava3.version'              : '3.1.10',
+            'spock.version'                : '2.3-groovy-4.0',
+    ]
+
+    // Note: the name of the dependency must be the prefix of the property name so properties in the pom are resolved correctly
+    bomPlatformDependencies = [
```
matrei, line 31: Is there a particular reason why we use snake_case and not kebab-case for the map keys?

dependencies.gradle
```diff
@@ -0,0 +1,102 @@
...
+            spock_bom      : "org.spockframework:spock-bom:${bomDependencyVersions['spock.version']}",
+            jackson_bom    : "com.fasterxml.jackson:jackson-bom:${bomDependencyVersions['jackson.version']}",
+            groovy_bom     : "org.apache.groovy:groovy-bom:${bomDependencyVersions['groovy.version']}"
+    ]
+
+    // Note: the name of the dependency must be the prefix of the property name so properties in the pom are resolved correctly
+    bomDependencies = [
```
matrei, line 38: Is there a particular reason why we use snake_case and not kebab-case for the map keys?

gradle/functional-test-config.gradle
```diff
@@ -21,6 +21,13 @@ configurations.configureEach {
     }
 }
 
+// work around for parallel builds due to https://github.com/bertramdev/asset-pipeline/issues/177
+if('assetCompile' in tasks.names) {
+    tasks.named('assetCompile').configure { Task task ->
+        task.outputs.dir rootProject.layout.buildDirectory.dir('asset-serialize')
+    }
+}
```
matrei: Simplify? (Also `tasks.named` does not need `.configure`, it is implied)
```groovy
tasks.matching { 'assetCompile' == it.name }.configureEach {
    outputs.dir rootProject.layout.buildDirectory.dir('asset-serialize')
}
```

grails-bootstrap/src/test/groovy/grails/io/IOUtilsSpec.groovy
matrei, line 1: `IOUtils` has been moved to the `grails-gradle-model` project. Should this `IOUtilsSpec` also move there?

grails-bootstrap/src/test/groovy/org/grails/build/parsing/CommandLineParserSpec.groovy
matrei, line 1: `CommandLineParser` has been moved to the `grails-gradle-model` project. Should this `CommandLineParserSpec` also move there?

grails-core/build.gradle
```diff
@@ -33,7 +33,7 @@ dependencies {
 
     // This only needs the datastore version property because of how source jars are being created.
     // Build will succeed, but assemble will fail without this because of task ordering in gradle.
```
matrei: Is this comment still relevant?

grails-core/src/main/groovy/grails/config/Config.groovy
matrei, line 1: Update license header year range?

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/api/ControllersDomainBindingApi.java
matrei, line 1: Update license header year range and change to `original author or authors`?

grails-bootstrap/src/main/groovy/org/grails/config/CodeGenConfig.groovy
matrei, line 1: Update license header year range?

grails-core/src/main/groovy/grails/util/GrailsClassUtils.java
matrei, line 1: Update license header year range?

grails-core/src/main/groovy/org/grails/config/EnvironmentAwarePropertySource.java
matrei, line 1: Update license header year range?

grails-core/src/main/groovy/org/grails/core/DefaultGrailsControllerClass.java
matrei, line 1: Update license header year range?

grails-core/src/main/groovy/org/grails/core/cfg/GroovyConfigPropertySourceLoader.groovy
matrei, line 1: Update license header year range?

grails-core/src/main/groovy/org/grails/plugins/DefaultGrailsPlugin.java
matrei, line 1: Update license header year range?

grails-dependencies/build.gradle
```diff
@@ -12,26 +12,32 @@ ext {
...
+                ":grails-interceptors",
+                ":grails-async-plugin",
+                ":grails-events-plugin",
+                "org.apache.grails:grails-gsp",
         ],
         'runtimeOnly': [
-                "com.h2database:h2:${property('h2.version')}",
+                "com.h2database:h2",
         ]
 ]
```
matrei: Align on single quotes?

grails-events/plugin/build.gradle
```diff
@@ -24,7 +24,6 @@ dependencies {
 
     // As this is a plugin these dependencies will be provided by the Grails application
     // This removes circular dependencies
-    compileOnly project(':grails-bootstrap')
```
matrei, line 27: As `GrailsNameUtils` used in `SpringEventTranslator` has been moved from `grails-bootstrap` to `grails-gradle-model`, should this dependency also be changed to `grails-gradle-model` instead of deleted?

Also there is a `testRuntimeOnly project(':grails-bootstrap')` further down that probably also should be changed to `grails-gradle-model`.

This also begs the question if we should create another layer with these kind of base utility classes (which only depends on groovy or java) so we are not having to rely on `gradle` projects in `core` projects and vice versa.

grails-gradle/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+allprojects {
+    repositories {
+        // mavenLocal()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+
+    for (Map.Entry<Object, Object> entry : props.entrySet()) {
+        project.ext[entry.key as String] = entry.value
+    }
```
matrei: Simplify?
```groovy
props.forEach { k, v ->
    project.ext.set(k as String, v)
}
```

grails-gradle/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    apply from: rootProject.layout.projectDirectory.file('dependencies.gradle')
+}
+
+ext {
+    ext.isReleaseVersion = Boolean.parseBoolean(System.getenv('GRAILS_PUBLISH_RELEASE'))
+    ext."signing.keyId" = project.findProperty("signing.keyId") ?: System.getenv('SIGNING_KEY')
+    ext."signing.password" = project.findProperty("signing.password") ?: System.getenv('SIGNING_PASSPHRASE')
+    ext."signing.secretKeyRingFile" = project.findProperty("signing.secretKeyRingFile") ?: "${System.properties['user.home']}${File.separator}.gnupg${File.separator}secring.gpg"
```
matrei: Here we are setting `ext` within `ext`?
Align on single quotes?

grails-gradle/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    isCiBuild = System.getenv().get('CI') as Boolean
+}
+if (project.ext.isReleaseVersion) {
```
matrei: No need to reference `project.ext` here?
`if (isReleaseVersion)` should be enough?

grails-gradle/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+        repositories {
+            sonatype {
+                def ossUser = System.getenv("SONATYPE_USERNAME") ?: project.hasProperty("sonatypeOssUsername") ? project.sonatypeOssUsername : ''
+                def ossPass = System.getenv("SONATYPE_PASSWORD") ?: project.hasProperty("sonatypeOssPassword") ? project.sonatypeOssPassword : ''
+                def ossStagingProfileId = System.getenv("SONATYPE_STAGING_PROFILE_ID") ?: project.hasProperty("sonatypeOssStagingProfileId") ? project.sonatypeOssStagingProfileId : ''
+                nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
+                username = ossUser
+                password = ossPass
+                stagingProfileId = ossStagingProfileId
+            }
```
matrei: Simplify?
```groovy
sonatype {
    nexusUrl = uri('https://s01.oss.sonatype.org/service/local')
    username = System.getenv('SONATYPE_USERNAME') ?: project.findProperty('sonatypeOssUsername')
    password = System.getenv('SONATYPE_PASSWORD') ?: project.findProperty('sonatypeOssPassword')
    stagingProfileId = System.getenv('SONATYPE_STAGING_PROFILE_ID') ?: project.findProperty('sonatypeOssStagingProfileId')
}
```

grails-gradle/buildSrc/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    // mavenLocal()
+    mavenCentral()
+    gradlePluginPortal()
+}
+
+dependencies {
+//    implementation "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:$gradleProperties.gradleLicensePluginVersion", {
+//        // Due to https://github.com/hierynomus/license-gradle-plugin/issues/161, spring must be excluded
+//        exclude group: 'org.springframework', module: 'spring-core'
+//    }
```
matrei: Remove comment?

grails-gradle/dependencies.gradle
```diff
@@ -0,0 +1,58 @@
...
+            'objenesis.version'                  : '3.4',
+            'profiles.version'                   : '10.0.2',
+            'gradle-spock.version'               : '2.3-groovy-3.0',
+            'gradle-nexus-publish-plugin.version': '2.0.0',
+            'spring-boot.version'                : '3.4.2',
+            'springloaded.version'               : '1.2.8.RELEASE',
+    ]
+
+    // Note: the name of the dependency must be the prefix of the property name so properties in the pom are resolved correctly
+    bomPlatformDependencies = [
```
matrei, line 31: Any particular reason to use snake_case instead of kebab-case for the map keys?

grails-gradle/docs-core/build.gradle
```diff
@@ -0,0 +1,128 @@
...
+            maven {
+                credentials {
+                    def u = System.getenv("MAVEN_PUBLISH_USERNAME") ?: ''
+                    def p = System.getenv("MAVEN_PUBLISH_PASSWORD") ?: ''
+                    username = u
+                    password = p
+                }
+                url System.getenv("MAVEN_PUBLISH_URL") ?: 'https://repository.apache.org/content/repositories/snapshots'
+            }
+        }
```
matrei: Simplify?
```groovy
maven {
    credentials {
        username = System.getenv('MAVEN_PUBLISH_USERNAME')
        password = System.getenv('MAVEN_PUBLISH_PASSWORD')
    }
    url = uri(System.getenv('MAVEN_PUBLISH_URL') ?: 'https://repository.apache.org/content/repositories/snapshots')
}
```

grails-gradle/gradle.properties
```diff
@@ -0,0 +1,8 @@
+# This groovy version should be the version built with grails applications and NOT the one used by this project
+# since this project must use gradle's groovy version.  This property will be used to set up the test projects
+# like grails would be to ensure proper test coverage
+e2eGroovyVersion=4.0.24
```
matrei, line 4: Should we be able to resolve this version now?

grails-gradle/gradle/docs-config.gradle
```diff
@@ -0,0 +1,39 @@
...
+    classpath = configurations.documentation
+    groovyClasspath = configurations.documentation
+    access = GroovydocAccess.PROTECTED
+    includeAuthor = true
+    includeMainForScripts = false
+    processScripts = false
+}
+
+configure([groovydoc]) {
+    destinationDir = project.file('build/docs/api')
```
matrei, line 24: Move to the configuration block above?

grails-gradle/gradle/docs-config.gradle
```diff
@@ -0,0 +1,39 @@
...
+}
+
+configure([javadoc]) {
+    options.encoding 'UTF-8'
+    options.docEncoding 'UTF-8'
+    options.charSet 'UTF-8'
+    options.jFlags '-Xms64M', '-Xmx512M'
+}
+
+tasks.named('build').configure {
```
matrei: No need to use `.configure`, as it is implied?

grails-gradle/gradle/docs-config.gradle
```diff
@@ -0,0 +1,39 @@
...
+configure([groovydoc]) {
+    destinationDir = project.file('build/docs/api')
+}
+
+configure([javadoc]) {
+    options.encoding 'UTF-8'
+    options.docEncoding 'UTF-8'
+    options.charSet 'UTF-8'
+    options.jFlags '-Xms64M', '-Xmx512M'
+}
```
matrei: Modernize?
```groovy
tasks.named('javadoc', Javadoc) {
    (options as StandardJavadocDocletOptions).with {
        encoding = 'UTF-8'
        docEncoding = 'UTF-8'
        charSet = 'UTF-8'
        addStringOption('-Xms64M')
        addStringOption('-Xmx512M')
    }
}
```

grails-gradle/gradle/java-config.gradle
```diff
@@ -0,0 +1,30 @@
...
+
+extensions.configure(JavaPluginExtension) {
+    // Explicit `it` is required here
+    it.withJavadocJar()
+    it.withJavadocJar()
+}
+
+tasks.withType(JavaCompile).configureEach {
+    it.options.deprecation = true
+    it.options.debug = true
```
matrei: Remove explicit `it`?

grails-gradle/model/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+        repositories {
+            maven {
+                credentials {
+                    def u = System.getenv("MAVEN_PUBLISH_USERNAME") ?: ''
+                    def p = System.getenv("MAVEN_PUBLISH_PASSWORD") ?: ''
+                    username = u
+                    password = p
+                }
+                url System.getenv("MAVEN_PUBLISH_URL") ?: 'https://repository.apache.org/content/repositories/snapshots'
+            }
```
matrei: Simplify like above example?

grails-gradle/model/src/main/groovy/grails/io/IOUtils.groovy
```diff
@@ -67,7 +67,7 @@ class IOUtils extends SpringIOUtils {
      */
     static String toString(Reader reader) {
         def writer = new StringWriter()
-        copy reader, writer
+        org.grails.io.support.SpringIOUtils.copy reader, writer
```
matrei: Do we need a qualified reference (`SpringIOUtils` is imported)?
Use method parentheses?

grails-gradle/model/src/main/groovy/grails/io/IOUtils.groovy
```diff
@@ -90,7 +90,7 @@ class IOUtils extends SpringIOUtils {
      */
     static void copy(InputStream input, Writer output, String encoding = null) {
         def reader = encoding ? new InputStreamReader(input, encoding) : new InputStreamReader(input)
-        copy(reader, output)
+        org.grails.io.support.SpringIOUtils.copy(reader, output)
```
matrei: Do we need a qualified reference (`SpringIOUtils` is imported)?

grails-gradle/model/src/main/groovy/grails/io/IOUtils.groovy
matrei, line 1: Update license header year range?

grails-gradle/model/src/main/groovy/org/grails/io/support/DefaultResourceLoader.java
matrei, line 1: Update license header year range (if we do the suggested changes in the file)?

grails-gradle/model/src/main/groovy/org/grails/io/support/DefaultResourceLoader.java
```diff
@@ -84,8 +84,8 @@ public ClassLoader getClassLoader() {
     }
 
     public Resource getResource(String location) {
-        if (location.startsWith(CLASSPATH_URL_PREFIX)) {
-            return new ClassPathResource(location.substring(CLASSPATH_URL_PREFIX.length()), getClassLoader());
+        if (location.startsWith(ResourceLoader.CLASSPATH_URL_PREFIX)) {
+            return new ClassPathResource(location.substring(ResourceLoader.CLASSPATH_URL_PREFIX.length()), getClassLoader());
```
matrei: Do we need the explicit reference when we are implementing the interface.

grails-gradle/plugins/build.gradle
```diff
@@ -0,0 +1,169 @@
...
+    if (!isReleaseVersion) {
+        repositories {
+            maven {
+                credentials {
+                    def u = System.getenv("MAVEN_PUBLISH_USERNAME") ?: ''
+                    def p = System.getenv("MAVEN_PUBLISH_PASSWORD") ?: ''
+                    username = u
+                    password = p
+                }
+                url System.getenv("MAVEN_PUBLISH_URL") ?: 'https://repository.apache.org/content/repositories/snapshots'
```
matrei: Simplify according to above suggestions?

grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy
matrei, line 1: Update license header year range?

grails-shell-cli/src/main/groovy/org/grails/cli/gradle/GradleInvoker.groovy
matrei, line 1: Update license header year range?

grails-web-core/src/main/groovy/org/grails/web/servlet/boostrap/DefaultGrailsBootstrapClass.java
matrei, line 1: Update license header year range?

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
matrei, line 1: Update license header year range?

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
matrei, line 1: Update license header year range?

matrei (approved):
Thank you @jdaugherty!

.github/workflows/gradle.yml
jdaugherty: build, etc have a dependency on skip check so it's not really needed.

.github/workflows/gradle.yml
jdaugherty: Yes, since it's not using the publish plugin.

dependencies.gradle
jdaugherty: It's to match the properties upstream with spring boot.  I wish we could use camel case, but it could be confusing to mix these with users.

dependencies.gradle
jdaugherty: See previous comment - it's to match the style of spring boot's bom (also jackson, and groovy)

.github/workflows/gradle.yml
matrei: Yes, but in order to reference it, does it not need to be included?

grails-bootstrap/src/test/groovy/grails/io/IOUtilsSpec.groovy
jdaugherty: Good catch

.github/workflows/gradle.yml
matrei: OK, but are we generating insecure checksums?

grails-gradle/build.gradle
jdaugherty: for quoted keys you have to use ext, but not for isRelease version.

grails-gradle/buildSrc/build.gradle
jdaugherty: The only reason i left this was in case we wanted to run the licnese plugin here since some of the dependencies are defined in this parent project.  I'll remove it for now, but we may want to revisit this.

grails-gradle/dependencies.gradle
jdaugherty: the only reason i used underscore was because this was from the example & underscore doesn't require quoting it.  I can swtich to kebab to be consistent

grails-gradle/build.gradle
matrei: I think we should extract the quoted keys and not use:
```groovy
ext {
    ext.'mykey' = 'myvalue'
}
```

grails-gradle/build.gradle
```diff
@@ -11,40 +11,36 @@ allprojects {
...
-                def ossPass = System.getenv("SONATYPE_PASSWORD") ?: project.hasProperty("sonatypeOssPassword") ? project.sonatypeOssPassword : ''
-                def ossStagingProfileId = System.getenv("SONATYPE_STAGING_PROFILE_ID") ?: project.hasProperty("sonatypeOssStagingProfileId") ? project.sonatypeOssStagingProfileId : ''
-                nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
-                username = ossUser
-                password = ossPass
-                stagingProfileId = ossStagingProfileId
+                nexusUrl = uri('https://s01.oss.sonatype.org/service/local/')
+                username = System.getenv('SONATYPE_USERNAME') ?: project.hasProperty('sonatypeOssUsername') ? project.findProperty('sonatypeOssUsername') : ''
+                password = System.getenv('SONATYPE_PASSWORD') ?: project.hasProperty('sonatypeOssPassword') ? project.findProperty('sonatypeOssPassword') : ''
+                stagingProfileId = System.getenv('SONATYPE_STAGING_PROFILE_ID') ?: project.hasProperty('sonatypeOssStagingProfileId') ? project.findProperty('sonatypeOssStagingProfileId') : ''
```
matrei: We don't need the `hasProperty`.

grails-events/plugin/build.gradle
```diff
@@ -24,12 +24,13 @@ dependencies {
...
+    compileOnly 'org.apache.grails.gradle:grails-gradle-model'
     compileOnly project(':grails-core')
 
     testImplementation 'org.spockframework:spock-core'
 
-    testRuntimeOnly project(':grails-bootstrap')
+    testRuntimeOnly 'org.apache.grails.gradle:grails-gradle-model'
+    testRuntimeOnly project(':grails-core')
```
matrei: This is not needed actually.

grails-gradle/gradle.properties
jdaugherty: No, we intentionally do not resolve the e2e groovy version.  These projects do not use the grails bom, etc and they simply include the verison of groovy.  Originally we inlined the groovy version, but feedback was given to extract it into a property.

PR #14317, opened 2025-04-17
Test fixes & grails-bom populating parent boms

jdaugherty:


jdaugherty:
I'm working on the data mapping changes and realized the bom isn't populating parent boms.  I'm going to go ahead and merge this but please let me know if you object @jamesfredley @matrei

PR #14326, opened 2025-04-21
License Audit - First Pass

jdaugherty:
Initial non-java/groovy source items are done.  Including:
1. adoc
2. properties
3. yml
4. xml
5. markdown
6. initial notice file

I have left the original author in the translation files to have a reference: 
* message.properties had the header added, but there are messages like this still in the files: 

          #
          # Translated by X - X@example.com
          #


jamesfredley (approved):
Remaining file extensions for second pass

css
gml
groovy
gson
gsp
java
js
jsp
map
no extension
sdkmanrc
sh
xml
xsd
yaml

I think we will also want to remove jquery, bootstrap, etc. via webjars in the test apps.  I will be glad to do that in the second pass PR.

PR #14668, opened 2025-04-24
CI, Build and Test updates

matrei:
- Stabilize flaky `CustomAutoTimestampSpec`
  - Address intermittent CI failures possibly caused by date values not incrementing as expected.
  - Enhance assertions by comparing `.time` values to reveal the actual millisecond differences.
  - Remove annotations that were added as a try to remedy the flakiness, to determine if they are warranted.
  - Omit cleanup of `insertOverwrite`, to determine if it is necessary for test stability.
  - Do one comparison per statement in `then` blocks.
- Fix `GroovyPageUnitTestMixinWithCustomViewDirSpec` on Windows
  - The test was failing on Windows due to line ending differences affecting string comparisons.
- Apply `asset-pipeline` Gradle plugin in some test applications with assets
  - This puts `org.graalvm.js:js-community` on the `runtimeClasspath` and avoids `ClassNotFoundException: org.graalvm.polyglot.Context`.
- Re-enable `TckTestSuite` on Windows as it does not seem to fail
- Update `gradle.yml`
- Fix an incorrect log pattern in functional test applications
- Update deprecated `converterClass` attributes in `logback.xml`
- Fix develocity authentication in `grails-gradle`
- Switch some Gradle log statements from `lifecycle` to `info` to clean up the log output a bit.
- Update Gradle Plugin e2e tests to use Groovy 2.0.25.
- Add `mavenCentral` repository to Gradle Plugin e2e tests.
- Switch to using `@PendingFeatureIf` for tests that only fail in `CI`.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ '17', '21', '23' ]
+        java: [ 17, 21, 23 ]
     runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -31,8 +31,6 @@ env:
 jobs:
   buildGradle:
     name: 'Build Gradle Plugins'
-    permissions:
-      contents: read  #  to fetch code (actions/checkout)
     strategy:
       fail-fast: false
       matrix:
@@ -57,8 +55,6 @@ jobs:
           --continue --stacktrace
   build:
     name: 'Build Grails-Core'
-    permissions:
-      contents: read  #  to fetch code (actions/checkout)
     strategy:
       fail-fast: false
       matrix:
@@ -86,8 +82,6 @@ jobs:
           -PskipHibernate5Tests
   functional:
     name: "Functional Tests"
-    permissions:
-      contents: read  #  to fetch code (actions/checkout)
     strategy:
       fail-fast: false
       matrix:
@@ -169,8 +163,6 @@ jobs:
   publishGradle:
     if: github.repository_owner == 'apache' && github.event_name == 'push'
     needs: [ buildGradle ]
-    permissions:
-      contents: read # limit to read access
     runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"
@@ -197,8 +189,6 @@ jobs:
   publish:
     if: github.repository_owner == 'apache' && github.event_name == 'push'
     needs: [ publishGradle, build, functional, hibernate5Functional, mongodbFunctional ]
-    permissions:
-      contents: read # limit to read access
     runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"
@@ -223,8 +213,6 @@ jobs:
     if: github.repository_owner == 'apache' && github.event_name == 'push'
     needs: [ publish ]
     runs-on: ubuntu-24.04
-    permissions:
-      contents: write # TODO: Comment why this is needed
     steps:
       - name: "📥 Checkout the repository"
         uses: actions/checkout@v4
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -25,7 +25,8 @@ on:
       - '[3-9]+.[3-9]+.x'
   workflow_dispatch:
 env:
-  # to prevent throttling of the github api, include the github token in an environment variable since the build will check for it
+  # To prevent throttling of GitHub API usage, include the GitHub token as an environment variable
+  # as the build will use it when fetching git tags in CreateReleaseDropDown.groovy
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 jobs:
   buildGradle:
@@ -35,9 +36,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [17, 21]
-        os: [ubuntu-latest]
-    runs-on: ${{ matrix.os }}
+        java: [ 17, 21 ]
+    runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"
         uses: actions/checkout@v4
@@ -52,7 +52,9 @@ jobs:
           develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY  }}
       - name: "🔨 Build project"
         working-directory: 'grails-gradle'
-        run: ./gradlew build assemble --continue --stacktrace
+        run: >
+          ./gradlew build assemble
+          --continue --stacktrace
   build:
     name: 'Build Grails-Core'
     permissions:
@@ -60,8 +62,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [17, 21, 23]
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        java: [ 17, 21, 23 ]
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: "📥 Checkout repository"
@@ -76,14 +78,12 @@ jobs:
         with:
           develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY  }}
       - name: "🔨 Build project"
-        id: build
         run: >
-          ./gradlew build 
-          assemble 
-          groovydoc 
-          --continue 
-          --stacktrace 
-          -PskipFunctionalTests -PskipMongodbTests -PskipHibernate5Tests
+          ./gradlew build assemble groovydoc
+          --continue --stacktrace
+          -PskipFunctionalTests
+          -PskipMongodbTests
+          -PskipHibernate5Tests
   functional:
     name: "Functional Tests"
     permissions:
@@ -92,7 +92,7 @@ jobs:
       fail-fast: false
       matrix:
         java: [ '17', '21', '23' ]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"
         uses: actions/checkout@v4
@@ -105,12 +105,13 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
         with:
           develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY  }}
-      - name: "🔨 Functional Tests"
+      - name: "🏃 Run Functional Tests"
         run: >
-          ./gradlew 
-          build 
-          --stacktrace 
-          -PonlyFunctionalTests -PskipMongodbTests -PskipHibernate5Tests
+          ./gradlew bootJar check
+          --continue --stacktrace
+          -PonlyFunctionalTests
+          -PskipHibernate5Tests
+          -PskipMongodbTests
   mongodbFunctional:
     name: "Mongodb Functional Tests"
     runs-on: ubuntu-24.04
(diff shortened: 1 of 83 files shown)


.github/workflows/gradle.yml
```diff
@@ -212,20 +213,18 @@ jobs:
...
+        run: ./gradlew publish
   docs:
     if: github.repository_owner == 'apache' && github.event_name == 'push'
-    needs: [publish]
-    runs-on: ubuntu-latest
+    needs: [ publish ]
+    runs-on: ubuntu-24.04
     permissions:
-      contents: write
+      contents: write # TODO: Comment why this is needed
```
jdaugherty: This isn't needed because we will commit to a different repo

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/TckTestSuite.groovy
```diff
@@ -21,14 +21,12 @@ package org.grails.datastore.gorm
...
 import org.junit.platform.suite.api.SelectClasses
 import org.junit.platform.suite.api.Suite
-import spock.lang.IgnoreIf
 
 /**
  * Use this class to run tck classes against the current implementation.
  *
  * @author graemerocher
  */
-@IgnoreIf({ os.windows }) // Fails on windows even though the NotInListSpec is in the TCK and runs
```
jdaugherty, line 31: I ran this multiple times and it kept failing in CI.  James F couldn't reproduce locally though.  We never could explain why it was failing.  We don't need it for CI, so I suggest we just disable if GITHUB_REF is present

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -94,6 +93,7 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
         def now = new Date()
         r.created = new Date(now.time)
         r.modified = r.created
+        sleep(50) // give the save a chance to set a different time
```
jdaugherty: Can't this just be sleep(1)?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -22,15 +22,9 @@ import grails.gorm.annotation.AutoTimestamp
 import grails.gorm.tests.GormDatastoreSpec
 import grails.persistence.Entity
 import org.grails.datastore.gorm.events.AutoTimestampEventListener
-import spock.lang.IgnoreIf
-import spock.lang.Retry
-import spock.lang.Stepwise
 
 import static grails.gorm.annotation.AutoTimestamp.EventType.CREATED
 
-@Retry // this test is flaky on CI due to https://github.com/apache/grails-data-mapping/issues/1877
```
jdaugherty, line 31: I think this is still failing because there is a shared, static variable thats being updated separately by another test.  The comment should stay

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -41,19 +35,21 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
             r = RecordCustom.get(r.id)
 
         then:"the custom lastUpdated and dateCreated are set"
-            r.modified != null && r.modified < new Date()
-            r.created != null && r.created < new Date()
+            r.modified != null
```
jdaugherty, line 39: You could add a sleep(1) and then restore these time checks

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -106,16 +106,15 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
         Date previousCreated = r.created
         Date previousModified = r.modified
         r.name = "Test 2"
+        sleep(50) // give the save a chance to set a different time
```
jdaugherty: Doesn't this just need to be sleep(1)?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -63,24 +59,27 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
...
-        now < r.modified
-        now < r.created
+        now.time < r.modified.time
+        now.time < r.created.time
 
         when:"An entity is modified"
         Date previousCreated = r.created
         Date previousModified = r.modified
         r.name = "Test 2"
+        sleep(50) // give the save a chance to set a different time
```
jdaugherty: sleep(1)?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -63,24 +59,27 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
         def now = new Date()
         r.created = new Date(now.time)
         r.modified = r.created
+        sleep(50) // give the save a chance to set a different time
```
jdaugherty: sleep(1)?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
```diff
@@ -41,19 +35,21 @@ class CustomAutoTimestampSpec extends GormDatastoreSpec {
...
-            r.modified != null && r.modified < new Date()
-            r.created != null && r.created < new Date()
+            r.modified != null
+            r.created != null
 
         when:"An entity is modified"
             Date previousCreated = r.created
             Date previousModified = r.modified
             r.name = "Test 2"
+            sleep(50) // give the save a chance to set a different time
```
jdaugherty: sleep(1)

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/TckTestSuite.groovy
matrei: It seems it is no longer failing in CI, so we should not need to disable it at all.

.github/workflows/gradle.yml
matrei: OK, so I can remove it?

I also see some `contents: read  #  to fetch code (actions/checkout)`
These should not be necessary, should they?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
matrei: Probably, I'll try it out.

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
matrei: Is it still failing? Where?

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
matrei: I don't think the time checks are necessary to fulfill the contract of what is tested.
That is why I removed it. Do you think I should add it back?

.github/workflows/gradle.yml
jdaugherty: read is still needed to checkout, write isn't

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
jdaugherty: it fails in parallel tests; i have had it fail locally sporadically and it seems to consistently fail on the windows 23 build

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
jdaugherty: I think we should

grails-gradle/plugins/src/e2eTest/resources/publish-projects/legacy-apply/child-project-with-unrelated-parent/build.gradle
```diff
@@ -20,6 +20,7 @@
 buildscript {
     repositories {
         maven { url "${System.getenv('LOCAL_MAVEN_PATH')}" }
+        mavenCentral()
```
jdaugherty, line 23: We should remove the grails repo & the apache repo.  They shouldnt' be needed.  the purpose of these tests were to have a minimum build file (so just include groovy as a minimum dependency), and the gradle plugin should be substituted by the gradle test kit.

jdaugherty, line 23: We should remove the grails repo & the apache repo.  They shouldnt' be needed.  the purpose of these tests were to have a minimum build file (so just include groovy as a minimum dependency), and the gradle plugin should be substituted by the gradle test kit.

grails-test-suite-uber/src/test/groovy/grails/test/mixin/GroovyPageUnitTestMixinWithCustomViewDirSpec.groovy
```diff
@@ -35,6 +35,6 @@ class GroovyPageUnitTestMixinWithCustomViewDirSpec extends Specification impleme
         def result = render(template: '/demo/myTemplate')
         
         then:
-        result == '\nthis is a custom template'
+        result.contains('this is a custom template')
```
jdaugherty: Should we just call trim() and compare the resulting string instead?

.github/workflows/gradle.yml
matrei: I think read is the default, so it is implied if you don't specify anything: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

jdaugherty: Le'ts give it a try then!

grails-test-suite-uber/src/test/groovy/grails/test/mixin/GroovyPageUnitTestMixinWithCustomViewDirSpec.groovy
matrei: OK, makes sense!

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/TckTestSuite.groovy
jdaugherty: We can try it, but in my deterministic ordering PR, I had it fail twice randomly.

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
matrei: >  it seems to consistently fail on the windows 23 build

I have not seen `CustomAutoTimestampSpec` fail even once on Windows/23 after the changes in this branch: https://github.com/apache/grails-core/actions/runs/14657676325/job/41135500844?pr=14668

Do I have to change anything to the test setup to trigger the failure?

grails-gradle/plugins/src/e2eTest/resources/publish-projects/legacy-apply/child-project-with-unrelated-parent/build.gradle
matrei: It seems the tests fail to find all dependencies if the repositories are removed.

jdaugherty: thats unfortunate; thank you for trying

grails-datamapping-core-test/src/test/groovy/org/grails/datastore/gorm/CustomAutoTimestampSpec.groovy
jdaugherty: Let's try it with your changes and then if it fails, I'll ignore it again?

jdaugherty (approved):

PR #14670, opened 2025-04-24
PR #1 - Working towards reproducible builds

jdaugherty:
This PR is working towards reproducible builds.  

You can run the script: 

       etc/bin/test-reproducible-build.sh

to do 2 builds, back to back, and compare their jar files to see what does not match.  This PR does not fully solve  reproducible grails build.  

It does: 
1. Set `SOURCE_DATE_EPOCH` in github actions to ensure the build date is always as of the commit date 
2. Use SOURCE_DATE_EPOCH in the gradle build, and set it as properties in case future processes need to use them
3. Adds a helper groovy script `generate-build-artifact-hashes` to generate hashes for any jar files that are considered distributed
4. adds the test script previously mentioned 
5. fixes sourcejar configuration
6. sets common encodings so across all platforms sources / docs /etc are treated the same
7. Removes "Built-By" and "Created-By" manifest attributes for reproducible manifests
8. Configures jars per Gradle's reproducibility recommendations: i.e. do not preserve timestamps, reproducible file order, and common permissions
9. disables jar files in projects that "staging" projects, such as documentation builds, test reporting, etc
10. Changes the AST transforms so that they do not make use of HashSet / HashMap (determined field ordering)
11. AST transforms are processed in a specific order with Groovy 4's TransformWithPriority improvement
12. Introduces a grails-common build for sharing classes between gorm & grails-core that have no specific grails dependency


matrei (commented):
Is it expected that all "Java CI" checks should fail?

jamesfredley:
The failure is coming from https://github.com/apache/grails-core/commit/0baef55ba86eb1e38bc84e615ffb5aef20e93c9a

jdaugherty:
I don't think my revert happened correctly

jdaugherty:
@matrei  I believe my reverts in core are correct now.  Apologies,I had push upstream initially instead of to this branch.

jdaugherty:
One of my discoveries  with this change is that the Gorm transforms expect to alternate;  that is if you have the following nodes to visit: 

            ClassNode BookService
            MethodNode countBooks

Then it's expected that the application order be: 

            For Class Node, apply Transactional Transform & Tenant Transform (simplified)
            For Method Node, apply Transactional Transform & Tenant Transform (simplified)

The Groovy 4 improvement (TransformWithPriority) is to apply in this order: 

            1. Class Node - Transactional Transform
            2. Method Node - Transactional Transform
            3. Class Node - Tenant Transform
            4. Method Node - Tenant Transform

We can't completely switch to the to TransformWithPriority because the transaction transform looks for values set by the previous visit to Class node.  You would get an error similar to "BookService#tenantId access is forbidden" and it would fail to compile. 

I've gone ahead and tried to document the run order in GroovyTransformOrder, and since both the groovy 4 TransformWithPriority & the grails datastore Ordered expect a higher value to be higher priority, they can be used interchangeably.  

A major side effect of this change is the transforms should still be applied in the same order, every time.  This should eliminate side effect compile behavior where they would execute in a different order pre groovy 4.  

Also, there still are differences amongst a generated domain class after this change.  I've tried to switch to ordered collections where possible, but there's still something along the domain path that's cause it to be different each time.  I am continuing to research ...

jdaugherty:
### Jars with Differences

		grails-async/core/build/libs/grails-async-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-cache/build/libs/grails-cache-7.0.0-SNAPSHOT.jar
		grails-controllers/build/libs/grails-controllers-7.0.0-SNAPSHOT.jar
		grails-converters/build/libs/grails-converters-7.0.0-SNAPSHOT.jar
		grails-core/build/libs/grails-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-data-hibernate5/core/build/libs/grails-data-hibernate5-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-data-hibernate5/dbmigration/build/libs/grails-data-hibernate5-dbmigration-7.0.0-SNAPSHOT.jar
		grails-data-hibernate5/grails-plugin/build/libs/grails-data-hibernate5-7.0.0-SNAPSHOT.jar
		grails-data-mongodb/bson/build/libs/grails-data-mongodb-bson-7.0.0-SNAPSHOT.jar
		grails-datamapping-core/build/libs/grails-datamapping-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-datamapping-core/build/libs/grails-datamapping-core-7.0.0-SNAPSHOT.jar
		grails-datamapping-support/build/libs/grails-datamapping-support-7.0.0-SNAPSHOT.jar
		grails-datamapping-tck-base/build/libs/grails-datamapping-tck-base-7.0.0-SNAPSHOT-javadoc.jar
		grails-datamapping-tck-domains/build/libs/grails-datamapping-tck-domains-7.0.0-SNAPSHOT.jar
		grails-datamapping-tck-tests/build/libs/grails-datamapping-tck-tests-7.0.0-SNAPSHOT-javadoc.jar
		grails-datamapping-tck-tests/build/libs/grails-datamapping-tck-tests-7.0.0-SNAPSHOT.jar
		grails-datastore-core/build/libs/grails-datastore-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-domain-class/build/libs/grails-domain-class-7.0.0-SNAPSHOT.jar
		grails-fields/build/libs/grails-fields-7.0.0-SNAPSHOT.jar
		grails-gsp/core/build/libs/grails-gsp-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-gsp/grails-sitemesh3/build/libs/grails-sitemesh3-7.0.0-SNAPSHOT.jar
		grails-gsp/grails-taglib/build/libs/grails-taglib-7.0.0-SNAPSHOT-javadoc.jar
		grails-gsp/grails-web-gsp-taglib/build/libs/grails-web-gsp-taglib-7.0.0-SNAPSHOT.jar
		grails-gsp/grails-web-gsp/build/libs/grails-web-gsp-7.0.0-SNAPSHOT-javadoc.jar
		grails-gsp/plugin/build/libs/grails-gsp-7.0.0-SNAPSHOT.jar
		grails-interceptors/build/libs/grails-interceptors-7.0.0-SNAPSHOT.jar
		grails-rest-transforms/build/libs/grails-rest-transforms-7.0.0-SNAPSHOT.jar
		grails-scaffolding/build/libs/grails-scaffolding-7.0.0-SNAPSHOT.jar
		grails-shell-cli/build/libs/grails-shell-cli-7.0.0-SNAPSHOT-javadoc.jar
		grails-test-core/build/libs/grails-test-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-views-core/build/libs/grails-views-core-7.0.0-SNAPSHOT-javadoc.jar
		grails-views-gson/build/libs/grails-views-gson-7.0.0-SNAPSHOT.jar
		grails-views-markup/build/libs/grails-views-markup-7.0.0-SNAPSHOT-javadoc.jar
		grails-views-markup/build/libs/grails-views-markup-7.0.0-SNAPSHOT.jar
		grails-web-common/build/libs/grails-web-common-7.0.0-SNAPSHOT-javadoc.jar
		grails-web-databinding/build/libs/grails-web-databinding-7.0.0-SNAPSHOT.jar
		grails-web-url-mappings/build/libs/grails-web-url-mappings-7.0.0-SNAPSHOT-javadoc.jar
		grails-web-url-mappings/build/libs/grails-web-url-mappings-7.0.0-SNAPSHOT.jar

### To investigate these issues, the following strategies can be employed: 
1. diff the jars to find what's different by unzipping them and then:
- for class files, decompile the class with IntelliJ and diff the result.  Reordering of the fields seems to be the most common issue.
- for binary files, use hex comparison to see how much is different
2. ensure common execution order by adding output to the various *Transform classes.  Note: OrderedGormTransformation  runs all of transforms for annotations with an annotation `GormASTTransformationClass` 
3. Ensure collections with reproducible order are used in transforms or any class changing an AST node. (I think most of this is done in this PR)


### Known Outstanding issues: 
1. the TCK implementation causes differences in the javadoc.jar.  

It appears the parent methods that spock adds to a spec, are out of order.  Based on the previous TCK work, and how it loads the same class name depending on the project, it seems the best action here may be to finally work on an extension or custom runner to run TCK based tests.  

The tests can continue to be shipped in a single jar file, but the base jar needs to do something different - maybe have a common class that's looked for and the behavior of the tests is influenced by that? 

2. domains still are not reproducible.  I'm seeing reordering of the fields.  It's probably related to the transforms being applied to each AnnotatedNode in bulk instead of applying 1, then another.  Decoupling TransactionalTransform & TenantTransform may help here.

jdaugherty:
I have not done any research on gson / markup / gsp differences. Some projects ship with gsps, so we'll need to also investigate those.

jdaugherty:
@jamesfredley @matrei I think we should merge this PR (assuming tests pass) and then create a ticket to track this in the long run.  If we have any volunteers to dig into the domain issue, that's likely the largest of all of these.

matrei:
Looks like tests pass. Feel free to merge.
I won't be able to look at the changes today, possibly tomorrow. Seems like we got some guidance in the grails-dev mailing list: https://lists.apache.org/thread/dvgm8n1r3okzvbdntdqomyylfh5x1l1w

PR #14671, opened 2025-04-24
use webjars dependencies to provide jquery and bootstrap js libraries

jamesfredley:
`popper.js` was not being used in these project, that I found.

Some test grails-apps needed the asset-pipeline gradle plugin applied.


grails-data-neo4j/examples/grails3-neo4j-hibernate/build.gradle
```diff
@@ -80,3 +84,7 @@ webdriverBinaries {
         geckodriver "$geckodriverVersion"
     }
 }
+
+apply {
+    from rootProject.layout.projectDirectory.file('gradle/webjar-asset-pipeline-config.gradle')
```
jdaugherty: I had avoided updating neo4j since it was only partially updated.  I assume these were just mass replaces?  Should we update all of these then or leave it for now?

gradle/webjar-asset-pipeline-config.gradle
jdaugherty, line 1: Can we rename this file to some how indicate it's only for functional and / or tests?  Maybe 'test-webjar-asset-config.gradle' ...

jdaugherty: actually, why can't this just be in functional-test-config?  Aren't we only ever uses assets in functional tests?

jdaugherty (commented):
I would just move the asset code into the functional test gradle config file.

You can leave the changes you made under grails-data-neo4j as-is since it's not included in the project.  But i would remove the webjar gradle config file.  Whenever someone tries to port neo4j, they'll just have to move those test projects and update.

grails-data-neo4j/examples/grails3-neo4j-hibernate/build.gradle
jamesfredley: We need to remove the jquery and bootstrap files, so I did the minimal amount of surgery for that purpose.  The projects will not work of course, since it has not been updated, but this small part will be there when someone has time to work on updating it for Grails 7.

gradle/webjar-asset-pipeline-config.gradle
jamesfredley: Glad to update the name.  It is only used in functional tests and those two in grails-data-neo4j which have not been updated and moved.  But it is not used in all of the functional tests, which was the reason for a separate config file.

jamesfredley: Also if we put this on all functional tests, we would have to add the `asset-pipeline` gradle plugin to all of them, including the spring boot only ones.

jamesfredley: I guess we could check if the plugin has been applied, though.

jdaugherty (approved):

PR #14680, opened 2025-04-29
Update to Gradle 8.14, Spring Boot 3.4.5 and Groovy 4.0.26

### LINKED ISSUE #14050: Groovy 4.0.26 and Spring Boot 3.4.3/Spring Framework 6.2.3
### Issue description

After milestone 2, update the following dependencies and resolve any compatibility issues.

Groovy 4.0.26
Spring Boot 3.4.3
Spring Framework 6.2.3

jamesfredley:
The PR currently updates the versions and does not attempt to consolidate the version locations, which likely should also be addressed.

There were two failing tests on Windows 11 for the external configuration tests.  Both were updated to be cross platform compatible, but one still fails and is ignored for now.


jdaugherty (approved):

jdaugherty:
Note: to stop finding versions in the old gradle projects that haven't been ported yet, I removed the assignment; we can fix if we ever bring it back into the gradle project

PR #14681, opened 2025-04-30
Use /content/groups/snapshot instead of /content/repositories/snapshots for access

jamesfredley:
https://repository.apache.org/content/groups/snapshot lists all snapshots instead of latest and will be used for access, but not for publishing

`in Nexus; groups are for access, not publishing. Instead, artifacts are published to hosted repositories like /repositories/snapshots, and the group /groups/snapshots aggregates this content, making it accessible via its URL.`

This also includes changes to 4 tests which are now passing after recent changes.

`grails-doc:docs` intermittent, unrelated failure:
 https://github.com/apache/grails-core/issues/14682

### DIFF
--- a/grails-gradle/bom/build.gradle
+++ b/grails-gradle/bom/build.gradle
@@ -112,7 +112,7 @@ publishing {
                     username = System.getenv('MAVEN_PUBLISH_USERNAME')
                     password = System.getenv('MAVEN_PUBLISH_PASSWORD')
                 }
-                url = System.getenv('MAVEN_PUBLISH_URL') ?: 'https://repository.apache.org/content/repositories/snapshots'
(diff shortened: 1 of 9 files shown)


matrei (commented):
Can you also publish to this url?

jdaugherty (approved):

jamesfredley:
@matrei I found a definite answer and am removing those changes.

> in Nexus; groups are for access, not publishing. Instead, artifacts are published to hosted repositories like /repositories/snapshots, and the group /groups/snapshots aggregates this content, making it accessible via its URL.

jamesfredley:
I put the `grails-doc:docs` failure in:

https://github.com/apache/grails-core/issues/14682

matrei (approved):

Issue #14138, opened 2025-03-31
Allow publishing only dependency / boms from the grails publish plugin


jdaugherty:
See https://github.com/apache/grails-core/pull/14093#discussion_r2020163558

For javaPlatforms, we should be able to publish with no sources.

PR #14684, opened 2025-05-01
Wrapper & Profiles merge

### LINKED ISSUE #14687: publishMavenPublicationToMavenLocal on rest-api-plugin and web-plugin fails
### Issue description

```
Execution failed for task ':grails-profiles-rest-api-plugin:publishMavenPublicationToMavenLocal'.
> Failed to publish publication 'maven' to repository 'mavenLocal'
   > Invalid publication 'maven': artifact file does not exist: '\grails-core\grails-profiles\rest-api-plugin\build\classes\profile\META-INF\grails-profile\profile.yml'
```

jdaugherty:
This PR:
1. merges grails-wrapper & the `expanded` wrapper that contains all of the dependencies; since users use `grails-wrapper` I went with the project names `grails-wrapper` & `grails-wrapper-impl`
2. merges the profiles into the project
3. rewrites the profile gradle logic to be cacheable, lazy, & support inheritance via a 'profileRuntimeApi` scope
4. removes the existing functional tests - they need rewritten to be in gradle.  We should open a bug ticket on this.  We can revert 9b8fa17d when we're ready to work on this.
5. removes the groovy jmx library from shell - this was included so profiles had access, but as part of the gradle refactor, you can now include dependencies there.  This may need added back to shell, but I think end projects should include the libraries they need to run the scripts they use instead of being included in the grails-shell-cli.
6. due to https://github.com/gradle/gradle/issues/2986 I did have to change the global dependency exclude, I spoke to @jamesfredley  and we agreed to minimize it to '.gitignore' and '.gitkeep'
7. the shell scripts are checked in under grails-wrapper/shell are still used manually - well need to change the process that stages these scripts for the wrapper in end user projects.  I believe these projects are `grails-profiles/base` and `grails-profiles/profile`.  We should create a ticket for this work so that when base builds, it copies from the associated location.
8. gradlew scripts are checked in under some of the profiles, we should just copy the version grails uses to match what we expect to run with.  We can take a similar approach, where we create a copy task in gradle that runs as part of build. 
9. enhances the extract dependencies task to use the project property 'pomArtifactId' instead of the project name when generating the bom dependencies
10. some tests are now passing that have been previously been marked with PendingFeature and the annotation is removed
11. I have removed the external JS project specific profiles - vue, angular, react.  These profiles are all extensions of the rest-api profile and this matches the forge decision to not include them.  Otherwise, we would be maintaining them every time angular is upgraded.  
12. https://github.com/apache/grails-core/issues/14138 is likely resolved by this since i removed the validation.  for now I left the TODO / commented the validation out

@matrei  would you be willing to do 7 or 8 above in a separate PR after this is merged?


jamesfredley:
On https://github.com/apache/grails-core/pull/14686/files I am still working towards it properly loading the `org.apache.grails.profiles` profiles and correctly generating and app using them.

In working on that I ran into the following.  I believe it exists on `rest-api-plugin` and `web-plugin`

```
Execution failed for task ':grails-profiles-rest-api-plugin:publishMavenPublicationToMavenLocal'.
> Failed to publish publication 'maven' to repository 'mavenLocal'
   > Invalid publication 'maven': artifact file does not exist: '\grails-core\grails-profiles\rest-api-plugin\build\classes\profile\META-INF\grails-profile\profile.yml'
```

jamesfredley (approved):
There is an issue with publishMavenPublicationToMavenLocal on two of the profiles.  I put it in a separate issue for now.

https://github.com/apache/grails-core/issues/14687

https://github.com/apache/grails-core/issues/14688 - issue to fix profile loading in grails-shell-cli

jdaugherty:
> There is an issue with publishMavenPublicationToMavenLocal on two of the profiles. I put it in a separate issue for now.
> 
> #14687
> 
> #14688 - issue to fix profile loading in grails-shell-cli

I fixed the publishing - those plugins do not have compiled code and the compile profile task was being used for 2 purposes.  I probably should split the task, but for now I made it work in either case.

jdaugherty:
I'm going to go ahead and merge this for end-to-end testing and to get the license reports clean.  We can revisit if anyone else has concerns.

PR #14696, opened 2025-05-02
Automate grailsw distribution

jdaugherty:
Previously we manually copied the wrapper file from the wrapper/shell to the associated profiles.  Now we'll have gradle do this for us and not check in those files.


jamesfredley:
@jdaugherty This is awesome.  

But when I try to run them in a generated app, I am seeing the following error.

```
grailsw
Error: Could not find or load main class grails.init.Start
Caused by: java.lang.ClassNotFoundException: grails.init.Start
```

jdaugherty:
> @jdaugherty This is awesome.
> 
> But when I try to run them in a generated app, I am seeing the following error.
> 
> ```
> grailsw
> Error: Could not find or load main class grails.init.Start
> Caused by: java.lang.ClassNotFoundException: grails.init.Start
> ```

What commands are you executing?  I'm able to launch the jar, so I suspect it's related to the bat file and I've missed something that i fixed in shellscript: grailsw

jamesfredley (approved):

PR #14708, opened 2025-05-03
Additional outstanding issues for grails-shell-cli

### LINKED ISSUE #14692: Reduce size of grails-shell-cli distribution
### Issue description

- Remove javadoc and source jars from dist directory, these are not used from this location and published to maven
- Remove media directory (6,610,788 bytes - 26 files) Grails logos and icon files, not used here at all

49,109,290 bytes
264 Files, 131 Folders

vs

31,277,374 bytes
116 Files, 128 Folders

### LINKED ISSUE #14707: grails-shell-cli:  No profile found for name [web] & ArtifactNotFoundException gradle-tooling-api
### Issue description

This only occurs when in a grails application directory.  If in any other directory the profile load correctly.

After generating an application via

./grails create-app Test1

then either starting interactive mode (./grails) or running another command such ./grails list-profiles

```
grails list-profiles --stacktrace
| Error Error occurred running Grails CLI: No profile found for name [web]. (NOTE: Stack trace has been filtered. Use --verbose to see entire trace.)
java.lang.IllegalStateException: No profile found for name [web].
        at org.grails.cli.GrailsCli.initializeProfile(GrailsCli.groovy:525)
        at org.grails.cli.GrailsCli.initializeApplication(GrailsCli.groovy:309)
        at org.grails.cli.GrailsCli.execute(GrailsCli.groovy:272)
        at org.grails.cli.GrailsCli.main(GrailsCli.groovy:162)
| Error Error occurred running Grails CLI: No profile found for name [web].
```

`org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.gradle:gradle-tooling-api:jar:8.12.1 in mavenCentral (https://repo1.maven.org/maven2)`

jamesfredley:
1. Add https://repo.grails.org/grails/core back allows gradle-tooling-api to load from https://repo.grails.org/ui/native/core/org/gradle/gradle-tooling-api/
2. Remove javadoc and source jars from dist directory, these are not used from this location
3. Remove most of media directory (6,610,788 bytes - 25 files)
Size comp
> 49,109,290 bytes
> 264 Files, 131 Folders
> 
> vs
> 
> 31,282,479 bytes
> 117 Files, 129 Folders
4. Fix No profile found for name [web] by Always creating a maven profile repository to lookup profile

### DIFF
--- a/gradle/assemble-config.gradle
+++ b/gradle/assemble-config.gradle
@@ -19,13 +19,14 @@
 
 tasks.register('installToHomeDist').configure { Task task ->
     task.inputs.files(tasks.named('jar'))
-    task.inputs.files(tasks.named('sourcesJar'))
-    task.inputs.files(tasks.named('javadocJar'))
     task.outputs.dir(distInstallDir)
 
     doLast {
         copy {
-            it.from layout.buildDirectory.dir('libs')
+            it.from layout.buildDirectory.dir('libs').map { it.asFileTree.matching {
+                include '**/*.jar'
+                exclude '**/*-sources.jar', '**/*-javadoc.jar'
+            } }
             it.into distInstallDir
         }
     }
--- a/gradle/assemble-config.gradle
+++ b/gradle/assemble-config.gradle
@@ -23,11 +23,9 @@ tasks.register('installToHomeDist').configure { Task task ->
 
     doLast {
         copy {
-            it.from layout.buildDirectory.dir('libs').map { it.asFileTree.matching {
-                include '**/*.jar'
-                exclude '**/*-sources.jar', '**/*-javadoc.jar'
-            } }
+            it.from layout.buildDirectory.dir('libs')
             it.into distInstallDir
+            it.exclude '*sources.jar', '*-javadoc.jar'
         }
     }
 }
--- a/gradle/assemble-root-config.gradle
+++ b/gradle/assemble-root-config.gradle
@@ -116,7 +116,7 @@ tasks.register('assemble', Zip).configure {
 
     into("grails-$grailsVersion") {
         from(rootProject.layout.projectDirectory) {
-            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'LICENSE', 'INSTALL'
+            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'LICENSE', 'INSTALL', 'NOTICE'
 
             def pathsToExclude = [
                     'ant/bin', 'src/grails', 'src/war'
--- a/gradle/assemble-root-config.gradle
+++ b/gradle/assemble-root-config.gradle
@@ -116,8 +116,7 @@ tasks.register('assemble', Zip).configure {
 
     into("grails-$grailsVersion") {
         from(rootProject.layout.projectDirectory) {
-            // Some of these are probably not needed as they are not present in the project folder
-            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'samples/', 'scripts/', 'LICENSE', 'INSTALL'
+            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'LICENSE', 'INSTALL'
 
             def pathsToExclude = [
                     'ant/bin', 'src/grails', 'src/war'
--- a/gradle/assemble-root-config.gradle
+++ b/gradle/assemble-root-config.gradle
@@ -117,7 +117,7 @@ tasks.register('assemble', Zip).configure {
     into("grails-$grailsVersion") {
         from(rootProject.layout.projectDirectory) {
             // Some of these are probably not needed as they are not present in the project folder
-            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/', 'samples/', 'scripts/', 'LICENSE', 'INSTALL'
+            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'samples/', 'scripts/', 'LICENSE', 'INSTALL'
 
             def pathsToExclude = [
                     'ant/bin', 'src/grails', 'src/war'
--- a/grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy
+++ b/grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy
@@ -507,12 +507,7 @@ class GrailsCli {
     private initializeProfile() {
         BuildSettings.TARGET_DIR?.mkdirs()
 
-        if(!new File(BuildSettings.BASE_DIR, "profile.yml").exists()) {
-            populateContextLoader()
-        }
-        else {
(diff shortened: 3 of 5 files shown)


gradle/assemble-config.gradle
```diff
@@ -19,13 +19,14 @@
...
 tasks.register('installToHomeDist').configure { Task task ->
     task.inputs.files(tasks.named('jar'))
-    task.inputs.files(tasks.named('sourcesJar'))
-    task.inputs.files(tasks.named('javadocJar'))
     task.outputs.dir(distInstallDir)
 
     doLast {
         copy {
-            it.from layout.buildDirectory.dir('libs')
+            it.from layout.buildDirectory.dir('libs').map { it.asFileTree.matching {
```
jdaugherty: Was there more than jar files being copied?  Why not add the exclude at the high level so you don't have to iterate over all of the files?

gradle/assemble-root-config.gradle
```diff
@@ -116,8 +116,7 @@ tasks.register('assemble', Zip).configure {
 
     into("grails-$grailsVersion") {
         from(rootProject.layout.projectDirectory) {
-            // Some of these are probably not needed as they are not present in the project folder
-            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/', 'samples/', 'scripts/', 'LICENSE', 'INSTALL'
+            include 'bin/grails', 'bin/grails.bat', 'lib/', 'media/icons/grails.icns', 'LICENSE', 'INSTALL'
```
jdaugherty: Shouldn't the notice be included now? 

Why do the media/icons need included?  Does the grails shell still launch a tray icon?

grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy
```diff
@@ -507,12 +507,7 @@ class GrailsCli {
     private initializeProfile() {
         BuildSettings.TARGET_DIR?.mkdirs()
 
-        if(!new File(BuildSettings.BASE_DIR, "profile.yml").exists()) {
-            populateContextLoader()
-        }
-        else {
-            this.profileRepository = createMavenProfileRepository()
-        }
+        this.profileRepository = createMavenProfileRepository()
```
jdaugherty, line 510: Wasn't this code there in case someone wanted to override the profile?  WOn't this always pull upstream now?

grails-wrapper-impl/src/main/groovy/grails/init/RunCommand.groovy
```diff
@@ -51,11 +51,13 @@ class RunCommand {
 
         GroovyClassLoader groovyClassLoader = new GroovyClassLoader(RunCommand.classLoader)
 
-        List<RepositoryConfiguration> repositoryConfigurations = [new RepositoryConfiguration("mavenCentral", new URI("https://repo1.maven.org/maven2"), true)]
+        List<RepositoryConfiguration> repositoryConfigurations = [new RepositoryConfiguration("mavenCentral", new URI("https://repo1.maven.org/maven2"), false)]
+
+        repositoryConfigurations.add(new RepositoryConfiguration("grailsCentral", new URI("https://repo.grails.org/grails/core"), true))
```
jdaugherty, line 56: I thought multiple repositories weren't working? 

If this is really for the tooling API, can we create a dedicated repo on repo.grails.org for it?

gradle/assemble-root-config.gradle
jamesfredley: The notice file can be added.

https://github.com/apache/grails-core/blob/9794089e793c642790cb2b90f51e66727c726781/gradle/assemble-root-config.gradle#L64 is the only file being used from /media

gradle/assemble-config.gradle
jamesfredley: Just jar files: compiled, source and javadoc.

Can you point me in the direction for doing this at a higher level?  This was the best path I could find while searching for a solution.  Ideally we would just grab the files from tasks.named('jar')

grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy
jamesfredley: I believe that is `~/.grails/settings.groovy`

https://github.com/apache/grails-core/blob/9fef7c0f3269d10998439e1eed2bb564020840fb/grails-shell-cli/src/main/groovy/org/grails/cli/GrailsCli.groovy#L83

This removed code was only running when wrapper or shell-cli and failed to correctly lookup the active profile.  Given how flakey it was and the repeated trouble it created, simplifying this code seemed like the right path forward.

gradle/assemble-root-config.gradle
jdaugherty: Yeah, so it's the icon for the shell that shows up on the task bar.  good to know.

grails-wrapper-impl/src/main/groovy/grails/init/RunCommand.groovy
jamesfredley: I'd like to bring `https://repo.grails.org/grails/core` back for now as any end project that requires dependencies from there will fail without this when running scripts.

Once we get more feedback on `tooling-api` I believe we can add repo.gradle.org/gradle/libs-releases also.

`MavenProfileRepository` was the location where multiple repos failed.  There were two other locations related to wrapper and shell-cli that function with multiple (GrailsDependencyVersions.groovy & RunCommand.groovy).

gradle/assemble-root-config.gradle
jamesfredley: NOTICE has been added.

matrei (approved):

gradle/assemble-config.gradle
jdaugherty: I was suggesting this below code.  Callign map will have gradle process all of the files, while using exclude will make use of the ant exclude so it will exclude as it copies.

        copy {
            it.from layout.buildDirectory.dir('libs')
            it.into distInstallDir
            it.exclude '*sources.jar', '*-javadoc.jar'
        }

jamesfredley: Got it, thank you.

gradle/assemble-config.gradle
```diff
@@ -23,11 +23,9 @@ tasks.register('installToHomeDist').configure { Task task ->
 
     doLast {
         copy {
-            it.from layout.buildDirectory.dir('libs').map { it.asFileTree.matching {
-                include '**/*.jar'
-                exclude '**/*-sources.jar', '**/*-javadoc.jar'
-            } }
+            it.from layout.buildDirectory.dir('libs')
             it.into distInstallDir
+            it.exclude '*sources.jar', '*-javadoc.jar'
```
matrei, line 28: Is there a reason why we have `*` on sources vs `*-` on javadoc?

jamesfredley: Copy & paste and I did not notice it, but this is being reworked substantially in another PR.

jdaugherty: https://github.com/apache/grails-core/pull/14714

Issue #14679, opened 2025-04-29
Reproducible Builds


jdaugherty:
## ASF Requirements
To begin releasing under the ASF, the Grails project must adhere to it's release policies.  The requirements of [Automated Release Signing](https://infra.apache.org/release-signing.html#automated-release-signing) require a [reproducible build](https://reproducible-builds.org/).    

## History of Custom Transformations
Grails makes heavy use of:
1. AST transforms
2. properties files
3. custom xml metadata files 

Historically, these transforms were unordered, and up until Groovy 4 it was not possible to order them.   There were a set of transforms where order did matter.  They were grouped by function - the grails-core framework transforms & the grails-data-mapping transforms.  To work around the Groovy limitation, Grails introduced both a `GlobalGrailsClassInjectorTransformation` (for grails-core) and `OrderedGormTransformation` (for grails-data-mapping).  These AST transformations would then call other transformers in their defined order.  

The problem with the Grails implementation is that as each annotated node was processed by the Groovy transform, it would call all of the Grails transformers.  This mean that the order of processing would often be:  Class Node (apply all transformers), Method Node (apply all transformers), etc.  While the Groovy 4 `TransformWithPriority` interface processes all nodes under a single transform, and then moves to the next highest priority transformer. 

So far, we have worked towards reproducible builds by moving as many transforms out of the custom grails & data-mapping transformers.  We've adopted the Groovy TransformWithPriority interface where possible.  We've switched our method iterations & various caches to use collections with determined sort orders.  We also have changed our xml files & properties files to write consistent dates based on the presence of a `SOURCE_DATE_EPOCH` environment variable that is set in CI based on the last commit date.

## GroovyTransformOrder
Because the order of AST Transforms was not guaranteed in Groovy 3, we determined the Grails 7 order by extensive debugging.  Once we added various combinations on a class and listed the resulting transforms, we created an order registry in the class [GroovyTransformOrder](https://github.com/apache/grails-core/blob/7.0.x/grails-common/src/main/groovy/org/apache/grails/common/compiler/GroovyTransformOrder.groovy).  This class lists the transforms in order from highest priority (first run) to lowest priority (last run).

## Future Deprecations
Debugging some of these transforms revealed that some are no longer heavy used and have limited test coverage.  In some cases, use was even discouraged (i.e. the mixin transform).  This ticket deprecates the Mixin transform since AST transforms should be adopted instead.  We will remove in a future Grails Version.

## Affect on Grails Application
[...]

PR #14710, opened 2025-05-03
reproducible builds - groovydoc 4, profile fixes, and ensuring gradle projects build docs

jdaugherty:
https://github.com/apache/grails-core/issues/14679 

* uses groovydoc 4
* fixes the profiles to be reproducible
* ensures source & javadoc jars are built as part of the test script
* ensures the gradle projects uses groovydoc instead of javadoc to generate documentation


grails-bootstrap/build.gradle
```diff
@@ -38,12 +38,10 @@ dependencies {
         exclude group: 'org.codehaus.groovy'
     }
 
-    // we have to continue to compile with groovy 3 only because of grails-shell
-    // the grails-shell will run under gradle and thus this must remain compatible
-    compileOnly "org.codehaus.groovy:groovy:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-templates:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-xml:$GroovySystem.version"
-    compileOnly "org.codehaus.groovy:groovy-ant:$GroovySystem.version"
+    compileOnly 'org.apache.groovy:groovy'
```
jdaugherty, line 41: I spoke to @jamesfredley about this change.  Since bootstrap is included by both gorm, and core, we should be compiling with 4.  We'll know if it breaks b/c shell still uses 3.

jamesfredley (approved):

jdaugherty:
I'm going to hold off on merging this for a bit.  Paul is continuing to investigate the groovy ordering issues and I want to use this PR to easily test the snapshot changes.

jdaugherty:
I have removed the bom changes so that only the build uses the snapshot build for documentation generation.

PR #14715, opened 2025-05-06
Rat adjustments based on grails-forge PR feedback

jamesfredley:
- Remove permission that is defaulted
- Make rat task lazily configured

Based on feedback from https://github.com/apache/grails-forge/pull/562, which started with copies of the same files from grails-core.

### DIFF
--- a/.github/workflows/rat.yml
+++ b/.github/workflows/rat.yml
@@ -28,8 +28,6 @@ on:
   workflow_dispatch:
 jobs:
   rat-audit:
-    permissions:
-      contents: read  #  to fetch code (actions/checkout)
     runs-on: ubuntu-latest
     steps:
       - name: "📥 Checkout repository"


jdaugherty (approved):

gradle/rat-root-config.gradle
```diff
@@ -17,8 +17,8 @@
 
 apply plugin: 'org.nosphere.apache.rat'
 
-tasks.named('rat') {
-    excludes = [
+tasks.named('rat').configure {
```
matrei: Please revert this again. According to my tests today it should not be needed. Sorry about that.

matrei (approved):

gradle/rat-root-config.gradle
jamesfredley: I removed the commit with `.configure {}` before merging this PR.

PR #14720, opened 2025-05-08
Buildscript Repository config consolidation for Tests

jamesfredley:
Repository config was already consolidated in the root `build.gradle` file.  This extends the consolidation for `buildscript{ repositories {}}`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,16 @@ allprojects {
         maven { url = 'https://repository.apache.org/content/groups/snapshots' }
         // mavenLocal() // Keep, this will be uncommented and used by CI (groovy-joint-workflow)
     }
+
+    buildscript {
+        repositories {
+            mavenCentral()
+            maven { url = 'https://repo.grails.org/grails/core' }
+            maven { url = 'https://oss.sonatype.org/content/repositories/snapshots' }
+            maven { url = 'https://repository.apache.org/content/groups/snapshots' }
+            // mavenLocal() // Keep, this will be uncommented and used by CI (groovy-joint-workflow)
+        }
+    }
 }
 
 subprojects {
--- a/grails-test-examples/app1/build.gradle
+++ b/grails-test-examples/app1/build.gradle
@@ -18,12 +18,6 @@
  */
 
 buildscript {
-    repositories {
-        // mavenLocal()
-        maven { url = 'https://repo.grails.org/grails/core' }
-        mavenCentral()
-        maven { url = 'https://repository.apache.org/content/groups/snapshots' }
-    }
     dependencies {
         classpath platform("org.apache.grails:grails-gradle-bom:${project.projectVersion}")
         classpath 'org.apache.grails:grails-gradle-plugins'
--- a/grails-test-examples/app2/build.gradle
+++ b/grails-test-examples/app2/build.gradle
@@ -18,11 +18,6 @@
  */
 
 buildscript {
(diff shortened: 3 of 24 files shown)


build.gradle
```diff
@@ -52,6 +52,16 @@ allprojects {
         maven { url = 'https://repository.apache.org/content/groups/snapshots' }
         // mavenLocal() // Keep, this will be uncommented and used by CI (groovy-joint-workflow)
     }
+
+    buildscript {
+        repositories {
+            mavenCentral()
+            maven { url = 'https://repo.grails.org/grails/core' }
```
jdaugherty, line 59: I do not agree with pulling into the entire repo grails org.  we need to know where our core dependencies come from and I intentionally had added the expanded scope on test apps, while leaving the core code isolated.  The intent longer term is to adopt isolated builds for tests and thus we need to maintain what others keep in their build files

jdaugherty (changes requested):

build.gradle
```diff
@@ -52,6 +52,16 @@ allprojects {
         maven { url = 'https://repository.apache.org/content/groups/snapshots' }
```
matrei, line 52: It think the Apache Snapshot repo will have to be defined below mavenLocal() if the locally built Groovy Snapshot is to be used in the Groovy Joint Workflow.

build.gradle
```diff
@@ -52,6 +52,16 @@ allprojects {
...
         // mavenLocal() // Keep, this will be uncommented and used by CI (groovy-joint-workflow)
     }
+
+    buildscript {
+        repositories {
+            mavenCentral()
+            maven { url = 'https://repo.grails.org/grails/core' }
+            maven { url = 'https://oss.sonatype.org/content/repositories/snapshots' }
+            maven { url = 'https://repository.apache.org/content/groups/snapshots' }
```
matrei, line 61: It think the Apache Snapshot repo will have to be defined below mavenLocal() if the locally built Groovy Snapshot is to be used in the Groovy Joint Workflow.

matrei (approved):

jamesfredley:
The reason I submitted this PR was to consolidate the buildscript{} repository configuration into a smaller number of locations.  Given we were already setting repositories{} under allprojects {}, I added it there.   There are a few conversations about adding repositories (gradle libs and spring milestones) and that currently would require changes in a large number of locations. 

I am wide open to thinking about other solutions.  The gradle plugin e2e tests also all currently include a repository list.

build.gradle
jamesfredley: agreed

build.gradle
jamesfredley: agreed

jdaugherty:
> The reason I submitted this PR was to consolidate the buildscript{} repository configuration into a smaller number of locations.  Given we were already setting repositories{} under allprojects {}, I added it there.   There are a few conversations about adding repositories (gradle libs and spring milestones) and that currently would require changes in a large number of locations. 
> 
> 
> 
> I am wide open to thinking about other solutions.  The gradle plugin e2e tests also all currently include a repository list.

The e2e tests should not be consolidated since we are testing our gradle plugin independently of Grails.  The only dependency is groovy since we need to compile source files.  These files are constructed to test specific scenarios too.

jamesfredley:
Replaced by https://github.com/apache/grails-core/pull/14743

PR #14726, opened 2025-05-09
Functional test repository config consolidation with Gradle Plugin

matrei:
This PR builds on #14720 and changes to use a Gradle plugin to selectively appy the buildscript repositories to only the functional test projects.


matrei:
This won't work as the plugin is applied to late.

PR #14727, opened 2025-05-09
Ensure we only remove update-wrapper argument.

markeganfuller:
Previously any arg would be removed assuming it was update-wrapper, despite the fact we can get here without an update-wrapper argument.

For us, this was causing our builds to use the wrong environment as:
```
-Dgrails.env=staging war
```
was being turned into
```
war
```


jamesfredley:
@markeganfuller Thank you, I have reviewed and merged.

PR #14732, opened 2025-05-13
Run workflow with Java 24 instead of 23, in addition to LTS versions …

jamesfredley:
…17 and 21

24 was released on March 18, 2025

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -60,7 +60,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 17, 21, 23 ]
+        java: [ 17, 21, 24 ]
         os: [ ubuntu-latest, windows-latest, macos-latest ]
     runs-on: ${{ matrix.os }}
     steps:
@@ -89,7 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 17, 21, 23 ]
+        java: [ 17, 21, 24 ]
     runs-on: ubuntu-24.04
     steps:
       - name: "📥 Checkout repository"


jdaugherty:
I don't think we can do this until Groovy supports Java 24.

matrei:
> I don't think we can do this until Groovy supports Java 24.

Does Groovy 4 even officially support Java 17?

> Groovy 4.0 requires Java 8+ with support for up to Java 16.
https://groovy-lang.org/install.html

matrei (approved):
The tests pass so let's do this!

jamesfredley:
Groovy is testing master with 24 and 25:
https://github.com/apache/groovy/blob/master/.github/workflows/groovy-build-test.yml#L58
https://github.com/apache/groovy/blob/master/.github/workflows/groovy-build-test.yml#L80

But are not currently testing 24 on GROOVY_4_0_X: https://github.com/apache/groovy/blob/GROOVY_4_0_X/.github/workflows/groovy-build-test.yml#L56

I am following up to confirm Java 24 support and then will update this PR.

jamesfredley:
Spring Boot is building with Java 24 against 3.5.x and 3.4.x:

3.5.x: https://github.com/spring-projects/spring-boot/blob/main/.github/workflows/ci.yml
3.4.x: https://github.com/spring-projects/spring-boot/blob/3.4.x/.github/workflows/ci.yml#L28

> Spring Boot 3.4.5 requires at least [Java 17](https://www.java.com/) and is compatible with versions up to and including Java 24

https://docs.spring.io/spring-boot/system-requirements.html

> Spring Boot 3.5.0-RC1 requires at least [Java 17](https://www.java.com/) and is compatible with versions up to and including Java 24.

https://docs.spring.io/spring-boot/3.5/system-requirements.html

And Spring Framework 6.2.x will support JDK 17-25 (expected). https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions#jdk-version-range

jdaugherty (approved):

PR #14745, opened 2025-05-16
#14729 - Corrupt Jar & War Files

### LINKED ISSUE #14729: Invalid or corrupt jarfile - bootjar & bootwar
### Issue description

./gradlew bootRun works fine.

The generated jar and war files are invalid or corrupt with Spring Boot 3.5.0-RC1 and 3.4.5

./gradlew bootJar
./gradlew bootWar

java -jar build/libs/demo-0.1.war

Error: Invalid or corrupt jarfile demo-0.1.war

java -jar build/libs/demo-0.1.jar

Error: Invalid or corrupt jarfile demo-0.1.jar

3.5.0-RC1 Example:  https://github.com/jamesfredley/invalidjarwar
3.4.5 Example: https://github.com/jamesfredley/invalidjarwar/tree/boot3.4.5

jdaugherty:
Fixes https://github.com/apache/grails-core/issues/14729

This is a rather large change.  There are multiple changes in this PR, including: 

1. by guaranteeing order of the AST Transformations, a side effect was that grails-plugin.xml & grails.factories are now properly generating for all projects in grails core.  This meant any existing copy of those files that was checked into resources had to be removed.  These duplicate files were 1 cause of the duplicated jar file

2. https://github.com/spring-projects/spring-framework/issues/34796 affected nested jar files and versions of spring boot after 3.4.3 all include the affected version of spring.  There doesn't appear to be a "spring-dependencies" equivalent of spring framework like there is for spring boot.  This means to override the spring version, every spring library has to be added to the bom and overridden.  Worse, Maven / Gradle do not yet support exclusions so this is actually a difficult problem that significantly complicates our bom.  For this reason, I've rolled us back to 3.4.3 until 3.5.0 comes out with the 6.2.7 spring framework fix.

3. In Gradle 8, tasks now run in process isolation via the worker api.  This means that the version of groovy *can* be different than Gradle.  The problem is the grails-gradle-model doesn't export the version of groovy since it assumes there isn't isolation.  So now it exports the version of gradle that it was built with since this library is used by both tasks and plugins

4. reworks the FindMainTask to properly set the main class and fail if a bootJar / bootRun  are enabled, but no Application class exists.  I discussed this behavior with @jamesfredley, and I think end grails app developers need to understand that those tasks will generate invalid jars so it's important to fail.  `grails-geb` is a good example of this, adding bootJar.enabled = false / bootRun.enabled = false will prevent the failure.  We could alternatively just log a warning, but the type of failures caused by invalid jars are incredibly hard to diagnosis.  Forcing a failure is a time saver.  This change will likely affect plugins the most.  We could make a further enhancement to detect if it's a plugin, and not hard fail in that case. @jamesfredley  should we ticket this optional enhancement? 

5. removes the 'doc' method on the grails plugin since it's been deprecated since Grails 1 from when Grails used to be both a Build System & an app.  This method previously generated documentation.

6. Cleans up the Upgrade7.MD for any entry that's been added to the grails-doc.  

7. Removes the duplicated codec plugin that is deprecated.  Updates the docs to reflect this change similar to the changes in gsp where we had other files in this scenario.

8. removes the remaining pathingJar code since gradle properly handles this with bootRun and our start scripts handle it otherwise.  People can use a shadowJar if they need to have their end application perform this function.

9. Fixes the grails-gsp plugin so that it properly nests files in the jar again.

10. In `DataBindingUtils`, I found some code that was calling class.getClass() and believe this is a bug in the domain databinding.  Tests pass after this change, and the original change that changed this from class to class.getClass() claimed it was done for tests. 

11. As part of reviewing the gradle output to debug the invalid docs, I've cleaned up several warnings that I thought were straight forward changes via intellij refactoring.

12. This change revealed there is a lack of test coverage for jar & war file starts.  We know that from the profiles a GroovyMXBean can be used to stop an application.  I think we should add tests that actually start each test app via jar / war file.  There aren't changes to address this in this PR, but we should likely ticket this.

13. There's a rather major problem with the functional tests: they pull the bom from the upstream repo instead of locally.  This is really only a problem when changing a dependency, which we don't often do.  This is due to https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/164 since we can't use the bom in the same gradle project.  As a "workaround" I've added a local repo that only contains the grails-bom to make testing the functional tests with version changes.  Running the task './gradlew :grails-bom:publishMavenPublicationToLocalBomRepository` will publish to that local repo, which typically will be empty, and then the functional tests will pull the dependencies.  

14. removes SDKman workflow, which was used prior to Grails 6.  The release to SDKman occurs in grails-forge:  https://github.com/apache/grails-forge/blob/7.0.x/.github/workflows/release.yml#L91

Since we do publish a valid gradle module for the bom, we could switch to using platforms exclusively in the functional tests, but I think that will defeat the purpose of these tests - end users are still using the spring dependency management plugin since gradle platforms do not support properties.  So until gradle upstream implements a property override for the platform versions, we need to continue to test with the dependency management plugin.  I think the workaround proposed here needs documented in a ticket and we need to track this longer term on how to fix this.


jdaugherty:
Hold off on this review, groovydoc isn't generating docs for this change correctly.

jdaugherty:
I reverted the java 17 cleanup - groovydoc still has an older language level set to it.  I've reached out to both the Gradle Community & Paul to see if we can get this fixed longer term.

jdaugherty:
I also do not error on find main task if it's a plugin (so we do not need the enhancement from bullet 4)

jdaugherty:
(this is ready to review, we merged the spring boot change separately so the tests can run with the correct version once it's published)

jamesfredley:
Issue for point 4:  https://github.com/apache/grails-core/issues/14748

grails-bom/build.gradle
```diff
@@ -161,4 +159,22 @@ ext {
...
+    publishing {
+        repositories {
+            maven {
+                name = 'localBom'
+                url = layout.buildDirectory.dir('repo')
+            }
+        }
+    }
+
+//    project.afterEvaluate {
```
jamesfredley: Can we remove this commented afterEvaluate {} block?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -43,14 +50,29 @@ class FindMainClassTask extends DefaultTask {
...
-            String mainClass = findMainClass()
-            if (mainClass != null) {
-                bootRun.mainClass.set(mainClass)
-                ExtraPropertiesExtension extraProperties = (ExtraPropertiesExtension) project.getExtensions().getByName("ext")
-                extraProperties.set("mainClassName", mainClass)
-            }
+
+        def bootRunTask = project.tasks.findByName('bootRun')
+        if (!bootRunTask || !bootRunTask.enabled) {
+            project.logger.info('The bootRun task will not so this must not be a runnable grails application. Skipping finding main class.')
```
jamesfredley: The bootRun task does not exist or is disabled, so this must not be a runnable grails application. Skipping finding main class.

jamesfredley (approved):

jamesfredley (approved):

grails-gradle/plugins/build.gradle
```diff
@@ -32,7 +30,10 @@ dependencies {
     // compile grails-gradle-plugin with the Groovy version provided by Gradle
     // to ensure build compatibility with Gradle, currently Groovy 3.0.x
     // see: https://docs.gradle.org/current/userguide/compatibility.html#groovy
-    compileOnly 'org.codehaus.groovy:groovy'
+    // TODO: tasks are isolated as of Gradle 7 so we must expose the version of the groovy.
```
matrei, line 35: - Why is this a TODO?
- Should this be `Gradle 8`? (the PR description says Gradle 8)
- Is "must" the correct word here? How has it been working up until now with `compileOnly`?

grails-gradle/plugins/build.gradle
```diff
@@ -32,7 +30,10 @@ dependencies {
...
+    // We could upgrade to groovy 4 for the tasks, but the plugin is not isolated and still
```
matrei: You are writing: `the plugin is`. Aren't there multiple plugins?

grails-gradle/plugins/build.gradle
```diff
@@ -32,7 +30,10 @@ dependencies {
...
+    // needs groovy 3.
+    implementation "org.codehaus.groovy:groovy"
```
matrei, line 38: I don't understand what we gain by exposing Groovy here. Can you elaborate? Which problem do we solve?
Single quotes?

UPGRADE7.md
```diff
@@ -17,32 +17,12 @@ limitations under the License.
...
-- the following deprecated classes were removed, please use the suggested replacement
-  * grails.core.GrailsTagLibClass -> grails.core.gsp.GrailsTagLibClass
-  * org.grails.core.artefact.TagLibArtefactHandler -> org.grails.core.artefact.gsp.TagLibArtefactHandler
-  * org.grails.core.DefaultGrailsTagLibClass -> org.grails.core.gsp.DefaultGrailsTagLibClass 
-- h2 2.x is stricter about reserved words
-  - https://github.com/apache/gorm-hibernate5/pull/910/commits/c8de45df204966ccc228b46b94beeb2142ae0f59 
-- [GROOVY-10621](https://issues.apache.org/jira/browse/GROOVY-10621) 
-  - Primitive booleans will no longer generate the form of isProperty & getProperty.  They will only generate isProperty()
+
+- NOTE: Several items have been directly integrated into the upgrade guide. Please refer there for the full list.
```
matrei: Suggestion: "Please refer to it for the complete list."

build.gradle
```diff
@@ -46,6 +46,12 @@ ext {
 
 allprojects {
     repositories {
+        // workaround for https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/164
+        maven {
```
matrei, line 50: I'm starting to think we should ditch the Dependency Management Plugin, and only use Gradle Platforms.
We are actually adding the `grails-bom` as a `platform` in generated apps.

grails-bom/build.gradle
```diff
@@ -161,4 +159,16 @@ ext {
 
 apply {
     from rootProject.layout.projectDirectory.file('gradle/publish-config.gradle')
+}
+
+// Workaround for https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/164
```
matrei, line 166: I'm starting to think we should ditch the Dependency Management Plugin, and only use Gradle Platforms. We are actually adding the grails-bom as a platform in generated apps.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -43,14 +50,29 @@ class FindMainClassTask extends DefaultTask {
...
+        def bootRunTask = project.tasks.findByName('bootRun')
+        if (!bootRunTask || !bootRunTask.enabled) {
+            project.logger.info('The bootRun task does not exist or is disabled, so this must not be a runnable grails application. Skipping finding main class.')
+            return
+        }
+
+        def bootJarTask = project.tasks.findByName(SpringBootPlugin.BOOT_JAR_TASK_NAME)
+        def bootWarTask = project.tasks.findByName(SpringBootPlugin.BOOT_WAR_TASK_NAME)
+        if ((!bootJarTask || !bootJarTask.enabled) && (!bootWarTask || !bootWarTask.enabled)) {
+            project.logger.info("There is neither a ${SpringBootPlugin.BOOT_JAR_TASK_NAME} or ${SpringBootPlugin.BOOT_WAR_TASK_NAME} task that will run. Skipping finding main Application class.")
```
matrei: Use lazy log message evaluation?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -43,14 +50,29 @@ class FindMainClassTask extends DefaultTask {
...
+            return
+        }
+
+        String mainClass = findMainClass()
+        if (mainClass) {
+            ExtraPropertiesExtension extraProperties = (ExtraPropertiesExtension) project.getExtensions().getByName('ext')
```
matrei: `def extraProperties = (ExtraPropertiesExtension) project.extensions.getByName('ext')`

Suggestions (readability):
- The type is inferred.
- Use Groovy getter syntax.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -43,14 +50,29 @@ class FindMainClassTask extends DefaultTask {
...
+            extraProperties.set('mainClassName', mainClass)
+
+            SpringBootExtension springBootExtension = project.getExtensions().getByType(SpringBootExtension)
```
matrei: `def springBootExtension = project.extensions.getByType(SpringBootExtension)`

Suggestions (readability):
- The type is inferred.
- Use Groovy getter syntax.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -103,15 +125,21 @@ class FindMainClassTask extends DefaultTask {
                 if (mainClass != null) {
                     mainClassFile.text = mainClass
                 } else {
-                    throw new RuntimeException("Could not find Application main class. Please set 'springBoot.mainClass'.")
+                    if (project.plugins.hasPlugin(GrailsPluginGradlePlugin)) {
+                        // this is ok if the project is a plugin because it's likely not going to be a runnable grails app
+                        project.logger.lifecycle("WARNING: this plugin project does not have an Application.class and thus the bootJar / bootRun will be invalid.")
```
matrei: Single quotes?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/run/FindMainClassTask.groovy
```diff
@@ -103,15 +125,21 @@ class FindMainClassTask extends DefaultTask {
...
+                    throw new RuntimeException("Could not find Application main class. Please set 'springBoot.mainClass' or disable BootJar & BootArchive tasks.")
                 }
             }
             return mainClass
         }
     }
 
     protected FileCollection resolveClassesDirs(SourceSetOutput output, Project project) {
-        output?.classesDirs ?: project.files(new File(project.buildDir, "classes/main"))
+        output?.classesDirs ?: project.files(new File(project.buildDir, 'classes/main'))
```
matrei: Use `project.layout` instead of deprecated `project.buildDir`?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
```diff
@@ -115,8 +89,9 @@ class GroovyPagePlugin implements Plugin<Project> {
             war.dependsOn compileGroovyPages
             war.duplicatesStrategy = DuplicatesStrategy.EXCLUDE
             if (war.name == 'bootWar') {
-                war.from(destDir)
-                war.into("WEB-INF/classes")
+                war.from(destDir) { CopySpec it ->
+                    it.into("WEB-INF/classes")
```
matrei: Single quotes?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
```diff
@@ -133,21 +108,16 @@ class GroovyPagePlugin implements Plugin<Project> {
             jar.duplicatesStrategy = DuplicatesStrategy.EXCLUDE
             if (!(jar instanceof War)) {
                 if (jar.name == 'bootJar') {
-                    jar.from(destDir)
-                    jar.into("BOOT-INF/classes")
+                    jar.from(destDir) { CopySpec it ->
+                        it.into("BOOT-INF/classes")
```
matrei: Single quotes?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -533,13 +542,22 @@ class GrailsGradlePlugin extends GroovyPlugin {
     protected void configureConsoleTask(Project project) {
         TaskContainer tasks = project.tasks
         if (!project.configurations.names.contains('console')) {
+            if (!tasks.names.contains('findMainClass')) {
+                project.logger.info("Project ${project.name} does not contain the findMainClass task so the console & shell tasks will not be created.")
```
matrei: Use lazy log message evalutation?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
-        def findMainClassTask = taskContainer.findByName("findMainClass")
-        if (findMainClassTask == null) {
-            findMainClassTask = project.tasks.register("findMainClass", FindMainClassTask).get()
-            findMainClassTask.mustRunAfter(project.tasks.withType(GroovyCompile))
-        } else if (!FindMainClassTask.class.isAssignableFrom(findMainClassTask.class)) {
-            def grailsFindMainClass = project.tasks.register("grailsFindMainClass", FindMainClassTask).get()
-            grailsFindMainClass.dependsOn(findMainClassTask)
-            findMainClassTask.finalizedBy(grailsFindMainClass)
+
+        def existingTask = taskContainer.findByName("findMainClass")
```
matrei: Single quotes?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+        if (existingTask == null) {
+            TaskProvider<FindMainClassTask> findMainClassTask = project.tasks.register("findMainClass", FindMainClassTask)
```
matrei: - Use `def`? (readability: The type is inferred)
- Single quotes?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+
+            Provider<RegularFile> mainClassFileContainer = project.layout.buildDirectory.file('resolvedMainClassName')
```
matrei: Use `def`? (readability: The type is inferred)

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+            Provider<String> mainClassProvider = project.providers.fileContents(mainClassFileContainer).asText.map {
```
matrei: Use def? (readability: The type is inferred)

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                it.trim()
+            }
+            project.tasks.withType(BootArchive).configureEach { BootArchive bootTask ->
```
matrei, line 629: Remove closure parameter and use `it` to reference the task?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                bootTask.dependsOn(findMainClassTask)
+                bootTask.inputs.file(mainClassFileContainer)
+                bootTask.mainClass.convention(mainClassProvider)
+            }
+            project.tasks.withType(BootRun).configureEach { BootRun it ->
```
matrei, line 634: Remove redundant closure parameter?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                it.dependsOn(findMainClassTask)
+                it.inputs.file(mainClassFileContainer)
+                it.mainClass.convention(mainClassProvider)
+            }
+
+            findMainClassTask.configure {
+                it.mustRunAfter(project.tasks.withType(GroovyCompile))
+            }
+            project.tasks.withType(ResolveMainClassName).configureEach {
+                it.dependsOn findMainClassTask
```
matrei: Use parentheses around method parameter?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -597,14 +615,34 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+            }
+        } else if (!FindMainClassTask.class.isAssignableFrom(existingTask.class)) {
+            project.logger.warn("Grails Projects typically register a findMainClass task to force the MainClass resolution for Spring Boot. This task already exists so this will not occur.")
```
matrei: Single quotes?

matrei (approved):
Great job James!
What's the story about the updated license headers?

grails-gradle/plugins/build.gradle
jdaugherty: The FindMainTask was not running until this PR.  Once I fixed it to run, it errors indicating groovy was not found.  The reason it errors is Gradle made tasks have process isolation so they can have their own library versions and not depend on gradle's classpath.  

The problem is this library is used by both plugins and tasks.  For plugins, there is no such isolation.  My quick fix was to expose groovy.  The todo is to either split this library or find another solution.  We could also just leave this as implementation to handle both cases.

grails-gradle/plugins/build.gradle
jdaugherty: The task doesn't run due process isolation (the finder will fail saying groovy is not found), so we have to expose it.

grails-gradle/plugins/build.gradle
jdaugherty: We technically have one plugin that everything inherits, but yes, I will update this to 'plugins'

build.gradle
jdaugherty: I 100% agree but we need the ability for end users to easily set versions.  I am debating whats better: ditch the dependency plugin inside of grails core or keep it to align with what end user apps use.  If gradle platforms supported property overrides, I would have suggested we already switch. I came close to writing a Gradle plugin to just add this support too.

jdaugherty:
> Great job James!
> 
> What's the story about the updated license headers?

IntelliJ was updating these.  Looks like I need to adjust its header.  I will revert these changes.

grails-gradle/plugins/build.gradle
matrei: Ah, thank you, that clears it up for me!

build.gradle
matrei: I think property version overrides are a really good feature but when needed it can be accomplished with Gradle Platforms as well, only not as easily, as you have to declare the dependency you want to be able to override beforehand.
We should rip that band-aid, it causes confusion to do both.

jdaugherty: I still maintain we shouldn't use both in end user apps - we should one or the other.  The property overrides make it very easy to set versions.  We should discuss this more outside of this PR.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jdaugherty: Can we update these all at once instead of piecemeal?  Have we run 2 runs and checked that the caching is the same?

jdaugherty:
1. Per separate discussion, I've added [skip tests] option to skip the tests to make publishing the bom easier.
2. After publishing the bom, I realized that the ErrorsControllerSpec is broken.  Retesting, it seems to be broken in all of 3. 4.x in spring boot.  The page being rendered in ErrorsControllerSpec appears to be the tomcat default 500 error page.  I've switched this back to 3.5.0-RC1 & setup constraints for spring framework 6.2.7.  Now both these tests + the nested jar issue is fixed.
4. Per separate discussion, the .configure {} & .configureEach changes are being tested here: https://github.com/matrei/grails-core/tree/simplify-build I'll try to follow-up this week and see if I can make this change everywhere instead of piecemeal.

PR #14751, opened 2025-05-16
Fix grails-shell-cli grailsVersion lookup and grails-bom resolution

jamesfredley:
`GrailsDependencyVersions.package.implementationVersion` is null

### DIFF
--- a/grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
+++ b/grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
@@ -52,7 +52,7 @@ class GrailsDependencyVersions implements DependencyManagement {
     }
 
     GrailsDependencyVersions(GrapeEngine grape) {
-        this(grape, [group: "org.apache.grails", module: "grails-bom", version: Environment.getGrailsVersion(), type: "pom"])
+        this(grape, [group: "org.apache.grails", module: "grails-bom", version: Environment.grailsVersion, type: "pom"])
     }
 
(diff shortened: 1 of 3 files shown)


jdaugherty:
It looks like the jar file isn't having an attribute added to it in forge.  We should fix it there instead of changing this.

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
```diff
@@ -52,7 +52,7 @@ class GrailsDependencyVersions implements DependencyManagement {
     }
 
     GrailsDependencyVersions(GrapeEngine grape) {
-        this(grape, [group: "org.apache.grails", module: "grails-bom", version: GrailsDependencyVersions.package.implementationVersion, type: "pom"])
+        this(grape, [group: "org.apache.grails", module: "grails-bom", version: Environment.grailsVersion, type: "pom"])
```
jdaugherty, line 55: I'm hesitant to change this.  I suspect it's the class loader that's causing this to be null.  

Does this work: 

         getClass().getClassLoader().getPackage('org.grails.cli.boot').implementationVersion

jdaugherty: Looking at this more, I think we need to be loading the shells reflectively so that the static initializers are called at runtime.  I accidentally pushed to 7.0.x for forge. can you retest with just that change and see if either of these PRs are still needed?

jdaugherty:
These changes were to fix a test.  The reason the package can't find the implementation version is because that's only set on the jar file.

PR #14753, opened 2025-05-17
#11795 preserve order of bean-defined basenames

### LINKED ISSUE #11795: resourcebundle basenames order not respected
### Steps to Reproduce

Define the PluginAwareResourceBundleMessageSource in resources.groovy and set the basenames array.

### Expected Behaviour
Grails respects the order of basenames defined, you one is able to overwrite existing keys with new ones.
This is especially useful if you migrate a project with thousands of translations in many languages and need to re-define key texts. For this, you use a basename which is defined first in the list. 
Of course you can play around with bundle naming, as this is the default behaviour of the PluginAwareResourceBundleMessageSource, but the AbstractMessageSource.setBasenames() doc says:
"The associated resource bundles will be checked sequentially when resolving a message code".
Also, Spring ReloadableResourceBundle uses a similar functionality.

### Actual Behaviour

Grails ignores the basenames order, so you cannot overwrite bundle keys in bundles you defined to be resolved first.

### Environment Information

- **Operating System**: Doesn't matter. macOS 15 in this case
- **Grails Version:** Grails 3 and 4 affected
- **JDK Version:** 8 or 11
- **Container Version (If Applicable):** Doesn't matter.

### Example Application

https://github.com/robertoschwald/GrailsPluginAwareBasenamesOrderBug

jdaugherty:
Rebasing https://github.com/apache/grails-core/issues/11795 onto 7.0.x

@robertoschwald  I see you submitted this PR sometime ago and it was never merged.  From what I can tell, it's still an issue.  I don't see any tests in the original PR, are you willing to add tests for this?  If so, I'm happy to merge this into 7.0.0.

The problem with bundle ordering materalized here when adding a PluginAwareResourceBundleMessageSource messageSource definition and sets own basenames to be able to overwrite other bundles keys in a centralized bundle file. Then the plugin basenames were first in the list, not last.


robertoschwald:
See #14755

matrei:
I would expect to see a failing test in `ResourceBundleMessageSourceSpec` after resetting the changes to `PluginAwareResourceBundleMessageSource`.

borinquenkid:
@jdaugherty I fixed it for him just take a quick look

PR #14755, opened 2025-05-19
Test-Case for #14753

robertoschwald:
Test-Case for #14753

### DIFF
--- a/grails-core/src/test/groovy/org/grails/spring/context/ResourceBundleMessageSourceSpec.groovy
+++ b/grails-core/src/test/groovy/org/grails/spring/context/ResourceBundleMessageSourceSpec.groovy
@@ -30,10 +30,12 @@ class ResourceBundleMessageSourceSpec extends Specification {
     void setup(){
         messages = new TestResource('messages.properties','''\
             foo=bar
+            shared.message=Messages Message
         '''.stripIndent().getBytes('UTF-8'))
          
         other = new TestResource('other.properties','''\
             bar=foo
+            shared.message=Other Message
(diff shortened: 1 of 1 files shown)


matrei:
Is this test supposed to fail before #14753 is merged?

robertoschwald:
This PRs base branch is [apache:issue-11795](https://github.com/apache/grails-core/tree/issue-11795), the base for PR #14753 
@jdaugherty rebased my original PR into a new one I cannot push to.

matrei:
OK, but when I run the test added in this PR, with branch `7.0.x` checked out, it succeeds.
I would expect it to fail before #14753 is merged.
It makes me wonder if the test validates what it's suppose to validate.
It's likely I'm missing something so I just want to hear if this is expected behavior.

robertoschwald:
The problem with bundle ordering materalized here when adding a PluginAwareResourceBundleMessageSource  messageSource definition and sets own basenames to be able to overwrite other bundles keys in a centralized bundle file. Then the plugin basenames were first in the list, not last. 
So yes, the test should work against 7.0.0, but also with #14753

jdaugherty:
Can we add a test for the problem that this fixes? to show it fails in 7 prior to the fix?

jdaugherty:
@jamesfredley I don't think this should have been merged.  We needed a test case that showed this was broken prior to the fixes

jamesfredley:
@jdaugherty Agreed.  I merged to two newest PRs together as a starting point, since it was disjointed and hard to follow across the set of PRs.

PR #14759, opened 2025-05-20
Use groovy 4 for grails-gradle-tasks

jamesfredley:
Groovy 3 was conflicting with Groovy 4 version used by some other Grails dependencies in the build, such as `org.apache.grails:grails-data-hibernate5`

Resolves:

```
A problem occurred configuring root project 'Test8'.
> Could not resolve all artifacts for configuration 'classpath'.
   > Could not resolve org.apache.groovy:groovy:4.0.26.
     Required by:
         root project : > org.apache.grails:grails-bom:7.0.0-SNAPSHOT:20250520.032711-84 > org.apache.groovy:groovy-bom:4.0.26
         root project : > org.apache.grails:grails-bom:7.0.0-SNAPSHOT:20250520.032711-84 > org.apache.grails:grails-spring:7.0.0-SNAPSHOT:20250520.032711-84 > org.apache.groovy:groovy-xml:4.0.26
      > Module 'org.apache.groovy:groovy' has been rejected:
           Cannot select module with conflict on capability 'org.codehaus.groovy:groovy:4.0.26' also provided by [org.codehaus.groovy:groovy:3.0.24(runtime)]
   > Could not resolve org.codehaus.groovy:groovy.
     Required by:
         root project : > org.apache.grails:grails-gradle-plugins:7.0.0-SNAPSHOT:20250520.001817-109 > org.apache.grails:grails-gradle-tasks:7.0.0-SNAPSHOT:20250520.001817-1
      > Module 'org.codehaus.groovy:groovy' has been rejected:
           Cannot select module with conflict on capability 'org.codehaus.groovy:groovy:3.0.24' also provided by [org.apache.groovy:groovy:4.0.26(groovyRuntimeElements)]
> There are 2 more failures with identical causes.
```

### DIFF
--- a/grails-gradle/tasks/build.gradle
+++ b/grails-gradle/tasks/build.gradle
@@ -29,10 +29,9 @@ group = 'org.apache.grails'
 dependencies {
     implementation platform(project(':grails-gradle-bom'))
 
-    implementation "org.codehaus.groovy:groovy"
+    implementation "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}"
 
     implementation project(':grails-gradle-model'), {
-        exclude group: 'org.apache.groovy'
         exclude group: 'org.spockframework'
     }
 
--- a/grails-gradle/tasks/build.gradle
+++ b/grails-gradle/tasks/build.gradle
@@ -31,9 +31,7 @@ dependencies {
 
     implementation "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}"
 
-    implementation project(':grails-gradle-model'), {
-        exclude group: 'org.spockframework'
-    }
+    implementation project(':grails-gradle-model')
 
     implementation 'io.github.gradle-nexus:publish-plugin'
     implementation 'org.springframework.boot:spring-boot-gradle-plugin'
--- a/grails-gradle/tasks/build.gradle
+++ b/grails-gradle/tasks/build.gradle
@@ -31,7 +31,10 @@ dependencies {
 
     implementation "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}"
 
-    implementation project(':grails-gradle-model')
+    // spock is leaking from the grails-gradle-bom through grails-gradle-model
+    implementation project(':grails-gradle-model'), {
+        exclude group: 'org.spockframework'
+    }
 
     implementation 'io.github.gradle-nexus:publish-plugin'
     implementation 'org.springframework.boot:spring-boot-gradle-plugin'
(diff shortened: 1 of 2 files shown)


grails-gradle/tasks/build.gradle
```diff
@@ -29,10 +29,9 @@ group = 'org.apache.grails'
 dependencies {
     implementation platform(project(':grails-gradle-bom'))
 
-    implementation "org.codehaus.groovy:groovy"
+    implementation "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}"
 
     implementation project(':grails-gradle-model'), {
-        exclude group: 'org.apache.groovy'
         exclude group: 'org.spockframework'
```
matrei, line 36: Why are we excluding `spock` here?
Is `spock` a transitive dependency of `grails-gradle-model`?

matrei (approved):
Sweet!

grails-gradle/tasks/build.gradle
jdaugherty: this was originally because it used spock that was compatible with groovy 4 only and this code was from the plugins project which was including the 3 compatible version.  I think the exclude can be removed now.

grails-gradle/tasks/build.gradle
```diff
@@ -29,10 +29,9 @@ group = 'org.apache.grails'
 dependencies {
     implementation platform(project(':grails-gradle-bom'))
 
-    implementation "org.codehaus.groovy:groovy"
+    implementation "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}"
```
jdaugherty, line 32: have you done a dependency check on the plugins project? we may need to exclude this there to make sure there isn't pollution in the dependency graph

grails-gradle/tasks/build.gradle
jdaugherty: (we need to confirm this exclude doesn't need moved to the plugins import of this though)

grails-gradle/tasks/build.gradle
jamesfredley: I am adding an exclude on grails-gradle-plugins to limit any pollution.   It might be a good idea to experiment with them all using Groovy 4 prior to RC1, but this will limit the scope of the change for now.

grails-gradle/tasks/build.gradle
jamesfredley: `exclude group: 'org.spockframework'` on `implementation project(':grails-gradle-model')` does not appear to be needed any longer based on the generated POM for grails-gradle-model.  Pushing a commit that removes it from two locations.

jdaugherty (approved):

PR #14764, opened 2025-05-21
Automatically retry snapshot publishing - 3 times

jamesfredley:
nick-fields/retry is on the ASF approved list

Automatically retries 3 times and reduces time needed to babysit the workflow when repository.apache.org is experiencing read timeouts.


jdaugherty (approved):

matrei (approved):
Should we add a comment explaining which problem we are solving by using the retry action (timeouts from https://repository.apache.org/content/repositories/snapshots)?

jamesfredley:
Comment added

`# Attempts to address: Could not write to resource 'https://repository.apache.org/content/repositories/snapshots/...' Read timed out`

PR #14765, opened 2025-05-22
Support disabling spring dependency management plugin & disable in core

jdaugherty:
Was troubleshooting the groovy reproducibility changes, and realized that spring dependency management was causing the version of groovy to cache unexpectedly.  Even after clearing / ignoring the cache.  

Given how much we've talked about adding this support, this PR adds an option to disable spring dependency management and uses it to turn it off in core.  

I intentionally am leaving this undocumented for us to run with it for awhile before advocating a larger adoption.

### DIFF
--- a/grails-gsp/spring-boot/build.gradle
+++ b/grails-gsp/spring-boot/build.gradle
@@ -40,7 +40,8 @@ dependencies {
 apply {
     // java-configuration must be applied first since tasks are now lazy registered
     from rootProject.layout.projectDirectory.file('gradle/java-config.gradle')
-    from rootProject.layout.projectDirectory.file('gradle/docs-config.gradle')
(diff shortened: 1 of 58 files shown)


grails-gsp/spring-boot/build.gradle
matrei, line 1: Is this project active? I don't see it in the `settings.gradle` file.
And the `grails-test-examples-gsp-spring-boot` project does not seem to be setup correctly.
It is added in `settings.gradle` but the `build.gradle` file is in an `app` subfolder.

matrei (approved):

grails-gsp/spring-boot/build.gradle
jdaugherty: good catch.  I fixed the spring boot test app & the spring boot gsp app.

jamesfredley (approved):

PR #14768, opened 2025-05-22
Configures a task to verify that essential Grails project directories exist and creates them if missing

### LINKED ISSUE #587: grails forge: services and domain directories are missing
Description
A fresh Grails 6 application is missing a number of directories

Steps to Reproduce
- Goto https://start.grails.org/
- Click on the "Generate Project" button and Download the Zip or upload to Github
- Open the Project in an IDE or view on file system
- Notice the following directories are missing:

1. grails-app/services
2. grails-app/domain
3. grails-app/taglib
4. grails-app/migrations
5. grails-app/conf/spring - addressed in 6.2.1 and 7.0.0
And the file grails-app/conf/spring/resources.groovy - addressed in 6.2.1 and 7.0.0

jamesfredley:
runs on project creation or sync in IntelliJ or on most other build tasks

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -776,7 +776,7 @@ class GrailsGradlePlugin extends GroovyPlugin {
 
         verifyGrailsProjectDirectoriesTask.configure { Task task ->
             task.group = 'other'
-
+            task.description = 'Ensures essential Grails project directories exist and creates them if missing.'
             task.doFirst {
                 ['grails-app/services', 'grails-app/domain',
                  'grails-app/taglib', 'grails-app/migrations'].each { String dir ->
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -162,6 +162,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
                 }
             """.stripIndent(16)
         }
+
+        verifyGrailsProjectDirectories(project)
     }
 
     private void configureGroovyCompiler(Project project) {
@@ -767,5 +769,33 @@ class GrailsGradlePlugin extends GroovyPlugin {
             fileCollection = fileCollection + it.filter({ File file -> !file.name.startsWith('spring-boot-devtools') })
         }
         fileCollection
+    }
+
+    private void verifyGrailsProjectDirectories(Project project) {
+        TaskProvider<Task> verifyGrailsProjectDirectoriesTask = project.tasks.register('verifyGrailsProjectDirectories')
+
+        verifyGrailsProjectDirectoriesTask.configure { Task task ->
+            task.group = 'other'
+
+            task.doFirst {
+                ['grails-app/services', 'grails-app/domain',
+                 'grails-app/taglib', 'grails-app/migrations'].each { String dir ->
+                    Directory directory = project.layout.projectDirectory.dir(dir)
+                    if (!directory.asFile.exists()) {
+                        directory.asFile.mkdirs()
+                    }
+                }
+            }
+        }
+
+        project.afterEvaluate {
+            // prepareKotlinBuildScriptModel runs when a project is created or synced in IntelliJ
+            // cleanGroovyCompilerConfig is run by most other tasks
+            ['prepareKotlinBuildScriptModel', 'cleanGroovyCompilerConfig'].each { String taskName ->
+                it.tasks.findByName(taskName)?.configure { Task dependent ->
+                    dependent?.dependsOn verifyGrailsProjectDirectoriesTask
+                }
+            }
+        }
     }
 }
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -806,10 +806,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
         project.afterEvaluate {
             // prepareKotlinBuildScriptModel runs when a project is created or synced in IntelliJ
             // cleanGroovyCompilerConfig is run by most other tasks
-            ['prepareKotlinBuildScriptModel', 'cleanGroovyCompilerConfig'].each { String taskName ->
-                it.tasks.findByName(taskName)?.configure { Task dependent ->
-                    dependent?.dependsOn verifyGrailsProjectDirectoriesTask
-                }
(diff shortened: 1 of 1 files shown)


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+
+    /**
+     * Configures a task to verify that essential Grails project directories exist and creates them if missing.
+     *
+     * This method registers a task named 'verifyGrailsProjectDirectories' that creates the following
+     * directories if they do not already exist:
+     * - grails-app/services
+     * - grails-app/domain
+     * - grails-app/taglib
+     * - grails-app/migrations
```
jdaugherty, line 782: not everyone has the migraiton plugin and not every project has them. I don't agree on creating this by default. 

I'm ok creating this if the migration plugin is applied to the project though.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                    }
+                }
+            }
+        }
+
+        project.afterEvaluate {
+            // prepareKotlinBuildScriptModel runs when a project is created or synced in IntelliJ
+            // cleanGroovyCompilerConfig is run by most other tasks
+            it.tasks.matching { Task task ->
+                task.name in ['prepareKotlinBuildScriptModel', 'cleanGroovyCompilerConfig'] }
```
jdaugherty, line 812: Instead of cleanGroovyCompilerConfig we should be a dependency of classes, since that's what triggers a compile.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+     *
+     * @param project The Gradle project for which the directories are verified and tasks are configured.
+     */
+    private void verifyGrailsProjectDirectories(Project project) {
+        TaskProvider<Task> verifyGrailsProjectDirectoriesTask = project.tasks.register('verifyGrailsProjectDirectories')
+
+        def grailsProjectDirectories = ['grails-app/services', 'grails-app/domain',
+                                 'grails-app/taglib', 'grails-app/migrations']
+
+        verifyGrailsProjectDirectoriesTask.configure { Task task ->
```
jdaugherty, line 795: This task has no outputs, so gradle can't cache it.  

`outputs.dirs(grailsProjectDirectories.collect { project.layout.projectDirectory.dir(it) })`

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+            task.group = 'build'
+            task.description = 'Ensures essential Grails project directories exist and creates them if missing.'
+            task.doFirst {
```
jdaugherty, line 798: Since this task will run before others, I'd expect this to be doLast() instead of doFirst()

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                grailsProjectDirectories.each { String dir ->
```
jdaugherty, line 799: I don't think you need to check if the directory exists, you can just call project.mkdir(dir):

        grailsProjectDirectories.each { project.mkdir(it) }

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                    Directory directory = project.layout.projectDirectory.dir(dir)
+                    if (!directory.asFile.exists()) {
+                        directory.asFile.mkdirs()
+                    }
+                }
+            }
+        }
+
+        project.afterEvaluate {
```
jdaugherty, line 808: After evaluate is the wrong solution here.  it should be only used if you can't configure the other tasks.  It's better to grab the tasks lazily and then configure them to depend on your task.

        project.tasks.named('prepareKotlinBuildScriptModel').configure { it.dependsOn(verifyGrailsProjectDirectoriesTask) }

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+            // prepareKotlinBuildScriptModel runs when a project is created or synced in IntelliJ
+            // cleanGroovyCompilerConfig is run by most other tasks
+            it.tasks.matching { Task task ->
```
jdaugherty, line 811: This is extremely bad.  It will unwrap every task and initialize them.  https://docs.gradle.org/current/userguide/task_configuration_avoidance.html covers that this shouldn't be used.  Use the .named() approach I mention previously.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -768,4 +770,49 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
     }
+
+    /**
+     * Configures a task to verify that essential Grails project directories exist and creates them if missing.
+     *
+     * This method registers a task named 'verifyGrailsProjectDirectories' that creates the following
+     * directories if they do not already exist:
+     * - grails-app/services
+     * - grails-app/domain
+     * - grails-app/taglib
```
jdaugherty, line 781: The more I think about this, we probably should have different gradle plugins doing some of this.  taglib wouldn't make sense in a rest app since they're likely only using grails for the backend.  I'd suggest the taglib creation to be done in the grails-gsp plugin?

matrei (commented):
I'm not sure it is a good idea to have the build make changes in the source trees.
I think creating these directories should only be part of project creation.

jdaugherty:
> I'm not sure it is a good idea to have the build make changes in the source trees. I think creating these directories should only be part of project creation.

The more I've thought about this, the more I agree.  

Can we not have this task run by default? 
Can we have forge create these empty directories on app create?

jamesfredley:
Creating the directories in Forge was the solution I had proposed:  https://github.com/apache/grails-forge/issues/587.  

This was a waste of time.

PR #14781, opened 2025-06-04
Verify build explanation and minor fix

sbglasius:
Added explanation on how to find staging repo id, and fixed a current directory that was wrong

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -71,7 +71,7 @@ For Example:
     verify.sh orgapachegrails-1030 v7.0.0-M4 /tmp/grails-verify
 ```
 
-The `staging repo id` can be found in awaiting https://github.com/apache/grails-core/actions/workflows/release.yml, click `publish`, open the stage `Create Staging Repository` and look for the line starting with `Created staging repository` 
+The `staging repo id` can be found under https://github.com/apache/grails-core/actions/workflows/release.yml, click `publish`, open the stage `Create Staging Repository` and look for the line starting with `Created staging repository` 
 
 ### Download the Staged Artifacts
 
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -71,6 +71,8 @@ For Example:
     verify.sh orgapachegrails-1030 v7.0.0-M4 /tmp/grails-verify
 ```
 
+The `staging repo id` can be found in awaiting https://github.com/apache/grails-core/actions/workflows/release.yml, click `publish`, open the stage `Create Staging Repository` and look for the line starting with `Created staging repository`
```
jdaugherty: You can also login to repository.apache.org and just click "Staging Repositories" on the left side.  That's how I've been doing it.

jdaugherty (approved):

jdaugherty:
Please do not merge this until after M4 has been released since we're waiting on the vote.

etc/bin/verify-jar-artifacts.sh
```diff
@@ -56,6 +56,8 @@ error() {
 }
 trap error ERR
 
+cd "${DOWNLOAD_LOCATION}"
```
jdaugherty, line 59: Ah, I always had the download_location = current directory in my testing.  Good catch!

jdaugherty (approved):

RELEASE.md
sbglasius: Yes, but if you're not committer, I assume it would also be possible to use my method?

jdaugherty: Oh absolutely, I agree with your proposal for that reason.

PR #14784, opened 2025-06-05
initial attempt at GitHub -> Mailing List notifications

jamesfredley:
https://lists.apache.org/list.html?commits@grails.apache.org is currently receiving most of these.

https://lists.apache.org/list.html?notifications@grails.apache.org, https://lists.apache.org/list.html?users@grails.apache.org and https://lists.apache.org/list.html?dev@grails.apache.org are not currently receiving any of these notification types

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,3 +1,20 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# see https://s.apache.org/asfyaml
 github:
   # deployment environments to delay releasing until the necessary steps have completed
   environments:
@@ -22,3 +39,14 @@ github:
         - id: jdaugherty
           type: User
       wait_timer: 0
+notifications:
+  # GitHub related
+  commits: commits@grails.apache.org
+  jobs: notifications@grails.apache.org
+  pullrequests_status: dev@grails.apache.org
+  pullrequests_comment: dev@grails.apache.org
+  pullrequests_bot_dependabot: dev@grails.apache.org
+  issues: users@grails.apache.org
+  issues_status: users@grails.apache.org
+  issues_comment: users@grails.apache.org
+  discussions: users@grails.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -43,9 +43,9 @@ notifications:
   # GitHub related
   commits: commits@grails.apache.org
   jobs: notifications@grails.apache.org
-  pullrequests_status: commits@grails.apache.org
-  pullrequests_comment: commits@grails.apache.org
-  pullrequests_bot_dependabot: commits@grails.apache.org
+  pullrequests_status: notifications@grails.apache.org
+  pullrequests_comment: notifications@grails.apache.org
+  pullrequests_bot_dependabot: notifications@grails.apache.org
   issues: notifications@grails.apache.org
   issues_status: notifications@grails.apache.org
   issues_comment: notifications@grails.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -46,7 +46,7 @@ notifications:
   pullrequests_status: commits@grails.apache.org
   pullrequests_comment: commits@grails.apache.org
   pullrequests_bot_dependabot: commits@grails.apache.org
-  issues: users@grails.apache.org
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# see https://s.apache.org/asfyaml
+
+notifications:
+  # GitHub related
+  commits: commits@grails.apache.org
+  jobs: notifications@grails.apache.org
+  pullrequests_status: dev@grails.apache.org
```
jdaugherty: Don't we want the dev list as a communication for non-automated?  

I would have assumed these would go under commits.

.asf.yaml
```diff
@@ -0,0 +1,29 @@
...
+  pullrequests_comment: dev@grails.apache.org
+  pullrequests_bot_dependabot: dev@grails.apache.org
+  issues: users@grails.apache.org
```
jdaugherty: Shouldn't issues go under notifications?  Same as dev - users is meant for people asking questions?

.asf.yaml
jamesfredley: I like both directions.  This would surface PRs in dev@ for those not subscribed to commits and notifications.

jdaugherty: I really don't want this because I'm already subscribed to github notifications and it would lead to what I consider spam on mailing list we're trying to get others to subscribe to.

.asf.yaml
jamesfredley: I am not dogmatic about these, but the thought was that emails to users@, github issues, github discussions and then if it ends up being possible to map slack #questions and potentially https://stackoverflow.com/feeds/tag?tagnames=grails&sort=newest, then all user related questions will be archived in the users@ mailing list.

.asf.yaml
jamesfredley: PR related notifications currently land in commits@.  If we don't surface PR notifications in dev@, I lean towards them going into notifications@ instead.

jamesfredley: My thought process is similar, since GitHub direct email notification already keep me updated.  

Here is what the quantity looks like:  https://lists.apache.org/list?commits@grails.apache.org:lte=1M:[PR]

jamesfredley: After looking at the monthly volume between 300 to 600, I agree PRs need to be in commits.  Updating PR.

.asf.yaml
jamesfredley: Example Issue with comments:  https://lists.apache.org/thread/dp3pp6tlphxv5h44mf8gnrfl79m65qgc

jdaugherty: Issues aren't really meant for questions, we should be pushing people to discussions or the user list for those.  I'd rather see issues go to notifications.  

We open a lot of issues for bugs / features, and the majority will be out of scope for users.  Ideally, we'd send slack & discussions to the user list.  We can update our issue template to even refer people to those locations to minimize ticket based questions.

jamesfredley: I have updated all issues notifications to go to notifications@

jdaugherty (approved):

jamesfredley:
I am onboard with guiding user questions to locations other than issues.  That will attempt to address my concern about users submitting questions as issues.  We already do that to some extent and may update it further.

image


https://github.com/apache/grails-core/issues/new/choose

https://github.com/apache/grails-core/tree/7.0.x/.github/ISSUE_TEMPLATE

.asf.yaml
```diff
@@ -0,0 +1,29 @@
...
+#
+# see https://s.apache.org/asfyaml
+
+notifications:
+  # GitHub related
+  commits: commits@grails.apache.org
+  jobs: notifications@grails.apache.org
+  pullrequests_status: commits@grails.apache.org
+  pullrequests_comment: commits@grails.apache.org
+  pullrequests_bot_dependabot: commits@grails.apache.org
```
matrei: I saw your previous conversation about putting pull request events in `dev` vs `commits` vs `notifications`. But I'm still not clear why we would choose to put them in `commits` rather than `notifications`.

>  After looking at the monthly volume between 300 to 600, I agree PRs need to be in commits. Updating PR.

Is this a problem in the `notifications` mailing list? What is the use case for the `notifications` mailing list?

jdaugherty: I am indifferent to having them in commits or notifications.  @jamesfredley what are your thoughts?

jdaugherty: As for the volume statement, I think James F put this here to confirm we should not direct these to the dev or user mailing lists.

jamesfredley: The volume statements were to gauge the level of activity that would end up in users@ and dev@, which we want to keep focused to user questions and development, respectively.  After reviewing the volume, it changed my mind about adding PRs and Issues to dev@ or users@.

For PRs notifications, I would be OK with `commits@` or `notifications@`, but was leaning slightly towards `commits@`, since PRs are closely related to the commits.  Either way is fine.

Groovy puts PRs in https://lists.apache.org/list.html?notifications@groovy.apache.org - that would keep commits clean, as only commits.

matrei: Thank you for the clarification. Then I think `notifications@` is the logical place for GitHub PR event messages, and keep the `commits@` purely for commits.

jamesfredley: PRs updated go to notifications@

jdaugherty (approved):

matrei (approved):

jamesfredley:
We will merge this after M4 is released and the 7.0.x branch is back to snapshot.

Issue #14780, opened 2025-06-04
Deprecated Gradle features used with Grails 7


Hardes1:
### Issue description

Hi, I am currently trying to support Grails 7 inside IntelliJ based IDEs, ticket is [here](https://youtrack.jetbrains.com/issue/IDEA-372363/Update-Grails-Plugin-to-support-Apache-Grails-changes). [Release notes](https://grails.org/blog/2024-12-23-grails-7-m1.html) says that it should be compatible with Gradle 8, but
I have noticed that the Grails plugins like `org.apache.grails.gradle.grails-web` or `org.apache.grails.gradle.grails-gsp` use deprecated Gradle features (the output of this warning can be seen when configuring a Gradle project).

While likely it doesn't affect the work of the framework, I can't write tests to ensure that the import is correct. Could you, please, fix it?

Example of the project you gave to me can be found [here](https://github.com/jamesfredley/grails-7.0.0-SNAPSHOT-APACHE)

jdaugherty:
Thank you for opening the ticket.  A recent build showing some of the deprecated Gradle features is here:  https://ge.grails.org/s/2cais25vnxmbs/deprecations 

At a high level, they are: 

1. (Gradle 9) setFileMode() usages in copyspec
2. (Gradle 9) setDirMode() usages in copyspec
3. (Gradle 10) Invocation of Task.project at execution time has been deprecated.
4. (Gradle 9) Relying on the convention for Test.classpath in custom Test tasks has been deprecated.
5. (Gradle 9) The StartParameter.isConfigurationCacheRequested property has been deprecated.
6. (Gradle 9) The IdeaModule.testSourceDirs property has been deprecated.


1 & 2 are planned to be fixed in the next RC.

3 needs fixed in the asset plugin for us to fully resolve, so we may not be able to resolve this in a timely manner.  

4 is deprecated, but the replacement is still considered experimental.

5 is dependent on the [Asciidoctor plugin](https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/751), so we may not be able to resolve this in a timely manner.

6 I can fix in the next RC.

Since some of these deprecations are dependent on external plugins / gradle standardizing an API, can you write tests by disabling the warning mode?  Either by setting the property `org.gradle.warning.mode=none` or by using an environment variable (maybe export GRADLE_OPTS="-Dorg.gradle.warning.mode=none")?

jdaugherty:
[...]

PR #14805, opened 2025-06-11
Gradle deprecations

jdaugherty:
Fixes some of the issues found by #14780


grails-wrapper/build.gradle
```diff
@@ -97,7 +97,9 @@ project.extensions.getByType(DistributionContainer).configureEach {
 
         from(startScripts) {
             into ''
-            fileMode = 0755
+            filePermissions { permissions ->
+                permissions.unix("755")
```
matrei: Do we want to be consistent on the method we use (`unix(String)` vs `unix(int)`)? I kind of like the `int` variant.

matrei (approved):

grails-wrapper/build.gradle
jamesfredley: I agree on int variant

jamesfredley (approved):

jdaugherty:
This should bring us down from 500+ to 43 deprecations

matrei:
Do we know why `testGrailsVersion()` is failing?
https://ge.grails.org/s/qces4gufrk4to/tests/overview?outcome=FAILED

PR #14808, opened 2025-06-12
Unnest the snapshot zip uploaded to workflow results

jamesfredley:
This does not run the publish steps until the PR is merged.

Example with nesting:  https://github.com/apache/grails-core/actions/runs/15598307628/artifacts/3311305695

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -266,7 +266,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: wrapper-artifacts
-          path: grails-wrapper/build/distributions/apache-grails-wrapper-*.zip
+          path: tmp/wrapper
   docs:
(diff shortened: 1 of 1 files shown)


matrei (approved):

.github/workflows/gradle.yml
```diff
@@ -266,7 +266,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: wrapper-artifacts
```
jdaugherty: shouldnt we add 'snapshot' to this?

jdaugherty (approved):

.github/workflows/gradle.yml
jamesfredley: Updated to `apache-grails-wrapper-SNAPSHOT-incubating-bin`

PR #14811, opened 2025-06-13
Use developmentOnly Gradle configuration scope for webjars

jamesfredley:


irllyliketoast (approved):
Looks good

matrei:
@jamesfredley I'd like to understand the reasoning behind this update — could you provide some context?

jdaugherty:
We want our functional tests matching how our app generators are generating grails apps.

matrei:
@jdaugherty OK, that makes sense, but with `developmentOnly`, I don't think the webjar assets will be loaded when running integration tests.

jdaugherty:
@matrei wouldn't this be a problem for end apps too then?  How would they load assets and test them?

matrei:
>  wouldn't this be a problem for end apps too then?  How would they load assets and test them?

Yes, I believe so.

PR #14812, opened 2025-06-14
Support specifying the publication name so that gradle plugins can be published

jdaugherty:
Working on the asset plugin, we need to be able to specify the maven publication name (pluginMaven) so gradle plugins can be published with the grails publish plugin


jdaugherty:
I'm going to go ahead and merge this so i can remain productive on the asset plugin.  @matrei @jamesfredley  can you let me know if you disagree with the change and I'll revise?

matrei:
@jdaugherty Looks good!

PR #14826, opened 2025-06-24
fix: update `grails-docs` coords in `GrailsDocGradlePlugin`

matrei:


jdaugherty (approved):
Wow, how did we miss this.  Good catch!

jamesfredley (approved):

PR #14827, opened 2025-06-24
Support not adding components from the grails publish plugin

jdaugherty:
In order to publish gradle plugins, we must let the [gradle publish plugin](https://plugins.gradle.org/docs/publish-plugin) handle adding components.


jamesfredley (approved):

jdaugherty:
@matrei  given how simple of a change this is to support gradle plugin publishing, I'm going to go ahead and merge it.  Let me know if you disagree or want to revisit this.

Issue #14142, opened 2025-01-10
Standardize on Gradle Platforms or Spring Dependency Management Gradle Plugin for grails-bom application


jamesfredley:
Gradle Platforms:  https://docs.gradle.org/current/userguide/platforms.html
`implementation platform("org.grails:grails-bom:$grailsVersion")`

vs 
io.spring.dependency-management plugin: https://docs.spring.io/spring-boot/gradle-plugin/managing-dependencies.html
```
dependencyManagement {
	imports {
		mavenBom "org.grails:grails-bom:$grailsVersion")
	}
}
```

Currently the `grails-gradle-plugin` automatically applies `grails-bom` via the Spring Dependency Management Gradle Plugin and also applies for spring-boot-dependencies, which is ALSO included in `grails-bom`  with an import scope.

```
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>${spring-boot.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>  
    </dependencies>
  </dependencyManagement>
```

We then apply `grails-bom` via Gradle Platforms in `build.gradle dependencies {}`, `buildSrc/build.gradle dependencies{}` and are in the process of adding it to `build.gradle buildscript{} dependencies {}` to make sure versions are applied and consistent.  Since `build.gradle buildscript{} dependencies {}` does not appear to support adding grails-bom via a Gradle Plugin to control versions, if we go the Gradle Platforms route, we should just put the lines in the build.gradle file, instead of magically applying via the grails-Gradle plugin, so that it is consistent in all locations.

We need to standardize on one or the other.

When Customizing/Overridding Managed Versions, which should be rare, overriding versions directly in the `dependencies{}` block, IE. `runtimeOnly "org.fusesource.jansi:jansi:1.18"`, is quicker/simpler/easier than the two examples provided by Spring.  And it is doing it in the normal dependencies location instead of having to use another mechanism.  

Andy Wilkinson, the io.spring.dependency-management plugin author:  https://stackoverflow.com/a/76540836/26842213

> The most notable difference in terms of performance is that Gradle's platform and enforcedPlatform support should be noticeably more efficient. Speaking as the main author of the dependency management plugin, unless you want to make use of the ability to override version properties, I would encourage you to use Gradle's platform and enforcedPlatform support.

[...]

PR #14829, opened 2025-06-24
fix: scope webjars as testAndDevelopmentOnly

jamesfredley:
Apply changes from https://github.com/apache/grails-forge/pull/600

### DIFF
--- a/grails-data-neo4j/examples/grails3-neo4j-hibernate/build.gradle
+++ b/grails-data-neo4j/examples/grails3-neo4j-hibernate/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     implementation "org.apache.grails:grails-data-hibernate5:$hibernateDatastoreVersion"
     implementation "org.hibernate:hibernate-ehcache:$hibernateEhcacheVersion"
 
+    testAndDevelopmentOnly platform(project(':grails-bom'))
     testAndDevelopmentOnly 'org.webjars.npm:jquery'
 
     runtimeOnly "cloud.wondrify:asset-pipeline-grails:$assetPipelineVersion"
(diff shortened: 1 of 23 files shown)


jdaugherty (approved):

matrei:
I'm not sure we need to add `grails-bom` as a platform to `testAndDevelopmentOnly` as it is a Spring configuration and we are adding the bom to the Spring Dependency Management Plugin in `GrailsGradlePlugin`. It seems to work without it.

jamesfredley:
I was glad to remove `grails-bom` as a platform on `testAndDevelopmentOnly`.  It appeared to be required when we switched from `implementation `to `developmentOnly `(https://github.com/apache/grails-core/actions/workflows/gradle.yml?query=branch%3Awebjars-developmentOnly)

jamesfredley:
```
* What went wrong:
Execution failed for task ':grails-test-examples-namespaces:bootJar'.
> Could not resolve all files for configuration ':grails-test-examples-namespaces:testAndDevelopmentOnly'.
   > Could not find org.webjars.npm:jquery:.
     Required by:
         project :grails-test-examples-namespaces
```

matrei:
> ```
> * What went wrong:
> Execution failed for task ':grails-test-examples-namespaces:bootJar'.
> > Could not resolve all files for configuration ':grails-test-examples-namespaces:testAndDevelopmentOnly'.
>    > Could not find org.webjars.npm:jquery:.
>      Required by:
>          project :grails-test-examples-namespaces
> ```

That's interesting. Gradle seems to be able to resolve the webjar versions in apps generated by Forge, without the platform added.

matrei:
Aha, this is why the platform is needed in the functional test apps:
https://github.com/apache/grails-core/blob/ead00352c90f2f7e263d64205302c3171873d99f/gradle/grails-extension-gradle-config.gradle#L33

jamesfredley:
Based on https://github.com/apache/grails-core/blob/ead00352c90f2f7e263d64205302c3171873d99f/gradle/grails-extension-gradle-config.gradle#L28-L34

Given we use project()  in these test projects, it appears we need to keep the dependency management plugin disabled and add grails-bom as a platform for testAndDevelopmentOnly.  

This is different than generated apps, but we are moving in this direction:  https://github.com/apache/grails-core/issues/14142

matrei (approved):

PR #14831, opened 2025-06-26
database-migration mapped to grails-data-hibernate5-dbmigration

jamesfredley:
The extra rule was mapping it to grails-data-hibernate6-dbmigration

"org[.]grails[.]plugins:database-migration|org.apache.grails:grails-data-hibernate6-dbmigration"

### DIFF
--- a/etc/bin/rename_gradle_artifacts.sh
+++ b/etc/bin/rename_gradle_artifacts.sh
@@ -184,7 +184,6 @@ declare -a gorm_mappings=(
   "org[.]grails[.]plugins:mongodb|org.apache.grails:grails-data-mongodb"
   "org[.]grails[.]plugins:hibernate6|org.apache.grails:grails-data-hibernate6"
   "org[.]grails[.]plugins:hibernate5|org.apache.grails:grails-data-hibernate5"
-  "org[.]grails[.]plugins:database-migration|org.apache.grails:grails-data-hibernate6-dbmigration"
   "org[.]grails[.]plugins:database-migration|org.apache.grails:grails-data-hibernate5-dbmigration"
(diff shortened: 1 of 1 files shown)


etc/bin/rename_gradle_artifacts.sh
```diff
@@ -184,7 +184,6 @@ declare -a gorm_mappings=(
   "org[.]grails[.]plugins:mongodb|org.apache.grails:grails-data-mongodb"
   "org[.]grails[.]plugins:hibernate6|org.apache.grails:grails-data-hibernate6"
   "org[.]grails[.]plugins:hibernate5|org.apache.grails:grails-data-hibernate5"
-  "org[.]grails[.]plugins:database-migration|org.apache.grails:grails-data-hibernate6-dbmigration"
```
jdaugherty, line 187: so the plan was to split this since we voted on this.  Instead of removing it should we just leave the source empty?

jamesfredley: I still expect us to publish `grails-data-hibernate5-dbmigration` and separately `grails-data-hibernate6-dbmigration` (when ready).  

The issue when you run `rename_gradle_artifacts.sh` is that `org.grails.plugins:database-migration` gets replaced with `org.apache.grails:grails-data-hibernate6-dbmigration` which does not currently exist and I think historical `org.grails.plugins:database-migration` dependencies should be replaced with `grails-data-hibernate5-dbmigration`

jamesfredley: Yes, I think, on source empty.

jdaugherty: Didn't realize this was the script instead of RENAME.md

jdaugherty (approved):

PR #14842, opened 2025-06-30
Replace google analytics with Matomo, analytics.apache.org

jamesfredley:
Also adjusts indention on addJsClass script.

### DIFF
--- a/grails-gradle/docs-core/src/main/template/style/guideItem.html
+++ b/grails-gradle/docs-core/src/main/template/style/guideItem.html
@@ -2,26 +2,34 @@
                       "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
-    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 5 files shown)


jdaugherty (approved):

grails-gradle/docs-core/src/main/template/style/guideItem.html
```diff
@@ -2,26 +2,34 @@
                       "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
-    <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script src="/cdn-cgi/apps/head/3cOPSgo5Omz84ycX7CvigfX4cpw.js"></script><script async src="https://www.googletagmanager.com/gtag/js?id=UA-82213539-2"></script>
```
jdaugherty, line 6: I had no idea that the publish guide templates would include tracking info.  I guess since we primarily use it that's ok, but several plugins also made use of this.  Should we really be exporting that?  Should this be something we have gated behind a feature flag or configuration?

jamesfredley: https://github.com/apache/grails-core/issues/14845

PR #14844, opened 2025-06-30
Correct dependency to `grails-codecs-core`

### LINKED ISSUE #14843: EncodeAsXXX Codecs not working in Grails 7
### Expected Behavior

In Grails < 7, when the grails-gsp plugin is enabled, the grails-codec package would be included. This package had many classes such as https://github.com/apache/grails-core/blob/6.0.x/grails-codecs/src/main/groovy/org/grails/plugins/codecs/Base64CodecExtensionMethods.groovy which would be added as default string methods via groovy.

uses of

String x="abc"
x.encodeAsBase64() 

would succeed

### Actual Behaviour

These classes have been moved to grails-codecs-core which is not included when using grails-gsp.

As a result, the above calls fail

`groovy.lang.MissingMethodException: No signature of method: java.lang.String.encodeAsBase64() is applicable for argument types: () values: []
Possible solutions: decodeBase64(), encodeAsJson(), encodeAsHTML4(), encodeAsJs(), encodeAsRaw()
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:72)
	at org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.unwrap(IndyGuardsFiltersAndSignatures.java:163)
	at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache(IndyInterface.java:321)`

### Steps To Reproduce

Codecs such as [Base64CodecExtensionMethods.groovy](https://github.com/apache/grails-core/blob/6.0.x/grails-codecs/src/main/groovy/org/grails/plugins/codecs/Base64CodecExtensionMethods.groovy) were moved out of grails-codecs into a new package called grails-codecs-core. How are these default groovy extensions loaded? Previously these would be loaded through the gsp plugin -> grails-codecs dependency. In latest Grails 7, these appear not to be loaded any more so calls to x.encodeAsBase64 / x.encodeAsMD5() no longer work.
Is this a bug / or a breaking change that is not documented? grails-codecs-core is not published in maven either?

### Environment Information

_No response_

### Example Application

_No response_

### Version

7.0.0-M4

matrei:
When migrating to Apache coordinates, `grails-codecs` was renamed to `grails-codecs-core`, and `grails-plugin-codecs` became `grails-codecs`. This commit updates a reference to `grails-codecs` that should have been changed to `grails-codecs-core` but was missed earlier.

Resolves gh-14843

### DIFF
--- a/grails-codecs/build.gradle
+++ b/grails-codecs/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     api project(':grails-web-core')
     api project(':grails-encoder')
     api 'org.apache.groovy:groovy'
-    runtimeOnly project(':grails-codecs')
+    runtimeOnly project(':grails-codecs-core')
 
     compileOnly 'jakarta.servlet:jakarta.servlet-api'
     compileOnly 'org.springframework:spring-test', {


jamesfredley (approved):

jdaugherty (commented):
Thank you for handling this!

PR #14890, opened 2025-07-10
Restore Sitemesh3GrailsPlugin

codeconsole:


gradle.properties
```diff
@@ -35,6 +35,7 @@ joddWotVersion=3.3.8
 elApiVersion=5.0.1
 defaultElImplementationVersion=5.0.0
 jspApiVersion=4.0.0
+sitemeshVersion=3.2.2
```
jdaugherty, line 38: We should name this sitemesh3Version or something so it's clear it's not 2.x.  Should we include sitemesh in the bom at all with both too?   Maybe we should just hard the covers in each plugin and leave off the bom so they don't ever try to resolve from the bom.

jdaugherty:
I'll try to get some tests added so that we can test including sitemesh3 or sitemesh2.  We'll probably want a validation where you can't include both.

jdaugherty:
Closing to reopen against the grails-core branch

gradle.properties
codeconsole: @jdaugherty that could be renamed to sitemesh3Version.  I would think it is ok to include both in the sitemesh 2 and sitemesh 3 in the bom

PR #14891, opened 2025-07-10
Restore Sitemesh3GrailsPlugin

jdaugherty:


jdaugherty:
Previously, sitemesh2 existed across 2 coordinates: `org.grails.plugins:sitemesh` and `org.grails:grails-web-sitemesh`.   From my research in Grails 6, to use sitemesh3, the grails-web-sitemesh library was still included in the application.  Per discussion with Scott, we wanted this to be an either/or choice, so if you choose sitemesh3 it will function like prior milestones did.  If you choose sitemesh2, the known issues will be fixed and the old behavior is restored.

Notes on this work:

### Decoupling
I've decoupled sitemesh2 completely from gsp with these changes.  Either `grails-sitemesh3` or `grails-layout` must be specified if using gsp and wanting the full functionality.  

### Forge
I've enhanced forge so that if you add the Sitemesh3 feature, it will use that instead of the default (which is grails-layout).

### Documentation
I've updated the upgrading guide to clarify this requirement to make sure people are aware.

### Testing
My real issue was what to do about testing since there are several known issues with the sitemesh 3 plugin.  We already have a testing app dedicated to sitemesh3 that's split off with `@PendingFeature`, but we know there are several other pieces of functionality broken with sitemesh 3.  For this reason, I added `SITEMESH3_TESTING_ENABLED` - if this env is set, it will run with sitemesh3 plugin instead of the default grails-layout.  This allows us to test sitemesh3 easily and see where it stands on becoming the default.  In addition to the `@PendingFeature` failures, [this build scan](https://ge.grails.org/s/gofohlggiqra6) is what will fail if using sitemesh3 only.

I've also not ported the applyLayout tests since there isn't a functionality equivalent in sitemesh3.  If/when we decide to remove sitemesh2, we'll need to make sure these tests are ported.

### Spring Boot
I reverted the changes to Spring Boot app, so it's assumed that if you're using spring boot and grails, you're using sitemesh 3.  I don't think the demand is too high for sitemesh2 + Spring Boot, so this seems like a reasonable compromise - otherwise we would have a lot more work to do and have to create 2 separate projects.  Also, we are not publishing this application so it's only an example.

jdaugherty:
@matrei I've renamed the plugin from grailsLayout to layout

codeconsole:
> From my research in Grails 6, to use sitemesh3, the grails-web-sitemesh library was still included in the application

Minor clarification: grails-web-sitemesh was [manually excluded](https://github.com/codeconsole/grails-plugin-sitemesh3/tree/6677be98c203ed689050eb0cb11e266816ad358a) in the build.gradle in Grails 6.2.x
There were only a few classes that were needed from that package and they were bundled in the plugin as [empty classes](https://github.com/codeconsole/grails-plugin-sitemesh3/tree/6677be98c203ed689050eb0cb11e266816ad358a/src/main/groovy/org/grails/web/sitemesh) with the same package structure. 

```gradle
configurations {
    all {
        // ... existing configurations
        exclude group:'org.grails.plugins', module:'sitemesh2'
        exclude group:'org.grails', module:'grails-web-sitemesh'
    }
}

dependencies {
    implementation("org.sitemesh:grails-plugin-sitemesh3:6.1.2")
    // ... existing dependencies
}
```

(excellent work btw... 😄)

grails-gsp/grails-layout/build.gradle
```diff
@@ -35,16 +35,27 @@ dependencies {
...
+    implementation project(':grails-core') // for plugin & groovy
     implementation project(':grails-gsp-core') // for grails layout preprocessor
-    api project(':grails-web-taglib')
+
     runtimeOnly project(':grails-web-jsp')
 
-    api project(':grails-web-layout')
 
+    testImplementation 'jakarta.servlet:jakarta.servlet-api'
+    testImplementation 'org.spockframework:spock-core'
```
sbglasius, line 53: `spock-core` mentioned twice in a row

grails-gsp/grails-layout/build.gradle
```diff
@@ -35,16 +35,27 @@ dependencies {
...
     testImplementation 'org.spockframework:spock-core'
+    testImplementation 'org.springframework:spring-test', {
+        // impl: MockHttpServletRequest
+    }
```
sbglasius, line 56: Closure not needed, if content is commented out.

grails-gsp/grails-layout/src/main/groovy/org/apache/grails/web/layout/GrailsContentBufferingResponse.java
```diff
@@ -33,7 +34,7 @@
 import com.opensymphony.sitemesh.ContentProcessor;
 import com.opensymphony.sitemesh.webapp.SiteMeshWebAppContext;
 
-public class GrailsContentBufferingResponse extends HttpServletResponseWrapper implements GrailsResponseMutator {
+public class GrailsContentBufferingResponse extends HttpServletResponseWrapper implements GrailsResponseMutator, StreamCharBuffer.LazyInitializingWriter {
```
sbglasius, line 37: Missing space before `LazyInitializingWriter`

grails-test-examples/gsp-layout/src/test/groovy/org/apache/grails/views/gsp/layout/FullGrailsLayoutLifeCycleTests.groovy
```diff
@@ -68,7 +65,7 @@ class FullGrailsLayoutLifeCycleTests extends AbstractGrailsTagTests {
 '''
         def result = applyLayout(layout, template)
 
-        assertEquals '''
+        org.junit.jupiter.api.Assertions.assertEquals '''
```
sbglasius: Why  not a static import, or at least an import?

sbglasius (approved):
I assume the many removed tests from sitemesh3 is on purpose.

So as you can see, just being nitpicky. I approve!

grails-gsp/grails-layout/build.gradle
matrei: > Closure not needed, if content is commented out.

@sbglasius This is by (my) design to not get comment drift and allow for multiple comment lines while still keeping it clear to which dependency the comments belong. In my mind, as the closure is a no-op, it does not hurt to use it like this.

jdaugherty: Going to leave this as @matrei suggested

grails-gsp/grails-layout/src/main/groovy/org/apache/grails/web/layout/GrailsContentBufferingResponse.java
jdaugherty: LazyInitializingWriter is a subclass of StreamCharBuffer, are you suggesting of adding a space after the period?  `StreamCharBuffer.LazyInitializingWriter` vs `StreamCharBuffer.  LazyInitializingWriter`

grails-test-examples/gsp-layout/src/test/groovy/org/apache/grails/views/gsp/layout/FullGrailsLayoutLifeCycleTests.groovy
jdaugherty: This is an intellij fail, will fix this.

jdaugherty:
> I assume the many removed tests from sitemesh3 is on purpose.

As part of the decoupling, since either sitemesh library can be used with gsp, I had to pull up several tests.  We have grails-layout & grails-sitemesh3 test projects and I've moved or duplicated the tests between those projects where appropriate.  Overall, there shouldn't be a removal of tests.

jamesfredley (approved):

PR #14895, opened 2025-07-11
Fix build performance issues

jprinet:
### Issue
The build performances are currently not optimal and could be improved.
Focusing on build cacheability, running twice the `build` task without any change and cache enabled on a fresh Gradle user home surfaces some performance issues, which can result in up to 26mn of CPU time wasted on a build.

### Experiment
I used an automated experiment to run the build twice ([GitHub workflow](https://github.com/gradle/develocity-oss-projects/actions/runs/16201427211/workflow))

The summary of the experiment can be checked [there](https://github.com/gradle/develocity-oss-projects/actions/runs/16201427211)

We can in particular notice [the second build](https://ge.solutions-team.gradle.com/s/6hnubh6rve352/timeline?cacheability=cacheable&outcome=success) from experiment 3 suffering from 171 unexpected cache misses, totalling almost 26mn of serial time.

Screenshot 2025-07-11 at 3 46 45 PM

### Fixes
Several fixes were applied and are isolated on specific commits for review convenience. Impacted tasks per commit are also provided in each subsection.

[Here](https://ge.solutions-team.gradle.com/s/yotz37bxxf3v4/performance/build-cache#summary-local) is a build with the fixes applied, demonstrating 100% cache hits

#### Apply `RelativePath` normalization on task inputs to improve cacheability 
- `aggregateDataMappingGroovydoc`
- `aggregateGroovydoc`
- `asciidoctor`
- `compileGsonViews`
- `compileProfile`

#### Ignore `grails.build.info` with runtime classpath normalization
- `compileIntegrationTestGroovy`
- `compileTestGroovy`
- `integrationTest`
- `test`

#### Improve Javadoc cacheability by using date instead of instant on options.bottom task input
- `javadoc`

#### Fix AsciiDoc task cache relocatability by using relative paths instead of absolute paths
- `asciidoc`

#### Fix PublishGuideTask task cache relocatability by using relative paths instead of absolute paths
- 'publishGuide'

#### Fix GroovyPageForkCompileTask cacheability by defining a task output
- `compileGroovyPages`


jdaugherty:
@jprinet I updated this branch - we had some known failures that have been fixed since you did this work.

jprinet:
> @jprinet I updated this branch - we had some known failures that have been fixed since you did this work.

Alright, thanks for the heads up. I am happy to trim what is already addressed 👍

jprinet:
Please note it's still a draft @jdaugherty as I need to confirm the fix for `GroovyPageForkCompileTask`

jdaugherty:
@jprinet  FYI: There's a known bug with the asset compiler (https://github.com/wondrify/asset-pipeline/issues/177) around concurrent modifications.  @davydotcom I think is taking a look at this, but I don't think it will be any time soon.  That's the reason we had to serialize on a common output if that task runs in parallel.  This should explain the concurrent modification exceptions.

Concerning the build info file, I only recently discovered that the metadata loading in grails is broken & discovered this file.  We need to apply the same fix to it we did for other property files when the SOURCE_EPOCH_DATE variable is set.  Instead of ignoring, I'd rather fix that (will do that today).

jdaugherty:
We appreciate help like this, we've come so far on the build that it is very welcome.  Thank you for helping!

grails-gradle/docs-core/src/main/groovy/grails/doc/gradle/PublishGuideTask.groovy
```diff
@@ -94,6 +104,14 @@ class PublishGuideTask extends DefaultTask {
         group = 'documentation'
     }
 
+    @Optional
+    @Input
+    Map<String, String> getRelativizedPropertiesWithFilePaths() {
+        return propertiesWithFilePaths.get().collectEntries { key, file ->
+            return [key, rootProjectDir.get().getAsFile().toPath().relativize(file.toPath()).toString()]
```
jprinet, line 111: This is not CC compatible and need to be reworked

build.gradle
```diff
@@ -86,6 +88,12 @@ subprojects {
     }
 
     apply from: rootProject.layout.projectDirectory.file('gradle/dependency-licenses.gradle')
+
+    normalization {
+        runtimeClasspath {
+            ignore '**/grails.build.info'
```
jdaugherty: so some other files that get output as part of the build process that constantly change: 

META-INF/grails.factories <- created as part of the groovy compile step via an AST transform
gsp/views.properties <- created as part of the gsp compiler
${output.resourcesDir}/META-INF/grails.build.info <- created by the task in GrailsGradlePlugin's 

I'm guessing only the build properties task is causing an issue though.  I've now made all of these consistent when the build is set to build reproducible

gradle/functional-test-config.gradle
```diff
@@ -57,7 +57,7 @@ configurations.configureEach {
 // work around for parallel builds due to https://github.com/wondrify/asset-pipeline/issues/177
 if ('assetCompile' in tasks.names) {
     tasks.named('assetCompile').configure { Task task ->
-        task.outputs.dir rootProject.layout.buildDirectory.dir('asset-serialize')
+        task.outputs.dir rootProject.layout.buildDirectory.dir("asset-serialize/${project.name}")
```
jdaugherty: I believe this needs reverted until https://github.com/wondrify/asset-pipeline/issues/177 is fixed.  This code was here as a workaround for that issue.

grails-gradle/docs-core/src/main/groovy/grails/doc/gradle/PublishGuideTask.groovy
jdaugherty: Can you help me understand what about it isn't?  Would a ConfigurableFileCollection fix this?

gradle/java-config.gradle
```diff
@@ -27,7 +27,7 @@ extensions.configure(JavaPluginExtension) {
 
 tasks.withType(Javadoc).configureEach { Javadoc it ->
     it.options.noTimestamp true // prevent the file header with the date
-    it.options.bottom "Generated ${formattedBuildDate} (UTC)"
+    it.options.bottom "Generated ${formattedBuildDateForCache} (UTC)"
```
jdaugherty: I think this will break our build reproducibility tests since it's no longer based on the SOURCE_DATE_EPOCH env.  How about we change the original formattedBuildDate to be date based instead: 

formattedBuildDate = System.getenv("SOURCE_DATE_EPOCH") ?
            DateTimeFormatter.ISO_INSTANT.format(buildInstant) :
            DateTimeFormatter.ISO_DATE.format(LocalDate.ofInstant(buildInstant as Instant, ZoneOffset.UTC))

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/json/GsonViewCompilerTask.groovy
```diff
@@ -36,6 +39,12 @@ import javax.inject.Inject
 @CacheableTask
 class GsonViewCompilerTask extends AbstractGroovyTemplateCompileTask {
 
+    @PathSensitive(PathSensitivity.RELATIVE)
```
jdaugherty: I'd suggest moving this up to the abstract class since the MarkupViewCompilerTask will have the same issue.

grails-gradle/docs-core/src/main/groovy/grails/doc/gradle/PublishGuideTask.groovy
jprinet: Please ignore my comment, I got confused by the CC report
```
Task `:grails-doc:publishGuide` of type `grails.doc.gradle.PublishGuideTask`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject'
```

I thought my change introduced the problem but I'm only using `DirectoryProperty rootProjectDir` which is just fine

gradle/functional-test-config.gradle
jprinet: Reverted with ca20c96beb2e7b1a605310ca45fce361acc258dd

build.gradle
jprinet: I am happy to revert this change as you will fix it when build is set to reproducible.
Sorry for the naive question, but is there a reason not to have a reproducible build always?

gradle/java-config.gradle
jprinet: applied on 85246f3a4d8e6543cccfccf18b3de531cbea4dbf

build.gradle
jprinet: Reverted `grails.build.info` normalization with a4afe97edef98d669f41b270f943c214a0101741

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/json/GsonViewCompilerTask.groovy
jprinet: Done with 7bcfc215154363f1ea1a7908ffe4979e97863c2b

jprinet:
It is now ready for review @jdaugherty

build.gradle
jdaugherty: In this case, I am contemplating ditching the date and adding just the version.  For now, the change you made is perfect!

For the build always being reproducible... setting SOURCE_DATE_EPOCH will make it always reproducible... the issue for us is there are multiple entry points to the GSP compiler.  Gradle is one of those entrypoints, but a running application in dev mode is also one.  Historically the caching was done in the compiler itself by setting the date of the compile on the generated class.  Setting that env variable tells the compiler do not compile files unless they have changed after that date - which could be problematic if that date is set.  I think we can get to the point where reproducible is not opt
in, but for now we have other tasks to solve.

jprinet: Makes sense, thanks for the explanation 👍

jdaugherty:
> It is now ready for review @jdaugherty 

@jprinet  
Thank you so much for this contribution.  I have to board a plane here shortly, so I can't do further testing locally yet.  I will finish this review later tonight after I can be at my computer.

jdaugherty:
@jprinet I think you've convinced me the property files should always be reproducible.  I will open a ticket and fix this in the coming week.  As for these contributions, thank you again so much.  They seem to work great locally and once the property files are always reproducible, I think it will help even more per your observations.

jprinet:
> @jprinet I think you've convinced me the property files should always be reproducible. I will open a ticket and fix this in the coming week. As for these contributions, thank you again so much. They seem to work great locally and once the property files are always reproducible, I think it will help even more per your observations.

It should be a great move for build performances 👍
Thanks again @jdaugherty for the assistance to get this one merged ❤️

PR #14896, opened 2025-07-11
Wrapper Improvements found by M5 Release

jdaugherty:
fix: improve the grails wrapper logic to better handle versions based on the allowed types that were added for the home directory management


jdaugherty:
@jamesfredley @matrei  this is breaking the current publish process, so I intend to go ahead and merge.  If we need to do further changes, please let me know here and I'll open a subsequent PR

PR #14902, opened 2025-07-13
Update GitHub repository homepage URL

jamesfredley:
Setting copied from https://github.com/apache/grails-static-website/blob/master/.asf.yaml

image

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -39,6 +39,8 @@ github:
         - id: jdaugherty
           type: User
       wait_timer: 0
+  description: "The Grails Web Application Framework"
+  homepage: https://grails.apache.org/
 notifications:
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -39,6 +39,8 @@ github:
         - id: jdaugherty
           type: User
       wait_timer: 0
+  description: "The Grails Web Application Framework"
```
jdaugherty: Isn't this better said as 'Grails - the Web Application Framework'

jdaugherty (approved):

.asf.yaml
jamesfredley: Updated.  It is stronger from a marketing standpoint.

PR #14908, opened 2025-07-16
Use version variables + minor copy and formatting adjustments

jamesfredley:


jdaugherty (commented):
Why use the version? for the final release we will have to set these anyhow... since the first release after it, it will change.  The milestone is supposed
to be a preview, so leaving 7.0.0 is reasonable.

jamesfredley:
@jdaugherty I have changed a few Grails {version} locations to Grails 7, where the context of the paragraph is fixed for all of Grails 7.0.x versions including -SNAPSHOT's.  I left {version} in a few locations where the docs should display the full grails version when published.

Since the docs are fixed when published, except for -SNAPSHOT, I think this is a good balance.

The other versions advance as we release, so I think variables are appropriate there, so we do not have to remember to make the change.

This PR now also includes some formatting and numbering fixes and relocates the Geb testcontainers subsections which had drifted to another section.

jdaugherty (approved):

PR #14929, opened 2025-07-23
Enhance CreateReleaseDropDownTask so it can be reused by grails-spring-security

jamesfredley:


grails-gradle/docs-core/src/main/groovy/grails/doc/dropdown/CreateReleaseDropDownTask.groovy
```diff
@@ -80,6 +87,9 @@ abstract class CreateReleaseDropDownTask extends DefaultTask {
         modifiedPagesDirectory = objects.directoryProperty().convention(project.layout.buildDirectory.dir("modified-pages"))
         gitTags = objects.fileProperty().convention(project.layout.buildDirectory.file('git-tags.txt'))
         minimumVersion = objects.property(SoftwareVersion).convention(new SoftwareVersion(major: 5))
+        docBaseUrl = objects.property(String).convention("https://docs.grails.org")
+        versionHtml = objects.property(String).convention("<p><strong>Version:</strong> " + projectVersion.get() + "</p>")
```
jdaugherty: The convention for this needs to be a provider so this can be evaluated only if the task executes.  Like this:

       .convention(project.provider{ "<p><strong>Version:</strong> " + projectVersion.get() + "</p>" })

jdaugherty (approved):
Other than the convention fix, this LGTM

grails-gradle/docs-core/src/main/groovy/grails/doc/dropdown/CreateReleaseDropDownTask.groovy
jamesfredley: Updated and retested with grails-doc:build on grails-core and aggregateDocs on grails-spring-security

PR #14936, opened 2025-07-26
Fix issue reported by GitHub CodeQL 

matrei:
Error from CodeQL:
Resolving XML external entity in user-controlled data
XML parsing depends on a without guarding against external entity expansion.


jdaugherty:
I saw this change suggestion, have you tested it through?  We still don't have end-to-end wrapper tests yet and as long as this works, I'm ok with merging this.

matrei:
> I saw this change suggestion, have you tested it through?  We still don't have end-to-end wrapper tests yet and as long as this works, I'm ok with merging this.

Yes, I have tested successfully parsing the maven xml files in question with these parser settings.

jamesfredley (approved):

jdaugherty (approved):

PR #14942, opened 2025-08-01
Update base, web and rest-api profiles to more closely match Forge

jamesfredley:
For web and rest-api profiles, plus necessary adjustments to base, this PR adjusts the dependencies to match the default dependencies generated by Forge.  There will likely be some debate regarding which is correct, but this is a starting place.

We still need to find a better solution for `integrationTestImplementation testFixtures("org.apache.grails:grails-geb")` which lives in base/skeleton/build.gradle and is currently applied for all profiles, since the yml conversion does not support `testFixtures(""`).  https://github.com/apache/grails-core/issues/14943 

plugin and web-plugin have significant differences to discuss before making changes:  https://github.com/apache/grails-core/issues/14944


grails-profiles/base/profile.yml
```diff
@@ -27,15 +27,10 @@ build:
         - eclipse
         - idea
         - org.apache.grails.gradle.grails-app
+        - groovy
```
jdaugherty: the grails-app plugin extends groovy.  I'm not sure of the benefit of adding that here.  I'd argue it shouldn't even exist in forge if it's there.

grails-profiles/rest-api/profile.yml
```diff
@@ -16,7 +16,6 @@ description: Profile for REST API applications
 features:
     defaults: 
         - hibernate5
-        - events
```
jdaugherty, line 19: event isn't included in forge?  they were split from hibernate5, so i'd have expected them to be included.  I'm not sure hibernate5 will work correctly without it.

grails-profiles/base/profile.yml
jamesfredley: Let's remove it from both, I'll update the PR.

grails-profiles/rest-api/profile.yml
jamesfredley: That is correct and was also the same with forge for Grails 6.

jamesfredley: looking at the hibernate5 projects the only event code I found was from https://github.com/apache/grails-core/tree/7.0.x/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/events

jamesfredley: Adding grails-events as a default dependency for all profiles and forge application types.

grails-profiles/base/profile.yml
```diff
@@ -19,6 +19,9 @@ skeleton:
 repositories:
     - "https://repository.apache.org/content/groups/public"
```
jdaugherty: Should we reorder these?  Maybe not even include RAO after infra blocking?

jdaugherty (approved):

grails-profiles/base/profile.yml
jamesfredley: I removed it from profile.yml and put it last in https://github.com/apache/grails-core/pull/14942/files#diff-1ef480bb951f8e0ea48fe8296d704c3f601fc15f790a86e84843875b95f5a9cb so it would exactly match the way we handled it in forge.

PR #14946, opened 2025-08-01
#14930 - make views reproducible by sorting views always

### LINKED ISSUE #14930: views.properties is not reproducible
### Issue description

The GSP compiler can produce a random order for views.properties.  This ticket is to ensure that file is sorted.

jdaugherty:
Fixes https://github.com/apache/grails-core/issues/14930


jamesfredley (approved):

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
```diff
@@ -148,7 +151,7 @@ class GroovyPageCompiler {
                 // write the view registry to a properties file (this is read by GroovyPagesTemplateEngine at runtime)
                 File viewregistryFile = new File(targetDir, "gsp/views.properties")
                 viewregistryFile.parentFile.mkdirs()
-                Properties views = new Properties()
+                Properties views = CollectionFactory.createSortedProperties(false)
```
matrei, line 154: Add comment why we need to use `SortedProperties`?

matrei (approved):

jamesfredley (approved):

Issue #13851, opened 2024-11-18
Move GrailsMessageSourceUtils.groovy to Grails Common


jamesfredley:
### Issue description

This will allow the removal of grails-core dependency from gorm-mongodb

https://github.com/grails/grails-core/blob/2a0b7d21305fb05fce181a91a83e99da55470b2b/grails-core/src/main/groovy/grails/util/GrailsMessageSourceUtils.groovy#L20

https://github.com/search?q=repo%3Agrails%2Fgorm-mongodb%20grails-core&type=code

jamesfredley:
Will be completed after https://github.com/grails/grails-core/issues/13850

PR #14952, opened 2025-08-04
fix: #14951 - make groovy extension merge for shadowJar reproducible

### LINKED ISSUE #14951: grails-cli shadowJar reproducibility
### Issue description

The grails-cli shadowJar contains both a grails-plugin.xml.asc & property files with differing dates due to the addition of merging groovy extension as of 7.0.0-RC1.  This ticket is to address these reproducibility issues.

jdaugherty:
Fixes https://github.com/apache/grails-core/issues/14951


grails-common/build.gradle
```diff
@@ -27,6 +27,8 @@ group = 'org.apache.grails.common'
 dependencies {
     implementation platform(project(':grails-bom'))
 
+    api 'org.apache.grails.gradle.common:grails-gradle-common'
```
matrei: I'm still looking at this PR (I'm traveling), but to me, it feels instinctively wrong that `grails-common` should depend on `grails-gradle-common`.

jdaugherty: We use the property utilities in both the AST & in Gradle.  Before, they were copy pasted.  I needed to put these utilities somewhere so they could be used by both & not copied.  The purpose of common was to include base libraries / utility classes.  I opt'd to split into a common & gradle-common because we needed to pull up the property utility.  Do you have another suggestion?

jdaugherty: Is it the name that's the problem?  Would grails-compile-common or grails-build-common be better names?  The grails-gradle-common project doesn't actually have any gradle dependencies, only a compileOnly groovy dependency.

matrei: Could we extract `grails-common` to an isolated build and use `includeBuild` where it is used?
That way, we would not create any dependencies between the `grails-gradle` project and the `grails-core` projects.
The `grails-common` project would then be at the top of the hierarchy and can be used anywhere without creating any other couplings.

jdaugherty: The issue is grails-common includes spring boot and we do not want that on the build classpath.  

I looked at removing that but when we move the message utilities we will still need that dependency

jdaugherty: If the concern is the naming, we can also name the projects: 

grails-common <- one included by both build & AST
grails-spring-common <- one included by grails-core / gorm

matrei: > The issue is grails-common includes spring boot and we do not want that on the build classpath.

Where does it include `spring-boot`?

jdaugherty: Sorry, spring core: 

    // the annotation metadata readers are used by both GORM & grails-core so it can be placed in a shared build
    implementation 'org.springframework:spring-core'

jdaugherty: The comment is out of date, but the messaging utilities will need it and common is what provides the share base

matrei: I'm not following. As the codebase looks right now, we can remove the dependency on `spring-core`. It's not used.
Which messaging utilities will be moved to `grails-common`?

jdaugherty: We still need a common inclusion point for spring and any utility classes that rely on Spring.

* https://github.com/apache/grails-core/issues/13851 is an example that will still cause us to deal with this problem.
* if we have to restore the annotation logic, it will go in the current grails-common

To be clear, I am proposing:
1. rename grails-common to grails-spring-common
2. rename grails-gradle-common to grails-common

since the objection seems to be based
on the name.  I also agree having gradle in the name could lead to someone including gradle dependencies - which would be bad because they would end up on application classpaths.

Combining these libraries will also cause a breaking change if we have to undo this due to needing to add back the annotation readers etc.  I am very against that because it could mean having to do another release candidate and delay 7.  I would rather we split them and recombine in 8 if we determine it's not needed.

matrei: Out of curiosity, does the `# SOURCE_DATE_EPOCH = 123456` comment added to the property files by `PropertyFileUtils.makePropertiesFileReproducible()` (and replacing the date added by `Properties.store()`) have any function or would it suffice to just remove the date comment (or that it was not written in the first place)?

jdaugherty: Since gradle uses that file to know when to recompile, we want that to change when the source date epoch does.  Otherwise it may not trigger a recompile.  So yes, it's needed.

matrei: > The issue is grails-common includes spring boot and we do not want that on the build classpath.

Is there also and issue with having `spring-core` as a dependency in `grails-gradle` projects?

jdaugherty: Yes, we don't want spring on the gradle build path.  It pulls in more dependencies than it needs to and will likely cause plugins to fail  - see teh license plugin as an example.  The goal is to get it off the path longer term.

matrei: > Since gradle uses that file to know when to recompile, we want that to change when the source date epoch does. Otherwise it may not trigger a recompile. So yes, it's needed.

Thanks, that is good info. I would like to understand this better.
Could you help me understand where/how Gradle uses one of the properties files updated by `FilePropertiesUtils` to know that it needs to trigger a recompile, that otherwise would not happen?

jdaugherty: The file is an input. The date in the file is what causes the file to change and when the input changes it triggers the recompile

grails-gradle/common/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+    // see: https://docs.gradle.org/current/userguide/compatibility.html#groovy
+    compileOnly 'org.codehaus.groovy:groovy'
+
+    // Testing
+    testImplementation platform(project(':grails-gradle-bom'))
+    testImplementation 'org.slf4j:slf4j-simple'
+    testImplementation('org.spockframework:spock-core') { transitive = false }
+    testImplementation 'org.codehaus.groovy:groovy-test-junit5'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.junit.platform:junit-platform-runner'
```
matrei: Do we need these?
```suggestion
```

grails-gradle/common/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+
+    // for easier setting of environment variables in tests
+    testImplementation 'uk.org.webcompere:system-stubs-core:2.1.8'
```
matrei: Group by configuration scope?
```suggestion
    // for easier setting of environment variables in tests
    testImplementation 'uk.org.webcompere:system-stubs-core:2.1.8'

    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
```

grails-gradle/common/src/main/groovy/org/apache/grails/gradle/common/PropertyFileUtils.groovy
```diff
@@ -0,0 +1,80 @@
...
+import org.codehaus.groovy.runtime.IOGroovyMethods
+
+import java.nio.charset.StandardCharsets
+import java.time.LocalDate
+import java.time.ZoneOffset
+import java.util.regex.Pattern
+
+@CompileStatic
+final class PropertyFileUtils {
+    private final static Pattern TIME_REGEX = Pattern.compile('^#(?:Sun|Mon|Tue|Wed|Thu|Fri|Sat)(?:,|\\s).*$')
```
matrei: Use Groovy pattern syntax?
```suggestion
    private final static Pattern TIME_REGEX = ~'^#(?:Sun|Mon|Tue|Wed|Thu|Fri|Sat)(?:,|\\s).*$'
```

grails-gradle/common/src/main/groovy/org/apache/grails/gradle/common/PropertyFileUtils.groovy
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private static void makePropertyLinesReproducible(List<String> lines, BufferedWriter writer) {
+        String sourceDateEpoch = determineSourceDateEpoch()
+
+        boolean dateReplaced = false
+        lines.each { String line ->
+            if (!dateReplaced && TIME_REGEX.matcher(line).matches()) {
+                dateReplaced = true
+                IOGroovyMethods.writeLine(writer, "# SOURCE_DATE_EPOCH = ${sourceDateEpoch}" as String)
```
matrei:
```suggestion
                IOGroovyMethods.writeLine(writer, "# SOURCE_DATE_EPOCH = $sourceDateEpoch")
```

grails-gradle/common/src/main/groovy/org/apache/grails/gradle/common/PropertyFileUtils.groovy
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private static String determineSourceDateEpoch() {
+        String sourceDateEpoch = System.getenv('SOURCE_DATE_EPOCH')
+        if (!sourceDateEpoch) {
+            sourceDateEpoch = LocalDate.now(ZoneOffset.UTC)
+                    .atStartOfDay(ZoneOffset.UTC)
+                    .toEpochSecond().toString()
+        }
+        sourceDateEpoch
```
matrei:
```suggestion
        System.getenv('SOURCE_DATE_EPOCH') ?:
                LocalDate.now(ZoneOffset.UTC)
                        .atStartOfDay(ZoneOffset.UTC)
                        .toEpochSecond().toString()

```

grails-forge/buildSrc/src/main/groovy/org/apache/grails/forge/buildlogic/shadowjar/GrailsGroovyExtensionTransformer.groovy
```diff
@@ -0,0 +1,112 @@
...
+import org.gradle.api.file.FileTreeElement
+
+/**
+ * Based on com.github.jengelman.gradle.plugins.shadow.transformers.GroovyExtensionModuleTransformer but with
+ * reproducible property file support
+ */
+@CompileStatic
+class GrailsGroovyExtensionTransformer implements Transformer {
+
+    private static final String GROOVY_EXTENSION_MODULE_DESCRIPTOR_PATH = "META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule"
```
matrei:
```suggestion
    private static final String GROOVY_EXTENSION_MODULE_DESCRIPTOR_PATH = 'META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule'
```

matrei (approved):

PR #14957, opened 2025-08-06
Add build test for grails-micronaut feature, consolidate tests

jamesfredley:


jamesfredley:
This test requires running bootRun, but we do not currently have a path to do this in these tests

PR #14985, opened 2025-08-18
Make reloading feature switch between options

jamesfredley:
Resolves:  Adding feature JRebel gives error: There can only be one of the following features selected: [spring-boot-devtools, jrebel]


jdaugherty (approved):

matrei (approved):
👍

Issue #14978, opened 2025-08-16
grails-gradle API Docs


jamesfredley:
### Issue description

Prior to consolidation api docs were located: https://apache.github.io/grails-gradle-plugin/latest/api/ & https://github.com/apache/grails-doc/tree/gh-pages/latest/api

Locations to Update with new URL:  

- https://github.com/apache/grails-core/tree/7.0.x/grails-gradle/plugins#readme

### Errors in .\guide\conf.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/Environment.html> | Cannot find file

### Errors in .\guide\plugins.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/BuildScope.html> | Cannot find file

### Errors in .\guide\single.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/BuildScope.html> | Cannot find file
* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/Environment.html> | Cannot find file

PR #14988, opened 2025-08-18
Cleanup broken documentation links, update links which have changed, standardize link syntax

jamesfredley:
This started by evaluating broken links in grails-doc, but expanded from there to correct the same URLs across the project.  It updates grails.org URLs to grails.apache.org and updates all links, which can be changed, to https.  

In some cases the best available URL was used to replace one that returned a 404.  In the rare case where there was no URL available, like in the instance of a 15 year old blog post, the URL was removed.

It now represents around 1,000 fixed or updated URLs.

There are a few `grails.util.*` api links which still 404, due to https://github.com/apache/grails-core/issues/14978

Broken links were identified and verified as fixed using https://github.com/lycheeverse/lychee/releases/tag/lychee-v0.19.1.  

grails-data and groovydocs were skipped and issues will be created to work on those broken URLs.  https://github.com/apache/grails-core/issues/14990 & https://github.com/apache/grails-core/issues/14991

```
./gradlew grails-doc:build
cd \grails-doc\build\docs
lychee-x86_64-windows.exe . --format markdown --output results.md --exclude-path api --exclude-path grails-data --exclude-path grails-data/hibernate5/api --exclude ".*.css$" --exclude "^http://localhost" --exclude-path guide/pages
```

review `results.md`

Here is the summary from the latest run.  https://dev.mysql.com/downloads/connector/j/ are false positives.  

# Summary

| Status        | Count |
|---------------|-------|
| 🔍 Total      | 97352 |
| ✅ Successful | 96709 |
| ⏳ Timeouts   | 0     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 637   |
| ❓ Unknown    | 0     |
| 🚫 Errors     | 6     |

```
## Errors per input

### Errors in .\guide\conf.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/Environment.html> | Cannot find file
* [403] <https://dev.mysql.com/downloads/connector/j/> | Rejected status code (this depends on your "accept" configuration): Forbidden

### Errors in .\guide\plugins.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/BuildScope.html> | Cannot find file

### Errors in .\guide\single.html

* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/BuildScope.html> | Cannot find file
* [ERROR] </grails-core/grails-doc/build/docs/api/grails/util/Environment.html> | Cannot find file
* [403] <https://dev.mysql.com/downloads/connector/j/> | Error (cached)
```


LICENSE
```diff
@@ -1,7 +1,7 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
```
jdaugherty: I tried to make this change previously and received push back because its text does not match the link text.

LICENSE
```diff
@@ -236,7 +236,7 @@ See licenses/LICENSE-CDDL.txt for the full license terms.
 
 --------------------------------------------------------------------------
 
-This product includes software developed by the OpenSymphony Group (http://www.opensymphony.com/). It uses Sitemesh2,
+This product includes software developed by the OpenSymphony Group (https://github.com/sitemesh). It uses Sitemesh2,
```
jdaugherty: This has to remain this way, the license requires this text

NOTICE
```diff
@@ -21,5 +21,5 @@ The original XML Schemas for Java EE Deployment Descriptors:
 may be obtained from:
 http://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/index.html
 
-This product includes software developed by the OpenSymphony Group (http://www.opensymphony.com/). It uses Sitemesh2,
+This product includes software developed by the OpenSymphony Group (https://github.com/sitemesh). It uses Sitemesh2,
```
jdaugherty: This has to remain this way because the license requires this text

gradle.properties
```diff
@@ -5,7 +5,7 @@
 # (the "License"); you may not use this file except in compliance with
 # the License.  You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#     https://www.apache.org/licenses/LICENSE-2.0
```
jdaugherty, line 8: This needs to match the official license text so it must remain

grails-doc/src/en/guide/async/events/asyncGormEvents.adoc
```diff
@@ -17,7 +17,7 @@ specific language governing permissions and limitations
 under the License.
 ////
 
-GORM defines a link:{gormApiDocs}/org/grails/datastore/mapping/engine/event/package-summary.html[number of useful events] that you can listen for.
```
jdaugherty, line 20: So we still publish both a gorm specific api and then all of the grails api.  Shouldn't we refer to the specific one? If not can we stop publishing it since its a duplicate?

grails-doc/src/en/guide/async/rxjava/asyncRxGorm.adoc
```diff
@@ -17,7 +17,7 @@ specific language governing permissions and limitations
 under the License.
 ////
 
-https://gorm.grails.org/{projectVersion}/rx/manual/[RxGORM] is new implementation of GORM that has the following goals:
+https://gorm.grails.org/latest/rx/manual/[RxGORM] is new implementation of GORM that has the following goals:
```
jdaugherty: We need to extract this into a variable.  If we ever start publishing those docs again, this will be wrong.

jdaugherty (changes requested):
Great Job!

My only concern is:

The license and notice files have to match exactly since it is spelled out that way in the licenses.  We can change headers, etc, but I don't believe we can change the LICENSE and NOTICE files

LICENSE
jamesfredley: I see, I'll put it back to match https://www.apache.org/licenses/LICENSE-2.0.txt exactly

gradle.properties
jamesfredley: Most of our headers use https:// - https://github.com/search?q=repo%3Aapache%2Fgrails-core%20https%3A%2F%2Fwww.apache.org%2Flicenses%2FLICENSE-2.0&type=code

LICENSE
jamesfredley: Changing these back, it's odd to link to a dead address and I left it in the open symphony license file, but after reading it is explicit on URL.

grails-doc/src/en/guide/async/events/asyncGormEvents.adoc
jamesfredley: Yes, /api/ has everything except for grails-gradle and grails-forge.  I leaned toward using the one with everything vs the specific one.  Removing the one from grails-data will have to be done as part of https://github.com/apache/grails-core/issues/14990

grails-doc/src/en/guide/async/rxjava/asyncRxGorm.adoc
jamesfredley: Updated

gradle.properties
jdaugherty: Headers are ok, I missed this and thought I was looking at the license file.  Thank you!

grails-doc/src/en/guide/async/events/asyncGormEvents.adoc
jdaugherty: The only downside is our packages are not segmented by project so it could be confusing using a unified groovydoc

jamesfredley: The following exist in the combined /api/ and a second location:

grails-data/api
grails-data/hibernate5/api
grails-data/mongodb/api

everything else, which was consolidated into grails core, only exists in /api/.  The groovydoc links currently point to the groovydoc without frames, which does make them a bit less overwhelming.

jamesfredley: @jdaugherty I created an issue which includes details on duplicated groovydocs for grails-data.  Let's work through that on a separate PR, there will actually be a few more groovydoc related ones.  https://github.com/apache/grails-core/issues/14997

jdaugherty (approved):

PR #14989, opened 2025-08-18
fix: sitemesh3 rendering

jdaugherty:
When we were forced to rename sitemesh -> grailsLayout due to the license audit of sitemesh2, it broke sitemesh 3.  The tags are being generated still as `grailsLayout:` in the GrailsLayoutPreprocessor so sitemesh3's SitemeshTagLib needs to use the same namespace.


jdaugherty:
I'm epxecting all tests to pass now; here's the local build scan:  https://ge.grails.org/s/ogwckaqwujlks

jamesfredley (approved):

matrei (approved):
Nice James!
What is the purpose of the `<body onload="test();">` function call in some of the `endToEnd` views?

jdaugherty:
> Nice James!
> 
> What is the purpose of the `<body onload="test();">` function call in some of the `endToEnd` views?

Sitemesh parses the html and I think this was meant as the ultimate test to make sure it was handling complicated attributes correctly.  I didn't write these tests - they are from the original Sitemesh 2 test code.

PR #14994, opened 2025-08-20
Update Feature info and Gradle templates

matrei:


grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/H2.java
```diff
@@ -20,12 +20,18 @@
...
+import org.grails.forge.feature.DefaultFeature;
+import org.grails.forge.feature.Feature;
+import org.grails.forge.options.Options;
+
+import java.util.Set;
 
 @Singleton
 @Primary
-public class H2 extends DatabaseDriverFeature {
+public class H2 extends DatabaseDriverFeature implements DefaultFeature {
```
jdaugherty, line 34: H2 isn't required unless GORM is being used.  What's the thinking on adding this as a default feature?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/MongoGorm.java
```diff
@@ -52,7 +52,7 @@ public String getTitle() {
 
     @Override
     public String getDescription() {
-        return "Configures GORM for MongoDB for Groovy applications";
+        return "Configure GORM for using MongoDB.";
```
jdaugherty, line 55: What are your thoughts about changing GORM -> Grails Data since we've renamed the packages?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/Neo4jGorm.java
```diff
@@ -44,7 +44,7 @@ public String getTitle() {
 
     @Override
     public String getDescription() {
-        return "Configures GORM for Neo4j for Groovy applications";
+        return "Configure GORM for using Neo4j";
```
jdaugherty, line 47: Should we set a max supported version or something?  If they select Neo4j for Grails 7, it will fail ... seems like we should tell the user this isn't supported yet.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/TestContainers.java
```diff
@@ -138,6 +138,6 @@ public String getCategory() {
     @Nullable
     @Override
     public String getThirdPartyDocumentation() {
-        return "https://www.testcontainers.org/";
+        return "https://www.testcontainers.org";
```
jdaugherty: @jamesfredley changed the url to java.testcontainers.org instead of www.testcontainers.org in the docs, so should we change that here too?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grails/GrailsConsole.java
```diff
@@ -41,9 +41,14 @@ public String getName() {
...
 
+    @Override
+    public String getTitle() {
+        return "Console";
+    }
+
     @Override
     public String getDescription() {
-        return "Starts the Grails console, which is an extended version of the regular Groovy console.";
+        return "A Groovy console for interactive runtime application management and debugging.";
```
jdaugherty, line 51: I don't think this is correct.  It's a grails console and it doesn't launch the interactive window.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
```diff
@@ -45,12 +45,12 @@ public String getName() {
...
     @Override
     public String getTitle() {
-        return "Grails Wrapper for Grails";
+        return "Grails Wrapper";
     }
 
     @Override
     public String getDescription() {
-        return "Gives you the ability to run grails shell from terminal.";
+        return "Enables running the Grails CLI directly without requiring a local Grails distribution installed.";
```
jdaugherty: The wrapper is really a light weight installer for the delegating CLI.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/Scaffolding.java
```diff
@@ -51,12 +51,14 @@ public String getName() {
 
     @Override
     public String getTitle() {
-        return "Grails Scaffolding Plugin";
+        return "Scaffolding";
```
jdaugherty, line 54: We distinguish between libraries for logback, but not for plugins.  Shouldn't we be consistent?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grails/GrailsConsole.java
matrei: Skärmbild 2025-08-20 205629

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/MongoGorm.java
matrei: Sounds good! But that is a big task.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/H2.java
matrei: Good catch! I'll add another condition for when to apply it.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/Neo4jGorm.java
matrei: Yes, we should probably disable it in some way.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
matrei: What would be a better description of the Grails Wrapper feature for a new user?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/Scaffolding.java
matrei: I started drifting into thinking in "Features", not implementations.
I don't know, it's not a always a clear cut.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/MongoGorm.java
jdaugherty: We don't have to do it for this PR, let's discuss on the weekly?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grails/GrailsConsole.java
jdaugherty: I had no idea this was supported.  Thank you.  Leave it as is.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
jdaugherty: Gradle refers to their wrapper as: 

       The Wrapper is a script (called gradlew or gradlew.bat) that invokes a declared version of Gradle, downloading it beforehand if necessary. I

But ours downloads the version of grails matching the project version. 

     A tiny wrapper script that auto-downloads the Grails Delegating CLI on demand and transparently delegates all commands to it.

jdaugherty: My main concern is that people understand it's a way to download the grails delegating CLI without checking in binaries.  It's meant to save space and be able to be checked in if not using sdkman

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/Scaffolding.java
jdaugherty: Let's leave it unless @jamesfredley thinks otherwise?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
matrei: >  is that people understand it's a way to download the grails delegating CLI without checking in binaries. It's meant to save space and be able to be checked in if not using sdkman

That is not how I see the `./grailsw` command with my end user hat on.
For me as an end user, it's a convenient shortcut to run/start the Grails CLI with the right version in a Grails project, plain and simple.

jamesfredley: So it's 'what it does' vs 'how it works'.  Here is a stab at combining both ideas into a single sentence.

`A tiny wrapper script that enables running the Grails CLI directly by auto-downloading and transparently delegating commands.`

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/Scaffolding.java
jamesfredley: I do not have a strong opinion.  Reviewing the changes it appears that 'plugin' was removed from the title for official apache grails plugins.  That aligns with the maven coordinate change and I think in general makes them appear as official parts of Grails, which could be a positive.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/Neo4jGorm.java
jamesfredley: Let's do this as a separate PR.  https://github.com/apache/grails-core/issues/15008

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/EHCache.java
```diff
@@ -55,12 +56,12 @@ public void apply(GeneratorContext generatorContext) {
 
     @Override
     public String getThirdPartyDocumentation() {
-        return "https://www.ehcache.org/";
+        return "https://www.ehcache.org";
```
jamesfredley: I think we should leave trailing slashes, since without is technically missing the path portion of the URL.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/EHCache.java
```diff
@@ -55,12 +56,12 @@ public void apply(GeneratorContext generatorContext) {
...
     }
 
     @Override
     public String getDocumentation() {
-        return "https://grails-plugins.github.io/grails-cache-ehcache/latest/";
+        return "https://grails-plugins.github.io/grails-cache-ehcache/latest";
```
jamesfredley: This causes two HTTP requests

image

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/GrailsCache.java
```diff
@@ -69,7 +70,7 @@ public String getCategory() {
 
     @Override
     public String getDocumentation() {
-        return "https://docs.grails.org/";
+        return "https://docs.grails.org/" + VersionInfo.getGrailsVersion() + "/guide/cache.html";
```
jamesfredley: In the recent URL update PRs I ended up removing " + VersionInfo.getGrailsVersion() + " since this is invalid for SNAPSHOT versions.  I put /latest/ in these URLs.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/MongoSync.java
```diff
@@ -71,7 +71,7 @@ public String getCategory() {
 
     @Override
     public String getDocumentation() {
-        return "https://docs.mongodb.com/drivers/java/sync/current/";
+        return "https://docs.mongodb.com/drivers/java/sync/current";
```
jamesfredley: two http requests

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/TestContainers.java
```diff
@@ -138,6 +138,6 @@ public String getCategory() {
     @Nullable
     @Override
     public String getThirdPartyDocumentation() {
-        return "https://java.testcontainers.org/";
+        return "https://java.testcontainers.org";
```
jamesfredley: missing path

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/migration/DatabaseMigrationPlugin.java
```diff
@@ -35,21 +36,22 @@ public String getName() {
...
     @Override
     public String getDescription() {
-        return "Adds support for Liquibase database migrations. The Database Migration plugin helps you manage database changes while developing Grails applications. ";
+        return "Add support for Liquibase database migrations. The Database Migration Plugin " +
+                "helps you manage database schema changes while developing Grails applications.";
     }
 
     public String getThirdPartyDocumentation() {
-        return "https://www.liquibase.org/";
+        return "https://www.liquibase.org";
```
jamesfredley: missing path

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/other/HibernateValidator.java
```diff
@@ -63,6 +63,6 @@ public String getCategory() {
 
     @Override
     public String getDocumentation() {
-        return "https://hibernate.org/validator/";
+        return "https://hibernate.org/validator";
```
jamesfredley: two http requests

grails-forge/grails-forge-core/src/test/groovy/org/grails/forge/feature/database/MongoSyncSpec.groovy
```diff
@@ -33,7 +33,7 @@ class MongoSyncSpec extends ApplicationContextSpec implements CommandOutputFixtu
 
         then:
         readme
-        readme.contains("https://docs.mongodb.com/drivers/java/sync/current/")
+        readme.contains("https://docs.mongodb.com/drivers/java/sync/current")
```
jamesfredley: https://www.mongodb.com/docs/drivers/java/sync/current/ is the new full URL

jamesfredley (commented):
I provided feedback on the open comment discussions and added some comments on the URLs.  URLs are top of mind after https://github.com/apache/grails-core/pull/14988 and https://github.com/apache/grails-core/pull/15003

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/Neo4jGorm.java
jdaugherty: I 100% agree doing this separately, but wanted to point out we should create a ticket.  Thanks @jamesfredley

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
jdaugherty: I'm ok with @jamesfredley 's suggestion.  Are you @matrei ?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/GrailsCache.java
jdaugherty: This could cause confusion when there are multiple major versions though.  I assume the documentation is shown on the CLI somehow?  If so, I'd argue we need the version in all of these.  

We could easily publish the snapshot version in the docs too, that would fix the snapshot problem.  It might also makes it better since right now after publishing a 6.x snapshot, we would technically have to republish a 7.x.  I could build the same "latest" detection logic into the snapshot publishing that's there for release in the github action.

jamesfredley: I have not seen where the documentation is exposed on start.grails.org or the forge cli.  There is a features schema on https://grailsforge-latest-cjmq3uyfcq-uc.a.run.app/swagger-ui/index.html#/, but documentation is not returned on endpoints like https://snapshot.grails.org/application-types/web/features

If we want to keep version here we should write another method in VersionInfo that returns 'SNAPSHOT' when the version end with -SNAPSHOT to be compatible with the documentation.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grailsWrapper/GrailsWrapper.java
matrei: Perfect!

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/GrailsCache.java
matrei: I have implemented a `VersionInfo.getDocumentationVersion()` method in this PR that will take `snapshot` into account.
As @jdaugherty pointed out, in the best of worlds we would publish version specific snapshots and this method would return `6.2.4-snapshot` and `7.0.0-snapshot`, but perhaps we can ticket that if we want that functionality.

matrei:
Hmm, that's a strange test failure in the Groovy Joint Validation Build: https://github.com/apache/grails-core/actions/runs/17173238880/job/48725756373?pr=14994#step:9:11029

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/cache/GrailsCache.java
jamesfredley: Sounds great, thank you.

jamesfredley (approved):

jamesfredley:
@matrei It was something in the GitHub action cache.

jdaugherty (approved):

PR #14999, opened 2025-08-20
fix: add asset-pipeline-bom to grails bom

jdaugherty:


dependencies.gradle
```diff
@@ -22,25 +22,25 @@
...
-            'jna.version'                        : '5.17.0',
-            'jquery.version'                     : '3.7.1',
-            'objenesis.version'                  : '3.4',
-            'gradle-spock.version'               : '2.3-groovy-3.0',
-            'spring-boot.version'                : '3.5.4',
-            'springloaded.version'               : '1.2.8.RELEASE',
+            'ant.version'                   : '1.10.15',
+            'asciidoctor-gradle-jvm.version': '4.0.4',
+            'asciidoctorj.version'          : '3.0.0',
+            'asset-pipeline-gradle.version' : '5.0.15',
```
jamesfredley, line 28: Given that https://repo1.maven.org/maven2/cloud/wondrify/asset-pipeline-bom/5.0.15/asset-pipeline-bom-5.0.15.module includes `asset-pipeline-gradle`, I think we should put asset-pipeline-bom.version in gradleBomDependencyVersions, move `'asset-pipeline-bom': "cloud.wondrify:asset-pipeline-bom:${bomDependencyVersions['asset-pipeline-bom.version']}",` up to gradleBomPlatformDependencies  and then add it to `combinedPlatforms = ['spring-boot-bom': gradleBomPlatformDependencies['spring-boot-bom']] + bomPlatformDependencies` so asset-pipeline-bom is in both boms.

jdaugherty: I intentionally did not add it to the gradle bom because it would then add conflicting dependencies that could cause issues for Gradle.  It's the same reason we have a gradle bom & grails bom.  

For the asset plugin, I thought only the gradle plugin gets put on the gradle build classpath.  Am I wrong in that assumption?  I don't see a benefit of adding the other libraries for that reason.I also don't think the gradle plugin ever should be included on the application classpath, so i don't think we want to mix these.

jdaugherty: I think the include of gradle of the asset bom was a mistake.  I'll get a PR to update the asset bom to remove it.

jdaugherty: https://github.com/wondrify/asset-pipeline/pull/390 will remove the gradle plugin from the asset bom

jamesfredley: I only saw asset-pipeline versions in the https://repo1.maven.org/maven2/cloud/wondrify/asset-pipeline-bom/5.0.15/asset-pipeline-bom-5.0.15.module , so I didn't think it would contribute to conflicting dependencies and simplify the grails setup, but I see your point on gradle only ever needing asset-pipeline-gradle and am OK not adding asset-pipeline-bom to gradleBomPlatformDependencies.

It would still be nice to have one asset-pipeline version, if possible, so we do not accidentally mismatch them in the two boms.

jdaugherty: If you add something to a bom, every dependency of that library gets added too - and then gradle has to resolve conflicts. 

I agree that one version would be nice, but I'm not sure the best way to accomplish that.  I think there's a another change required here where we go back to properties in gradle.properties, but I think it's out of scope for this change.  Are you ok with this change for now and we can follow-up on simplifying later?

dependencies.gradle
```diff
@@ -72,37 +72,37 @@ ext {
...
-            'liquibase-hibernate5.version' : '4.27.0',
-            'mongodb.version'              : '5.5.1',
-            'rxjava.version'               : '1.3.8',
-            'rxjava2.version'              : '2.2.21',
-            'rxjava3.version'              : '3.1.11',
-            'selenium.version'             : '4.34.0',
-            'spock.version'                : '2.3-groovy-4.0',
-            'sitemesh.version'             : '2.6.0',
-            'starter-sitemesh.version'     : '3.2.2',
+            'asset-pipeline-bom.version'  : '5.0.15',
```
jamesfredley, line 75: `'asset-pipeline-bom.version' : gradleBomDependencyVersions['asset-pipeline-gradle.version'],`

would get us down to a single asset-pipeline version.  I don't love it entirely, since the flip would make more sense, but this is structured with gradle-bom first.

jamesfredley (approved):

dependencies.gradle
jdaugherty: This isn't compatible with the bom version page that is published.  It assumes the versions are in the matching map.

jamesfredley: OK

matrei (approved):

PR #15002, opened 2025-08-21
Fix/hibernate session integration tests 14920

rgc-code:
I've successfully implemented a solution to address the Hibernate session and transaction issue in Grails
  integration tests. The solution provides a way for integration tests to run with the same transactional context      
  as the application.

  What was implemented:

  1. @WithSession annotation
  (grails-testing-support-core/src/main/groovy/grails/testing/mixin/integration/WithSession.groovy):
    - New annotation that binds a Hibernate session without starting a transaction
    - Can be applied at class or method level
    - Supports specifying specific datasources
  2. Session binding infrastructure:
    - GrailsTestSessionInterceptor - Manages session binding without transactions
    - WithSessionSpecExtension - Spock extension for handling @WithSession
    - WithSessionTransformation - AST transformation for the annotation
  3. Enhanced test infrastructure:
    - Updated GrailsTestMode to support bindSession flag
    - Modified GrailsTestInterceptor to handle session-only binding
    - Updated IntegrationSpecConfigurerExtension to enable session binding by default
  4. Test examples:
    - WithSessionIntegrationSpec - Demonstrates usage of @WithSession
    - SessionBindingComparisonSpec - Shows differences between test modes
  5. Documentation:
    - Added comprehensive documentation in the integration testing guide

  How it solves the problem:

  The issue described three scenarios:
  - Without @Rollback: No session bound → tests fail for operations that work in the app
  - With @Rollback: Both session and transaction → tests pass for operations that fail in the app
  - With @WithSession: Session without transaction → tests match application behavior

  The new @WithSession annotation provides a session binding that matches the OISV (Open Session In View) pattern      
  used in running applications, where:
  - SELECT operations work (session available)
  - save() without flush works (no transaction needed)
  - save(flush: true) fails without @Transactional (matches runtime behavior)

  This ensures integration tests accurately reflect the application's runtime behavior, preventing false positives     
   and false negatives in test results.


jamesfredley:
@RAJ-debug858 Thank you.  Can you resubmit with the a target branch of 7.0.x?

When I try to make this change I get "There are no new commits between base branch '7.0.x' and head branch 'fix/hibernate-session-integration-tests-14920'", so I am guessing some of your new commits are not present on your forked branch.

jdaugherty:
FYI: you can add your branch as a remote, fetch, cherry-pick your change, and then adjust for where its supposed to be.  GitHub is showing 10,000+ commits so my guess is this change was made against an earlier branch before we created the mono repo.

aether-dev07:
@jamesfredley Thank you! I’ve rebased my changes onto the 7.0.x branch and force-pushed the updated branch. The PR now targets 7.0.x as requested.

aether-dev07:
- grails-test-core/build.gradle (modified)
  - grails-testing-support-core/build.gradle (modified)
  - The moved files in grails-testing-support-core/src/main/groovy/org/grails/test/
  - grails-testing-support-core/src/main/resources/META-INF/services/org.spockframework.runtime.extension.IGlobalE     
  xtension (modified)
  - The fixed Java file with the import change

grails-testing-support-core/build.gradle
```diff
@@ -47,7 +47,8 @@ dependencies {
 
     api project(':grails-databinding')
     api project(':grails-datamapping-core')
-    api project(':grails-test-core')
```
jdaugherty, line 50: test-core is meant to be the parent to this dependency. Why are you removing it?

grails-test-core/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
     implementation platform(project(':grails-bom'))
 
     api 'org.springframework:spring-tx'
+
```
jdaugherty: The extra newline should be removed

grails-testing-support-core/build.gradle
```diff
@@ -48,6 +48,7 @@ dependencies {
     api project(':grails-databinding')
     api project(':grails-datamapping-core')
     api project(':grails-test-core')
+    api('org.spockframework:spock-core') { transitive = false }
```
jdaugherty, line 51: this is incudes by test-core so it should be unnecessary

grails-testing-support-core/src/main/groovy/org/grails/testing/context/junit4/GrailsTestConfiguration.java
```diff
@@ -29,7 +29,7 @@
  * @author Graeme Rocher
  * @since 3.0
  */
-@ContextConfiguration(loader = GrailsApplicationContextLoader.class)
+@ContextConfiguration(loader = SpringBootContextLoader.class)
```
jdaugherty: Why aren't we using the Grails Context?

jdaugherty (changes requested):
Hibernate considers OSIV an anti pattern.  I understand that previous Grails applications have used it and I fully support adding test support for the feature, but we should not make this behavior the default given that it's considered bad practice upstream.

grails-testing-support-core/build.gradle
aether-dev07: Restored grails-test-core dependency in grails-testing-support-core

jdaugherty:
Can't the test behavior be driven by whether the application uses OSIV?

grails-testing-support-core/src/main/groovy/org/grails/testing/context/junit4/GrailsTestConfiguration.java
aether-dev07: okay we should use the Grails-specific context loader since it provides Grails
  application context setup rather than just generic Spring Boot context. The GrailsApplicationContextLoader
  extends SpringBootContextLoader but adds Grails-specific initialization through GrailsApp.

grails-test-core/src/main/groovy/org/grails/test/spock/IntegrationSpecConfigurerExtension.groovy
```diff
@@ -60,7 +60,10 @@ class IntegrationSpecConfigurerExtension implements IAnnotationDrivenExtension<A
 
         IntegrationSpecMethodInterceptor(ApplicationContext applicationContext) {
             this.applicationContext = applicationContext
-            this.mode = new GrailsTestMode(autowire: true, wrapInTransaction: true, wrapInRequestEnvironment: true)
+            // By default, bind session for integration tests to match application behavior
```
jdaugherty, line 64: The application behavior depends if OSIV is configured and if hibernate is used.  We should have our tests default to best behaviors, or even better, detect the configuration and then only apply it based on the application configuration.

grails-testing-support-core/src/main/resources/META-INF/services/org.spockframework.runtime.extension.IGlobalExtension
```diff
@@ -1 +1 @@
-org.grails.testing.spock.TestingSupportExtension
+org.grails.test.spock.WithSessionSpecExtension
```
jdaugherty, line 1: What about the original extension?  Why are we replacing the old behavior instead of augmenting it?

grails-testing-support-core/src/main/groovy/org/grails/test/support/GrailsTestSessionInterceptor.groovy
```diff
@@ -0,0 +1,187 @@
...
+import org.slf4j.LoggerFactory
+
+/**
+ * Binds Hibernate sessions to the test thread without starting transactions.
+ * This mimics the OISV (Open Session In View) pattern used in running applications.
+ */
+@CompileStatic
+class GrailsTestSessionInterceptor {
+    
+    private static final Logger LOG = LoggerFactory.getLogger(GrailsTestSessionInterceptor)
```
jdaugherty, line 40: This is a groovy class, is there a reason you can't use `@Slf4j` ?

grails-test-core/src/main/groovy/org/grails/test/support/GrailsTestInterceptor.groovy
```diff
@@ -30,6 +30,7 @@ class GrailsTestInterceptor {
     private String[] testClassSuffixes
 
     private GrailsTestTransactionInterceptor transactionInterceptor
+    private GrailsTestSessionInterceptor sessionInterceptor
```
jdaugherty, line 34: This is not compiling.

jdaugherty (changes requested):

jamesfredley:
This needs refactored to be an optional feature vs replacing the existing feature.

borinquenkid:
@rgc-code If you want this kept open update it with the changes requested

PR #15004, opened 2025-08-21
Remove embedded-mongodb feature and references

### LINKED ISSUE #14983: grails-embedded-mongodb plugin for Grails 7?
### Issue description

Will/should this be updated for Grails 7 since we are using testcontainers now?

[grails-embedded-mongodb](https://github.com/grails-plugins/grails-embedded-mongodb)

If not updated we should removed the feature from grails-forge

jamesfredley:
Eliminated the embedded-mongodb feature, its documentation, implementation, tests, and dependencies from the project. Updated MongoDB feature configuration and related documentation to reflect this removal.


jdaugherty (approved):
I am ok removing the embedded plugin, but shouldn't we change the feature to use the testing library I added to core?  Should we just include that as the default testImplementation in the mongodb feature?

PR #15005, opened 2025-08-21
Hide servlet features, should be selected via --servlet instead

jamesfredley:
the servlet container should be selected
via --servlet
jetty      none       tomcat     undertow

Selecting via features does not properly override ServletImpl


jdaugherty (approved):

matrei (approved):
Should we also simplify the condition in the `shouldApply` methods of `SpringBoot{ServletContainer}Feature` classes.

```java
public boolean shouldApply(ApplicationType applicationType, Options options, Set<Feature> selectedFeatures) {
    return options.getServletImpl() == ServletImpl.JETTY;
}
```

jamesfredley:
@matrei I updated the shouldApply() methods for the 3 servlet container features and retested.  They are concise and understandable now.

PR #15016, opened 2025-08-26
Enforce code style

matrei:
This PR adds CodeNarc and Checkstyle Gradle Plugins to enforce code styles.
It adds a Gradle task `codeStyle` to run all `codenarc*` and `checkstyle*` tasks exclusively.

It also adds a Gradle property check for `skipCodeStyle` to skip code style checking when running `build` or `check`.
Existing GitHub workflows are updated to use this property and a new workflow is added that does only code style checks.

```console
# Run all style checks locally
./gradlew codeStyle

# Include style checks during build/check
./gradlew check
./gradlew build

# Skip style checks (e.g., for quick iterations or CI main pipelines)
./gradlew check -PskipCodeStyle
./gradlew build -PskipCodeStyle
```


.github/workflows/codestyle.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: ./gradlew codeStyle
+      - name: "📤 Upload Failure Reports"
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: core-reports
+          path: |
+            **/build/reports/checkstyle/
+            **/build/reports/codenarc/
+  check_gradle_plugin_projects:
```
jamesfredley, line 55: Do we need a job here for grails-forge?

matrei: Good catch!

.github/workflows/gradle.yml
```diff
@@ -256,7 +275,9 @@ jobs:
         working-directory: 'grails-gradle'
         run: >
           ./gradlew publish aggregateChecksums aggregatePublishedArtifacts
-          -Dorg.gradle.internal.publish.checksums.insecure=true --no-build-cache --rerun-tasks
+          --no-build-cache
+          --rerun-tasks
+          -Dorg.gradle.internal.publish.checksums.insecure=true
```
jdaugherty: the insecure publishing is no longer required since we use the grails publish plugin and it sets this up by default

jdaugherty (approved):
Can we update DEVELOPMENT.md for the new properties & codestyle workflows?

jamesfredley (approved):

PR #15019, opened 2025-08-27
Forge fixes

### LINKED ISSUE #14972: Misc Forge Feature issues
These are some issues I noticed when going through the available features in the Grails Forge:

- [x] Adding feature `JRebel` gives error: `There can only be one of the following features selected: [spring-boot-devtools, jrebel]`
- [x] Adding feature `Fat/Shaded Jar` gives error: `Internal Server Error: Failed to find a match for the provided id: [shadow]`
- [ ] ~~Adding feature `AssertJ` gives error: `AssertJ requires JUnit`~~
- [ ] ~~Adding feature `Hamcrest` gives error: `Hamcrest requires JUnit`~~
- [x] Adding feature `H2 Database` gives error: `No features have been selected. Please choose one or more features and try again.`
- [x] Adding feature `Embedded Jetty` gives error: `There can only be one of the following features selected: [spring-boot-starter-tomcat, spring-boot-starter-jetty]`
- [x] Adding feature `Embedded Undertow` gives error: `There can only be one of the following features selected: [spring-boot-starter-undertow, spring-boot-starter-tomcat]`
---
- [x] Feature `Quartz` adds `4.0.0-SNAPSHOT` version (`4.0.0-M3` is released)
- [x] Feature `Web Console` adds `7.0.0-M2` version (`7.0.0-M3` is released). The url associated with Web Console documentation (https://plugins.grails.org/plugin/sheehan/console) is a 404.
- [x] Feature `Logback Groovy Config` adds version `1.12.4` (`1.14.5` is the latest)
- [x] Feature `Micronaut Http Client` adds version `4.6.5` but `4.9.9` is the latest and companion library `micronaut-serde-jackson` is `2.11.0` but `2.15.1` is the latest.
---
- [x] Feature `Testcontainers` is in the `Database` category

matrei:
Resolves #14972


grails-forge/grails-forge-core/src/main/resources/pom.xml
```diff
@@ -55,8 +55,8 @@
         </dependency>
         <dependency>
             <groupId>com.gradleup.shadow</groupId>
-            <artifactId>shadow-gradle-plugin</artifactId>
-            <version>8.3.6</version>
+            <artifactId>com.gradleup.shadow.gradle.plugin</artifactId>
+            <version>9.0.2</version>
         </dependency>
         <dependency>
             <groupId>com.github.erdi</groupId>
```
jdaugherty, line 62: Looking over this view - erdi caught my eye. I didn't think we use this anymore since it's incompatible, should the web driver binaries plugin be removed?

jdaugherty (approved):

grails-forge/grails-forge-core/src/main/resources/pom.xml
matrei: Yes, I think you are right as `webdriver-binaries-gradle-plugin` is not working with Gradle 8.7+ and the project seems to be dead: erdi/webdriver-binaries-gradle-plugin#44
I'll remove it.

matrei: @jdaugherty Actually, I see that @jamesfredley has put in a lot of effort to keep the `webdriver-binaries` solution alive a bit longer so I reverted that removal and let him decide whether it stays or goes.

jamesfredley: I think we should keep webdriver-binaries as a non-default, optional feature for Grails 7 and then remove it for Grails 8.  That gives users transition time and it works with up to Gradle 8.6 with up to JDK 21.

Language in the title and description about these limitations.

https://github.com/apache/grails-core/blob/00b8008c907807502f2d5a7605f5ae7aa3745907/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/test/GebWithWebDriverBinaries.java#L53-L62

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/other/ShadePlugin.java
```diff
@@ -68,7 +68,7 @@ public void apply(GeneratorContext generatorContext) {
         generatorContext.addHelpLink("Shadow Gradle Plugin", "https://plugins.gradle.org/plugin/com.gradleup.shadow");
         GradlePlugin.Builder builder = GradlePlugin.builder()
                 .id("com.gradleup.shadow")
-                .lookupArtifactId("shadow");
+                .lookupArtifactId("com.gradleup.shadow.gradle.plugin");
```
jamesfredley: This looks slightly different on the other plugins referenced by id, with a version from POM.xm.

https://github.com/apache/grails-core/blob/00b8008c907807502f2d5a7605f5ae7aa3745907/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/reloading/Jrebel.java#L47-L50

https://github.com/apache/grails-core/blob/00b8008c907807502f2d5a7605f5ae7aa3745907/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/asciidoctor/Asciidoctor.java#L65-L68

jamesfredley (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/other/ShadePlugin.java
matrei: Yes, you are right, I used the wrong one: https://mvnrepository.com/artifact/com.gradleup.shadow
It does not matter in this case but I'll change it to `shadow-gradle-plugin`.

grails-forge/grails-forge-core/src/main/resources/pom.xml
jamesfredley: There is also code which adjusts the Gradle and Java version when `geb-with-webdriver-binaries` is selected, with comments in generated config.

jdaugherty: Makes sense.  Thank you!

PR #15031, opened 2025-08-28
Explicitly type render(object) to help IntelliJ ambiguous method overloading

jdaugherty:
@matrei @jamesfredley Later versions of intellij show render() calls with a map as an ambiguous method call.  Testing locally, it seems that intellij prefers (Object object) instead of just (object).  What are your thoughts about making this change? It shouldn't matter for groovy / compiler, but it seems to help IntelliJ.


jamesfredley (approved):

matrei (approved):

sbglasius:
@jdaugherty it's a welcomed change!

PR #15034, opened 2025-08-29
Formatting helpers

matrei:
Initial formatting helper files for IntelliJ.
Might not cover every rule in the code styles yet, but can be iterated upon.


jdaugherty (approved):
Thank you

PR #15036, opened 2025-08-29
Support for Gorm Entities with same name, but different packages

codeconsole:
Currently RestfulServiceController will not work for situations where you have domain objects with the same name.

This scenario is more likely for complex websites that use plugins with domain objects.

For instance:

```
grails-app/domain/org/website/User.groovy
grails-app/domain/org/website/plugin/User.groovy
```

```groovy
package org.website

import org.springframework.security.provisioning.UserDetailsManager

@Scaffold(GormService<User>)
class UserService implements UserDetailsManager {
    // implementation 
}
```

```groovy
package org.website.plugin

@Scaffold(GormService<User>)
class MyPluginUserService  {
    //  implementation 
}
```

This introduces a fix that will resolve the appropriate service class for the respective GormService


jdaugherty (changes requested):
We need to add tests for this to merge it.

codeconsole:
@jdaugherty the scaffolding plugin doesn't have any tests? Do you know of any?  

This is a fix.  If there are no existing tests, it is better to merge than to leave broken, no?

jdaugherty:
@codeconsole We should add a functional test project.  DomainServiceLocator appears more complex than it should be.  How do we know an existing workflow didn't break without a test?

codeconsole:
@jdaugherty 
> DomainServiceLocator appears more complex than it should be. 

I've simplified it to its original intent. It now just only uses services of type GormService

> How do we know an existing workflow didn't break without a test?

RestfulServiceController is a new class that is just an example of the **Service Layer** pattern.  There are no existing workflows and it would be silly to just add 1 test to the entire plugin that only tests this.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Resolves the appropriate service bean for a given domain class by:
+ *   - Scanning only beans of type GormService
+ *   - Matching via: ((GormEntity<?>) service.getResource()).instanceOf(domainClass)
+ *
+ * Keeps a single shared cache for the whole app.
+ */
+public final class DomainServiceLocator {
+
+    private static final ConcurrentMap<Class<?>, GormService<?>> CACHE = new ConcurrentHashMap<>();
```
jdaugherty, line 42: GormService will be a bean, what if someone has reloading enabled?  Won't this then break their ability to reload?  One solution would be to move this to a bean itself.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
```diff
@@ -0,0 +1,98 @@
...
+        return found;
+    }
+
+    /** Clear cache (useful in tests/dev reloads). */
+    public static void clear() {
+        CACHE.clear();
+    }
+
+    private static <T extends GormEntity<T>> GormService<T> findService(Class<T> domainClass) {
+        ApplicationContext ctx = Holders.getGrailsApplication().getMainContext();
```
jdaugherty, line 67: If we're going to rely on beans, why is this class itself not a bean?  Is it used earlier in the init process?

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
```diff
@@ -19,45 +19,52 @@
...
 import grails.rest.RestfulController
-import grails.util.Holders
-import org.grails.datastore.gorm.GormEntityApi
+import org.grails.datastore.gorm.GormEntity
 
 @Artefact('Controller')
 @ReadOnly
-class RestfulServiceController<T> extends RestfulController<T> {
+@CompileStatic
+class RestfulServiceController<T extends GormEntity<T>> extends RestfulController<T> {
```
jdaugherty, line 31: By changing a definition of this class that's widely extended, this may potentially break every subclass and require recompilation. We've stated we would not make these changes after RC2.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
```diff
@@ -0,0 +1,98 @@
...
+import org.grails.datastore.gorm.GormEntity;
+
+/**
+ * Resolves the appropriate service bean for a given domain class by:
+ *   - Scanning only beans of type GormService
+ *   - Matching via: ((GormEntity<?>) service.getResource()).instanceOf(domainClass)
+ *
+ * Keeps a single shared cache for the whole app.
+ */
+public final class DomainServiceLocator {
```
jdaugherty, line 40: We should add a functional test for scaffolding - we should be able to easily add 2 examples of this scenario to prove this works.

jdaugherty (commented):
I've added this to Wednesday's developer meeting to see where people stand on this.  I tried to point out in the review some possible problem points.  I'm not 100% sure if it would cause compilation issues - but if it does we have to agree to do another RC.

jamesfredley:
For the scaffolding functional tests, the following existing test projects are a good place to copy as a quick starting point.  These all use Geb w/ TestContainers.

https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/geb
https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/gsp-layout
https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/app1

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
codeconsole: @jdaugherty where is a scenario where T does not extend GormEntity<T> ?

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
codeconsole: @jdaugherty cache is not used in development mode.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
codeconsole: @jdaugherty it's a utility class that is only used post initialization.   There is no need for it to be a bean.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
codeconsole: @jdaugherty Are there existing functional tests for scaffolding?  if so, do you have an explanation how they work?

codeconsole:
> For the scaffolding functional tests, the following existing test projects are a good place to copy as a quick starting point. These all use Geb w/ TestContainers.
> 
> https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/geb https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/gsp-layout https://github.com/apache/grails-core/tree/7.0.x/grails-test-examples/app1

@jamesfredley are there already functional tests for the scaffolding plugin in those examples? Where exactly are you saying to put it?

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
jdaugherty: We discussed the risk on Wednesday; we're willing to put this in 7.0.0 without changing this given the low adoptance of this.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/DomainServiceLocator.java
jdaugherty: our functional test apps are under `grails-test-examples/*`.  We do not have an existing app.  We only want you to generate one and test the 2 scenarios.  We do not think we need to test all of scaffolding, but as we add more features we need to start enhancing the tests.

jdaugherty: For example: 

Add a subproject: `grails-test-examples/scaffolding` (can copy gsp or sitemesh if you want as a basis, the only thing that needs updated is settings.gradle & adding the files from grails-test-examples/scaffolding).  Then just add your domain scenario & a geb test that calls the screen / makes sure it doesn't error.  

We're not asking you to test every functionality of scaffolding.  Just to grow the test coverage in this area as we make enhancements.

jdaugherty:
I took the app you created at https://github.com/codeconsole/scaffold-app and integrated it under grails-test-examples/scaffolding.  

I added a test to invoke the user/index and it works correctly.  `community/user/index` though is erroring: 

image

codeconsole:
@jdaugherty you missed https://github.com/codeconsole/scaffold-app/commit/60cc372383a9e81b134907c684fa42eab566c883

jdaugherty (approved):

jdaugherty:
@jamesfredley  can you please review?

jamesfredley (approved):

PR #15040, opened 2025-09-02
Update to Asset Pipeline 5.0.16

jdaugherty:
As a follow-up to https://github.com/apache/grails-core/pull/14953, since we depend on the asset pipeline, the coordinate change for grails-i18n broke forge.  This should resolve it once 5.0.16 is findable on maven central.


jamesfredley (approved):

grails-forge/grails-forge-cli/src/test/groovy/org/grails/forge/cli/CommandSpec.groovy
```diff
@@ -70,6 +70,7 @@ class CommandSpec extends Specification {
         String[] args = builder.toString().split(' ')
         ProcessBuilder pb = new ProcessBuilder(args)
         pb.environment().put('JAVA_HOME', System.getenv('JAVA_HOME') ?: System.getProperty('java.home'))
+        pb.environment().put('GRAILS_REPO_URL', System.getenv('GRAILS_REPO_URL') ?: null)
```
matrei, line 73: From https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ProcessBuilder.html#environment():
The returned map does not permit null keys or values. Attempting to insert or query the presence of a null key or value will throw a NullPointerException.

PR #15052, opened 2025-09-05
fix: wrapper should only consider release types that are a higher priority than the wrapper version when packaged

jdaugherty:
@matrei had a ClassNotFoundException thrown when testing the 7.0.0-RC2 wrapper locally.  Upon researching the problem, the cause was he had an old snapshot version under the directory we download grails releases to.  The reason this was a problem is we were returning snapshot builds as eligible & the wrapper assumes snapshots are more up to date than a release.  

This line is the problem: https://github.com/apache/grails-core/blob/2cb067e0176f86bc40a17375523b5eb52503a6d3/grails-wrapper/src/main/java/grails/init/Start.java#L151 

That code should only return release types later than the version of the packaged wrapper:

      return new LinkedHashSet<>(myVersion.releaseType.upTo());
   

I wanted to add test coverage to the wrapper though, so I've refactored the helper methods for preferred version & allowed types to make this easier to test.   The test "allowed release types - no preferred version - non-development for non-release" makes sure this issue doesn't happen in the future.

With the refactoring, the jar size is 30KB, which is still smaller than the gradle wrapper size (44KB) so I'm considering the extra classes as an acceptable.

### DIFF
--- a/grails-wrapper/src/main/java/grails/init/GrailsWrapper.java
+++ b/grails-wrapper/src/main/java/grails/init/GrailsWrapper.java
@@ -1,0 +1,82 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package grails.init;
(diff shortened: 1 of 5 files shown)


jamesfredley (approved):

grails-wrapper/src/main/java/grails/init/GrailsWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.stream.Collectors;
+
+import grails.proxy.SystemPropertiesAuthenticator;
+
+/**
+ * The purpose of this class is to download the expanded Grails wrapper jars into GRAILS_WRAPPER_HOME (`.grails` in the project root)
```
matrei: Is this information correct? (`.grails` in the _project_ root)

matrei (approved):
Thank you!

grails-wrapper/src/main/java/grails/init/GrailsWrapper.java
jamesfredley: It was for a short moment, but then went back to ~/.grails/

jdaugherty: I'll update this - we changed the default but you can still change where the wrapper home is by defining GRAILS_WRAPPER_HOME

PR #15053, opened 2025-09-05
Update GitHub references and default project slug for documentation

jamesfredley:
Updated CONTRIBUTING.md to clarify issue tracking for documentation. Changed the default githubSlug in CreateReleaseDropDownTask from 'apache/grails-doc' to 'apache/grails-core'.


jdaugherty (approved):
Shouldn't we error if the githubSlug isn't defined?

PR #15054, opened 2025-09-05
Add external references for dependency graph

jamesfredley:
Introduced `.github/dependency-graph/external-references.yml` to associate the repository with published Maven packages. This enables GitHub to display accurate 'Used by' information for the repository.  Example Used By section:  https://github.com/protocolbuffers/protobuf

See https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/exploring-the-dependencies-of-a-repository#changing-the-used-by-package
This file associates this repository with packages published to external registries (Maven Central)
so that GitHub can display the correct "Used by" package on the repository page.
The Package URL (purl) format is defined at https://github.com/package-url/purl-spec


jdaugherty (approved):
We should put the old package names as well.

PR #15055, opened 2025-09-05
build reproducibility & verification issues after testing 7.0.0-RC2

jdaugherty:
1. reproducibility issues

@matrei noticed that he was having reproducibility issues for grails-redis when run under docker desktop.  These issues did not exist for me under orbstack or when running locally.

Debugging this, the source jars were different because the file listings inside of the source jars were different.  Specifically, the listing across source sets - sometimes grails-app/init would be first and sometimes it would be later.  

Digging into this, we do define the source / resource directories based on file system order instead of a deterministic order.  This PR sorts those to help these issues in the future (we may need to open an upstream gradle bug when using reproducibleFileOrder=true if this continues).  From what I can tell, the gradle source uses lists / linkedHashSets so as long as it's added in the same order, it's processed in the order.

2. reproducible script issues 

@matrei  noted the reproducible script was failing because after a manual diff of all files, no differences were found.  The script did not exit properly or ensure a diff file existed.  @paulk-asert  has previously mentioned these issues in prior releases but I couldn't reproduce.  Going forward this should resolve this.

3. KEYS file verification

As part of 7.0.0-RC2 we started verifying both the grails github action files & the svn files to make sure they are identical.  We now will verify the KEYS files are identical so they're always kept in sync with grails-core.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -512,6 +512,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
         for (String f in grailsAppResourceDirs) {
             grailsResourceDirs.add(project.file("grails-app/${f}"))
         }
+        // force a defined order for build reproducibility
+        grailsResourceDirs.sort { File a, File b -> a.name <=> b.name }
         grailsResourceDirs
     }
 
@@ -527,6 +529,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
             }
         }
         grailsSourceDirs.add(project.file('src/main/groovy'))
+        // force a defined order for build reproducibility
+        grailsSourceDirs.sort { File a, File b -> a.name <=> b.name }
         grailsSourceDirs
     }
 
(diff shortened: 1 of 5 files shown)


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -512,6 +512,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
         for (String f in grailsAppResourceDirs) {
             grailsResourceDirs.add(project.file("grails-app/${f}"))
         }
+        // force a defined order for build reproducibility
+        grailsResourceDirs.sort { File a, File b -> a.name <=> b.name }
```
matrei: `.sort` returns the list, so the return on the next line is redundant.

Possible simplification of method:
```groovy
@CompileStatic
protected List<File> resolveGrailsResourceDirs1(Project project) {
    ['src/main/resources']
        .addAll(grailsAppResourceDirs.collect { 'grails-app/' + it })
        .collect { project.file(it) }
        .sort { it.name } // sort for build reproducibility
}
```

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -527,6 +529,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
             }
         }
         grailsSourceDirs.add(project.file('src/main/groovy'))
+        // force a defined order for build reproducibility
+        grailsSourceDirs.sort { File a, File b -> a.name <=> b.name }
         grailsSourceDirs
```
matrei: Same here that `.sort` already returns the list.

Possible simplification:
```groovy
grailsSourceDirs
    .tap { add(project.file('src/main/groovy')) }
    .sort { it.name } // sort for build reproducibility
```

matrei (approved):

jamesfredley (approved):

PR #15060, opened 2025-09-08
Revise upgrading guide for Grails 7: grails-i18n plugin

jamesfredley:
This did not make it into the RC2 docs, but it will be there for the final GA.


jdaugherty (approved):
Thank you.  I completely forgot about this!

PR #15066, opened 2025-09-10
Remove additivity attribute from StackTrace logger

jamesfredley:
Prevents:  `WARN in Logger[StackTrace] - No appenders present in context [default] for logger [StackTrace]`.

The 'additivity="false"' attribute was removed from the StackTrace logger configuration in multiple logback-spring.xml files and the logback.rocker.raw template. This change simplifies the logger configuration and ensures it extends the root logger.

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
@@ -39,7 +39,7 @@ under the License.
     </root>
 
     <springProfile name="development">
-        <logger name="StackTrace" level="ERROR" additivity="false" />
+        <logger name="StackTrace" level="ERROR" />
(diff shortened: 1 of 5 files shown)


jdaugherty (commented):
So we're saying we should effectively configure all loggers as the root loggers and not allow messages to go to multiple logs?  Isn't the additivity configuration so that an exception logs to both the console log + an error log?

matrei (approved):
Good catch! With no appender and additivity disabled, events didn’t bubble up and were discarded.

jamesfredley:
The default generated logback config throws the following error when there is an exception in a controller.  I am not sure how it made it ~1 year that way, but this was focused on resolving that issue.  

`WARN in Logger[StackTrace] - No appenders present in context [default] for logger [StackTrace].`

I did not explore "exception logs to both the console log + an error log", just resolved the issue above, but the end user could go in that direction instead of the default generated simpler config.

PR #15067, opened 2025-09-11
Geb 8

### LINKED ISSUE #15065: Geb 8
### Issue description

Try out if it's feasible to upgrade to Geb 8 for Grails 7.

matrei:
Closes #15065


matrei:
@JonasPammer Would you like to look over these changes?

build.gradle
```diff
@@ -70,6 +70,7 @@ allprojects {
             url = 'https://repository.apache.org/content/groups/staging'
             content {
                 includeModuleByRegex('org[.]apache[.]grails[.]gradle', 'grails-publish')
+                includeModuleByRegex('org[.]apache[.]groovy[.]geb', 'geb-.*')
```
jdaugherty, line 73: I thought you could only have one of these includes?  does this effectively or the expressions?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -162,18 +162,10 @@ class WebDriverContainerHolder {
...
         }
 
-        // If `GebConfig` instantiates a `RemoteWebDriver` without using it's `remoteAddress` constructor,
-        // the `RemoteWebDriver` will be instantiated using the `webdriver.remote.server` system property.
-        String existingPropertyValue = System.getProperty('webdriver.remote.server')
+        // If a custom `GebConfig` instantiates a `RemoteWebDriver` without using it's `remoteAddress` constructor,
+        // the `RemoteWebDriver` will be instantiated using the `webdriver.remote.server` system property,
+        // so we need to set that value.
         System.setProperty('webdriver.remote.server', currentContainer.seleniumAddress.toString())
-        gebConfig.driver // This will implicitly call `createDriver()`
```
jdaugherty, line 169: Wasn't the intent of this code to force the test container always?

matrei: > Wasn't the intent of this code to force the test container always?

I think the intent was to force using the `currentContainer.seleniumAddress` when creating the driver.

build.gradle
matrei: Yes, I believe this should create two rules with logical OR.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: As the cautious/unecessary restore was removed, it will just happen at any the next call to [.driver](https://github.com/apache/groovy-geb/blob/4266eeeea6f59e4718339b93283327bf176563b9/module/geb-core/src/main/groovy/geb/Browser.groovy#L172)

JonasPammer:
now this is what i call generics, haha. I would not even know half of these functions and groovy/thread-nanigans, especially after having used solely django at work for last few months. :shushing_face:  :stuck_out_tongue_winking_eye: 

as long as the system env is set for when https://github.com/apache/groovy-geb/blob/4266eeeea6f59e4718339b93283327bf176563b9/module/geb-core/src/main/groovy/geb/Browser.groovy#L172 is called its still the same thing as my original thought/impl. :+1:

matrei:
> now this is what i call generics, haha. I would not even know half of these functions and groovy/thread-nanigans, especially after having used solely django at work for last few months. 🤫 😜
> 
> as long as the system env is set for when https://github.com/apache/groovy-geb/blob/4266eeeea6f59e4718339b93283327bf176563b9/module/geb-core/src/main/groovy/geb/Browser.groovy#L172 is called its still the same thing as my original thought/impl. 👍

Haha, I thought it might be overkill as well, but what if you mix `ContainerGebSpec` and regular `GebSpec` tests in the same project?

However, I don't think our GebConfig integration is working perfectly yet. I tried switching to `new RemoteWebDriver(new FirefoxOptions())` and it blows up as the container is browser specific and seems to default to chrome.
```console
Could not start a new session. Response code 500. Message: Could not start a new session. No nodes support the capabilities in the request
Host info: host: 'f1b81917b1f3', ip: '172.17.0.4'
Build info: version: '4.34.0', revision: '2a4c61c498'
System info: [os.name](http://os.name/): 'Linux', os.arch: 'amd64', os.version: '6.10.14-linuxkit', java.version: '21.0.8'
Driver info: driver.version: unknown
Build info: version: '4.34.0', revision: '707dcb4246*'
System info: [os.name](http://os.name/): 'Linux', os.arch: 'amd64', os.version: '6.8.0-79-generic', java.version: '17.0.14'
Driver info: org.openqa.selenium.remote.RemoteWebDriver
Command: [null, newSession {capabilities=[Capabilities {acceptInsecureCerts: true, browserName: firefox, grails:gebConfigUsed: true, moz:firefoxOptions: {prefs: {remote.active-protocols: 1}}}]}]
Capabilities {acceptInsecureCerts: true, browserName: firefox, grails:gebConfigUsed: true, moz:firefoxOptions: {prefs: {remote.active-protocols: 1}}}
org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: Could not start a new session. No nodes support the capabilities in the request
```

This is now fixed in [fcc4d82](https://github.com/apache/grails-core/pull/15067/commits/fcc4d821af41f36b53a4f20428f7fa6693ea032b)

build.gradle
jdaugherty: Wow, so it does:  https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/repositories/RepositoryContentDescriptor.html 

I had no idea!

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -245,19 +268,47 @@ class WebDriverContainerHolder {
...
+    private static DockerImageName getDefaultDockerImageName() {
+        DockerImageName.parse("$DEFAULT_DOCKER_IMAGE_NAME:$seleniumVersion")
+    }
+
+    private static DockerImageName createDockerImageName(Object configuredBrowser) {
+        validateConfiguredBrowser(configuredBrowser)
+        DockerImageName.parse("selenium/standalone-$configuredBrowser:$seleniumVersion")
+    }
+
+    private static void validateConfiguredBrowser(Object browser) {
```
JonasPammer: why only allow a hardcoded subset?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -59,99 +64,130 @@ import grails.plugin.geb.serviceloader.ServiceRegistry
...
+                throw new IllegalStateException(
+                        'The "driver" config value must be a Closure that returns an instance of RemoteWebDriver.'
+                )
+            }
+            gebConfigFileFound = true
+
+            // Prepare for creating a suitable container matching the driver
+            // configured in GebConfig.groovy, or more specifically,
+            // specified by the `configuredBrowser` property.
+            dockerImageName = createDockerImageName(gebConf.rawConfig.configuredBrowser)
```
JonasPammer: regression for GebConfig's that don't have this special var - add a default here?
we already imply Chrome at `new RemoteWebDriver(container.seleniumAddress, new ChromeOptions()` when no GebConfig existed on classpath

grails-geb/README.md
```diff
@@ -161,6 +161,28 @@ To enable tracing, set the following system property:
   
 This allows you to opt in to tracing when an OpenTelemetry collector is available.
 
+#### GebConfig.groovy and using non-default browser settings
+Provide a `GebConfig.groovy` on the test runtime classpath (commonly `src/integration-test/resources`, but any location on the test classpath works) to customize the browser.
+
+To make this work, ensure:
+1. The `driver` property in your `GebConfig` is a `Closure` that returns a `RemoteWebDriver` instance.
+2. You set a custom `configuredBrowser` property so that `ContainerGebSpec` can start a matching container (e.g. "chrome", "firefox", "edge").
```
JonasPammer: as pointed out some projects potentially have GebConfig next to ContainerGebConfig. I saw other extensions always prefix their variables with their extension (e.g. sauceLabsBrowser), maybe we should too to avoid confusion? e.g. `containerBrowser` (in reference to containergebspec)

JonasPammer (commented):
even opening grails-core brtings down my pc at home, much more when running tests, so could only do github overview

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: Missing `configuredBrowser` property in `GebConfig.groovy` will result in error message:
```console
Illegal 'configuredBrowser' property value in GebConfig.groovy: [null]. Valid values are [chrome, edge, firefox]
```

I don't think we can have a default for `configuredBrowser`, as that will risk the container not matching the configured driver.
Are you suggesting we revert/default to chrome driver and container if the `GebConfig.groovy` is missing the `configuredBrowser` property?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: At the moment, these three seem to be the only ones that have supporting container images: https://java.testcontainers.org/modules/webdriver_containers/#other-browsers

grails-geb/README.md
matrei: Yes, that is a better name!

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: Yes, default to chrome.
If possible, put a try-catch where the "could not find container matching driver restriction" appears, and put this message as a possible suggestion/fix from us before the printStackTrace.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: Yeah but don't know if we should impose a restriction here, for it could change or disallow adaptability by another.

JonasPammer:
> However, I don't think our GebConfig integration is working perfectly yet. I tried switching to `new RemoteWebDriver(new FirefoxOptions())` and it blows up as the container is browser specific and seems to default to chrome.
> 
> ```
> Could not start a new session. Response code 500. Message: Could not start a new session. No nodes support the capabilities in the request
> Host info: host: 'f1b81917b1f3', ip: '172.17.0.4'
> Build info: version: '4.34.0', revision: '2a4c61c498'
> System info: [os.name](http://os.name/): 'Linux', os.arch: 'amd64', os.version: '6.10.14-linuxkit', java.version: '21.0.8'
> Driver info: driver.version: unknown
> Build info: version: '4.34.0', revision: '707dcb4246*'
> System info: [os.name](http://os.name/): 'Linux', os.arch: 'amd64', os.version: '6.8.0-79-generic', java.version: '17.0.14'
> Driver info: org.openqa.selenium.remote.RemoteWebDriver
> Command: [null, newSession {capabilities=[Capabilities {acceptInsecureCerts: true, browserName: firefox, grails:gebConfigUsed: true, moz:firefoxOptions: {prefs: {remote.active-protocols: 1}}}]}]
> Capabilities {acceptInsecureCerts: true, browserName: firefox, grails:gebConfigUsed: true, moz:firefoxOptions: {prefs: {remote.active-protocols: 1}}}
> org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: Could not start a new session. No nodes support the capabilities in the request
> ```
> 
> This is now fixed in [fcc4d82](https://github.com/apache/grails-core/pull/15067/commits/fcc4d821af41f36b53a4f20428f7fa6693ea032b)

Weird, because [it](https://github.com/testcontainers/testcontainers-java/blob/51219646dca72ad267e575bf25d0b60208c60b42/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L195) uses [getStandardImageForCapabilities](https://github.com/testcontainers/testcontainers-java/blob/51219646dca72ad267e575bf25d0b60208c60b42/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L246)
Will try to reproduce

----

Update, i forgot about the chicken and egg problem. Can't call `.withCapabilities(((RemoteWebDriver)gebConf.createDriver()).getCapabilities())` because it tries to connect (`java.lang.NullPointerException: Cannot invoke "org.testcontainers.containers.BrowserWebDriverContainer.getSeleniumAddress()" because "this.container" is null`). Hiding

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: Pushed some changes, have a look.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: Good point! Pushed changes.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -243,6 +244,9 @@ class WebDriverContainerHolder {
             browser.driver.manage().timeouts().pageLoadTimeout(Duration.ofSeconds(settings.pageLoadTimeout))
         if (settings.scriptTimeout != DEFAULT_TIMEOUT_SCRIPT)
             browser.driver.manage().timeouts().scriptTimeout(Duration.ofSeconds(settings.scriptTimeout))
+        if (settings.atCheckWaiting != DEFAULT_AT_CHECK_WAITING) {
```
JonasPammer: to be honest, now that we read GebConfig like normal Browser instantiaton, I don't even  know if we should keep our GebConfig'able browser/driver settings as part of our custom GrailsGebSettings (this includes for example timeouts).

i.e. rather than cc17f14a48eebd897116408c371a94300e943c94
```
        env:
          # Make Geb tests more resilient in slow CI environments
          JAVA_TOOL_OPTIONS: -Dgrails.geb.atCheckWaiting=true
```

do

`grails-test-examples/geb-gebconfig/src/integration-test/resources/GebConfig.groovy`
```groovy
atCheckWaiting = System.getenv('CI') != null
```

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: I didn't mean for us to force an own RemoteWebDriver with ChromeOptions **just** because containerBrowser property is missing, but to (pseudo code - need to find the spot)

```groovy
                } catch (SessionNotCreatedException e){
                    log.error("Failed to create remote browser session: $e.message. Did you set a 'containerBrowser' property corresponding to the driver in GebConfig.groovy?", e)
                }
```

which probably will imply removing the gebConfigFileActive logic

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
```diff
@@ -66,15 +68,21 @@ class GrailsGebSettings {
...
-        restartRecordingContainerPerTest = Boolean.parseBoolean(
-                System.getProperty('grails.geb.recording.restartRecordingContainerPerTest', 'true')
+        restartRecordingContainerPerTest = getBooleanProperty(
+                'grails.geb.recording.restartRecordingContainerPerTest',
+                true
         )
         implicitlyWait = getIntProperty('grails.geb.timeouts.implicitlyWait', DEFAULT_TIMEOUT_IMPLICITLY_WAIT)
         pageLoadTimeout = getIntProperty('grails.geb.timeouts.pageLoad', DEFAULT_TIMEOUT_PAGE_LOAD)
         scriptTimeout = getIntProperty('grails.geb.timeouts.script', DEFAULT_TIMEOUT_SCRIPT)
+        atCheckWaiting = getBooleanProperty('grails.geb.atCheckWaiting', DEFAULT_AT_CHECK_WAITING)
```
JonasPammer: even if we wanna keep this GrailsGebSetting (other comment), it accepts more than a boolean - https://groovy.apache.org/geb/manual/current/#at-check-waiting

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: The good thing with `GrailsGebSettings` is that we can use system properties without changes or additions to source code.
I wish this was possible with stock Geb.

The option to use `GebConfig` is also available.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
matrei: Yes, but I thought it was overkill to implement logic for handling that.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: ok, i'd think if the user wants that he could also add system property code in his gebconfig but sure does not hurt, only will be weird when some things are settable and some not through this dep.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
JonasPammer: Shouldn't it just be getStringProperty? no need for own validation of correctness, it will fail later if incorrect anyway too

matrei: Wouldn't a String imply a "wait preset name"? https://groovy.apache.org/geb/manual/current/#content-dsl-wait

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: My thinking is, that it should not be necessary to have a `GebConfig` to run `ContainerGebSpec`s (convention over configuration).
For the `grails-core` functional tests for example, we have no `GebConfigs` (except one).
And if we wanted to set `atCheckWaiting=true` we would have to add a `GebConfig` to each functional test project.
Now we can just set a system property in CI.

> only will be weird when some things are settable and some not through this dep

I agree, but this is already the case with stock Geb: https://groovy.apache.org/geb/manual/current/#system-properties

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
JonasPammer: ah ok yes i thought it would just pick the right one with trial and error. yes your implementation is best as we set the default timeouts with environment too! :+1:

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
JonasPammer: [nice](https://github.com/apache/grails-core/pull/15067/commits/79b8708c8f39b4f77d11ce7c124dd9ac4ee246e6)!

JonasPammer (approved):
Great Thanks

.github/workflows/gradle.yml
```diff
@@ -173,6 +173,9 @@ jobs:
         with:
           develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY  }}
       - name: "🏃 Run Functional Tests"
+        env:
+          # Make Geb tests more resilient in slow CI environments
+          JAVA_TOOL_OPTIONS: -Dgrails.geb.atCheckWaiting.enabled=true
```
jdaugherty: Why make CI work differently then locally?  Does our JDK read this by default?  I would have assumed we would have defined this property as part of the test functional test configuration instead.

grails-geb/README.md
```diff
@@ -161,6 +177,28 @@ To enable tracing, set the following system property:
   
 This allows you to opt in to tracing when an OpenTelemetry collector is available.
 
+#### GebConfig.groovy and using non-default browser settings
+Provide a `GebConfig.groovy` on the test runtime classpath (commonly `src/integration-test/resources`, but any location on the test classpath works) to customize the browser.
```
jdaugherty, line 181: I assume this GebConfig is optional?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
```diff
@@ -65,17 +73,48 @@ class GrailsGebSettings {
...
     private static int getIntProperty(String propertyName, int defaultValue) {
         Integer.getInteger(propertyName, defaultValue) ?: defaultValue
     }
 
+    private static Number getNumberProperty(String propertyName, Number defaultValue) {
+        def propValue = System.getProperty(propertyName)
+        if (propValue) {
+            try {
+                if (propValue.contains('.')) {
+                    return Double.parseDouble(propValue)
```
jdaugherty: Does the geb api really use Doubles and not BigDecimals?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -270,58 +367,117 @@ class WebDriverContainerHolder {
...
+
+            static <T> T withProperty(String key, String value, Closure<T> body) {
+                propertiesWrappedOnFirstAccess // Access property to trigger property wrapping
+                def map = OVERRIDDEN_SYSTEM_PROPERTIES.get()
+                def prev = map.put(key, value)
+                try {
+                    return body.call()
+                } finally {
+                    if (prev == null) map.remove(key) else map[key] = prev
+                    if (map.isEmpty()) OVERRIDDEN_SYSTEM_PROPERTIES.remove()
```
jdaugherty, line 469: Should this be removed?  What recreates the map?  Is the assumption the thread won't be reused and will always terminate?

jdaugherty (approved):
Great Job!  Only minor comments ...

.github/workflows/gradle.yml
jdaugherty: I think this is my only concern. This seems better suited in the gradle files than in the CI configuration.

matrei: > Why make CI work differently then locally?

It should only be a concern in low resource environments, to not get false positives because the runner can't serve the page fast enough. We see these random failures in CI.

> Does our JDK read this by default?4

Gradle will automatically pass through the `JAVA_TOOL_OPTIONS` to the JVM, without needing `systemProperty 'grails.geb.atCheckWaiting.enabled', System.getProperty('grails.geb.atCheckWaiting.enabled')` in the build file.

>  I would have assumed we would have defined this property as part of the test functional test configuration instead.

We could do that, I think this way is more flexible.

grails-geb/README.md
matrei: > I assume this GebConfig is optional?

Yes, `GebConfig` is optional.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
matrei: > Does the geb api really use Doubles and not BigDecimals?

The Geb API uses `Number`. Should I create a `BigDecimal` instead of a `double`?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
matrei: No, it should stay.
```groovy
def c = ClosureDecorators.withSystemProperty({
    println System.getProperty('my.custom.property') // should print "customValue"
}, 'my.custom.property', 'customValue')
c()
new Thread({
    sleep 1000
    c()
}).start()
println System.getProperty('my.custom.property') // should print original value, probably null
```
```console
customValue
null

customValue
```

Without it:
```console
customValue
customValue

customValue
```

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
jdaugherty: I would have assumed BigDecimal is a better default

.github/workflows/gradle.yml
jdaugherty: Can we at least add a comment to the functional test config so we know this configuration is present / different in CI?

grails-geb/src/testFixtures/groovy/grails/plugin/geb/GrailsGebSettings.groovy
matrei: Changed to `BigDecimal`.

.github/workflows/gradle.yml
matrei: I moved setting `atCheckWaiting` to `functional-test-config.gradle`.

PR #15069, opened 2025-09-12
feature: allow disabling micronaut configuration by the grails plugin

jdaugherty:
We know that the micronaut support could have issues due to the lack of test coverage.  This PR adds a flag that allows disabling the auto configuration in the grails plugin so that if an issue arises, one can debug it locally.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
@@ -68,8 +68,9 @@ class GrailsExtension {
     boolean springDependencyManagement = true
 
     /**
-     * Whether the micronaut dependencies should be auto added to the project on detection of micronaut & the platform
-     * version enforced
+     * Whether the Micronaut `annotationProcessor` dependencies should be auto-added to the project
+     * on detection of the `grails-micronaut` plugin, and the version defined by the `micronautPlatformVersion`
+     * Gradle property is enforced.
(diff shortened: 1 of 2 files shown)


matrei (commented):
If this is for debugging, would it be more suitable with a Gradle property than changing the API of the Grails extension?

jdaugherty:
It's allowing the end user to opt out of functionality permanently if what we've done conflicts or prevents them from using micronaut.  A property wouldn't make sense since this setting is to turn off the functionality for the build permanently.

jamesfredley (approved):

matrei:
> It's allowing the end user to opt out of functionality permanently

OK, if it is not only for debugging, then an extension property makes more sense.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
```diff
@@ -67,6 +67,12 @@ class GrailsExtension {
      */
     boolean springDependencyManagement = true
 
+    /**
+     * Whether the micronaut dependencies should be auto added to the project on detection of micronaut & the platform
+     * version enforced
+     */
```
matrei:
```suggestion
    /**
     * Whether the Micronaut `annotationProcessor` dependencies should be auto-added to the project
     * on detection of the `grails-micronaut` plugin, and the version defined by the `micronautPlatformVersion`
     * Gradle property is enforced.
     */
```

matrei (approved):

PR #15070, opened 2025-09-16
Gorm service enhancements

codeconsole:
Slowing migrating this to how it was intended.

Services are kind of useless if they can't filter/query.  

GormService was originally modeled after`RestfulController<T>`, but that wasn't really necessary. 

This is another step that migrates towards making the scaffold service more useful 

My final version of GormService allows querying by params and counting by params.  Currently I am using this as a base class that I extend with a more meaningful implementation.   

# Allows this

```groovy
@Scaffold(domain = User)
class UserService {
    List<User> list(Map args) {
        args.firstName? User.findAllByFirstName(args.firstName, args) : User.list(args)
    }

    Long count(Map args) {
        args.firstName? User.countByFirstName(args.firstName) : User.count()
    }
} 

@Scaffold(RestfulServiceController<User>)
class UserController {}
```
now works and is properly paginated:
http://localhost:8080/user/index?firstName=scott

### DIFF
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
@@ -54,7 +54,7 @@ class GormService<T extends GormEntity<T>> {
         resource.list(args)
     }
 
-    Long count() {
+    Long count(Map args) {
         resource.count()
     }
 
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
@@ -63,7 +63,7 @@ class GormService<T extends GormEntity<T>> {
         if (readOnly) {
             return
         }
-        ((GormEntityApi) queryForResource(id)).delete(flush: true)
+        ((GormEntityApi) get(id)).delete(flush: true)
     }
 
     @Transactional
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
@@ -46,12 +46,8 @@ class GormService<T extends GormEntity<T>> {
         resourceName = GrailsNameUtils.getPropertyName(resource)
     }
 
-    protected T queryForResource(Serializable id) {
-        resource.get(id)
-    }
-
     T get(Serializable id) {
-        queryForResource(id)
+        resource.get(id)
     }
 
     List<T> list(Map args) {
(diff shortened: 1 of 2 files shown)


grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
```diff
@@ -46,19 +46,15 @@ class GormService<T extends GormEntity<T>> {
...
-        queryForResource(id)
+        resource.get(id)
     }
 
     List<T> list(Map args) {
         resource.list(args)
     }
 
-    Long count() {
+    Long count(Map args) {
```
jdaugherty, line 57: This parameter addition seems unused?  It doesn't look like there's a count method on domains that takes an argument.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
```diff
@@ -49,8 +50,9 @@ class RestfulServiceController<T extends GormEntity<T>> extends RestfulControlle
     }
 
     @Override
+    @CompileDynamic
```
jdaugherty: Why is this necessary?

codeconsole: In order to access `params`

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
codeconsole: It is meant for overriding.   You can do this:

```groovy
@Scaffold(domain = User)
class UserService {
    List<User> list(Map args) {
        args.firstName? User.findAllByFirstName(args.firstName, args) : User.list(args)
    }

    Long count(Map args) {
        args.firstName? User.countByFirstName(args.firstName) : User.count()
    }
} 
```

then in your scaffold controller, you could just hit /user/index?firstName=scott with no additional changes.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
matrei: To be able to compile statically, we can implement `WebAttributes` or do:
```groovy
protected Integer countResources() {
    def params = ((GrailsWebRequest) RequestContextHolder.currentRequestAttributes()).params
    getService().count(params)
}
```

matrei: Scratch that, Controllers of course already implement WebAttributes via transform, and should be able to compile statically anyway.

matrei (approved):

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/RestfulServiceController.groovy
jdaugherty: As @matrei stated, I don't think the CompileDynamic should be necessary.  I'm assuming you're having to use it because you're not using `@GrailsCompileStatic`?

codeconsole: works without it.  IDE was throwing an error, but gradle builds fine

PR #15076, opened 2025-09-17
14977 - multiproject reloading support

### LINKED ISSUE #14977: Development Mode for multiprojects should use classes instead of jar files
### Issue description

https://docs.grails.org/snapshot/guide/plugins.html#_plugins_and_multi_project_builds

Indicates you should be able to add `exploded=true` to the plugins `gradle.properties`

and the following to the web application` build.gradle`

```
grails {
    plugins {
        implementation project(':myplugin')
    }
}
```

https://github.com/apache/grails-core/blob/388387c9e3ba109af2f4fe1dd3b9117080162fbf/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy#L100-L118

But this is not causing the plugin to be reloaded in the the app when the plugin is modified.

https://github.com/apache/grails-core/issues/14974#issuecomment-3192511671 - alternate example with working reloading, which works great

```
tasks.named('bootRun') {
    doFirst {
        classpath = classpath.filter { file ->
            !file.path.contains('base.jar')  // Exclude the JAR to avoid duplicates
        } + project(':base').sourceSets.main.output  // Add classes and resources output
    }
}
```

jdaugherty:
While multiproject reloading was possible before Grails 6, it was rather difficult to configure, the following had to be done: 

1. the property `grails.run.active` was set
2. the plugin project had the property `exploded` set to `true` prior to the application of the `grails-plugin` Gradle plugin
3. plugins were added to the Grails Application project via the `plugins` block inside of the `grails` extension in `build.gradle` instead of the `dependencies` block
4. the property `exploded` was set on the `grails` extension in `build.gradle` of the Grails Application project

This PR cleans up this process by adding a new gradle plugin `org.apache.grails.gradle.grails-exploded`, now the following has to be true for reloading to work: 

1. In the plugin project, apply the gradle plugin `org.apache.grails.gradle.grails-exploded`
2. In the application project, define plugins inside of the `plugins` block of the `grails` extension in `build.gradle`:

As part of this PR, I have also: 

1. cleaned up old variables on GrailsGradlePlugin that are no longer used since we adopted the grails-bom
2. moved the groovy compilation configuration back into compileGroovy.  

For the groovy compile configuration changes, the core issue is we need to be sure a non-cached version of the configuration file is generated whenever groovy would recompile / not cache.  The issue is to do this we need to depend on the runtimeClasspath - if the build changes we need to rerun, if the project switches from application to a plugin we need to rerun, if the project name changes rerun, if the version changes rerun, etc.  This means that any task that interacts with the runtime classpath, now needs to declare a dependency on our customization task.  This really makes the build fragile and is the reason I've moved the configuration back into doFirst { }.  When gradle changes the configuration to be provider based, I think we can revisit this.

### DIFF
--- a/grails-doc/src/en/guide/plugins/creatingAndInstallingPlugins.adoc
+++ b/grails-doc/src/en/guide/plugins/creatingAndInstallingPlugins.adoc
@@ -20,7 +20,7 @@ under the License.
 
 ==== Creating Plugins
 
-Creating a Grails plugin is a simple matter of running the command:
+Creating a Grails plugin is a simple matter of running one of the supported plugin generators:
 
 Grails Shell CLI
 [source,groovy]
@@ -120,7 +120,7 @@ Adds Quartz job scheduling features
     Closure doWithSpring()......
 ----
 
-==== Plugin Configuration
+===== Plugin Configuration
 
 A Grails plugin can be configured in one of the following files:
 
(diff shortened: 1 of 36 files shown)


jdaugherty:
FYI: The majority of the files in this PR are copied test projects to ensure these changes work as expected.

sbglasius (approved):

jamesfredley (approved):

PR #15087, opened 2025-09-19
feature: generate sboms for all published binary jar files

### LINKED ISSUE #14963: Validate Dependency Licenses at build time & generate SBOMs
### Issue description

Per the weekly meeting (8/6) we need to apply the license plugin to ensure we don't have unexpected license changes in our dependencies going forward.  The plugin is already on the build classpath, but it's not currently being applied everywhere.

jdaugherty:
With the recent [compromises of NPM](https://www.sonatype.com/blog/ongoing-npm-software-supply-chain-attack-exposes-new-risks), it's important we publish an SBOM.  

This change generates sboms per the recommended [standard](https://cyclonedx.org/).  The files will not be published individually at this time (publish plugin needs enhancements).  Instead, the files will be published inside of the binary jar files at META-INF/sbom.json.

Other notes on this PR:
* As part of testing the cyclonedx gradle plugin, there are some quality issues with the plugin.  It does not support the latest best practices for gradle.  There is a 3.x version upcoming, but it's still alpha and has other issues (you can't set the component type).
* I discovered we were generating jar files with just license / ASF policy files, so I've stopped generating those files as of this change (grails-dependencies-*,etc)
* I'm not creating an aggregate sbom for the Grails framework.  Right now, an individual sbom for every project will be generated.
* I'm not widely documenting this feature at the start since we'll want to review all of these sboms.  Security software is the main target of these files, so it will find them on it's own.


jdaugherty:
I've gone ahead and updated this PR to also strictly validate licenses based on current versions.  This means that as we update dependencies in the future, we can adjust our exceptions/ mappings and be confident there hasn't been a license change.

jamesfredley (approved):
Great work!

matrei:
I'm getting:
```console
Unpermitted License found for bom dependency: pkg:maven/org.hdrhistogram/HdrHistogram@2.2.2?type=jar in project grails-dependencies-starter-web : CC0-1.0
```
when running `./gradlew clean cyclonedxBom`: https://ge.grails.org/s/y74nwrngrwqtq

jdaugherty:
The starter projects don't generate a jar so they don't run the license check by default.  I guess I could depend on assemble instead ... let me make that change and fix the fallout

jdaugherty:
@matrei I populated most of the other licenses that the ASF considers acceptable from https://www.apache.org/legal/resolved.html & run the validation on assemble as well. This fixes the error you're seeing from the spring actuator project.

matrei:
Do we need to include the license texts in the SBOMs?

jdaugherty:
> Do we need to include the license texts in the SBOMs?

I'm not sure if I understand this - are you proposing we don't include the text or are you asking us to?  

The SPDX code is what identifies the license text.  The cyclonedx is a json standard, so we're simply including what the upstream authors have published.  Most of the licenses are included via a base64 encoding.  We could technically strip the text and ensure the id is present.  I'm inclined to use what the standard does by default since this is meant for security review.

matrei:
> I'm not sure if I understand this - are you proposing we don't include the text or are you asking us to?
> 
> The SPDX code is what identifies the license text. The cyclonedx is a json standard, so we're simply including what the upstream authors have published. Most of the licenses are included via a base64 encoding. We could technically strip the text and ensure the id is present. I'm inclined to use what the standard does by default since this is meant for security review.

I'm wondering if there is any downside to set the `includeLicenseText` property to `false` (https://github.com/CycloneDX/cyclonedx-gradle-plugin?tab=readme-ov-file#configuration-properties).

jdaugherty:
> > I'm not sure if I understand this - are you proposing we don't include the text or are you asking us to?
> 
> > 
> 
> > The SPDX code is what identifies the license text. The cyclonedx is a json standard, so we're simply including what the upstream authors have published. Most of the licenses are included via a base64 encoding. We could technically strip the text and ensure the id is present. I'm inclined to use what the standard does by default since this is meant for security review.
> 
> 
> 
> I'm wondering if there is any downside to set the `includeLicenseText` property to `false` (https://github.com/CycloneDX/cyclonedx-gradle-plugin?tab=readme-ov-file#configuration-properties).

One downside is there isn't an SPDX code for the sitemesh license.  The actual standard only requires id or name though.  https://cyclonedx.org/docs/1.6/json/#tab-pane_components_items_licenses_oneOf_i0_items_license_oneOf_i0

Why would you want to not include the license text?

matrei:
> Why would you want to not include the license text?

I was thinking about jar size.

jdaugherty:
Let me check the sizes but they are zipped so I wasn’t expecting a
significant increase.  Will circle back this afternoon.

jdaugherty:
File sizes for the hibernate-core.jar:
1. has license text: 655K
2. no license text: 588K

I guess with the base64 encoding, zip can't optimize the sizes so it seems reasonable to turn off the text include.

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+
+    def sbomOutputLocation = findProperty('sbomOutputLocation')
+    it.jsonOutput.set(sbomOutputLocation.get())
+    it.outputs.file(sbomOutputLocation)
+
+    // cyclonedx does not support "choosing" the license placed in the sbom
+    // see: https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/16
+    it.doLast {
+        // ordered so that first value is the most preferred, this list is from https://www.apache.org/legal/resolved.html
+        def preferredLicenses = ['Apache-2.0', 'EPL-1.0', 'BSD-3-Clause', 'EPL-2.0', 'MIT', 'MIT-0', '0BSD', 'UPL-1.0', 'CC0-1.0', 'ICU', 'Xnet', 'NCSA', 'W3C', 'Zlib', 'AFL-3.0', 'MS-PL', 'PSF-2.0', 'APAFML', 'BSL-1.0', 'WTFPL', 'Unlicense', 'HPND', 'EPICS', 'TCL']
```
matrei: Multi-line for readability?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+                        'pkg:maven/org.hibernate.common/hibernate-commons-annotations@5.1.2.Final?type=jar': 'LGPL-2.1-only', // hibernate 5 is LGPL, we are migrating to ASF license in hibernate 7
+                        'pkg:maven/org.hibernate/hibernate-core-jakarta@5.6.15.Final?type=jar': 'LGPL-2.1-only', // hibernate 5 is LGPL, we are migrating to ASF license in hibernate 7
+                ],
+                'grails-data-hibernate5-dbmigration': [
+                        'pkg:maven/javax.xml.bind/jaxb-api@2.3.1?type=jar': 'CDDL-1.1', // api export
+                ],
+        ]
+
+        def pickLicense = { String bomRef, List licenseChoices ->
+            if(!bomRef) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+                throw new GradleException("No bomRef found for a dependency of ${project.name}, cannot pick license")
+            }
+
+            logger.info('Picking license for {} from {} choices', bomRef, licenseChoices.size())
+            if(licenseMapping.containsKey(bomRef)) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+                // There are several reasons that cyclone will get the license wrong, usually due to upstream not publishing information or publishing it incorrectly
+                // see the licenseMapping map above for details
+                def licenseId = licenseMapping[bomRef]
+                logger.lifecycle('Forcing license for {} to {}', bomRef, licenseId)
+
+                def licenseBlock = licenses[licenseId]
+                if(!licenseBlock) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+apply plugin: 'org.cyclonedx.bom'
+
+project.ext.setProperty('sbomOutputLocation', project.layout.buildDirectory.file("${findProperty('pomArtifactId') ?: project.name}-${projectVersion}-sbom.json"))
+
+def sbomTask = tasks.named('cyclonedxBom', CycloneDxTask)
+sbomTask.configure { CycloneDxTask it ->
+    // the 2.x version of Cyclonedx uses a legacy syntax & helpers for setting inputs so the syntax below
+    // is required until the 3.x version is GA
+    it.setProjectType(findProperty('sbomProjectType') ?: Component.Type.FRAMEWORK.name())
+    it.@componentName.set(findProperty('pomArtifactId') ?: project.name)
```
matrei: Screenshot_20250923_093025

```groovy
    it.setProjectType(findProperty('sbomProjectType')?.toString() ?: Component.Type.FRAMEWORK.name())
    it.@componentName.set(findProperty('pomArtifactId')?.toString() ?: project.name)
```
or

```groovy
    it.setProjectType((findProperty('sbomProjectType') ?: Component.Type.FRAMEWORK.name()) as String)
    it.@componentName.set((findProperty('pomArtifactId') ?: project.name) as String)
```

will provide the correct type (String) to the methods.

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+
+            def licenseIds = licenseChoices.findAll { it instanceof Map && it.license instanceof Map && it.license.id }
+            def foundLicense = preferredLicenses.find { p -> licenseIds.any { it.license.id == p } }
+            if (foundLicense) {
+                return licenseIds.find { it.license.id == foundLicense }
+            }
+
+            def defaultLicense = licenseChoices[0] // pick the first one found
+            def defaultLicenseId = defaultLicense.license.id as String
+            if(defaultLicenseId == null) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+                throw new GradleException("Could not determine License id for dependency: ${bomRef} in project ${project.name} for value ${defaultLicense}")
+            }
+            if(!(defaultLicenseId in preferredLicenses)) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+                def projectLicenseExemptions = licenseExceptions[project.name] ?: [:]
+                def permittedLicense = projectLicenseExemptions.get(bomRef) == defaultLicenseId
+                if(!permittedLicense) {
```
matrei: Space after `if`?

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        sbomOutputLocation.get().with { rewriteSbom(it.asFile) }
+    }
+}
+
+// for now, we only publish the sbom inside of the binary jar (our bom projects won't have a jar file)
+pluginManager.withPlugin('java') {
+    tasks.named('assemble').configure {
+        dependsOn tasks.named('cyclonedxBom')
```
matrei: Simplify?
```groovy
dependsOn('cyclonedxBom')
```

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    if (!project.findProperty('skipJavaComponent')) {
+        tasks.named('jar', Jar).configure { Jar jar ->
```
matrei: Type is redundant?
```groovy
tasks.named('jar').configure { Jar jar ->
```

gradle/sbom-config.gradle
```diff
@@ -0,0 +1,242 @@
...
+            jar.dependsOn tasks.named('cyclonedxBom')
```
matrei: Simplify?
```groovy
jar.dependsOn('cyclonedxBom')
```

matrei (approved):
Great work James! :tada:

gradle/sbom-config.gradle
jdaugherty: I would have assumed codenarc was applied to our gradle files, but apparently not.  Do you know if it's possible to apply our code style rules to gradle files?

jamesfredley (approved):

PR #15091, opened 2025-09-23
fix(deps): clarify and update usage of `caffeine`

matrei:
Remove `caffeine` dependency from projects where it is unused and add clarifying comments where it it used or necessary for compilation or groovydoc generation.


borinquenkid:
@matrei I think you can change this to a real PR and I will approve it

borinquenkid:
@matrei still working on this?

PR #15095, opened 2025-09-24
Test JDK 25 support

jamesfredley:
Replaces JDK 24 with JDK 25 in the build matrix, JdkVersion enum, and related tests to align with latest LTS and Spring Boot compatibility.


jamesfredley:
Currently several roadblocks exist with Groovy 3 in Gradle 8.14.3 when building grails-core with Java 25

PR #15099, opened 2025-09-25
Update cyclonedx to `2.4.1`

matrei:


gradle/sbom-config.gradle
```diff
@@ -33,83 +33,91 @@ import java.time.temporal.ChronoUnit
...
-project.ext.setProperty('sbomOutputLocation', project.layout.buildDirectory.file("${findProperty('pomArtifactId') ?: project.name}-${projectVersion}-sbom.json"))
+ext {
+    sbomOutputLocation = layout.buildDirectory.file(
+            "${findProperty('pomArtifactId') ?: name}-$projectVersion-sbom.json"
+    )
+}
 
-def sbomTask = tasks.named('cyclonedxBom', CycloneDxTask)
-sbomTask.configure { CycloneDxTask it ->
+tasks.withType(CycloneDxTask).configureEach {
```
jdaugherty, line 42: I thought in 2.4.x they reuse the same task definition and we should only be configuring the cyclonedxBom task name?

gradle.properties
```diff
@@ -47,8 +47,8 @@ yakworksHibernateGroovyProxyVersion=1.1
 # Build dependency versions not managed by BOMs
 apacheRatVersion=0.8.1
 gradleChecksumPluginVersion=1.4.0
-# note: the cyclonedx 3.0.0-alpha-1 still does not set the project correctly, so we must use the older version
-gradleCycloneDxPluginVersion=2.4.0
+# note: the cyclonedx 3 requires Gradle 9
```
jdaugherty, line 50: Is this documented somewhere?  I thought 3.x is intended to work with Gradle 8.x.  The issue is it just didn't generate the correct output.  They only recently updated their project to Gradle 9 (in the last month), but I thought it was still compatible? .

matrei: According to the upgrade guide at:
https://github.com/CycloneDX/cyclonedx-gradle-plugin/releases/tag/cyclonedx-gradle-plugin-3.0.0-alpha-0

> New System Requirements in 3.0.0
> 
>     Java 17 or newer (previously supported older versions)
>     Gradle 9.0 or newer (previously supported Gradle 8.0+)
> 
> The plugin might be still compatible with older versions, but it's not tested.
> 
> ⚠️ Important: If you cannot upgrade to these requirements, continue using version 2.x.x.

gradle/sbom-config.gradle
matrei: It's a bit confusing as the current README is for 3.0.0, which is in alpha.
In version 2.4, I don't think there are multiple different tasks.

jdaugherty: That's where my confusion was, thank you.

jdaugherty (approved):

Issue #15100, opened 2025-09-26
Grails Array/List access performance severely degraded compared to pure groovy or java


val235:
### Expected Behavior

I 've been writing some code that does tight loop based iteration as well as random access in lists and array. My test scripts have been as fast as expected, but i noticed that if the same iteration loops were triggered through a grails controller, they suffered from severe performance degradation.  I would have expected simple looping and array, no matter how its triggered, be blazingly fast given that the JVM optimizes these down at a very low level.

### Actual Behaviour

I've simplified the looping to a very basic example (attached), its just a tight outer loop that iterates 500000 reps, and an inner loop that iterates over a short array of strings, and extracts the element from the array, using basic index based accessor (array[i]) What I found is that exact same groovy code triggered from the controller is magnitudes of order slower than then if the exact same code is triggered from a groovy script. As a second test point, I converted the same loop into a pure java class, and that one runs as fast as expected in all situations.

 For the slow down we are talking 
 - 6 msec for pure java (doesnt matter how its triggered)
 - 10-18 msec for groovy triggered via a script
 - 400-600 mse for groovy triggered via Grails Controller

I've tried various combination of @CompileStatic and @GrailsCompileStatic none make a difference. After profiling I noticed that when running from a controller the hotspot is around various java reflective array accessors. So the Grails compilation or tooling is leaving some invoke-dynamics around arrays that end up using reflection which is much slower and would account for this slow down.

this happens for very basic array operations, instead of using the highly optimized low level operators, they default to reflection based classes

array[i] 
array.length

<img width="1456" height="150" alt="Image" src="https://github.com/user-attachments/assets/94575ff2-ead7-4e86-bb7d-05f768898f20" />

<img width="1684" height="94" alt="Image" src="https://github.com/user-attachments/assets/39513f0e-079b-4944-badf-b1a14671e26a" />

<img width="1587" height="216" alt="Image" src="https://github.com/user-attachments/assets/f80daa81-4ebb-4fec-887b-8807106acf4f" />

### Steps To Reproduce

Im attaching a very basic grails app, generated via Grails App Forge. 

I added 4 files, 
- slow.loop.MainController  -controller that triggers  slow groovy looper and fast java looper and reports the timing
- slow.loop.TinyLoopGroovy.groovy - the groovy based loop that suffers from slowdown when run from controller
- slow.loop.TinyLoopJava.java - the java based version for the groovy looper above, doesnt suffer from any slowdown
[...]

PR #15104, opened 2025-10-01
15100 - remove previous optimization for older JDKs & larger projects

jdaugherty:
Removes the previous tiered compilation configuration that is the cause of the performance issues in #15100 .  I'm documenting the original values so that people can restore them as needed - it was originally only there for larger size projects.


davydotcom (approved):
Thanks this lines up with the previous issue I raised. I do not feel these are necessary settings by default anymore. Especially with groovy4 using indy by default

jamesfredley (approved):

PR #15105, opened 2025-10-01
15079 - grails-micronaut improvements

### LINKED ISSUE #15079: Improved Micronaut Feature for Forge
### Issue description

Since micronaut is now an [optional](https://docs.grails.org/snapshot/guide/upgrading.html#_12_4_micronaut_in_grails_is_now_supported_via_the_plugin_grails_micronaut) dependency in Grails.  We need to add a feature that generates this setup by default. 

The HttpClient should depend on this feature.

jdaugherty:
1. make grails-micronaut work with micronaut-http-client correctly
2. setup Application.groovy so the grails-micronaut plugin works correctly

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
@@ -38,10 +38,10 @@ import groovy.transform.CompileStatic
 @if(applicationType.equals(ApplicationType.PLUGIN) || applicationType.equals(ApplicationType.WEB_PLUGIN)) {
 @@PluginSource
 }
+@@CompileStatic
 @if(features.contains("grails-micronaut")) {
 @@EnableMicronaut
 }
-@@CompileStatic
 class Application extends GrailsAutoConfiguration {
     static void main(String[] args) {
         GrailsApp.run(Application, args)
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
@@ -37,6 +37,7 @@ import grails.boot.config.GrailsAutoConfiguration
 import grails.plugins.metadata.PluginSource
 }
 
+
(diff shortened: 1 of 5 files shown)


jdaugherty:
fixes #15079

jamesfredley (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
```diff
@@ -35,6 +38,9 @@ import groovy.transform.CompileStatic
 @if(applicationType.equals(ApplicationType.PLUGIN) || applicationType.equals(ApplicationType.WEB_PLUGIN)) {
 @@PluginSource
 }
+@if(features.contains("grails-micronaut")) {
+@@EnableMicronaut
+}
```
matrei, line 47: Put below `@CompileStatic`?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
```diff
@@ -25,6 +25,9 @@ under the License.
 
 package @project.getPackageName()
 
+@if(features.contains("grails-micronaut")) {
+import io.micronaut.spring.boot.starter.EnableMicronaut
+}
```
matrei, line 33: Use import groups?
```groovy

import io.micronaut.spring.boot.starter.EnableMicronaut

import grails.boot.GrailsApp
import grails.boot.config.GrailsAutoConfiguration
```

matrei (commented):
I'm a bit concerned about the handling of the Micronaut Platform version.
It is duplicated in `GrailsMicronaut` and `MicronautHttpClient` and there are still versions for `micronaut-http-client` and `micronaut-serde-jackson` in `resources/pom.xml`.

jdaugherty:
> I'm a bit concerned about the handling of the Micronaut Platform version. It is duplicated in `GrailsMicronaut` and `MicronautHttpClient` and there are still versions for `micronaut-http-client` and `micronaut-serde-jackson` in `resources/pom.xml`.

So the platform version is duplicated because I didn't see a way to share / lookup the version based on a configured value.  They can technically diverge over time, but if they do one will likely cause an override of the other.  Do you have a suggestion for how to lookup that value? The goal of this ticket was to get it functioning, and I didn't want to try to invent new concepts in the forge framework.

matrei (approved):

PR #15109, opened 2025-10-03
First Step in optimizing GSP Performance on Grails 7. Safe small optimizations to start

davydotcom:
I have been asked to take a look at Grails GSP Performance. This is an older codebase and I was able to identify some low hanging fruit to start this journey that is pretty safe for release.

There are multiple parts that need to be considered with GSP performance. The differences in cold start performance in development mode as well as production mode rendering.

Cold start involves compiler overhead. This means any changes in groovy4 compiler performance @paulk-asert can have a significant impact in this area. Fortunately, from what I can tell, there does not seem to be a huge difference in performance here in my testing.

I am aware that the compiler (since groovy 3) runs slower on LARGE files. In an effort to reduce file size, I have shorthanded `printHtmlParts(Integer)` to `h(Integer)` when the gsp is first being converted to a `groovy` file. This should help with just scanning and size for the compilation process.

I have also started swapping out commons logging with `SLF4J`. Why you may ask? There are several spots where there are wrapped if statements if `debugEnabled`. This is no longer necessary on `SLF4J` and simply reads better.

I also have started adding javadoc headers to some of the GSP classes so it is easier to follow what is going on in the code. As this is not the easiest to read through.

## Big Optimizations

Whenever a page or partial is rendered, a new instance of `GroovyPage` is created. This is never cached even though 99% of the class is thread safe (except for the OutputStack). We have made changes in the already cached `GroovyPageMetaInfo` class that now instantiates a `ThreadLocal` cache of the page. In addition, the `Constructor` pointer is also cached for concurrent access to reduce reflection overhead. There is an excessive amount of operations that are performed before a page is rendered in the `GroovyPage#initRun()`. I have moved some of this out into `initCommonRun()` that only needs to be run per instance, as opposed to per execution of the pages `run()`. This can be taken farther by a lot but did not want to create high risk issues to start. The other big move was to create the `htmlPartsSet` in the meta info class instead of during construction of each groovy page instance.

Making this change shows approximately 30% in overhead reduction.

### Test Scenarios

**First Scenario: Render 10,000 nested template renders in a `<g:each>` of a row:**

Render Time after a few warmups BEFORE Changes: 200ms avg
Render Time after a few warmups AFTER Changes: 150ms avg

**Second Scenario: Render 10,000 nested template renders using `<g:render collection=''>`**
Render Time after a few warmups BEFORE Changes: 63ms avg
Render Time after a few warmups AFTER Changes: 40ms avg

**Third Scenario: Render LARGE HTML Page (lots of html parts)**

Render Time after a few warmups BEFORE Changes: 2ms avg
Render Time after a few warmups AFTER Changes: 1ms avg 

These Tests were run on an M3 Max Macbook Pro . I would imagine even more gains on different hardware.

### What's next.

We need to aim to further reduce the instantiation logic within `GroovyPage` it's a short lived object that needs to be super quick to instantiate. There are common singleton reference classes that are associated at the start like `TagLibraryLookup` that should be removed.

We need to shorten this pipeline overall. There are a lot more spots I can see that need optimized and will continue to dig farther. The TagLibraryLookup could also stand a rethink, but that is for another day.

I would appreciate any thoughts on this PR and look forward to hearing if this improves overall performance for others.

### DIFF
--- a/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
+++ b/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
@@ -44,7 +44,6 @@ import org.grails.taglib.encoder.OutputContextLookup;
  * @since 0.5
  */
 public class GroovyPageWritable implements Writable {
-    //get SLF4J logger instead
     private static final Logger LOG = LoggerFactory.getLogger(GroovyPageWritable.class);
     private static final String GSP_NONE_CODEC_NAME = "none";
     private GroovyPageMetaInfo metaInfo;
--- a/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
+++ b/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
@@ -26,13 +26,10 @@ import java.io.Reader;
 import java.io.Writer;
 import java.util.LinkedHashMap;
 import java.util.Map;
-
 import groovy.lang.Binding;
 import groovy.lang.Writable;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import grails.util.Environment;
 import org.grails.taglib.AbstractTemplateVariableBinding;
 import org.grails.taglib.TemplateVariableBinding;
@@ -47,7 +44,8 @@ import org.grails.taglib.encoder.OutputContextLookup;
  * @since 0.5
  */
 public class GroovyPageWritable implements Writable {
-    private static final Log LOG = LogFactory.getLog(GroovyPageWritable.class);
+    //get SLF4J logger instead
+    private static final Logger LOG = LoggerFactory.getLogger(GroovyPageWritable.class);
     private static final String GSP_NONE_CODEC_NAME = "none";
     private GroovyPageMetaInfo metaInfo;
     private OutputContextLookup outputContextLookup;
@@ -125,9 +123,7 @@ public class GroovyPageWritable implements Writable {
                 // only try to set content type when evaluating top level GSP
                 boolean contentTypeAlreadySet = outputContext.isContentTypeAlreadySet();
                 if (!contentTypeAlreadySet) {
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("Writing output with content type: " + metaInfo.getContentType());
-                    }
+                    LOG.debug("Writing output with content type: {}", metaInfo.getContentType());
                     outputContext.setContentType(metaInfo.getContentType()); // must come before response.getWriter()
                 }
             }
@@ -139,7 +135,9 @@ public class GroovyPageWritable implements Writable {
 
             GroovyPage page = null;
             try {
-                page = (GroovyPage) metaInfo.getPageClass().newInstance();
+
+                page = metaInfo.getPageClassInstance();
+
             } catch (Exception e) {
                 throw new GroovyPagesException("Problem instantiating page class", e);
             }
@@ -294,4 +292,6 @@ public class GroovyPageWritable implements Writable {
             in.close();
         }
     }
+
+
 }
--- a/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
+++ b/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
@@ -26,11 +26,15 @@ import java.io.Reader;
 import java.io.Writer;
 import java.util.LinkedHashMap;
 import java.util.Map;
+
 import groovy.lang.Binding;
 import groovy.lang.Writable;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import grails.util.Environment;
+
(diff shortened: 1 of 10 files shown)


grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageWritable.java
```diff
@@ -47,7 +44,8 @@
  * @since 0.5
  */
 public class GroovyPageWritable implements Writable {
-    private static final Log LOG = LogFactory.getLog(GroovyPageWritable.class);
+    //get SLF4J logger instead
```
jdaugherty: No reason to use the comment; we should migrate to slf4j over time

codeconsole:
Isn't gsp currently exclusive commons logging? 

Not sure if adding org.slf4j dependencies is a good idea if that is the case. 

Spring Framework is 100% [commons logging](https://repo.maven.apache.org/maven2/org/springframework/spring-core/7.0.0-M9/spring-core-7.0.0-M9.pom)

davydotcom:
This isn't made for spring . It's made for grails

codeconsole:
> This isn't made for spring . It's made for grails

@davydotcom GSP should be independent of Spring and Grails. What is being gained by switching out the logger and requiring another dependency in the classpath?

jdaugherty:
I've reformatted some of the files using IntelliJ to fix the codeStyle errors.  It seems some of this reformat caused parts of the file that were unchanged to be formatted.  This is probably ok, but trying to give a heads up for other users.

jdaugherty:
> > This isn't made for spring . It's made for grails
> 
> @davydotcom GSP should be independent of Spring and Grails. What is being gained by switching out the logger and requiring another dependency in the classpath?

According to Spring, slf4j is a supported logging platform:  https://docs.spring.io/spring-framework/reference/core/spring-jcl.html

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
```diff
@@ -30,29 +30,30 @@ import groovy.transform.CompileStatic
...
+ * These are read at runtime by the {@link org.grails.gsp.GroovyPagesTemplateEngine} class.
  *
  * @author Graeme Rocher
  * @since 1.2
  */
 @CompileStatic
 class GroovyPageCompiler {
 
-    private static final Log LOG = LogFactory.getLog(GroovyPageCompiler)
+    private static final Logger LOG = LoggerFactory.getLogger(GroovyPageCompiler)
```
sbglasius, line 56: Why not just `@Slf4j('LOG')` and you get `log` statements wrapped in `if(log.isDebugenabled())` "for free"

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
```diff
@@ -225,11 +224,11 @@ class GroovyPageCompiler {
                 gpp.generateGsp(gsptarget)
                 gsptarget.flush()
                 // write static html parts to data file (read from classpath at runtime)
-                File htmlDataFile = new File(new File(targetDir, packageDir),  className + GroovyPageMetaInfo.HTML_DATA_POSTFIX)
+                File htmlDataFile = new File(new File(targetDir, packageDir), className + GroovyPageMetaInfo.HTML_DATA_POSTFIX)
```
sbglasius, line 227: Seems redundant to create two `new File(targetDir, packageDir)` here and in line 231

sbglasius (approved):

codeconsole:
> According to Spring, slf4j is a supported logging platform: https://docs.spring.io/spring-framework/reference/core/spring-jcl.html

@jdaugherty  Of course it is. It is bundled with the Spring Boot starter, but why force another dependency on gsp which should be independent of both grails and spring?  what value is being added by switching it other than personal preference?

```
    73737  02-01-1980 00:00   BOOT-INF/lib/commons-logging-1.3.5.jar
   348513  02-01-1980 00:00   BOOT-INF/lib/log4j-api-2.24.3.jar
    23834  02-01-1980 00:00   BOOT-INF/lib/log4j-to-slf4j-2.24.3.jar
   306760  02-01-1980 00:00   BOOT-INF/lib/logback-classic-1.5.18.jar
   627250  02-01-1980 00:00   BOOT-INF/lib/logback-core-1.5.18.jar
     6349  02-01-1980 00:00   BOOT-INF/lib/jul-to-slf4j-2.0.17.jar
    23834  02-01-1980 00:00   BOOT-INF/lib/log4j-to-slf4j-2.24.3.jar
    69908  02-01-1980 00:00   BOOT-INF/lib/slf4j-api-2.0.17.jar
```

sbglasius:
I would like to add, that this all looks good and very well thought through.

davydotcom:
Worth mentioning, this does not break existing GSP compiles. we left the old method names such that the old compiled GSP Files would still function.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
davydotcom: good call out, I have a change local to fix this but given closeness to release its part of some larger changes. can likely wait

PR #15111, opened 2025-10-04
Instant should persist the same way (BsonType.DATE_TIME) as LocalDateTime and Date

codeconsole:
```groovy
class Person {
    LocalDateTime created
    Date updated
    Instant modified
}
```

results in inconsistent mongo storage. 
```
  {
    _id: '68e0824c605bbe8f074c52ba',
    created: ISODate('2025-10-04T02:11:24.201Z'),
    updated: ISODate('2025-10-04T02:11:24.201Z'),
    modified: Long('1759543884201'),
    version: Long('0')
  }
```

Spring Data persists Instant the [same way as Date](https://github.com/spring-projects/spring-data-commons/blob/834368c0526ed410932dea7f731fe441c9888e8f/src/main/java/org/springframework/data/convert/Jsr310Converters.java#L181-L191).  So for mongodb it is stored as a BSON DateTime (displayed as ISODate(...)). That’s the correct and expected mapping.  Grails, however, stores it as a Long.

Aggregation stages or queries that expect date types:
$year, $month, $week, $dayOfWeek
$dateTrunc, $dateAdd, $dateSubtract
$dateToString, $dateFromParts

so none of this works with the current Instant > Long implementation 

Similar to Spring Data, the following should be interchangeable without requiring a database update:
```groovy
    Date created
    LocalDateTime created
    Instant created
```


matrei (commented):
LGTM! Good catch!

However, how will this effect existing applications using `grails-data-mongodb`?
Should we document that the field type in the database must be updated when people upgrade?

jdaugherty (changes requested):
My only concern with this is the upgrade guide was not updated to mention this change.

codeconsole:
> However, how will this effect existing applications using `grails-data-mongodb`? 

@matrei If anyone out there is actually using Instant, it will thrown an exception.  The data must be fixed.  Fortunately, there is an easy mongosh solution which I posted in UPGRADE7.md

> Should we document that the field type in the database must be updated when people upgrade?

done and a mongosh update example was provided.

UPGRADE7.md
```diff
@@ -43,6 +43,15 @@ Experienced while upgrading modules for Grails 7
           }
       }
 
+## Mongo DB
```
jdaugherty: We need to update the ascii doc instructions since thats what is published.  This file is die for removal now that the docs are updated

jdaugherty (approved):

PR #15113, opened 2025-10-04
Include dependencies in POM and publish Gradle module JSON

### LINKED ISSUE #15110: grails-dependencies-* are missing POM dependencies
### Issue description

https://repository.apache.org/content/repositories/snapshots/org/apache/grails/grails-dependencies-assets/

https://repository.apache.org/content/repositories/snapshots/org/apache/grails/grails-dependencies-starter-web/

https://repository.apache.org/content/repositories/snapshots/org/apache/grails/grails-dependencies-test/

jamesfredley:
Set skipJavaComponent to false in assets, starter-web, and test build.gradle files to include Java component configuration in these Grails meta-projects.

The publication must be a proper Java component to publish the Gradle module json file.

Resolves:  https://github.com/apache/grails-core/issues/15110

### DIFF
--- a/grails-dependencies/assets/build.gradle
+++ b/grails-dependencies/assets/build.gradle
@@ -26,7 +26,7 @@ version = projectVersion
 group = 'org.apache.grails'
 
 ext {
-    skipJavaComponent = true
+    skipJavaComponent = false
     pomDescription = 'A meta-project that aggregates all Grails assets files under one place.'
 }
 
--- a/grails-dependencies/starter-web/build.gradle
+++ b/grails-dependencies/starter-web/build.gradle
@@ -26,7 +26,7 @@ version = projectVersion
 group = 'org.apache.grails'
 
 ext {
-    skipJavaComponent = true
+    skipJavaComponent = false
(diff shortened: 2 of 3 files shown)


matrei (approved):
Another option would be to remove the property setting.
But in this case maybe it is better to set the property and add a comment why it is explicitly set to `false`.

matrei (commented):
Now that I think a bit more about this, I'm not sure why the `skipJavaComponent` property was added and if this defeats it's purpose in some other area.

I think @jdaugherty would be able to answer that as he introduced `skipJavaComponent`.

jdaugherty:
So i have been working on another
change to move all of this.  I would prefer we wait to merge this because this is changing code I have already touched.

The core issue is there is a bug where we were publishing empty jar files with just asf policy files.

jdaugherty:
Made obsolete by https://github.com/apache/grails-core/pull/15114

jamesfredley:
https://github.com/apache/grails-core/pull/15114 fully addresses https://github.com/apache/grails-core/issues/15110, so close this PR.

PR #15118, opened 2025-10-06
7.1.x AutoTmestamp Enhancements + grails.importGrailsAnnotations GrailsExtension

codeconsole:
Introduce `@CreatedDate`, `@LastModifiedDate`, `@CreatedBy`, and  `@LastModifiedBy` similar to [Spring Data](https://docs.spring.io/spring-data/jpa/reference/data-commons/auditing.html)

Example App: `./gradlew :grails-test-examples-scaffolding:bootRun`

Introduce grails.importGrailsCommonAnnotations to automatically import common grails annotations and jakarta.validation.constraints.*

build.gradle
```gradle
grails {
    importJavaTime = true
    importGrailsCommonAnnotations = true // kitchen sink
    starImports = ['java.util.concurrent', 'groovy.transform'] // fully customizable
}
```

allows the following with no imports
```groovy
class User {
    String firstName
    String lastName
    String email
    @CreatedDate Instant created
    @LastModifiedDate Instant modified
}

class Car {
    String make
    String model
    @CreatedBy User owner
    @LastModifiedBy User driver
    @CreatedDate Instant purchased
    @LastModifiedDate Instant lastDriven
}
```

```groovy
    @Bean
    AuditorAware<User> springSecurityAuditorAware() {
        return new AuditorAware< User >() {
            @Override
            Optional<String> getCurrentAuditor() {
                Optional.ofNullable(SecurityContextHolder.context)
                        .map { it.authentication }
                        .filter { it?.authenticated }
                        .map { (User) it.principal }
            }
        }
    }
```

```groovy
@Scaffold(GormService<User>)
class UserService implements UserDetailsService {
    @Override
    UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
        return User.findByEmail(username)
    }
}
```

```groovy
import grails.plugin.scaffolding.RestfulServiceController

@Scaffold(RestfulServiceController<User>)
class UserController {}
```


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -226,17 +226,47 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                starImports.add('grails.gorm.annotation')
+            }
+
+            // Check for grails-scaffolding (grails.plugin.scaffolding.annotation.*)
+            def scaffoldingDep = project.configurations.getByName('compileClasspath').dependencies.find { Dependency d ->
+                d.group == 'org.apache.grails' && d.name == 'grails-scaffolding'
+            }
+            if (scaffoldingDep) {
+                starImports.add('grails.plugin.scaffolding.annotation')
+            }
```
sbglasius, line 257: While I understand why you added it while at it, this technically does not have anything to do with time. 
It should probably be in a subsequent PR?

sbglasius (approved):

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: @sbglasius I agree it could of been in a separate PR, but it's kind of a gray area because the added config is `importGrailsAnnotations` and that is just adding 1 annotation `@Scaffold`.  Plus there is a lot I want to address and the latency in getting PRs reviewed is a bit slow at the moment for what I want to get added/fixed.

The motivation for this PR is actually for `importGrailsAnnotations`.    I was originally against 
@matrei  's suggest of having 2 different annotations for AutoTimestamp because of the extra import statement on every domain class, but with `importGrailsAnnotations`, I don't mind. 

 In fact, with `importGrailsAnnotations`, I am even willing to get rid of `@AutoTimestamp` now. @matrei what are your thoughts about getting rid of `@AutoTimestamp`?

jdaugherty:
Per the weekly meeting, we'll merge this into 7.1.  The vote to create that branch is going now - once 72 hrs have passed, we'll create the branch.

jdaugherty:
@codeconsole Can you update this PR now that you've made your mongo update date fixes?

codeconsole:
> @codeconsole Can you update this PR now that you've made your mongo update date fixes?

Done

jdaugherty:
I've updated 7.1.x with the latest 7.0.x changes.  Can you please merge 7.1.x into this PR so the changes are reflected?

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/timestamp/AuditorAware.java
```diff
@@ -0,0 +1,58 @@
...
+ * This will be used to populate @CreatedBy and @LastModifiedBy annotated fields
+ * in domain objects.
+ *
+ * <p>Implementations should be registered as Spring beans. The type parameter
+ * should match the type of the auditor field in your domain classes (e.g., User,
+ * Long, String, etc.).</p>
+ *
+ * <p>Example implementation:</p>
+ * <pre>
+ * &#64;Component
```
jdaugherty: You can remove the html escaping if you wrap the code with `{@code` and `}`.  Unfortunately, you still need the pre to preserve line breaks.

grails-datamapping-validation/src/main/groovy/org/grails/datastore/gorm/validation/constraints/eval/DefaultConstraintEvaluator.java
```diff
@@ -305,6 +306,11 @@ protected boolean isConstrainableProperty(PersistentProperty persistentProperty,
             return NameUtils.isNotConfigurational(propertyName);
         }
         else {
+            // Check if property has @CreatedDate or @LastModifiedDate annotations
+            if (AutoTimestampUtils.hasAutoTimestampAnnotation(persistentProperty)) {
```
jdaugherty: Why can't constraints be added for the auditor type?  It's perfectly reasonable if the auditor is a String, to set a max string (i.e. if you wanted a value greater than 255)

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+     * recognized. In production, the result is cached to avoid repeated reflection calls.</p>
+     *
+     * @param persistentProperty The persistent property to check
+     * @return The auto-timestamp type (CREATED, UPDATED, or NONE)
+     */
+    public static AutoTimestampType getAutoTimestampType(PersistentProperty<?> persistentProperty) {
+        Property mappedForm = persistentProperty.getMapping().getMappedForm();
+
+        // In development mode, always detect fresh to support class reloading
+        if (Environment.isDevelopmentMode()) {
```
jdaugherty: We need another way to enable/disable caching.  GORM was meant to work without grails, and this breaks for non-Grails.

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/CreatedBy.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * A property annotation used to automatically populate a field with the current auditor
+ * upon GORM insert events. The current auditor is retrieved from an {@link org.grails.datastore.gorm.timestamp.AuditorAware}
+ * bean registered in the Spring application context.
+ *
+ * <p>Example usage:</p>
+ * <pre>
+ * class Book {
```
jdaugherty, line 33: See my other comment about `{@code` so we can remove the escaping.

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/LastModifiedBy.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * A property annotation used to automatically populate a field with the current auditor
+ * upon GORM insert and update events. The current auditor is retrieved from an
+ * {@link org.grails.datastore.gorm.timestamp.AuditorAware} bean registered in the Spring application context.
+ *
+ * <p>Example usage:</p>
+ * <pre>
+ * class Book {
```
jdaugherty, line 33: See my other comment about {@code so we can remove the escaping.

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/CreatedDate.java
```diff
@@ -0,0 +1,36 @@
...
+package grails.gorm.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A property annotation used to apply auto-timestamping on a field
+ * upon gorm insert events. This is an alias for @AutoTimestamp(EventType.CREATED).
```
jdaugherty, line 28: Examples exist for the other annotations, why not add one for this one?

grails-datastore-core/build.gradle
```diff
@@ -45,6 +45,9 @@ dependencies {
     implementation platform(project(':grails-bom'))
 
     api project(':grails-common')
+    api 'org.apache.grails.gradle:grails-gradle-model', {
```
jdaugherty: 1. grails-gradle-model includes groovy3, etc and must be excluded. 
2. Environment is a grails concept, but gorm can be used in non-grails applications.  

I think we need another solution here.  This is likely a breaking change.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/Property.groovy
```diff
@@ -137,6 +137,36 @@ class Property implements Cloneable {
...
     private EnumType enumType
+    /**
+     * The auto-timestamp type for this property, cached to avoid repeated reflection calls
+     */
+    AutoTimestampType autoTimestampType
+
+    /**
+     * Enum representing the type of auto-timestamp annotation on a property
+     */
+    static enum AutoTimestampType {
```
jdaugherty: 1. Can we pull this into it's own file?  
2. Is "AutoTimestamp" really a good name for createdBy / updatedBy?  Would Tracked or TrackedEntityType be a better name?

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+ * <p>Supports the following annotations (both GORM and Spring Data variants):</p>
+ * <ul>
+ *   <li>@CreatedDate / @grails.gorm.annotation.CreatedDate - automatically set on insert</li>
+ *   <li>@LastModifiedDate / @grails.gorm.annotation.LastModifiedDate - automatically set on insert and update</li>
+ *   <li>@CreatedBy / @grails.gorm.annotation.CreatedBy - automatically populated with current auditor on insert</li>
+ *   <li>@LastModifiedBy / @grails.gorm.annotation.LastModifiedBy - automatically populated with current auditor on insert and update</li>
+ *   <li>@AutoTimestamp - GORM-specific annotation for backwards compatibility</li>
+ * </ul>
+ *
+ * <p>Caching is automatically disabled in development mode ({@link Environment#isDevelopmentMode()})
```
jdaugherty: In large applications this will significantly slow down development mode.  Is this to support reloading?  It would be better to add that method on the grails plugin for gorm that triggers on reload/refresh and then clear the cache correctly there.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -226,17 +226,50 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+
+        List<String> starImports = []
+
+        // Add java.time if enabled
+        if (grails.importJavaTime) {
+            starImports.add('java.time')
+        }
+
+        // Add Grails annotation packages and common validation annotations if enabled
+        if (grails.importGrailsCommonAnnotations) {
```
jdaugherty, line 239: Instead of hard coding the default imports, can we store these defaults as a list on the GrailsExtension and then use them here if this is set? This makes it clear what's be included to the end user.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -226,17 +226,50 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+            // Always add jakarta.validation.constraints
+            starImports.add('jakarta.validation.constraints')
+
+            // Check for grails-datamapping-core (grails.gorm.annotation.*)
+            def datamappingCoreDep = project.configurations.getByName('compileClasspath').dependencies.find { Dependency d ->
```
jdaugherty, line 244: This is going to run every time and for an app with lots of dependencies this could be a slow down.  Seems like the list based approach may be better and then people can opt in or out of them?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -226,17 +226,50 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
+                d.group == 'org.apache.grails.data' && d.name == 'grails-datamapping-core'
+            }
+            if (datamappingCoreDep) {
+                starImports.add('grails.gorm.annotation')
+            }
+
+            // Check for grails-scaffolding (grails.plugin.scaffolding.annotation.*)
+            def scaffoldingDep = project.configurations.getByName('compileClasspath').dependencies.find { Dependency d ->
```
jdaugherty, line 252: I wouldn't consider scaffolding a default.  I think this needs a separate setting.  My app doesn't actually use scaffolding so this is a problem for me.

jdaugherty (changes requested):
My largest problem is the breaking change with the isDevelopmentMode().  We need a better solution here.  Can you give your reasoning for disabling the caching in development mode?

codeconsole:
> My largest problem is the breaking change with the isDevelopmentMode(). We need a better solution here. Can you give your reasoning for disabling the caching in development mode?

@jdaugherty You are looking at it from the wrong perspective.  I am introducing caching when not in development mode.  The current implementation has no caching.  The performance benefit is so negligible that I suspect you wouldn't even notice.

jdaugherty:
We should just always cache.  We don't need the dependency that way.

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/AutoTimestamp.java
```diff
@@ -29,7 +29,9 @@
  *
  * @author Scott Murphy Heiberg
  * @since 7.0
+ * @deprecated Use {@link CreatedDate} for creation timestamps or {@link LastModifiedDate} for update timestamps instead
  */
+@Deprecated
```
matrei:
```suggestion
@Deprecated(forRemoval = true)
```

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/CreatedBy.java
```diff
@@ -0,0 +1,56 @@
...
+ *     &#64;CreatedBy
+ *     Long creatorId
+ * }
+ * </pre>
+ *
+ * <p>The field type should match the type parameter of your {@link org.grails.datastore.gorm.timestamp.AuditorAware}
+ * implementation (e.g., String, Long, User, etc.).</p>
+ *
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/CreatedDate.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/LastModifiedBy.java
```diff
@@ -0,0 +1,56 @@
...
+ *     &#64;LastModifiedBy
+ *     Long lastModifierId
+ * }
+ * </pre>
+ *
+ * <p>The field type should match the type parameter of your {@link org.grails.datastore.gorm.timestamp.AuditorAware}
+ * implementation (e.g., String, Long, User, etc.).</p>
+ *
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datamapping-core/src/main/groovy/grails/gorm/annotation/LastModifiedDate.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A property annotation used to apply auto-timestamping on a field
+ * upon gorm insert and update events. This is an alias for @AutoTimestamp(EventType.UPDATED).
+ *
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/timestamp/AuditorAware.java
```diff
@@ -0,0 +1,58 @@
...
+ *                 .map(SecurityContext::getAuthentication)
+ *                 .filter(Authentication::isAuthenticated)
+ *                 .map(Authentication::getName);
+ *     }
+ * }
+ * </pre>
+ *
+ * @param <T> the type of the auditor (e.g., User, Long, String)
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+ * to ensure annotation changes are picked up during class reloading.</p>
+ *
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
matrei:
```suggestion
 * @since 7.1
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AuditMetadataUtils.java
```diff
@@ -0,0 +1,202 @@
...
+                                } else {
+                                    return AutoTimestampType.CREATED;
+                                }
+                            }
+                        } catch (Exception e) {
+                            // If we can't read the value, default to CREATED
+                            return AutoTimestampType.CREATED;
+                        }
+                    }
+                }
```
matrei, line 161:
```suggestion
                    switch (annotation.annotationType().getName()) {
                        case CREATED_DATE_ANNOTATION, CREATED_DATE_SPRING_ANNOTATION -> {
                            return AutoTimestampType.CREATED;
                        }
                        case LAST_MODIFIED_DATE_ANNOTATION, LAST_MODIFIED_DATE_SPRING_ANNOTATION -> {
                            return AutoTimestampType.UPDATED;
                        }
                        case CREATED_BY_ANNOTATION, CREATED_BY_SPRING_ANNOTATION -> {
                            return AutoTimestampType.CREATED_BY;
                        }
                        case LAST_MODIFIED_BY_ANNOTATION, LAST_MODIFIED_BY_SPRING_ANNOTATION -> {
                            return AutoTimestampType.UPDATED_BY;
                        }
                        case AUTO_TIMESTAMP_ANNOTATION -> {
                            // For @AutoTimestamp, check the EventType value
                            try {
                                Object eventTypeValue = annotation.annotationType()
                                        .getMethod("value")
                                        .invoke(annotation);

                                if (eventTypeValue != null) {
                                    String eventTypeName = eventTypeValue.toString();
                                    if (eventTypeName.equals("UPDATED")) {
                                        return AutoTimestampType.UPDATED;
                                    } else {
                                        return AutoTimestampType.CREATED;
                                    }
                                }
                            } catch (Exception e) {
                                // If we can't read the value, default to CREATED
                                return AutoTimestampType.CREATED;
                            }
                        }
                    }

```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+
+        return AutoTimestampType.NONE;
+    }
+
+    /**
+     * Checks if a property has any auto-timestamp or auditing annotation.
+     *
+     * @param persistentProperty The persistent property to check
+     * @return true if the property has any supported annotation (@CreatedDate, @LastModifiedDate,
+     *         @CreatedBy, @LastModifiedBy, or @AutoTimestamp) from either GORM or Spring Data
```
matrei:
```suggestion
     * @return true if the property has any supported annotation ({@code @CreatedDate}, {@code @LastModifiedDate},
     *         {@code @CreatedBy}, {@code @LastModifiedBy}, or {@code @AutoTimestamp}) from either GORM or Spring Data
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+     */
+    public static boolean hasAutoTimestampAnnotation(PersistentProperty<?> persistentProperty) {
+        return persistentProperty != null && getAutoTimestampType(persistentProperty) != AutoTimestampType.NONE;
+    }
+
+    /**
+     * Checks if a property has a @CreatedDate annotation (GORM or Spring Data) or @AutoTimestamp(CREATED).
```
matrei:
```suggestion
     * Checks if a property has a {@code @CreatedDate} annotation (GORM or Spring Data) or {@code @AutoTimestamp(CREATED)}.
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @param persistentProperty The persistent property to check
+     * @return true if the property represents a creation timestamp
+     */
+    public static boolean isCreatedTimestamp(PersistentProperty<?> persistentProperty) {
+        return getAutoTimestampType(persistentProperty) == AutoTimestampType.CREATED;
+    }
+
+    /**
+     * Checks if a property has a @LastModifiedDate annotation (GORM or Spring Data) or @AutoTimestamp(UPDATED).
```
matrei:
```suggestion
     * Checks if a property has a {@code @LastModifiedDate} annotation (GORM or Spring Data) or {@code @AutoTimestamp(UPDATED)}.
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @param persistentProperty The persistent property to check
+     * @return true if the property represents an update timestamp
+     */
+    public static boolean isUpdatedTimestamp(PersistentProperty<?> persistentProperty) {
+        return getAutoTimestampType(persistentProperty) == AutoTimestampType.UPDATED;
+    }
+
+    /**
+     * Checks if a property has a @CreatedBy annotation (GORM or Spring Data).
```
matrei:
```suggestion
     * Checks if a property has a {@code @CreatedBy} annotation (GORM or Spring Data).
```

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AutoTimestampUtils.java
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @param persistentProperty The persistent property to check
+     * @return true if the property represents a creation auditor
+     */
+    public static boolean isCreatedBy(PersistentProperty<?> persistentProperty) {
+        return getAutoTimestampType(persistentProperty) == AutoTimestampType.CREATED_BY;
+    }
+
+    /**
+     * Checks if a property has a @LastModifiedBy annotation (GORM or Spring Data).
```
matrei:
```suggestion
     * Checks if a property has a {@code @LastModifiedBy} annotation (GORM or Spring Data).
```

grails-datamapping-validation/src/main/groovy/org/grails/datastore/gorm/validation/constraints/eval/DefaultConstraintEvaluator.java
```diff
@@ -305,6 +306,11 @@ protected boolean isConstrainableProperty(PersistentProperty persistentProperty,
             return NameUtils.isNotConfigurational(propertyName);
         }
         else {
+            // Check if property has @CreatedDate or @LastModifiedDate annotations
```
matrei, line 316: The condition below returns true for all "timestamp" annotations?

grails-datamapping-validation/src/main/groovy/org/grails/datastore/gorm/validation/constraints/eval/DefaultConstraintEvaluator.java
matrei: This is true, as auditor can be different types, and constraints influence the database schema.

grails-fields/grails-app/taglib/grails/plugin/formfields/FormFieldsTagLib.groovy
```diff
@@ -618,7 +618,8 @@ class FormFieldsTagLib {
             }
         } else {
             properties = list ? domainModelService.getListOutputProperties(domainClass) : domainModelService.getInputProperties(domainClass,
-                    exclusionType == ExclusionType.Input ? exclusionsInput : exclusionsDisplay)
+                    exclusionType == ExclusionType.Input ? exclusionsInput : exclusionsDisplay,
+                    exclusionType == ExclusionType.Input)
```
matrei, line 622: `getInputProperties(PersistenProperty, List<String>, boolean)` is not part of the `DomainModelService` interface.

grails-fields/src/main/groovy/org/grails/scaffolding/model/DomainModelServiceImpl.groovy
matrei, line 1: This class is no longer implementing `getInputProperties(PersistentEntity, List)` from the `DomainModelService` interface.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
matrei: > what are your thoughts about getting rid of @AutoTimestamp?

@codeconsole Seems reasonable :+1:

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/timestamp/AuditorAware.java
```diff
@@ -0,0 +1,58 @@
...
+ *                 .map(SecurityContext::getAuthentication)
+ *                 .filter(Authentication::isAuthenticated)
+ *                 .map(Authentication::getName);
+ *     }
+ * }
+ * }</pre>
+ *
+ * @param <T> the type of the auditor (e.g., User, Long, String)
+ * @author Scott Murphy Heiberg
+ * @since 7.0
```
jdaugherty: This is still showing 7.0

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/model/AuditMetadataUtils.java
```diff
@@ -0,0 +1,172 @@
...
+                            } catch (Exception e) {
+                                // If we can't read the value, default to CREATED
+                                return AuditMetadataType.CREATED;
+                            }
+                            break;
+                    }
+                }
+            }
+        } catch (Exception ignored) {
+            // If reflection fails, return NONE
```
jdaugherty, line 164: Would it be useful to add log.debug for these ignored exceptions?

jdaugherty (approved):
I still think it's a bad idea to being walking the dependency graph on every apply.  I won't stop this merge since the default is to not include them though.  

I think the only thing that needs to be done to merge this is to update the documentation to inform people of these features.

codeconsole:
https://github.com/apache/grails-core/pull/15232

PR #15127, opened 2025-10-09
7.0.0-RC2 - PROJECT_TARGET_DIR is never used or set.

codeconsole:
# build is hardcoded in the plugin and results in an exception if you specify any other directory

```
./gradlew bootRun --project-prop buildDir=build-7.0.0-SNAPSHOT 
```

results in 

```
java.io.FileNotFoundException: /grails-demo/build/.grailspid (No such file or directory)
        at java.base/java.io.FileOutputStream.open0(Native Method)
        at java.base/java.io.FileOutputStream.open(FileOutputStream.java:255)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:210)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:171)
        at java.base/java.io.FileWriter.<init>(FileWriter.java:98)
        at grails.util.Environment.writeDevToolsPidFile(Environment.groovy:383)
        at grails.util.Environment.isDevtoolsRestart(Environment.groovy:370)
```

## Motivation - Run 2 Different Versions of Grails on Different Ports with the Same Source
```
./gradlew bootRun --project-prop buildDir=build-7.0.0-RC2 -PgrailsVersion=7.0.0-RC2 --args="--server.port=8081"
./gradlew bootRun --project-prop buildDir=build-7.0.0-SNAPSHOT -PgrailsVersion=7.0.0-SNAPSHOT --args="--server.port=8082"
```


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -427,6 +427,7 @@ class GrailsGradlePlugin extends GroovyPlugin {
     @CompileStatic
     protected String configureGrailsBuildSettings(Project project) {
         System.setProperty(BuildSettings.APP_BASE_DIR, project.projectDir.absolutePath)
+        System.setProperty(BuildSettings.PROJECT_TARGET_DIR, project.layout.buildDirectory.get().asFile.name)
```
jdaugherty, line 430: Going back to April 2024, this never appears to have been set previously.  Why set it now?

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
```diff
@@ -253,7 +253,7 @@ class BuildSettings {
         }
         BASE_DIR = System.getProperty(APP_BASE_DIR) ? new File(System.getProperty(APP_BASE_DIR)) : (IOUtils.findApplicationDirectoryFile() ?: new File('.'))
         GRAILS_APP_DIR_PRESENT = new File(BASE_DIR, 'grails-app').exists() || new File(BASE_DIR, 'Application.groovy').exists()
-        TARGET_DIR = new File(BASE_DIR, 'build')
+        TARGET_DIR = System.getProperty(PROJECT_TARGET_DIR) ? new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR)) : new File(BASE_DIR, 'build')
```
jdaugherty, line 256: What's the use case for having this set? The reason we removed PROJECT_TARGET_DIR is because it wasn't used.  You can easily inject this via gradle, and BuildSettings is really a left over artifact from the original grails build system.  We are trying to remove this functionality over time since it's specific to dev mode and gradle is the better place for that logic.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jdaugherty: I don't believe this is needed because configureForkSettings copies any grails property and sets it on javaexec / test.

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
jdaugherty: I think this is a breaking change too.  Going back to 3/29/24 in the logs, this never overrode TARGET_DIR.

codeconsole: If you set a buildDir via gradle in anything other than build, bootRun fails.
```
./gradlew bootRun --project-prop buildDir=build-7.0.0-RC2 
```

@jdaugherty how is it breaking? if it is not set, it reverts to the way it was previously.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: @jdaugherty the `build` dir should be customizable and not hard coded.

matrei:
I think this is a possible candidate for 7.0.1.
We need to ship 7.0.0 to lay some stable ground so we can continue moving from there.

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
jdaugherty: It's a breaking change b/c it didn't previously behave like this.  We're introducing a function here and I still am not clear why we need to support this.  Why do we need to support a custom build directory? We're trying to remove BuildSettings.groovy completely in the longer run.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jdaugherty: We haven't allowed that to be set since the retirement of the original grails build configuration.  What's the use case for this?  I'm not saying we shouldn't allow this, but for 7.0.0 it's a feature, not a bug.  My concern is scope for 7.0.0 release.

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
codeconsole: @jdaugherty because you shouldn't hard code `build`.  It is not a guaranteed directory name and customizable by gradle.  This doesn't have to go in 7.0.0.   This is just trying to fix `java.io.FileNotFoundException: /grails-demo/build/.grailspid (No such file or directory)` which is caused because build directory is not automatically created.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: @jdaugherty the reason I am calling it a bug is because it results in an exception `java.io.FileNotFoundException: /grails-demo/build/.grailspid (No such file or directory)`
If it works, let's put it in 7.0.1

codeconsole:
@matrei @jdaugherty 
Use case example app: https://github.com/codeconsole/datetime-demo

```
./run.sh          # Starts Grails 7.0.0 on port 8081
./run.sh snapshot # Starts Grails 7.0.1-SNAPSHOT on port 8082
```

jdaugherty (approved):
So until we merge the reloading behavior looking for the pid in the build directory, it seems like we have to support this.  Eventually, we should abstract that configuration on the gradle side instead of pushing it into the grails code itself.  @matrei if you agree, I'm ok merging.

matrei (approved):

PR #15134, opened 2025-10-12
Add kapa.ai AI assistant widget to docs templates

jamesfredley:
The templates are not `DRY`.

Integrated the kapa.ai AI assistant widget into guideItem.html, layout.html, menu.html, and referenceItem.html templates. This provides users with an AI-powered assistant for Apache Grails documentation, including example questions, branding, and consent requirements.

Modeled after https://hudi.apache.org/ integration

Preview Widget: https://demo.kapa.ai/widget/apache-grails


jamesfredley:
image

fernando88to:
This is a really cool initiative. One suggestion is to leave this link in the project's README.MD or in the documentation itself.

jdaugherty (commented):
this is changing the default templates to be very ASF specific- shouldn't we instead put the templates in the grails doc project and not publish them?

matrei (approved):

matrei:
> this is changing the default templates to be very ASF specific- shouldn't we instead put the templates in the grails doc project and not publish them?

Isn't the page tracker already making the templates project specific?

jamesfredley:
@fernando88to Do you think it would be OK to add an `Ask AI` section which links to the current version of docs, which will have the current version of AI Widget?

@jdaugherty @matrei These 4 templates historically had the Grails Google Analytics tracker, which was transitioned to Apache/Matomo with the move to ASF.  I'd like to move forward with these Kapa.ai changes, so they are included on latest and 7.0.x, with the release.  I'm onboard with a follow-up PR which moves these changes, but would still want the templates centralized for guide generation on the projects which live at Apache.

jdaugherty:
@matrei  I pushed back when we changed the tracker too and was told we would address later - now we're making it even worse.  

Several plugins I've converted have used these tasks.  If we're going to continue to customize them, we should just stop publishing the doc helper so we can do our own thing.  

Otherwise, the task takes a template directory (set styleDir).  To make this work, we just have to add folder under grails-docs & then set the `styleDir` on the doc task to that directory.  

I don't think we should move forward unless we do one of these items - stop publishing the docs task or use external templates.  Both of these items are trivial to accomplish. I'm happy to adjust the PR if we're worried about timing.

jdaugherty:
@jamesfredley as for the analytics, I think we were intentionally trying to track third parties in the past.  Now we're not due to ASF policy, so that's why analytics were there previously.

matrei:
> @matrei I pushed back when we changed the tracker too and was told we would address later - now we're making it even worse.
> 
> Several plugins I've converted have used these tasks. If we're going to continue to customize them, we should just stop publishing the doc helper so we can do our own thing.
> 
> Otherwise, the task takes a template directory (set styleDir). To make this work, we just have to add folder under grails-docs & then set the `styleDir` on the doc task to that directory.
> 
> I don't think we should move forward unless we do one of these items - stop publishing the docs task or use external templates. Both of these items are trivial to accomplish. I'm happy to adjust the PR if we're worried about timing.

I understand and I think you are right that it would be nice to have the styleDir configurable. My point was, that the tracker already being in the templates, makes them unusable anyway for other projects so this should not necessarily hold up this PR.

jdaugherty:
We originally agreed we would clean this up after the tracker was added, and we're now making the situation even worse. 

I'll push an update to this PR to make it private then.

jdaugherty (approved):
I've removed grails-doc-core from publishing.  If we really want to publish this, we should be setting the styleDir to specify a different template directory.

fernando88to:
The kapa.ai knowledge base at the link https://grails.apache.org/docs/7.1.1/ is still Grails 7.0.0. Is it possible to update it?

jamesfredley:
It was refreshing weekly, but now it appears it has to be manual. It is ingesting https://grails.apache.org/docs/latest/

PR #15141, opened 2025-10-14
Test Demonstrating Mongo beforeUpdate changes are not persisted

codeconsole:


codeconsole:
The issue is with field access.

Issue #15120, opened 2025-10-08
7.0.0-RC2: @AutoTimestamp and lastUpdated not persisting correctly with Mongo DB


codeconsole:
### Expected Behavior

@AutoTimestamp should update every save that has modified data.

### Actual Behaviour

date is updated on the domain object, but not persisted to the database even when other updated properties are persisted correctly.

### Steps To Reproduce

```
git clone --depth 1 --single-branch https://github.com/codeconsole/mongo-test && cd mongo-test && ./gradlew bootRun
```
Create a new User http://localhost:8080/user/create
Notate Last Updated and Modified dates. 

### Bug 1
Edit the user http://localhost:8080/user/edit/1 and change the name.
Notice only Last Updated is change and NOT Modified.

### Bug 2
Edit the user http://localhost:8080/user/edit/1 and **_DO NOT_** change the name. Just Save 
Notate Last Updated and Modified dates are BOTH modified.  Neither should change because there are no updates.

The behavior with hibernate leaves both dates unchanged.
You can verify by stopping the app and checking out the h2 branch.

### Example Application
https://github.com/codeconsole/mongo-test

### Version

7.0.0-RC2

codeconsole:
BUG 2, marks dirty even if it is the only property that is dirty

https://github.com/apache/grails-core/blob/9434fd146ad3baa844365cbba922c1293405fa2e/grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/codecs/PersistentEntityCodec.groovy#L254-L256

codeconsole:
Strange how hasChanged() is true when $changedProperties is null.  There are situations where trackChanges has yet to be called.

https://github.com/apache/grails-core/blob/9434fd146ad3baa844365cbba922c1293405fa2e/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckable.groovy#L72-L79

codeconsole:
https://github.com/apache/grails-core/pull/15118

codeconsole:
Closed with https://github.com/apache/grails-core/pull/15122

PR #15143, opened 2025-10-14
Take snapshot prior to mongo update then mark any changes during beforeUpdate and update events as dirty.

codeconsole:
Hibernate overrides all dirty checking in the GormInstanceApi

https://github.com/apache/grails-core/blob/7.0.x/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/HibernateGormInstanceApi.groovy#L77-L80

which uses hibernate's api for comparing against an internal snapshot.

Currently mongodb does not persist any updates during Events and AutoTimestamp events because the changes are not marked dirty and must be manually marked dirty.

https://grails.apache.org/docs/latest/grails-data/hibernate5/manual/index.html#eventsAutoTimestamping

This fix takes a snapshot before all events and then marks any changes as dirty props to persistence.

### DIFF
--- a/grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/BeforeUpdatePropertyPersistenceSpec.groovy
+++ b/grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/BeforeUpdatePropertyPersistenceSpec.groovy
@@ -34,7 +34,8 @@ class BeforeUpdatePropertyPersistenceSpec extends GrailsDataTckSpec<GrailsDataMo
     void setupSpec() {
         manager.domainClasses.addAll([UserWithBeforeUpdate, UserWithBeforeUpdateAndAutoTimestamp])
     }
-    
+
+    @Issue('GRAILS-15139')
     void "Test that properties set in beforeUpdate are persisted"() {
         given: "A user is created"
         def user = new UserWithBeforeUpdate(name: "Fred")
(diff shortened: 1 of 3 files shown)


grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/BeforeUpdatePropertyPersistenceSpec.groovy
```diff
@@ -119,8 +116,7 @@ class BeforeUpdatePropertyPersistenceSpec extends GrailsDataTckSpec<GrailsDataMo
         user.random.length() == 5
     }
 
-    @PendingFeature
-    @Issue('GPMONGODB-XXX')
+    @Issue('GRAILS-15120')
```
matrei, line 120: Issue #15120 reports that `lastUpdated` and `@AutoTimestamp` where both not working.
But as I understand it, `lastUpdated` was actually working but not `@AutoTimestamp`?
Can we also create a test with a property that has `@AutoTimestamp`?

codeconsole: @matrei I added @AutoTimestamp Date modified to the test

matrei (approved):
Thank you!

jdaugherty (approved):

jdaugherty:
Thank you!

Issue #15147, opened 2025-10-16
7.0.0-RC2 & 7.0.0 BUG - ControllerActionTransformer - GroovyCastException binding objects to controller method arguments. 


codeconsole:
[This commit](https://github.com/apache/grails-core/commit/7c56295d6beb8f731dbabf310acf59aaf9b7788e) introduced biding exceptions on several methods in my Controller, but not all of them.

# Summary
Inconstant binding resulting in GroovyCastExceptions

```groovy
class ContestController {
    def dq(long id, long imageId) {}  // WORKS FINE!
    def submit(long id, long images) {} // GroovyCastException: Cannot cast object '1000311' with class 'java.lang.String' to class 'long'
    def prepareQuestionnaire(Long id) { // GroovyCastException: Cannot cast object '1000271' with class 'java.lang.String' to class 'java.lang.Long'
}
```

this worked fine until August 7th

```groovy
class ContestController {
    def submit(long images) { params.long('id')} 
}
```

results in 

```
2025-10-16T10:15:25.907-07:00 ERROR 69893 --- [cat-handler-402] 
o.g.web.errors.GrailsExceptionResolver   : Cannot cast object '1000311' with class 'java.lang.String' to class 'long'
org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object '1000311' with class 'java.lang.String' to class 
'long'
    at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104) 
```

which is weird because there is not a complete stack trace.

Workaround:
```groovy
class ContestController {
    def submit(long id, long images) {}
}
```

_No response_

### Example Application

I am unable to replicate this in an example app and there are other similar situations in my app that work fine.

### Version

7.0.0

codeconsole:
Another occurrence
```groovy
"/$controller/$action?/$id?(.$format)?"
```
```groovy
def prepareQuestionnaire(Long id) {
// exception occurs before method execution 
}
```
```
org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object '1000271' with class 'java.lang.String' to class 'java.lang.Long'
at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache ( org/codehaus.groovy.vmplugin.v8/IndyInterface.java:321 )
```

codeconsole:
[...]

PR #15152, opened 2025-10-18
Revert "fix: #14947 - class cast exception due to not using local variable

codeconsole:
This reverts commit 7c56295d6beb8f731dbabf310acf59aaf9b7788e.

which causes 

https://github.com/apache/grails-core/issues/15147


jamesfredley:
Can you confirm that the revert reintroduces this bug:  https://github.com/apache/grails-core/issues/14947#issue-3285500647

jdaugherty:
We should add a test for this instead of reverting and see if Paul can help instead.

codeconsole:
Superceded by https://github.com/apache/grails-core/pull/15153

Issue #15164, opened 2025-10-22
Change g:format to accept date, datetime and time


jamesfredley:
### Issue description

Change g:format to take date, datetime and time, instead of single date format

codeconsole:
@jamesfredley it already does that but the default behavior is to render:

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy#L187

otherwise there is a type parameter for date, time, and anything else 

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy#L193-L201

codeconsole:
https://github.com/apache/grails-core/pull/15146

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-fields/grails-app/taglib/grails/plugin/formfields/FormFieldsTagLib.groovy#L949-L973

https://github.com/apache/grails-core/pull/15118 - introduces second precision on date picker and time-stamping

Date picker should also be updated to support time only as well as optional millisecond and nanosecond precision for lossless date updating.

jamesfredley:
@codeconsole @jdaugherty I created this issue during the discussion in the weekly meting today, but it's sounds like we might already have what we need.

codeconsole:
@jamesfredley yeah, I know, I am just adding context because we need to discuss it more.  I still don't think we have a final solution and I am looking for everyones input. I want this issue to start the discussion.

This PR shows what everything looks like using DATETIME
https://github.com/apache/grails-core/pull/15165

codeconsole:
```yml
grails:
  databinding:
    dateFormats: # used when parsing posted form fields
      - "yyyy-MM-dd'T'HH:mm:ss.SSSX"
      - "yyyy-MM-dd'T'hh:mm:ss'Z'"
      - "yyyy-MM-dd HH:mm:ss.S z"
      - "yyyy-MM-dd HH:mm:ss.S"
      - "yyyy-MM-dd'T'HH:mm:ssZ"
      - "yyyy-MM-dd'T'HH:mm:ss"
      - "yyyy-MM-dd HH:mm:ss"
      - "yyyy-MM-dd HH:mm"
      - "yyyy-MM-dd"
[...]

PR #15165, opened 2025-10-22
default fields dates to datetime precision

codeconsole:
Alternative solution for `<f:all` rendering of date objects with time precision.

### NOTE: This proposal puts rendering IN SYNC with create/edit input tag rendering.  Currently the fields plugin ONLY supports MINUTE precision for input rendering using the date picker.

The documentation for [datePicker](https://grails.apache.org/docs/snapshot/ref/Tags%20-%20GSP/datePicker.html) is incorrect.  "and second of the day."  Seconds have never been supported, but are set to be added in with https://github.com/apache/grails-core/pull/15118

### Show DateTime - Proposal (contingent on Locale)

| Type              | Value              |
|--------------------|--------------------|
| Calendar           | 10/8/25, 12:48 AM  |
| Instant            | 10/8/25, 12:48 AM  |
| Offset Date Time   | 10/8/25, 12:48 AM  |
| Zoned Date Time    | 10/8/25, 12:48 AM  |
| Date               | 10/8/25, 12:48 AM  |
| SQL Date           | 10/8/25            |
| Time               | 12:48 AM           |
| Local Time         | 12:48 AM           |
| Timestamp          | 10/8/25, 12:48 AM  |
| Local Date         | 10/8/25            |
| Local Date Time    | 10/8/25, 12:48 AM  |

### Show DateTime - Current Behavior

| Type              | Value                     |
|--------------------|---------------------------|
| Calendar           | 2025-10-08 00:48:46 PDT  |
| Instant            | 2025-10-08 00:48:46 PDT  |
| Offset Date Time   | 2025-10-08 00:48:46 PDT  |
| Zoned Date Time    | 2025-10-08 00:48:46 PDT  |
| Date               | 2025-10-08 00:48:46 PDT  |
| SQL Date           | 10/8/25                  |
| Time               | 12:48 AM                 |
| Local Time         | 12:48 AM                 |
| Timestamp          | 2025-10-08 00:48:46 PDT  |
| Local Date         | 10/8/25                  |
| Local Date Time    | 2025-10-08 00:48:46 PDT  |

### DIFF
--- a/grails-fields/grails-app/taglib/grails/plugin/formfields/FormFieldsTagLib.groovy
+++ b/grails-fields/grails-app/taglib/grails/plugin/formfields/FormFieldsTagLib.groovy
@@ -966,7 +966,7 @@ class FormFieldsTagLib {
             case Instant:
             case ZonedDateTime:
             case OffsetDateTime:
-                g.formatDate(date: model.value)
+                g.formatDate(date: model.value, type: 'DATETIME')
                 break
             default:
                 g.fieldValue(bean: model.bean, field: model.property)
--- a/grails-fields/src/test/groovy/grails/plugin/formfields/DisplayWidgetSpec.groovy
+++ b/grails-fields/src/test/groovy/grails/plugin/formfields/DisplayWidgetSpec.groovy
@@ -38,12 +38,12 @@ class DisplayWidgetSpec extends AbstractFormFieldsTagLibSpec implements TagLibUn
 
 	void 'f:displayWidget without template and a date value renders the formatted date'() {
 		expect:
-		applyTemplate('<f:displayWidget bean="personInstance" property="dateOfBirth"/>', [personInstance: personInstance]) == applyTemplate('<g:formatDate date="${personInstance.dateOfBirth}"/>', [personInstance: personInstance])
+		applyTemplate('<f:displayWidget bean="personInstance" property="dateOfBirth"/>', [personInstance: personInstance]) == applyTemplate('<g:formatDate date="${personInstance.dateOfBirth}" type="DATETIME"/>', [personInstance: personInstance])
 	}
 
 	void 'f:displayWidget without template and an instant value renders the formatted date'() {
 		expect:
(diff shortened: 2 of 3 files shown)


sbglasius (approved):

matrei:
So, if I understand this correctly, before this change, the `date.format` and `default.date.format` message properties were used to decide the formatting. Now, it will always use `FormatStyle.SHORT` for both date and time + timezone.

Also, should all these types be formatted with timezone?

codeconsole:
> So, if I understand this correctly, before this change, the `date.format` and `default.date.format` message properties were used to decide the formatting. Now, it will always use `FormatStyle.SHORT` for both date and time + timezone.
> 
> Also, should all these types be formatted with timezone?

@matrei this ticket will be for coming up with a final solution
https://github.com/apache/grails-core/issues/15164

I described the current behavior there:

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-fields/grails-app/taglib/grails/plugin/formfields/FormFieldsTagLib.groovy#L949-L973 


currently `g.formatDate(date: model.value)` is called which result in no type so that results in a default format:

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy#L187

all this ticket does is set the type to datetime which results in the following behavior;

https://github.com/apache/grails-core/blob/c27742f9ad64b361ad39dabe9e090e938dc256ca/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy#L193-L201

matrei:
I'm OK with this being a two-step solution to not break backwards-compatibility, but I don't think we should output timezone for types that do not hold that information.

How would that be able to render the stored value correctly? For example, a user in timezone `CEST` enters a value in a `LocalDateTime` field. The `LocalDateTime` type does not store timezone information. The server that renders this value will use `ZoneId.systemDefault()` and output for example `PDT` which is a completely different time than what the user has entered.

matrei:
I might have misread your tables above.
However, I don't seem to be getting the output formatting from the first table when applying this patch locally.

matrei (approved):
I have tried out this change and it seems to be a real improvement to the current rendering behavior.

Thanks @codeconsole!

PR #15179, opened 2025-10-27
use stripped PROJECT_TARGET_DIR for setting TARGET_DIR

codeconsole:
`grails.` is stripped out of all build settings when the jvm is forked.

This correctly sets the TARGET_DIR after it has been stripped by using `project.target.dir` 

### Difference from 7.0.0
```groovy
TARGET_DIR = new File(BASE_DIR, 'build')
```
to
```groovy
TARGET_DIR = new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR.substring(7)) ?: 'build')
```

### DIFF
--- a/grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
+++ b/grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
@@ -253,7 +253,7 @@ class BuildSettings {
         }
         BASE_DIR = System.getProperty(APP_BASE_DIR) ? new File(System.getProperty(APP_BASE_DIR)) : (IOUtils.findApplicationDirectoryFile() ?: new File('.'))
         GRAILS_APP_DIR_PRESENT = new File(BASE_DIR, 'grails-app').exists() || new File(BASE_DIR, 'Application.groovy').exists()
-        TARGET_DIR = System.getProperty(PROJECT_TARGET_DIR) ? new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR)) : new File(BASE_DIR, 'build')
+        TARGET_DIR = new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR.substring(7)) ?: 'build')
         RESOURCES_DIR = !GRAILS_APP_DIR_PRESENT ? null : (System.getProperty(PROJECT_RESOURCES_DIR) ? new File(System.getProperty(PROJECT_RESOURCES_DIR)) : new File(TARGET_DIR, 'resources/main'))
     }
 }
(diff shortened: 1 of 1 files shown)


codeconsole:
https://github.com/apache/grails-core/pull/15127

matrei:
Why is `grails.` stripped in the first place in `configureForSettings`?

The message of the commit that added that code in 2015 is:
`Infrastructure support to allow changing server port and enable HTTPS from command line`

That implies to me that this is done to convert `grails.server.port` -> `server.port` and `grails.server.port.https` -> `server.port.https` and possibly some other properties.

Would it not be more transparent to explicitly convert the Spring Boot properties that we care about in `configureForkSettings` instead of bulk converting every system property that starts with `grails.`?

codeconsole:
> Why is `grails.` stripped in the first place in `configureForSettings`?
> 
> The message of the commit that added that code in 2015 is: `Infrastructure support to allow changing server port and enable HTTPS from command line`
> 
> That implies to me that this is done to convert `grails.server.port` -> `server.port` and `grails.server.port.https` -> `server.port.https` and possibly some other properties.
> 

@matrei this PR isn't stripping anything.  All this PR does is reference the historically correct `target.dir`.  The plugin has always stripped them. I started down the path of not stripping them, but since it has always been like that, it might be beyond the scope for 7.x

> Would it not be more transparent to explicitly convert the Spring Boot properties that we care about in configureForkSettings instead of bulk converting every system property that starts with grails.?

yes, I agree with everything you are saying, but we should prob not be that invasive until 8.x, no?

codeconsole:
An alternative for this PR is not reference the constant at all.
```gradle
TARGET_DIR = new File(BASE_DIR, System.getProperty('target.dir') ?: 'build')
```

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
```diff
@@ -253,7 +253,7 @@ class BuildSettings {
         }
         BASE_DIR = System.getProperty(APP_BASE_DIR) ? new File(System.getProperty(APP_BASE_DIR)) : (IOUtils.findApplicationDirectoryFile() ?: new File('.'))
         GRAILS_APP_DIR_PRESENT = new File(BASE_DIR, 'grails-app').exists() || new File(BASE_DIR, 'Application.groovy').exists()
-        TARGET_DIR = System.getProperty(PROJECT_TARGET_DIR) ? new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR)) : new File(BASE_DIR, 'build')
+        TARGET_DIR = new File(BASE_DIR, System.getProperty(PROJECT_TARGET_DIR.substring(7)) ?: 'build')
```
matrei: I agree with your comment, in this case it's better to not use the constant.
Also `System.getProperty` has a version that takes a default value:
```groovy
System.getProperty('project.target.dir', 'build')
```

matrei (approved):

grails-gradle/model/src/main/groovy/grails/util/BuildSettings.groovy
codeconsole: yeah, that's what I meant.

PR #14737, opened 2025-05-15
Groovy 5 compatibility

matrei:
This is the start of checking compatibility with Groovy 5 and eventually adding a CI workflow for Groovy 5.0.0-SNAPSHOT.

Superseded by https://github.com/apache/grails-core/pull/15183

PR #15183, opened 2025-10-29
Groovy 5 Compatibility

matrei:
Supersedes #14737


jamesfredley:
https://github.com/apache/grails-core/commit/9574fe8f7f6a1d230b895c8b1df76fd5b6860df9 - should be a comprehensive list of the remaining Groovy 5 issues which need to be addressed, not necessary the solutions.

jdaugherty:
> [9574fe8](https://github.com/apache/grails-core/commit/9574fe8f7f6a1d230b895c8b1df76fd5b6860df9) - should be a comprehensive list of the remaining Groovy 5 issues which need to be addressed, not necessary the solutions.

It seems the way we solved this was making it dynamically resolve, while before it was compilestatic.  These don't seem like real solutions - the developer impact of method resolution is going to be very painful.

jdaugherty:
@jamesfredley we should recreate this PR so your change shows correctly in the diff history.

jamesfredley:
### Why GormEntity Works but HibernateEntity Didn't
The issue is **NOT** that Groovy 5 prohibits static methods in traits. `GormEntity` has 87+ static methods and works fine.
The issue is **joint compilation with Java files that directly import the trait**.
| Trait | Module | Static Methods | Java Files Import It? | Result |
|-------|--------|----------------|----------------------|--------|
| `GormEntity` | grails-datamapping-core | 87 | **No** | Works |
| `HibernateEntity` | grails-data-hibernate5-core | 5 | **Yes** (`HibernateMappingContext.java`) | Fails |
| `MongoEntity` | grails-data-mongodb | 18 | No | Works |
| `Neo4jEntity` | grails-data-neo4j | 10 | No | Works |
### The Technical Problem
When a **Java file in the same module** imports a Groovy trait with static methods that return generic type parameters (`D`), the Groovy stub generator creates invalid Java code:
```java
// Generated stub (INVALID Java)
@groovy.transform.Generated() static java.util.List<D> findAllWithSql(java.lang.CharSequence sql);
//                                                   ^ ERROR: non-static type variable D 
//                                                            cannot be referenced from static context
```

In Java, you cannot use a class-level type parameter (`D`) in a static method signature. The Groovy trait works because Groovy handles this differently at runtime, but the Java stub generator doesn't account for this.

---
## Protential Solution
### Approach: Remove Java Import via Reflection
Instead of removing the static methods from the trait, we modified `HibernateMappingContext.java` to load `HibernateEntity` via reflection, avoiding the need for Java stubs:

Updated commit with this change: https://github.com/apache/grails-core/commit/425d2da3b633fa471ad64dbb21a92c1327a5268b

### Approach: move class locations to other modules or change Java Class to Groovy Class



@jdaugherty these commits are not indented to be part of the real PR they are just a map for the real commit, it identifies everything that breaks trying to build.

jdaugherty:
@matrei  should we close this in favor of https://github.com/apache/grails-core/pull/15557 ?

jamesfredley:
All commits from this PR exist in https://github.com/apache/grails-core/pull/15557

Issue #15193, opened 2025-10-31
"" Spring Profile results in no configuration being loaded


jdaugherty:
### Issue description

This code was added in Grails 7: https://github.com/apache/grails-core/blob/7f9a2c0fbc8bc27e23a53d39514bc648f54133e7/grails-bootstrap/src/main/groovy/org/grails/config/NavigableMap.groovy#L190

The problem is boot run doesn't always define a spring profile which results in no configuration being loaded.

PR #15194, opened 2025-10-31
fix - #15193 fix empty spring profile causing configuration to not load

jdaugherty:
This bug is preventing my application from loading any configuration in development mode because `spring.profiles.active` is not defined.  I'm specifically failing on loading of an `application.groovy` file without an environment specific configuration.

### DIFF
--- a/grails-bootstrap/src/main/groovy/org/grails/config/NavigableMap.groovy
+++ b/grails-bootstrap/src/main/groovy/org/grails/config/NavigableMap.groovy
@@ -18,8 +18,6 @@
  */
 package org.grails.config
 
-import java.util.regex.Pattern
-
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import groovy.transform.EqualsAndHashCode
@@ -37,15 +35,6 @@ import org.slf4j.LoggerFactory
 class NavigableMap implements Map<String, Object>, Cloneable {
 
     private static final Logger LOG = LoggerFactory.getLogger(NavigableMap)
-
-    private static final Pattern SPLIT_PATTERN = ~/\./
-    private static final String SPRING_PROFILES = 'spring.profiles.active'
-    private static final String SPRING = 'spring'
-    private static final String CONFIG = 'config'
-    private static final String ACTIVATE = 'activate'
-    private static final String ON_PROFILE = 'on-profile'
-    private static final String PROFILES = 'profiles'
-    private static final String SUBSCRIPT_REGEX = /((.*)\[(\d+)\]).*/
 
     final NavigableMap rootConfig
     final List<String> path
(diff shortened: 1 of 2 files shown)


jdaugherty:
@matrei some of the earlier reports of application.groovy not loading configuration is likely due to this bug.

jdaugherty:
Fixes #15193

grails-bootstrap/src/main/groovy/org/grails/config/NavigableMap.groovy
```diff
@@ -187,7 +189,7 @@ class NavigableMap implements Map<String, Object>, Cloneable {
     private static boolean springProfileExclude(Map sourceMap, String path) {
 
         // Is there an active Spring profile?
-        def activeSpringProfile = System.getProperty(SPRING_PROFILES)
+        def activeSpringProfile = System.getProperty(SPRING_PROFILES) ?: Environment.current.name
```
jdaugherty: Alternatively to this change, we could change line 228.  It seemed most reasonable to default the environment name instead

sbglasius (approved):

matrei:
I think it might make more sense to change the condition on line 228 instead as `springActiveProfile` will always be `true` with this change.

So instead of defaulting `springActiveProfile` to `Environment.current.name`, would it work with?:
```groovy
if (!sourceMapProfile1 && !sourceMapProfile2) {
```
on line 228?

jdaugherty:
> I think it might make more sense to change the condition on line 228 instead as `springActiveProfile` will always be `true` with this change.
> 
> 
> 
> So instead of defaulting `springActiveProfile` to `Environment.current.name`, would it work with?:
> 
> ```groovy
> 
> if (!sourceMapProfile1 && !sourceMapProfile2) {
> 
> ```
> 
> on line 228?

Well, I thought that originally too but if we run boot run its development mode but no profile is set.  Shouldn't dev only config apply in that case?  Is it valid to not have a profile?

matrei:
@jdaugherty Does this problem also exist in any of the functional test apps?
Otherwise, could you replicate the problem in a minimal app?
It's not clear to me what's going on.

jdaugherty:
> @jdaugherty Does this problem also exist in any of the functional test apps?
> 
> Otherwise, could you replicate the problem in a minimal app?
> 
> It's not clear to me what's going on.

It won't ever exist in a test because the tests explicitly run with a test profile.

matrei:
Yes, but you can start the functional-test apps with bootRun.

jdaugherty:
How?  The test annotations explicitly set the profile. 

My original question is still unanswered: is it valid to not have a profile? Also, shouldn't the profile always be equal to the grails environment?

jdaugherty:
@matrei I misread your original comment, in terms of reproducing with the functional test apps by using their bootRun, let me get back to you.

jdaugherty:
@matrei  

To reproduce with a functional test app in `grails-core`, you can test with `issue-11102`.  The issue is earlier versions of grails recommended this customization to bootRun: 

        String springProfilesActive = 'spring.profiles.active'
        systemProperty springProfilesActive, System.getProperty(springProfilesActive)

Which maps the profile to an empty string, instead of null.  I've updated the PR to use groovy truth to address this scenario.

jdaugherty:
@matrei I merged your proposed change.  thank you for taking a look at this.  I assume you're ok with me merging this then?

matrei (approved):

PR #15195, opened 2025-10-31
7.0.x Bug - Don't call a resource constructor until GORM has been initialized by Spring

codeconsole:
# The Problem

```
Caused by: java.lang.IllegalStateException: org.springframework.boot.web.servlet.context.AnnotationConfigServletWebServerApplicationContext@48300ca8 has not been refreshed yet
        at org.springframework.context.support.AbstractApplicationContext.assertBeanFactoryActive(AbstractApplicationContext.java:1262)
        at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1280)
        at org.grails.web.context.ServletEnvironmentGrailsApplicationDiscoveryStrategy.findGrailsApplication(ServletEnvironmentGrailsApplicationDiscoveryStrategy.groovy:54)
        at grails.util.Holders.findApplication(Holders.java:126)
        at org.grails.plugins.web.controllers.api.ControllersDomainBindingApi.autowire(ControllersDomainBindingApi.java:90)
        at org.grails.plugins.web.controllers.api.ControllersDomainBindingApi.initialize(ControllersDomainBindingApi.java:50)
        at com.pixoto.community.CategoryParticipant.<init>(CategoryParticipant.groovy)
        at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
        ... 43 common frames omitted
```

This is only an issue during a Spring Dev Tools restart, but this fixes the issue.

``` groovy
  this.resource = resource.getDeclaredConstructor().newInstance() as GormAllOperations<T>
```
results in a domain object instantiation during the constructor, which:
  1. Triggers Grails GORM enhancements
  2. Calls ControllersDomainBindingApi.initialize()
  3. Tries to access Holders.findApplication()
  4. Fails during DevTools restart when ApplicationContext is being switched

# The Fix
Lazy initialize the resource

### DIFF
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
@@ -35,7 +35,7 @@ import org.grails.datastore.gorm.GormEntityApi
 @CompileStatic
 class GormService<T extends GormEntity<T>> {
 
-    private GormAllOperations<T> gormStaticApi
+    GormAllOperations<T> gormStaticApi
     Class<T> resource
     String resourceName
     String resourceClassName
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
@@ -35,7 +35,8 @@ import org.grails.datastore.gorm.GormEntityApi
 @CompileStatic
 class GormService<T extends GormEntity<T>> {
 
-    GormAllOperations<T> gormStaticApi
+    @Lazy
+    GormAllOperations<T> gormStaticApi = GormEnhancer.findStaticApi(resource) as GormAllOperations<T>
     Class<T> resource
     String resourceName
     String resourceClassName
@@ -48,24 +49,16 @@ class GormService<T extends GormEntity<T>> {
         resourceName = GrailsNameUtils.getPropertyName(resource)
     }
 
-    protected GormAllOperations<T> getGormStaticApi() {
-        if (gormStaticApi == null) {
-            // Lazy initialization - happens on first access when GORM is ready
-            gormStaticApi = GormEnhancer.findStaticApi(resource) as GormAllOperations<T>
-        }
-        return gormStaticApi
-    }
-
     T get(Serializable id) {
-        getGormStaticApi().get(id)
+        gormStaticApi.get(id)
     }
 
     List<T> list(Map args) {
-        getGormStaticApi().list(args)
+        gormStaticApi.list(args)
     }
 
     Long count(Map args) {
-        getGormStaticApi().count()
+        gormStaticApi.count()
     }
 
     @Transactional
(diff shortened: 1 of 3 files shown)


codeconsole:
@matrei if you recall, we did this `this.resource = resource.getDeclaredConstructor().newInstance() as GormAllOperations<T>` so we could `@CompileStatic`

but I think it is best to keep resource a `Class<T>` for 2 reasons:
1. Simplify the `DomainServiceLocator` lookup
2. Keep in line with  https://github.com/apache/grails-core/blob/7f9a2c0fbc8bc27e23a53d39514bc648f54133e7/grails-rest-transforms/src/main/groovy/grails/rest/RestfulController.groovy#L46-L47

Also, I found it unnecessary to instantiate an instance if we separate `GormAllOperations<T>` and `Class<T>`

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
```diff
@@ -34,28 +35,37 @@ import org.grails.datastore.gorm.GormEntityApi
 @CompileStatic
 class GormService<T extends GormEntity<T>> {
 
-    GormAllOperations<T> resource
+    private GormAllOperations<T> gormStaticApi
```
jdaugherty: This field is now private, while it was public before.  This is a breaking change for a non-major release ...

codeconsole: @jdaugherty this is something that is not documented that nobody uses, but I have no problem removing private.  done.

matrei: From what I can tell, there was no `gormStaticApi` property before this change, so it should be `private` as it is lazily instantiated in the getter.

However, the `resource` property has changed type, which is breaking.

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
```diff
@@ -34,28 +35,37 @@ import org.grails.datastore.gorm.GormEntityApi
...
         resourceName = GrailsNameUtils.getPropertyName(resource)
     }
 
+    protected GormAllOperations<T> getGormStaticApi() {
+        if (gormStaticApi == null) {
+            // Lazy initialization - happens on first access when GORM is ready
+            gormStaticApi = GormEnhancer.findStaticApi(resource) as GormAllOperations<T>
+        }
+        return gormStaticApi
+    }
```
matrei: Should this be thread-safe?

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
jdaugherty: Is there a reason we renamed the field?

codeconsole: > but I think it is best to keep resource a Class<T> for 2 reasons:

@matrei @jdaugherty I explained my rational for the name change in the comment above

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/GormService.groovy
codeconsole: It could be for redundancy. Worst case scenario an old reference would be replaced, but I just made it @Lazy to prevent the possibility

jdaugherty:
I tried to update this PR for the 7.1.x branch @codeconsole can you confirm these are the changes you were expecting?

jdaugherty (approved):
Let's update the documentation for 7.1 so people know to update if they're using this.  Then you can merge this.

PR #15196, opened 2025-11-01
7.1.x Fix Enum JSON/XML Serialization for Round-Trip Compatibility

codeconsole:
# The Problem

  Grails had asymmetric enum handling that broke round-trips:

  POST/PUT (What you send):
```json
{"stage":"SUBMIT"}
```
  GET (What you received back):
```json
{"stage":{"enumType":"com.example.ChallengeStage","name":"SUBMIT"}}
```
Result: You couldn't POST back what you GET!

# The Fix

  Enums now serialize the same way they deserialize - as simple strings:

  POST/PUT (What you send):
```json
{"stage":"SUBMIT"}
```
  GET (What you get back):
```json
{"stage":"SUBMIT"}
```
 
Result: What you POST is what you GET.

### Technical Changes

  - JSON.java: Handle enums as primitives (like strings, numbers, booleans)
  - XML.java: Handle enums as primitives (like strings, numbers, booleans)
  - Removed dead code: EnumMarshaller classes (no longer needed)
  - Updated tests: Verify symmetric serialization/deserialization


gradle.properties
```diff
@@ -13,11 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-projectVersion=7.0.1-SNAPSHOT
```
jdaugherty: You are asking to merge this into 7.0.x, but this is on the 7.1.x code line?

jdaugherty:
So this is one of the first things I do in a Grails app.  I 100% agree it needs fixed, but unfortunately this is a breaking change since we're registering the marshaller by default.  Our support policy says we won't break except on major releases.  I think to accept this change in 7.x, we have to not register these by default & then in 8.x we can register them by default.

gradle.properties
codeconsole: 7.1.x

codeconsole:
> So this is one of the first things I do in a Grails app. I 100% agree it needs fixed, but unfortunately this is a breaking change since we're registering the marshaller by default. Our support policy says we won't break except on major releases. I think to accept this change in 7.x, we have to not register these by default & then in 8.x we can register them by default.

@jdaugherty I am not registering anything. I am deleting the marshallers which could be added back in by someone if they really wanted them.

matrei:
As this could break someones parsing of data returned from an endpoint, we could keep backwards compatibility by defaulting to the old behavior and add a feature flag to enable the new behavior in 7.1 and then replace and remove the old behavior in 8.

codeconsole:
> As this could break someones parsing of data returned from an endpoint, we could keep backwards compatibility by defaulting to the old behavior and add a feature flag to enable the new behavior in 7.1 and then replace and remove the old behavior in 8.

@matrei I think 8 is too far off for something that IMO is completely broken.    Does this POST request work?
```
curl -X POST 'https://localhost:8080/test' \
  -H 'Content-Type: application/json' \
  -d '{"stage":{"enumType":"com.example.ChallengeStage","name":"SUBMIT"}}'
```
if the answer is no, then this is clearly a bug and needs to be fixed.

There is a reason why you have to submit json like this:
```
curl -X POST 'https://localhost:8080/test' \
  -H 'Content-Type: application/json' \
  -d '{"stage":"SUBMIT"}'
```
even form submission sends enum String values
```
curl -X POST 'https://localhost:8080/test'\
  -H 'Content-Type: application/x-www-form-urlencoded' \
  --data 'stage=SUBMIT'
```

matrei:
> I think 8 is too far off

With a feature flag, the new serialization format would be available in 7.1.

> something that IMO is completely broken

I don't think it's broken, it just does not work like you would like.
And I agree with you.

But it is a breaking change, and we should avoid that in a minor version update.

jdaugherty:
I'm in agreement with @matrei .  I originally looked at this and assumed we were registering new marshallers, but my point was that we can't break behavior - even bad behavior in a point release.  We've stated we would only do it in a major release.  

Everyone seems to agree this is really bad behavior though.  I'm willing to discuss in the weekly and if there is consensus we can deviate from our policy.

codeconsole:
Well the trade off is doing it with a marshaller or not.  Marshaller seems overkill.

https://github.com/apache/grails-core/blob/c899909ab5e44c070505c87d883b5b4cdbfa9b50/grails-converters/src/main/groovy/grails/converters/JSON.java#L179-L197

@matrei define feature flag.  A config setting in application.yml?

matrei:
> define feature flag. A config setting in application.yml?

@codeconsole Yes, a config value. However it is set.

codeconsole:
merged in 7.0.x https://github.com/apache/grails-core/pull/15212

sbglasius:
@codeconsole Good work!

PR #15199, opened 2025-11-03
Forge: Be explicit with zip directories and entry metadata

### LINKED ISSUE #15186: Invalid Timestamp Error when running Gradle tasks on generated project
### Issue description

When generating a Grails 7.0.0 zip from https://start.grails.org and running for example `bootRun`, the following error occurs:

```console
Execution failed for task ':processResources'.
> Invalid Timestamp -3600000 for 'file '/home/mattias/Downloads/httpclienttest/grails-app/conf/spring''.
```

Updating `grails-app/conf/spring/resources.groovy` fixes that.

```console
Execution failed for task ':processResources'.
> Invalid Timestamp -3600000 for 'file '/home/mattias/Downloads/httpclienttest/grails-app/views/layouts''.
```

Updating `grails-app/views/layouts/main.gsp` fixes that.

matrei:
Fixes gh-15186

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
@@ -32,7 +32,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.FileTime;
-import java.time.Instant;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -42,6 +41,7 @@ public class ZipOutputHandler implements OutputHandler {
     private final File zip;
     private final String directory;
     private final Set<String> createdDirs = new HashSet<>();
+    private final FileTime lastModified;
 
     public ZipOutputHandler(Project project) throws IOException {
         File baseDirectory = new File(".").getCanonicalFile();
@@ -53,18 +53,18 @@ public class ZipOutputHandler implements OutputHandler {
         zip.createNewFile();
         zipOutputStream = new ZipArchiveOutputStream(Files.newOutputStream(zip.toPath()));
         directory = project.getName();
+        lastModified = FileTime.from(OutputUtils.createLastModified());
     }
 
     public ZipOutputHandler(OutputStream outputStream) {
-        zip = null;
-        zipOutputStream = new ZipArchiveOutputStream(outputStream);
-        directory = null;
+        this(null, outputStream);
     }
 
     public ZipOutputHandler(String projectName, OutputStream outputStream) {
         zip = null;
         zipOutputStream = new ZipArchiveOutputStream(outputStream);
         directory = projectName;
+        lastModified = FileTime.from(OutputUtils.createLastModified());
     }
 
     @Override
@@ -84,7 +84,6 @@ public class ZipOutputHandler implements OutputHandler {
     @Override
     public void write(String path, Template contents) throws IOException {
         String entryName = (directory != null ? StringUtils.prependUri(directory, path) : path);
-        FileTime lastModified = FileTime.from(Instant.now());
 
         // ensure parent directories exist as explicit dir entries
         // https://github.com/apache/grails-core/issues/15186
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
@@ -109,7 +109,9 @@ public class ZipOutputHandler implements OutputHandler {
 
     private void createParentDirs(String entryName, FileTime lastModified) throws IOException {
         int slash = entryName.lastIndexOf('/');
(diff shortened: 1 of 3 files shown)


grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
```diff
@@ -78,10 +83,19 @@ public boolean exists(String path) {
 
     @Override
     public void write(String path, Template contents) throws IOException {
-        ZipArchiveEntry zipEntry = new ZipArchiveEntry(directory != null ? StringUtils.prependUri(directory, path) : path);
-        if (contents.isExecutable()) {
-            zipEntry.setUnixMode(UnixStat.FILE_FLAG | 0755);
-        }
+        String entryName = (directory != null ? StringUtils.prependUri(directory, path) : path);
+        FileTime lastModified = FileTime.from(Instant.now());
```
jdaugherty: If source date epoch is set can we parse that time instead?  now can be the default otherwise ...

grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
```diff
@@ -92,4 +106,29 @@ public void close() throws IOException {
         zipOutputStream.finish();
         zipOutputStream.close();
     }
+
+    private void createParentDirs(String entryName, FileTime lastModified) throws IOException {
+        int slash = entryName.lastIndexOf('/');
```
jdaugherty, line 110: Do zip files not use the system file separator?

grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
matrei: > If source date epoch is set can we parse that time instead? now can be the default otherwise ...

I think this code only run as part of the containerized API service.

It turned out, the problem had nothing to do with build reproducibility: https://github.com/apache/grails-core/issues/15186#issuecomment-3479553169

jdaugherty: The code is in forge-core so its possible the cli could trigger it

grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
matrei: I believe, in this case, the entry names are already normalized on slash. Only slash is valid for zip files entries.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
matrei: What is the advantage of setting the last modified date of zip file entries in newly generated project zips to `SOURCE_DATE_EPOCH`?

jdaugherty: So that if we want to test build reproducibility we can ... 2 generated projects would match exactly for code built separately.

matrei: But, I don't think that would involve zip files?

matrei: The only other place where we reference `System.getenv('SOURCE_DATE_EPOCH')` in production code is in `PropertyFileUtils`.

I can add it if you think it would be useable, I just want to understand why, and if it's really going to be used?
(`ZipOutputHandler` is only used in `ZipCreateController` which is in `grails-forge-api`).

In that case, should I also add it to `FileSystemOutputHandler`?

jdaugherty: It's really just for best practice - any build we publish should be reproducible.  Before your change, it tried to write out a consistent time, and after it's going to write out .now() for every file.  I can see the argument that the container won't be published, but core is and others could use it.  

I also think longer term, we should stage forge so we can make it easier to test staging / make sure it works - but I think we have higher priorities right now.

jamesfredley (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/io/ZipOutputHandler.java
jdaugherty: I spoke to @jamesfredley more about this too.  Longer term, I want to publish forge to a staging area that we can also test as part of the vote process - to ensure it's not broken.  We'll want this reproducible for that testing too.  @matrei if you still feel strongly about this, let's discuss in the weekly?

matrei: @jdaugherty I have updated the PR to take `SOURCE_DATE_EPOCH` into account.

jdaugherty (approved):

Issue #15201, opened 2025-11-03
RenderTagLib is no longer accessible when including grails-gsp


jdaugherty:
### Issue description

In prior versions of Grails, the RenderTagLib was accessible like ApplicationTagLib, and other tag libs in grails-gsp.  This broke because it's dependency was changed from api to implementation. It's common to require these includes in Tests so this ticket will add back an api dependency in the testing-support-web package.

PR #15202, opened 2025-11-03
fix: #15201 RenderTagLib should be accessible if the gsp plugin is applied

jdaugherty:
RenderTagLib should be just as accessible as ApplicationTagLib, etc.  Otherwise, users have to include internal grails packages (grails-web-gsp-taglib)

### DIFF
--- a/grails-testing-support-web/build.gradle
+++ b/grails-testing-support-web/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     implementation platform(project(':grails-bom'))
 
     api project(':grails-gsp')
+    api project(':grails-web-gsp-taglib') // Makes RenderTagLib available for mocking in tests
     api project(':grails-rest-transforms')
     api project(':grails-interceptors')
     api 'jakarta.servlet:jakarta.servlet-api'


matrei:
> RenderTagLib should be just as accessible as ApplicationTagLib, etc. Otherwise, users have to include internal grails packages (grails-web-gsp-taglib)

How are you using `RenderTagLib`, that does not work at the moment?

jdaugherty:
I'm registering it in unit tests, and it's no longer exposed on the classpath like it was previously and like the other gsp tag libs are.

jamesfredley (approved):

matrei:
> I'm registering it in unit tests, and it's no longer exposed on the classpath like it was previously and like the other gsp tag libs are.

Then, in my view, you are explicitly using it in your test implementation and should declare it in the `testImplementation` configuration.

`@Taglib` is a runtime retention annotation and annotated classes are not needed on the application compile classpath. The other taglibs are available only because they are in `grails-gsp` and `grails-gsp` must be on the compile classpath as it has the `ControllerTagLibraryTraitInjector` transformation.

I'm not going to stop this from being merged, but I think we should be restricting our dependency graphs instead of opening them up, to be able to reduce coupling going forward.

One solution would be to add `grails-web-gsp-taglib` as an API dependency in `grails-testing-support-web` (with an explaining comment).

jdaugherty:
@matrei  Let's discuss further in the weekly?

matrei (approved):

Issue #14887, opened 2025-07-08
GRAILS_REPO_URL support for apps generated by profiles


jdaugherty:
### Issue description

forge, and all tooling except the profiles supports the notion of GRAILS_REPO_URL to force the addition of another maven repo (we use this to force the staging repo).  The profiles currently do not have this concept, so the generated applications have to be manually modified.  This ticket is to investigate adding such support to the profile if the env is set.

PR #15206, opened 2025-11-04
Post 7.0.1 release adjustments

matrei:
Generalizations and simplifications to the release workflow and scripts.


.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
-        run: |
-          echo "repository_name=${GITHUB_REPOSITORY##*/}" >> $GITHUB_OUTPUT
       - name: "📥 Checkout repository"
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
+          ref: ${{ env.TAG }}
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: ${{ github.ref }}
-          fetch-depth: 1
```
jdaugherty, line 48: The fetch by ref, depth, and tag is required to correctly calculate the tags for the doc generation.  This needs reverted

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
-        run: |
-          find . -depth \( -type f -o -type d \) -exec touch -d "@${SOURCE_DATE_EPOCH}" {} +
+      - name: "📅 Ensure source files use common date"
+        run: find . -depth \( -type f -o -type d \) -exec touch -d "@${SOURCE_DATE_EPOCH}" {} +
       - name: '🔐 Set up GPG'
         run: |
           echo "${{ secrets.GRAILS_GPG_KEY }}" | gpg --batch --import
           gpg --list-keys
-        env:
-          GPG_KEY_ID: ${{ secrets.GPG_KEY_ID }}
```
jdaugherty, line 60: we should still set the secret to a variable and then reference that variable so that when we debug scripts we can see the full script contents in github actions - otherwise github will mask the secret

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
-        run: echo "value=${GITHUB_REF:11}" >> $GITHUB_OUTPUT
-      - name: '⚙️ Run pre-release'
+          develocity-access-key: ${{ secrets.GRAILS_DEVELOCITY_ACCESS_KEY }}
+      - name: "⚙️ Run pre-release"
         uses: apache/grails-github-actions/pre-release@asf
         env:
-          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+          RELEASE_VERSION: ${{ env.VERSION }}
           RELEASE_SCRIPT_PATH: '.github/scripts/setReleasedGrailsVersion.sh'
-      - name: 'Fetch release Commit Hash'
```
jdaugherty, line 85: Didn't we determine this is still required because the pre-release replaces the tag and we want to reference the updated tag?

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
-            grails-forge/grails-cli/build/distributions/apache-grails-${{ steps.release_version.outputs.value }}-bin.zip.asc
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - name: "Create Staging Repository"
+            grails-forge/grails-cli/build/distributions/apache-grails-${{ env.VERSION }}-bin.zip
+            grails-forge/grails-cli/build/distributions/apache-grails-${{ env.VERSION }}-bin.zip.sha512
+            grails-forge/grails-cli/build/distributions/apache-grails-${{ env.VERSION }}-bin.zip.asc
+      - name: "✨ Create Staging Repository"
         env:
-          GRAILS_PUBLISH_RELEASE: "true"
```
jdaugherty, line 163: This should still be set

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
+      - name: "📅 Store common build date" # to ensure a reproducible build
         run: echo "SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)" >> "$GITHUB_ENV"
-      - name: "Ensure source files use common date"
-        run: |
-          find . -depth \( -type f -o -type d \) -exec touch -d "@${SOURCE_DATE_EPOCH}" {} +
+      - name: "📅 Ensure source files use common date"
+        run: find . -depth \( -type f -o -type d \) -exec touch -d "@${SOURCE_DATE_EPOCH}" {} +
       - name: '🔐 Set up GPG'
         run: |
           echo "${{ secrets.GRAILS_GPG_KEY }}" | gpg --batch --import
```
jdaugherty: Isn't this GPG_KEY_ID

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
+      - name: "✅ Close Staging Repository"
         env:
-          GRAILS_PUBLISH_RELEASE: 'true'
           NEXUS_PUBLISH_USERNAME: ${{ secrets.NEXUS_STAGE_DEPLOYER_USER }}
           NEXUS_PUBLISH_PASSWORD: ${{ secrets.NEXUS_STAGE_DEPLOYER_PW }}
-          NEXUS_PUBLISH_URL: 'https://repository.apache.org/service/local/'
+          NEXUS_PUBLISH_URL: ${{ vars.STAGING_URL }}
           NEXUS_PUBLISH_STAGING_PROFILE_ID: ${{ secrets.STAGING_PROFILE_ID }}
-          NEXUS_PUBLISH_DESCRIPTION: '${{ steps.extract_repository_name.outputs.repository_name }}:${{ steps.release_version.outputs.value }}'
-          SIGNING_KEY: ${{ secrets.GPG_KEY_ID }}
```
jdaugherty, line 231: Are we sure this isn't required?  I thought removing this would trigger a validation error

.github/workflows/release.yml
```diff
@@ -21,267 +21,220 @@ permissions: { }
...
-          tag_name: ${{ github.event.release.tag_name }}
-          files: build/PUBLISHED_ARTIFACTS.txt
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - name: "Generate Build Date file"
+      - name: "📅 Generate build date file"
         run: echo "$SOURCE_DATE_EPOCH" >> build/BUILD_DATE.txt
-      - name: "Upload Build Date file"
+      - name: "📤 Upload build date, checksums and published artifact files"
         uses: softprops/action-gh-release@v2
```
jdaugherty, line 222: We should make sure this publishes to a non-draft release by setting that argument on the action

.github/workflows/release.yml
```diff
@@ -294,77 +247,86 @@ jobs:
...
-          zip -r "apache-grails-${{ needs.publish.outputs.release_version }}-src.zip" grails -x 'grails/.git/*' -x 'grails/.github/*'
+        run: >
+          zip -r
+          apache-grails-${VERSION}-src.zip
+          grails
+          -x 'grails/.git/*'
+          -x 'grails/.github/*'
       - name: '🔐 Set up GPG'
         run: |
           echo "${{ secrets.GRAILS_GPG_KEY }}" | gpg --batch --import
```
jdaugherty: Please see previous comment about the secret; we need to use an environment variable

.github/workflows/release.yml
```diff
@@ -294,77 +247,86 @@ jobs:
...
         with:
-          tag_name: ${{ github.event.release.tag_name }}
+          tag_name: ${{ env.TAG }}
           files: |
-            apache-grails-*-src.zip
-            apache-grails-*-src.zip.sha512
-            apache-grails-*-src.zip.asc
-      - name: "Remove CHECKSUMS.txt asset from release"
-        working-directory: "grails"
+            apache-grails-${{ env.VERSION }}-src.zip
```
jdaugherty: We abstracted so much already - shouldn't we abstract the basename of our distributions?

.github/workflows/release.yml
```diff
@@ -375,116 +337,119 @@ jobs:
...
           set -e
-
-          if svn ls https://dist.apache.org/repos/dist/dev/grails --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive; then
-            echo "Dev Folder 'grails' already exists — skipping creation"
+          if svn ls https://dist.apache.org/repos/dist/dev/grails --username "${SVN_USERNAME}" --password "${SVN_PASSWORD}" --non-interactive; then
+            echo "Dev Folder [grails] already exists — skipping creation"
           else
-            echo "Dev Folder 'grails' does not exist, creating"
-            svnmucc --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --non-interactive \
+            echo "Dev Folder [grails] does not exist, creating"
```
jdaugherty: Lets abstract the project host into a variable?

.github/workflows/release.yml
```diff
@@ -533,10 +523,12 @@ jobs:
...
     name: "VOTE SUCCEEDED - Publish Documentation"
-    needs: [ publish, source, upload ] # TODO Once we have confirmed release won't fail, add it as a dependency here
+    needs: [ publish, source, upload, release ]
     runs-on: ubuntu-24.04
     steps:
-      - name: Maximize build space
+      - name: "📝 Establish release version"
+        run: echo "VERSION=${TAG#v}" >> "$GITHUB_ENV"
+      - name: "⚙️ Maximize build space"
         uses: easimon/maximize-build-space@c28619d8999a147d5e09c1199f84ff6af6ad5794
```
jdaugherty, line 540: What is this for?

.github/workflows/release.yml
matrei: The default `fetch-depth` is `1`.

Aah, I missed that the docs are generated in this job, I'll add back `fetch-tags: true`.

.github/workflows/release.yml
matrei: No, I did not change this. And, to my knowledge, `GPG_KEY_ID` was unused.

.github/workflows/release.yml
matrei: I don't agree that that use case warrants complicating the workflow. Isn't it easy enough to change if debugging is needed. To my knowledge, `GPG_KEY_ID` is unused here.

.github/workflows/release.yml
matrei: It's fetched on line 424 instead.
As long as the checkout is done with `ref: ${{ env.TAG }}` it will be ok.

.github/workflows/release.yml
matrei: It is! At the workflow level.
But, I should probably change to `true` with quotes.

.github/workflows/release.yml
matrei: Yes, but I'm unsure how that works.
I think the documentation is unclear:
```yml
draft:
  description: "Creates a draft release. Defaults to false"
  required: false
```

.github/workflows/release.yml
matrei: I did not change this line.

.github/workflows/release.yml
matrei: I did that in the other repositories using `REPO_NAME` which suited them.
Then I saw that it doesn't match here, so I thought we would make another iteration for that, but I wanted to discuss first.

.github/workflows/release.yml
matrei: @jamesfredley Added it when he worked on docs publishing.

.github/workflows/release.yml
jdaugherty: I'm ok waiting - but I think making it generic will make it easier longer term.  It's up to you then.

.github/workflows/release.yml
jdaugherty: I think this was a mistake.  From a security perspective, we don't want to write a file out on the file system with a secret hard coded.  We only want that file to contain an env variable.

.github/workflows/release.yml
jdaugherty: I'm pretty sure we tried an iteration like this because of that default and it failed.  We had to go back and add it in 80ffc9952153669e621f92e9dd150c25a0347b71

.github/workflows/release.yml
jdaugherty: I'm looking at the repo: https://github.com/softprops/action-gh-release 

The readme has this: 
draft 	Boolean 	Indicator of whether or not this release is a draft

So i think we want draft: false to filter draft releases?

.github/workflows/release.yml
jdaugherty: Sorry, I'm trying to advocate that we should fix this so that the files on the filesystem dont' have hardcoded secrets =)

.github/workflows/release.yml
jdaugherty: Looking at the README later, it does look like thats to update the value =(  Ignore this suggestion.

.github/workflows/release.yml
matrei: I have added `DIST_NAME: apache-grails` for this.

.github/workflows/release.yml
matrei: I have added `SVN_PROJECT: grails` for this.

.github/workflows/release.yml
matrei: Apparently `fetch-tags: true` does not work with `fetch-depth: 1`.
https://github.com/actions/checkout/issues/1471

I changed to `fetch-depth: 0` with `filter: tree:0` instead.

.github/workflows/release.yml
matrei: Fixed

.github/workflows/release.yml
matrei: Hmm, why would finding and closing the staging repository need the signing key?

.github/workflows/release.yml
matrei: Fixed

etc/bin/verify.sh
```diff
@@ -114,11 +114,12 @@ set -e
...
+echo "☑️   Set the override repo: 'export GRAILS_REPO_URL=https://repository.apache.org/content/groups/staging'"
+echo "☑️   Run the wrapper ShellApp: cd ${DOWNLOAD_LOCATION}/apache-grails-wrapper-${VERSION}-bin/ShellApp && ./gradlew bootRun --init-script ${DOWNLOAD_LOCATION}/use-staging-repo.gradle"
+echo "☑️   Run the wrapper ForgeApp: cd ${DOWNLOAD_LOCATION}/apache-grails-wrapper-${VERSION}-bin/ForgeApp && ./gradlew bootRun --init-script ${DOWNLOAD_LOCATION}/use-staging-repo.gradle"
+echo "☑️   Run the cli ShellApp: cd ${DOWNLOAD_LOCATION}/apache-grails-${VERSION}-bin/bin/ShellApp && ./gradlew bootRun --init-script ${DOWNLOAD_LOCATION}/use-staging-repo.gradle"
+echo "☑️   Run the cli ForgeApp: cd ${DOWNLOAD_LOCATION}/apache-grails-${VERSION}-bin/bin/ForgeApp && ./gradlew bootRun --init-script ${DOWNLOAD_LOCATION}/use-staging-repo.gradle"
+# BELOW STEP NEEDS CLARIFICATION
 echo "☑️   Add the local repos to the application and then run the shell cli from one of the applications and ensure all commands show as expected - pay attention to the scaffolding ones since they are dynamically loaded"
```
matrei: @jdaugherty Could we re-phrase this last instruction? I cannot make sense of it.

.github/workflows/release.yml
jdaugherty: Lets try it!

etc/bin/verify.sh
jdaugherty: The profile generated applications do not add the staging repos directly to the gradle files (https://github.com/apache/grails-core/issues/14887).  As a work around, the verify.sh script passes a local-repos.gradle file to gradle commands so that dependencies can be found.  Except we cannot do that when running the interactive grails shell.  Thus:

1. add the repo manually to an app generated so that the interactive shell does not fail to resolve
2. run the interactive shell under the generated application directory that has had the repos added to the build.gradle
3. type help, which should look up commands dynamically on the class path
4. ensure the scaffolding command is listed 

This is testing the dynamic resolution of commands in the grails shell for commands not built in (scaffolding is an example)


Given all of this, can you help me fix this statement now?

jdaugherty (approved):

jdaugherty (approved):

PR #15213, opened 2025-11-06
Allow greedy parameters (+) prior to a format paramter that consume everything but the last .

codeconsole:
# Why?
Grails by default doesn't use mime types to render json.  It encourages using $format.  If you want to handle multiple mime types with the same controller while supporting ids with decimals, this becomes impossible due to how regular expression url mappings are handled in Grails.   This PR offers the capability of restricting format to not include periods by specifying the $id attribute to be greedy with the + symbol. 

# Problem
Currently when a url parameter is specified prior prior to a $format parameter, the parameter only consumes everything up until the first dot.   This results in undesired behavior on sites that's use parameters for usernames that include 1 or more periods. 

For instance

"/$controller/$action?/$id?(.$format)?" 
/user/show/bob.smith.json  results in controller = 'UserController' id = 'bob'. format = 'smith.json'

# The solution

Introduce the + symbol after a url parameter to make it greedy.

Syntax:

  - "/$id+(.$format)?" - Required id, greedy (splits at last dot)
  - "/$id+?(.$format)?" - Optional id, greedy (splits at last dot)

Examples with this Fix

"/$controller/$action?/$id+?(.$format)?" 

  /user/show/est.test.json  → id=test.test, format=json
  /user/show/test.json       → id=test, format=json
  /user/show/foo.bar.baz.xml → id=foo.bar.baz, format=xml

### DIFF
--- a/grails-doc/src/en/guide/upgrading/upgrading60x.adoc
+++ b/grails-doc/src/en/guide/upgrading/upgrading60x.adoc
@@ -754,3 +754,109 @@ grails:
 * **8.0**: `SimpleEnumMarshaller` will become the default
 
 The legacy `org.grails.web.converters.marshaller.json.EnumMarshaller` and `org.grails.web.converters.marshaller.xml.EnumMarshaller` classes are marked as `@Deprecated(forRemoval = true, since = "7.0.2")` and will be removed in Grails 8.0.
+
+===== 12.27 Greedy Extension Parameter Matching in URL Mappings
+
+Grails 7.1 introduces a new greedy extension parameter marker (`+`) for URL mappings that provides more intuitive handling of file extensions in URLs with multiple dots.
+
+====== The Problem with Default Behavior
+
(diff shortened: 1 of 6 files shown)


jdaugherty:
This regex change does not appear to compile: 

Caused by:
        java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 3
        ^/(*)(\.(*))?/??$
           ^

codeconsole:
> This regex change does not appear to compile:
> 
> Caused by: java.util.regex.PatternSyntaxException: Dangling meta character '_' near index 3 ^/(_)(.(*))?/??$ ^

@jdaugherty accidentally added a couple parens when I fixed codenarc spacing. Should be good now

sbglasius (approved):

jdaugherty (approved):

jdaugherty:
We should update the new features for 7.1.x to mention this.  thank you for your contribution

PR #15150, opened 2025-10-17
Apply Generated annotation

gustavoharff:
This PR is the same as https://github.com/apache/grails-data-mapping/pull/1507, but for Grails 7.

Related to #14506, #14364

The only difference to the original PR is the GormEntityDirtyCheckable class, which was created in Grails 7.

PR #15214, opened 2025-11-06
Do not mark AST transformation methods as generated

gustavoharff:
Follow up #15150

Jacoco isn't displaying coverage lines on methods annotated with Transactional (or other annotation that apply AST transformation).

I think that has been applied to display a more clean list of methods in the tree, but I guess the coverage lines is more important than that.

I did some tests and below you can see the results.

### Current

No coverage lines are displayed.

```groovy
markAsGenerated(classNode, renamedMethodNode)
```

image

### Marking the original method as generated

I don't think this is the best way, it feels counterintuitive to me.

```groovy
markAsGenerated(classNode, methodNode)
```

image

### Not marking methods as generated

It has a larger methods list than other ways.

image

---

I chose not to mark the methods as generated, let me know what you think.


jdaugherty:
This appears to be a follow-up to https://github.com/apache/grails-core/pull/15150

jdaugherty:
Given that we made this change on 10/17 and it recently was released, I'm going to go ahead and merge this since it's basically a revert.

PR #15221, opened 2025-11-10
Gradle Plugin Fix - remove calls to setSystemProperty() and set on individual tasks instead

### LINKED ISSUE #15222: base.dir can be set incorrectly for bootRun and other gradle tasks
### Issue description

base.dir can be set incorrectly for bootRun and other gradle tasks because setProperty() is called and the Grails Gradle plugins are now written to be more parallel capable.

jdaugherty:
This was an extremely frustrating bug for multiproject builds.  The gradle plugins were calling System.setProperty() as part of any grails gradle plugin setup - which means the set Property is for the life of that jvm instance.  Since gradle is threaded, with multiproject builds, it resulted in a race condition / random setting of this field.  This meant that the basedir was incorrectly set for bootRun, etc.  It was even complicated more in that if it was set for that daemon process, it would remain set.  

We already set other properties as part of task execution, so this PR takes the same approach.  This does mean that the properties are no longer set for gradle code, but after testing, I think this is ok.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -133,8 +133,6 @@ class GrailsGradlePlugin extends GroovyPlugin {
 
         registerFindMainClassTask(project)
 
-        configureGrailsBuildSettings(project)
-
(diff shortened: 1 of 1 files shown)


jdaugherty:
Fixes https://github.com/apache/grails-core/issues/15222

jamesfredley (approved):

codeconsole (approved):

matrei (approved):

PR #15225, opened 2025-11-13
Apply `java-library` Gradle Plugin to Grails Plugin projects

matrei:
Grails plugins are Java libraries and should have access to the `api` dependency configuration by default.


sbglasius (approved):
Good addition

PR #15232, opened 2025-11-15
7.1.x importGrailsCommonAnnotations dependency check fix

codeconsole:
importGrailsCommonAnnotations resolution check wasn't checking the full hierarchy

so only jakarta.validation.constraints was being resolved.

this fixes the dependency check and also creates a helper method that can be used to check any dependency.


matrei (approved):

codeconsole:
https://github.com/apache/grails-core/pull/15118

PR #15234, opened 2025-11-17
fix: Re-implemented ContainerRenderer on AbstractJsonViewContainerRenderer

### LINKED ISSUE #14199: Restore implements ContainerRenderer<C, T> on AbstractJsonViewContainerRenderer<C,T>
The interface XXX cannot be implemented more than once with different arguments.

implements ContainerRenderer<C, T> was causing Render to be implement more than once. It is currently removed from class AbstractJsonViewContainerRenderer<C,T>

https://github.com/grails/grails-views/pull/581/files#diff-b2646b6110b34798952582e9a501678112c0ed5d810ac556902857099652a31e

The way it worked historically would still require Groovy allowing classes to extend multiple interfaces with different generics.  Groovy allowed this for traits to support Grails 7, but would not allow it for interfaces, since Java does not allow that.  https://issues.apache.org/jira/browse/GROOVY-11508
https://issues.apache.org/jira/browse/GROOVY-5106
https://github.com/apache/groovy/commit/9d0d67197eb8cd333e973b08d4d0f4b7f3f41475

### LINKED ISSUE #15228: respond with Errors does not render view errors/_errors.gson
### Expected Behavior

When I call 

```
respond (validateable.errors)
```

I expect my `grails-app/views/errors/_errors.gson` view to be rendered

### Actual Behaviour

That the general renderer of my Errors class is rendered, ignoring `errors/_errors.gson` 

### Steps To Reproduce

I have noticed, that in Grails 6.2.3 when calling respond(cmd.errors) and having a `views/errors/_errors.gson` it gets called through the `grails.plugin.json.view.mvc.JsonViewResolver` and `grails.plugin.json.renderer.ErrorsJsonViewRenderer` but in Grails 7.0.2 the view resolver is not called, instead the `DefaultJsonRenderer` is used, and hence the `_errors.gson` is not used.
As far as I can see, the `JsonViewResolver#initialize()` is called, but the registration is not read when `org.grails.plugins.web.rest.render.DefaultRendererRegistry#findContainerRenderer` is called.

I will proceed with writing a functional test in `grails-test-examples/views-functional-tests/src/test/groovy/grails/plugin/json/view/test/JsonViewUnitTestSpec.groovy`

### Environment Information

Ubuntu 24.04
JDK 17 or 21

### Example Application

https://github.com/sbglasius/grails-respond-gson-bug

### Version

7.0.2

sbglasius:
* Re-enabled gson view resolver
* `ContainerRenderer` extends `Renderer` fix to generics parameter
  * consequence fix, that `Renderer` calls `render(T render, ...` changed to `render(Object render, ...` and thus contract changed back to pre Grails 7.0.x
* New test project to verify outcome.
Fixes #14199 
Fixes #15228

### DIFF
--- a/grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/BookSpec.groovy
+++ b/grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/BookSpec.groovy
@@ -40,7 +40,7 @@ class BookSpec extends HttpClientSpec {
     def setupSpec() {
         objectMapper = new ObjectMapper()
     }
-
+    
     @RunOnce
     @BeforeEach
     void init() {
@@ -56,22 +56,17 @@ class BookSpec extends HttpClientSpec {
         HttpClientResponseException e = thrown()
         e.response.status == HttpStatus.UNPROCESSABLE_ENTITY
         e.response.headers.getFirst(HttpHeaders.CONTENT_TYPE).isPresent()
-        // This has changed somewhere along the way
-        // e.response.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/vnd.error;charset=UTF-8'
-        // to ->
-        e.response.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/json;charset=UTF-8'
-        objectMapper.readTree(e.response.body().toString()) == objectMapper.readTree('''
+        e.response.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/vnd.error;charset=UTF-8'
+        objectMapper.readTree(e.response.body().toString()) == objectMapper.readTree("""
             {
-                "errors": [
-                    {
-                        "object": "functional.tests.Book",
-                        "field": "title",
-                        "rejected-value": null,
-                        "message": "Property [title] of class [class functional.tests.Book] cannot be null"
-                    }
-                ]
-            }
-        ''')
+              "message": "Property [title] of class [class functional.tests.Book] cannot be null",
+              "path": "/book/index",
+              "_links": {
+                "self": {
+                  "href": "$baseUrl/book/index"
+                }
+              }
+            }""")
     }
 
     void 'Test REST view rendering'() {
--- a/grails-rest-transforms/src/main/groovy/grails/rest/render/ContainerRenderer.groovy
+++ b/grails-rest-transforms/src/main/groovy/grails/rest/render/ContainerRenderer.groovy
@@ -24,7 +24,7 @@ package grails.rest.render
  * @author Graeme Rocher
  * @since 2.3
  */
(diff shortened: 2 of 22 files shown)


matrei (approved):
Looks solid! Good find and fix. Thanks for adding functional test.

grails-rest-transforms/src/main/groovy/grails/rest/render/Renderer.groovy
```diff
@@ -39,5 +39,5 @@ interface Renderer<T> extends MimeTypeProvider {
      * @param object The object to render
      * @param context The {@link RenderContext}
      */
-    void render(T object, RenderContext context)
+    void render(Object object, RenderContext context)
```
jdaugherty, line 42: Since this is part of the rest transforms, that means this is an API breaking change. I think we have to do this, but I'm wondering if it should be a 7.1.x bug fix and we just ship 7.1.x sooner.

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/BookSpec.groovy
```diff
@@ -56,22 +56,17 @@ class BookSpec extends HttpClientSpec {
...
-                    {
-                        "object": "functional.tests.Book",
-                        "field": "title",
-                        "rejected-value": null,
-                        "message": "Property [title] of class [class functional.tests.Book] cannot be null"
-                    }
-                ]
-            }
-        ''')
+              "message": "Property [title] of class [class functional.tests.Book] cannot be null",
```
jdaugherty, line 62: I assume if there isn't an error.gson, then it does render errors like before?

jdaugherty (commented):
My only concern is the API change and putting this in 7.0.x.  I understand that this is a critical bug fix, but given that we were wanting to do 7.1.x in November anyhow, should we instead promote 7.1.x sooner?

jamesfredley (approved):
Thank you @sbglasius for finding a solution.  My changes in September 2024 allowed grails-views to compile on Groovy 4, but I was worried we would run into some issue, at some point and these issues tracked that, for many months:
https://github.com/apache/grails-views/issues/582
https://github.com/apache/grails-core/issues/14199

grails-rest-transforms/src/main/groovy/grails/rest/render/Renderer.groovy
sbglasius: I disagree, that it's a breaking change, since that's how it was in Grails 6. It was changed in Grails 7.

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/BookSpec.groovy
sbglasius: Yes, I found it because my client had a custom `_errors.gson` file that was not rendered.

PR #15236, opened 2025-11-18
Minor corrections to documentation (whatsnew)

### LINKED ISSUE #15233: Version mismatch in documentation
### Issue description

Image

matrei:
Closes #15233

### DIFF
--- a/grails-doc/src/en/guide/introduction/whatsNew.adoc
+++ b/grails-doc/src/en/guide/introduction/whatsNew.adoc
@@ -21,7 +21,7 @@ This section covers all the new features introduced in Grails 7
 
 === Overview
 
-Grails {GrailsVersion} is a major release that includes new features, improvements, and dependency upgrades.
+Grails 7 is a major release that includes new features, improvements, and dependency upgrades.
 This release focuses on enhancing the developer experience, improving performance, and ensuring compatibility with the latest technologies.
 
 For detailed information on how to upgrade to Grails 7, including major dependency changes, please see the xref:upgrading#upgrading60x[Upgrading from Grails 6 to Grails 7] section.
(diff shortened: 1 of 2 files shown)


grails-doc/src/en/guide/introduction/whatsNew.adoc
```diff
@@ -21,7 +21,7 @@ This section covers all the new features introduced in Grails 7
 
 === Overview
 
-Grails {GrailsVersion} is a major release that includes new features, improvements, and dependency upgrades.
```
jdaugherty, line 24: I think I tried to make this change before and @jamesfredley pushed back on this.  Should we instead change line 20?

jamesfredley: @jdaugherty and I debated a prior PR and I think we split the difference with fixed versions vs variables.  I am onboard with changing this one.

Or we could add GrailsMajorVersion in https://github.com/apache/grails-core/blob/d08a8b10d4249b14a5304b5ceb55c03993cfd5d9/grails-doc/build.gradle and use that.

jamesfredley (approved):

grails-doc/src/en/guide/introduction/whatsNew.adoc
matrei: What happens here when we release the next minor version (7.1).
Don't we need to rewrite this page anyway.

jamesfredley: Historically:

https://grails.apache.org/docs/6.0.0/guide/introduction.html#whatsNew vs https://grails.apache.org/docs/6.1.0/guide/introduction.html#whatsNew

https://grails.apache.org/docs/5.0.0/guide/introduction.html#whatsNew vs
https://grails.apache.org/docs/5.2.0/guide/introduction.html#whatsNew

PR #15252, opened 2025-11-22
Fix links to controllers with namespaces in create-app index.gsp

codeconsole:
Currently the default index.gsp is displaying the wrong links with controllers with namespaces. This fixes it so that the namespace is included in the url.

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grails/templates/urlMappings.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/grails/templates/urlMappings.rocker.raw
@@ -27,6 +27,7 @@ package @project.getPackageName()
 class UrlMappings {
     static mappings = {
 @if(applicationType == ApplicationType.WEB || applicationType == ApplicationType.WEB_PLUGIN) {
+        "/$namespace/$controller/$action?/$id?(.$format)?" {}
         "/$controller/$action?/$id?(.$format)?"{
             constraints {
                 // apply constraints here
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
@@ -71,7 +71,7 @@
                 <ul>
                     <g:each var="c" in="${grailsApplication.controllerClasses.sort { it.fullName } }">
                         <li class="controller">
-                            <g:link controller="${c.logicalPropertyName}">${c.fullName}</g:link>
+                            <g:link namespace="${c.namespace}" controller="${c.logicalPropertyName}">${c.fullName}</g:link>
                         </li>
                     </g:each>
                 </ul>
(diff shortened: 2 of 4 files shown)


matrei:
Will this actually change the rendered links?

I tried it, and with two controllers:

```groovy
package com.example

class DemoController {

    def index() {
        render 'DEMO'
    }
}
```

and

```groovy
package com.example.test

class DemoController {

    static namespace = 'custom'

    def index() {
        render 'CUSTOM/DEMO'
    }
}
```

both controller links generated by `index.gsp` point to `/demo/index`, regardless of the changes in this PR.

I might be misunderstanding something, I have not worked much with namespaced controllers.

codeconsole:
@matrei do you have a url mapping rule for your namespace?

```groovy
    static mappings = {
        "/$namespace/$controller/$action?/$id?(.$format)?" {}
        "/$controller/$action?/$id?(.$format)?"{}

        "/"(view:"/index")
        "500"(view:'/error')
        "404"(view:'/notFound')
    }
```

matrei:
> @matrei do you have a url mapping rule for your namespace?
> 
> ```groovy
>     static mappings = {
>         "/$namespace/$controller/$action?/$id?(.$format)?" {}
>         "/$controller/$action?/$id?(.$format)?"{}
> 
>         "/"(view:"/index")
>         "500"(view:'/error')
>         "404"(view:'/notFound')
>     }
> ```

Aha, OK, thanks! Then shouldn't this also be the default starter template for `UrlMappings`?

codeconsole:
> Aha, OK, thanks! Then shouldn't this also be the default starter template for `UrlMappings`?

we could if you think namespaces should be default behavior.

matrei:
What would be the drawback in adding this to the `UrlMappings` template?

Does it make sense to add namespace code to some of the templates (`index.gsp`) but not to all places (`UrlMappings.groovy`) needed for it to actually work?

codeconsole:
No drawback, just potentially an unnecessary mapping that could just be deleted by the end user.  

It makes sense to add it to index.gsp just because the page generates bad links if you do add namespaces. 
We can add it to the mappings to eliminate any potential confusion.

sbglasius:
+1 for adding mapping as well

codeconsole:
> +1 for adding mapping as well

@sbglasius @matrei  I added the mapping in the last commit

sbglasius (approved):

jamesfredley:
I have run into this issue and worked around it by adding `namespace=''` to the tags and am in favor of fixing it via this PR.

Given the generated application `UrlMappings.groovy` includes the placeholder for constraints, I think we should put it on both. 

```
       "/$namespace/$controller/$action?/$id?(.$format)?" {
            constraints {
                // apply constraints here
            }
        }

        "/$controller/$action?/$id?(.$format)?" {
            constraints {
                // apply constraints here
            }
        }
```

I believe the same issue exists for the following in addition to <g:link>.  Currently the URL will not include the namespace unless explicitly added to the tag.  This can be a separate PR.  

```
<g:form namespace="administration" ... >
<f:table namespace="administration" .. >
```

jamesfredley (approved):

codeconsole:
@jamesfredley unless there is some logic somewhere that actually adds constraints, I think adding another constraints block might be a bit redundant and add confusion.  
```
            constraints {
                // apply constraints here
            }
```

Does it serve any purpose other than providing an example of how constraints might be added?  What percentage of users ever modify this constraints block?  Any?


```
<g:form namespace="administration" ... >
<f:table namespace="administration" .. >
```
Are these examples from the scaffold templates?  If so, I ave no problem making that modification.    However, I do recommend in practice using custom templates for each namespace which was enabled by a previous PR I did.

Issue #15258, opened 2025-11-25
bom does not include properties for dependency versions


codeconsole:
The bom has dramatically changed from versioned properties in [7.0.0-M1](https://repo1.maven.org/maven2/org/grails/grails-bom/7.0.0-M1/grails-bom-7.0.0-M1.pom), M2, [M3](https://repo1.maven.org/maven2/org/grails/grails-bom/7.0.0-M3/grails-bom-7.0.0-M3.pom), [M4](https://repo1.maven.org/maven2/org/apache/grails/grails-bom/7.0.0-M4/grails-bom-7.0.0-M4.pom), [RC1](https://repo1.maven.org/maven2/org/apache/grails/grails-bom/7.0.0-RC1/grails-bom-7.0.0-RC1.pom) and [7.0.0-RC2](https://repo1.maven.org/maven2/org/apache/grails/grails-bom/7.0.0-RC2/grails-bom-7.0.0-RC2.pom) to hard coded versions in [7.0.0](https://repo1.maven.org/maven2/org/apache/grails/grails-bom/7.0.0/grails-bom-7.0.0.pom)

After versioned properties had been used over the course of 9 months since M1 was released on 12/23/24 through RC2 on 9/3/25, seems like a bug broke them before final was released on 10/15/25?

Here is example of how it used to work [7.0.0-M3 app](https://github.com/codeconsole/grails-bom-demo/commit/d70ce75bb035687eda34837052c0bdc23f2ee5a1) showing how easy it was to use **_any version_** of Groovy or Spring Boot.

All you had to do before was add 2 LOC to `gradle.properties` and everything worked perfectly.

### 7.0.0-M3 running Spring Boot 3.5.8 and Groovy 4.0.29
```properties
spring-boot.version=3.5.8
groovy.version=4.0.29
```
You could literally override any version group in the bom with a single property change.

### 7.0.0-7.0.3 - Broken and Dangerous

Now it appears that when you set a new version for anything that has transitive dependencies, the grails-bom overrides the transitive dependencies and you get mixed versions of dependencies.

### Version
7.0.0-7.0.3

codeconsole:
Fixed versions so they would display, but this is WAY TOO VERBOSE in current form:
```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #15260, opened 2025-11-25
#15258 - fix - restore bom properties

jdaugherty:
This PR will fail until https://github.com/apache/grails-gradle-publish/pull/16 is merged.  Properties in the bom were broken because of not doing a deferred lookup of the pomCustomization closure.  This change adopts the provider syntax for pomCustomization so that the property isn't looked up until execution


jdaugherty:
This fixes the root issue of the properties not being generated due to the parallel resolution order of Gradle.  Once the publish change is merged, we can merge this , and then re-evaluate the issue after both projects are published to maven for testing.

codeconsole (approved):

matrei (approved):

PR #15269, opened 2025-12-01
Centralizing Gradle Logic - java-config

jdaugherty:
Follow-up to https://github.com/apache/grails-core/pull/15114 for `java-config.gradle`

This PR also introduces a SharedProperties plugin that when applied will "inherit" all gradle.properties up to the root directory.  For duplicate property handling, it will not replace the property if previously defined.

PR #15273, opened 2025-12-02
Make GSP Gradle plugin order-independent

matrei:


jdaugherty (approved):

jdaugherty:
I don't think we can merge this into 7.x: https://github.com/apache/grails-core/pull/15269#issuecomment-3602954401

grails-core/src/test/groovy/grails/util/GrailsUtilTests.java
```diff
@@ -33,7 +33,7 @@ public class GrailsUtilTests {
 
     @Test
     public void testGrailsVersion() {
-        assertEquals("7.0.4-SNAPSHOT", GrailsUtil.getGrailsVersion());
+        assertEquals("7.0.5-SNAPSHOT", GrailsUtil.getGrailsVersion());
```
jdaugherty: We need to update this to 8.0.0-SNAPSHOT

jdaugherty:
I'd suggest we merge the other PR first, then update this one so we can go ahead and remove the 'groovy' plugins in this one after the other is merged upward.

PR #15274, opened 2025-12-02
Moving to a shared author/developer list

jdaugherty:
See https://lists.apache.org/thread/c33fx3djh17tl67y0g3szhvsq98cq2jc

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
@@ -192,30 +192,128 @@ class PublishPlugin implements Plugin<Project> {
     }
 
     private static List<MavenPomDeveloper> getProjectDevelopers(Project project) {
-        // Note: id is typically the github user id if the user has a github account (historically, not always true)
+        // Note: id is typically the github user id if the user has a github account
+        // Note: these lists are sorted alphabetically by section
         [
             // Founders
             founder('dierk', 'Dierk König', project),
             founder('glaforge', 'Guillaume LaForge', project),
             founder('graemerocher', 'Graeme Rocher', project),
-            founder('stevendevijver', 'Steven Devijver', project),
-            // Developers - `active` contributors (should have an ASF account)
+            founder('devijvers', 'Steven Devijver', project),
+            // Developers
+            // - `active` contributors (should have an ASF account)
+            // - supports the current framework
+            developer('bkoehm','Brian Koehmstedt', project),
+            developer('borinquenkid','Walter B Duque de Estrada', project),
             developer('codeconsole', 'Scott Murphy Heiberg', project),
+            developer('davydotcom','David Estes', project),
             developer('jamesfredley', 'James Fredley', project),
             developer('jdaugherty', 'James Daugherty', project),
+            developer('jpammer', 'Jonas Pammer', project),
             developer('lhotari', 'Lari Hotari', project),
             developer('matrei', 'Mattias Reichel', project),
+            developer('paulk','Paul King', project),
             developer('sbglasius', 'Søren Berg Glasius', project),
             developer('sdelamo', 'Sergio del Amo', project),
-            // Past Developers, Criteria:
+            developer('tbrasmussen', 'Thomas Rasmussen', project),
+            // Past Developers
             // - non-apache members go here until their ASF account is created
             // - non-active, no contributions across mailing lists, commits, or grails processes in the last 12 months
+            // - contributors in the pom previously
+            // - significant contributors (i.e. they were involved in Grails Development or a member of a company that was)
+            emeritus('ColinHarrington','Colin Harrington', project),
+            emeritus('JasonTypesCodes', 'Jason Schindler', project),
+            emeritus('ZacharyKlein','Zachary Klein', project),
+            emeritus('aitortxu','Aitor Alzola', project),
+            emeritus('alexanderzeillinger','Alexander Zeillinger', project),
+            emeritus('alkemist', 'Luke Daley', project),
+            emeritus('alvarosanchez','Álvaro Sánchez-Mariscal', project),
+            emeritus('anshbansal','Aseem Bansal', project),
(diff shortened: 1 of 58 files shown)


jdaugherty:
I reordered the names alphabetically by section.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
```diff
@@ -191,6 +191,156 @@ class PublishPlugin implements Plugin<Project> {
...
+            emeritus('burtbeckwith', 'Burt Beckwith', project),
+            emeritus('davidkron', 'David Kron', project),
+            emeritus('delight', 'Konstantinos Kostarellis', project),
+            emeritus('erawat','Erawat Chamanont', project),
+            emeritus('erichelgeson','Eric Helgeson', project),
+            emeritus('fordguo','Ford Guo', project),
+            emeritus('houbie','Ivo Houbrechts', project),
+            emeritus('jameskleeh', 'James Kleeh', project),
+            emeritus('jbrisbin','Jon Brisbin', project),
+            emeritus('jeffscottbrown', 'Jeff Brown', project),
```
sbglasius, line 243: I am fairly sure, that JSB was one of the founders or joined very early?

sbglasius: I have reached out to him to ask.

sbglasius (approved):

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
sbglasius: JSB Says he was part of G2One from Grails 0.3 IMO that makes him a co-founder 🙂

jdaugherty: The articles we put out didn't list him as a founder.  I'm happy to indicate him as one though.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
```diff
@@ -191,6 +191,156 @@ class PublishPlugin implements Plugin<Project> {
         }
     }
 
+    private static List<MavenPomDeveloper> getProjectDevelopers(Project project) {
```
matrei: I am not a fan of the usage of `get*` method names for non-getters.
Could we name it:
- `findDevelopers(Project project)`
- `buildDeveloperList(Project project)`
- `createDeveloperList(Project project)`
- or something else...

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
```diff
@@ -191,6 +191,156 @@ class PublishPlugin implements Plugin<Project> {
...
+            developer('jdaugherty', 'James Daugherty', project),
+            developer('jpammer', 'Jonas Pammer', project),
+            developer('lhotari', 'Lari Hotari', project),
+            developer('matrei', 'Mattias Reichel', project),
+            developer('paulk','Paul King', project),
+            developer('sbglasius', 'Søren Berg Glasius', project),
+            developer('sdelamo', 'Sergio del Amo', project),
+            developer('tbrasmussen', 'Thomas Rasmussen', project),
+            // Past Developers
+            // - non-apache members go here until their ASF account is created
```
matrei: Does "non-apache members" refer to active contributors?
If that is the case, would it be more fitting with the `Contributor` role, rather than the `Developer Emeritus` role?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
```diff
@@ -191,6 +191,156 @@ class PublishPlugin implements Plugin<Project> {
...
+
+    private static MavenPomDeveloper emeritus(String id, String name, Project project) {
+        pomDeveloper('Developer Emeritus', id, name, project)
+    }
+
+    private static MavenPomDeveloper contributor(String id, String name, Project project) {
+        pomDeveloper('Contributor', id, name, project)
+    }
+
+    private static MavenPomDeveloper pomDeveloper(String role, String id, String name, Project project) {
```
matrei: A developer can have more than one role (and also other properties).
Should we take that into account from the start?
On the other hand, these methods are all `private` so we can easily adjust later if needed.

matrei (approved):

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
jdaugherty: My proposal was to only give them one role for simplicity.  Technically almost all of the founders are inactive (well Graeme has responded when asked directly), but they're not actively maintaining the framework anymore.  I modeled our approach after Groovy's approach.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/PublishPlugin.groovy
jdaugherty: I was trying to imply there are people who have historically contributed a significant contribution (developer emeritus instead of contributor) but they do not have an active ASF account.  Until they create an active ASF account, we wouldn't move them to developer. In other words, it's a requirement to have an ASF account to be considered an active developer.

matrei: But are these "non-apache members" counted as active or not?
If they are active, I don't think "Developer Emeritus" is the right role.
If they are not active, why do we need this comment?

jdaugherty: emeritus means non-active.  I was just trying to inform people on how to be marked developer.  We can remove the comment since it's implied by the developer requirement.

PR #15275, opened 2025-12-05
Issue 14379 remove dependency to tomcat embed logging log4j

### LINKED ISSUE #14379: Remove dependency to tomcat-embed-logging-log4j
It is deprecated and is incompatible with latest tomcat production releases (is not published for 8.5 and 9.0+ branches).

Related to the same issue in grails core https://github.com/grails/grails-core/issues/11631

dauer:
Issue 14379 Remove dependency to tomcat-embed-logging-log4j

Fixes https://github.com/apache/grails-core/issues/14379

The dependency `tomcat-embed-logging-log4j` is old, unmaintained and
should not be used.

Updated Hibernate documentation and related example.

Removed `tomcat-embed-logging-log4j` dependency from GraphQL
The GraphQL code is not active in the grails-core project but I removed
the dependency anyway - it should not be there if we reintroduce the
code again.

### DIFF
--- a/grails-data-hibernate5/docs/src/docs/asciidoc/gettingStarted/outsideGrails.adoc
+++ b/grails-data-hibernate5/docs/src/docs/asciidoc/gettingStarted/outsideGrails.adoc
@@ -17,21 +17,21 @@ specific language governing permissions and limitations
 under the License.
 ////
 
-If you wish to use GORM for Hibernate outside of a Grails application you should declare the necessary dependencies for GORM and the database you are using, for example in Gradle:
+If you wish to use GORM for Hibernate outside a Grails application you should declare the necessary dependencies for GORM and the database you are using, for example in Gradle:
 
 [source,groovy,subs="attributes"]
 ----
-compile "org.apache.grails.data:grails-data-hibernate5-core:{version}"
-runtime "com.h2database:h2:1.4.192"
-runtime "org.apache.tomcat:tomcat-jdbc:8.5.0"
-runtime "org.apache.tomcat.embed:tomcat-embed-logging-log4j:8.5.0"
-runtime "org.slf4j:slf4j-api:1.7.10"
+implementation platform("org.apache.grails:{version}")
+implementation "org.apache.grails.data:grails-data-hibernate5-core"
+runtimeOnly "com.h2database"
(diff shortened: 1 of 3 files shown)


grails-data-hibernate5/docs/src/docs/asciidoc/gettingStarted/outsideGrails.adoc
```diff
@@ -17,21 +17,21 @@ specific language governing permissions and limitations
...
-compile "org.apache.grails.data:grails-data-hibernate5-core:{version}"
-runtime "com.h2database:h2:1.4.192"
-runtime "org.apache.tomcat:tomcat-jdbc:8.5.0"
-runtime "org.apache.tomcat.embed:tomcat-embed-logging-log4j:8.5.0"
-runtime "org.slf4j:slf4j-api:1.7.10"
+implementation platform("org.apache.grails:grails-bom:{version}")
+implementation "org.apache.grails.data:grails-data-hibernate5-core"
+runtimeOnly "com.h2database"
+runtimeOnly "org.apache.tomcat:tomcat-jdbc"
+runtimeOnly "org.slf4j:slf4j-nop"
```
jdaugherty, line 28: Why recommend nop instead of api? Don't we want the slf4j logging?

dauer: I just picked the bare minimum - using `api` is probably a better choise, for a more realistic example.
I wrote a short example for testing, and added `nop` to get rid of this message:
```
> Task :Main.main()
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
```

matrei: Shouldn't `slf4j-api` only be used as `implementation` or `compileOnly` (if it is provided) and only if the slf4j API is used in the implementation of a library. In that case the downstream user chooses the runtime implementation. In this case, I think a slf4j implementation (like `slf4j-nop`) is the right choice.

matrei (approved):

PR #15289, opened 2025-12-14
7.1.x cli add-field and add-property for domain field and property additions

codeconsole:
add-property CLI Command

  Overview

  The add-property command adds a new property to an existing Grails domain class, optionally with validation
  constraints using either Grails constraints or Jakarta Validation annotations.

  Usage

  grails add-property <DOMAIN-CLASS> <FIELD:TYPE> [OPTIONS]

  Arguments

  | Argument     | Description                            | Example                     |
  |--------------|----------------------------------------|-----------------------------|
  | DOMAIN-CLASS | The name of the domain class           | Book, com.example.Author    |
  | FIELD:TYPE   | Field name and type separated by colon | title:String, pages:Integer |

  Supported Field Types

  - String
  - Integer
  - Long
  - Boolean
  - Date
  - BigDecimal
  - Double
  - Float
  - Short
  - Byte
  - Character

  Options

  | Option                     | Description                                                            |
  |----------------------------|------------------------------------------------------------------------|
  | --nullable                 | Mark the field as nullable                                             |
  | --not-nullable             | Mark the field as NOT nullable (generates `@NotNull` or nullable: false) |
  | --blank                    | Allow blank values (String fields only)                                |
  | --not-blank                | Disallow blank values (generates `@NotBlank` or blank: false)            |
  | --max-size <N>             | Maximum size constraint (String fields only)                           |
  | --min-size <N>             | Minimum size constraint (String fields only)                           |
  | --constraint-style <STYLE> | Constraint style: grails (default), jakarta, or both                   |

  Constraint Styles

  | Style   | Description                                                      |
  |---------|------------------------------------------------------------------|
  | grails  | Uses Grails static constraints block (default)                   |
  | jakarta | Uses Jakarta Validation annotations (`@NotNul`l, `@NotBlank`, `@Size`) |
  | both    | Uses both Grails constraints AND Jakarta annotations             |

  Examples

  Basic field (no constraints)

  add-field Book title:String
  Result:
```groovy
  class Book {
      String title
  }
```
  With Grails constraints (default)

  add-field Book title:String --not-nullable --max-size 255
  Result:
```groovy
  class Book {
      String title

      static constraints = {
          title nullable: false, maxSize: 255
      }
  }
```
  With Jakarta Validation annotations

  add-field Book title:String --not-nullable --not-blank --max-size 255 --constraint-style jakarta
  Result:
```groovy
  import jakarta.validation.constraints.NotNull
  import jakarta.validation.constraints.NotBlank
  import jakarta.validation.constraints.Size

  class Book {
      @NotNull
      @NotBlank
      @Size(max = 255)
      String title
  }
```
  With both constraint styles

  add-field Book title:String --not-nullable --max-size 255 --constraint-style both
  Result:
```groovy
  import jakarta.validation.constraints.NotNull
  import jakarta.validation.constraints.Size

  class Book {
      @NotNull
      @Size(max = 255)
      String title

      static constraints = {
          title nullable: false, maxSize: 255
      }
  }
```
  Constraint Mapping

  | Option         | Grails Constraint | Jakarta Annotation |
  |----------------|-------------------|--------------------|
  | --not-nullable | nullable: false   | @NotNull           |
  | --nullable     | nullable: true    | (none)             |
  | --not-blank    | blank: false      | @NotBlank          |
  | --blank        | blank: true       | (none)             |
  | --max-size N   | maxSize: N        | @Size(max = N)     |
  | --min-size N   | minSize: N        | @Size(min = N)     |

  Error Handling

  - Domain class not found: Displays error message suggesting to run create-domain-class first
  - Field already exists: Displays error message indicating the field already exists
  - Invalid field type: Displays error with list of supported types
  - Invalid constraint options: Displays specific validation error (e.g., --blank only valid for String)


grails-forge/grails-forge-cli/build.gradle
```diff
@@ -80,6 +80,8 @@ dependencies {
     annotationProcessor 'io.micronaut:micronaut-inject-java'
 
     implementation project(':grails-forge-core')
+    implementation "org.apache.grails.bootstrap:grails-bootstrap:$projectVersion"
+    implementation "org.codehaus.groovy:groovy:$groovyVersion"
```
sbglasius: Groovy 4.x is org.apache.groovy! This is polluting the classpath.

sbglasius (changes requested):

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
+/**
+ * Utility class for modifying domain class source files to add fields.
+ *
+ * @since 7.0
+ */
+@CompileStatic
+class DomainFieldModifier {
```
jdaugherty, line 39: Can you help me understand why this is in bootstrap and not the CLI itself?  Current files in this project all relate to classes that need shared across grails-core, gradle, and gorm.  I only see you using this in the CLI libraries, why not put it in one of those or add a new project that's shared between the CLI for common dependencies?

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+        if (exactMatch.exists()) {
+            return exactMatch
+        }
+
+        String simpleClassName = className.contains('.') ? className.substring(className.lastIndexOf('.') + 1) : className
+        File found = null
+
+        domainDir.eachFileRecurse { File file ->
+            if (file.name == "${simpleClassName}.groovy") {
+                found = file
```
jdaugherty: what if there are duplicate class names across different packages? I think you had an example where you had an Admin & User view - couldn't multiple domains exist in that case with a different table name configuration?

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+            }
+
+            for (PropertyNode prop : classNode.properties) {
+                if (prop.name == fieldName) {
+                    return true
+                }
+            }
+
+            for (FieldNode field : classNode.fields) {
+                if (field.name == fieldName && !field.name.startsWith('$') && !field.name.startsWith('__')) {
```
jdaugherty: I know Grails uses these prefix conventions for fields that should not be found, but can we at least extract these prefixes into a variable?  Can you check if such a variable already exists so we don't forget to change this file over time?

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+                    return true
+                }
+            }
+
+            return false
+        } catch (Exception e) {
+            return domainFile.text.contains("${fieldName}")
```
jdaugherty: If it refuses to parse, shouldn't we error?  What's the scenario here?

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+        if (field.usesGrailsConstraints()) {
+            String constraintLine = field.toConstraintLine()
+            if (constraintLine) {
+                if (points.hasConstraintsBlock) {
+                    int constraintInsertIndex = points.constraintInsertLine + linesAdded
+                    lines.add(constraintInsertIndex, '        ' + constraintLine)
+                } else {
+                    int constraintBlockIndex = fieldInsertIndex + 1
+                    lines.add(constraintBlockIndex, '')
+                    lines.add(constraintBlockIndex + 1, '    static constraints = {')
```
jdaugherty, line 199: Shouldn't formatting be handled independently of line addition?

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
```diff
@@ -0,0 +1,347 @@
...
+            }
+        }
+        imports
+    }
+
+    /**
+     * Parses the Groovy source file and returns the main class node.
+     */
+    private ClassNode parseClass(File sourceFile) {
+        CompilerConfiguration config = new CompilerConfiguration()
```
jdaugherty, line 250: What's the context of where this is supposed to be used?  We usually used forked threads so not to continue expand the memory of the running process.

grails-bootstrap/src/main/groovy/grails/codegen/model/FieldDefinition.groovy
```diff
@@ -0,0 +1,268 @@
...
+    enum ConstraintStyle {
+        /** Use Grails static constraints block */
+        GRAILS,
+        /** Use Jakarta Validation annotations */
+        JAKARTA,
+        /** Use both Grails constraints and Jakarta annotations */
+        BOTH
+    }
+
+    static final Set<String> SUPPORTED_TYPES = [
```
jdaugherty: Why only the built in types?  Seems like enum would be a reasonable one ...

grails-forge/grails-forge-cli/build.gradle
jdaugherty: this is in the forge-cli, which is still a micronaut app and on groovy 3.

jdaugherty (changes requested):
The only concern I have with this PR is the decision to put it in grails-bootstrap.  I am assuming end user apps won't be generating fields at runtime of a Grails application so this is the wrong spot to place it.

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: the `grails.codegen.model` package is in grails-bootstrap.  This is adding those 2 generation classes to this package that already exists.  This needs to be shared across BOTH clis and profiles.

grails-bootstrap is already a dependency of grails-shell-cli, putting the classes there means:
  - profiles can use it (via shell-cli → bootstrap)
  - shell-cli can use it (already depends on bootstrap)
  - forge-cli can use it (just needs to depend on bootstrap which is part of this PR)

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: added duplicate domain name / different package support

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: done

grails-bootstrap/src/main/groovy/grails/codegen/model/FieldDefinition.groovy
codeconsole: I was intentionally trying to keep this simple for the first phase.  I've expanded it a bit.

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: ok

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: parseClass() is only called by fieldExists() which parses a single domain class file per add-field invocation

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
codeconsole: are you suggesting to reformat the entire file after making changes?

matrei:
This looks OK to me.
I would probably prefer `add-property` or `add-domain-property` instead of `add-field`.

(We could move code generation classes to it's own module in Grails 8 if we want to get them out of `grails-bootstrap`.)

codeconsole:
> This looks OK to me. I would probably prefer `add-property` or `add-domain-property` instead of `add-field`.
> 
> (We could move code generation classes to it's own module in Grails 8 if we want to get them out of `grails-bootstrap`.)

@matrei  I modeled it after [Spring Roo](https://docs.spring.io/spring-roo/docs/current/reference/html/)
In Java, it is obviously adding a field, but in Groovy there seems to be confusion.

Is the Groovy consensus still that it is a property because it has no access modifier?  This is from 2014:
[[groovy-user] Property vs Field](https://groups.google.com/g/groovy-user/c/2OM-_PDklfk?pli=1)

grails-bootstrap/src/main/groovy/grails/codegen/model/DomainFieldModifier.groovy
jdaugherty: I was trying to suggest extracting the spacing instead of hard coding the spaces into the argument.  

i.e. 

        int indent = 5
        "${indent*1}static constraints = {" 

This way you're not hard coding various length spaces and later if we have to change the indent / wrap the code, we don't have to manually modify every line.

matrei:
> Is the Groovy consensus still that it is a property because it has no access modifier?

@codeconsole
https://groovy-lang.org/objectorientation.html#_fields_and_properties

codeconsole:
> > Is the Groovy consensus still that it is a property because it has no access modifier?
> 
> https://groovy-lang.org/objectorientation.html#_fields_and_properties

@matrei 

ok, that is pretty clear.  Thank you. Since there is no access modifier, I agree that we should use add-property for property additions. 

Since add-field could also be useful, I think we should have add-field for fields and add-property for properties.

add-field now _**requires**_ passing an access modifier.

matrei:
> Since add-field could also be useful, I think we should have add-field for fields and add-property for properties.

@codeconsole What is the use case for adding a field to a domain class, and does that use case warrant a shell command?

codeconsole:
> > Since add-field could also be useful, I think we should have add-field for fields and add-property for properties.
> 
> @codeconsole What is the use case for adding a field to a domain class, and does that use case warrant a shell command?

@matrei well, it's kind of a freebie due to inheritance.  The original premise behind this PR was to slowly introduce codegen capabilities to where you could build a workable demo with cli commands.   I've had use cases for static fields in my domain objects. 

I suppose I am indifferent.  Adding properties is what is most important.   Should we get rid of add-field altogether?

matrei:
> > > Since add-field could also be useful, I think we should have add-field for fields and add-property for properties.
> > 
> > 
> > @codeconsole What is the use case for adding a field to a domain class, and does that use case warrant a shell command?
> 
> @matrei well, it's kind of a freebie due to inheritance. The original premise behind this PR was to slowly introduce codegen capabilities to where you could build a workable demo with cli commands. I've had use cases for static fields in my domain objects.
> 
> I suppose I am indifferent. Adding properties is what is most important. Should we get rid of add-field altogether?

@codeconsole My preference would be to remove `add-field` as it might make things more complicated for a user who has to think/remember which one to pick.

codeconsole:
@matrei ok, I will get rid of it.

matrei (approved):

PR #15307, opened 2026-01-08
Update default application start page

matrei:
Suggestion for re-haul of the default application start page.

Attaching rendered files for review:
[Welcome_to_Grails_files.zip](https://github.com/user-attachments/files/24563145/Welcome_to_Grails_files.zip)


jamesfredley (approved):

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
```diff
@@ -41,7 +41,7 @@
     <li class="nav-item dropdown">
         <a href="#" class="nav-link dropdown-toggle" role="button" data-bs-toggle="dropdown" aria-expanded="false">Installed Plugins<span class="caret"></span></a>
         <ul class="dropdown-menu dropdown-menu-right">
-            <g:each var="plugin" in="${applicationContext.getBean('pluginManager').allPlugins}">
+            <g:each var="plugin" in="${applicationContext.getBean('pluginManager').allPlugins.sort { it.name.toLowerCase() }}">
```
jdaugherty: Isn't the ordering actually important?  It's the order the plugins loaded which can be used to debug issues if there are conflicts between plugins?

jdaugherty: We could call withIndex() to inject the original order.  Then you can have a title to show the original order - this way the view shows them sorted, but it's still possible to easily get the original order.

matrei:
@jdaugherty Valid point. How about this?

Screenshot_20260108_175146

jeffscottbrown:
@matrei FYI... For that screenshot, I would expect periodically to see questions from folks asking what the numbers in the parens represent.

jdaugherty:
I'd suggest just prefixing the numbers with '#' to signify it's an ordering, otherwise it looks good.

jdaugherty (approved):

matrei:
> FYI... For that screenshot, I would expect periodically to see questions from folks asking what the numbers in the parens represent.

@jeffscottbrown Yes, you are right. I have updated this PR with bigger change of the default application start page.

matrei:
@jamesfredley @jdaugherty 
After feedback I thought about making a bigger update to the default application start page in Grails 7.1.0.
There is a rendered static html version attached in the PR description for review/feedback.

jdaugherty:
@matrei  given the new changes, would you mind including a screenshot of what this is supposed to look like?

matrei:
> @matrei given the new changes, would you mind including a screenshot of what this is supposed to look like?

@jdaugherty 
Screenshot_20260112_161422

jdaugherty:
I merged the changes from 7.0.x up to 7.1/8 and then updated this PR for 7.1.x

grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
```diff
@@ -1,81 +1,86 @@
...
-                    </h6>
-                    <p class="card-text">Get feedback and share your experience with other Grails developers in the community <a href="https://slack.grails.org" target="_blank">Slack channel</a>.</p>
-                </div>
+            <div class="col-12 col-md-4">
+                <a class="card h-100 text-decoration-none shadow-sm border-1"
+                   href="https://slack.grails.org" target="_blank" rel="noopener">
+                    <div class="card-body p-4">
+                        <div class="d-flex align-items-center justify-content-between mb-2">
+                            <h6 class="card-title mb-0 fw-semibold">Join the Community</h6>
+                            <asset:image src="slack.svg" alt="Grails Slack" width="34" height="34"/>
```
jdaugherty, line 67: Should we reference the mailing lists, github, & slack?

jdaugherty (commented):
The render looks great and a basic look at the changes look great too.

Also ...
* Thank you for using SVG files to reduce our build sizes.
* Thank you for not applying the license headers since these are templates any app will receive. 
* We may want to consider updating references to slack to point people to the mailing lists / github discussions as well.

matrei:
Replaced by #15317

PR #15309, opened 2026-01-09
Add missing application.yml examples to docs

### LINKED ISSUE #15167: Add missing application.yml examples for each application.groovy example in documentation
There are many locations in the documentation with application.groovy examples.  Since the default in generated application is application.yml, we should add application.yml examples in each location.

Ajay9704:
This PR addresses issue #15167 by adding missing application.yml examples for every application.groovy configuration example in the Grails documentation. This ensures users have examples in both configuration formats, which is particularly helpful since the default generated application uses application.yml.

Key Improvements:
Every configuration example that had only Groovy code now has a corresponding YAML example
All YAML examples follow proper indentation and Grails configuration conventions
The documentation is now consistent with Grails' default use of application.yml in new projects
All changes maintain the existing documentation structure and style
Comprehensive coverage across all major configuration areas

### DIFF
--- a/grails-doc/src/en/guide/REST/domainResources.adoc
+++ b/grails-doc/src/en/guide/REST/domainResources.adoc
@@ -88,6 +88,21 @@ grails.mime.types = [
 ]
 ----
 
+Or the equivalent in `application.yml`:
+
+[source,yaml]
+----
+grails:
+    mime:
+        types:
+            json:
+              - 'application/json'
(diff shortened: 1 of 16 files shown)


jamesfredley (approved):

jamesfredley:
@Ajay9704 Thank you, this is excellent and will be very helpful for end users.

Issue #14194, opened 2025-03-31
Revisit gsp / web project structure


jdaugherty:
https://github.com/apache/grails-core/pull/14093#discussion_r2020168768 

There is not a clear dependency chain of projects to implement gsp.  We should pull up classes where they're used by parent libraries and make it clear which projects need included for which functionality.

PR #15320, opened 2026-01-17
Groovy 4.0.30, staged groovy verison support, multiple repo support in GRAILS_REPO_URL, & profile support for GRAILS_REPO_URL

jdaugherty:
I think these are the required changes to support a staged version of groovy.  This PR also switches us to 4.0.30 - we shouldn't merge it until Groovy is released, but this is a good time to test staged releases used in dev.


jdaugherty:
More work will need to be done to allow groovy staging builds to be included - the generated applications can't find the staging repo since it's not set.

jdaugherty:
finally decided to just create a repo settings plugin in build-logic; this allows us to centrally define the repositories for all of grails projects in core in `build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsRepoSettingsPlugin.groovy`  @jamesfredley I think we've discussed this before and hopefully this makes it easier to maintain.

For the forge tests, I'm not sure what to do.  We already define an override for GRAILS_REPO_URL so the tests pickup the local grails-core code.  We need to leave that, but somehow we need to either add the staging repo or our repo settings plugin.  I'm not sure if we should publish our repo settings plugin since that could lead to a burden on repository.apache.org @matrei  do you have any ideas here? 

For now, I think when groovy 4.0.30 releases we should merge this PR either way. We can address the forge issue longer term.

jdaugherty:
We could also just support multiple repos in GRAILS_REPO_URL by using a separator ...

jamesfredley:
@jdaugherty For Staged Groovy version support in the forge generated applications, I think we should make the same repositories changes in:

https://github.com/apache/grails-core/blob/7.0.x/grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java#L41-L62

This will only impact Grails snapshots.

jdaugherty:
> @jdaugherty For Staged Groovy version support in the forge generated applications, I think we should make the same repositories changes in:
> 
> 
> 
> https://github.com/apache/grails-core/blob/7.0.x/grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java#L41-L62
> 
> 
> 
> This will only impact Grails snapshots.

So duplicate the change instead of importing the plugin?

jamesfredley:
I see that plugin now, if it can be used in the grails-forge generated application, then let's use it.

jdaugherty:
> I see that plugin now, if it can be used in the grails-forge generated application, then let's use it.

I've added support to profiles for the GRAILS_REPO_URL & also made the GRAILS_REPO_URL be able to have multiple repos by using `;` as a separator.  

As for how to include that settings plugin, we'd have to publish it.  The more I thought about this, I think we can just get away pointing at the staging repo url as a secondary url.  What are your thoughts on this?

jdaugherty:
I've also fixed the grails wrapper from defaulting to the ASF repo - it will still use it as a last resort, but this should alleviate any load on RAO

jdaugherty:
For now, i've hard coded the staging repo on the forge test as a secondary repo.  This causes the tag lib to pass locally (prior to groovy 4.0.30 publish): 

image

jamesfredley:
@jdaugherty I believe adjusting the regexp filter https://github.com/apache/grails-core/blob/7.0.x/grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java#L52 will resolve this for grails-forge generated test apps.

jdaugherty:
This change will also fix: https://github.com/apache/grails-core/issues/14887

jdaugherty:
So @jamesfredley & I figured out that overriding the groovy version does not work when using the spring dependency management plugin for 2 reasons: 
1. the apply exclusions was set to false (due to bug https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/368)
2. for the exclusions to work, specific versions must be overridden in downstream boms

I added all of the groovy-bom dependencies to our bom so the dependency management plugin works correctly when we upgrade to a version of groovy newer than spring boot.

jdaugherty:
One other change this PR addresses, is it further refines the expected release type for snapshot builds.  The build.gradle of a forge generated app (from a grails snapshot) will now have the following: 

      repositories {
          maven {
             url = 'file:/Users/jdaugherty/sandbox/external/grails-core/build/local-maven'
          }
          mavenCentral()
          maven {
             url = 'https://repo.grails.org/grails/restricted'
          }
          maven {
             url = 'https://repository.apache.org/content/groups/snapshots'
             content {
                includeVersionByRegex('org[.]apache[.]grails.*', '.*', '.*-SNAPSHOT')
             }
             content {
                includeVersionByRegex('org[.]apache[.]groovy.*', 'groovy.*', '.*-SNAPSHOT')
             }
             mavenContent {
                snapshotsOnly()
             }
          }
          maven {
             url = 'https://repository.apache.org/content/groups/staging'
             content {
                includeVersionByRegex('org[.]apache[.]grails[.]gradle', 'grails-publish', '.*')
             }
             content {
                includeVersionByRegex('org[.]apache[.]groovy.*', 'groovy.*', '.*')
             }
             mavenContent {
                releasesOnly()
             }
          }
      }

jdaugherty:
Forge tests are now failing because of asset compilation failing: 

         > Task :assetCompile FAILED
         Exception in thread "main" java.lang.NoClassDefFoundError: groovy/json/JsonSlurper
                 at asset.pipeline.AssetCompiler.main(AssetCompiler.groovy:228)
         Caused by: java.lang.ClassNotFoundException: groovy.json.JsonSlurper
                 at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
                 at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
                 at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
                 ... 1 more

jdaugherty:
So it looks like the exclusions added to grails-gsp are now being properly applied.  The side effect is several of the libraries that are included by Gradle are not being included when using spring dependency management plugin.

These appear the ones needed under grails-gsp: 
1. grails-encoder exports groovy-json 
2. grails-web-common exports spring-webmvc

Removing the excludes fixes the interceptor tests.  I suspect we'll find more as we test.

jdaugherty:
@matrei @jamesfredley this is finally passing after i discovered the exclusions now get applied when using the dependency management plugin.

jdaugherty:
Fixes: https://github.com/apache/grails-core/issues/15324

jdaugherty:
I cleaned up the commit history / squashed all of the WIP.  Also created tickets to document this work for 7.0.6

matrei:
> So it looks like the exclusions added to grails-gsp are now being properly applied. The side effect is several of the libraries that are included by Gradle are not being included when using spring dependency management plugin.
> 
> These appear the ones needed under grails-gsp:
> 
>     1. grails-encoder exports groovy-json
> 
>     2. grails-web-common exports spring-webmvc
> 
> 
> Removing the excludes fixes the interceptor tests. I suspect we'll find more as we test.

I'm not sure this is the right solution.

Why is it the responsibility of `grails-encoder` to provide `groovy-json` via `grails-gsp`, even though it is not used in `grails-gsp` API scope?
And why is it the responsibility of `grails-gsp` to provide `spring-webmvc` from `grails-web-common`?

What if the application does not use the GSP plugin?

I think we can do this workaround for now, but keep the exclusions (commented out) with a comment explaining why they are not excluded. I'll add that change for review.

jdaugherty:
> > So it looks like the exclusions added to grails-gsp are now being properly applied. The side effect is several of the libraries that are included by Gradle are not being included when using spring dependency management plugin.
> > These appear the ones needed under grails-gsp:
> > ```
> > 1. grails-encoder exports groovy-json
> > 
> > 2. grails-web-common exports spring-webmvc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Removing the excludes fixes the interceptor tests. I suspect we'll find more as we test.
> 
> I'm not sure this is the right solution.
> 
> Why is it the responsibility of `grails-encoder` to provide `groovy-json` via `grails-gsp`, even though it is not used in `grails-gsp` API scope? And why is it the responsibility of `grails-gsp` to provide `spring-webmvc` from `grails-web-common`?
> 
> What if the application does not use the GSP plugin?
> 
> I think we can do this workaround for now, but keep the exclusions (commented out) with a comment explaining why they are not excluded. I'll add that change for review.

I think the original intent was web-common would provide common web libraries, but we've added excludes.  We definitely have the various dependency "starters" that say we provide them.  It seems that gsp or the gson projects are where these have historically come from.  We need to revisit on what we exclude; we don't want end apps having to include some of these lower libraries.  All of the gsp projects need re-architected to make it clear what's being provided where.

jdaugherty:
As for adding them back, I'm fine with that.  @matrei are you ok with this change then?

matrei (approved):

jdaugherty:
@matrei  FYI: we have https://github.com/apache/grails-core/issues/14194 ; i'm trying to go through and mark some of these tickets as "design-required" so we can discuss and focus on them.  We'll probably want to do this in the next 6 months as part of the Grails 8 work?

jamesfredley (approved):
I reviewed the base and web profiles, tested wrapper with snapshot and release versions and audited the generated POMs.  

Summary of Changed Exclusions

grails-gsp POM (grails-gsp-7.0.6-SNAPSHOT.pom)

Change #1: grails-encoder dependency
Dependency: org.apache.grails.web:grails-encoder:7.0.6-SNAPSHOT
7.0.5 (6 exclusions):
- ❌ org.apache.groovy:groovy-json ← REMOVED IN 7.0.6
- ❌ org.slf4j:slf4j-api
- ❌ org.springframework:spring-web
- ❌ org.slf4j:jcl-over-slf4j
- ❌ org.apache.grails:grails-core
- ❌ org.apache.groovy:groovy

7.0.6 (5 exclusions):
- ❌ org.slf4j:slf4j-api
- ❌ org.springframework:spring-web
- ❌ org.slf4j:jcl-over-slf4j
- ❌ org.apache.grails:grails-core
- ❌ org.apache.groovy:groovy
- 
Impact: groovy-json will now be included as a transitive dependency

---
Change #2: grails-web-common dependency

Dependency: org.apache.grails.web:grails-web-common:7.0.6-SNAPSHOT

7.0.5 (6 exclusions):
- ❌ org.apache.grails:grails-databinding
- ❌ org.apache.grails:grails-core
- ❌ org.apache.grails:grails-encoder
- ❌ org.springframework:spring-webmvc ← REMOVED IN 7.0.6
- ❌ org.springframework:spring-context-support
- ❌ org.apache.groovy:groovy-templates

7.0.6 (5 exclusions):
- ❌ org.apache.grails:grails-databinding
- ❌ org.apache.grails:grails-core
- ❌ org.apache.grails:grails-encoder
- ❌ org.springframework:spring-context-support
- ❌ org.apache.groovy:groovy-templates
- 
Impact: spring-webmvc will now be included as a transitive dependency

jdaugherty:
FYI: The latest groovy versions haven't sync'd to maven central yet.  There's a maven central request open (`115831`) to address this issue. Once it's fixed, I'll merge this.

PR #15323, opened 2026-01-18
fix:negative number bug-15178

### LINKED ISSUE #15178: Negative number bug with gsp and fields and Norwegian locale in browser.
### Expected Behavior

It should be straight forward entering and editing negative numbers for both Long and BigDecimal fields.

### Actual Behaviour

Long numbers:
I have not been able to figure a way to enter negative long numbers.

When editing a long that was previously negative it is not pre-filled, an empty field is displayed

BigDecimal numbers
Can be entered and edited using using typographic minus (ordinal value e28892). 

### Steps To Reproduce

_No response_

### Environment Information

- Ubuntu 24.04
- java 17.0.17
- Grails 7.0.0 (same issue on grails 5.3.2)

### Example Application

https://github.com/yggdrasildy/negativenumbers

### Version

7.0.0

Ajay9704:
Fix: Negative number bug with GSP and fields in Norwegian locale (#15178)

Problem:
- Cannot enter negative Long numbers in Norwegian locale
- Editing existing negative Long shows empty field in Norwegian locale

Solution:
- Added normalizeMinusSign() method to convert Unicode minus (−) to ASCII minus (-)
- Applied fix to ValidationTagLib, FormFieldsTagLib, and FormatTagLib
- Added test cases to verify the fix works correctly

Symptoms addressed:
- Negative Long numbers can now be properly entered and edited
- Editing existing negative Long numbers shows correct pre-filled values


Ajay9704:
@jdaugherty  can you trigger the checks ? . If there are anything that I should modify it will be better in this draft mode , once I am done, I will  change to ready mode .

Issue #14991, opened 2025-08-18
Fix broken links in generated groovydocs


jamesfredley:
### Issue description

Broken links can be identified and verified as fixed using https://github.com/lycheeverse/lychee/releases/.  

```
./gradlew grails-doc:build
cd \grails-doc\build\docs\
lychee-x86_64-windows.exe . --format markdown --output results.md --exclude ".*.css$" --exclude "^http://localhost"
```

review `results.md`

jamesfredley:
@sanjana2505006

sanjana2505006:
Hello @jamesfredley, 
thank you for the ping! I’ll start looking into the broken groovydocs links and follow the steps outlined to identify and fix them.

sanjana2505006:
I’ve raised PR #15327 for this. Please let me know if you have any feedback or would like any changes I’d be happy to update it.

PR #15327, opened 2026-01-21
docs: fix broken external and internal links in Groovydocs

sanjana2505006:
# Changes i did:

1. Surgical Link Correction: Modernized the `docs-dependencies.gradle`
 resolution logic and added a professional post-processing script (`fix_groovydoc_links.py`) to repair malformed relative paths for inner classes and map 20+ external API roots (Spring, Hibernate, Jakarta, GORM, etc.).

1. Verification Tooling: Included a parallelized `check_links.py` utility for high-speed documentation validation.

This PR fixes issue #14991

### DIFF
--- a/TEST_FAILURES.md
+++ b/TEST_FAILURES.md
@@ -1,0 +1,25 @@
+# Test Failures Summary
+Generated on: 2026-04-18 00:42:05
+
+Found 12 failures.
+
+## Module: grails-data-hibernate7-dbmigration
+| Class | Test | Type | Message |
+| :--- | :--- | :--- | :--- |
+| liquibase.ext.hibernate.snapshot.HibernatePrimaryKeySnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateForeignKeySnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateViewSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateSequenceSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateSchemaSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateIndexSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.extension.TableGeneratorSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateTableSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateCatalogSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration |
+| liquibase.ext.hibernate.snapshot.HibernateUniqueConstraintSnapshotGeneratorSpec | initializationError | java.lang.IllegalStateException | java.lang.IllegalStateException: Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration |
+
+## Module: grails-test-examples-gorm
+| Class | Test | Type | Message |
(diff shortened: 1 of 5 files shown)


sanjana2505006:
A small number of residual 'broken' links remain (approx. 30 reports). 
My investigation confirms these are artifacts of the Groovydoc generator itself, they are 'phantom' links to generic type parameters (e.g., T1.html) or internal compiler classes that the tool references in HTML but does not actually generate as files. These do not represent errors in the documentation configuration.

jamesfredley:
@sanjana2505006 Thank you.  On this issue would it be possible to use the python scripts to identify the source of the broken link and fix them either in the javadoc/groovy doc in each .groovy or .java file and Gradle configuration for the build?  This is preferable to having a script fix them each time the project is built.

After that, if we still need to have a script, since our project is based on the Groovy programming language we prefer Groovy Scripts vs Python scripts.  I don't think this should be need though, if we fix the root sources of the broken links.  The broken references just built up over the last 20 years as the whole Java ecosystem has evolved significantly.

sanjana2505006:
@jamesfredley Thank you for the clarification. I’ve updated the PR to align with your suggestions.
I fixed the broken links at the source by adjusting visibility where Groovydoc couldn’t generate pages, expanded the gdoc.link configuration in docs-dependencies.gradle to cover Hibernate/GORM/Jakarta EE, and removed the Python scripts entirely. Any remaining fixes are now handled via a native Groovy Gradle task.

I’ve also squashed everything into a single commit. The docs now build cleanly with improved link resolution using the project’s native tooling.
Please let me know if you’d like anything adjusted further.

sanjana2505006:
@jamesfredley Whenever you get a chance, please review and let me know if any changes are needed

sanjana2505006:
Hello @jamesfredley, just a gentle follow-up on this PR. Whenever you have time, could you please share your feedback or let me know if any further changes are needed from my side? Happy to update anything required. Thank you

jdaugherty:
There are license failures on this branch and unnecessary formatting to
unrelated areas.  Can you please fix these so we can review?

sanjana2505006:
@jdaugherty I've updated the PR, please take a look whenever you get a chance and let me know if there's anything else that needs to be changed.

grails-doc/build.gradle
```diff
@@ -121,6 +122,11 @@ combinedGroovydoc.configure { Groovydoc gdoc ->
     gdoc.outputs.dir(gdoc.destinationDir)
 }
 
+def fixGroovydocLinks = tasks.register('fixGroovydocLinks', FixGroovydocLinksTask) {
```
jdaugherty, line 125: It's fine to write a task to help identify bad links, but we shouldn't be "patching" our docs like this.  We should fix the source issue. The task needs to be changed to an audit task or even a build failure that can be triggered via a separate CI (like we do CI or the version mismatches)

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/Tokens.java
```diff
@@ -19,23 +19,24 @@
...
 /**
- * NOTE: Based on work done by on the GSP standalone project (https://gsp.dev.java.net/)
+ * NOTE: Based on work done by on the GSP standalone project
+ * (https://gsp.dev.java.net/)
  *
  * Interface defining an enumeration of tokens for the different scriptlet types
  *
  * @author Troy Heninger
  * @author Graeme Rocher
  *
```
jdaugherty, line 29: The formatting changes to this file are unnecessary and should be reverted.

grails-doc/build.gradle
sanjana2505006: I've converted the task into an audit-only check that fails the build (via `-PauditGroovydocLinks`) instead of patching the HTML.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/Tokens.java
sanjana2505006: I've reverted the unrelated formatting changes and squashed everything into a single clean commit. Let me know what you think!

grails-doc/build.gradle
jdaugherty: No reason to use a property to control behavior.  Let's just use a task that's well named?

sanjana2505006: I have refactored the task 

1. Renamed the task class to `AuditGroovydocLinksTask` and the task instance to `auditGroovydocLinks`.
2. Removed the project property check; the task now performs an audit by default.
3. Removed all patching logic. The task now only verifies the links and fails the build if violations are detected, forcing us to fix the source issues instead of post-processing the HTML.
Let me know if the new structure looks good.

TEST_FAILURES.md
```diff
@@ -0,0 +1,25 @@
+# Test Failures Summary
```
jdaugherty, line 1: What generated this? Let's remove it from being checked in.

gradle/docs-dependencies.gradle
```diff
@@ -31,16 +31,10 @@ dependencies {
...
 String resolveProjectVersion(String artifact) {
-    String version = configurations.runtimeClasspath
-            .resolvedConfiguration
-            .resolvedArtifacts
-            .find {
-                it.moduleVersion.id.name == artifact
-            }?.moduleVersion?.id?.version
-    if (!version) {
-        return null
+    def component = configurations.runtimeClasspath.incoming.resolutionResult.allComponents.find {
```
jdaugherty, line 34: Are these changes necessary?

grails-doc/build.gradle
```diff
@@ -278,7 +284,7 @@ createReleaseDropdownTask.configure {
 
 def docsTask = tasks.register('docs', Sync)
 docsTask.configure { Sync it ->
-    it.dependsOn(combinedGroovydoc, createReleaseDropdownTask, ':grails-data-docs-stage:docs')
+    it.dependsOn(combinedGroovydoc, auditGroovydocLinks, createReleaseDropdownTask, ':grails-data-docs-stage:docs')
```
jdaugherty, line 287: The audit should be a task we run, but it shouldn't be run by default unless we're going to capture it in the workflow summary of the build.

sanjana2505006:
I have an exam shortly, but I'll make sure to look into this and resolve this as a priority once I'm done

borinquenkid:
Superseded by #16005, which rebases this work onto current 7.0.x and addresses the outstanding review feedback (removes the stray TEST_FAILURES.md, and decouples auditGroovydocLinks from the default docs build per @jdaugherty's last comment so it can't fail future docs builds on a false positive). Original commits and authorship preserved. Thank you @sanjana2505006 for the original fix and the persistence through several rounds of review — closing this one in favor of #16005 so it doesn't keep going stale.

Issue #14319, opened 2025-04-19
Cannot enable parallel test execution in grails-gsp


jdaugherty:
### Issue description

While working on the performance of the test run, I noticed that grails-gsp (the plugin) forces the max test parallel to 1.  Changing this locally to be parallel, it succeed.  Changing it in the CI to be parallel fails with the following: 

        RestfulReverseUrlRenderingTests > testFormTagRendering FAILED
         Condition not satisfied:
         output == '<form action="/book" method="post" name="myForm" id="myForm" >save</form>'
         | |
         | false
         | 8 differences (90% similarity)
         | <form action="/(restfulB)ook" method="post" name="myForm" id="myForm" >save</form>
         | <form action="/(b-------)ook" method="post" name="myForm" id="myForm" >save</form>
         <form action="/restfulBook" method="post" name="myForm" id="myForm" >save</form>
         at org.grails.web.mapping.RestfulReverseUrlRenderingTests.testFormTagRendering(RestfulReverseUrlRenderingTests.groovy:30)
        FormTagLibTests > testHiddenFieldWithZeroValue PASSED
        RestfulReverseUrlRenderingTests > testFormTagRenderGETRequest FAILED
         Condition not satisfied:
         output == '<form action="/book" method="get" name="myForm" id="myForm" >create</form>'
         | |
         | false
         | 8 differences (90% similarity)
         | <form action="/(restfulB)ook" method="get" name="myForm" id="myForm" >create</form>
         | <form action="/(b-------)ook" method="get" name="myForm" id="myForm" >create</form>
         <form action="/restfulBook" method="get" name="myForm" id="myForm" >create</form>
         at org.grails.web.mapping.RestfulReverseUrlRenderingTests.testFormTagRenderGETRequest(RestfulReverseUrlRenderingTests.groovy:40)

jdaugherty:
For now i've ignored these tests in CI

jdaugherty:
We have applied the develocity plugin which autoretries tests.  The downside is it's causing our test runs to be longer.

PR #15335, opened 2026-01-24
Flaky Build Test Cleanup

jamesfredley:
## Summary

This PR optimizes CI workflow performance, fixes multiple flaky tests caused by static state pollution during parallel test execution, and adds developer experience improvements. After fixing the root causes of test flakiness, Develocity test retries have been removed.

### Build Performance
- Use `--rerun-tasks` flag for reliable CI builds
- Disable test caching in CI via `outputs.cacheIf { !isCiBuild }`
- Disable Groovy compilation cache in CI (`GroovyCompile` tasks) to ensure AST transformations are always applied
- Increase `maxParallelForks` from 3 to 4 (matches GitHub runner cores)
- Increase `forkEvery` from 10-25 to 50 (reduces JVM restart overhead)
- Enable Develocity test distribution for CI builds
- Remove hardcoded `maxParallelForks = 1` restriction from grails-gsp module
- **Enable `cancel-in-progress` for PR workflow runs** - Cancels outdated workflow runs when new commits are pushed to a PR across all CI workflows (`gradle.yml`, `codestyle.yml`, `codeql.yml`, `groovy-joint-workflow.yml`, `rat.yml`)
- Add new `buildRerunTasks` job for building Grails-Core with rerun tasks enabled

### CI Matrix Reduction
- **Build job**: Changed from full OS/Java cross-product to specific combinations:
  - Ubuntu: Java 17, 21, 25
  - macOS: Java 21 only
  - Windows: Java 25 only
- **Other jobs** (`buildGradle`, `functional`, `mongodbFunctional`, `hibernate5Functional`): Changed from `[17, 21]` to `[17, 25]`
- **Grails Forge build**: Changed to `[17, 21]`

### Test Retry Removal
After fixing the flaky tests, Develocity test retries have been removed from:
- `gradle/test-config.gradle` - Removed testRetry block
- `build-logic/docs-core/build.gradle` - Removed develocity testRetry block
- `grails-data-hibernate5/dbmigration/build.gradle` - Removed entire testRetry task configuration
- `grails-gradle/gradle/test-config.gradle` - Removed develocity testRetry block
- `grails-forge/buildSrc/build.gradle` - Removed unused test-retry plugin dependency
- `grails-forge/gradle/test-config.gradle` - Removed commented-out retry code and plugin apply

### Root Cause Analysis

When tests run in parallel (`maxParallelForks > 1`), static state can leak between tests causing intermittent failures. The main culprits identified:

1. **`HttpServletResponseExtension.@mimeTypes`** - Static cache initialized by one test used by another
2. **`HttpServletResponseExtension.getMimeTypes()`** - Was calling `webRequest.applicationContext` without null check, causing NPE when `GrailsWebRequest.lookup()` returns null during parallel test runs
3. **`HttpServletRequestExtension.getMimeTypes()`** - Was using servlet context attributes which get polluted by parallel tests due to `GrailsUnitTest` static fields
4. **`Holders` state** - `grailsApplication`, `pluginManager` cleared mid-test by parallel cleanup
5. **`Validateable.constraintsMapInternal`** - Constraint cache evaluated before `doWithConfig()` registers shared constraints
6. **`ConstraintEvalUtils.defaultConstraintsMap`** - Global shared constraints cache keyed by config identity hash
7. **`SystemPropertyTenantResolver`** - System property `gorm.tenantId` pollution between multi-tenancy tests
8. **`RequestContextHolder`** - ThreadLocal request attributes not properly reset
9. **`GrailsApplication.artefactInfo`** - URL mapping artefacts polluting parallel tests
10. **`LinkGenerator.urlMappingsHolder`** - Stale reference to old URL mappings holder after bean recreation

### Key Fix Pattern: Holders.clear() in Cleanup

**Problem**: Calling `Holders.clear()` in `cleanup()` or `@AfterEach` clears ALL Holders state (including `grailsApplication`) which can affect parallel tests that depend on `Holders.grailsApplication` for mime type resolution.

**Solution**:
- Use targeted cleanup (only reset what you set)
- Or use `cleanupSpec()`/`@AfterAll` which runs after all tests in the class complete
- The `GrailsUnitTest` trait already handles `Holders.clear()` in `cleanupClass()`

### Key Fix Pattern: Ensure grailsApplication Initialization Before Clearing Caches

**Problem**: Clearing static caches in `setup()` before `grailsApplication` is initialized causes `doWithConfig()` to not be called, meaning custom MIME types and shared constraints are not registered.

**Solution**: Access the `config` property first in `setup()` to trigger lazy initialization of `grailsApplication`, which calls `doWithConfig()`. Then clear the caches so they will be repopulated from the fresh config.

#### 1. MIME Type Lookup Isolation (Production Code Fixes)

**`HttpServletResponseExtension.getMimeTypes()`** was calling `webRequest.applicationContext` without checking if `webRequest` is null. When `GrailsWebRequest.lookup()` returns null during parallel test execution, this caused a NullPointerException.

**Fix**: Added null-safe operator (`?.`) when accessing `webRequest.applicationContext`.

**`HttpServletRequestExtension.getMimeTypes()`** was using `WebApplicationContextUtils.getWebApplicationContext(request.servletContext)` to look up MIME types. This fails in parallel tests because `GrailsUnitTest` uses static `_servletContext` which gets polluted across tests.

**Fix**: Now prioritizes `GrailsWebRequest.lookup()` (thread-local) before falling back to servlet context lookup.

#### 2. Mime Type Cache Isolation (~20 files)
Added `setup()` and `cleanup()` methods to clear `HttpServletResponseExtension.@mimeTypes` before/after each test, ensuring `grailsApplication` is initialized first:
- `RespondMethodSpec`, `WithFormatContentTypeSpec`, `ContentFormatControllerTests`
- `ContentNegotiationSpec`, `JSONBindingToNullSpec`, `MimeUtilitySpec`
- `MimeTypesConfigurationSpec`, `AcceptHeaderParserSpec`, `RequestAndResponseMimeTypesApiSpec`
- `HalJsonRendererSpec`, `JsonRendererSpec`, `BaseDomainClassRendererSpec`
- `RestfulControllerSpec`, `WithFormatSpec`, `DefaultXmlRendererSpec`
- `VndErrorRenderingSpec`, `HtmlRendererSpec`, `DefaultRendererRegistrySpec`
- And others across modules

#### 3. Multi-Tenancy Test Isolation (~18 files)
Used Spock `@RestoreSystemProperties` annotation for cleaner system property isolation, combined with clearing `SystemPropertyTenantResolver`:
- `DatabasePerTenantSpec`, `SchemaPerTenantSpec`, `PartitionMultiTenancySpec`
- `MultiTenancySpec`, `SingleTenantSpec`, `SchemaMultiTenantSpec`
- `MongoStaticApiMultiTenancySpec`, `SchemaBasedMultiTenancySpec`, `SingleTenancySpec` (MongoDB)
- `MongoConnectionSourcesSpec`, `MultiTenancySpec` (MongoDB)
- `CurrentTenantTransformSpec`, `TenantServiceSpec`, `MultiTenantServiceTransformSpec`
- And Hibernate/MongoDB connection specs

#### 4. Command Object Constraint Isolation (~15 files)
Tests that use shared constraints via `doWithConfig()` need proper cache clearing. Uses reflection to clear the private `constraintsMapInternal` field in the Validateable trait:
- `CommandObjectsSpec.groovy`, `CommandObjectNoDataSpec.groovy`
- `ValidateableTraitSpec.groovy`, `ValidateableTraitAdHocSpec.groovy`
- `CommandObjectConstraintGettersSpec`, `SerializableValidateableSpec`
- `ValidateableMockSpec`, `DefaultASTValidateableHelperSpec`
- `JsonApiSpec`, `CommandObjectNullabilitySpec`
- `ControllerUnitTestMixinTests`, `JsonBindingWithExceptionHandlerSpec`
- And others

**Note**: A public API method `clearConstraintsMapCache()` will be added in Grails 7.1 (see PR #15346). This PR uses reflection to access the Groovy 4 trait static field helper interface to clear the cache without modifying the public API, which would be a breaking change in a patch release.

#### 5. Holders.clear() Fixes (4 files)
Replaced aggressive `Holders.clear()` with targeted cleanup:
- `WithFormatContentTypeSpec` - Removed `Holders.clear()` from `cleanup()`
- `BeanBuilderTests` - Changed to `Holders.setPluginManager null` (only reset what setUp sets)
- `DomainClassTraitSpec` - Moved `Holders.clear()` from `cleanup()` to `cleanupSpec()`

#### 6. RequestContextHolder Cleanup (~10 files)
Added `RequestContextHolder.resetRequestAttributes()` in setup/cleanup:
- `LinkGeneratorWithUrlMappingsSpec`, `EncodePathFromURISpec`
- `HTMLJSCodecIntegrationSpec`, `DefaultGrailsApplicationAttributesTests`
- `FlashScopeWithErrorsTests`, `StreamCharBufferSpec`, `WebUtilsTests`
- `BindingOutsideRequestSpec`
- `UrlMappingsWithHttpMethodSpec`, `RestfulUrlMappingSpec`

#### 7. URL Mapping Artefact Isolation (~4 files)
Clear URL mapping artefacts and refresh linkGenerator urlMappingsHolder reference:
- `FormTagLibTests` - Clear `artefactInfo` in setup/cleanup, refresh linkGenerator
- `AbstractGrailsTagTests` - Clear URL mappings in `tearDown()` for all subclasses
- `RestfulReverseUrlRenderingTests` - Clear URL mappings and re-register in setup
- `ControllerUnitTestMixin.mockArtefact()` - Refresh linkGenerator.urlMappingsHolder after bean recreation

#### 8. System Property Isolation (~20 files)
Fixed tests to use `@RestoreSystemProperties` for proper system property isolation:
- `UrlMappingMatcherSpec`, `YamlPropertySourceLoaderSpec`
- `NavigableMapSpringProfilesSpec`, `GrailsPlaceHolderConfigurerCorePluginRuntimeSpec`
- `DevelopmentModeWatchSpec`, `GrailsArtefactTransformerSpec`
- `CreateAppCommandSpec`, `BookSpec`
- Neo4j and MongoDB auto-configuration specs
- And others

## Dependency Updates

- Groovy 3.0.25 in `grails-data-graphql`, `grails-data-neo4j`, `grails-forge` (Java 25 compatibility)
- Gradle wrapper 8.14.4 for generated apps (`gradleWrapperProperties.rocker.raw`)
- Gradle tooling API 8.14.4 (`gradle.properties`)
- `.sdkmanrc` updates for Java 17.0.18-librca

## Related PRs

- **#15346** - Adds `clearConstraintsMapCache()` public API to Validateable trait (targeting 7.1.x)

## Files Changed

- **~100 files** modified across flaky test fixes, CI optimizations, test retry removal, and production code improvements


jamesfredley:
First run: https://github.com/apache/grails-core/actions/runs/21306755309?pr=15335

jdaugherty:
In my early testing, using the cache was not always consistent because of how some of our dependencies are included. Part of the reason we force a rerun of tests was to ensure everything was functional.  I am ok disabling the cache for every commit, but I think we still need to run all of the tests periodically.

I would strongly urge we fix the reason the develocity plugin retries the tests first - sometimes it reruns all of the tests 3+ times.  Most of the tests that fail are in grails-gsp.  If you checkout the last build, look at flaky tests to find the problematic ones.  I suspect if we fixed these, the github build would go down by over half of its run time.

jdaugherty:
Oh and this scenario too: changing something in core might only be tested by the end to end test in forge, but forge doesn't depend on grails so gradle would just not run those tests because the cache would see nothing changed.  We had lots of issues earlier on in 7.0 dev that were only caught by the functional tests and gradle would not run them always because the cache would not trigger a change. This was a reoccurring theme that I ended up wasting a lot of time chasing in 7.0 dev.

jamesfredley:
Apparently `cleanTest` will only force the rerun of tests when Gradle cache is disabled.  I'm trying another approach.  

Still experiments to see if anything reduces this from an hour, before proposing changes.

jamesfredley:
Looks pretty good compared to recent runs, although there is significant variability 

image
image


vs

image
image
image

jamesfredley:
This ended up being more of a test cleanup than a speed improvement PR.

matrei:
Could we leverage https://spockframework.org/spock/docs/2.4/all_in_one.html#_restoresystemproperties instead of clearing system properties explicitly?

grails-mimetypes/src/test/groovy/grails/web/mime/MimeUtilitySpec.groovy
```diff
@@ -30,6 +31,11 @@ import spock.lang.Specification
  */
 class MimeUtilitySpec extends Specification {
 
+    void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 36: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-mimetypes/src/test/groovy/org/grails/plugins/web/mime/MimeTypesConfigurationSpec.groovy
```diff
@@ -21,11 +21,17 @@ package org.grails.plugins.web.mime
...
 import grails.web.mime.MimeType
+import org.grails.web.mime.HttpServletResponseExtension
 import org.springframework.context.ApplicationContext
 import spock.lang.Specification
 
 class MimeTypesConfigurationSpec extends Specification {
 
+    void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 32: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-mimetypes/src/test/groovy/org/grails/web/mime/AcceptHeaderParserSpec.groovy
```diff
@@ -41,6 +42,8 @@ class AcceptHeaderParserSpec extends Specification {
     def config
 
     void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 46: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and clean this up afterward?

grails-mimetypes/src/main/groovy/org/grails/web/mime/HttpServletRequestExtension.groovy
```diff
@@ -75,7 +76,17 @@ class HttpServletRequestExtension {
     static MimeType[] getMimeTypes(HttpServletRequest request) {
         MimeType[] result = (MimeType[]) request.getAttribute(GrailsApplicationAttributes.REQUEST_FORMATS)
         if (!result) {
-            WebApplicationContext context = WebApplicationContextUtils.getWebApplicationContext(request.servletContext)
+            // First try to get context from GrailsWebRequest (more reliable in tests with parallel execution)
```
matrei: Should we adapt production code for tests? Is it possible to clean up pollution in tests instead?

grails-mimetypes/src/test/groovy/org/grails/web/servlet/mvc/RequestAndResponseMimeTypesApiSpec.groovy
```diff
@@ -53,11 +55,14 @@ class RequestAndResponseMimeTypesApiSpec extends Specification{
     }
 
     void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 59: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-rest-transforms/src/test/groovy/org/grails/plugins/web/rest/render/hal/HalJsonRendererSpec.groovy
```diff
@@ -84,7 +86,13 @@ class HalJsonRendererSpec extends Specification {
         ShutdownOperations.runOperations()
     }
 
+    void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 91: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-test-suite-uber/src/test/groovy/org/grails/plugins/web/rest/render/BaseDomainClassRendererSpec.groovy
```diff
@@ -50,12 +50,18 @@ import org.springframework.mock.web.MockHttpServletRequest
...
+import org.grails.web.mime.HttpServletResponseExtension
 import org.springframework.web.context.request.RequestContextHolder
 import org.springframework.web.context.support.GenericWebApplicationContext
 import spock.lang.Specification
 
 abstract class BaseDomainClassRendererSpec extends Specification {
 
+    void setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 62: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-test-suite-uber/src/test/groovy/org/grails/web/codecs/HTMLJSCodecIntegrationSpec.groovy
```diff
@@ -57,6 +57,10 @@ public class HTMLJSCodecIntegrationSpec extends Specification {
         GrailsWebMockUtil.bindMockWebRequest()
         registry = GrailsWebRequest.lookup().getEncodingStateRegistry()
     }
+
+    def cleanup() {
+        org.springframework.web.context.request.RequestContextHolder.resetRequestAttributes()
```
matrei: Import `RequestContextHolder`?

grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/DefaultGrailsApplicationAttributesTests.groovy
```diff
@@ -50,6 +50,11 @@ class DefaultGrailsApplicationAttributesTests {
         request.setAttribute(GrailsApplicationAttributes.CONTROLLER, controller)
     }
 
+    @org.junit.jupiter.api.AfterEach
```
matrei: Import `AfterEach`?

grails-test-suite-uber/src/test/groovy/org/grails/web/util/StreamCharBufferSpec.groovy
```diff
@@ -66,6 +66,10 @@ class StreamCharBufferSpec extends Specification {
         codecOut=new GrailsPrintWriter(out.getWriterForEncoder(htmlCodecClass.encoder, EncodingStateRegistryLookupHolder.getEncodingStateRegistryLookup().lookup()))
     }
 
+    def cleanup() {
+        org.springframework.web.context.request.RequestContextHolder.resetRequestAttributes()
```
matrei: Import `RequestContextHolder`?

grails-test-suite-uber/src/test/groovy/org/grails/web/util/WebUtilsTests.groovy
```diff
@@ -49,6 +50,8 @@ class WebUtilsTests {
     @BeforeEach
     void setUp() {
         RequestContextHolder.resetRequestAttributes()
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 54: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
```diff
@@ -30,6 +31,21 @@ class CommandObjectNoDataSpec extends Specification implements GrailsWebUnitTest
...
+     * This is necessary because the Validateable trait caches constraints in a static field,
+     * and in parallel test execution, the constraints may be evaluated before doWithConfig()
+     * has registered the shared constraint 'isProg'.
+     *
+     * Also clear ConstraintEvalUtils.defaultConstraintsMap which caches shared constraints
+     * globally. In parallel test execution, another test's config may have been cached,
+     * causing the 'isProg' shared constraint to not be found.
+     */
+    def setup() {
+        ConstraintEvalUtils.clearDefaultConstraints()
```
matrei, line 64: These are guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-test-suite-web/src/test/groovy/org/grails/web/mime/ContentFormatControllerTests.groovy
```diff
@@ -33,6 +33,11 @@ import spock.lang.Specification
  */
 class ContentFormatControllerTests extends Specification implements ControllerUnitTest<ContentController>, DomainUnitTest<Gizmo> {
 
+    def setup() {
+        // Clear the static mimeTypes cache to ensure proper test isolation in parallel test runs
+        HttpServletResponseExtension.@mimeTypes = null
```
matrei, line 38: This is a guard for previous pollution (that in the best of worlds should not be necessary). Should we be good citizens and also clean this up afterward?

grails-validation/src/main/groovy/grails/validation/Validateable.groovy
```diff
@@ -103,6 +103,18 @@ trait Validateable {
...
 
+    /**
+     * Clears the cached constraints map, forcing re-evaluation on next access.
+     * This is primarily useful in testing scenarios where shared constraints
+     * may need to be re-evaluated after configuration changes.
+     *
+     * @since 7.0
+     */
+    @Generated
+    static void clearConstraintsMap() {
```
matrei: This is adding to the public API, preferable (if not breaking) added to 7.1.

grails-validation/src/test/groovy/grails/validation/ValidateableTraitAdHocSpec.groovy
```diff
@@ -27,6 +27,16 @@ import spock.lang.Specification
...
 class ValidateableTraitAdHocSpec extends Specification {
 
+    /**
+     * Clear the static constraints cache for classes that use shared constraints.
+     * This is necessary because the Validateable trait caches constraints in a static field,
+     * and in parallel test execution, the constraints may be evaluated before configuration
+     * has registered the shared constraints.
+     */
+    void setup() {
+        PersonAdHocSharedConstraintsValidateable.clearConstraintsMap()
```
matrei: Should we also clean up after?

grails-validation/src/test/groovy/grails/validation/ValidateableTraitSpec.groovy
```diff
@@ -36,6 +36,16 @@ import java.lang.reflect.Method
...
 class ValidateableTraitSpec extends Specification {
 
+    /**
+     * Clear the static constraints cache for classes that use shared constraints.
+     * This is necessary because the Validateable trait caches constraints in a static field,
+     * and in parallel test execution, the constraints may be evaluated before configuration
+     * has registered the shared constraints.
+     */
+    void setup() {
+        SharedConstraintsValidateable.clearConstraintsMap()
```
matrei: Should we also clean up after?

matrei (commented):
This is good stuff, thank you @jamesfredley

grails-mimetypes/src/main/groovy/org/grails/web/mime/HttpServletRequestExtension.groovy
jamesfredley: I had the same concern until stumbling on the following, there are also a few other reasons in the PR description above, but this was the one that made be lean towards fixing the production code.  


**Consistency with `HttpServletResponseExtension`**: The sister class `HttpServletResponseExtension.getMimeTypes()` already uses `GrailsWebRequest.lookup()` as the primary lookup method. This change makes `HttpServletRequestExtension` consistent.

jamesfredley:
@RestoreSystemProperties will replace all the system properties changes.

grails-test-examples/gsp-spring-boot/script/.sdkmanrc
```diff
@@ -1,3 +1,3 @@
-java=17.0.12-librca
+java=17.0.18-librca
```
jdaugherty, line 1: Should we even keep .sdkmanrc here?  It was left over from moving this project into the mono repo, yes?

grails-validation/src/main/groovy/grails/validation/Validateable.groovy
jdaugherty: We should probably be more explicit about this name - it's only clearing a cache.  I'd suggest the name `clearConstraintsMapCache()`

.github/workflows/gradle.yml
```diff
@@ -91,7 +100,6 @@ jobs:
         run: >
           ./gradlew build :grails-shell-cli:installDist groovydoc
           --continue
-          --rerun-tasks
```
jdaugherty, line 94: This rerun has cause issues in the functional tests b/c of our transitive dependencies.  I'm fine doing it for the regular build, but we need to introduce a build that doesn't have caching if we do this.  I'd suggest just running it weekly.

.github/workflows/gradle.yml
```diff
@@ -176,8 +184,8 @@ jobs:
         run: >
           ./gradlew bootJar check
           --continue
-          --rerun-tasks
           --stacktrace
+          --configuration-cache-problems=warn
```
jdaugherty: Same as core, I think if the tasks don't rerun then it's going to not rebuild with our AST transforms sometimes, which will cause the tests to execute with the old code.

.github/workflows/gradle.yml
```diff
@@ -210,10 +218,10 @@ jobs:
         env:
           GITHUB_MAVEN_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
         run: >
-          ./gradlew bootJar cleanTest check
+          ./gradlew bootJar check
           --continue
-          --rerun-tasks
```
jdaugherty, line 215: Same AST transform issue / rerun issues.

.github/workflows/gradle.yml
```diff
@@ -243,10 +251,10 @@ jobs:
...
```
jdaugherty, line 248: Same AST transform issue / rerun issues.

jdaugherty (dismissed):
My main concern is with the removal of rerun-tasks, we won't always reapply our AST transformers.  We experienced this in the 7.x development and a lot of time was lost trying to figure out when this broke.  I think this can easily be solved with a build without caching that runs weekly or something.

Thank you for doing all of the test cleanup- it's been sorely needed.

grails-mimetypes/src/test/groovy/grails/web/mime/MimeUtilitySpec.groovy
jamesfredley: Good point, will update all 8 locations.

.github/workflows/gradle.yml
jamesfredley: Based on the commit where `--rerun-tasks` was added, I was working under the assumption that it was added solely to force tests to rerun.  I found that `outputs.cacheIf { !isCiBuild }` works better to force tests to rerun.

Since `--rerun-tasks` is fixing more than just test reruns, is there a specific Gradle task or task that we want to ensure always run?  It is possible to force just a task or small set of tasks to run, which should be faster than `--rerun-tasks`.

jdaugherty: The problem is it depends where the change is; that's why I think it's better to take the caching approach and just create a separate workflow that we can check periodically.

If we clean up our dependencies, this should not be an issue longer term.  I think it's transitive while gradle doesn't fully understand our outputs.

grails-test-examples/gsp-spring-boot/script/.sdkmanrc
jamesfredley: The .sdkmanrc file in grails-test-examples/gsp-spring-boot/script/ appears to be for running the standalone Groovy script example with the Groovy CLI, not for building the project.   https://github.com/apache/grails-core/blob/0bd4622d11c2c611dd38c53df3ab659887410ab0/grails-test-examples/gsp-spring-boot/script/README.md

grails-validation/src/main/groovy/grails/validation/Validateable.groovy
jamesfredley: Updated.

grails-validation/src/test/groovy/grails/validation/ValidateableTraitSpec.groovy
jamesfredley: Added

grails-validation/src/test/groovy/grails/validation/ValidateableTraitAdHocSpec.groovy
jamesfredley: Added

grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
jamesfredley: Added

grails-test-suite-uber/src/test/groovy/org/grails/web/servlet/DefaultGrailsApplicationAttributesTests.groovy
jamesfredley: fixed

grails-test-suite-uber/src/test/groovy/org/grails/web/codecs/HTMLJSCodecIntegrationSpec.groovy
jamesfredley: fixed

grails-test-suite-uber/src/test/groovy/org/grails/web/util/StreamCharBufferSpec.groovy
jamesfredley: fixed

grails-validation/src/main/groovy/grails/validation/Validateable.groovy
jamesfredley: Agreed, will create something which fixes tests another way for now and potentially create another PR with this method against 7.1.x.

jamesfredley: Updated and moved public API change to https://github.com/apache/grails-core/pull/15346

.github/workflows/gradle.yml
jamesfredley: Another potential solution, matching how the PR forces tests to rerun.  https://github.com/apache/grails-core/pull/15335/commits/a9bb5730658f3e489f41a227dd4c9a2ed87f0248

grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
```diff
@@ -20,16 +20,64 @@
...
+import spock.lang.Isolated
 import spock.lang.Specification
 
+/**
+ * Tests for command object validation without DataTest trait.
+ * This spec is marked @Isolated because it modifies global shared constraints
+ * via doWithConfig() which affects ConstraintEvalUtils.defaultConstraintsMap - a static
+ * cache shared across all tests in the same JVM fork.
+ */
+@Isolated
```
matrei: My understanding of `@Isolated` is that it is only useful when running Spock with parallel execution (https://spockframework.org/spock/docs/2.4/all_in_one.html#parallel-execution). We are not running parallel execution, we are running multiple forks (separate JVM processes, each running one Spec at a time), so Spock will run tests sequentially, in each fork.

grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectsSpec.groovy
```diff
@@ -23,11 +23,23 @@ import grails.artefact.Artefact
...
 import spock.lang.Issue
 import spock.lang.Specification
 
+/**
+ * Tests for command object binding and validation.
+ * This spec is marked @Isolated because it modifies global shared constraints
+ * via doWithConfig() which affects ConstraintEvalUtils.defaultConstraintsMap - a static
+ * cache shared across all tests in the same JVM fork.
+ */
+@Isolated
```
matrei: My understanding of `@Isolated` is that it is only useful when running Spock with parallel execution (https://spockframework.org/spock/docs/2.4/all_in_one.html#parallel-execution). We are not running parallel execution, we are running multiple forks (separate JVM processes, each running one Spec at a time), so Spock will run tests sequentially, in each fork.

grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
jamesfredley: I see, will refactor another way.

matrei:
Many comments mention `parallel` test execution. I think that might be confusing. We are, to my knowledge, not running tests in parallel in the same process. The parallelism comes from Gradle running multiple forks which are already isolated from each other.
https://spockframework.org/spock/docs/2.4/all_in_one.html#parallel-execution

jamesfredley:
> Many comments mention `parallel` test execution. I think that might be confusing. We are, to my knowledge, not running tests in parallel in the same process. The parallelism comes from Gradle running multiple forks which are already isolated from each other. https://spockframework.org/spock/docs/2.4/all_in_one.html#parallel-execution

I'll adjust the comments, since fork and parallel have different meanings for these flaky test failures.

gradle/test-config.gradle
```diff
@@ -47,10 +61,9 @@ tasks.withType(Test).configureEach {
     useJUnitPlatform()
     jvmArgs += java17moduleReflectionCompatibilityArguments
     develocity {
-        testRetry {
-            maxRetries = configuredTestParallel == 1 ? 1 : 2
-            maxFailures = 20
-            failOnPassedAfterRetry = true
+        testDistribution {
+            enabled = isCiBuild
+            maxLocalExecutors = configuredTestParallel
```
matrei: Do we have remote agents available/configured?

grails-mimetypes/src/main/groovy/org/grails/web/mime/HttpServletRequestExtension.groovy
matrei: OK, but can we remove the mentioning of tests in the comments?

matrei:
> > Many comments mention `parallel` test execution. I think that might be confusing. We are, to my knowledge, not running tests in parallel in the same process. The parallelism comes from Gradle running multiple forks which are already isolated from each other. https://spockframework.org/spock/docs/2.4/all_in_one.html#parallel-execution
> 
> I'll adjust the comments, since fork and parallel have different meanings for these flaky test failures.

I think it might be clearer to talk about test environment pollution.

.sdkmanrc
```diff
@@ -1,3 +1,3 @@
-java=17.0.17-librca
+java=17.0.18-librca
```
matrei, line 2: Remember to also keep this in sync with `.github/workflows/release.yml` and `etc/bin/Dockerfile`. Maybe add a comment here as a reminder?

jamesfredley: https://github.com/apache/grails-core/issues/15352

jamesfredley: Updated all three for now.  New Issue to find better solutions.

jdaugherty: Dockerfiles can take args, creating a shell script to spin up the verification container is completely reasonable.  Also, the gradle properties github action can load property files like sdkman, so both the release workflow & dockerfile versions should be easily loaded from .sdkmanrc.

jamesfredley:
all the version updates were moved out to a separate PR and merged, since they were missed when we updated Gradle, Groovy and Java versions.

.github/workflows/gradle.yml
jamesfredley: --rerun-tasks has been restored in all jobs except build grails core and a duplicate job that reruns tasks on just Ubuntu with Java 17 has been added.  This should give us the coverage while saving cumulative time.

grails-mimetypes/src/main/groovy/org/grails/web/mime/HttpServletRequestExtension.groovy
jamesfredley: Thank you, the comment has been cleaned up so it does not reference tests.

gradle/test-config.gradle
jamesfredley: I am going to carve this off into a research issue since they are not configured, but are supported.  I'll add it to the issue for migrating to Apache's instance.

jamesfredley: It was previously in place for 'testBisect'

matrei (approved):

jdaugherty (approved):

jdaugherty:
This fixed https://github.com/apache/grails-core/issues/14319

Issue #15326, opened 2026-01-20
Document Plugin Best Practices


jdaugherty:
### Issue description

The Grails documentation does a great job on how to create a plugin, but after the upgrade to Grails 7 we have learned several best practices.  These practices should be added to the documentation. 

Should be based on https://github.com/grails-plugins/grails-server-timing

Some topics that need included: 
1. Build: Using a multiproject build to separate example/test code from plugin code
2. Continuous Builds: Adopting CI via GitHub actions (https://github.com/apache/grails-github-actions)
3. Publishing: Publishing configuration (https://github.com/apache/grails-gradle-publish)
4. Init: Publishing a github template for getting started
5. Documentation: How to publish documentation 
6. Development: Easy environment configuration by adopting sdkman
7. Licensing: Ensuring licensing is clear and recommending the ASF license by default
8. Build: Gradle multiproject design using composition instead of if/else in the root of the project
9. Discoverability: Adding the plugin to our index of plugins (https://github.com/apache/grails-plugins-metadata) 
10. Advanced Design: when to include the Grails Plugin Gradle Plugin and went to not. 
11. Testing: testing approaches and how to ensure the project is maintained longer term.

Ajay9704:
@jdaugherty  I've submitted a pull request that addresses this issue, please review and let me know if anything is required from my end

jamesfredley:
https://github.com/grails-plugins/grails-server-timing is the Grails Plugin Gold Standard

PR #15336, opened 2026-01-24
added plugin documents

Ajay9704:
added plugin best practices documentation

i worked on adding comprehensive documentation for plugin best practices as mentioned in issue #15326. honestly it took me quite a while to understand all the different aspects of plugin development in grails 7 but i think i got most of it covered.

what i did:

- wrote up how to set up multiproject builds with gradle to keep plugin code separate from example apps and tests
- included github actions setup for ci/cd (copied from the apache repos to make sure its correct)
- documented the publishing process using grails-gradle-publish plugin 
- added info about creating github templates for new plugins
- explained how to publish documentation with github pages
- put in some notes about using sdkman for environment setup (makes life easier)
- wrote about licensing recommendations (apache license seems to be the way to go)
- included guidance on gradle composition vs conditional logic (learned this was important)
- added how to get plugins listed in the grails plugin index
- documented when to use the grails plugin gradle plugin vs simpler approaches
- wrote about testing strategies for keeping plugins maintainable long term

the documentation is in the usual grails doc format with asciidoctor syntax. i tried to make it practical with real examples instead of just theory. hope i got everything right - please let me know if anything needs fixing.

files i touched:
- added new file: grails-doc/src/en/guide/plugins/pluginBestPractices.adoc
- updated toc.yml to include the new section
- updated plugins.adoc to link to the new content

### DIFF
--- a/grails-doc/src/en/guide/plugins.adoc
+++ b/grails-doc/src/en/guide/plugins.adoc
@@ -20,3 +20,5 @@ under the License.
 Grails is first and foremost a web application framework, but it is also a platform. By exposing a number of extension points that let you extend anything from the command line interface to the runtime configuration engine, Grails can be customised to suit almost any needs. To hook into this platform, all you need to do is create a plugin.
 
 Extending the platform may sound complicated, but plugins can range from trivially simple to incredibly powerful. If you know how to build a Grails application, you'll know how to create a plugin for <<providingBasicArtefacts,sharing a data model>> or some static resources.
+
+For professional plugin development, see <<pluginBestPractices,Plugin Best Practices>> for comprehensive guidelines on creating maintainable, production-ready plugins following industry standards.
--- a/grails-doc/src/en/guide/plugins/pluginBestPractices.adoc
+++ b/grails-doc/src/en/guide/plugins/pluginBestPractices.adoc
@@ -1,0 +1,919 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+////
+
+==== Plugin Best Practices
+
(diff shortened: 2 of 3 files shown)


Ajay9704:
@jdaugherty  can you trigger the workflows so i can see the build status

Ajay9704:
@jdaugherty  all checks are successful , please review my pr and let me know if changes are required from my end

borinquenkid (changes requested):
*Review drafted with assistance from Claude Code (Anthropic's CLI agent) — I reviewed and confirmed the findings below before posting.*

Thanks for taking this on — plugin best-practices docs are genuinely needed. I checked the examples against the actual `grails-gradle` codebase and found several that reference things that don't exist, which would break a plugin author's build if followed as written. Requesting changes before this can merge.

## Critical: fabricated Gradle plugin IDs

The only plugin IDs actually registered in `grails-gradle/plugins/build.gradle` are: `grails-app`, `grails-gsp`, `grails-gson`, `grails-markup`, `grails-plugin`, `grails-profile`, `grails-web`, `grails-publish-profile`, `grails-exploded`, `grails-integration-test`.

This doc introduces several that don't exist anywhere in the codebase:
- `id "org.apache.grails.gradle.grails-publish"` — no such plugin. The real publish-related one is `grails-publish-profile` (`GrailsProfilePublishGradlePlugin`), and it serves a different purpose (profile publishing).
- `id "org.apache.grails.gradle.grails-github-pages"` with a `githubPages { repoUri ... targetBranch ... }` block — I grepped the whole repo and this plugin/DSL doesn't exist at all.
- The `grailsPublish { githubSlug, license, developers, portalUrl, ... }` DSL doesn't match any extension registered by any plugin here — it looks like it's carried over from the old Grails 3 / Bintray-era community publish plugin.

Anyone copying these examples will hit `Plugin with id '...' not found`.

## Outdated reference

The "Plugin Discoverability" section has readers add a `"bintrayPackage"` object to `grails-plugins.json`. Bintray shut down in 2021 — this should be dropped or replaced with whatever the current registration process actually is.

## Duplicates/contradicts existing docs

`creatingAndInstallingPlugins.adoc` already has a correct "Plugins and Multi-Project Builds" section that uses the real `grails-exploded` plugin for reload support. This new doc covers the same ground but never mentions `grails-exploded`, and instead introduces the fictional plugins above — so we'd end up with two docs giving conflicting guidance for the same task.

## Scope

A good chunk of the page (generic CI matrix builds, Jacoco config, a generic Spock `calculate()` test, a generic `setup-dev.sh` script) isn't Grails-specific and could be trimmed to keep the page focused and easier to keep accurate long-term.

## Ask

Could you re-verify the publishing/GitHub Pages sections against what the project actually uses today (worth checking `CONTRIBUTING.md` and current community plugin repos for the real pattern), drop the Bintray reference, and reconcile the multiproject section with the existing `creatingAndInstallingPlugins.adoc` content? Happy to take another look once that's updated.

PR #15346, opened 2026-01-25
Add clearConstraintsMapCache() public API to Validateable trait

jamesfredley:
## Summary

Add a new static method `clearConstraintsMapCache()` to the `Validateable` trait that allows clearing the cached constraints map, forcing re-evaluation on next access.

## Motivation

The `Validateable` trait caches evaluated constraints in a private static field `constraintsMapInternal`. During parallel test execution, constraints may be evaluated before `doWithConfig()` has registered shared constraints, causing test failures where shared constraint names (like `'isProg'`) are not found.

This new public API method enables proper test isolation by allowing tests to clear the constraint cache in setup/cleanup methods.

## API Addition

```groovy
/**
 * Clears the cached constraints map, forcing re-evaluation on next access.
 * This is useful in testing scenarios where shared constraints may need
 * to be re-evaluated after configuration changes, particularly during
 * parallel test execution.
 *
 * @since 7.1
 */
@Generated
static void clearConstraintsMapCache()
```

## Usage Example

```groovy
class CommandObjectsSpec extends Specification implements ControllerUnitTest<TestController> {
    
    Closure doWithConfig() {{ config ->
        config['grails.gorm.default.constraints'] = {
            isProg inList: ['Emerson', 'Lake', 'Palmer']
        }
    }}

    def setup() {
        ConstraintEvalUtils.clearDefaultConstraints()
        Artist.clearConstraintsMapCache()
    }

    def cleanup() {
        ConstraintEvalUtils.clearDefaultConstraints()
        Artist.clearConstraintsMapCache()
    }
}
```

## Why This is a Feature (7.1)

Per the Grails versioning policy, adding a new public method to a trait is a backward-compatible feature addition that belongs in a MINOR release (7.1.x), not a PATCH release (7.0.x).

A companion change in PR #15335 (targeting 7.0.x) uses reflection to achieve the same test isolation without modifying the public API.

## Files Changed

- `grails-validation/src/main/groovy/grails/validation/Validateable.groovy` - Add new `clearConstraintsMapCache()` method
- `grails-validation/src/test/groovy/grails/validation/ValidateableTraitSpec.groovy` - Add test isolation
- `grails-validation/src/test/groovy/grails/validation/ValidateableTraitAdHocSpec.groovy` - Add test isolation
- `grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectsSpec.groovy` - Add test isolation
- `grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy` - Add test isolation

### DIFF
--- a/grails-geb/src/testFixtures/groovy/grails/plugin/geb/serviceloader/ServiceRegistry.groovy
+++ b/grails-geb/src/testFixtures/groovy/grails/plugin/geb/serviceloader/ServiceRegistry.groovy
@@ -24,7 +24,7 @@ import groovy.transform.CompileStatic
  * A service registry that loads and caches different service types using the Java
  * {@link java.util.ServiceLoader}, while allowing overriding which instance to return.
  * <p>
- * This class provides thread-safe service loading and caching, supporting parallel test execution.
+ * This class provides thread-safe service loading and caching to prevent test environment pollution.
  * It provides both automatic service discovery through {@link java.util.ServiceLoader}
  * and explicit replacement of the returned service instance for customization.
  * </p>
--- a/grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
+++ b/grails-test-suite-web/src/test/groovy/org/grails/web/commandobjects/CommandObjectNoDataSpec.groovy
@@ -20,6 +20,7 @@
 package org.grails.web.commandobjects
 
 import grails.testing.web.GrailsWebUnitTest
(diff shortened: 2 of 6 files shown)


jdaugherty (approved):

matrei (approved):
Could we change the wording in the comments to move away from "parallel test  execution" as that is another concept that is not in play here?

jamesfredley:
All comments updated to a variant of "prevents test environment pollution"

PR #15358, opened 2026-01-28
7.1.x Fix for UrlMapping greedy not working when no id is present

codeconsole:
Follow up to https://github.com/apache/grails-core/pull/15213

```groovy
"/$controller/$action?/$id?(.$format)?"
```

resulted in `/user.json` not matching. only `/user` would match. This fixes that.

Is seems like the best pathway forward to support formats and string $id's with .'s is the following mapping:

```groovy
  // First: greedy with format constraint                                                                                              
  "/$controller/$action/$id+(.$format)?" {                                                                                             
      constraints {                                                                                                                    
          format(inList: ['json', 'xml', 'html', 'csv'])                                                                               
      }                                                                                                                                
  }                                                                                                                                    
  // Second: fallback for IDs with dots but no valid format                                                                            
  "/$controller/$action/$id"()  
```

  | URL | id | format | Which mapping? |                                                                                               
  |-----|-----|--------|----------------|                                                                                              
  | `/user/show/bob.smith` | `bob.smith` | `null` | Fallback (no valid format) |                                                       
  | `/user/show/bob.smith.json` | `bob.smith` | `json` | Greedy (valid format) |                                                       
  | `/user/show/bob.smith.jones` | `bob.smith.jones` | `null` | Fallback (no valid format) |                                           
  | `/user/show/bob.smith.jones.xml` | `bob.smith.jones` | `xml` | Greedy (valid format) |

### DIFF
--- a/grails-web-url-mappings/src/test/groovy/grails/web/mapping/UrlMappingsWithGreedyExtensionSpec.groovy
+++ b/grails-web-url-mappings/src/test/groovy/grails/web/mapping/UrlMappingsWithGreedyExtensionSpec.groovy
@@ -227,4 +227,223 @@ class UrlMappingsWithGreedyExtensionSpec extends AbstractUrlMappingsSpec {
             info.urlData.hasGreedyExtensionParam()
             info.urlData.hasOptionalExtension()
     }
+
+    void "Test that greedy id with optional action matches controller-only URL with format"() {
+        given: "A complex URL mapping with optional action and optional greedy id"
+            def urlMappingsHolder = getUrlMappingsHolder {
+                "/$controller/$action?/$id+?(.$format)?"()
+            }
+
+        when: "Matching a URL with just controller and format (no action, no id)"
+            def info = urlMappingsHolder.match('/mobile.json')
+
+        then: "The controller and format are correctly extracted without greedy behavior affecting controller"
+            info != null
+            info.parameters.controller == 'mobile'
+            info.parameters.action == null
+            info.parameters.id == null
+            info.parameters.format == 'json'
+    }
+
+    void "Test that greedy id works correctly when all parameters are present"() {
+        given: "A complex URL mapping with optional action and optional greedy id"
+            def urlMappingsHolder = getUrlMappingsHolder {
+                "/$controller/$action?/$id+?(.$format)?"()
+            }
+
+        when: "Matching a URL with all parameters including id with dots"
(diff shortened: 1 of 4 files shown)


grails-web-url-mappings/src/test/groovy/grails/web/mapping/UrlMappingsWithGreedyExtensionSpec.groovy
```diff
@@ -227,4 +227,223 @@ class UrlMappingsWithGreedyExtensionSpec extends AbstractUrlMappingsSpec {
...
+            def urlMappingsHolder = getUrlMappingsHolder {
+                "/$controller/$action?/$id+?(.$format)?"()
+            }
+
+        when: "Matching a URL with controller containing dots and format"
+            def info = urlMappingsHolder.match('/test.test.json')
+
+        then: "The URL /test.test.json should NOT apply greedy to controller"
+            info != null
+            println "DEBUG: controller=${info.parameters.controller}, action=${info.parameters.action}, id=${info.parameters.id}, format=${info.parameters.format}"
```
sbglasius: Is leaving a `println` intended? Personally I prefer to annotate the Spec with `@Slf4j` and use `log.debug` instead

grails-web-url-mappings/src/test/groovy/grails/web/mapping/UrlMappingsWithGreedyExtensionSpec.groovy
```diff
@@ -227,4 +227,223 @@ class UrlMappingsWithGreedyExtensionSpec extends AbstractUrlMappingsSpec {
...
+            def urlMappingsHolder = getUrlMappingsHolder {
+                "/$controller/$action?/$id+?(.$format)?"()
+            }
+
+        when: "Matching a URL where the id contains a dot but there's no real format"
+            def info = urlMappingsHolder.match('/user/show/bob.smith')
+
+        then: "Check what actually happens - does .smith become the format?"
+            info != null
+            println "DEBUG bob.smith: controller=${info.parameters.controller}, action=${info.parameters.action}, id=${info.parameters.id}, format=${info.parameters.format}"
```
sbglasius: same as above

matrei: Remove debug println?

matrei (approved):

PR #15361, opened 2026-01-29
7.1.x - Simplify star imports check

codeconsole:
Rather than searching the dependency hierarchy, just check to see if the class is in the classpath.

enabled by: 
```gradle
grails {
    importGrailsCommonAnnotations = true
}
```

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -18,6 +18,8 @@
  */
 package org.grails.gradle.plugin.core
 
+import java.util.zip.ZipFile
+
 import grails.util.BuildSettings
 import grails.util.Environment
 import grails.util.GrailsNameUtils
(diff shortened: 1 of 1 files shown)


sbglasius (approved):
When tests pass, then this PR is approved.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -263,29 +263,35 @@ ${importStatements}
...
+                    }
+                }
+            } catch (Exception ignored) {
             }
-        } catch (Exception e) {
-            project.logger.debug("Failed to check for dependency ${group}:${name} in ${configurationName}: ${e.message}")
-            return false
         }
+        return false
     }
```
matrei:
```suggestion
    private static boolean isClassOnClasspath(FileCollection classpath, String className) {
        def classEntry = className.replace('.', '/') + '.class'
        classpath.files.any { f ->
            try {
                if (f.file && f.name.endsWith('.jar')) {
                    new ZipFile(f).withCloseable { zip ->
                        zip.getEntry(classEntry) != null
                    }
                } else if (f.directory) {
                    new File(f, classEntry).exists()
                } else {
                    false
                }
            } catch (Exception ignored) {
                false
            }
        }
    }

```

matrei (approved):

PR #15394, opened 2026-02-17
test: stabilize flaky scaffolding Geb tests for slow CI environments

jamesfredley:
## Summary

Stabilizes flaky `UserControllerSpec` and `UserCommunityControllerSpec` Geb tests in `grails-test-examples/scaffolding` that intermittently fail on slower CI runners (observed on JDK 25 in [Functional Tests (25)](https://github.com/apache/grails-core/actions/runs/22083891105/job/63814394408?pr=15393)).

## Problem

The test expects page title `User List` but intermittently gets `Please sign in` because:

1. **No wait after login** — `setup()` clicks the login button then immediately proceeds to the feature method. On slower environments, the login POST hasn't completed, so the session cookie isn't set when `go 'user/index'` executes, causing Spring Security to redirect to the login page.
2. **No waitFor on assertions** — `title == 'User List'` asserts immediately without waiting for async page load, unlike every other Geb spec in this repo.

These were the **only two Geb specs** not updated with `waitFor` patterns during the stability improvements in e8adf0143f.

## Changes

- Add `waitFor { title != 'Please sign in' }` after login click to verify auth completes before proceeding
- Wrap `title == 'User List'` assertions with `waitFor { }` for async page load handling
- Add `waitFor` for logout button visibility before clicking in `cleanup()`
- Remove redundant `go '/'` before `to LoginPage` (the `to` command navigates directly)

## Evidence

| Signal | Detail |
|--------|--------|
| Same test passed on `7.0.x` | [Run 22018179657](https://github.com/apache/grails-core/actions/runs/22018179657) ✅ |
| Same test passed on JDK 17 & 21 | Same CI run, only JDK 25 failed |
| Failure pattern | Classic auth timing flake — `Please sign in` instead of `User List` |
| Established pattern | All other Geb specs use `waitFor` extensively (see `ScaffoldingFunctionalSpec`, `FieldsValidationSpec`, `PaginationSpec`) |

### DIFF
--- a/grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/UserCommunityControllerSpec.groovy
+++ b/grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/UserCommunityControllerSpec.groovy
@@ -38,7 +38,6 @@ class UserCommunityControllerSpec extends ContainerGebSpec {
     void cleanup() {
         try {
             go 'logout'
-            waitFor { $('input', value: 'Log Out').displayed }
             $('input', value: 'Log Out').click()
         }
         catch (ignored) {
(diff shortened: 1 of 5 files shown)


jdaugherty:
So I thought I had added waitFor and there was push back to instead set the global property that was added to change the default wait.  I might be thinking of the spring security project instead.  @matrei are you ok with this approach?

jdaugherty:
Looks @matrei cleaned up this.  I'm good to merge this if he is.

matrei (approved):

PR #15409, opened 2026-02-18
Load plugin.yml/plugin.groovy earlier so @Conditional* annotations can be used

### LINKED ISSUE #15408: Plugin configuration is loaded after `@Conditional` checks in Spring
### Issue description

While working on an example Grails plugin, I discovered that plugin.yml is loaded *after* Spring checks the various `@ConditionalOnProperty` annotations.  We could move plugins completely to the autoconfiguration workflow, but the first step is to get configuration loading earlier in the process to ensure it's accessible via spring supported annotating wiring.

jdaugherty:
Fixes https://github.com/apache/grails-core/issues/15408 

Assisted-by: OpenCode <[opencode@opencode.ai](mailto:opencode@opencode.ai)>
Assisted-by: Claude <[claude@anthropic.com](mailto:claude@anthropic.com)>


grails-core/src/main/groovy/grails/plugins/GrailsPlugin.java
```diff
@@ -263,6 +263,19 @@ public interface GrailsPlugin extends ApplicationContextAware, Comparable, Grail
...
+     *
+     * <p>Plugin configuration is loaded early in the Spring Boot lifecycle by
+     * {@link org.grails.config.GrailsPluginEnvironmentPostProcessor} and registered as a
+     * property source named {@code <pluginName>-plugin.yml} or {@code <pluginName>-plugin.groovy}.
+     * This method retrieves that property source from the application context's environment.</p>
+     *
+     * @return the plugin's property source from the environment, or {@code null} if the
+     *         GrailsApplication main context is not yet available or no matching property source exists
+     */
+    @Deprecated(forRemoval = true)
```
sbglasius: I think that if it's deprecated it should say why.

sbglasius (approved):

grails-core/src/main/groovy/grails/plugins/GrailsPlugin.java
jdaugherty: I was proposing this be removed because I don't see a usage for it - you can fetch the configuration but when would it be valid to look at the shipped config? btw, i had marked this as draft to see what @matrei thinks of this. This is a rather large change for 7.0.x so we may want to do some further testing.

jamesfredley:
If this does not require changes in end Grails apps I am OK with this going into 7.0.x, if we think the risk is too high, then 7.1.x is also OK.

grails-core/src/main/groovy/org/grails/config/GrailsPluginEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,641 @@
...
+     * <p>The plugin class is instantiated to read its {@code loadAfter}, {@code loadBefore},
+     * and {@code dependsOn} properties via a {@link BeanWrapper}, mirroring the approach
+     * used by {@link org.grails.plugins.DefaultGrailsPlugin}.</p>
+     *
+     * <p>Configuration is discovered by probing for both {@code plugin.yml} and
+     * {@code plugin.groovy} relative to the plugin class, replicating the approach from
+     * {@link org.grails.plugins.AbstractGrailsPlugin#readPluginConfiguration}. If both
+     * exist, a {@link RuntimeException} is thrown.</p>
+     */
+    PluginInfo extractPluginInfo(Class<?> pluginClass) {
```
jamesfredley: calls `pluginClass.getDeclaredConstructor().newInstance()` to read `loadAfter`/`loadBefore`/`dependsOn`. At this point in the Spring Boot lifecycle, there's no `GrailsApplication`, no application context, no plugin manager. If any plugin's no-arg constructor has side effects or requires dependencies, this could fail.

The code handles this gracefully (catches exceptions, continues with default ordering), but plugins with complex constructors will silently get default ordering rather than their declared ordering. This could subtly change the order configs are loaded compared to the existing behavior.

grails-core/src/main/groovy/org/grails/config/GrailsPluginEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,641 @@
...
+
+    /**
+     * Finds a plugin configuration resource at the given path relative to the plugin class,
+     * replicating the logic in {@link grails.io.IOUtils#findResourceRelativeToClass}.
+     *
+     * @param pluginClass the plugin class to resolve relative to
+     * @param configPath the path to probe (e.g., {@code "/plugin.yml"} or {@code "/plugin.groovy"})
+     * @return the resource wrapping the configuration URL, or {@code null} if not found
+     */
+    Resource getConfigurationResource(Class<?> pluginClass, String configPath) {
```
jamesfredley: reimplements the logic from `IOUtils.findResourceRelativeToClass` instead of calling it. The logic is nearly identical, but there's one difference: the original uses `BuildSettings.BUILD_CLASSES_PATH` constant for the dev-mode path check, while the new code hardcodes `"build/classes/groovy/main"`. If `BuildSettings.BUILD_CLASSES_PATH` ever changes, these would diverge. However, using `IOUtils` directly would pull in the `grails-gradle-model` dependency which may not be desirable at this early lifecycle point.

grails-core/src/main/groovy/org/grails/config/GrailsPluginEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,641 @@
...
+     * and extracts plugin class names using SAX parsing (same approach as
+     * {@link org.grails.plugins.CorePluginFinder}).
+     */
+    List<String> scanPluginDescriptors() {
+        List<String> pluginClassNames = new ArrayList<>();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+
+        try {
+            Enumeration<URL> resources = classLoader.getResources(CORE_PLUGIN_PATTERN);
+            SAXParser saxParser = SpringIOUtils.newSAXParser();
```
jamesfredley: A single `SAXParser` instance is created and reused across all `grails-plugin.xml` files. Per the SAX spec, `SAXParser` instances are not guaranteed to be reusable after `parse()`. The `javax.xml.parsers.SAXParser` javadoc says "An implementation of SAXParser is NOT guaranteed to behave as per the specification if it is used concurrently by two or more threads." More importantly, some implementations don't support reuse at all. A new parser should be created per file, or `reset()` should be called between uses.

jamesfredley (approved):
Looks solid.  A few minor items to take a look at.  

The `GrailsPluginConfigurationClass` test removed the "plugin with both yml and groovy throws exception" test. That validation still exists in `GrailsPluginEnvironmentPostProcessor.readPluginConfiguration()` (line 293), but there's no unit test covering it in the new code. The `GrailsPluginEnvironmentPostProcessorSpec` should add a test for this.

jdaugherty:
This change is incomplete since the plugin configuration will be loaded potentially when the plugin is disabled.  We may be able to post process already loaded configuration and remove them, but I'm not sure without further research.  

Here's the loading order in Spring: 

• SpringApplication.run(...)
• Prepare SpringApplicationRunListeners
• Prepare Environment
Create/attach ConfigurableEnvironment
Invoke all EnvironmentPostProcessors
• Create ApplicationContext
• Prepare context
• Refresh context (bean creation begins)

The problem is the plugin manager has different discovery mechanisms (but very similar).  Worst, one of those mechanisms appears to require the application context, which won't exist when we need to load configuration.  I *think* the need for the application context (or rather the parent application context) isn't actually used though.  

I'm doing further research on this to see how best to unify the plugin manager mechanisms so we can share the logic in both places.

grails-core/src/main/groovy/org/grails/config/GrailsPluginEnvironmentPostProcessor.java
jdaugherty: See loadCorePluginsFromResources in the CorePluginFinder; this is how that code has always worked.  This isn't threaded and this is how it's worked historically, so I'm not inclined to change it.

grails-core/src/main/groovy/org/grails/config/GrailsPluginEnvironmentPostProcessor.java
jdaugherty: I'd argue at this point they shoudln't be using Holders either.  These dependencies would have to be autowired and that would happen after the constructor creation, so  i view this as low risk.

jdaugherty:
I've been working on this and will likely push a significant update in the next day or so.  The core issue is the plugin filtering needs shared between the environment & the application post processors.  The problem: the plugin manager is what currently does this filtering & it's a bean.  The environment post processor can't use a bean b/c the context doesn't exist yet.  The only practical place the filtering is used is when in unit tests it filters the plugins.  I think the right solution is either moving the filter further out & sharing it between the plugin manager / EPP, or just moving the plugin manager away from being a bean completely.

jdaugherty:
This change was way larger than I wanted it to be. I ended up having to do this change manually because the complicated lifecycle made it so AI couldn't easily work in this area.  As a summary: to determine what plugin configuration to load, you have to know the order & what plugins were loaded.  This requires a lot of information, including: 
* plugin version / name
* grails version (not supported => not loaded)
* loadbefore / loadafter / dependencies fields 
* environment support (wrong environment => not loaded)
* evictions (if evicted, never loaded)

Worst, there isn't a stable interface for a Grails Plugin.  Several properties are 'dynamic' in that they can support a single string, a list, or even a map.  All of this is also determined at runtime via some rather complicated logic.   I introduced a GrailsPluginDiscovery class that determines the order plugins should be loaded in (which is also different than the sorted order that later processing happens against). The majority of these changes are "lift" and shifting code to an earlier process & then referencing that bean to get the various orders so when the plugins are actually loaded they load the same way the configuration did.  I very well could have made an error in the lift and shift, but I did it incrementally because it kept being so fragile.  Most of the private methods on the various GrailsPluginManagers were moved to helpers on a GrailsPluginUtils class.  

My other issue was I needed a way to "share" data between the bootstrap process & the actual bean loading processing (I did not want to load the data twice and assume it would load the same).  Spring has a bootstrapContext that supports promoting shared classes to beans, and that's the approach I've taken in this PR.  However, there is one spec test - EmbeddedContainerWithGrailsSpec - that explicitly tests an embedded context.  Since this is a test, I'm assuming that bootstrap process works in an embedded container, but if we take this change we need to explicitly test the embedded tomcat server because I had to implement a crude bootstrap process to get the test to pass (are the functional tests enough for this? they start fine...).  My guess is if anyone is programmatically starting a GrailsApplication, then they'll have to simulate this same process (unless they're invoking it via the spring mechanism).

I also split the upgrade documentation into a 7.1 & 7.0 section. I believe there's more work to do on documenting this PR, but I wanted the actual change reviewed before trying to finalize the documentation. The most noteable impacts to this change are: if any tests are "mocking" the plugin environment, and not using a standard mechanism, then the bootstrap process that populates the plugin order will now need mocked too. Note: if you're using the testing-support libraries, this is done for you.

Finally, as part of testing this, we need to make sure logging works when a bootstrap error occurs. In some of my tests, I didn't see logging on an exception from the GrailsPluginDiscovery class.  I think it was because of the test logging configuration, but if we decide to move forward with this, it's one of those critical items we need to check.

This change is way too large for 7.0.  Maybe even for 7.1 due to the scope of changes.  I think if we leave this in 7.1, we should do a milestone release of 7.1 given the impact this could have.  Otherwise, we need to wait until Grails 8.  There are some class renames in this PR, that I can probably undo to limit the scope, but I think the renamed files are considered internal to Grails and it's best to probably leave them for clarity.  I'd rather see this support added sooner than later given that a key feature of Spring can't be used without it.  It's just unfortunate how much work had to be done to get this working. 

Oh, and I also discovered there are several classes in core that don't appear to be used anymore.  I'm assuming some of them were there for troubleshooting, but we should eventually cleanup the dead code in this area. There are other optimizations that can be done too - for example the plugin load order should probably be calculated before trying to load.  Right now, it just iterates over and over against the delayed plugin list until all plugins are found in the expected order - effectively bruteforcing the load order.

@jamesfredley & @matrei  take a look at this PR now and see what you think.  All of the tests are passing locally for me now.  I've also published this version locally and started my large Grails application without issue (which means it works with many, external & existing Grails plugins).  I also confirmed the startup order was roughly the same in my application before/after this change (there is some randomness to this due to the classpath discovery order and not all plugins having explicit loadbefore/loadafter). My preference would be a 7.1.0-M1 and put this there.

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
+        try {
+            GrailsPluginDiscovery pluginDiscovery = bootstrapContext.get(GrailsPluginDiscovery.class);
+            Collection<GrailsPluginInfo> plugins = pluginDiscovery.getLoadOrderedPlugins(environment);
+            loadPluginConfigurations(plugins, environment);
+        } catch (Exception e) {
+            LOG.warn("Error loading Grails plugin configurations early: {}. " +
```
jamesfredley: consider LOG.error - since silent degradation of @ConditionalOnProperty is hard to debug

jamesfredley (approved):
We should consider Adding tests for "both yml and groovy throws exception" 

Should document Breaking constructor changes for anyone subclassing GrailsApplicationPostProcessor or DefaultGrailsPluginManager

grails-bootstrap/src/main/groovy/grails/plugins/GrailsPluginInfo.java
```diff
@@ -60,7 +60,7 @@ public interface GrailsPluginInfo {
     String getFullName();
 
     /**
-     * Returns the location of the Resource that represents the plugin descriptor (the *GrailsPlugin.groovy file)
+     * Returns the location of the Resource that represents the plugin descriptor (the *GrailsPlugin.groovy or grails-plugin.yml file)
```
matrei: Isn't the `grails-plugin.yml` created from the `*GrailsPlugin.groovy` file so this is essentially the same thing.

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
matrei, line 1: Why not Groovy?

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
jdaugherty: These have traditionally been warnings, but I think this is a good idea.  Changing to error ...

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
```diff
@@ -0,0 +1,593 @@
...
+            findPlugins(environment);
+        }
+
+        return orderedPlugins;
+    }
+
+    /**
+     * @return the order the plugins were loaded in.
+     */
+    public Collection<GrailsPluginInfo> getLoadOrderedPlugins(Environment environment) {
```
matrei: Refrain from using the get* ideom for non-getters?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
```diff
@@ -0,0 +1,593 @@
...
+import grails.plugins.GrailsVersionUtils;
+import grails.plugins.PluginFilter;
+import grails.plugins.exceptions.PluginException;
+import grails.util.Metadata;
+import org.apache.grails.core.plugins.filters.PluginFilterRetriever;
+import org.grails.core.io.CachingPathMatchingResourcePatternResolver;
+import org.grails.io.support.GrailsResourceUtils;
+
+/**
+ * This class provides the canonical implementations of Grails Plugin Discovery.
```
matrei: I don't understand this sentence. Does it mean: "This class finds and provides access to the Grails plugins on the classpath?

matrei:
@jdaugherty Why did you choose to write new classes in Java, and not Groovy?

grails-testing-support-core/src/main/resources/META-INF/services/grails.plugins.GrailsPluginConfigFilter
matrei, line 1: What is this file used for?

grails-core/src/main/groovy/org/grails/plugins/MockGrailsPluginManager.java
```diff
@@ -28,14 +28,20 @@
...
 
 /**
  * @author Graeme Rocher
  * @since 0.4
  */
 public class MockGrailsPluginManager extends AbstractGrailsPluginManager {
     public MockGrailsPluginManager(GrailsApplication application) {
-        super(application);
+        super(application, new MockGrailsPluginDiscovery());
+        loadPlugins();
```
matrei:
```suggestion
        this(application, new MockGrailsPluginDiscovery());
```

grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsApplicationBuilder.groovy
```diff
@@ -151,11 +152,20 @@ class GrailsApplicationBuilder {
...
     protected void prepareContext(ConfigurableApplicationContext applicationContext, ConfigurableBeanFactory beanFactory) {
-        registerGrailsAppPostProcessorBean(beanFactory)
+        def discovery = registerPluginDiscoveryBean(beanFactory)
+        registerGrailsAppPostProcessorBean(beanFactory, discovery)
         AnnotationConfigUtils.registerAnnotationConfigProcessors((BeanDefinitionRegistry) beanFactory)
         new ConfigDataApplicationContextInitializer().initialize(applicationContext)
     }
 
+    protected GrailsPluginDiscovery registerPluginDiscoveryBean(ConfigurableBeanFactory beanFactory) {
+        GrailsPluginDiscovery discovery = new GrailsPluginDiscovery()
```
matrei: Use `def` for local variables where the type can be inferred?

grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsApplicationBuilder.groovy
```diff
@@ -151,11 +152,20 @@ class GrailsApplicationBuilder {
...
+        // we must load the classpath since the plugin manager needs to find the default plugins
+        discovery.setPluginFilter(new IncludingPluginFilter(includePlugins ?: DEFAULT_INCLUDED_PLUGINS))
```
matrei: Use Groovy property setter?

grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsApplicationBuilder.groovy
```diff
@@ -151,11 +152,20 @@ class GrailsApplicationBuilder {
...
+        (beanFactory as DefaultListableBeanFactory).registerSingleton(GrailsPluginDiscovery.BEAN_NAME, discovery)
```
matrei: Why cast?

grails-test-suite-uber/src/test/groovy/org/grails/plugins/PluginLoadOrderTests.groovy
```diff
@@ -62,16 +67,21 @@ class ThreeGrailsPlugin {
         def three = gcl.loadClass("ThreeGrailsPlugin")
         def four = gcl.loadClass("FourGrailsPlugin")
         def five = gcl.loadClass("FiveGrailsPlugin")
-        def pluginManager = new DefaultGrailsPluginManager([one,two,three, four,five] as Class[],
-            new DefaultGrailsApplication())
-
-        pluginManager.loadCorePlugins = false
+        GrailsPluginDiscovery pluginDiscovery = new GrailsPluginDiscovery([one, two, three, four, five] as Class[],)
```
matrei: `def`?

grails-test-suite-uber/src/test/groovy/org/grails/commons/GrailsPluginManagerTests.groovy
```diff
@@ -116,12 +117,13 @@ hibernate {
     }
 
     void testDefaultGrailsPluginManager() {
-        DefaultGrailsPluginManager manager = new DefaultGrailsPluginManager(RESOURCE_PATH,ga)
-        assertEquals(1, manager.getPluginResources().length)
+        def discovery = new GrailsPluginDiscovery(RESOURCE_PATH)
+        DefaultGrailsPluginManager manager = new DefaultGrailsPluginManager(ga, discovery)
```
matrei: `manager` is unused and can be removed

grails-test-suite-uber/src/test/groovy/org/grails/commons/GrailsPluginManagerTests.groovy
```diff
@@ -116,12 +117,13 @@ hibernate {
...
+        assertEquals(1, discovery.getPluginResources().length)
     }
 
     void testLoadPlugins() {
-        GrailsPluginManager manager = new DefaultGrailsPluginManager(RESOURCE_PATH,ga)
+        GrailsPluginManager manager = new DefaultGrailsPluginManager(ga, new GrailsPluginDiscovery())
```
matrei: `def`?

grails-test-suite-uber/src/test/groovy/org/apache/grails/core/plugins/filters/PluginFilterFactoryTests.java
```diff
@@ -16,59 +16,73 @@
...
+
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@SuppressWarnings("rawtypes")
 public class PluginFilterFactoryTests {
 
+    MockEnvironment createEnvironment(String includes, String excludes) {
+        MockEnvironment env = new MockEnvironment();
```
matrei: `var`?

grails-test-suite-uber/src/test/groovy/org/apache/grails/core/plugins/filters/PluginFilterFactoryTests.java
```diff
@@ -16,59 +16,73 @@
...
+    public void testIncludeFilterOne() {
         PluginFilterRetriever fb = new PluginFilterRetriever();
-        PluginFilter bean = fb.getPluginFilter("one", null);
-        assertTrue(bean instanceof IncludingPluginFilter);
+        PluginFilter bean = fb.getPluginFilter(createEnvironment("one", null));
+        assertInstanceOf(IncludingPluginFilter.class, bean);
 
         IncludingPluginFilter filter = (IncludingPluginFilter)bean;
-        Set suppliedNames = filter.getSuppliedNames();
+        Set<String> suppliedNames = filter.getSuppliedNames();
```
matrei: `var` (and also above)?

grails-test-suite-uber/src/test/groovy/org/apache/grails/core/plugins/filters/PluginFilterFactoryTests.java
```diff
@@ -16,59 +16,73 @@
...
+    public void testIncludeFilter() {
         PluginFilterRetriever fb = new PluginFilterRetriever();
-        PluginFilter bean = fb.getPluginFilter("one, two", " three , four ");
-        assertTrue(bean instanceof IncludingPluginFilter);
+        PluginFilter bean = fb.getPluginFilter(createEnvironment("one, two", " three , four "));
+        assertInstanceOf(IncludingPluginFilter.class, bean);
 
         IncludingPluginFilter filter = (IncludingPluginFilter)bean;
-        Set suppliedNames = filter.getSuppliedNames();
+        Set<String> suppliedNames = filter.getSuppliedNames();
```
matrei: `var` (and also above)?

grails-test-suite-uber/src/test/groovy/org/apache/grails/core/plugins/filters/PluginFilterFactoryTests.java
```diff
@@ -16,59 +16,73 @@
...
+    public void testExcludeFilter() {
         PluginFilterRetriever fb = new PluginFilterRetriever();
-        PluginFilter bean = fb.getPluginFilter(null, " three , four ");
-        assertTrue(bean instanceof ExcludingPluginFilter);
+        PluginFilter bean = fb.getPluginFilter(createEnvironment(null, " three , four "));
+        assertInstanceOf(ExcludingPluginFilter.class, bean);
 
         ExcludingPluginFilter filter = (ExcludingPluginFilter)bean;
-        Set suppliedNames = filter.getSuppliedNames();
+        Set<String> suppliedNames = filter.getSuppliedNames();
```
matrei: `var` (and also above)?

grails-test-suite-uber/src/test/groovy/org/apache/grails/core/plugins/filters/PluginFilterFactoryTests.java
```diff
@@ -16,59 +16,73 @@
...
         assertTrue(suppliedNames.contains("four"));
     }
 
     @Test
-    public void testDefaultFilter() throws Exception {
+    public void testDefaultFilter() {
         PluginFilterRetriever fb = new PluginFilterRetriever();
-        PluginFilter bean = fb.getPluginFilter(null, null);
-        assertTrue(bean instanceof IdentityPluginFilter);
+        PluginFilter bean = fb.getPluginFilter(createEnvironment(null, null));
```
matrei: `var` and also above?

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
```diff
@@ -71,17 +78,18 @@ class GrailsMicronautGrailsPlugin extends Plugin {
         [plugins, pluginsFromContext].each { pluginsToProcess ->
             pluginsToProcess
                     .findAll { it.propertySource != null }
```
matrei, line 80: You have deprecated `GrailsPlugin.propertySource` but it is still used here?

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
```diff
@@ -71,17 +78,18 @@ class GrailsMicronautGrailsPlugin extends Plugin {
...
-                                )
-                        )
+                    .each { plugin ->
+                        // Look up the plugin's property source from the Spring environment,
+                        // where it was loaded by GrailsEnvironmentPostProcessor
+                        String pluginName = plugin.name
+                        String ymlSourceName = pluginName + PLUGIN_YML_SUFFIX
+                        String groovySourceName = pluginName + PLUGIN_GROOVY_SUFFIX
+                        org.springframework.core.env.PropertySource<?> springPs =
+                                springPropertySources.get(ymlSourceName) ?: springPropertySources.get(groovySourceName)
```
matrei: 4 x `def`

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
```diff
@@ -71,17 +78,18 @@ class GrailsMicronautGrailsPlugin extends Plugin {
...
+                        if (springPs instanceof EnumerablePropertySource) {
+                            log.debug('Loading configurations from {} plugin to the parent Micronaut context', pluginName)
+                            micronautEnv.addPropertySource(PropertySource.of("grails.plugins.${pluginName}", (Map) springPs.getSource(), --priority))
```
matrei: Use Groovy property getter?

grails-core/src/test/groovy/org/grails/plugins/GrailsPluginTests.groovy
```diff
@@ -162,16 +165,36 @@ class TestGrailsPlugin {
...
-        def pluginManager = new DefaultGrailsPluginManager([test1] as Class[], application)
-
-pluginManager.loadPlugins()
+        
+        // Create and set up application context
+        def appCtx = new GenericApplicationContext()
+        application.setMainContext(appCtx)
+        
+        // Create discovery bean configured for unit tests
+        GrailsPluginDiscovery discovery = new GrailsPluginDiscovery(new Class<?>[]{test1})
```
matrei: `def`?

grails-core/src/test/groovy/org/grails/plugins/GrailsPluginTests.groovy
```diff
@@ -162,16 +165,36 @@ class TestGrailsPlugin {
...
 
         String originalEnv = System.getProperty(Environment.KEY)
         try {
             System.setProperty(Environment.KEY, Environment.PRODUCTION.getName())
 
-            pluginManager = new DefaultGrailsPluginManager([test1] as Class[], application)
+            // Create new application and context for production environment test
+            application = new DefaultGrailsApplication()
+            appCtx = new GenericApplicationContext()
+            application.setMainContext(appCtx)
```
matrei: Groovy property setter?

grails-core/src/test/groovy/org/grails/plugins/GrailsPluginTests.groovy
```diff
@@ -162,16 +165,36 @@ class TestGrailsPlugin {
...
+            
+            // Create new discovery bean for production environment test
+            discovery = new GrailsPluginDiscovery(new Class<?>[]{test1})
+            discovery.setLoadClasspathPlugins(false)
```
matrei: Groovy property setter?

grails-core/src/test/groovy/org/grails/plugins/GrailsPluginConfigurationClass.groovy
```diff
@@ -50,81 +43,18 @@ class GrailsPluginConfigurationClass extends GrailsAutoConfiguration {
         final String grailsVersion = '4.0.1'
         def gcl = new GroovyClassLoader()
 
-        GrailsPlugin plugin = new MockTestGrailsPlugin(gcl.parseClass("""class TestGrailsPlugin {
+        GrailsPlugin plugin = new DefaultGrailsPlugin(gcl.parseClass("""class TestGrailsPlugin {
```
matrei: `def`?

grails-core/src/test/groovy/org/grails/plugins/GrailsPluginConfigurationClass.groovy
```diff
@@ -50,81 +43,18 @@ class GrailsPluginConfigurationClass extends GrailsAutoConfiguration {
...
         def version = '1.0.0'
         def grailsVersion = '$grailsVersion'
         def loadAfter = ['testTwo']
 }"""), grailsApplication)
 
-        GrailsPlugin plugin2 = new MockTestTwoGrailsPlugin(gcl.parseClass("""class TestTwoGrailsPlugin {
+        GrailsPlugin plugin2 = new DefaultGrailsPlugin(gcl.parseClass("""class TestTwoGrailsPlugin {
```
matrei: `def`?

grails-core/src/test/groovy/org/grails/plugins/BinaryPluginSpec.groovy
```diff
@@ -72,67 +71,57 @@ class BinaryPluginSpec extends Specification {
...
-        def descriptor = new BinaryGrailsPluginDescriptor(new ByteArrayResource(testBinary.getBytes('UTF-8')), ['org.grails.plugins.TestBinaryResource'])
-        MockConfigBinaryGrailsPlugin.YAML_EXISTS = true
-        MockConfigBinaryGrailsPlugin.GROOVY_EXISTS = true
-        new MockConfigBinaryGrailsPlugin(descriptor)
+        def descriptor = new GrailsPluginDescriptor(new ByteArrayResource(testBinary.getBytes('UTF-8')), ['org.grails.plugins.TestBinaryResource'], [])
+        def binaryPlugin = new BinaryGrailsPlugin(TestBinaryGrailsPlugin, descriptor, new DefaultGrailsApplication())
 
         then:
-        thrown(RuntimeException)
+        binaryPlugin.getPropertySource() == null
```
matrei: Groovy property getter?

grails-core/src/test/groovy/org/grails/plugins/BinaryPluginSpec.groovy
```diff
@@ -72,67 +71,57 @@ class BinaryPluginSpec extends Specification {
...
-        def descriptor = new BinaryGrailsPluginDescriptor(new ByteArrayResource(testBinary.getBytes('UTF-8')), ['org.grails.plugins.TestBinaryResource'])
-        MockConfigBinaryGrailsPlugin.YAML_EXISTS = false
-        MockConfigBinaryGrailsPlugin.GROOVY_EXISTS = true
-        def binaryPlugin = new MockConfigBinaryGrailsPlugin(descriptor)
+        grailsApp.setMainContext(appCtx)
 
         then:
-        binaryPlugin.propertySource.getProperty('bar') == "foo"
+        binaryPlugin.getPropertySource() != null
+        binaryPlugin.getPropertySource().getProperty('foo') == 'bar'
```
matrei: Groovy property getters?

grails-core/src/test/groovy/org/grails/core/io/ResourceLocatorSpec.groovy
```diff
@@ -18,9 +18,13 @@
  */
 package org.grails.core.io
 
-import groovy.xml.XmlSlurper
+
```
matrei, line 26: Remove empty line?

grails-core/src/test/groovy/org/grails/core/io/ResourceLocatorSpec.groovy
```diff
@@ -73,9 +80,15 @@ class ResourceLocatorSpec extends Specification {
...
 
             def resource = new MockBinaryPluginResource(str.bytes)
-            def descriptor = new BinaryGrailsPluginDescriptor(resource, ['org.grails.plugins.TestBinaryGrailsPlugin'])
+            def descriptor = new GrailsPluginDescriptor(resource, ['org.grails.plugins.TestBinaryGrailsPlugin'], [])
             resource.relativesResources['static/css/main.css'] = new ByteArrayResource(''.bytes)
-            def binaryPlugin = new BinaryGrailsPlugin(TestBinaryGrailsPlugin, descriptor, new DefaultGrailsApplication())
+            
+            def grailsApp = new DefaultGrailsApplication()
+            def appCtx = new org.springframework.context.support.GenericApplicationContext()
+            grailsApp.setMainContext(appCtx)
```
matrei: - Use Groovy property getter?
- Inline `new GenericApplicationContext()`
- Import `org.springframework.context.support.GenericApplicationContext`

grails-core/src/test/groovy/org/grails/core/io/ResourceLocatorSpec.groovy
```diff
@@ -73,9 +80,15 @@ class ResourceLocatorSpec extends Specification {
             def xml = new XmlSlurper().parseText(str)
```
matrei: `xml` is unused

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.grails.core.plugins
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+import java.net.URLClassLoader
```
matrei: Unused import

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+class GrailsPluginUtilsSpec extends Specification {
+
+    @Unroll
+    def "isPluginVersionCompatible should check that plugin with grailsVersion=#pluginGrailsVersion is compatible with grails #grailsVersion"() {
+        when:
+        def compatible = GrailsPluginUtils.isPluginVersionCompatible(
+            "1.0.0",  // pluginVersion
+            pluginGrailsVersion,  // pluginSupportedVersion
+            grailsVersion,  // grailsVersion
+            "test-plugin"  // pluginDescription
```
matrei: Single quotes?

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+        grailsVersion | pluginGrailsVersion        || expectedCompatible
+        "1.0"         | "3.3.1 > *"                || false
+        "2.5"         | "3.0.1"                    || false
+        "3.0.0"       | "3.3.10 > *"               || false
+        "3.3.10"      | "4.0.0 > *"                || false
+        "4.0.1"       | "3.0.0.BUILD-SNAPSHOT > *" || true
+        "4.0.1"       | "4.0.1"                    || true
+        "4.0.1"       | "3.0.1"                    || false
+        "4.0.1"       | "3.3.1 > *"                || true
+        "4.0.1"       | "3.3.10 > *"               || true
```
matrei: Single quotes?

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    def "isPluginVersionCompatible should handle null grailsVersion"() {
+        when:
+        def compatible = GrailsPluginUtils.isPluginVersionCompatible(
+            "1.0.0",
+            "3.3.10 > *",
+            null,  // grailsVersion is null
+            "test-plugin"
```
matrei: Single quotes?

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+        compatible == true
+    }
+
+    def "isPluginVersionCompatible should handle null pluginSupportedVersion"() {
+        when:
+        def compatible = GrailsPluginUtils.isPluginVersionCompatible(
+            "1.0.0",
+            null,  // pluginSupportedVersion is null
+            "4.0.1",
+            "test-plugin"
```
matrei: Single quotes?

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginUtilsSpec.groovy
```diff
@@ -0,0 +1,333 @@
...
+        compatible == true
+    }
+
+    def "isPluginVersionCompatible should handle @ in pluginSupportedVersion"() {
+        when:
+        def compatible = GrailsPluginUtils.isPluginVersionCompatible(
+            "1.0.0",
+            "4.0.0@",  // contains @
+            "3.0.0",
+            "test-plugin"
```
matrei: Single quotes?

grails-core/src/test/groovy/org/apache/grails/core/plugins/GrailsPluginDiscoverySpec.groovy
```diff
@@ -0,0 +1,303 @@
...
+package org.apache.grails.core.plugins
+
+import spock.lang.Specification
+
+class GrailsPluginDiscoverySpec extends Specification {
+
+     def "scanPluginDescriptorResources discovers descriptors with rich info from classpath"() {
+         when:
+         def descriptors = GrailsPluginUtils.scanPluginDescriptorResources(
+                 Thread.currentThread().getContextClassLoader())
```
matrei: Groovy property getter?

grails-core/src/test/groovy/grails/plugins/DefaultGrailsPluginManagerSpec.groovy
```diff
@@ -19,79 +19,40 @@
...
         given:
-        GrailsApplication app = stubGrailsApplicationWithVersion("4.0.1")
-        DefaultGrailsPluginManager sut = buildPluginsManager(app)
-        DefaultGrailsPlugin plugin = stubPluginWithGrailsVersion(app, pluginGrailsVersion)
+        GrailsApplication app = Mock(GrailsApplication)
+        GrailsPluginDiscovery discovery = Mock(GrailsPluginDiscovery)
 
         when:
-        def version = sut.getPluginGrailsVersion(plugin)
+        DefaultGrailsPluginManager manager = new DefaultGrailsPluginManager(app, discovery)
```
matrei: `def` x 3

grails-core/src/test/groovy/grails/plugins/DefaultGrailsPluginManagerSpec.groovy
```diff
@@ -19,79 +19,40 @@
...
         given:
-        GrailsApplication app = stubGrailsApplicationWithVersion(grailsVersion)
-        DefaultGrailsPluginManager sut = buildPluginsManager(app)
-        DefaultGrailsPlugin plugin = stubPluginWithGrailsVersion(app, pluginGrailsVersion)
+        GrailsApplication app = Mock(GrailsApplication)
+        GrailsPluginDiscovery discovery = Mock(GrailsPluginDiscovery)
 
         when:
-        def compatible = sut.isCompatiblePlugin(plugin)
+        DefaultGrailsPluginManager manager = new DefaultGrailsPluginManager(app, discovery)
```
matrei: `def` x 3

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
```diff
@@ -0,0 +1,121 @@
...
+package org.apache.grails.core.plugins;
+
+import org.springframework.core.io.Resource;
+
+/**
+ * Lightweight value class holding plugin metadata needed for ordering and
+ * configuration loading. Wraps {@link GrailsPluginLoadMetadata}
+ * with an additional configuration resource reference.
+ */
+public final class GrailsPluginInfo {
```
matrei: Why `final`?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
```diff
@@ -0,0 +1,121 @@
...
+    private final boolean dynamic;
+
+    /**
+     * Creates a new {@code PluginInfo}.
+     *
+     * @param metadata the plugin metadata from {@link GrailsPluginDiscovery}
+     * @param configResource the plugin's configuration resource ({@code plugin.yml} or
+     *        {@code plugin.groovy}), or {@code null} if no config file exists
+     */
+    GrailsPluginInfo(GrailsPluginDescriptor pluginDescriptor, GrailsPluginLoadMetadata metadata, Resource configResource, boolean dynamic) {
```
matrei: Why `package-private`?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginLoadMetadata.java
```diff
@@ -0,0 +1,89 @@
...
+ * @param name the logical plugin name (e.g., "core", "myPlugin")
+ * @param grailsVersion the grailsVersion this plugin supports
+ * @param pluginClass the plugin's class
+ * @param loadAfterNames plugin names this plugin should load after
+ * @param loadBeforeNames plugin names this plugin should load before
+ * @param dependsOnNames plugin names this plugin depends on (used for
+ *        transitive dependency resolution during filtering, not for
+ *        sort ordering)
+ * @param environments the environments this plugin is enabled for, or empty if enabled for all environments
+ * @param status the status of the plugin
```
matrei: There is no `status` param, should it be `enabled`?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
```diff
@@ -0,0 +1,121 @@
...
+        return metadata().grailsVersion();
+    }
+
+    public String[] evictions() {
+        return metadata.evictions();
+    }
+
+    public GrailsPluginDescriptor pluginDescriptor() {
+        return pluginDescriptor;
+    }
```
matrei: Why the record-style naming instead of getters?

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
```diff
@@ -75,7 +75,7 @@ abstract class Plugin implements GrailsApplicationLifeCycle, GrailsApplicationAw
     /**
      * Whether the plugin is enabled
      */
-    boolean enabled = true
+    public boolean enabled = true
```
matrei, line 78: Why `public`?

grails-core/src/main/groovy/org/grails/plugins/AbstractGrailsPluginManager.java
```diff
@@ -124,10 +144,8 @@ protected void checkInitialised() {
     }
 
     public GrailsPlugin getFailedPlugin(String name) {
-        if (name.indexOf('-') > -1) {
-            name = GrailsNameUtils.getPropertyNameForLowerCaseHyphenSeparatedName(name);
-        }
-        return failedPlugins.get(name);
+        GrailsPluginInfo metadata = pluginDiscovery.getFailedPlugins().get(GrailsPluginUtils.normalizePluginName(name));
```
matrei: `var`?

grails-core/src/main/groovy/org/grails/plugins/AbstractGrailsPlugin.java
```diff
@@ -93,32 +80,43 @@ public AbstractGrailsPlugin(Class<?> pluginClass, GrailsApplication application)
...
+     * early in the application lifecycle by
+     * {@link grails.boot.config.GrailsEnvironmentPostProcessor} and registered as named
+     * property sources in the environment. This method looks up the property source by the
+     * expected name ({@code "<pluginName>-plugin.yml"} or {@code "<pluginName>-plugin.groovy"}).</p>
+     *
+     * @return the plugin's property source, or {@code null} if no configuration was loaded
+     *         or the application context is not yet available
+     */
     @Override
+    @Deprecated(forRemoval = true)
```
matrei: since?

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.boot.BootstrapRegistryInitializer;
+
+import org.apache.grails.core.plugins.GrailsPluginDiscovery;
+
+/**
+ * Registers the {@link GrailsPluginDiscovery} in the Spring Boot Bootstrap context so it can be accessed during
+ * the early lifecycle of the application & later promoted to actual bean.
+ *
+ * <p>This ensures that both the early-lifecycle
+ * {@link GrailsEnvironmentPostProcessor} and the
```
matrei: Use full path?

grails-core/src/main/groovy/grails/plugins/GrailsPluginSorter.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.function.Function;
+
+/**
+ * Canonical topological sort for Grails plugins based on {@code loadAfter}
+ * and {@code loadBefore} declarations.
+ *
+ * <p>This class provides the single shared implementation of plugin ordering that
+ * is used by both {@link DefaultGrailsPluginManager} (at runtime, operating on
+ * {@link GrailsPlugin} instances) and
+ * {@link GrailsEnvironmentPostProcessor} (early in the
```
matrei: Use full path?

grails-core/src/main/groovy/grails/plugins/GrailsPluginSorter.java
```diff
@@ -0,0 +1,218 @@
...
+     * @param nameExtractor     extracts the logical plugin name from a plugin
+     * @param loadAfterExtractor extracts the {@code loadAfter} names from a plugin
+     * @param loadBeforeExtractor extracts the {@code loadBefore} names from a plugin
+     * @param pluginLookup      resolves a plugin by its logical name (returns {@code null}
+     *                          if the named plugin is not present)
+     * @return a new list with the plugins in topological order
+     */
+    public static <T> List<T> sort(
+            List<T> plugins,
+            Function<T, String> nameExtractor,
```
matrei: `nameExtractor` is unused

grails-core/src/main/groovy/grails/plugins/GrailsPlugin.java
```diff
@@ -263,6 +263,20 @@ public interface GrailsPlugin extends ApplicationContextAware, Comparable, Grail
...
+     * <p>Plugin configuration is loaded early in the Spring Boot lifecycle by
+     * {@code GrailsEnvironmentPostProcessor} and registered as a
+     * property source named {@code <pluginName>-plugin.yml} or {@code <pluginName>-plugin.groovy}.
+     * This method retrieves that property source from the application context's environment.</p>
+     *
+     * @deprecated no longer used to populate plugin-specific configuration, see {@code GrailsEnvironmentPostProcessor}
+     * @return the plugin's property source from the environment, or {@code null} if the
+     *         GrailsApplication main context is not yet available or no matching property source exists
+     */
+    @Deprecated(forRemoval = true)
```
matrei: `since`?

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
matrei, line 1: Why Java (general question for all new classes)?

matrei (commented):
This was a great effort @jdaugherty, thank you!

As changes a some of the public API of core functionality, I think it would be wise to target 8.0.x.

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
jdaugherty: There was precedent for java for the other processors and it's typically more performant.  So I left this as java.

grails-bootstrap/src/main/groovy/grails/plugins/GrailsPluginInfo.java
jdaugherty: There are 2 code paths - binary plugins load from grails-plugin.yml & source based plugins load from GrailsPlugin.  Source based plugins that are loaded dynamically do not have a grails-plugin.yml.  I think this really exists today for tests only.  I was trying to call out these 2 code paths in this comment.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
jdaugherty: Not just the classpath; it's responsible for finding configured Grails Plugins.  The problem is the existing framework has 2 code paths - one that dynamically injects plugins (I think this is really just for tests) and one for searching on the classpath.  I can change this to: "This class is responsible for locating Grails Plugins that should be loaded."

grails-testing-support-core/src/main/resources/META-INF/services/grails.plugins.GrailsPluginConfigFilter
jdaugherty: I originally had pulled this out into a service loader so I could discontinue all of the dynamic test registration, but I later released just how much that code path was used and how much fo a breaking change it was for unit tests.  This is unused now, and I'll remove it.

grails-core/src/main/groovy/grails/boot/config/GrailsEnvironmentPostProcessor.java
jdaugherty: The classes in this area have historically been java.  Looking through the commit history, there have been numerous performance problems in this area and conversion to java was one of the "fixes".  I did not want to introduce a new pattern as I was only trying to abstract the existing logic.  I think if we introduce a hard contract, we could change this with some significant refactoring.

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
jdaugherty: I started to move several of the classes from grails.plugins into an apache equivalent.  Since i'm accessing across packages now, I had to make it public.

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
jdaugherty: I deprecated it to help us find external usages of it.  I'm not convinced it's used for anything but here.  I was trying to defer a decision on this until later.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
jdaugherty: This originally started out as a record, but I later changed it when I realized there were nested classes.  I left this to make the lambda function references clear: `GrailsPluginInfo::name`

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
jdaugherty: There was an existing convention in Grails where org.grails => public grails.* => non-public.  Since I'm switching to the org.apache.grails.* package, I didn't want to have to create separate packages (it's non-compliant with the java modular system).  My solution was to mark classes that would have gone in grails.* as package private instead.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
jdaugherty: See this comment: https://github.com/apache/grails-core/pull/15409/changes#r2926329927 

Effectively, I don't think this should be overridden as it's a private API and since it's going in what I would consider a public package, I wanted to be clear on the purpose.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
jdaugherty: I originally tried to have 1 record that was just loaded into, but then I realized that there were nested holders all over the place.  My initial attempt was to combine them into one, and my assumption is we'd clean this stuff up later - most likely in Grails 8.  

The other main benefit was just to pretty names when reference lambda functions; i.e. `GrailsPluginInfo::name`

jdaugherty:
> This was a great effort @jdaugherty, thank you!
> 
> As changes a some of the public API of core functionality, I think it would be wise to target 8.0.x.

The problem I have with 8.0 is:
1. this works with all of the plugins in my current application with no breakage (that I can tell). 
2. a core piece of Spring 3.x is broken currently without this change
3. there is a precedent of us already committing breaking changes to 7.1.
4. we can do a milestone to see if there's impact and then revert if there is?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
jdaugherty: I'm going to refactor the GrailsPluginDiscovery class so that it has a common interface & then we ship a default implementation.  I'm also going to add an init() method instead of initing on the first time this is called.  This way the getters can be true getters.

jamesfredley:
A potential start on backward-compatibility bridge shims for 7.1.x:
https://github.com/jdaugherty/grails-core/pull/5

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
matrei: I think we need to have and use an alternative code path here if this code path is to be deprecated.

grails-core/src/test/groovy/org/grails/core/io/ResourceLocatorSpec.groovy
matrei: Why are we parsing this string?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
matrei: - I think it is a mistake to mix these conventions. `getName` is also very clear for lambda references.
- Also, I think we should try to remove the word `Grails` from class names where we can. It is not adding any value.
- On the Java vs Groovy question, I think we should try to be a flag holder of the Groovy language and if there is no documented gain of using Java, we should write statically compiled Groovy.

matrei:
> The problem I have with 8.0 is:
> 
This is excellent, but is it relevant?
>     1. this works with all of the plugins in my current application with no breakage (that I can tell).
`@ConditionalOnProperty` was introduced in Spring Boot 1.1
>     2. a core piece of Spring 3.x is broken currently without this change
Yeah, but not in this way, and that was a hard call.
>     3. there is a precedent of us already committing breaking changes to 7.1.
Unfortunately, not many projects try out milestones.
>     4. we can do a milestone to see if there's impact and then revert if there is?

If we can get the API backwards compatible, I have no problem including this in 7.1. I see that @jamesfredley already started this work in https://github.com/jdaugherty/grails-core/pull/5.

jamesfredley:
I plan to approve this once the backward-compatibility bridge shims are in place for 7.1.x.  Thank you @jdaugherty

jdaugherty:
@matrei @jamesfredley I believe I've updated all of the shims so that the old way can still be used (the plugin discovery process will just reinit in this case).

jdaugherty:
I'm expecting all of these tests to now pass.  I've also started my main application against this branch and made sure there are no issues with application start.  I've previously mentioned this, but longer term we need to better define our plugin dependencies in core so they load in the same order.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
jdaugherty: I'm not comfortable rewriting this from Java to Groovy in a minor release. I'm happy to take on this work for 8.0 though.  Let's ticket this and I'll do it in 8.0?

grails-micronaut/src/main/groovy/org/apache/grails/micronaut/GrailsMicronautGrailsPlugin.groovy
jdaugherty: I've remove the depreciation for now.

jdaugherty:
@matrei  I think I've responded to all of your feedback.  Concerning this comment: 

`@ConditionalOnProperty was introduced in Spring Boot 1.1`

While it was introduced in Spring Boot 1.1 & did not work in priors Grails Versions, Grails 7 was the first release where we started using `@Configuration` & `@ConditionalOnProperty`.  For example: `UrlMappingsAutoConfiguration` uses `@ConditionalOnProperty`.  In previous Grails versions, one could use the plugin architecture to replace beans, in Grails 7 this isn't possible due the bug fixed by this PR.

Given that we have adopted a solution that our end users can't override with Grails Plugins, I view this as a critical issue and it should be fixed in 7.1.x.

jdaugherty:
Looking like I missed responding to this one too:  @matrei 

    This is excellent, but is it relevant?

    1. this works with all of the plugins in my current application with no breakage (that I can tell).

Grails-core doesn't depend on external plugins.  My internal application is using plugins across github.com/grails-plugins, github.com/gpc, custom (and nested) plugins, core plugins, & several apache grails-spring-security plugins.  I'm just pointing out that I've regression tested these changes to limit the impact given the scope of these changes.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @return Plugins that were discovered during the bootstrap process, no ordering guaranteed
+     */
+    Collection<GrailsPluginInfo> getPlugins();
+
+    /**
+     * @param pluginName a plugin name to search for, the name may not be normalized
+     * @return the plugin information if found
+     */
+    GrailsPluginInfo getPlugin(String pluginName);
```
matrei: Use a method name that does not use the getter idiom, like `findPlugin(String pluginName)`?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
```diff
@@ -0,0 +1,148 @@
...
+     * @param name the name of the plugin - it may not be normalized
+     * @return true if a plugin with the given name exists, false otherwise
+     */
+    boolean hasPlugin(String name);
+
+    /**
+     * @param plugin the plugin to get the observers for
+     * @return the plugins that are observing the given plugin, empty if none
+     */
+    Collection<GrailsPluginInfo> getPluginObservers(GrailsPluginInfo plugin);
```
matrei: Use a method name that does not use the getter idiom, like `findPluginObservers(GrailsPluginInfo plugin)`?

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
matrei: There is still:
- `getPlugin(String pluginName)`
- `getPlugin(String pluginName, Object version)`
- `getPluginObservers(GrailsPluginInfo plugin)`

Can we use `findPlugin*`?

grails-core/src/test/groovy/org/grails/core/io/ResourceLocatorSpec.groovy
matrei: Question remains

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginInfo.java
matrei: > Why the record-style naming instead of getters?

The question remains, the class naming and Java vs Groovy was broader sidetracks.

matrei (commented):
I also see some unresolved comments from the last review.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
jdaugherty: I refactored this method into 2 separate processes.  There's now an init() call instead of initializing on the get.  This means the getters are just getters now.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
jdaugherty: I refactored this method into 2 separate processes. There's now an init() call instead of initializing on the get. This means the getters are just getters now.

grails-core/src/main/groovy/org/apache/grails/core/plugins/GrailsPluginDiscovery.java
matrei: A getter does typically not take parameters.

jdaugherty:
@matrei I've updated this to move away from the record syntax, rename the classes, & to change `get*(_)` to `find*(_)`.  Please let me know if there's anything else that I've missed.

jdaugherty:
Mongo test failed because it couldn't reach the mongo server: 

       org.gradle.internal.remote.internal.ConnectException: Could not connect to server [873da1e7-52b1-4b27-b072-abcf59eb5992 port:32787, addresses:[/127.0.0.1]]. Tried addresses: [/127.0.0.1].

I expect a retry will fix this.  This is passing locally.

matrei (approved):

Issue #15321, opened 2026-01-17
Disable invoke dynamic by default


jdaugherty:
See associated discussion here: https://lists.apache.org/thread/6vh8ld1bjh31z9p0682fkn1q5hrlyt7y 

This ticket is to disable invoke dynamic by default until progress is made on https://github.com/apache/grails-core/issues/15293

PR #15415, opened 2026-02-19
test(ci): add indy matrix to functional test CI jobs

jamesfredley:
## Summary

- Add `indy: [false, true]` matrix dimension to all functional test CI jobs to ensure Groovy invokedynamic is tested both enabled and disabled
- Add `-PgrailsIndy` property toggle to `grails-extension-gradle-config.gradle` so all 45 test example projects respect the CI matrix setting
- Include indy status in job names for clear CI dashboard visibility

## Context

PR #15375 moved indy configuration from generated `build.gradle` files to the Grails Gradle Plugin (`GrailsExtension.indy`, default: `false`). After that merge, CI was only testing with the default `indy=false`, leaving `indy=true` completely untested. If a user sets `grails { indy = true }`, we had zero CI coverage for that code path.

### `gradle/grails-extension-gradle-config.gradle`
- Added conditional `indy` property toggle inside the `grails { }` block
- When `-PgrailsIndy=true` is passed, sets `grails { indy = true }` on all test example projects
- All 45 test example build files apply this config, so no per-project changes needed

### `.github/workflows/gradle.yml`
- **Functional Tests**: 3 Java versions x 2 indy = 6 jobs (was 3)
- **Hibernate5 Functional Tests**: 2 Java versions x 2 indy = 4 jobs (was 2)
- **MongoDB Functional Tests**: 2 Java versions x 2 MongoDB versions x 2 indy = 8 jobs (was 4)
- **Build Grails Forge**: 2 Java versions x 2 indy = 4 jobs (was 2)
- Job names now include `indy=true/false` for dashboard clarity

## Note on Forge Tests

The forge `test-core` tests use Gradle TestKit, which creates an isolated build environment. The generated apps resolve the Grails Gradle plugin from Maven repositories rather than the local composite build, so the `-PgrailsIndy` property is passed to the forge build itself. Full end-to-end indy coverage for forge-generated apps would require additional work to inject `grails { indy = true }` into the Rocker template or TestKit configuration.

Relates to #15321, follow-up to #15375

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -136,11 +136,12 @@ jobs:
           -PonlyCoreTests
           -PskipCodeStyle
   buildForge:
-    name: "Build Grails Forge"
+    name: "Build Grails Forge (Java ${{ matrix.java }}, indy=${{ matrix.indy }})"
     strategy:
       fail-fast: false
       matrix:
         java: [ 17, 21 ]
+        indy: [ false, true ]
     runs-on: ubuntu-24.04
     steps:
       - name: "Output Agent IP" # in the event RAO blocks this agent, this can be used to debug it
@@ -163,6 +164,7 @@ jobs:
           ./gradlew build
           --continue
           --stacktrace
+          -PgrailsIndy=${{ matrix.indy }}
           -PskipCodeStyle
           -PskipTests
       - name: "🔨 Build project with tests"
@@ -173,6 +175,7 @@ jobs:
           --continue
           --rerun-tasks
           --stacktrace
+          -PgrailsIndy=${{ matrix.indy }}
           -PskipCodeStyle
       - name: "✅ Verify combined CLI"
         run: |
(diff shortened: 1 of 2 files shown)


jdaugherty (approved):
I like this and I think this makes sense, but will this impact our run times?  @matrei what are your thoughts?

jamesfredley:
We might be able to trim the number of jobs, by only running indy on a subset.

jamesfredley:
@jdaugherty It puts more load on github, but the large set of jobs took 41m 36s and the longest running task was grails-core on windows 41m 30s 

Additionally the longest running overall was still the joint validation build at 51m 51s, we could save a few minutes there by using the published Groovy snapshot instead of building it, which should be equivalent.

jdaugherty (approved):

jdaugherty:
Given that this is just matrix'ing the jobs, I'm going to proceed with merging this change.  @matrei  let me know if you disagree and we can revisit.

PR #15418, opened 2026-02-19
Fix @Where and DetachedCriteria query methods ignoring @Transactional(connection)

### LINKED ISSUE #15416: Data Service query methods ignore class-level @Transactional(connection) - @Where and DetachedCriteria operations route to wrong datasource
## Summary

Data Service query methods that use `DetachedCriteria` internally ignore the class-level `@Transactional(connection = '...')` annotation and silently route queries to the default datasource instead of the specified connection. This affects `@Where`-annotated methods and all methods implemented via `AbstractDetachedCriteriaServiceImplementor` (including `count()`, `list()`, `findAll*()`, `findOneBy*()`, etc.).

The `save()` method is NOT affected because `AbstractSaveImplementer` correctly uses `findConnectionId(abstractMethodNode)`.

## Root Cause

Two implementers call `findConnectionId(newMethodNode)` instead of `findConnectionId(abstractMethodNode)`:

1. **`AbstractWhereImplementer.groovy` line 99:**
   ```groovy
   Expression connectionId = findConnectionId(newMethodNode)  // BUG
   ```

2. **`AbstractDetachedCriteriaServiceImplementor.groovy` line 78:**
   ```groovy
   Expression connectionId = findConnectionId(newMethodNode)  // BUG
   ```

`findConnectionId` resolves the `@Transactional` annotation via `TransactionalTransform.findTransactionalAnnotation(methodNode)`, which falls back to `methodNode.getDeclaringClass()`. The `newMethodNode` (the generated implementation method) declares on the generated `$ServiceImplementation` class, which does NOT carry the `@Transactional` annotation. The `abstractMethodNode` (the original interface/abstract class method) declares on the class that has the annotation.

This causes `findConnectionId` to return null, so the `withConnection(connectionId)` call on the DetachedCriteria is never added, and queries run against the default datasource.

## Affected Code

- `grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractWhereImplementer.groovy` line 99
- `grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractDetachedCriteriaServiceImplementor.groovy` line 78

## Affected Data Service Methods

Any Data Service method that goes through `AbstractDetachedCriteriaServiceImplementor` or `AbstractWhereImplementer`:
- `@Where`-annotated methods
- `count()` / `countBy*()`
- `list()` / `find()` / `findAll*()`
- `findBy*()` / `findOneBy*()`
- Any method returning domain class iterables

**NOT affected:** `save()` (uses `AbstractSaveImplementer` which correctly uses `abstractMethodNode`)

## Example

```groovy
@Service(Metric)
@Transactional(connection = 'secondary')
interface MetricDataService {

    Metric save(String name, Double amount)  // WORKS - saves to secondary

    @Where({ amount >= minAmount })
    List<Metric> findByMinAmount(Double minAmount)  // BUG - queries default
}
```

`findByMinAmount()` will query the default datasource instead of `secondary`.

## Reproducer

https://github.com/jamesfredley/grails-15416-where-connection-routing

```bash
git clone https://github.com/jamesfredley/grails-15416-where-connection-routing.git
cd grails-15416-where-connection-routing
./gradlew integrationTest
```

The test passes by asserting the **buggy** behavior: the `@Where` query returns data from the default datasource instead of secondary. See inline comments in the test for what correct behavior should be.

## Fix

Change `findConnectionId(newMethodNode)` to `findConnectionId(abstractMethodNode)` in both files:

**AbstractWhereImplementer.groovy line 99:**
```groovy
// Before (bug):
Expression connectionId = findConnectionId(newMethodNode)
// After (fix):
Expression connectionId = findConnectionId(abstractMethodNode)
```

**AbstractDetachedCriteriaServiceImplementor.groovy line 78:**
```groovy
// Before (bug):
Expression connectionId = findConnectionId(newMethodNode)
// After (fix):
Expression connectionId = findConnectionId(abstractMethodNode)
```

Both `abstractMethodNode` and `newMethodNode` are available in the `doImplement` method signature.

## Context

Discovered while fixing a similar bug pattern in PR #15395. That PR introduced `findConnectionId` calls in several other implementers (SaveImplementer, DeleteImplementer, FindAndDeleteImplementer) which correctly use `abstractMethodNode`. The two implementers listed above pre-existed on the `7.0.x` branch with the incorrect `newMethodNode` usage.

jamesfredley:
## Summary

Fixes #15416

Data Service methods using `@Where` annotations or `DetachedCriteria`-based queries (`count()`, `list()`, `findBy*()`) were ignoring the class-level `@Transactional(connection)` annotation, causing queries to execute against the default datasource instead of the specified connection.

Reproducer: https://github.com/jamesfredley/grails-15416-where-connection-routing

## Root Cause

Two issues were found:

### 1. Wrong method node passed to `findConnectionId()`

Both `AbstractWhereImplementer` and `AbstractDetachedCriteriaServiceImplementor` called `findConnectionId(newMethodNode)` instead of `findConnectionId(abstractMethodNode)`.

- `newMethodNode` belongs to the generated `$ServiceImplementation` class, which does not carry the `@Transactional` annotation
- `abstractMethodNode` belongs to the original interface/abstract class where `@Transactional(connection)` is declared
- `findConnectionId()` resolves the connection by walking up to `methodNode.getDeclaringClass()` to find `@Transactional` - so passing the wrong node meant it could never find the annotation

This is the same class of bug fixed in #15395 for `SaveImplementer`, `DeleteImplementer`, and `FindAndDeleteImplementer`.

### 2. `build()` after `withConnection()` in `AbstractWhereImplementer`

In the `@Where` code path, the original order was:
1. `query = new DetachedCriteria(Foo)`
2. `query = query.withConnection('secondary')` - sets `connectionName`
3. `query = query.build(closure)` - internally calls `clone()`, which does **not** copy `connectionName`

The `clone()` call inside `build()` creates a new `DetachedCriteria` instance without the connection setting, so the connection was silently lost.

Fixed by reordering to: `new DetachedCriteria` -> `build(closure)` -> `withConnection(name)`.

> **Note**: `AbstractDetachedCriteria.clone()` not copying `connectionName` is arguably a separate bug. The non-`@Where` path in `AbstractDetachedCriteriaServiceImplementor` is not affected because it does not call `build()` after `withConnection()`.

## Changes

| File | Change |
|------|--------|
| `AbstractWhereImplementer.groovy` | `findConnectionId(newMethodNode)` -> `findConnectionId(abstractMethodNode)` + reorder `build()`/`withConnection()` |
| `AbstractDetachedCriteriaServiceImplementor.groovy` | `findConnectionId(newMethodNode)` -> `findConnectionId(abstractMethodNode)` |

### Unit Tests (`WhereConnectionRoutingSpec`) - 5 tests
- Verifies `@Where` method generates `withConnection()` call when `@Transactional(connection)` is present
- Verifies `@Where` method does NOT generate `withConnection()` when no connection specified
- Verifies `DetachedCriteria` `count()` generates `withConnection()` call
- Verifies `DetachedCriteria` `list()` generates `withConnection()` call
- Verifies `DetachedCriteria` `findByName()` generates `withConnection()` call

### Integration Tests (`WhereQueryMultiDataSourceSpec`) - 5 tests
- Uses two real H2 in-memory databases (default + secondary)
- Domain class mapped with `datasource 'ALL'`
- Inserts data only into secondary, verifies queries route correctly:
  - `@Where`-annotated method returns data from secondary
  - `@Where`-annotated method returns empty from default (proving isolation)
  - `count()` routes to secondary
  - `list()` routes to secondary
  - `findByName()` routes to secondary

All existing `ServiceTransformSpec` tests (30) continue to pass with no regressions.

### DIFF
--- a/grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
+++ b/grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
@@ -28,7 +28,6 @@ import grails.gorm.annotation.Entity
 import grails.gorm.services.Service
 import grails.gorm.services.Where
 import grails.gorm.transactions.Transactional
-import org.grails.datastore.gorm.GormEnhancer
 import org.grails.datastore.gorm.GormEntity
 import org.grails.datastore.mapping.core.DatastoreUtils
 import org.grails.orm.hibernate.HibernateDatastore
@@ -60,13 +59,11 @@ class WhereQueryMultiDataSourceSpec extends Specification {
     }
 
     void cleanup() {
-        def secondaryApi = GormEnhancer.findStaticApi(Item, 'secondary')
-        secondaryApi.withNewTransaction {
-            secondaryApi.executeUpdate('delete from Item')
+        Item.secondary.withNewTransaction {
+            Item.secondary.deleteAll(Item.secondary.list())
         }
-        def defaultApi = GormEnhancer.findStaticApi(Item)
-        defaultApi.withNewTransaction {
-            defaultApi.executeUpdate('delete from Item')
+        Item.withNewTransaction {
+            Item.deleteAll(Item.list())
         }
     }
 
@@ -138,20 +135,14 @@ class WhereQueryMultiDataSourceSpec extends Specification {
     }
 
     private void saveToSecondary(String name, Double amount) {
-        def api = GormEnhancer.findStaticApi(Item, 'secondary')
-        api.withNewTransaction {
-            def instanceApi = GormEnhancer.findInstanceApi(Item, 'secondary')
-            def item = new Item(name: name, amount: amount)
-            instanceApi.save(item, [flush: true])
+        Item.secondary.withNewTransaction {
+            new Item(name: name, amount: amount).secondary.save(flush: true)
         }
     }
 
     private void saveToDefault(String name, Double amount) {
-        def api = GormEnhancer.findStaticApi(Item)
-        api.withNewTransaction {
-            def instanceApi = GormEnhancer.findInstanceApi(Item)
-            def item = new Item(name: name, amount: amount)
-            instanceApi.save(item, [flush: true])
+        Item.withNewTransaction {
+            new Item(name: name, amount: amount).save(flush: true)
         }
     }
 }
--- a/grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
+++ b/grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
@@ -1,0 +1,187 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.orm.hibernate.connections
+
+import org.hibernate.dialect.H2Dialect
+import spock.lang.AutoCleanup
+import spock.lang.Issue
+import spock.lang.Shared
+import spock.lang.Specification
+
+import grails.gorm.annotation.Entity
+import grails.gorm.services.Service
+import grails.gorm.services.Where
+import grails.gorm.transactions.Transactional
+import org.grails.datastore.gorm.GormEnhancer
+import org.grails.datastore.gorm.GormEntity
+import org.grails.datastore.mapping.core.DatastoreUtils
+import org.grails.orm.hibernate.HibernateDatastore
+
+@Issue("https://github.com/apache/grails-core/issues/15416")
+class WhereQueryMultiDataSourceSpec extends Specification {
+
+    @Shared Map config = [
+            'dataSource.url':"jdbc:h2:mem:defaultDB;LOCK_TIMEOUT=10000",
+            'dataSource.dbCreate': 'create-drop',
+            'dataSource.dialect': H2Dialect.name,
+            'dataSource.formatSql': 'true',
+            'hibernate.flush.mode': 'COMMIT',
+            'hibernate.cache.queries': 'true',
+            'hibernate.hbm2ddl.auto': 'create-drop',
+            'dataSources.secondary':[url:"jdbc:h2:mem:secondaryDB;LOCK_TIMEOUT=10000"],
+    ]
+
+    @Shared @AutoCleanup HibernateDatastore datastore = new HibernateDatastore(
+            DatastoreUtils.createPropertyResolver(config), Item
+    )
+
+    @Shared ItemQueryService itemQueryService
+
+    void setupSpec() {
+        itemQueryService = datastore
+                .getDatastoreForConnection('secondary')
+                .getService(ItemQueryService)
+    }
+
+    void setup() {
+        def api = GormEnhancer.findStaticApi(Item, 'secondary')
+        api.withNewTransaction {
+            api.executeUpdate('delete from Item')
+        }
+        GormEnhancer.findStaticApi(Item).withNewTransaction {
+            GormEnhancer.findStaticApi(Item).executeUpdate('delete from Item')
+        }
+    }
+
+    void "@Where query routes to secondary datasource"() {
+        given:
+        saveToSecondary('Cheap', 10.0)
+        saveToSecondary('Expensive', 500.0)
+
+        when:
+        def results = itemQueryService.findByMinAmount(100.0)
+
+        then:
+        results.size() == 1
+        results[0].name == 'Expensive'
+    }
+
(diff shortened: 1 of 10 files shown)


grails-datamapping-core/src/test/groovy/grails/gorm/services/WhereConnectionRoutingSpec.groovy
```diff
@@ -0,0 +1,207 @@
...
+        Class impl = service.classLoader.loadClass("\$FooServiceImplementation")
+
+        then:"The impl is valid"
+        impl.getAnnotation(Transactional) != null
+        impl.getAnnotation(Transactional).connection() == 'secondary'
+        impl.getMethod("list").getAnnotation(Implemented).by() == FindAllImplementer
+        impl.getMethod("findAllByTitle", String).getAnnotation(Implemented).by() == FindAllByImplementer
+        impl.getMethod("find", Serializable).getAnnotation(Implemented).by() == FindOneImplementer
+    }
+
```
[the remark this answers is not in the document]
jamesfredley: Added `findByName(String)` to the unit test's service interface and asserted it maps to `FindOneByImplementer`. Both unit and integration tests now cover `findByName` routing.

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
```diff
@@ -0,0 +1,187 @@
...
+    void setupSpec() {
+        itemQueryService = datastore
+                .getDatastoreForConnection('secondary')
+                .getService(ItemQueryService)
+    }
+
+    void setup() {
+        def api = GormEnhancer.findStaticApi(Item, 'secondary')
+        api.withNewTransaction {
+            api.executeUpdate('delete from Item')
```
jdaugherty: Is this from test pollution? Why isn't this in cleanup?

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractWhereImplementer.groovy
```diff
@@ -96,16 +96,16 @@ abstract class AbstractWhereImplementer extends AbstractReadOperationImplementer
             body.addStatement(
                     declS(queryVar, ctorX(getDetachedCriteriaType(domainClassNode), args(classX(domainClassNode.plainNodeReference))))
             )
-            Expression connectionId = findConnectionId(newMethodNode)
+            body.addStatement(
+                    assignS(queryVar, callX(queryVar, 'build', closureExpression))
```
jdaugherty, line 100: These changes are in core, but there are additional tests in hibernate.  That strongly tells me the test should be part of the TCK.

jdaugherty (commented):
The change is great, but what do you think about adding a test to the tck?

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractWhereImplementer.groovy
jamesfredley: This is going to be complex, since TCK does not currently support multiple data sources, but I will present a potential solution, since I think it should.

jdaugherty: If we want to ticket it for the future I'm ok with that too.

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
jamesfredley: Fixed - moved the data deletion from `setup()` to `cleanup()` so each test cleans up after itself regardless of pass/fail. See commit 98e9b51.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractWhereImplementer.groovy
jamesfredley: Done - added a full TCK test (`WhereQueryConnectionRoutingSpec`) with domain class, service interface, and multi-datasource support in both the Hibernate and MongoDB TCK managers. The test validates @Where, count, list, and findByName all route to the correct secondary datasource.

Verified passing on:
- Hibernate5: 5/5 pass
- MongoDB: 5/5 pass
- SimpleMap: skipped (no multi-datasource support)

See commit 98e9b51.

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
```diff
@@ -0,0 +1,188 @@
...
+        def found = itemQueryService.findByName('Unique')
+
+        then:
+        found != null
+        found.name == 'Unique'
+        found.amount == 77.0
+    }
+
+    private void saveToSecondary(String name, Double amount) {
+        def api = GormEnhancer.findStaticApi(Item, 'secondary')
```
jdaugherty: If we have the domain object, why aren't we testing the public apis?  It looks like you're calling the static, internal implementation instead.

jdaugherty (approved):

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/connections/WhereQueryMultiDataSourceSpec.groovy
jamesfredley: Good call - replaced all `GormEnhancer.findStaticApi`/`findInstanceApi` usage with the public domain class API (`Item.secondary.withNewTransaction`, `item.secondary.save()`, `Item.secondary.deleteAll`) in 5a50372074. Matches the pattern used in `MultipleDataSourcesWithEventsSpec`.

jamesfredley: This came from an issue preventing Data Services from using @CompileStatic with injected services.  But not the case here.

PR #15419, opened 2026-02-20
Fix exists() cross-join caused by duplicate CriteriaQuery root

### LINKED ISSUE #14334: Domain exists method cross joining entire table with Hibernate
I've been tracing some performance issues and found an issue with the `GormEntity.exists` when using Hibernate.

In [AbstractHibernateGormStaticApi](https://github.com/apache/grails-data-mapping/blob/00fbc39170146c6ae844c66c7e2056a4e7565ba6/grails-data-hibernate5/gorm/src/main/groovy/org/grails/orm/hibernate/AbstractHibernateGormStaticApi.groovy#L217) each call to `criteriaQuery.from` creates a new query root.  The resulting HQL looks like this:

```
select count(generatedAlias0) 
from SomeDomain as generatedAlias1, 
SomeDomain as generatedAlias0 
where generatedAlias1.id=1L 
```

It looks like it just needs to have one of the calls to `criteriaQuery.from` eliminated.

jamesfredley:
## Summary

Fixes #14334

`GormEntity.exists()` was performing a cartesian product (cross-join) against the entire table due to `criteriaQuery.from()` being called twice in `AbstractHibernateGormStaticApi.exists()`.

Reproducer: https://github.com/scottlollman/grails-data-mapping-issue-2071

## Root Cause

`AbstractHibernateGormStaticApi.exists()` created two query roots:

```groovy
Root queryRoot = criteriaQuery.from(persistentEntity.javaClass)          // root 1
// ...
criteriaQuery.select(criteriaBuilder.count(criteriaQuery.from(...)))     // root 2 (BUG)
```

Each call to `criteriaQuery.from()` adds a new root to the query. Two roots on the same table produce a cross-join:

```sql
-- Before (cross-join scans entire table)
select count(generatedAlias0)
from SomeDomain as generatedAlias1, SomeDomain as generatedAlias0
where generatedAlias1.id=1L

-- After (single root, simple indexed lookup)
select count(generatedAlias0)
from SomeDomain as generatedAlias0
where generatedAlias0.id=1L
```

While the boolean result was technically correct (non-zero = true), the query performed a full table scan for every `exists()` call.

## Fix

One-line change - reuse the existing `queryRoot` variable instead of calling `criteriaQuery.from()` a second time:

```groovy
criteriaQuery.select(criteriaBuilder.count(queryRoot))
```

### Unit Tests (`ExistsCrossJoinSpec`) - 4 tests
Integration tests using a standalone `HibernateDatastore` with H2:
- `exists` returns true for existing entity
- `exists` returns false for non-existent id
- `exists` does not produce a cross-join (SQL captured via Hibernate `StatementInspector`, verified no cross-join or comma-join pattern)
- `exists` with multiple rows returns correct result

### Functional Tests (`ExistsSpec`) - 3 tests
Full Grails integration tests in `grails-test-examples/gorm` using the existing `Product` domain class:
- `exists` returns true for persisted entity
- `exists` returns false for non-existent id
- `exists` returns correct result with multiple rows in table

Existing `ReadOperationSpec` tests continue to pass.

### DIFF
--- a/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/AbstractHibernateGormStaticApi.groovy
+++ b/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/AbstractHibernateGormStaticApi.groovy
@@ -245,7 +245,7 @@ abstract class AbstractHibernateGormStaticApi<D> extends GormStaticApi<D> {
                     //TODO: Remove explicit type cast once GROOVY-9460
                     criteriaBuilder.equal((Expression<?>) idProp, id)
             )
-            criteriaQuery.select(criteriaBuilder.count(criteriaQuery.from(persistentEntity.javaClass)))
+            criteriaQuery.select(criteriaBuilder.count(queryRoot))
             Query criteria = session.createQuery(criteriaQuery)
             HibernateHqlQuery hibernateHqlQuery = new HibernateHqlQuery(
                     hibernateSession, persistentEntity, criteria)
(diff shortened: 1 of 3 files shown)


jdaugherty (approved):
I'm assuming the test failure is due to test flakyness

PR #15420, opened 2026-02-20
Adopt AntBuilder groovydoc with javaVersion support

### LINKED ISSUE #15385: Adopt newer groovydoc options
### Issue description

Per the weekly, we need to adopt groovydoc using antbuilder so we can take advantage of newer java language versions since https://github.com/gradle/gradle/issues/33659#issuecomment-3843849481 is not merged

jamesfredley:
## Summary

Replaces Gradle built-in Groovydoc task execution with direct AntBuilder invocation of the Groovy `org.codehaus.groovy.ant.Groovydoc` Ant task, enabling the `javaVersion` parameter introduced in Groovy 4.0.27 (GROOVY-11668). The AntBuilder logic is centralized into a layered plugin architecture in `build-logic/`:

- **`GroovydocEnhancerPlugin`** - Generic, publishable base plugin with all AntBuilder logic
- **`GrailsGroovydocPlugin`** - Thin Grails-specific layer that applies the base and sets the Matomo footer

Also fixes a pre-existing bug in `resolveProjectVersion()` where external groovydoc links (geb, testcontainers, spring, spring-boot) were silently omitted.

## Problem

Gradle `Groovydoc` task does not expose the `javaVersion` property (gradle/gradle#33659 is not merged). Without it, groovydoc defaults to `JAVA_11` language level when parsing Java source files, causing failures when processing Java 17+ features like sealed classes, records, and pattern matching.

### 1. AntBuilder groovydoc with javaVersion

For each Groovydoc task:
1. Clear Gradle built-in `@TaskAction` execution
2. Replace with a `doLast` that uses AntBuilder to call `org.codehaus.groovy.ant.Groovydoc` directly
3. Pass `javaVersion: JAVA_17` (derived from `gradle.properties`) to enable correct Java 17+ source parsing

#### Base plugin: `GroovydocEnhancerPlugin` (`org.apache.grails.buildsrc.groovydoc-enhancer`)

Generic, publishable plugin with no Grails-specific logic:
- Replaces Gradle's built-in Groovydoc execution with AntBuilder for `javaVersion` support
- Registers `documentation` configuration for Ant classpath
- Resolves source directories from source sets or `ext.groovydocSourceDirs`
- Supports external doc links via `ext.groovydocLinks`
- Handles `GroovydocAccess` enum via `Class.forName` (not on build-logic compile classpath)
- Handles Gradle's mixed `Property<T>` / plain-value API via `resolveGroovydocProperty()`

`GroovydocEnhancerExtension` provides:
- `javaVersion` (default: `JAVA_17`) - target Java version for Groovydoc output
- `javaVersionEnabled` (default: `true`) - allows disabling for Groovy 3.x (Forge uses 3.0.25)
- `useAntBuilder` (default: `true`) - switch back to Gradle's built-in task when gradle/gradle#33659 merges
- `footer` - customizable HTML footer for generated docs

#### Grails plugin: `GrailsGroovydocPlugin` (`org.apache.grails.buildsrc.groovydoc`)

Thin Grails-specific layer:
- Applies `GroovydocEnhancerPlugin`
- Sets `footer` to the Matomo analytics snippet

#### Switch-back path

When Gradle merges their `javaVersion` PR, set `groovydocEnhancer { useAntBuilder = false }` to skip the AntBuilder action replacement and use Gradle's built-in task instead.

### 3. Bug fix: resolveProjectVersion()

Fixed a pre-existing bug where `resolveProjectVersion()` in `docs-dependencies.gradle` never returned the resolved version string. The function had an early return for null but was missing the return of the actual version value, causing all external groovydoc links to be silently omitted.

### Files changed

| File | Scope | Notes |
|------|-------|-------|
| `build-logic/plugins/build.gradle` | Plugin registration | Registers both `groovydocEnhancer` and `grailsGroovydoc` plugin IDs |
| `build-logic/.../GroovydocEnhancerPlugin.groovy` | Base plugin | Generic AntBuilder execution, source resolution, links |
| `build-logic/.../GroovydocEnhancerExtension.groovy` | Base extension | `javaVersion`, `javaVersionEnabled`, `useAntBuilder`, `footer` |
| `build-logic/.../GrailsGroovydocPlugin.groovy` | Grails plugin | Applies base plugin, sets Matomo footer |
| `gradle/docs-dependencies.gradle` | Central config (~90 modules) | Plugin apply + deps + dynamic links + resolveProjectVersion fix |
| `gradle/docs-config.gradle` | Per-module setup | Plugin apply + `includeInApiDocs` |
| `grails-gradle/gradle/docs-config.gradle` | grails-gradle config | Plugin apply + deps + custom dest dir |
| `grails-forge/gradle/doc-config.gradle` | Forge (Groovy 3.0.25) | Plugin apply (`javaVersionEnabled = false`) + deps |
| `grails-data-hibernate5/docs/build.gradle` | Hibernate5 docs | Plugin apply + deps + custom source collection |
| `grails-data-mongodb/docs/build.gradle` | MongoDB docs | Plugin apply + deps + custom source collection |
| `grails-doc/build.gradle` | Aggregate groovydoc | Simplified - footer from plugin |
| `grails-data-docs/stage/build.gradle` | Data mapping aggregate | Simplified - footer from plugin |

### Technical details

- A `resolveGroovydocProperty()` helper handles the mix of plain values and `Property<T>` wrappers in Gradle `Groovydoc` task API
- Source directories are passed via `ext.groovydocSourceDirs` for aggregate tasks, or derived from source sets for per-module tasks
- The `documentation` configuration already includes `groovy-ant` in the central config; hibernate5 and mongodb docs needed it added
- Disabled groovydoc tasks (test suites) are unaffected - Gradle checks `enabled` before running any actions
- Dynamic links (geb-spock, testcontainers, spring-core, spring-boot) use `doFirst` to populate `ext.groovydocLinks`, plugin reads them in `doLast`

## Testing

- [x] Per-module groovydoc: `:grails-core:groovydoc`, `:grails-bootstrap:groovydoc`
- [x] grails-gradle module: `:grails-gradle-plugins:groovydoc`
- [x] Aggregate groovydoc: `:grails-doc:aggregateGroovydoc` (3,929 HTML files generated)
- [x] Data mapping aggregate: `:grails-data-docs-stage:aggregateDataMappingGroovydoc`
- [x] Code style check: `./gradlew codeStyle` passes
- [x] Forge groovydoc (with `javaVersionEnabled = false`)
- [x] All 7 groovydoc builds verified after plugin split
- [x] Matomo confirmed in all 6,295+ HTML files across all builds

Closes #15385


build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsGroovydocPlugin.groovy
```diff
@@ -0,0 +1,186 @@
...
+import org.gradle.api.attributes.Bundling
+import org.gradle.api.attributes.Category
+import org.gradle.api.attributes.Usage
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.SourceSet
+import org.gradle.api.tasks.SourceSetContainer
+import org.gradle.api.tasks.javadoc.Groovydoc
+
+@CompileStatic
+class GrailsGroovydocPlugin implements Plugin<Project> {
```
jdaugherty, line 27: What are your thoughts about keeping this plugin generic?  i.e. do not put specific grails-core configuration in it and instead configure it like we configured groovydoc before?  That way if gradle merges the upstream change, we don't have to separate out all of the configuration.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsGroovydocPlugin.groovy
```diff
@@ -0,0 +1,186 @@
...
+                )
+
+                List<Map<String, String>> links = resolveLinks(gdoc)
+                String sourcepath = sourceDirs.collect { it.absolutePath }.join(File.pathSeparator)
+
+                Map<String, Object> antArgs = [
+                        destdir: destDir.absolutePath,
+                        sourcepath: sourcepath,
+                        packagenames: '**.*',
+                        windowtitle: gdoc.windowTitle ?: '',
```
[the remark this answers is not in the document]
jamesfredley: This is not correct. The Ant `groovydoc` task does not have a separate `classpath` attribute - the `taskdef` classloader (loaded with the `documentation` configuration) IS the type resolution classpath. Groovydoc parses sources syntactically rather than compiling them, so the `taskdef` classpath provides everything needed for type resolution.

The `documentation` configuration includes `groovy-all`, `groovy-ant`, `groovy-templates`, and any project-specific runtime dependencies. This is verified working - all 6,295+ HTML files generate correctly across all 7 groovydoc builds.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsGroovydocPlugin.groovy
```diff
@@ -0,0 +1,186 @@
...
+            gdoc.actions.clear()
+            gdoc.doLast {
+                File destDir = gdoc.destinationDir
+                destDir.mkdirs()
+
+                List<File> sourceDirs = resolveSourceDirectories(gdoc, project)
+                if (sourceDirs.isEmpty()) {
+                    project.logger.lifecycle("Skipping groovydoc for ${gdoc.name}: no source directories found")
+                    return
+                }
```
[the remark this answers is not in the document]
jamesfredley: Acknowledged, though this is not a practical issue. The Ant groovydoc task uses `sourcepath` + `packagenames` by design - it does not support nested `fileset` elements.

The excludes configured on the Gradle task (`META-INF/**`, `*.yml`, `*.properties`, `*.xml`, `Application.groovy`, `Bootstrap.groovy`, `resources.groovy`) are either non-code files that groovydoc naturally ignores (it only processes `.groovy` and `.java` files) or convention files that do not exist in framework source directories. The generated output is identical to what the previous per-module AntBuilder configurations produced.

gradle/docs-dependencies.gradle
```diff
@@ -52,52 +44,30 @@ String resolveProjectVersion(String artifact) {
...
 
-    doFirst {
+    gdoc.doFirst {
+        List<Map<String, String>> links = []
         def gebVersion = resolveProjectVersion('geb-spock')
-        if(gebVersion) {
-            gdoc.link("https://groovy.apache.org/geb/manual/${gebVersion}/api/", 'geb.')
+        if (gebVersion) {
+            links << [packages: 'geb.', href: "https://groovy.apache.org/geb/manual/${gebVersion}/api/"]
         }
```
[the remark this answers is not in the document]
jamesfredley: Good catch - this is correct. This was a pre-existing bug on 7.0.x (not introduced by this PR). The function had an early `return null` for the not-found case but was missing the implicit return of `version` as the last expression. In Groovy, the `if` statement evaluates to `null` when the condition is false and there is no `else`, so the function always returned null.

Fixed in bf365318bb by adding `version` as the last expression in the function. External groovydoc links (geb, testcontainers, spring, spring-boot) should now be included correctly.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsGroovydocPlugin.groovy
jamesfredley: Addressed in bf365318bb. The plugin is now split into two layers:

1. **`GroovydocEnhancerPlugin`** (`org.apache.grails.buildsrc.groovydoc-enhancer`) - Generic base plugin with all the AntBuilder logic, source resolution, link support, and `documentation` configuration. No Grails-specific code. Publishable and reusable by anyone.

2. **`GrailsGroovydocPlugin`** (`org.apache.grails.buildsrc.groovydoc`) - Thin layer that applies the base plugin and sets `footer` to the Matomo analytics snippet. This is the only Grails-specific piece.

The base extension also includes a `useAntBuilder` flag (default `true`). When Gradle merges their `javaVersion` support (gradle/gradle#33659), setting `groovydocEnhancer { useAntBuilder = false }` will skip the AntBuilder action replacement and let Gradle's built-in task run - no code changes needed to switch back.

matrei (approved):
Excellent!

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerExtension.groovy
```diff
@@ -0,0 +1,98 @@
...
+     * analytics scripts, copyright notices, or custom branding.
+     *
+     * <p>Defaults to an empty string (no footer).</p>
+     */
+    final Property<String> footer
+
+    @Inject
+    GroovydocEnhancerExtension(ObjectFactory objects, Project project) {
+        javaVersion = objects.property(String).convention(
+                "JAVA_${GradleUtils.findProperty(project, 'javaVersion') ?: '17'}" as String
```
jdaugherty: javaVersion should always be set or it should error?

jdaugherty (commented):
Have you checked the build scans to ensure no new warnings have appeared with adding this?

jamesfredley:
@jdaugherty Deprecations look the same

Before: https://develocity.apache.org/s/ocm5dcxmesrps/deprecations (this is run later, but was off 7.0.x)
After: https://develocity.apache.org/s/ovm3zrom23m7o/deprecations

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerExtension.groovy
jamesfredley: I believe it will be null, unless you set it in the end project and we are defaulting to 17 here given that is the base for Grails 7.

jdaugherty: Use a provider and it will be set correctly.  project.provider { /* your existing JAVA_ code */ }

jamesfredley: Done in 3702656 - wrapped the convention in `project.provider { ... }` so `javaVersion` resolves lazily at execution time.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerExtension.groovy
```diff
@@ -89,7 +89,7 @@ class GroovydocEnhancerExtension {
     @Inject
     GroovydocEnhancerExtension(ObjectFactory objects, Project project) {
         javaVersion = objects.property(String).convention(
-                "JAVA_${GradleUtils.findProperty(project, 'javaVersion') ?: '17'}" as String
+                project.provider { "JAVA_${GradleUtils.findProperty(project, 'javaVersion') ?: '17'}" as String }
```
jdaugherty: You shouldn't need the default of 17 now.

jamesfredley: The end grails project will not have javaVersion set in the project, unless they are attempting to overriode and change the version lower or higher for groovydoc support.  17 is here as the automatic default.

jdaugherty: The end grails app?  This is for our applications and we define `javaVersion` in the grade.properties - which is copied to every project in this build.  This comment doesn't make sense to me. If it's not copied, we have a bug that's why we should leave it with no default.

jamesfredley: grails-core does, but if we were to publish this and another project used it ... was what I was thinking about.  IE the end Grails App.

jdaugherty: There are so many other changes that have to be made to publish this.  Lets do whats best for grails core and address that when/if we decide to publish?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy
```diff
@@ -0,0 +1,182 @@
...
+    @CompileDynamic
+    private static void configureAntBuilderExecution(Project project, GroovydocEnhancerExtension extension) {
+        project.tasks.withType(Groovydoc).configureEach { gdoc ->
+            if (!extension.useAntBuilder.get()) {
+                return
+            }
+
+            gdoc.actions.clear()
+            gdoc.doLast {
+                def destDir = gdoc.destinationDir.tap { it.mkdirs() }
```
sanjana2505006, line 90: Small suggestion here, since `destinationDir` is deprecated in Gradle 8 and getting removed in Gradle 9, should we maybe use `destinationDirectory.get().asFile` instead? Just thought it might help avoid some deprecation warnings down the line! Let me know if that makes sense.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy
```diff
@@ -0,0 +1,182 @@
...
+                            gdoc.name
+                    )
+                    return
+                }
+
+                project.ant.taskdef(
+                        name: 'groovydoc',
+                        classname: 'org.codehaus.groovy.ant.Groovydoc',
+                        classpath: docConfig.asPath
+                )
```
sanjana2505006: Just noticed this while reading through, could we use `gdoc.groovyClasspath` here instead of looking up the configuration manually? 
Since a task's `groovyClasspath` is already tracked as an input by Gradle, using it directly might be a little safer for the build cache. 
Maybe something like this?
```groovy
def classpath = gdoc.groovyClasspath
if (!classpath || classpath.empty) {
    project.logger.warn('Skipping groovydoc for {}: groovyClasspath is empty', gdoc.name)
    return
}

project.ant.taskdef(
        name: 'groovydoc',
        classname: 'org.codehaus.groovy.ant.Groovydoc',
        classpath: classpath.asPath
)

File: build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy (hunk 7 above)
Comment (jamesfredley) [reply build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy:90]: Thanks for the suggestion! However, the `destinationDir` deprecation in Gradle 8/9 applies to `AbstractCompile` subclasses (`JavaCompile`, `GroovyCompile`, etc.), not to `Groovydoc`. The `Groovydoc` task extends `SourceTask` and its `destinationDir` is a plain `File` property with no deprecation - even in [Gradle 9.3.1](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.javadoc.Groovydoc.html). There is no `destinationDirectory` (DirectoryProperty) equivalent on `Groovydoc`.

File: build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy (hunk 8 above)
Comment (jamesfredley) [reply build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy]: Good catch! Done in 1137d89 - now uses `gdoc.groovyClasspath` directly instead of the redundant `project.configurations.findByName('documentation')` lookup.

File: build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy [hunk 9]
```diff
@@ -0,0 +1,182 @@
...
+            if (!extension.useAntBuilder.get()) {
+                return
+            }
+
+            gdoc.actions.clear()
+            gdoc.doLast {
+                def destDir = gdoc.destinationDir.tap { it.mkdirs() }
+                def sourceDirs = resolveSourceDirectories(gdoc, project)
+                if (sourceDirs.isEmpty()) {
+                    project.logger.lifecycle(
```
jdaugherty: I don't think this should ever happen in our project, we should exception instead.

sanjana2505006: If this is meant to be a generic plugin down the road, wouldn't an exception break the build if someone applies it to a module that only has resources? Just wondering if returning early is safer since it mimics how Gradle usually skips empty source tasks.

jdaugherty: I think addressing the needs of the generic plugin and our needs are very different.  The immediate goal was to address ours and as part of any refactoring / extraction, we would then address the public needs.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy
sanjana2505006: Great, glad I could help!

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy
sanjana2505006: Ah, that is great to know, thank you for the explanation, that makes perfect sense.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerPlugin.groovy
sanjana2505006: That’s completely fair. Makes sense to keep it focused on `grails-core` for now and handle the extraction when the time comes. Thank you for clarifying!

jamesfredley: Updated in two locations.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovydocEnhancerExtension.groovy
jdaugherty: @jamesfredley this is my only last concern - again to force failures b/c we don't want core to be misconfigured.  Can we remove this default & merge this PR?  We can then re-address any extraction separately.

jamesfredley:
Addressed @jdaugherty's feedback in db5a52b - removed the `?: '17'` default from `GroovydocEnhancerExtension`. Now throws `IllegalStateException` with a clear message if `javaVersion` is missing from `gradle.properties`.

PR #15430, opened 2026-02-21
feat: publish Groovydoc Enhancer Plugin for use in end-user Grails apps

jamesfredley:
## Summary

Potential path to allow end-user Grails applications to use modern Java features (17+) and still generate accurate Groovydoc documentation.

Gradle's built-in Groovydoc task does not expose the `javaVersion` parameter (added in Groovy 4.0.27 via [GROOVY-11668](https://issues.apache.org/jira/browse/GROOVY-11668)), so projects using Java 17+ features (sealed classes, records, pattern matching, text blocks, etc.) fail to generate Groovydoc. This PR publishes the `GroovydocEnhancerPlugin` (from #15420) as a standalone Gradle plugin that end-user Grails applications can consume.

### New Published Module: `grails-gradle-groovydoc`
- **Maven coordinate**: `org.apache.grails.gradle:grails-gradle-groovydoc`
- **Plugin ID**: `org.apache.grails.gradle.groovydoc`
- Located in `grails-gradle/groovydoc/` following the existing grails-gradle subproject pattern
- Automatically included in `grails-gradle-bom` (subproject auto-inclusion)
- Manually added to `grails-bom` via `gradleBuildProjects` map for end-user app BOM resolution

### Forge Integration (DefaultFeature)
- New `GroovydocEnhancer` feature class applied to **all** generated Grails applications by default
- Adds `grails-gradle-groovydoc` as a buildscript classpath dependency (version managed by `grails-bom`)
- Applies `org.apache.grails.gradle.groovydoc` plugin via `apply plugin`

### Profile Integration
- Updated `base/profile.yml` to include the plugin for `grails` CLI-based app generation
- Both forge and profile paths now generate apps with Groovydoc enhancer support

### Build-Logic
- `build-logic/` retains its own copy of the plugin for the framework build itself (bootstrap constraint - build-logic cannot depend on grails-gradle artifacts without circular dependency)
- The published `grails-gradle-groovydoc` module is the canonical version for end-user consumption

## Architecture Note

There are two copies of the enhancer plugin:
1. **`build-logic/plugins/`** - convention plugin for building grails-core itself (unpublished, internal)
2. **`grails-gradle/groovydoc/`** - published plugin for end-user Grails apps

This duplication exists because `build-logic` is evaluated before `grails-gradle` builds, creating a bootstrap dependency that prevents `build-logic` from consuming the published artifact directly. Future work could explore composite build substitution or a shared source set to eliminate this duplication.

## Testing
- `grails-gradle-groovydoc` builds successfully standalone
- `grails-bom` POM generation succeeds with the new artifact
- `grails-forge-core` compiles with the new feature class
- Code style checks pass

## Depends On
- #15420

### DIFF
--- a/build-logic/plugins/build.gradle
+++ b/build-logic/plugins/build.gradle
@@ -62,6 +62,10 @@ gradlePlugin {
             id = 'org.apache.grails.gradle.grails-code-style'
             implementationClass = 'org.apache.grails.buildsrc.GrailsCodeStylePlugin'
         }
+        register('groovydocEnhancer') {
+            id = 'org.apache.grails.buildsrc.groovydoc-enhancer'
+            implementationClass = 'org.apache.grails.buildsrc.GroovydocEnhancerPlugin'
+        }
         register('grailsGroovydoc') {
             id = 'org.apache.grails.buildsrc.groovydoc'
             implementationClass = 'org.apache.grails.buildsrc.GrailsGroovydocPlugin'
--- a/build-logic/plugins/build.gradle
+++ b/build-logic/plugins/build.gradle
@@ -62,6 +62,10 @@ gradlePlugin {
             id = 'org.apache.grails.gradle.grails-code-style'
             implementationClass = 'org.apache.grails.buildsrc.GrailsCodeStylePlugin'
         }
+        register('grailsGroovydoc') {
+            id = 'org.apache.grails.buildsrc.groovydoc'
+            implementationClass = 'org.apache.grails.buildsrc.GrailsGroovydocPlugin'
+        }
         register('grailsRepoSettings') {
             id = 'org.apache.grails.buildsrc.repo'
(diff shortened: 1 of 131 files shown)


jamesfredley:
A Third "commons" build appears to be the only way to remove the duplication.

jamesfredley:
Or moving it to a separate repo, as we have done with publishing.

jdaugherty:
In order to publish this, its going to have to go to a separate repository...

jamesfredley:
@jdaugherty I will add that to agenda for the meeting, next week.

jamesfredley:
This will either end up as a separate Apache Grails repo or possible over on Apache Groovy.

jdaugherty:
@jamesfredley Groovy has suggested they move forward with publishing their fork of this plugin. I think we should close this & push on them to maintain / publish instead.

PR #15436, opened 2026-02-22
feature - add `@DatabaseCleanup` test annoation to facilicate cleaning up after a test or tests

jdaugherty:
After reviewing several of the grails-data PRs submitted by @jamesfredley I realized that a pattern we implemented in our Grails app would be beneficial for everyone: rather than use `@Rollback`, etc we use a clean up process that truncates all tables with data.  I've extracted our internal logic into a generic spock extension & then enhanced it with the assistance of AI to be a little more generic.  Only H2 & Postgres (AI did this one, while the H2 one was lifted from our internal implementation) are implemented for now.  

There aren't really many dependencies with this feature. I intentionally did not depend on many grails projects to ensure it can be used to test in grails-core too.  The spock extension requires `@SpringBootTest` to use the `DefaultApplicationContextResolver` to find the application context - this is added automatically by using `@Integration`.  We could relax this requirement if a custom resolver is provided instead. 

The annotation can be placed at the class level to clear all data after each test, or individually on each method - only the methods annotated will truncate the database after running. It should be rather performant since it issues truncates instead of iterating over every row (FK's etc are disabled during the cleanup process too).  There's also some optional debug that can be turned on.

I'm adding this to 7.0.x since it doesn't impact any of the existing code & we can then make use of it in the various grails-data tests to be DRY.  Simply annotate the test, and the database is always clean now at the start of every test - regardless of new transaction testing, etc.  This also means with the combination of `@Stepwise` you can now test multi-transaction processes and roll back the data reliably. 

Assisted-by: OpenCode <[opencode@opencode.ai](mailto:opencode@opencode.ai)>
Assisted-by: Claude <[claude@anthropic.com](mailto:claude@anthropic.com)>


jdaugherty:
The AI didn't do a great job on the postgres implementation.  I've cleaned it up a lot and I think this should all be working / passing now.  It should be ready for review.

jamesfredley (commented):
e

grails-testing-support-cleanup-h2/src/main/groovy/org/apache/grails/testing/cleanup/h2/H2DatabaseCleaner.groovy
```diff
@@ -0,0 +1,131 @@
...
+        if (!schemaName) {
+            log.warn('Could not resolve schema name for datasource, skipping cleanup')
+            stats.stop()
+            return stats
+        }
+
+        Sql sql = new Sql(dataSource)
+        try {
+            sql.execute('SET REFERENTIAL_INTEGRITY FALSE')
+            String query = "SELECT table_name, row_count_estimate FROM information_schema.tables WHERE table_schema = '${schemaName}' AND row_count_estimate > 0" as String
```
jamesfredley: the query filters AND row_count_estimate > 0. H2's row_count_estimate is an estimate, it is not guaranteed to be accurate. After rapid inserts and deletes (or without a recent ANALYZE), a table can have rows but still show row_count_estimate = 0. Those tables will be silently skipped, leaving test data behind.

grails-testing-support-cleanup-postgresql/src/main/groovy/org/apache/grails/testing/cleanup/postgresql/PostgresDatabaseCleanupHelper.groovy
```diff
@@ -0,0 +1,126 @@
...
+                try {
+                    connection.close()
+                }
+                catch (Exception ignored) {
+                    // ignore
+                }
+            }
+        }
+
+        throw new IllegalStateException("Because postgres defaults to the search_patch when currentSchema isn't defined, a schema should always be found")
```
jamesfredley: search_patch should be search_path

gradle/publish-root-config.gradle
```diff
@@ -76,6 +76,8 @@ def publishedProjects = [
         'grails-taglib',
         'grails-test-core',
         'grails-testing-support-core',
+        'grails-testing-support-cleanup-core',
```
jamesfredley: postgresql is missing here

grails-testing-support-cleanup-core/src/main/groovy/org/apache/grails/testing/cleanup/core/DatabaseCleanupStats.groovy
```diff
@@ -0,0 +1,200 @@
...
+        sb.toString()
+    }
+
+    /**
+     * Formats a timestamp in milliseconds as an ISO 8601 string.
+     *
+     * @param timeMillis the timestamp in milliseconds since epoch
+     * @return the formatted time string
+     */
+    private static String formatTime(long timeMillis) {
```
jamesfredley: This also exists in DatabaseCleanupInterceptor (line 168)

grails-testing-support-cleanup-h2/src/main/groovy/org/apache/grails/testing/cleanup/h2/H2DatabaseCleaner.groovy
jdaugherty: We currently have a test run that has over 25,000 tests and several of them test in performance.  I have never seen this behavior in H2.  Can you point me to the document to indicate where this is true for H2?

jamesfredley (approved):

grails-testing-support-cleanup-h2/src/main/groovy/org/apache/grails/testing/cleanup/h2/H2DatabaseCleaner.groovy
jdaugherty: FYI: https://github.com/h2database/h2database/blob/cfe04fd3de9b32fc0540284bb921b9b3ec34e0cf/h2/src/docsrc/help/information_schema.csv#L912 it looks like with H2 it's not estimated.

jamesfredley: I have never seen this directly.  Given your experience, I think we can move forward with it as it is.

> The approximate number of rows if known or some default value if unknown. For regular tables contains the total number of rows including the uncommitted rows.  
https://www.h2database.com/html/systemtables.html#information_schema_tables

matrei (approved):

PR #15441, opened 2026-02-24
refactor(tests): replace GormEnhancer.find*Api calls with public GORM APIs

jamesfredley:
## Summary

Tests were calling `GormEnhancer.findStaticApi()`, `findInstanceApi()`, and `findValidationApi()` to reach into GORM internals directly. This PR replaces all such usages with the same public APIs an end user would call.

## Changes

| File | Before | After |
|------|--------|-------|
| `WhereQueryConnectionRoutingSpec` | `GormEnhancer.findStaticApi(Domain, connection)` for tx/list | `Domain."${connectionName}"` named-datasource property + `item.save()`/`item.delete()` |
| `FooIntegrationSpec` / `BarIntegrationSpec` | `findStaticApi(Foo).datastore` | `Foo.withNewSession { it.datastore instanceof ... }` |
| `SaveWithFailOnErrorDefaultSpec` | Mutate internal `GormInstanceApi.failOnError` field | Test per-call `save(failOnError: true/false)` behaviour |
| `DeepValidateWithSaveSpec` | `findInstanceApi` + mock `GormValidateable` | Real entity + `entity.save(deepValidate: true/false)` |
| `DataServiceMultiTenantMultiDataSourceSpec` (test) | `findStaticApi` not-null assertion | Functional save/count via `metricService` |
| `DataServiceMultiTenantMultiDataSourceSpec` (`MetricService`) | `findStaticApi(Metric, 'analytics')` in service body | `Metric.executeQuery/executeUpdate` (transaction already routes to `analytics`) |
| `DataServiceMultiDataSourceSpec` | `findStaticApi(Product, 'books')` for setup/count/HQL | `productService.deleteAll()` + `@Query` methods on `ProductService` |
| `ValidationSpec` (neo4j) | `findValidationApi(cls).validator = null` | `session.mappingContext.addEntityValidator(entity, null)` |

## Verification

All affected tests pass:
- `SaveWithFailOnErrorDefaultSpec` — 4 tests PASSED
- `DeepValidateWithSaveSpec` — 2 tests PASSED
- `DataServiceMultiDataSourceSpec` — 20 tests PASSED
- `DataServiceMultiTenantMultiDataSourceSpec` — 7 tests PASSED

The `grails-data-graphql` and `grails-data-neo4j` modules are not active in the current build (not yet updated for Grails 7); their files are cleaned up but cannot be run locally.

Relates to the principle added in #15440: tests should use only public APIs as an end user would.


jamesfredley:
Addressed both Copilot review comments in commit 303415c:

**DeepValidateWithSaveSpec** (comment on line 54): Fixed. The spec now installs a `CascadingValidator` mock via `manager.session.mappingContext.addEntityValidator()` and asserts that `mockValidator.validate(entity, _, true/false)` is called with the correct `cascade` argument on each `save()` — the same public-API pattern already used in `ValidationSpec.groovy`.

**ValidationSpec** (comment on line 40): Confirmed — `AbstractMappingContext.addEntityValidator(entity, null)` is indeed a no-op (guarded by `validator != null`). However, `ValidationSpec.groovy` lives in `grails-data-neo4j`, which the module README explicitly states is not included in the Grails 7 build. This is a pre-existing issue outside the scope of this PR.

jdaugherty:
The neo4j ValidationSpec doesn't appear to have been changed?  I think that code should just be reverted.

jdaugherty:
@jamesfredley the neo4j comment still stands.

jamesfredley:
Reverted the neo4j ValidationSpec change - restored to original `GormEnhancer.findValidationApi` usage. Sorry about the delay on that.

PR #15446, opened 2026-02-24
fix(grails-gradle): change indy log level from lifecycle to info

### LINKED ISSUE #15431: Redundant Warning with Sub Projects - Grails: Groovy invokedynamic (indy) is disabled to improve performance (see issue #15293).   To enable invokedynamic: grails { indy = true } in build.gradle
### Unexpected Behavior

Introduces too much noise on startup and should only occur once.

```
> Configure project :
Grails: Groovy invokedynamic (indy) is disabled to improve performance (see issue #15293).
        To enable invokedynamic: grails { indy = true } in build.gradle

> Configure project :modules:custom-plugin
A sourcesJar task was not found, creating one.
Grails: Groovy invokedynamic (indy) is disabled to improve performance (see issue #15293).
        To enable invokedynamic: grails { indy = true } in build.gradle

> Task :bootRun
Grails application running at http://localhost:8090 in environment: development
```

### Version

7.0.7

jamesfredley:
## Summary

- Change indy disabled message from `logger.lifecycle` to `logger.info` so it doesn't print on every build

## Context

Follow-up to #15375. The `lifecycle` log level causes the indy message to appear on every Gradle invocation, which is noisy. Using `info` keeps the message available via `--info` or `--debug` without cluttering default output.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -234,8 +234,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
                 c.groovyOptions.optimizationOptions.indy = indyEnabled
             }
             if (!indyEnabled) {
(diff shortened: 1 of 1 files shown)


jamesfredley:
Resolves: https://github.com/apache/grails-core/issues/15431#issuecomment-3939971266

PR #15447, opened 2026-02-24
fix: where query composition fails when variable is re-assigned in if/else blocks

jamesfredley:
## Summary

- Fix `DetachedCriteriaTransformer` AST transform to track variables that are re-assigned (not just declared with initializer), so `where` query composition works correctly inside `if/else` blocks
- Add `junctions` to `AbstractDetachedCriteria.clone()` for completeness

## Problem

When building dynamic GORM `where` queries with conditional logic, re-assigned variables are silently ignored:

```groovy
def query
if (someCondition) {
    query = DomainClass.where { visible == true }  // This was silently ignored!
} else {
    query = DomainClass.where { visible == false }
}
query = query.where { name == 'test' }  // This chained where also broken
```

The `DetachedCriteriaTransformer` only tracked variables declared with an initializer (`def query = DomainClass.where{...}`) via `visitDeclarationExpression()`, but not variables that were re-assigned (`query = DomainClass.where{...}`) which are `BinaryExpression` nodes. When the transformer encountered `query.where{...}`, it couldn't find `query` in its `detachedCriteriaVariables` map, so the closure body was left untransformed. At runtime, the untransformed closure's property access resolved via `propertyMissing` to a non-mutating clone, and the criteria were silently dropped.

## Reproducer

Sample Grails 7 app demonstrating the bug: https://github.com/jamesfredley/grails-where-query-reassignment-bug

## Root Cause

The `DetachedCriteriaTransformer` AST transform has a gap in its variable tracking:

1. `visitDeclarationExpression()` handles `def query = DomainClass.where{...}` - registers variable in `detachedCriteriaVariables`
2. **No handler** existed for re-assignment `query = DomainClass.where{...}` - this is a `BinaryExpression`, not a `DeclarationExpression`
3. When the transformer encounters `query.where{visible == true}`:
   - `detachedCriteriaVariables.get("query")` returns `null` (never tracked)
   - Fallback: `var.getType()` returns `Object` (declared as `def query` with no initializer)
   - Neither path matches, so the closure body is NOT transformed
4. At runtime, the untransformed closure evaluates as a no-op

## Fix

1. **`DetachedCriteriaTransformer.java`**: Added `visitBinaryExpression()` override that detects assignments where the RHS is a `DomainClass.where{...}` call or `new DetachedCriteria(DomainClass)` constructor, and registers the LHS variable in `detachedCriteriaVariables` with the correct type. Also updates the variable's declared type to `DetachedCriteria<DomainClass>` so subsequent `query.where{...}` calls are properly transformed.

2. **`AbstractDetachedCriteria.groovy`**: Added `junctions` to the `clone()` method to ensure complex AND/OR query compositions are preserved during cloning.

## Tests

Added 3 new test cases to `WhereMethodSpec.groovy`:
- Re-assigned variable in `if/else` blocks (both branches verified)
- Re-assigned variable without initializer (`def query; query = Person.where{...}`)
- Triple-chained `where` on re-assigned variable

All 80 WhereMethodSpec tests pass. Code style check passes.

### DIFF
--- a/grails-datamapping-core-test/src/test/groovy/grails/gorm/tests/WhereMethodSpec.groovy
+++ b/grails-datamapping-core-test/src/test/groovy/grails/gorm/tests/WhereMethodSpec.groovy
@@ -1553,6 +1553,73 @@ class Project {
         }
     }
 
+    def "Test where query composition with re-assigned variable in if/else blocks"() {
+        given: "A bunch of people"
+        createPeople()
+
+        when: "A where query variable is assigned inside an if/else block and then chained"
+        def condition = true
+        def query
+        if (condition) {
+            query = Person.where { lastName == 'Simpson' }
+        } else {
+            query = Person.where { lastName == 'Rubble' }
+        }
+        query = query.where { age > 10 }
+        def results = query.list()
+
+        then: "Both criteria from the if block and the chained where are applied"
+        results.size() == 2
+        results.every { it.lastName == 'Simpson' && it.age > 10 }
+
+        when: "The else branch is taken instead"
+        condition = false
+        if (condition) {
+            query = Person.where { lastName == 'Simpson' }
+        } else {
+            query = Person.where { lastName == 'Rubble' }
+        }
+        query = query.where { age > 10 }
+        results = query.list()
+
+        then: "Both criteria from the else block and the chained where are applied"
+        results.size() == 1
+        results[0].lastName == 'Rubble'
+        results[0].firstName == 'Barney'
+    }
+
+    def "Test where query composition with re-assigned variable without initializer"() {
+        given: "A bunch of people"
+        createPeople()
+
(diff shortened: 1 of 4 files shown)


davydotcom:
this is huge. I have lost count how many times this bites us. Thanks for finding and fixing, glad I finally remembered to bring it up :/

davydotcom (approved):
This exactly describes a scenario we run into often. 

We normally have a workaround but when doing composable queries based on user input parameters, this reveals itself often. Thanks for finding this root cause!

jdaugherty (approved):
I assume we should add a functional test, but nice work.  Thank you.

jamesfredley:
It was added to an existing functional test on https://github.com/apache/grails-core/pull/15447/changes/4db4ed81261143b51f6bbdfab951fb6f9832ec10

PR #15456, opened 2026-02-25
Fix DetachedCriteria join behavior for get() in Hibernate

davydotcom:
## Summary
This PR now fixes two related DetachedCriteria/Hibernate issues:

1. `where { ... }.join(...).get()` did not consistently honor join handling like `.list()`
2. `join('association').property('association.field')` did not auto-create aliases for nested projection paths unless `createAlias(...)` was explicitly called

### Root cause
`DetachedCriteria` joins are applied via `DynamicFinder.applyDetachedCriteria(...)`, which can call:
- `query.join(property)` or
- `query.join(property, joinType)`

In `AbstractHibernateQuery`, only `join(String)` was overridden. `join(String, JoinType)` fell back to the base `Query` implementation, so Hibernate-specific join state used by single-result execution could diverge.

### Why `.list()` and `.get()` differed
- `.list()` path (`AbstractHibernateQuery.list()`) applies fetch strategies and executes list semantics.
- `.get()` uses `query.singleResult()` path, which relies on Hibernate query join state (`hasJoins`) when evaluating unique-result execution behavior.

### Fix
Add `AbstractHibernateQuery.join(String property, JoinType joinType)` override that:
- sets `hasJoins = true`
- records `joinTypes.put(property, joinType)`
- applies `FetchMode.JOIN` to criteria/detachedCriteria

### Symptom
`Team.where { ... }.join('club').property('club.name').list()` / `.get()` raised:
- `QueryException: could not resolve property: club.name of: Team`

Unless users manually did `createAlias('club','club')`.

### Root cause
Detached-criteria projections are adapted via `HibernateProjectionAdapter` through `ProjectionList.add(...)`, and nested association property paths were not normalized to Hibernate aliases before projection adaptation.

### Fix
In `AbstractHibernateQuery`:
- Add projection-path normalization that auto-creates aliases for nested association paths (e.g. `club.name`)
- Normalize property-based projections before passing to `HibernateProjectionAdapter`
- Keep behavior safe for non-association and embedded paths

## Tests
Added/updated `DetachedCriteriaJoinSpec` coverage:
- `check get honours join and eagerly loads association`
- `check list with join and projected association property works without explicit alias`
- `check get with join and projected association property works without explicit alias`
- `check list with association subquery plus join and projection works`

All pass with:
- `./gradlew :grails-data-hibernate5-core:test --tests grails.gorm.tests.DetachedCriteriaJoinSpec`

### DIFF
--- a/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateQuery.java
+++ b/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateQuery.java
@@ -696,6 +696,17 @@ public abstract class AbstractHibernateQuery extends Query {
     }
 
     @Override
+    public Query join(String property, JoinType joinType) {
+        this.hasJoins = true;
+        this.joinTypes.put(property, joinType);
+        if (criteria != null)
+            criteria.setFetchMode(property, FetchMode.JOIN);
+        else if (detachedCriteria != null)
+            detachedCriteria.setFetchMode(property, FetchMode.JOIN);
+        return this;
+    }
+
+    @Override
     public Query select(String property) {
         this.hasJoins = true;
         if (criteria != null)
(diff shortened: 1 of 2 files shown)


borinquenkid (approved):
It looks good to me.

jamesfredley (approved):

PR #15462, opened 2026-02-26
Fix composite ID criteria projection regression (#14516)

### LINKED ISSUE #14516: GORM 7 Regression for Criteria projection
Given these domain classes: 

```groovy
class Role {
    String authority
}
class User {
    String username
}
class UserRole implements Serializable {
    private static final long serialVersionUID = 1

    User user
    Role role

    static mapping = {
        id composite: ['user', 'role']
        version false
    }
}
```

and this query: 

```groovy
@ReadOnly
List<Long> findAllRoleIdsByUser(Long userId) {
    def c = UserRole.createCriteria()
    c.list {
        projections {
            role {
                property('id')
            }
        }
        user {
            eq('id', userId)
        }
    } as List<Long>
}
```

With Grails 3.3.11 and GORM 6.1.12.RELEASE it works.

For Grails 4.0.2 and GORM 7.0.2, it fails with: 

`Caused by: java.lang.IllegalArgumentException: Unable to locate Attribute  with the the given name [role] on this ManagedType [example.grails.UserRole]`

It seems to be related with the composite key of `UserRole` Removing that key gets rids of the error. 

### Environment Information

- **Operating System**: MacOS
- **Grails Version:** 4.0.2
- **JDK Version:** 1.8
- **Container Version (If Applicable):** TODO

### Example Application

https://github.com/grails-core-issues-forks/grails-core-issue-11505

Steps to reproduce with the example app

*Grails 3*

```
cd grailsthree;
./gradlew bootRun
curl localhost:8080/home
{"ids":[1,2]}
```
*Grails 4*

```
cd grailsfour;
./gradlew bootRun
curl localhost:8080/home
Unable to locate Attribute  with the the given name [role] on this ManagedType [example.grails.UserRole]. Stacktrace follows:
...
```

jamesfredley:
## Summary

Fixes #14516 - Criteria queries on domains with composite IDs (`id composite: ['user', 'role']`) fail when navigating composite key component associations.

## Symptom

```groovy
class UserRole implements Serializable {
    User user
    Role role
    static mapping = { id composite: ['user', 'role'] }
}

// Throws: IllegalArgumentException: Unable to locate Attribute
//         with the given name [role] on ManagedType [demo.UserRole]
UserRole.createCriteria().list {
    projections {
        role { property('id') }
    }
    user { eq('id', userId) }
}
```

This worked in GORM 6 but regressed in GORM 7.

## Root Cause

`AbstractHibernateCriteriaBuilder.invokeMethod()` uses the JPA Metamodel to determine property types:

```java
final EntityType<?> entityType = metamodel.entity(targetClass);
final Attribute<?, ?> attribute = entityType.getAttribute(name);  // THROWS for composite ID components
```

The JPA Metamodel's `EntityType.getAttribute()` does not expose composite ID components as regular attributes. In GORM 6, the older Hibernate-specific API handled this differently.

## Fix

Wrap `entityType.getAttribute(name)` in a try-catch. When it throws `IllegalArgumentException` for a composite ID component:

1. Check if the property type (`pd.getPropertyType()`) is a managed entity via `metamodel.entity()`
2. If it is, treat the property as an association (same code path as `attribute.isAssociation()`)
3. If not, re-throw the original exception

**Before:**
```java
final Attribute<?, ?> attribute = entityType.getAttribute(name);
if (attribute.isAssociation()) { ... }
```

**After:**
```java
Attribute<?, ?> attribute = null;
try {
    attribute = entityType.getAttribute(name);
} catch (IllegalArgumentException e) {
    // Fall back to checking if property type is a managed entity
    Class<?> propertyType = pd.getPropertyType();
    try {
        metamodel.entity(propertyType);
        // Treat as association - create alias, invoke closure, etc.
        ...
        return name;
    } catch (IllegalArgumentException ignored) {
        throw e;  // Not an entity, re-throw
    }
}
if (attribute.isAssociation()) { ... }
```

## Changes

| File | Change |
|------|--------|
| `AbstractHibernateCriteriaBuilder.java` | Handle composite ID components in `invokeMethod()` by catching `IllegalArgumentException` from `getAttribute()` and falling back to entity type checking |
| `CompositeIdCriteria.groovy` | Add 2 new tests for composite ID criteria projections and restrictions |

## Verification

- 578 existing tests pass (0 failures)
- 2 new tests added covering projections and restrictions on composite ID component associations
- All 5 CompositeIdCriteria tests pass

## Reproducer App

https://github.com/jamesfredley/grails-composite-id-criteria-14516

### DIFF
--- a/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
+++ b/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
@@ -1806,7 +1806,45 @@ public abstract class AbstractHibernateCriteriaBuilder extends GroovyObjectSuppo
             if (pd != null && pd.getReadMethod() != null) {
                 final Metamodel metamodel = sessionFactory.getMetamodel();
                 final EntityType<?> entityType = metamodel.entity(targetClass);
-                final Attribute<?, ?> attribute = entityType.getAttribute(name);
+
+                Attribute<?, ?> attribute = null;
+                try {
+                    attribute = entityType.getAttribute(name);
+                } catch (IllegalArgumentException e) {
+                    // Composite ID components may not be registered as JPA Metamodel
+                    // attributes. Fall back to checking if the property type is a managed
+                    // entity so the criteria builder can navigate associations that form
+                    // part of a composite key (e.g. UserRole with composite: ['user', 'role']).
+                    Class<?> propertyType = pd.getPropertyType();
+                    try {
+                        metamodel.entity(propertyType);
+                        // Property type is a managed entity - treat as association
+                        Class oldTargetClass = targetClass;
+                        targetClass = propertyType;
+                        if (targetClass.equals(oldTargetClass) && !hasMoreThanOneArg) {
+                            joinType = org.hibernate.sql.JoinType.LEFT_OUTER_JOIN.getJoinTypeValue();
+                        }
+                        associationStack.add(name);
+                        final String associationPath = getAssociationPath();
+                        createAliasIfNeccessary(name, associationPath, joinType);
+                        logicalExpressionStack.add(new LogicalExpression(AND));
+                        invokeClosureNode(callable);
+                        aliasStack.remove(aliasStack.size() - 1);
+                        if (!aliasInstanceStack.isEmpty()) {
+                            aliasInstanceStack.remove(aliasInstanceStack.size() - 1);
+                        }
+                        LogicalExpression logicalExpression = logicalExpressionStack.remove(logicalExpressionStack.size() - 1);
+                        if (!logicalExpression.args.isEmpty()) {
+                            addToCriteria(logicalExpression.toCriterion());
+                        }
+                        associationStack.remove(associationStack.size() - 1);
+                        targetClass = oldTargetClass;
+                        return name;
+                    } catch (IllegalArgumentException ignored) {
+                        // Not a managed entity type - re-throw original exception
+                        throw e;
+                    }
+                }
 
                 if (attribute.isAssociation()) {
                     Class oldTargetClass = targetClass;
--- a/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
+++ b/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
@@ -1841,8 +1841,9 @@ public abstract class AbstractHibernateCriteriaBuilder extends GroovyObjectSuppo
                         targetClass = oldTargetClass;
                         return name;
                     } catch (IllegalArgumentException ignored) {
-                        // Not a managed entity type - re-throw original exception
-                        throw e;
+                        // Not a managed entity type - wrap the original exception to preserve stack trace
+                        throw new IllegalArgumentException(
+                                "Unable to locate attribute [" + name + "] on entity [" + entityType.getName() + "]", e);
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
```diff
@@ -1806,7 +1806,45 @@ else if (paginationEnabledList) {
...
+                        if (!aliasInstanceStack.isEmpty()) {
+                            aliasInstanceStack.remove(aliasInstanceStack.size() - 1);
+                        }
+                        LogicalExpression logicalExpression = logicalExpressionStack.remove(logicalExpressionStack.size() - 1);
+                        if (!logicalExpression.args.isEmpty()) {
+                            addToCriteria(logicalExpression.toCriterion());
+                        }
+                        associationStack.remove(associationStack.size() - 1);
+                        targetClass = oldTargetClass;
+                        return name;
```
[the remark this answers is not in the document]
jamesfredley: The duplication is intentional. The composite ID fallback (catch block) handles a fundamentally different case from the normal association path:

1. **Normal path** (line 1850+): Uses `getClassForAssociationType(attribute)` which relies on the JPA `Attribute` object being available
2. **Composite ID fallback** (catch block): The JPA `Attribute` is *not available* (that's why the `getAttribute()` threw) - we use `pd.getPropertyType()` from the `BeanWrapper` instead

These paths cannot be unified because `getClassForAssociationType()` requires the JPA `Attribute` that doesn't exist for composite ID components. The fallback deliberately uses a different mechanism (`BeanUtils.getPropertyDescriptor`) to resolve the association type. Merging them would require a synthetic `Attribute` wrapper, adding complexity without benefit for this edge case.

grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
```diff
@@ -1806,7 +1806,45 @@ else if (paginationEnabledList) {
...
+                    } catch (IllegalArgumentException ignored) {
+                        // Not a managed entity type - re-throw original exception
+                        throw e;
```
jdaugherty: This should be wrapped.  Rethrowing looses the original internal line it threw from

grails-data-hibernate5/core/src/test/groovy/grails/gorm/tests/compositeid/CompositeIdCriteria.groovy
```diff
@@ -79,6 +79,49 @@ class CompositeIdCriteria extends Specification {
...
+    CompositeIdToMany compositeIdToMany = new CompositeIdToMany(author:_author, book:_book).save(failOnError:true, flush:true)
+
+    when: "querying with projections navigating composite ID component associations"
+    def results = CompositeIdToMany.createCriteria().list {
+      projections {
+        book {
+          property('id')
+        }
+      }
+      author {
```
jdaugherty, line 95: what about eq('author', author) as a test too?

jdaugherty (changes requested):

grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateCriteriaBuilder.java
jamesfredley: Fixed in 020e56a - now wraps with a descriptive message and the original exception as cause:
```java
throw new IllegalArgumentException(
    "Unable to locate attribute [" + name + "] on entity [" + entityType.getName() + "]", e);
```

grails-data-hibernate5/core/src/test/groovy/grails/gorm/tests/compositeid/CompositeIdCriteria.groovy
jamesfredley: Added in 020e56a - new test "test that eq on composite id component entity works" exercises `eq('author', _author)` directly on the composite ID component. All 6 tests pass.

jdaugherty (approved):
We may want to test a proxy scenario too

davydotcom (approved):

Issue #14814, opened 2025-06-14
7.0.0-M4 BUG - Unexpected URLMapping Resolution


codeconsole:
### Expected Behavior

This worked fine prior but broke M4.  If you have a plugin with a controller, it takes priority over the controller with the same name even if it is in a different namespace IF you run a jar.  If you run via bootRun, it works fine.

plugin/grails-app/controllers/UserController
```groovy
class UserController {
     static namespace = 'myplugin'

     def index() {}
}
```

grails-app/controllers/UserController
```groovy
class UserController {
     def index() {}
}
```

UrlMappings .groovy
```
"/$controller/$action?/$id?(.$format)?"
```

grails bootRun
http://localhost/user/index -> grails-app/controllers/UserController.index

grails bootJar
java -jar build/.../sample.jar
http://localhost/user/index ->plugin/grails-app/controllers/UserController.index

### Version

7.0.0-M4

jdaugherty:
I created this branch: https://github.com/apache/grails-core/tree/url-mapping-testing

Added a controller "DuplicateController" in grails-test-examples/app1/.../DuplicateController
Added a controller "DuplicateController" in grails-test-examples/plugins/loadfirst/...DuplicateController

Ran the gradle build:

        ./gradlew build

changed to the libs directory:

        cd grails-test-examples/app1/build/libs

ran the bootJar:

        java -jar grails-test-examples-app1-0.1.jar

Visited: http://localhost:8080/duplicate which shows "app controller" so I can't immediately reproduce this.  Can you please provide an application to reproduce it?  

In the past these bugs have been because of a bug in the gradle plugin where the groovy compiler script isn't picked up.  Is it possible you're still using the older gradle plugins?  

Can you look in the plugin build directory, there should be a grailsGroovyCompilerConfig, is there?  For example in the branch I created, it has this content for compile: 

        withConfig(configuration) {
            inline(phase: 'CONVERSION') { source, context, classNode ->
                classNode.putNodeMetaData('projectVersion', '0.1-SNAPSHOT')
                classNode.putNodeMetaData('projectName', 'grails-test-examples-plugins-loadafter')
                classNode.putNodeMetaData('isPlugin', 'true')
            }
        }

If this script is missing, the behavior described in this ticket would exist.

codeconsole:
@jdaugherty this issue is actually **much deeper** and no long think it has to do with UrlMappings

[...]

PR #15464, opened 2026-02-26
BUG 7.0.x - Fixes prioritization of a non-namespaced controller when no namespace is defined

codeconsole:
During `registerController()`, namespaced plugin controllers register themselves at null-namespace keys as fallbacks. The deferred merge then runs for app controllers. 

Previously, `putIfAbsent` meant an app's non-namespaced controller could never reclaim a null-namespace slot already occupied by a namespaced plugin controller. 

Now, when all three conditions are met — the key has no namespace, the deferred controller has no namespace, and the existing controller does have a namespace — the non-namespaced controller wins, because it's the exact match for a null-namespace lookup.

### Example

create a grails app with the default UrlMappings.groovy
```groovy
        "/$controller/$action?/$id?(.$format)?"{
            constraints {
                // apply constraints here
            }
        }
```

Now create a namespace controller and a non-namespace controller.

Even though no namespace is defined in the above mapping, Grails will randomly select between the 2 controllers. 
The radom behavior is determined on start up and will consistently chose the same controller after startup.  So sometimes you will get an app that starts up and only uses the namespace controller and other times you will get the non-namespace controller.

This behavior is independent of byte code.   The way I discovered it was by running a multiple instance application. 
Some instances would serve one way and others would server a different way.  All of the instances used the same jar / byte code / server configuration.

### DIFF
--- a/grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/AbstractGrailsControllerUrlMappings.groovy
+++ b/grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/AbstractGrailsControllerUrlMappings.groovy
@@ -61,7 +61,17 @@ abstract class AbstractGrailsControllerUrlMappings implements UrlMappings {
         }
 
         for (Map.Entry<ControllerKey, GrailsControllerClass> entry: deferredMappings.entrySet()) {
-            mappingsToGrailsControllerMap.putIfAbsent(entry.key, entry.value)
+            ControllerKey key = entry.key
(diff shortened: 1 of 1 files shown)


codeconsole:
Fixes https://github.com/apache/grails-core/issues/14814

sbglasius (approved):
The arguments set forth in this PR seams reasonable.

If possible a test would be nice, but given the nature of this fix, I can see why it's absent.

matrei (approved):
LGTM

PR #15470, opened 2026-02-27
Add for removal to previously deprecated methods and classes

matrei:
This will make it easier to remove these in 8 or later.


jdaugherty (approved):
This seems reasonable.  It also reminded me of the GrailsDomainClass vs PersitentEntity change.  Should we consider removing domainclass?  I know it's a major change but in line with hibernate it may be worth it.

matrei:
> This seems reasonable. It also reminded me of the GrailsDomainClass vs PersitentEntity change. Should we consider removing domainclass? I know it's a major change but in line with hibernate it may be worth it.

I looked at `GrailsDomainClass`, but skipped on as it was not an easy removal like the rest of these. But we should definitely figure out the correct way to deal with that deprecation also. There is also `JsonBuilder` in `grails-converters`

jamesfredley (approved):

jdaugherty:
@matrei both James and I agree on this but this PR is a draft.  Can we move forward with this?

jdaugherty (approved):

PR #15475, opened 2026-03-01
fix: normalize Unicode minus sign in GSP number formatting for locale compatibility

### LINKED ISSUE #15178: Negative number bug with gsp and fields and Norwegian locale in browser.

jamesfredley:
## Summary

Fixes #15178 - Negative numbers with Norwegian locale (and similar locales) produce Unicode minus sign (U+2212) instead of ASCII hyphen-minus (U+002D) when rendered in input type=number fields via the f:field tag. This violates the WHATWG HTML spec which requires ASCII hyphen-minus for valid floating-point number values.

## Changes

**Source (1 file):**
- grails-fields/.../FormFieldsTagLib.groovy - Modified getNumberFormatter() to normalize DecimalFormatSymbols.minusSign to ASCII hyphen-minus. This method is called exclusively from renderNumericInput() which outputs input elements, making the fix isolated to the input rendering pipeline.

**Tests (1 file):**
- grails-fields/.../DefaultInputRenderingSpec.groovy - Added data-driven Spock test covering 5 locales (nb_NO, nn_NO, sv_SE, fi_FI, en_US) to verify negative number input values use ASCII minus.

## Design Decisions

Per reviewer feedback, this fix is now scoped to **only** the FormFieldsTagLib input rendering path:
- **No changes to FormatTagLib.formatNumber** - This is a display tag (g:formatNumber) and should preserve locale-specific formatting including Unicode minus.
- **No changes to ValidationTagLib.formatValue** - This is used in both display and form contexts and should not be globally modified.
- **Fix only in FormFieldsTagLib.getNumberFormatter()** - This is explicitly input-specific, called from renderNumericInput() which produces input elements where the WHATWG spec mandates ASCII minus.

## Technical Details

- Java DecimalFormat uses DecimalFormatSymbols.minusSign which varies by locale (e.g., Norwegian uses U+2212)
- The WHATWG HTML spec defines valid floating-point number for input type=number as requiring U+002D (ASCII hyphen-minus)
- The normalization is instance-local (per-formatter), not mutating any shared state
- For locales that already use ASCII minus, the assignment is a no-op

## Verification

- All DefaultInputRenderingSpec tests pass (including 5 new locale tests)
- All existing grails-fields tests remain passing
- No changes to display formatting behavior (g:formatNumber, g:fieldValue)


grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy
```diff
@@ -327,13 +327,25 @@ class FormatTagLib implements TagLibrary {
             number = decimalFormat.parse(number as String)
         }
 
+        // Normalize Unicode minus sign (U+2212) to ASCII hyphen-minus (U+002D)
+        // for HTML compatibility (fixes grails-core#15178)
+        DecimalFormatSymbols formatSymbols = decimalFormat.decimalFormatSymbols
+        formatSymbols.minusSign = '-' as char
```
jdaugherty: Isn't this assuming we are formatting for an input?  From what I can tell, this tag isn't for an input, so this is not the right change

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/ValidationTagLib.groovy
```diff
@@ -471,7 +471,12 @@ class ValidationTagLib implements TagLibrary {
         PropertyEditor editor = registry.findCustomEditor(value.getClass(), propertyPath)
         if (editor) {
             editor.setValue(value)
-            return !(value instanceof Number) ? editor.asText?.encodeAsHTML() : editor.asText
+            String formattedValue = editor.asText
+            if (value instanceof Number && formattedValue != null) {
+                DecimalFormatSymbols formatSymbols = new DecimalFormatSymbols(webRequest.getLocale())
+                formattedValue = formattedValue.replace(formatSymbols.minusSign, '-' as char)
+            }
+            return !(value instanceof Number) ? formattedValue?.encodeAsHTML() : formattedValue
```
jdaugherty: Again, this doesn't appear to be inout specific

jdaugherty (changes requested):
These changes seem to not be the right approach:
1. they assume all output is for an html input
2. they do not handle only that locale case, but rather them all
3. they are not isolated and specific - we should be changing the property binding and output for only this locale and only where used in an input 
4. they have side effect behavior; wouldn't it be better to implement or extend our own formatter and document these rules?  how have others handled this?  does the formatter already have an ascii solution?

This change feels rushed and hacky because its handling a specific case in generic code

jamesfredley:
Addressed all review feedback. The fix is now scoped to only FormFieldsTagLib.getNumberFormatter() which is explicitly input-specific (called from renderNumericInput() for HTML input elements). All changes to FormatTagLib, ValidationTagLib, and functional tests have been reverted. Added data-driven Spock test covering nb_NO, nn_NO, sv_SE, fi_FI, and en_US locales.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/ValidationTagLib.groovy
jamesfredley: PR updated with a completely new approach.

jdaugherty (approved):

jdaugherty:
The change looks much better now!

PR #15476, opened 2026-03-01
docs: add Spring Security plugin upgrade instructions

### LINKED ISSUE #15388: Add Spring Security upgrade instructions to documentation
### Issue description

From the weekly meeting, we noticed that while the rename.md file & associated script reference the new library names, these mappings were not explicitly called out in the upgrade notes.  This ticket is to add a blurb about spring security plugin, where it was combined, and the new package names.

jamesfredley:
## Summary

- Add section 12.27 to the upgrade guide documenting the Spring Security plugin changes in Grails 7
- Documents the consolidation of previously separate plugin repos (core, ACL, CAS, LDAP, OAuth2, REST, UI) into a single `apache/grails-spring-security` repository
- Lists all artifact coordinate changes in a table, highlighting that `spring-security-core` was renamed to `grails-spring-security` (dropped the `-core` suffix)
- Calls out the less obvious REST submodule renames (`rest-gorm` -> `rest-datamapping`, `rest-grailscache` -> `grails-cache`)
- References the existing coordinate rename script and BOM for automated migration
- Fix incorrect artifact names in `RENAME.md` and `etc/bin/rename_gradle_artifacts.sh`:
  - `grails-spring-security-grails-cache` -> `grails-spring-security-rest-grailscache` (verified against actual `publishArtifactId`)
  - `spring-security-rest-testapp` -> `spring-security-rest-testapp-profile` (verified against actual `publishArtifactId`)

Fixes #15388


RENAME.md
```diff
@@ -131,7 +131,7 @@ Below is a reference of all migrated artifacts - both their old and new name.
...
 | org.grails.plugins                | spring-security-rest                            | org.apache.grails                               | grails-spring-security-rest                                   |                        |                        | grails-spring-security |
 | org.grails.plugins                | spring-security-rest-gorm                       | org.apache.grails                               | grails-spring-security-rest-datamapping                       |                        |                        | grails-spring-security |
-| org.grails.plugins                | spring-security-rest-grailscache                | org.apache.grails                               | grails-spring-security-grails-cache                           |                        |                        | grails-spring-security |
+| org.grails.plugins                | spring-security-rest-grailscache                | org.apache.grails                               | grails-spring-security-rest-grailscache                       |                        |                        | grails-spring-security |
```
jdaugherty, line 134: Why not fix the artifact id instead?  I think this was a mistake

grails-doc/src/en/guide/upgrading/upgrading60x.adoc
```diff
@@ -865,3 +865,87 @@ grails:
 * **8.0**: `SimpleEnumMarshaller` will become the default
 
 The legacy `org.grails.web.converters.marshaller.json.EnumMarshaller` and `org.grails.web.converters.marshaller.xml.EnumMarshaller` classes are marked as `@Deprecated(forRemoval = true, since = "7.0.2")` and will be removed in Grails 8.0.
+
+===== 12.27 Spring Security Plugins
```
jdaugherty, line 869: We have intentionally kept security docs separate.  I am not sure we should be putting this detail here.  Instead, why not link to the security docs?

jdaugherty (changes requested):

grails-doc/src/en/guide/upgrading/upgrading60x.adoc
jamesfredley: Was going off https://github.com/apache/grails-core/issues/15388, but we can move or close this.

jdaugherty: I think we just need to link to the doc.  i am ok changing the rename script / rename.md since we discussed it

jamesfredley: I think we will need to create a good location on https://apache.github.io/grails-spring-security/.  Consolidating those docs, might be part of the effort, so this might need to wait.

jdaugherty: I'm against putting all of this in the main grails-core docs, referencing it is completely and 100% acceptable.  We are rapidly expanding docs and they will diverge since we may update in different orders.  I'd rather we fix the spring security docs since that's where that info lives.  

My main worry here is this information getting out of date and that causes more problems then not having them.

jdaugherty: Here's the necessary changes to create a single guide view of the spring security plugin: https://github.com/apache/grails-spring-security/pull/1211

jamesfredley:
This is replaced by https://github.com/apache/grails-spring-security/pull/1211 and https://github.com/apache/grails-core/issues/15516

Issue #14986, opened 2025-08-18
grails-doc disable publishing of /guides/pages/*


jamesfredley:
### Issue description

The html files generated in /guides/pages/* are already included in html files in /guides/*, where they belong.  The html files generated in /guides/pages/* have broken links, css, js, etc, since they are built to be included in html files in /guides/*

jdaugherty:
I'm not sure this can be removed.  You have to generate documentation to a different directory in gradle to ensure that gradle can cache the output.  I believe the current design writes to a directory, then copies that documentation & other artifacts to a shared directory.  I think this ticket can be closed @jamesfredley

jamesfredley:
We could potentially exclude them from being published to avoid:  

https://github.com/apache/grails-website/tree/asf-site-production/docs/latest/guide/pages
https://grails.apache.org/docs/latest/guide/pages/RSSAndAtom.html

mika-collins:
I have submitted a PR to address the issue. Please review and let me know if I need to make any changes.

PR #15477, opened 2026-03-01
fix: disable publishing of /guide/pages/** in grails-doc/build.gradle

mika-collins:
This PR addresses issue #14986. The html files generated in /guide/pages/ have broken links, css, js, etc. 

**Changes:**
Added an `exclude` rule to the `docs` Sync task in `grails-doc/build.gradle` to prevent the `/guide/pages/` directory from being included in the final documentation output. 

**Verification:**
Verified locally by running `.\gradlew :grails-doc:docs` and confirming that `build/docs/guide/pages/` is no longer present in the output, while the primary guide remains.

### DIFF
--- a/grails-doc/build.gradle
+++ b/grails-doc/build.gradle
@@ -294,7 +294,11 @@ docsTask.configure { Sync it ->
 
     it.outputs.dir mergedDocsDir
 
-    it.from manualDocsDir, apiDocsDir, dataDocsDir
+    it.from(manualDocsDir) {
+        exclude "guide/pages/**"
+    }
+
+    it.from apiDocsDir, dataDocsDir
     it.into mergedDocsDir
 }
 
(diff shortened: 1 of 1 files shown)


grails-doc/build.gradle
```diff
@@ -294,7 +294,11 @@ docsTask.configure { Sync it ->
 
     it.outputs.dir mergedDocsDir
 
-    it.from manualDocsDir, apiDocsDir, dataDocsDir
+    it.from(manualDocsDir) {
+        exclude "guide/pages/**"
```
jamesfredley, line 300: @mika-collins Let's add a comment here, something like the following

```
// Exclude page fragments - they have broken links/CSS/JS and are already
// included in the assembled guide pages (see #14986)
```

jamesfredley:
The build failure was unrelated to these changes.  Merged latest 7.0.x changes into this PR branch.

grails-doc/build.gradle
mika-collins: Thanks! I have added the comment to the most recent commit.

jamesfredley (approved):

PR #15478, opened 2026-03-01
New Feature: HttpClient Testing Support

matrei:
This pull request adds a new `org.apache.grails:grails-testing-support-httpclient` module and refactors the HTTP client usage in integration tests to use it, improving consistency and maintainability of test code.


.github/workflows/groovy-joint-workflow.yml
```diff
@@ -159,6 +159,7 @@ jobs:
         run: >
           ./gradlew build
           -x groovydoc
+          -PgebAtCheckWaiting
```
jdaugherty, line 168: Well that explains why this has been flakey!

grails-doc/src/en/guide/introduction/whatsNew.adoc
```diff
@@ -94,3 +94,14 @@ The `g:form` tag now automatically provides csrf protection when Spring Security
 A Grails banner was introduced in Grails 7 that is displayed on application startup.
 From Grails 7.1 onwards, this banner now shows versions of foundational dependencies
 and can be xref:conf.adoc#customizing-the-banner[customized].
+
+==== Grails HTTP Client Testing Support (Grails 7.1+)
```
jdaugherty, line 98: Oof ... I forgot there was a whatsNew doc, and have been adding this stuff to the upgrade guide instead. Good catch.

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/api/NamespacedBookSpec.groovy
```diff
@@ -16,142 +16,103 @@
...
-            rsp.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/json;charset=UTF-8'
-            rsp.body().api == 'version 1.0 (Namespaced)'
-            rsp.body().title == 'API - The Shining'
+        def response = http('/api/book')
+
+        then: 'The response is correct'
+        response.expectJson(200, 'Content-Type': 'application/json;charset=UTF-8', [
+                api: 'version 1.0 (Namespaced)',
+                title: 'API - The Shining'
+        ])
```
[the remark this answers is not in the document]
matrei: This is not correct. Map param first becomes named params in Groovy.

gradle/publish-root-config.gradle
```diff
@@ -79,6 +79,7 @@ def publishedProjects = [
         'grails-testing-support-dbcleanup-core',
         'grails-testing-support-dbcleanup-h2',
         'grails-testing-support-dbcleanup-postgresql',
+        'grails-testing-support-httpclient',
```
jdaugherty: Trying to think longer term here; I'd like to add the ersatz server as a default configuration as well.  This way you can test calling your code & you can test your code calling a mock endpoint.  From a naming perspective, should we call this `grails-testing-support-http-client` and then later we'd call the other `grails-testing-support-http-server`?  The reason I ask is I can envision eventually having a `grails-testing-support-http-core` that's share between them.

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/ProductSpec.groovy
```diff
@@ -16,209 +16,161 @@
...
-        HttpRequest request = HttpRequest.GET('/products')
-        HttpResponse<String> resp = client.toBlocking().exchange(request, String)
-        Map body = objectMapper.readValue(resp.body(), Map)
+        def response = http('/products')
 
         then:
-        resp.status == HttpStatus.OK
-        resp.headers.getFirst(HttpHeaders.CONTENT_TYPE).isPresent()
-        resp.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/hal+json;charset=UTF-8'
+        response.expectHeaders(200, 'Content-Type': 'application/hal+json;charset=UTF-8')
```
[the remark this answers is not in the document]
matrei: Not correct, it uses the named param feature in Groovy.

grails-testing-support-httpclient/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+    id 'org.apache.grails.buildsrc.sbom'
+    id 'org.apache.grails.gradle.grails-code-style'
+}
+
+version = projectVersion
+group = 'org.apache.grails'
+
+ext {
+    pomTitle = 'Grails Http Client Testing Support'
+    pomDescription =
```
jdaugherty, line 33: I can't make the comment at the folder level, but should we add a README.md with a basic description of this project as well?

grails-testing-support-httpclient/src/main/groovy/org/apache/grails/testing/httpclient/HttpClientSupport.groovy
```diff
@@ -0,0 +1,418 @@
...
+    /**
+     * Shared singleton client reused across tests.
+     */
+    private static volatile HttpClient sharedClient
+
+    /**
+     * @return the shared singleton client instance, creating it on first access.
+     */
+    HttpClient getHttpClient() {
+        def client = sharedClient
```
jdaugherty: * I'm assuming the HttpClient is thread safe?  
* I was expecting 1 client per spec class, but this appears to share it across all tests
* Why not make use of the Spock infrastructure to create these in a spock extension?

grails-test-examples/app1/src/integration-test/groovy/functionaltests/flow/FlashChainForwardSpec.groovy
```diff
@@ -40,95 +38,69 @@ import grails.testing.mixin.integration.Integration
...
-            location
+                new URL(location).path + (new URL(location).query ? "?${new URL(location).query}" : '') :
+                location
         
-        def request2 = HttpRequest.GET(redirectPath)
+        def headers = [:] as Map<String, String>
         if (sessionCookie) {
-            request2 = request2.header('Cookie', sessionCookie)
+            headers.put('Cookie', sessionCookie)
         }
```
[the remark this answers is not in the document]
matrei: It's good enough for this test case.

grails-testing-support-httpclient/src/main/resources/META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule
```diff
@@ -0,0 +1,3 @@
+moduleName=grails-testing-support-httpclient
+moduleVersion=1.0
+extensionClasses=org.apache.grails.testing.httpclient.HttpResponseExtensions
```
jdaugherty: Does this mean these extensions would be injected on the code being tested too?  Wouldn't that potentially cause a different behavior in test vs prod/dev?  Would it be better to use the `@Delegate` pattern and wrap the http response returned by this library so there's no side effect behavior?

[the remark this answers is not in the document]
matrei: Good catch!

grails-test-examples/cache/src/integration-test/groovy/com/demo/AdvancedCachingIntegrationSpec.groovy
```diff
@@ -67,257 +60,146 @@ class AdvancedCachingIntegrationSpec extends ContainerGebSpec {
...
         when: "evicting all and fetching again"
-        client.toBlocking().exchange(HttpRequest.GET('/advancedCaching/evictAllKeyCache'), String)
-        HttpResponse<String> second = client.toBlocking().exchange(
-            HttpRequest.GET('/advancedCaching/getDataByKey?key=anykey'),
-            String
-        )
-        def secondData = new JsonSlurper().parseText(second.body()).data
+        http('/advancedCaching/evictAllKeyCache')
+        def second = http('/advancedCaching/getDataByKey?key=anykey',)
+        def secondData = second.json().data
```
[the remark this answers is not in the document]
matrei: Good catch!

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/api/NamespacedBookSpec.groovy
```diff
@@ -16,142 +16,103 @@
...
-        then: 'The rsponse is correct'
-        rsp.status() == HttpStatus.OK
-        rsp.headers.getFirst(HttpHeaders.CONTENT_TYPE).isPresent()
-        rsp.headers.getFirst(HttpHeaders.CONTENT_TYPE).get() == 'application/json;charset=UTF-8'
-        rsp.body().api == 'version 1.0 (Namespaced)'
-        rsp.body().title == 'API - The Shining'
+        def response = http('/api/book/testRender')
+
+        then: 'The responseonse is correct'
+        response.expectJson(200, 'Content-Type': 'application/json;charset=UTF-8', [
```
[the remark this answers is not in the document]
matrei: Good catch!

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/api/NamespacedBookSpec.groovy
```diff
@@ -16,142 +16,103 @@
...
+                api: 'version 1.0 (Namespaced)',
+                title: 'API - The Shining'
+        ])
     }
 
-    void 'test rspond(foo, view: ..) in controllers with namespaces works'() {
+    void 'test responseond(foo, view: ..) in controllers with namespaces works'() {
         when: 'A request is sent to a controller with a namespace'
```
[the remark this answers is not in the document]
matrei: Good catch!

grails-doc/src/en/guide/testing/integrationTesting.adoc
```diff
@@ -286,6 +286,124 @@ Example usage:
...
+    }
+}
+----
+
+`HttpClientSupport` uses the JDK `HttpClient` and applies default request and connect timeouts of 60 seconds and
+a redirect policy of always follow redirects. You can use helpers `httpRequestWith(Closure configurer)` and
+`httpClientWith(Closure configurer)` to customize your own client and requests starting from the defaults.
+
+Additional request helper methods available include `httpPut()`, `httpPatch()`, `httpDelete()`, `httpOptions()`, and many assertion methods on responses.
+
```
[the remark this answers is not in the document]
matrei: The first issue is not true, it's using Groovy named params.
Good catch on the second issue!

jdaugherty (commented):
This is such an incredible amount of work: 
1. it's a clear API
2. it doesn't pollute our dependencies
3. it's useful for all Grails apps and we adopted it ourselves (so the benefits are immediately clear in this PR)
4. we own the API so we can enhance it more rapidly than Micronaut or Spring Boot 

I have some minor concerns, mainly: 
1. the assumption is this is used in specs, but we're not following the recommended pattern for spec enhancement (using an extension)
2. the naming - I'd like to propose an alternative with the assumption of a future addition
3. could including this library with extensions cause a difference in behavior between prod/dev and test?  Ancillary to this, I think using a delegate pattern may be better for testing, but I can definitely see the http client enhancements useful for a grails app itself.  Interested to discuss this one.

jdaugherty:
I know this is a new feature, but i'd argue we should back port it.  We don't have to announce it for 7.0.x, but the reason it should exist in 7.0.x is for easier maintenance given how much clean up on the tests have been done

gradle/publish-root-config.gradle
matrei: I'll rename to open up for more testing support http modules 👍

grails-testing-support-httpclient/src/main/groovy/org/apache/grails/testing/httpclient/HttpClientSupport.groovy
matrei: To my understanding, HttpClient is immutable and thread-safe. I manages it's own private thread pool if a custom `Executor` is not explicitly set. I did have one client per spec initially, but I don't think that is necessary.

matrei:
Thank you @jdaugherty for excellent feedback!
I'm re-hauling the PR again, new changes coming soon.

grails-testing-support-httpclient/src/main/groovy/org/apache/grails/testing/httpclient/HttpClientSupport.groovy
jdaugherty: Makes sense, the only change I would suggest then is the spock extension.

grails-testing-support-httpclient/build.gradle
matrei: Readme added

grails-testing-support-httpclient/src/main/groovy/org/apache/grails/testing/httpclient/HttpClientSupport.groovy
matrei: For various reasons, I went in another direction than Spock extension for this library.
HttpClient is now auto-configured as a Spring bean that can be injected in tests.

grails-testing-support-httpclient/src/main/resources/META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule
matrei: This is a really good point, I removed the Groovy Extension.

grails-testing-support-http-client/README.md
```diff
@@ -0,0 +1,57 @@
...
+    void 'health endpoint responds OK'() {
+        expect:
+        http.get('/health')
+            .expectStatus(200)
+            .expectContains('UP')
+    }
+}
+```
+
+For non-`@Integration` Spring Boot tests, you can opt in explicitly with `@EnableHttpClientSupport`.
```
jdaugherty: Does this still work with your bean wiring configuration change?

grails-testing-support-core/src/main/groovy/grails/testing/spring/IntegrationTestAutoConfiguration.groovy
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Marker annotation for Spring test configuration classes that should be auto-imported into
+ * Integration tests by {@code IntegrationTestAstTransformation}.
+ *
+ * @since 7.1
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target([ElementType.TYPE])
+@TestConfiguration(proxyBeanMethods = false)
+@interface IntegrationTestAutoConfiguration {
```
jdaugherty, line 37: So instead of a spock extension, you went with the bean approach.  I assume that's why this change is now for 7.1.x instead of 7.0.x?  I don't think it's a problem to use the test configuration approach, but can you give background on what led you down this path?

grails-testing-support-http-client/README.md
matrei: Yes

grails-testing-support-core/src/main/groovy/grails/testing/spring/IntegrationTestAutoConfiguration.groovy
matrei: As it is a new feature, it was always intended it for 7.1.x.
For proper IDE support I think we need an object to work with.

matrei: @jdaugherty So, a bit more context on why I went the `@TestConfiguration` route. I read [here](https://spockframework.org/spock/docs/2.4/known_issues.html#_using_traits_with_specifications) that using Traits on Specifications is not supported/recommended.

Now that I think about this some more, maybe this was overly cautious, as we apparently do that a lot already in Grails testing support:

- `UrlMappingsUnitTest`
- `TagLibUnitTest`
- `EnvironmentAwareSpec`
- `GrailsUnitTest`
- `GrailsWebUnitTest`
- `DataTest`
- `JsonViewTest`
- `MocksDomains`
- etc.

I'll revert to using a trait...

jdaugherty: My biggest concern of putting this in 7.1.x is this actually makes it easier to test grails-core.  Would it be possible to put this in 7.0.x and only publish it in 7.1.x for external use?  I want to make merges easier ...

matrei: I think that should be possible by referencing `grails-testing-support-http-client` as a project dependency instead of as a "usual" dependency in the test projects and remove it from publishing.

jdaugherty:
I see 7.1.x commits in the commit history for merging to 7.0.x  @matrei

matrei:
Replaced by #15505

PR #15482, opened 2026-03-03
fix: handle null argument in GORM domain class constructor (Groovy 4 regression)

davydotcom:
## Problem

Calling `new DomainClass(null)` on a GORM domain class throws:

```
groovy.lang.GroovyRuntimeException: Could not find matching constructor for: DomainClass(null)
```

This is a regression from Grails 6 / Groovy 3, where `new DomainClass(null)` was treated equivalently to `new DomainClass()`.

## Root Cause

In Groovy 3, the runtime resolved `new SomeClass(null)` by matching it to the implicit map-based constructor that Groovy provides for POGOs. When `null` was passed, the runtime simply created the object via the no-arg constructor without setting any properties.

Groovy 4 changed how constructor resolution works at runtime. The implicit map constructor is no longer matched when `null` is passed as the argument, causing the `GroovyRuntimeException`.

This is a common pattern in Grails applications — for example, when a variable that may be `null` is passed to a domain constructor:

```groovy
def props = maybeGetProperties() // could return null
def obj = new MyDomain(props)    // fails in Grails 7 / Groovy 4
```

## Fix

The `GormEntityTransformation` (the `@Entity` AST transform) now injects an explicit `Map` constructor into every GORM entity class. The constructor:

- Accepts a `Map` parameter, allowing `null` to match at runtime
- If the map is not `null`, sets properties via `InvokerHelper.setProperties(this, args)` — the same mechanism Groovy uses internally for map constructors
- If `null`, does nothing (equivalent to the no-arg constructor)
- Also ensures a no-arg constructor is preserved, since adding any explicit constructor prevents Groovy from auto-generating the default one

The constructor is injected early in the transformation (before trait composition), so trait `\$init\$` methods are properly wired into it by the Groovy trait composer.

## Tests

Added `NullConstructorArgSpec` in `grails-data-hibernate5/core` with three tests:
1. No-arg constructor works (baseline)
2. Null constructor argument works the same as no-arg
3. Domain created with null constructor argument can be saved

All 598 existing hibernate5 core tests continue to pass with 0 failures.

### DIFF
--- a/grails-data-hibernate5/core/src/test/groovy/grails/gorm/tests/NullConstructorArgSpec.groovy
+++ b/grails-data-hibernate5/core/src/test/groovy/grails/gorm/tests/NullConstructorArgSpec.groovy
@@ -1,0 +1,57 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


jdaugherty:
Closing this, i think teh core issue was the rest transforms weren't included in your project.

PR #15492, opened 2026-03-11
fix #15491 - Bug: DetachedCriteria.distinct().property() projections exclude rows where the projected property value is null

### LINKED ISSUE #15491: DetachedCriteria.distinct().property() projections excludes rows where the projected property value is null
### Issue description

Commit ddae6ee795 ("Auto-create aliases for joined association projections") introduced a regression in AbstractHibernateQuery.calculateProjectionPropertyName(). When a projection references a dotted association path like warehouse.id, the method auto-creates a Hibernate alias via getOrCreateAlias(). That method defaults to INNER_JOIN when no explicit join type is registered, which filters out all rows where the association is null.

jdaugherty:
Commit ddae6ee795 ("Auto-create aliases for joined association projections") introduced a regression in AbstractHibernateQuery.calculateProjectionPropertyName(). When a projection references a dotted association path like warehouse.id, the method auto-creates a Hibernate alias via getOrCreateAlias(). That method defaults to INNER_JOIN when no explicit join type is registered, which filters out all rows where the association is null.

Assisted-by: OpenCode <opencode@opencode.ai>
Assisted-by: Claude <claude@anthropic.com>

### DIFF
--- a/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateQuery.java
+++ b/grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/query/AbstractHibernateQuery.java
@@ -708,7 +708,13 @@ public abstract class AbstractHibernateQuery extends Query {
             }
             associationPath.append(token);
 
-            CriteriaAndAlias criteriaAndAlias = getOrCreateAlias(associationPath.toString(), generateAlias(token));
+            // Use LEFT JOIN for auto-created projection aliases so that rows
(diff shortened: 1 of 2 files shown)


jamesfredley (approved):

borinquenkid (approved):
Summary: LGTM (Looks Good To Me)

The transition from a default join to an explicit JoinType.LEFT in AbstractHibernateQuery.java correctly addresses the reported issue. By ensuring that auto-created projection aliases use a left join, we prevent Hibernate from filtering out rows where the associated property is null, which is the expected behavior for GORM projections.

PR #15500, opened 2026-03-12
Remove remnants of `doWithWebDescriptor`

matrei:
Also, clean up the javadoc in some classes related to Grails plugins.


grails-core/src/main/groovy/grails/plugins/GrailsPluginManager.java
```diff
@@ -262,8 +260,8 @@ public interface GrailsPluginManager extends ApplicationContextAware {
     String getPluginPath(String name, boolean forceCamelCase);
 
     /**
-     * Looks up the plugin that defined the given instance. If no plugin
-     * defined the instance then null is returned.
+     * Looks up the plugin that defined the given instance.
+     * If no plugin is defined the instance, then null is returned.
```
jamesfredley: minor, but maybe one of these: 

- "If no plugin defined the instance, then null is returned." (keep original active voice)
- "If no plugin is defined for the instance, then null is returned."

jamesfredley (approved):

grails-core/src/main/groovy/grails/plugins/GrailsPluginManager.java
matrei: Did a bigger overhaul of the javadoc in `GrailsPluginManager`.

matrei:
The `GrailsPluginMetadataTests` needs #15518 to be merged first to succeed.

jdaugherty:
@matrei can you please update this PR now that the other is merged so the tests show clean?

matrei:
> @matrei can you please update this PR now that the other is merged so the tests show clean?

@jdaugherty The PR is now updated

jdaugherty (approved):

PR #15505, opened 2026-03-14
new module `grails-testing-support-http-client`

matrei:
Only used internally and not published in 7.0.x.
Will be published for external use in 7.1.x.

Replaces #15478


grails-test-examples/app1/src/integration-test/groovy/functionaltests/interceptors/InterceptorAdvancedMatchingSpec.groovy
```diff
@@ -36,233 +34,201 @@ import grails.testing.mixin.integration.Integration
...
+        def response = http('/api/advancedMatching/index')
 
         then: "only namespace interceptor matches (others exclude index)"
-        response.status.code == 200
-        response.body().interceptors.contains('namespace:api')
-        !response.body().interceptors.contains('excludes:index,reset')
-        !response.body().interceptors.contains('multiRule:show|list')
-        !response.body().interceptors.contains('method:POST')
+        response.expectStatus(200)
+        with(response.json()) {
```
jdaugherty, line 215: Is this a spock feature?  How do these lines inside the with get asserted?

grails-test-examples/micronaut/src/integration-test/groovy/micronaut/MicronautErsatzAdvancedSpec.groovy
```diff
@@ -136,37 +124,6 @@ class MicronautErsatzAdvancedSpec extends Specification {
         ersatz.verify()
     }
 
-    void "ANY method matcher in ersatz matches GET request from Micronaut HttpClient"() {
```
jdaugherty, line 139: These tests (and others in this file) appear to be testing the behavior of the eratz server - I agree with removing them, but I suspect there is missing test coverage - maybe the AI didn't understand the intent was to call code that then uses a declarative client that calls the mock server? @jamesfredley we may have some test gaps here.

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/TeamSpec.groovy
```diff
@@ -104,15 +68,11 @@ class TeamSpec extends HttpClientSpec {
     @IgnoreIf({ System.getenv('GITHUB_REF') })
     void 'Test HAL rendering'() {
         when:
-        HttpRequest request = HttpRequest.GET('/teams/hal/1')
-        HttpResponse<String> resp = client.toBlocking().exchange(request, String)
+        def response = http('/teams/hal/1')
+        def lang = "${System.properties.getProperty('user.language')}_${System.properties.getProperty('user.country')}"
```
jdaugherty: Is this actually used?

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/MultipartBody.groovy
```diff
@@ -0,0 +1,191 @@
...
+        private Charset charset = StandardCharsets.UTF_8
+
+        /**
+         * Adds a text form field.
+         *
+         * @param name form field name
+         * @param value field value
+         * @return this builder
+         */
+        MultipartBuilder addPart(String name, String value) {
```
jdaugherty, line 82: Parts do not necessarily have to be a String value.  A good example of this caused me to open this issue recently:  https://github.com/cjstehno/ersatz/issues/196

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/JsonUtils.groovy
```diff
@@ -0,0 +1,212 @@
...
+     * Verifies that the response JSON is structurally equal to the expected object tree.
+     *
+     * <p>Map keys are compared independent of declaration order.</p>
+     *
+     * @param response HTTP response with a JSON body
+     * @param expected expected JSON-compatible object (for example {@code Map}, {@code List}, scalar)
+     * @throws AssertionFailedError if the JSON trees differ
+     */
+    static void verifyJsonTree(HttpResponse<?> response, Object expected) {
+        compareJsonTree(new JsonSlurper().parseText(response.body() as String), expected)
```
jdaugherty: This is really a general comment for all of the JsonSlurper usage in this library: Maybe consider a helper to configure the JsonSlurper?  There are different types of parsers available per JsonParserType and I've had to switch between them in testing before.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy
```diff
@@ -0,0 +1,48 @@
...
+     * <p>
+     * The closure is cloned and executed against a {@link groovy.xml.MarkupBuilder}
+     * delegate using {@link Closure#DELEGATE_FIRST}.
+     *
+     * @param dsl the closure that produces the XML markup
+     * @return the rendered XML string
+     */
+    static String toXml(@DelegatesTo(strategy = Closure.DELEGATE_FIRST, value = MarkupBuilder) Closure<?> dsl) {
+        def writer = new StringWriter()
+        def markupBuilder = new MarkupBuilder(writer)
```
jdaugherty: Same as the JsonSlurper, being able to specify configuration for the MarkupBuilder would be useful.  i.e. whether attributes are escaped.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/HttpClientSupportSpec.groovy
```diff
@@ -0,0 +1,570 @@
...
+
+        then:
+        latch.await(5, TimeUnit.SECONDS)
+        otherThreadClientRef.get() != null
+        mainThreadClient.is(otherThreadClientRef.get())
+    }
+
+    void 'httpRequestWith uses hard-coded default request timeout'() {
+        given:
+        def testClient = new TestClient('http://localhost:8080')
```
jdaugherty, line 84: One of the lessons learned from Spring was they generally do not advise testing http clients with mock calls.  They instead suggest mocking real responses.  I've suggested the eratz server for this in the past, and I think it would be good to use here instead.

jdaugherty (approved):
I really love how much this has cleaned up the tests.  We should 100% merge it, but I'd like to see some future improvement (or even in this pr if you're willing):
1. a way to configure the json/xml helpers 
2. better support for multipart for api testing
3. actually mocking responses instead of mocking method calls in the http client test

I also noticed you removed the usage of enums - specifically HttpStatus.  Generally, I'm a fan of test code not referencing enum values so the test breaks if there is a change.  With that said, I'm not sure the HttpStatus values will ever change.  Spring has an equivalent & we could use it instead.  I'm somewhat indifferent here, but thought I'd put this out there in case we should discuss further.

grails-test-examples/app1/src/integration-test/groovy/functionaltests/interceptors/InterceptorAdvancedMatchingSpec.groovy
jdaugherty: I never knew ... https://spockframework.org/spock/docs/2.4/all_in_one.html#_using_with_for_expectations

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/MultipartBody.groovy
matrei: Doesn't `addPart(String name, String filename, String contentType, byte[] bytes)` work for that use case?

jdaugherty: How do you define this part:  

         Content-Disposition: form-data; name="myJsonObject"
         Content-Type: application/json
   
          {"myKey":"AAAAAAAAAA"}

jdaugherty: Nevermind, I see later on a method on line 123 that will do it

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/JsonUtils.groovy
matrei: I have added support for configuring XML and JSON formatting and parsing.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy
matrei: I have added support for configuring XML and JSON formatting and parsing.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/HttpClientSupportSpec.groovy
matrei: I have switched to use Ersatz.

jdaugherty (approved):

grails-test-examples/views-functional-tests/src/integration-test/groovy/functional/tests/TeamSpec.groovy
matrei: I'm not sure, the test is also disabled in CI.
I'll remove it and the lang field from the expected JSON, and enable the test for CI again.

matrei:
> I also noticed you removed the usage of enums - specifically HttpStatus. Generally, I'm a fan of test code not referencing enum values so the test breaks if there is a change. With that said, I'm not sure the HttpStatus values will ever change. Spring has an equivalent & we could use it instead. I'm somewhat indifferent here, but thought I'd put this out there in case we should discuss further.

I think that if Enums (or Constants) are not providing any value, but rather detracts from readability, we should not use them in tests.

jdaugherty (approved):

jamesfredley (approved):

PR #15522, opened 2026-03-21
7.1.x url mapping greedy extension 3

codeconsole:
Final fix for greedy url mapping parameter

Verified with:
```groovy
        // /scott.smith.json, /scott.json, /12345.json, /scott, /12345
        "/${id}+(.$format)?"(controller:'user', action:'profile') { 
            constraints {
                format(inList: ['json', 'xml'])
            }
        }
        // Fallback for IDs with dots but no valid format (e.g. /scott.smith → id=scott.smith)
        "/${id}"(controller:'user', action:'profile')
```

 I had found another edge case that wasn't working. This PR should get the job done.


codeconsole:
#15213
#15358

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/RegexUrlMapping.java
```diff
@@ -253,8 +253,8 @@ protected Pattern convertToRegex(String url) {
                 // For /(*)?+(\.(*))?  we want regex: /(.+?)?(?:\.([^/.]+))? (optional, greedy)
                 // Only apply greedy regex if this logical URL actually includes the greedy token
                 String processed = urlEnd
```
jamesfredley, line 255: Nit: these comments still reference the old regex (`.+?`). Now that the pattern is `[^/]+?`, the comments should be updated to match:

```java
// For /(*)+(\.(*))?  we want regex: /([^/]+?)(?:\.([^/.]+))?  (required, greedy)
// For /(*)?+(\.(*))?  we want regex: /([^/]+?)?(?:\.([^/.]+))? (optional, greedy)
```

jamesfredley (approved):

jamesfredley:
I merge 7.1.x in to get all the CI fixes after the 3rd party action lockdown after https://infra.apache.org/blog/trivy_security_incident.html

jdaugherty (approved):
Lets update the out of date comment; otherwise I am good to merge

PR #15525, opened 2026-03-23
Add default validation for $namespace, $controller, and $action url mapping variables

codeconsole:
Validates namespace, controller and action.  If validation fails, triggers next mapping.
Literals override wildcards.
```groovy
        "/$namespace/$controller/$action?/$id?(.$format)?" {}
        "/$controller/$action?/$id?(.$format)?"{}
        "/community" { controller: 'topic', action: 'home' } // literals win even when namespace or controller CommunityController exists
        group "/community", namespace: 'community', { 
              "/$controller/$action?"()
              group "/files", controller: 'fileServing', namespace: null, { // allow null namespace overrides on group mappings
                     "/$imageId"()
              }
        } 
        "/$username"(controller: 'profile', action: 'show'
```

Without validation, you have to a do something messy like this:
```
        "/$username"(ontroller: 'profile', action: 'show') {
            constraints {
                username(validator: { val -> !(val in reserved) })
            }
        }
```

I can't think why you would want to disable this feature, but it can be disabled via:

```yaml
grails:                                                                                                                                                    
    web:                                                                                                                                                   
        url:                                                                                                                                               
           mapping:                                                                                                                                       
               validateWildcards: false 
```

Does not affect performance:

  1. Calls info.configure(webRequest) for each match — sets params on the web request
  2. Does a ConcurrentHashMap.get() for the ControllerKey lookup
  3. Now additionally: info.hasWildcardCaptures() (three instanceof checks) when the lookup misses

  Steps 1 and 2 were already happening before this change. The only new cost is step 3, which is negligible.


grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/AbstractGrailsControllerUrlMappings.groovy
```diff
@@ -45,16 +47,20 @@ import org.grails.web.servlet.mvc.GrailsWebRequest
...
     UrlMappings urlMappingsHolderDelegate
     UrlConverter urlConverter
+    boolean validateWildcardMappings
     ConcurrentHashMap<ControllerKey, GrailsControllerClass> mappingsToGrailsControllerMap = new ConcurrentHashMap<>()
     ConcurrentHashMap<ControllerKey, GrailsControllerClass> deferredMappings = new ConcurrentHashMap<>()
 
     AbstractGrailsControllerUrlMappings(GrailsApplication grailsApplication, UrlMappings urlMappingsHolderDelegate, UrlConverter urlConverter = null) {
         this.urlMappingsHolderDelegate = urlMappingsHolderDelegate
         this.urlConverter = urlConverter
+        this.validateWildcardMappings = grailsApplication?.config?.getProperty(
```
sbglasius: Why would `grailsApplication` be `null`? or `config` for that matter? Except perhaps when testing?

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/AbstractGrailsControllerUrlMappings.groovy
```diff
@@ -45,16 +47,20 @@ import org.grails.web.servlet.mvc.GrailsWebRequest
...
+                Settings.WEB_URL_MAPPING_VALIDATE_WILDCARDS, Boolean, true) ?: true
         def controllerArtefacts = grailsApplication.getArtefacts(ControllerArtefactHandler.TYPE)
```
sbglasius, line 64: Since right here `grailsApplication` is not safe referenced.

sbglasius (approved):

matrei (approved):

codeconsole:
I am doing some final testing and want to do 1 more doc update before merging

matrei:
@codeconsole There is an existing config setting with the config key:
```yml
grails:
    urlmapping:
        cache:
            maxsize: 1000
```
Should we use that namespace instead for validateWildcards?
```yml
grails:
    urlmapping:
        validateWildcards: false
        cache:
            maxsize: 1000
```

codeconsole:
> @codeconsole There is an existing config setting with the config key:
> 

@matrei 
yeah, I will create another PR

PR #15532, opened 2026-03-25
Fix overlapping task outputs preventing build cache usage

ribafish:
## Summary

Two issues prevent Gradle build cache from working for GroovyCompile, Checkstyle, and CodeNarc tasks:

1. **`GrailsGradlePlugin.configureGroovyCompiler()`** adds the same output file (`build/grailsGroovyCompilerConfig.groovy`) to every `GroovyCompile` task in a project. When a project has multiple `GroovyCompile` tasks (e.g. `compileGroovy`, `compileTestGroovy`, `compileIntegrationTestGroovy`), Gradle detects overlapping outputs and refuses to cache any of them.

2. **`GrailsCodeStylePlugin`** redirects all Checkstyle/CodeNarc XML reports to `rootProject.buildDir/reports/codestyle/{tool}/{project.name}.xml`. Projects with multiple source sets (e.g. `grails-cache` has `ast` + `main`) have multiple tasks writing to the same report file, again causing overlapping outputs.

## Fix

- Use task-specific config file names (`grailsGroovyCompilerConfig-{taskName}.groovy`) so each `GroovyCompile` task has a unique output path.
- Include the task name in report file paths (`{project.name}-{taskName}.xml`) so each Checkstyle/CodeNarc task has a unique report location.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsAppBaseDirProvider.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsAppBaseDirProvider.groovy
@@ -1,0 +1,50 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.gradle.plugin.core
+
+import groovy.transform.CompileStatic
+
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.process.CommandLineArgumentProvider
+
+import grails.util.BuildSettings
+
+/**
+ * Provides the {@code -Dgrails.build.base.dir} system property to forked JVM tasks
+ * while using {@link PathSensitivity#RELATIVE} so that the absolute project directory
+ * path does not break build cache relocatability.
+ */
+@CompileStatic
+class GrailsAppBaseDirProvider implements CommandLineArgumentProvider {
+
+    @InputDirectory
+    @PathSensitive(PathSensitivity.RELATIVE)
+    final File appBaseDir
+
+    GrailsAppBaseDirProvider(File appBaseDir) {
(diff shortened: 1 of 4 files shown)


jdaugherty:
@ribafish Good catch!  Thank you for your contribution. @matrei

matrei (approved):
LGTM! Thank you @ribafish

ribafish:
Added two more commits on top of the overlapping outputs fix:

**Normalize `APP_BASE_DIR` system property** — `configureForkSettings()` sets `BuildSettings.APP_BASE_DIR` to `project.projectDir.absolutePath` as a system property on all `Test` and `JavaExec` tasks. Since system properties are `@Input` on `Test`, this absolute path becomes part of the cache key and prevents cache hits when building from a different directory. Replaced with a `CommandLineArgumentProvider` using `@PathSensitive(RELATIVE)`.

**Make SBOM timestamp reproducible** — The CycloneDX SBOM rewrite uses `buildDate` (defaults to `Instant.now()` without `SOURCE_DATE_EPOCH`) for the metadata timestamp, which also changes the `serialNumber` UUID. Since the SBOM is embedded in every jar, this cascades cache misses through the compile classpath of downstream projects. Fixed by using `Instant.EPOCH` when `SOURCE_DATE_EPOCH` is not set. Release builds continue to use the real build date.

**Remaining: `SourceTask` path sensitivity** — `GroovyCompile`, `Checkstyle`, `CodeNarc`, and `Groovydoc` tasks still have relocatability issues in [experiment 3](https://github.com/gradle/develocity-build-validation-scripts/blob/main/Gradle.md#experiment-3---validate-local-build-caching---different-locations) (validates local build caching by building the same commit from two different checkout directories and verifying cache hits on the second build). This happens because `SourceTask.getSource()` defaults to `@PathSensitive(ABSOLUTE)` in Gradle itself, which cannot be fixed in grails-core. Additionally, since `grails-gradle` is an included build compiled from Groovy, the task implementation bytecode differs between directories (Groovy compiler embeds source paths), causing "Task implementation changed" for all tasks whose class originates from the included build. This is a known limitation of experiment 3 with Groovy-compiled included builds.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsAppBaseDirProvider.groovy
```diff
@@ -0,0 +1,53 @@
...
+ * The directory is marked {@link Internal} rather than {@code @InputDirectory} because
+ * the project directory encompasses task output directories (e.g. {@code build/}).
+ * Declaring it as {@code @InputDirectory} causes Gradle to report implicit dependency
+ * violations between the consuming task (e.g. {@code test}) and every task that writes
+ * into the project directory (e.g. {@code compileIntegrationTestGroovy}, {@code jar}).
+ * The actual inputs that matter for caching (classpath, source sets) are already tracked
+ * by their respective tasks.
+ */
+@CompileStatic
+class GrailsAppBaseDirProvider implements CommandLineArgumentProvider {
```
jdaugherty, line 40: I'd just like to say: this is genius.  I've been trying to figure out a way to fix this and I very much appreciate this solution.

jdaugherty (approved):
@matrei  Can you please check this out one more time?  If you agree, I'll merge.

matrei (approved):
Awesome, thank you!

jdaugherty:
@ribafish I opened https://github.com/gradle/gradle/issues/38395 - apparently sometime between the latest gradle release and gradle 8, this stopped working as i would expect.  =(

PR #15533, opened 2026-03-26
Migrate `GrailsPluginManagerTests` to Spock

matrei:
Replaces legacy JUnit 3 test, which was previously not running, as there is currently no support for JUnit 3 in the build.


matrei:
https://develocity.apache.org/s/wvsh3rbrhbfby/tests/overview?class=org.grails.commons.GrailsPluginManagerTests

sbglasius (approved):
Excellent work

PR #15537, opened 2026-03-30
Fix build cache hits for FindMainClassTask and GroovyPageForkCompileTask

ribafish:
### FindMainClassTask: remove upToDateWhen preventing cache hits
- Remove `outputs.upToDateWhen` closure that checked if the cache file exists — per Gradle docs, returning false prevents both up-to-date checks **and** build cache loading, making `@CacheableTask` ineffective
- Remove redundant manual cache file deletion in the task action — Gradle automatically cleans task outputs before re-execution

Both checks are redundant since Gradle already tracks the `@OutputFile` annotation and knows to re-execute when the output is missing.

### GroovyPageForkCompileTask: fix overlapping outputs
- `compileGroovyPages` and `compileWebappGroovyPages` both wrote to `build/gsp-classes/main`, causing Gradle to detect overlapping outputs and disable caching for both tasks
- Give `compileWebappGroovyPages` its own output directory (`build/gsp-classes/webapp`) and update War/Jar packaging to include both directories
- Runtime classpath is unaffected because `output.dir('gsp-classes')` already registers the parent directory


davydotcom (approved):

matrei (approved):
Thank you @ribafish!

PR #15540, opened 2026-03-31
Add test for GSPs in src/main/webapp

matrei:


jdaugherty (approved):
Thank you! I say we merge since this is just test coverage for the recent PR.

PR #15542, opened 2026-04-01
change url mapping wildcard validation config value and address edge case direct mapping

codeconsole:
Follow up to https://github.com/apache/grails-core/pull/15525

per @matrei 
changed the url mapping wildcard validation config value 
`grails.web.url.mapping.validateWildcards` -> `grail.urlmapping.validateWildcards`

and 

addressed issue where direct mapping should win over wildcard.


sbglasius (approved):

jamesfredley (approved):

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/AbstractGrailsControllerUrlMappings.groovy
```diff
@@ -225,30 +223,32 @@ abstract class AbstractGrailsControllerUrlMappings implements UrlMappings {
...
-            (otherMatches + wildcardActionMatches) as UrlMappingInfo[]
-        } else {
-            (wildcardActionMatches + otherMatches) as UrlMappingInfo[]
+        // When wildcard validation is enabled, promote validated wildcard matches
+        // (e.g., $action? resolving to a real action) only when they have strictly fewer
+        // non-routing URL captures — meaning they matched a more specific URL pattern.
+        // Same wildcard status: preserve URL matcher's original order (stable sort).
+        // When validation is disabled, preserve original URL matcher order entirely.
+        if (validateWildcardMappings) {
+            matches.sort(true) { a, b ->
```
jdaugherty, line 238: @codeconsole  do these get run every time?  Is there a performance concern here?

codeconsole: @jamesfredley yes, typically 2-5 matches for any given URL, so the sort is sorting a tiny list. The `nonRoutingParamCount()` does a `keySet().count { }` iteration over 1-3 params per call. For a 3-element sort, that's ~6 comparisons × ~2 param iterations = ~12 operations. Negligible compared to the HTTP overhead, regex matching in `matchAll()`, and controller dispatch that surround it.

PR #15554, opened 2026-04-04
Spring Boot Follow-ups

jdaugherty:
@jamesfredley your AI was hallucinating on these tests.  The embedded servlet context class is now 'AnnotationConfigServletWebServerApplicationContext'.  The tests pass once this was updated. 

I've also restored the `@CompileStatic` that was removed by your PR in the GrailsApplicationBuilder

### DIFF
--- a/grails-core/build.gradle
+++ b/grails-core/build.gradle
@@ -44,15 +44,12 @@ dependencies {
     implementation 'com.github.ben-manes.caffeine:caffeine'
     api 'org.apache.groovy:groovy'
     api 'org.springframework.boot:spring-boot'
+    api 'org.springframework.boot:spring-boot-web-server'
     api 'org.springframework:spring-core'
     api 'org.springframework:spring-tx'
     api 'org.springframework:spring-beans'
     api 'org.springframework:spring-context'
     api 'org.springframework.boot:spring-boot-autoconfigure'
-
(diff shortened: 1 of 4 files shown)


jdaugherty:
@jamesfredley with the split of the tomcat classes into their own library, since we're forced to use the embedded tomcat, I think we're going to have to add `org.springframework.boot:spring-boot-tomcat` to our generated apps.  The tomcat starter being added as a test dependency is likely why the tests are working in the spring boot 4 branch.  We need to: 

1. Update the tomcat feature so the correct libraries are included since we're not using the starter
2. Remove /disable the undretow feature since undertow is not yet on the servlet 6.1 api so boot can't support it
3. possibly create the new project to make tomcat specific dependencies more clear

grails-core/build.gradle
```diff
@@ -44,16 +44,13 @@ dependencies {
     implementation 'com.github.ben-manes.caffeine:caffeine'
     api 'org.apache.groovy:groovy'
     api 'org.springframework.boot:spring-boot'
+    api 'org.springframework.boot:spring-boot-web-server'
```
jdaugherty, line 47: FYI: Spring-boot-web-server was split from spring-boot.  We need to include it for end apps to work the same since we don't support the other options.

jamesfredley (approved):

PR #15563, opened 2026-04-07
Do not fail mergeTestReports for partial previous test task runs

jdaugherty:
If a user: 
1. runs integrationTest - and it's terminated (OOM, etc) before finish 
2. then runs smokeTest 

The mergeTestReports will fail b/c it's partially run, an error like this will show: 

      Caused by: java.lang.IllegalStateException: Test outputs data file '$PROJECT_DIR/build/test-results/gebTest/binary/output.bin' exists but the index file '$PROJECT_DIR/build/test-results/gebTest/binary/output.bin.idx' does not
        at org.gradle.api.internal.tasks.testing.junit.result.TestOutputStore$Reader.<init>(TestOutputStore.java:231)
        at org.gradle.api.internal.tasks.testing.junit.result.TestOutputStore.reader(TestOutputStore.java:391)
        at org.gradle.api.internal.tasks.testing.junit.result.TestOutputStoreBackedResultsProvider.<init>(TestOutputStoreBackedResultsProvider.java:28)
        at org.gradle.api.internal.tasks.testing.junit.result.BinaryResultBackedTestResultsProvider.<init>(BinaryResultBackedTestResultsProvider.java:27)
        at org.gradle.util.internal.CollectionUtils.collect(CollectionUtils.java:192)
        at org.gradle.api.internal.tasks.testing.LegacyTestReportGenerator.lambda$createAggregateProvider$1(LegacyTestReportGenerator.java:54)
        at org.gradle.api.internal.tasks.testing.LegacyTestReportGenerator.hasResults(LegacyTestReportGenerator.java:71)
        at org.gradle.api.tasks.testing.TestReport.generateReport(TestReport.java:186)
        ... 125 more

This is because the index file isn't generated until after the task finishes.  This PR simply removes these invalid states so the test report task doesn't fail.

### DIFF
--- a/grails-test-examples/test-phases/grails-app/controllers/testphases/UrlMappings.groovy
+++ b/grails-test-examples/test-phases/grails-app/controllers/testphases/UrlMappings.groovy
@@ -25,10 +25,6 @@ class UrlMappings {
             constraints {
             }
         }
-
-        '/'(view: '/index')
-        '500'(view: '/error')
-        '404'(view: '/notFound')
     }
 
 }
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/TestPhasesGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/TestPhasesGradlePlugin.groovy
@@ -25,6 +25,8 @@ import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ConfigurationContainer
+import org.gradle.api.file.Directory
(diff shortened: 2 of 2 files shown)


grails-test-examples/test-phases/grails-app/controllers/testphases/UrlMappings.groovy
```diff
@@ -25,10 +25,6 @@ class UrlMappings {
             constraints {
             }
         }
-
```
jdaugherty, line 28: FYI: This may be a regression with 7.1.  There was a swallowed exception I noticed while working on this change because these views didn't exist in this project.  It may be related to Scott's wildcard mapping change.  Although, I think it doesn't have a real impact so it's probably ok.

matrei (approved):

PR #15564, opened 2026-04-08
fix 4x exception logging

codeconsole:
## Fix duplicate stack trace logging in DefaultStackTraceFilterer                                                                                                                                                                            
                                                                                                                                      
  When an exception occurs in a Grails controller, the same error is logged multiple times to the `StackTrace` logger. For a typical  
  `InvocationTargetException` wrapping a `RuntimeException`, this produces 3-4 separate ERROR log entries for a single request:
                                                                                                                                      
  1. `StackTrace` logger: Full stack trace of the wrapper exception
  2. `StackTrace` logger: Full stack trace of the root cause
  3. `StackTrace` logger: Full stack trace of the wrapper again (condensed)                                                           
  4. `GrailsExceptionResolver` logger: The exception with request context
                                                                                                                                      
  Entries 1-3 all come from `DefaultStackTraceFilterer.filter()`. The `filter(Throwable source, boolean recursive)` method walks the  
  cause chain and calls `filter(Throwable source)` for each throwable. But `filter(Throwable source)` logs the full stack trace via   
  `STACK_LOG.error(FULL_STACK_TRACE_MESSAGE, source)` every time it trims a trace — so each cause in the chain gets its own log entry.
                                                                  
  This is compounded by the fact that `GrailsExceptionResolver.logStackTrace()` then logs the exception a final time with request     
  context. The filterer and the resolver were written independently — the filterer doesn't know the resolver will also log, and the
  resolver doesn't know the filterer already logged.                                                                                  
                                                                  
  ### Fix

  Extract the stack trace trimming logic into a private `doFilter(Throwable)` method that filters without logging. The recursive walk 
  calls `doFilter()` for each cause in the chain. The public `filter(Throwable source)` method calls `doFilter()` and then logs the
  full stack trace **once** for the top-level exception (which naturally includes its causes in the log output).                      
                                                                  
  This reduces the `StackTrace` logger output from N entries (one per cause) to 1 entry, while preserving all existing behavior:      
   
  - Stack traces are still recursively filtered for all causes                                                                        
  - The full unfiltered trace is still logged before trimming (for debugging)
  - The `StackTraceFilterer` interface is unchanged                                                                                   
  - Custom `GrailsExceptionResolver` subclasses are unaffected
  - All existing tests pass without modification 

### The name of this class is DefaultStackTraceFilterer, should the intent not be to filter stack traces?


## Exception Log Context — User and IP Address                                                                                        
                                                       
  `GrailsExceptionResolver` now appends a single parenthesised clause of per-request context to its log headline:                       
                                                                                                                                        
  ```                                                                                                                                   
  RuntimeException occurred when processing request: [GET] /admin/users/42 (ip: 198.51.100.42, user: alice)                             
  ```                                                                                                                                   
  
  logAuditor is **enabled by default**. When neither applies to a given request, the headline collapses to the pre-7.1 form.         
                                                                              
  ### `grails.exceptionresolver.logAuditor` _(default: `true`)_                                                                         
                                                                              
  Appends `user: <value>` when the application has registered an `AuditorAware` bean (the same bean used to populate `@CreatedBy` /     
  `@LastModifiedBy`). The appended value is `toString()` of whatever the bean returns — typically a username or numeric id. Applications
   without an `AuditorAware` bean see no change. Discovery is reflective, so `grails-web-mvc` gains no new dependency on                
  `grails-datamapping-core`.                                                  

  Opt out:

  ```yaml
  grails:
      exceptionresolver:
          logAuditor: false
  ```

  ### `grails.exceptionresolver.logRemoteAddr` _(default: `false`)_                                                                      
  
  Appends `ip: <value>` using `HttpServletRequest.getRemoteAddr()`. Behind a reverse proxy, set `server.forward-headers-strategy:       
  native` so Tomcat's `RemoteIpValve` rewrites `remoteAddr` from `X-Forwarded-For` (the valve only trusts the header when the immediate
  peer is in the configured trusted-proxy list — spoofed headers on direct requests are ignored).                                       
                                                                              
  Opt in:

  ```yaml
  grails:
      exceptionresolver:
          logRemoteAddr: true
  ```                                                                                                                                   
  
  #### Custom resolution                                                                                                                
                                                                              
  Subclasses can override the protected `resolveRemoteAddr(HttpServletRequest)` to plug in a different strategy. Returning `null` or an 
  empty string omits the `ip:` entry.
                                                                                                                                        
  ```java                                                                     
  public class MyExceptionResolver extends GrailsExceptionResolver {
      @Override
      protected String resolveRemoteAddr(HttpServletRequest request) {
          return myCustomIpExtraction(request);                                                                                         
      }
  }                                                                                                                                     
  ```                                                                         

  ### Log format impact

  Any log parser, alert rule, or aggregation query that keys off the exact exception headline needs to tolerate the optional ` (ip:     
  <value>, user: <value>)` suffix. Both features can be disabled together to restore the previous format.


jdaugherty:
We need tests to accept a change like this.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java
```diff
@@ -90,27 +90,38 @@ public Throwable filter(Throwable source, boolean recursive) {
...
-            }
-
-            // Only trim the trace if there was some application trace on the stack
-            // if not we will just skip sanitizing and leave it as is
-            if (!newTrace.isEmpty()) {
-                // We don't want to lose anything, so log it
+            boolean modified = doFilter(source);
+            if (modified) {
+                // Log the full stack trace once for the top-level exception (includes causes)
                 STACK_LOG.error(FULL_STACK_TRACE_MESSAGE, source);
```
jdaugherty: Isn't this a different meaning now?  Before it would have been the full stack trace while now it's the filtered.  Wouldn't a better solution be to set the logging to trace or something for these detailed stack traces ?  Instead of this change why not disable STACK_LOG if you don't want this output?

jdaugherty:
The more I have thought about this, the more I realize this is not a bug fix.  This is a feature - we are not trimming stacktraces that got longer due to groovy 4. We are instead hiding traces that get logged multiple times.  I think the original behavior needs to stay and we need to make the new feature opt in.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java
codeconsole: @jdaugherty sensible defaults.  logging a stracktrace 4x that has no useful information (especially with the groovy 4 clutter) quickly leads to a scenario where you can run out of disk space fast.  

I have yet to see an example where the 4x logging provides anything useful.  do you have one?

matrei:
I would probably label this as a bug. Or is this behavior intentional?

This is the log output when visiting `stackTrace/show`:
```groovy
package com.example

class StackTraceController {

    def show() {
        display()
    }

    private void display() {
        notHere
    }
}
```

```console
>                                                                            <
>                            ____           _ _                              <
>                           / ___|_ __ __ _(_) |___                          <
>                          | |  _| '__/ _` | | / __|                         <
>                          | |_| | | | (_| | | \__ \                         <
>                           \____|_|  \__,_|_|_|___/                         <
>                          https://grails.apache.org                         <
>                                                                            <

2026-04-09T13:48:34.488+02:00 ERROR 2325051 --- [p1601249101-110] StackTrace                               : Full Stack Trace:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614)
	at org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:751)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1622)
	at org.eclipse.jetty.ee10.websocket.servlet.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:195)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.CharacterEncodingFilter.doFilterInternal(CharacterEncodingFilter.java:201)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.ServerHttpObservationFilter.doFilterInternal(ServerHttpObservationFilter.java:110)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.CorsFilter.doFilterInternal(CorsFilter.java:91)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$MappedServlet.handle(ServletHandler.java:1555)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.dispatch(ServletChannel.java:822)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.handle(ServletChannel.java:438)
	at org.eclipse.jetty.ee10.servlet.ServletHandler.handle(ServletHandler.java:470)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:575)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.handle(SessionHandler.java:717)
	at org.eclipse.jetty.server.handler.ContextHandler.handle(ContextHandler.java:1102)
	at org.eclipse.jetty.server.Handler$Wrapper.handle(Handler.java:740)
	at org.eclipse.jetty.server.handler.EventsHandler.handle(EventsHandler.java:81)
	at org.eclipse.jetty.server.Server.handle(Server.java:182)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:721)
	at org.eclipse.jetty.server.internal.HttpConnection.onFillable(HttpConnection.java:416)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:322)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:981)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1211)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1166)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: com.example.StackTraceController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at com.example.StackTraceController.propertyMissing(StackTraceController.groovy)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:343)
	at groovy.lang.MetaClassImpl.invokeMissingProperty(MetaClassImpl.java:852)
	at groovy.lang.MetaClassImpl$11.getProperty(MetaClassImpl.java:2181)
	at org.codehaus.groovy.runtime.callsite.GetEffectivePogoPropertySite.getProperty(GetEffectivePogoPropertySite.java:90)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:341)
	at com.example.StackTraceController.display(StackTraceController.groovy:10)
	at com.example.StackTraceController.show(StackTraceController.groovy:6)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 55 common frames omitted

2026-04-09T13:48:34.498+02:00 ERROR 2325051 --- [p1601249101-110] StackTrace                               : Full Stack Trace:

groovy.lang.MissingPropertyException: No such property: notHere for class: com.example.StackTraceController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at com.example.StackTraceController.propertyMissing(StackTraceController.groovy)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:343)
	at groovy.lang.MetaClassImpl.invokeMissingProperty(MetaClassImpl.java:852)
	at groovy.lang.MetaClassImpl$11.getProperty(MetaClassImpl.java:2181)
	at org.codehaus.groovy.runtime.callsite.GetEffectivePogoPropertySite.getProperty(GetEffectivePogoPropertySite.java:90)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:341)
	at com.example.StackTraceController.display(StackTraceController.groovy:10)
	at com.example.StackTraceController.show(StackTraceController.groovy:6)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614)
	at org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:751)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1622)
	at org.eclipse.jetty.ee10.websocket.servlet.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:195)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.CharacterEncodingFilter.doFilterInternal(CharacterEncodingFilter.java:201)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.ServerHttpObservationFilter.doFilterInternal(ServerHttpObservationFilter.java:110)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.springframework.web.filter.CorsFilter.doFilterInternal(CorsFilter.java:91)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$MappedServlet.handle(ServletHandler.java:1555)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.dispatch(ServletChannel.java:822)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.handle(ServletChannel.java:438)
	at org.eclipse.jetty.ee10.servlet.ServletHandler.handle(ServletHandler.java:470)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:575)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.handle(SessionHandler.java:717)
	at org.eclipse.jetty.server.handler.ContextHandler.handle(ContextHandler.java:1102)
	at org.eclipse.jetty.server.Handler$Wrapper.handle(Handler.java:740)
	at org.eclipse.jetty.server.handler.EventsHandler.handle(EventsHandler.java:81)
	at org.eclipse.jetty.server.Server.handle(Server.java:182)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:721)
	at org.eclipse.jetty.server.internal.HttpConnection.onFillable(HttpConnection.java:416)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:322)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:981)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1211)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1166)
	at java.base/java.lang.Thread.run(Thread.java:1583)

2026-04-09T13:48:34.502+02:00 ERROR 2325051 --- [p1601249101-110] StackTrace                               : Full Stack Trace:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:751)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1622)
	at org.eclipse.jetty.ee10.websocket.servlet.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:195)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$MappedServlet.handle(ServletHandler.java:1555)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.dispatch(ServletChannel.java:822)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.handle(ServletChannel.java:438)
	at org.eclipse.jetty.ee10.servlet.ServletHandler.handle(ServletHandler.java:470)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:575)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.handle(SessionHandler.java:717)
	at org.eclipse.jetty.server.handler.ContextHandler.handle(ContextHandler.java:1102)
	at org.eclipse.jetty.server.Handler$Wrapper.handle(Handler.java:740)
	at org.eclipse.jetty.server.handler.EventsHandler.handle(EventsHandler.java:81)
	at org.eclipse.jetty.server.Server.handle(Server.java:182)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:721)
	at org.eclipse.jetty.server.internal.HttpConnection.onFillable(HttpConnection.java:416)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:322)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:981)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1211)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1166)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: com.example.StackTraceController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at com.example.StackTraceController.display(StackTraceController.groovy:10)
	at com.example.StackTraceController.show(StackTraceController.groovy:6)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 44 common frames omitted

2026-04-09T13:48:34.524+02:00 ERROR 2325051 --- [p1601249101-110] o.g.web.errors.GrailsExceptionResolver   : MissingPropertyException occurred when processing request: [GET] /stackTrace/show
No such property: notHere for class: com.example.StackTraceController. Stacktrace follows:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at org.eclipse.jetty.ee10.servlet.ServletHolder.handle(ServletHolder.java:751)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1622)
	at org.eclipse.jetty.ee10.websocket.servlet.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:195)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.FilterHolder.doFilter(FilterHolder.java:205)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1594)
	at org.eclipse.jetty.ee10.servlet.ServletHandler$MappedServlet.handle(ServletHandler.java:1555)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.dispatch(ServletChannel.java:822)
	at org.eclipse.jetty.ee10.servlet.ServletChannel.handle(ServletChannel.java:438)
	at org.eclipse.jetty.ee10.servlet.ServletHandler.handle(ServletHandler.java:470)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:575)
	at org.eclipse.jetty.ee10.servlet.SessionHandler.handle(SessionHandler.java:717)
	at org.eclipse.jetty.server.handler.ContextHandler.handle(ContextHandler.java:1102)
	at org.eclipse.jetty.server.Handler$Wrapper.handle(Handler.java:740)
	at org.eclipse.jetty.server.handler.EventsHandler.handle(EventsHandler.java:81)
	at org.eclipse.jetty.server.Server.handle(Server.java:182)
	at org.eclipse.jetty.server.internal.HttpChannelState$HandlerInvoker.run(HttpChannelState.java:721)
	at org.eclipse.jetty.server.internal.HttpConnection.onFillable(HttpConnection.java:416)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:322)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:981)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1211)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1166)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: com.example.StackTraceController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at com.example.StackTraceController.display(StackTraceController.groovy:10)
	at com.example.StackTraceController.show(StackTraceController.groovy:6)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 44 common frames omitted

```

matrei:
This is the log output with these changes:
```console
>                                                                            <
>                            ____           _ _                              <
>                           / ___|_ __ __ _(_) |___                          <
>                          | |  _| '__/ _` | | / __|                         <
>                          | |_| | | | (_| | | \__ \                         <
>                           \____|_|  \__,_|_|_|___/                         <
>                          https://grails.apache.org                         <
>                                                                            <
            grails-test-examples-app5: 0.1 | JVM: BellSoft 17.0.18            
        Grails: 7.1.0-SNAPSHOT | Groovy: 4.0.31 | Spring Boot: 3.5.13         
                                Spring: 6.2.17                                

2026-04-09 15:32:18.490 ERROR --- [nio-8080-exec-1] StackTrace                               : Full Stack Trace:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: app5.TargetController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at app5.TargetController.display(TargetController.groovy:41)
	at app5.TargetController.show(TargetController.groovy:37)
	... 13 common frames omitted

2026-04-09 15:32:18.517 ERROR --- [nio-8080-exec-1] o.g.web.errors.GrailsExceptionResolver   : MissingPropertyException occurred when processing request: [GET] /target/show
No such property: notHere for class: app5.TargetController. Stacktrace follows:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: app5.TargetController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at app5.TargetController.display(TargetController.groovy:41)
	at app5.TargetController.show(TargetController.groovy:37)
	... 13 common frames omitted
```

jdaugherty:
I think the intent is to disable the logging on this filter, yes?  Then you don't see the the full stacktraces.

jdaugherty:
Why not change the default configuration to just hide logs from grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java then there wouldn't be a duplicate log right? We also appear to have removed the cut off package, which means it could filter our own package

matrei:
> Why not change the default configuration to just hide logs from grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java then there wouldn't be a duplicate log right?

Yes, that might be one option/solution!
But I wonder if logging the stacktrace 4 times is the right behavior?

```xml
<logger name="StackTrace" level="OFF" />
```

```console
>                                                                            <
>                            ____           _ _                              <
>                           / ___|_ __ __ _(_) |___                          <
>                          | |  _| '__/ _` | | / __|                         <
>                          | |_| | | | (_| | | \__ \                         <
>                           \____|_|  \__,_|_|_|___/                         <
>                          https://grails.apache.org                         <
>                                                                            <
            grails-test-examples-app5: 0.1 | JVM: BellSoft 17.0.18            
        Grails: 8.0.0-SNAPSHOT | Groovy: 4.0.31 | Spring Boot: 3.5.13         
                                Spring: 6.2.17                                

2026-04-09 16:33:53.713 ERROR --- [nio-8080-exec-1] o.g.web.errors.GrailsExceptionResolver   : MissingPropertyException occurred when processing request: [GET] /target/show
No such property: notHere for class: app5.TargetController. Stacktrace follows:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:215)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:192)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:110)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:80)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:70)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: app5.TargetController
	at grails.artefact.gsp.TagLibraryInvoker$Trait$Helper.propertyMissing(TagLibraryInvoker.groovy:125)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at app5.TargetController.display(TargetController.groovy:41)
Caused by: groovy.lang.MissingPropertyException: No such property: notHere for class: app5.TargetController

	at app5.TargetController.show(TargetController.groovy:37)
	... 13 common frames omitted
```

matrei:
Curiosly enough, setting the system property `grails.full.stacktrace` to `true` also silences the log output from `DefaultStackTraceFilterer`.

matrei:
Hmm, it looks to me like the console logging from `DefaultStackTraceFilterer` is a side effect of preparing the stack trace for output in the GSP view.

So, if that is the case, disabling logging output for `DefaultStackTraceFilterer` in application logback.xml seems like the most correct option to fix this bug.

```xml
<logger name="StackTrace" level="OFF" />
```

However, In that case I do think the logger should have a more descriptive name, like `ViewStackTrace`.

jdaugherty:
Yeah, I think this is just "debug" output in case something is being truncated.  I think instead of making this change we should just set `<logger name="StackTrace" level="OFF" />` by default.  Then if people need the more detailed loggign, they can still obtain it.

codeconsole:
> Yeah, I think this is just "debug" output in case something is being truncated. I think instead of making this change we should just set `<logger name="StackTrace" level="OFF" />` by default. Then if people need the more detailed loggign, they can still obtain it.

I am fully aware that `<logger name="StackTrace" level="OFF" />`.  Why would you ever want to force that requirement?

What other framework defaults to printing a stack trace 4x? Micronaut? Spring Boot? Anything??  
We need sensible defaults.  If someone really wants to eat up their logs, they can just create their own 

```
public class MultiLoggingExceptionResolver extends GrailsExceptionResolver {
// code displays exception 4 times.
}
```

and register it

@jdaugherty, in @matrei examples above, what information does the 4x version have that the 1x version does not?

Why would we want to silence logs from a class named DefaultStackTraceFilterer? Isn't the job of a stack trace filter to filter stack traces instead of replicate them?

jdaugherty:
With the forward and async actions I suspect its important.  Thats probably the only real reason to turn it on.  Sensible default would be to configure logging with it just off.  It will still log in their error controller - the stacktrace logger is just for details.

codeconsole:
@jdaugherty there is no reason for 4 exceptions. 1 of which is obviously a bug.  I don't think adding `<logger name="StackTrace" level="OFF" />` to every generated grails app is a good idea. It just adds confusion.

What about just adding a configuration to opt in the stack trace display?  The last change introduces:
```groovy
   /**
     * Whether the exception resolver should also emit the exception on the separate
     * {@code StackTrace} logger in addition to its own request-context log entry.
     * Defaults to {@code false}; set to {@code true} to restore the historical two-logger
     * behaviour, which allows routing the trace to a separate appender via logback config.
     */
    String SETTING_LOG_FULL_STACKTRACE = 'grails.exceptionresolver.logFullStackTrace'
```

so it is opt in instead of opt out.

codeconsole:
@jdaugherty @matrei if we want to make this more useful, we could log the spring security user id and ip address with the exception.  This is what I do on my own apps.    Could just add some logic to detect if Spring Security is configured and then display the user id and ip address.  We already display the request url.   Can provide an opt out option as well.

jdaugherty:
I think this requires more discussion and we said we were going to call a vote yesterday.  I'm going to proceed with a vote and we can discuss this in the weekly meeting for Grails 8 or via a patch release for 7.1.

codeconsole:
@jdaugherty that's fine, this can go in a patch release, but what's the hold up? did the last commits not address your concerns?

jdaugherty:
Since there was push back, I just wanted to add this to the weekly meeting.  I assume we'll merge it this week if everyone agrees.

codeconsole:
@matrei @jdaugherty  I added user id and ip address resolution so stack traces can be traced to the request logs.  please take a look.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java
```diff
@@ -81,32 +89,32 @@ public Throwable filter(Throwable source, boolean recursive) {
         if (recursive) {
```
sanjana2505006, line 89: I noticed that `filter(source)` is called again after the cause chain loop. Since the loop already filters the `source` throwable (when `current == source`), is the second call on line 89 necessary?

codeconsole: @sanjana2505006 that's true about the old code, but after this PR, the early return makes the two calls mutually exclusive now.

jdaugherty:
@codeconsole I'll take a look at this tonight and try to get back to you on the review.  Thank you for your patience.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java
```diff
@@ -81,32 +89,32 @@ public Throwable filter(Throwable source, boolean recursive) {
...
+        if (newTrace.isEmpty()) {
+            // filter with no cut-off so at least there is some trace
+            newTrace = filterTraceWithCutOff(trace, null);
+        }
 
-            // Only trim the trace if there was some application trace on the stack
-            // if not we will just skip sanitizing and leave it as is
-            if (!newTrace.isEmpty()) {
-                // We don't want to lose anything, so log it
-                STACK_LOG.error(FULL_STACK_TRACE_MESSAGE, source);
```
jdaugherty, line 105: We discussed this and we think this should stay with a configuration setting.

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
```diff
@@ -36,7 +36,7 @@ class GrailsVersionSpec extends Specification {
         "3.2.0"         | true
         "3.1.0"         | true
         "3.3.0"         | true
-        "7.1.0"         | false
+        "7.1.0"         | true
```
jdaugherty: Can you please update this branch? I've fixed this by increasing the value, we need a "false" case for this test.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
```diff
@@ -261,6 +266,70 @@ protected void logStackTrace(Exception e, HttpServletRequest request) {
...
+     * Resolves the client address to include in the exception log headline. The default
+     * returns {@link HttpServletRequest#getRemoteAddr()} — the container's view of the
+     * TCP peer, which reflects forwarded-header handling only when the servlet container
+     * is configured to trust a proxy chain (for example Spring Boot's
+     * {@code server.forward-headers-strategy}). Subclasses can override this to apply a
+     * different resolution strategy; the returned value (null or empty to omit) is
+     * appended verbatim as {@code ip: <value>}.
+     */
+    protected String resolveRemoteAddr(HttpServletRequest request) {
+        return request.getRemoteAddr();
```
jdaugherty, line 335: This can be wrong for proxy based solutions.  This can also be considered PII in some areas so I think we may want to disable this by default.

jdaugherty (dismissed):

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
codeconsole: done

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
codeconsole: no defaults to false

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java
codeconsole: Consulted AI and it thinks it is a really bad idea:

can you just introduce another config value and default it to false?

It would work, but the mess comes back for opt-in users. Let me trace it.

  ## Setup                                                                                                                            
   
  - Keep `logFullStackTrace` where it is (resolver emits once per request exception when on).                                         
  - Add `logFullStackTraceOnFilter` (default `false`). When `true`, `filter(Throwable)` emits to `STACK_LOG` as it did pre-PR.      
                                                                                                                                      
  ## Output matrix for a 3-cause request exception                                                                                  
                                                                                                                                      
  | logFullStackTrace | logFullStackTraceOnFilter | STACK_LOG entries | Resolver entries | Total |                                    
  |---|---|---|---|---|
  | `false` (default) | `false` (default) | 0 | 1 | 1 |                                                                               
  | `true` | `false` | 1 | 1 | 2 |                                                                                                    
  | `false` | `true` | 3 (one per cause via filter loop) | 1 | 4 |
  | `true` | `true` | 4 (1 from resolver + 3 from filter loop) | 1 | 5 |                                                              
                                                                                                                                    
  The bottom two rows are exactly the per-cause duplication the PR fixes. The user who opts into `logFullStackTraceOnFilter` for      
  non-resolver coverage pays for it with 3+ `STACK_LOG` entries every time a request exception happens.                               
   
  ## Non-resolver sanitize path                                                                                                       
                                                                                                                                    
  | logFullStackTraceOnFilter | STACK_LOG entries from one `sanitizeRootCause(ex)` call |                                             
  |---|---|
  | `false` | 0 |                                                                                                                     
  | `true` | 1 |                                                                                                                    

  This is the only clean win. But it only kicks in when the flag is on, at which point the user is also paying the 4-entry penalty on 
  every request exception.                                                                                                            
   
  ## Plus the layering problem                                                                                                        
                                                                                                                                    
  `DefaultStackTraceFilterer` lives in `grails-bootstrap`, which doesn't depend on `grails-core` where `Config` lives. Options:
                                                                                                                                      
  1. **Pull `Config` into bootstrap** — module cycle.
  2. **Setter on the filterer** — resolver reads the flag and calls `filterer.setLogFullStackTraceOnFilter(flag)` at construction.    
  Every other caller (`new DefaultStackTraceFilterer()` in `TestStacktraceSanitizer`, user subclasses, plugin code) defaults to       
  `false` because nobody wires the flag for them.                                                                                     
  3. **System property** (`Boolean.getBoolean("grails.full.stacktrace.onFilter")`) — matches the existing                             
  `SYS_PROP_DISPLAY_FULL_STACKTRACE` pattern in this file. Avoids the module coupling. Costs YAML-config discoverability.             
   
  Option 3 is the cleanest plumbing-wise but means users opt in via JVM flag rather than `application.yml`. A bit inconsistent with   
  the other `grails.exceptionresolver.*` settings, all of which live in YAML.                                                         
   
  ## Cleaner alternative                                                                                                              
                                                                              
  Document one line in the upgrade guide for teams relying on non-resolver sanitize coverage:

  ```java
  // before: relied on side-effect log
  Throwable cleaned = GrailsUtil.sanitizeRootCause(ex);                                                                               
  log.error("job failed", cleaned);
                                                                                                                                      
  // after: explicit                                                          
  Throwable cleaned = GrailsUtil.sanitizeRootCause(ex);
  DefaultStackTraceFilterer.STACK_LOG.error(StackTraceFilterer.FULL_STACK_TRACE_MESSAGE, cleaned);                                    
  log.error("job failed", cleaned);                                                                                                   
```                                                                                                                                      
  No new flag. No layering workaround. No duplicate-per-cause regression for users who happen to need non-resolver coverage.          
                                                                              
  My call                                                                                                                             
                                                                              
  If you want a concrete non-resolver path you want covered, better the two-line migration. If the project wants blanket pre-PR   compatibility, the second flag works but I'd go the system-property route (option 3) to avoid the layering mess, and I'd document  clearly that turning it on without also turning logFullStackTrace off produces N+1 entries per exception.                           
                                                                              
  Honestly though: the second flag is solving the problem by re-exposing the old mechanism. The cleaner fix is the two-line code  change at the actual call sites that want the behavior. That's explicit, discoverable at the code site, and doesn't involve tuning two interacting flags.

**Review by sbglasius — APPROVED** [review]

File: grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java (hunk 3 above)
Comment (jdaugherty) [reply grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/DefaultStackTraceFilterer.java:105]: a config setting preserves both and we agreed in the weekly that this was the way forward, I am -1 on this without this

**Review by jdaugherty — COMMENTED** [review]
I'm good with this change now, but I'm not sure if it fits in 7.1.x.  7.2.x is an immediate merge for me.  @matrei  what are your thoughts?

File: grails-web-mvc/src/main/groovy/org/grails/web/errors/AuditorAwareLookup.java
Comment (matrei) [inline grails-web-mvc/src/main/groovy/org/grails/web/errors/AuditorAwareLookup.java:1]: I found the logic a bit hard to grasp as the `resolve()` method always returns `true`.
Using Groovy we could also abstract away some boilerplate:
```groovy
package org.grails.web.errors

import java.lang.reflect.Method

import groovy.transform.PackageScope
import groovy.util.logging.Slf4j

import org.springframework.beans.BeansException
import org.springframework.context.ApplicationContext
import org.springframework.util.ClassUtils

/**
 * Optional integration with the GORM {@code AuditorAware} bean. When the audit API
 * class is on the classpath and a bean is registered, {@link AuditorAwareLookup#getCurrentAuditor()}
 * returns its result so the exception resolver can reuse the same "current user"
 * resolution that {@code @CreatedBy} uses. The lookup is reflective so that this
 * module does not compile-time depend on {@code grails-datamapping-core}.
 *
 * <p>Any failure — missing class, missing bean, invocation error — resolves to an
 * empty {@code Optional}. Exception logging must never be blocked by a broken
 * auditor lookup.</p>
 */
@Slf4j
@PackageScope
@CompileStatic
class AuditorAwareLookup {

    private static final String AUDITOR_AWARE_CLASS =
            'org.grails.datastore.gorm.timestamp.AuditorAware'

    private final ApplicationContext applicationContext

    @Lazy volatile
    private AuditorLookup auditorLookup = lookupAuditor()

    AuditorAwareLookup(ApplicationContext applicationContext) {
        this.applicationContext = applicationContext
    }

    Optional<?> getCurrentAuditor() {
        auditorLookup.currentAuditor
    }

    private AuditorLookup lookupAuditor() {
        try {
            def classLoader = applicationContext?.classLoader
            if (!classLoader || !ClassUtils.isPresent(AUDITOR_AWARE_CLASS, classLoader)) {
                return AuditorLookup.empty()
            }
            def type = ClassUtils.forName(AUDITOR_AWARE_CLASS, classLoader)
            try {
                return new AuditorLookup(
                        applicationContext.getBean(type),
                        type.getMethod('getCurrentAuditor')
                )
            }
            catch (BeansException ignored) {
                return AuditorLookup.empty()
            }
        }
        catch (Throwable t) {
            log.debug('AuditorAware integration unavailable', t)
            return AuditorLookup.empty()
        }
    }

    @CompileStatic
    private static class AuditorLookup {

        private final Object auditorAwareBean
        private final Method getCurrentAuditorMethod

        AuditorLookup(Object auditorAwareBean, Method getCurrentAuditorMethod) {
            this.auditorAwareBean = auditorAwareBean
            this.getCurrentAuditorMethod = getCurrentAuditorMethod
        }

        static AuditorLookup empty() {
            new AuditorLookup(null, null)
        }

        Optional<?> getCurrentAuditor() {
            if (auditorAwareBean == null || getCurrentAuditorMethod == null) {
                return Optional.empty()
            }
            try {
                def result = getCurrentAuditorMethod.invoke(auditorAwareBean)
                return result instanceof Optional ? (Optional<?>) result : Optional.empty()
            }
            catch (ReflectiveOperationException e) {
                log.debug('AuditorAware#getCurrentAuditor invocation failed', e)
                return Optional.empty()
            }
        }
    }
}
```

grails-core/src/main/groovy/grails/config/Settings.groovy
```diff
@@ -292,6 +292,41 @@ interface Settings {
      * The parameters to exclude from logging
      */
     String SETTING_EXCEPTION_RESOLVER_PARAM_EXCLUDES = 'grails.exceptionresolver.params.exclude'
+    /**
+     * Whether the exception resolver should also emit the exception on the separate
+     * {@code StackTrace} logger in addition to its own request-context log entry.
+     * Defaults to {@code false}; set to {@code true} to restore the historical two-logger
+     * behaviour, which allows routing the trace to a separate appender via logback config.
+     */
+    String SETTING_LOG_FULL_STACKTRACE = 'grails.exceptionresolver.logFullStackTrace'
```
matrei, line 301: It would have been nice with `grails.exceptionResolver.logFullStackTrace`, but I agree with keeping consistency.

grails-doc/src/en/guide/conf/config/logging/loggingFullStackTraces.adoc
```diff
@@ -0,0 +1,212 @@
...
+calls `getCurrentAuditor()` on that bean and includes ` user: <value>` in the clause, reusing the same auditor
+resolution that populates `@CreatedBy` and `@LastModifiedBy` fields. This avoids registering a second "current
+user" bean purely for logging.
+
+The appended value is the `toString()` of whatever the `AuditorAware` implementation returns — typically a
+username string or a numeric id, depending on the type parameter chosen for the bean. If no `AuditorAware` bean
+is registered, or if it returns `Optional.empty()` for the current request (for example, an unauthenticated
+request), no `user:` entry is emitted.
+
+Enabled by default. To keep user identifiers out of exception logs — for example to satisfy a PII separation
```
matrei: I think this should be disabled by default. Won't we risk logging email addresses otherwise? Same GDPR problem as IP-addresses.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
```diff
@@ -261,6 +266,70 @@ protected void logStackTrace(Exception e, HttpServletRequest request) {
...
+
+    protected boolean shouldLogAuditor() {
+        Config config = grailsApplication != null ? grailsApplication.getConfig() : null;
+        return config != null && config.getProperty(Settings.SETTING_LOG_AUDITOR, Boolean.class, true);
+    }
+
+    protected boolean shouldLogRemoteAddr() {
+        Config config = grailsApplication != null ? grailsApplication.getConfig() : null;
+        return config != null && config.getProperty(Settings.SETTING_LOG_REMOTE_ADDR, Boolean.class, false);
+    }
```
matrei, line 323: Do we need to resolve these on every invocation?

matrei (approved):
Some minor comments.

I think this turned out great @codeconsole, thank you!

grails-web-mvc/src/main/groovy/org/grails/web/errors/AuditorAwareLookup.java
codeconsole: Good catch — the `boolean` return on `resolve()` was dead. Changed it to `void` and gated the caller on `bean == null` instead, which removes the misleading `!resolve() || bean == null` check. See f5d28ec50b.

Appreciate the Groovy rewrite — left it as Java since the rest of `grails-web-mvc/.../errors/*.java` is Java and the reflective lookup doesn't really benefit from `@Lazy` / Groovy sugar here. Happy to revisit if you'd rather standardize on Groovy in this package.

grails-core/src/main/groovy/grails/config/Settings.groovy
codeconsole: Yeah, agreed — I went lowercase to match the existing `grails.exceptionresolver.*` keys on this interface (`grails.exceptionresolver.params.exclude`, `grails.exceptionresolver.logRequestParameters`), so the four new keys (`logFullStackTrace`, `logAuditor`, `logRemoteAddr`, `logFullStackTraceOnFilter`) stay consistent with that prefix. Happy to look at a repo-wide camelCase normalization in a separate PR if we want it.

grails-doc/src/en/guide/conf/config/logging/loggingFullStackTraces.adoc
codeconsole: Good call. Flipped the default to `false` in e4b96a83bc — same PII reasoning as `logRemoteAddr`. Apps that want the appended ` user: <value>` clause now opt in with:

```yaml
grails:
    exceptionresolver:
        logAuditor: true
```

Updated `loggingFullStackTraces.adoc` ("Disabled by default…") and the section 2.12 blurb in `upgrading71x.adoc` to match. With both `logAuditor` and `logRemoteAddr` off-by-default, the headline collapses to its pre-7.1 form unless the app explicitly opts in.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
codeconsole: Fair — config doesn't change at runtime so re-reading it per resolved exception is wasteful. Resolved in e4b96a83bc: added a private `resolveLogFlags()` (double-checked, `volatile boolean logFlagsResolved`) that reads the three properties on first use and caches them into `boolean` fields. The three `shouldLog*` predicates now just call `resolveLogFlags()` and return the cached value, so subsequent exceptions pay only a volatile read.

Kept the methods as `protected` so subclasses overriding any `shouldLog*` predicate bypass the cache entirely, same as before.

Issue #15469, opened 2026-02-26
Migrate configuration metadata to @ConfigurationProperties with annotation processor for Grails 8


jamesfredley:
## Summary

Grails currently hand-maintains `spring-configuration-metadata.json` files across multiple modules. Spring Boot provides an annotation processor (`spring-boot-configuration-processor`) that auto-generates this metadata from `@ConfigurationProperties`-annotated classes at compile time. Adopting this approach would eliminate manual metadata maintenance and prevent config drift.

### Hand-Maintained JSON (prone to drift)
We maintain `spring-configuration-metadata.json` by hand in 9 grails-core modules, grails-spring-security, and asset-pipeline. During the PR #15426 / #1210 work, we discovered **62 properties had drifted** out of sync in Spring Security alone (83 of 145 were present).

### Modules that already have @ConfigurationProperties classes but lack the processor
These 5 modules already define config as POJOs but don't use the annotation processor to auto-generate metadata:

- `grails-databinding` - `DataBindingConfigurationProperties.java`
- `grails-web-url-mappings` - `GrailsCorsConfiguration.groovy`
- `grails-cache` - `CachePluginConfiguration.groovy`
- `grails-views-gson` - `JsonViewConfiguration.groovy`
- `grails-views-markup` - `MarkupViewConfiguration.groovy`

### Modules that use Groovy DSL closures for config
These cannot use the annotation processor without refactoring:

- `grails-core` - config via `application.groovy` merging
- `grails-spring-security` - `DefaultSecurityConfig.groovy` (Groovy closure DSL, 145 properties)
- `grails-web-core` - mixed approach

### Phase 1: Quick wins (add processor to existing @ConfigurationProperties modules)

Add the annotation processor dependency to the 5 modules that already have `@ConfigurationProperties` classes:

```groovy
annotationProcessor "org.springframework.boot:spring-boot-configuration-processor"
```

This would auto-generate the metadata JSON at compile time from the existing annotated classes. Javadoc on fields becomes the property description. Field initializers become the defaults. Zero manual JSON maintenance for these modules.

### Phase 2: Migrate Groovy DSL config to @ConfigurationProperties POJOs

Refactor configuration that currently lives in Groovy DSL closures (like `DefaultSecurityConfig.groovy`) into `@ConfigurationProperties`-annotated Java or Groovy classes. For example:

**Before** (Groovy DSL - cannot be processed):
```groovy
security {
    rememberMe {
        cookieName = 'grails_remember_me'
[...]

PR #15566, opened 2026-04-09
Phase 1: Enable Spring Boot configuration metadata processor for existing @ConfigurationProperties modules (#15469)

vigneshsiva11:
This pull request addresses part of the work for issue number 15469. It makes Spring Boots configuration metadata processor work in some modules that already use the ConfigurationProperties annotation.

Here are the modules that were updated:

- grails-databinding
- grails-web-url-mappings
- grails-cache
- grails-views-gson
- grails-views-markup

These are the changes that were made:

- A new dependency was added:

annotationProcessor "org.springframework.boot:spring-boot-configuration-processor"

- The Grails BOM was applied to the annotation processor configuration. This means the dependency will be resolved correctly when the project is built.

So how does the process work?

When the code is compiled, the Spring Boot configuration processor automatically makes a file called

META-INF/spring-configuration-metadata.json

This keeps the configuration metadata in sync with the source code. It also reduces the need to manually keep metadata files up to date. The default values and property names come from the existing ConfigurationProperties classes.

The above is what was checked to make sure everything works:

- Java 17 was. Set up in the development container
- The Gradle compilation was completed successfully for the updated modules
- The generated metadata was checked in the build output for an example like the grails-databinding module
- 
Some things to note:

This pull request only does the part of the work for issue number 15469.

The configuration metadata for DSL will be done in later parts.

### DIFF
--- a/grails-cache/build.gradle
+++ b/grails-cache/build.gradle
@@ -38,6 +38,8 @@ ext {
 
 dependencies {
     implementation platform(project(':grails-bom'))
+    annotationProcessor platform(project(':grails-bom'))
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
     api project(':grails-core')
     api project(':grails-web-boot')
--- a/grails-databinding/build.gradle
+++ b/grails-databinding/build.gradle
@@ -35,6 +35,8 @@ group = 'org.apache.grails'
 dependencies {
 
     implementation platform(project(':grails-bom'))
+    annotationProcessor platform(project(':grails-bom'))
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
     api project(':grails-core')
     api project(':grails-web-core')
--- a/grails-views-gson/build.gradle
+++ b/grails-views-gson/build.gradle
@@ -36,6 +36,8 @@ ext {
 
 dependencies {
     implementation platform(project(':grails-bom'))
+    annotationProcessor platform(project(':grails-bom'))
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
     api project(':grails-views-core') // Used in public API
     api project(':grails-rest-transforms') // Used in public API
--- a/grails-views-markup/build.gradle
+++ b/grails-views-markup/build.gradle
@@ -36,6 +36,8 @@ ext {
 
 dependencies {
     implementation platform(project(':grails-bom'))
+    annotationProcessor platform(project(':grails-bom'))
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
     api project(':grails-views-core') // Used in public API
     api project(':grails-core') // Used in public API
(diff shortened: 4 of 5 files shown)


jdaugherty:
Hi @vigneshsiva11 did you coordinate this change with @jamesfredley ?  I think he already has some of this work in another PR.

grails-cache/build.gradle
```diff
@@ -37,6 +37,8 @@ ext {
 
 dependencies {
     implementation platform(project(':grails-bom'))
+    annotationProcessor platform(project(':grails-bom'))
```
jdaugherty, line 41: Annotation processors are incompatible with incremental groovy compilation and can break the groovy compiler when used.

vigneshsiva11:
Hi @jdaugherty and @jamesfredley

I want to thank you for taking the time to review my work and for the guidance you provided.

I made some changes to this PR based on the feedback you gave me.
I removed the annotation processor wiring from the modules that are focused on Groovy.
I also made sure that the scope of this change is narrow and only includes databinding for now, which is what we are calling Phase 1.

This change should help with the issue of incremental compilation, and it keeps this PR small and easy to review.

I went ahead. Pushed the updated commit. Then I checked the builds again to ensure everything is working correctly after I made these changes.

If this PR is still getting in the way of the work that @jamesfredley is doing, then I am more than happy to make some adjustments or even close this PR if that is what you would prefer.

jamesfredley:
@vigneshsiva11 Can I get you to rebase this on 8.0.x branch, that is the earliest this can occur and current focus on work.

vigneshsiva11:
Hi @jamesfredley I have updated the Phase 1 branch to work with the "8.0.x" version. Pushed the changes. The base for the pull request is now the "8.0.x" branch. I did a rebase. Then pushed the updated changes to the repository.

jdaugherty:
I'm going to close this work as it's already been started in parallel by @jamesfredley and this implementation uses an annotation processor which we have previously discussed & decided as a non-viable solution.  @vigneshsiva11 It looks like you started this work before consulting with the team and this is one of the issues we have discussed heavily in the weekly meetings.  Can you please be sure we indicate to proceed on this issue before opening PRs further on this matter?

vigneshsiva11:
Thanks for clarifying, @jdaugherty. So, the team's already looked into the annotation processor approach and ruled it out for Grails due to compatibility issues with Groovy and incremental compilation. You've saved me some time by filling me in on the context, especially since this has been discussed internally before.

I shouldn't have invested time in this PR without checking the current direction with the team first; that was my mistake. But I've learned from it, and going forward, I'll make sure to touch base with the team on the issue thread before starting any implementation work on design-sensitive topics.

And I'm still keen to contribute to Grails, particularly when it comes to configuration improvements for Grails 8. If there are specific areas that need help, like the proposed Grails-specific configuration processor, or other tasks that align with the team's direction, I'd be happy to lend a hand. Thanks again for your feedback and guidance.

PR #15567, opened 2026-04-10
8.0.x hibernate7 dev

borinquenkid:
This is a plethora of refactorings and bug fixes


jdaugherty:
I would like to merge this outside of this PR into the branch so we can review the same code in one place - if we merge / review on this PR, then it will be "hidden" on the 8.0.x-hibernate7 PR.  @borinquenkid are you ok if we do this?

PR #15569, opened 2026-04-11
Agents fix and add skills

borinquenkid:
Fixed the format of the SKILL.md and added 3 more
hibernate
styling
testing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,19 @@
+<!--
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -69,6 +73,10 @@ export GRADLE_OPTS="-Xms2G -Xmx5G"
 | **grails-developer** | `.agents/skills/grails-developer/SKILL.md` | Grails 7 apps, GORM, controllers, views |
```
sanjana2505006, line 73: Since this branch (`8.0.x-hibernate7)` is versioned as `8.0.0-SNAPSHOT` and requires JDK 21, we should update this description to refer to Grails 8 instead of Grails 7 to stay consistent with the branch goals.

PR #15571, opened 2026-04-11
Refactor: Centralize Spring Boot configuration metadata setup (Phase 2 of #15469)

vigneshsiva11:
## **Overview**

This change is the step after the first part. It makes the Spring Boot configuration metadata generation better by improving the structure and consistency.

### Key Improvements

* I put the annotation processor configuration in one place

* Removed the setup from each module's build file

* i still only enable it for "grails-databinding."

* Made sure the metadata generation works as it should without adding risk

This part is about making the code better and putting things together. I do not use the processor for things that I did before.

---

## **Problem**

In this part I only enabled configuration metadata generation for "grails-databinding" because of some problems.

* There were risks with incremental compilation

* It was not stable when i used annotation processors for Groovy- modules

But this also caused some issues.

* i had to do the configuration many times

* It was harder to maintain. i could easily make mistakes

### This Change Fixes

* I do not have to do the configuration many times

* I have control over annotation processing

* It is easier to add this feature to things in the future

---

### **Build Logic Changes**

* i made a shared configuration for annotation processors

* i put this logic in a place

### **Modules Updated**

* `grails-databinding`

* Now it uses the shared processor configuration

* It works the same as before

### **Removed**

* i removed the annotation processor setup from each module's build file

### **Key Characteristics**

* i still have to enable annotation processing manually

* It only works for "grails-databinding."

* It does not change how things are compiled or what the output metadata looks like

---

## **Issues**

* This is related to issue #15469

* It is the step, after #15566

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
@@ -23,11 +23,13 @@ import java.nio.charset.StandardCharsets
 import java.util.concurrent.atomic.AtomicBoolean
 
 import groovy.transform.CompileStatic
+import groovy.transform.TypeCheckingMode
 
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.DuplicatesStrategy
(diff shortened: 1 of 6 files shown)


jdaugherty:
We cannot use an annotation processor.  Per previous PR, I'm going to close this.

PR #15574, opened 2026-04-14
Cleanup and align `application.yml` in generated applications

matrei:
Many config values in generated `application.yml` files are redundant as they only restate the default values.
This is creating a lot of confusing boilerplate that's never changed, gets copied around, gets outdated and only leads to confusion.

I guess it has served as some form of documentation of the available configuration options, but I think we should move this to the Guide documentation as is already in the pipeline.

There are still some config values left in the `application.yml` files that actually do not match default values so these are left in the file for the time being.

### DIFF
--- a/grails-doc/src/en/guide/upgrading/upgrading60x.adoc
+++ b/grails-doc/src/en/guide/upgrading/upgrading60x.adoc
@@ -869,3 +869,43 @@ The legacy `org.grails.web.converters.marshaller.json.EnumMarshaller` and `org.g
 ==== 13. Upgrading the Spring Security Plugin
 
 If your application uses the Grails Spring Security plugin, please consult the https://apache.github.io/grails-spring-security/snapshot/guide/index.html#upgrading-from-previous-versions[Upgrading from Previous Versions] section of the Spring Security plugin documentation for specific upgrade instructions related to Grails 7.
+
+==== 14. Changes in CLI- and Forge-generated configuration files
+
+From Grails 7.0.11 onwards, the CLIs (`grails-shell-cli` and `grails-forge-cli`) and the https://start.grails.org[Grails Forge website] no longer include several previously generated properties in `application.yml`.
+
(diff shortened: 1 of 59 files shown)


jdaugherty:
I do not think we can simplify these changes in 7.x.  8.x yes, but only if we add the associated configuration defaults we've previously talked about.

matrei:
> I do not think we can simplify these changes in 7.x. 8.x yes, but only if we add the associated configuration defaults we've previously talked about.

@jdaugherty I have not touched any production code. This is purely generated files from the CLIs. The expected loaded configuration is the same (except for `grails.urlmapping.cache.maxsize: 1000` which defaults to `5000` in code).

jdaugherty:
@matrei the generated code is production code; we've had bugs opened in the past b/c of this and suddenly removing that generated code without documenting the defaults is a major limitation.

matrei:
> @matrei the generated code is production code; we've had bugs opened in the past b/c of this and suddenly removing that generated code without documenting the defaults is a major limitation.

This is not code, it's a configuration file. Bugs are ok, if there is one we will fix it.
We should absolutely document the available configuration options and their defaults.

This cleaned up version of the `application.yml` file is only generated for new Grails apps. No existing apps will be influenced by this change.

jdaugherty:
The problem is we don't have the documentation and if you're converting, the only way to previously get those values was to generate the apps.  I agree we should clean up this up, but not until we have documentation to see what the defaults are.

matrei:
@jdaugherty I have added documentation to the upgrade guide. We also have pending #15426 which will add all configuration properties to the Guide documentation.

jdaugherty:
Related: https://github.com/apache/grails-core/pull/15426

jdaugherty (approved):

jdaugherty:
Per discussion, I approved under the assumption we'll merge the related PR & we'll address this in a better way post 7.1.x (likely 8.x)

jamesfredley (approved):

PR #15576, opened 2026-04-15
Clean up labels used in the forge ui

matrei:
image

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/GrailsGsp.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/GrailsGsp.java
@@ -18,6 +18,7 @@
  */
 package org.grails.forge.feature.view;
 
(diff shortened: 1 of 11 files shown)


jamesfredley (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/GrailsGsp.java
```diff
@@ -51,12 +52,21 @@ public boolean shouldApply(ApplicationType applicationType, Options options, Set
...
     public String getTitle() {
-        return "Grails GSP";
+        return "GSP";
     }
 
+    @NonNull
     @Override
     public String getName() {
-        return "grails-gsp";
+        return "gsp";
```
jdaugherty, line 61: This will technically mismatch but it doesn't appear to break forge, so I'm ok with this.

jdaugherty (approved):
Looks great!

PR #15583, opened 2026-04-18
Native gorm mongodb String to ObjectId conversion

codeconsole:
# GORM MongoDB: `storedAs` Identifier Coercion

## The Problem

Before this change, GORM MongoDB had a silent asymmetry in how it handled identifier types:

| Code path | `String id` declared + `_id: ObjectId` on disk |
|-----------|-------------------------------------------------|
| **Scan read** (`.list()`) | ✅ Decoder falls back via Spring ConversionService |
| **Point lookup** (`get(hex)`) | ❌ Returns `null` — query sends `{_id: "<hex>"}` as BSON String |
| **Update** (`save()`) | ❌ Throws misleading `OptimisticLockingException` |
| **Batch** (`getAll([hex])`, `findAllByIdInList`, `in('id', [...])`) | ❌ Returns empty — `$in` list sent as BSON Strings |

The reader was forgiving, but the query and write paths weren't. This made "declare `String id` on a domain that previously used `ObjectId id`" a silent landmine — scans looked fine, everything else quietly broke.

The ergonomic consequence: teams were forced to either keep `ObjectId id` (which serializes to `{"timestamp":..., "date":...}` in JSON and requires `new ObjectId(...)` at every call site) or do a full data migration of every `_id` value in storage.

## The Fix

Two new knobs that let a domain keep `String id` ergonomics while persisting `_id` as a BSON ObjectId — no data migration required.

### Per-domain mapping option

```groovy
import org.bson.types.ObjectId

class Person {
    String id

    static mapping = {
        id storedAs: ObjectId
    }
}
```

### Global config default

```yaml
grails:
  mongodb:
    stringIdsDefaultStoredAs: objectid   # or 'string' (default — current behavior)
```

Per-domain `storedAs` always wins over the global default. Natural-key domains opt out with `id storedAs: String`.

## What Changed

10 production files across three modules, all coerce between declared type and storage type:

| Layer | File | Change |
|-------|------|--------|
| Core interface | `IdentityMapping.java` | New `default Class<?> getStoredAs() { return null; }` |
| Core property | `Property.groovy` | New `Class<?> storedAs` field |
| Core factory | `MappingFactory.java` | Both `createDefaultIdentityMapping` overloads expose `storedAs` dynamically (composite-key-safe) |
| Mongo config | `MongoSettings.groovy` | New `SETTING_STRING_IDS_DEFAULT_STORED_AS` |
| Mongo settings | `AbstractMongoConnectionSourceSettings.groovy` | New `String stringIdsDefaultStoredAs` field |
| Mongo context | `MongoMappingContext.java` | Reads global default; applies to String-id domains without explicit `storedAs`; warns on unrecognized values |
| Mongo encoder | `IdentityEncoder.groovy` | String → ObjectId on write (with `ObjectId.isValid` guard for natural keys) |
| Mongo query | `MongoQuery.java` | `IdEquals` + `In` handlers coerce to storage type (fixes `get(hex)`, `findAllByIdInList`, criteria `in('id', ...)`) |
| Mongo persister | `MongoCodecEntityPersister.groovy` | Coerces keys in `retrieveEntity` and `retrieveAllEntities` (fixes `get` and `getAll`) |
| Mongo session | `MongoCodecSession.groovy` | Coerces `nativeKey` in update/delete filters (fixes saves and deletes) |

## Why It's Better

- **Ergonomics + performance, not either/or.** Domain code sees `String id` (clean JSON, no `new ObjectId(...)` dance, no `[object Object]` bugs at HTTP boundaries). Storage keeps 12-byte BSON ObjectId `_id` (half the index size of hex strings, embedded creation timestamp, native Mongo sort order).
- **No data migration required.** Legacy documents with `_id: ObjectId(...)` continue to load via the existing decoder fallback; new writes produce BSON ObjectId too — everything stays consistent on disk.
- **The silent failure modes are gone.** `get(hex)`, `findAllByIdInList`, saves, deletes, and criteria in-lists all now match stored ObjectIds when the domain opts in. The misleading `OptimisticLockingException` on save (which previously pointed developers at concurrency rather than the real id-type mismatch) no longer fires.
- **Fully backward compatible.** Default config is `null`/`string` — nothing changes for existing apps unless they explicitly opt in.
- **Natural-key safety.** Combining `generator: 'assigned'` with `storedAs: ObjectId` and a non-hex value (e.g. a slug) falls back to BSON String rather than throwing `IllegalArgumentException` deep in the BSON pipeline.

## Testing

**23 tests across 2 new specs**, all against a real MongoDB testcontainer or real `MongoMappingContext` init:

- `StringIdWithObjectIdStorageSpec` (18 tests) — documents the original bugs, then proves each code path (read, point lookup, save, batch `getAll`, `findAllByIdInList`, criteria `in('id', ...)`, delete, update of legacy ObjectId-`_id` docs, non-hex natural-key fallback, JSON serialization demo).
- `StringIdDefaultStoredAsConfigSpec` (5 tests) — global config plumbing: default off, global on, per-domain override, declared-type filter (ObjectId-id domains ignore the String-id global), unrecognized-value safe fallback.

Full `:grails-data-mongodb-core:test` + `:grails-data-hibernate5-core:test` green. Checkstyle clean on all three modified modules. Three rounds of automated code review completed, each round surfaced real bugs, all fixed.

## Documentation

- `objectMapping/idGeneration.adoc` — new section "Decoupling the Declared Type from the Storage Type" + "Global Default" + "Caveats".
- `gettingStarted/advancedConfig.adoc` — one-line pointer to the new setting.


grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
```diff
@@ -41,6 +41,25 @@ class IdentityEncoder implements PropertyEncoder<Identity> {
...
+                String hex = id.toString()
+                // Guard against natural-key strings accidentally paired with storedAs: ObjectId.
+                // new ObjectId(<non-hex>) throws IllegalArgumentException, which would surface
+                // deep inside the BSON write pipeline. Fall through to writeString for consistency
+                // with the converter-based paths (MongoCodecSession, MongoCodecEntityPersister).
+                if (ObjectId.isValid(hex)) {
+                    writer.writeObjectId(new ObjectId(hex))
+                    return
+                }
+            }
```
sanjana2505006, line 56: I noticed that if `ObjectId.isValid(hex)` is false, it falls through to write a BSON String. While this prevents a crash during save, the `Converter` added in `MongoMappingContext` returns `null` for invalid hex strings. This creates an asymmetry where a document can be saved with a non-hex ID (like a slug) but then becomes unreachable via `Domain.get(id)` because the query will target `{_id: null}`.

Should we ensure the query path also falls back to the original String if it's not a valid hex, to maintain consistency?

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/query/MongoQuery.java
```diff
@@ -138,11 +138,66 @@ public void handle(EmbeddedQueryEncoder queryEncoder, IdEquals criterion, Docume
...
+                        }
+                    }
+                    values = coerced;
+                }
+
+                inQuery.put(IN_OPERATOR, values);
+                String propertyName = getPropertyName(entity, in);
+                query.put(propertyName, inQuery);
+            }
+        });
```
sanjana2505006: The coercion logic for the `In` operator (and `IdEquals`) seems to specifically target the root entity's identity property. Have we considered how this affects queries on associations?

For example, if `Author` has `storedAs: ObjectId`, a query like `Post.findAllByAuthor(authorInstance)` might miss because the `Author` ID in the criteria isn't being coerced to `ObjectId` before the query is sent to MongoDB. We might need to make the coercion more generic for any property that points to an entity with a `storedAs` mapping.

grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
codeconsole: Good catch, thanks... fixed.

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
```diff
@@ -36,7 +36,7 @@ class GrailsVersionSpec extends Specification {
         "3.2.0"         | true
         "3.1.0"         | true
         "3.3.0"         | true
-        "7.1.0"         | false
```
jdaugherty: We should fix this with an arbitrary large value instead of just setting to true.

jdaugherty (commented):
@borinquenkid I know you've recently worked on mongo. Would you mind taking a look at this PR? 

@copilot can you please review this PR too?

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
jdaugherty: I've updated this in the 7.1 branch after seeing it fail on multiple reviews.  You should be able to update your base branch now.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/MongoCodecEntityPersister.groovy
```diff
@@ -175,6 +178,32 @@ class MongoCodecEntityPersister extends ThirdPartyCacheEntityPersister<Object> {
         new Document(AbstractMongoObectEntityPersister.MONGO_ID_FIELD, key)
     }
 
+    /**
+     * Coerce an identifier value to the {@code storedAs} type declared in the entity's id
+     * mapping, so that point-lookup queries target the BSON type actually on disk.
+     * See {@code IdentityMapping#getStoredAs()}.
+     */
+    protected Object coerceIdToStoredType(Object key, PersistentEntity entity) {
```
borinquenkid, line 195: I don't see any test coverage for this

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/query/MongoQuery.java
```diff
@@ -138,11 +138,78 @@ public void handle(EmbeddedQueryEncoder queryEncoder, IdEquals criterion, Docume
                 Object value = criterion.getValue();
                 MappingContext mappingContext = entity.getMappingContext();
                 PersistentProperty identity = entity.getIdentity();
-                Object converted = mappingContext.getConversionService().convert(value, identity.getType());
+                // Prefer the configured storage type ('storedAs' on the id mapping) so query BSON
```
borinquenkid, line 145: Is there a unit test for this?

grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/bugs/StringIdDefaultStoredAsConfigSpec.groovy
```diff
@@ -0,0 +1,132 @@
...
+import spock.lang.Specification
+
+/**
+ * Exercises the global config {@code grails.mongodb.stringIds.defaultStoredAs}.
+ *
+ * Uses {@link MongoMappingContext} directly (no live MongoDB) — this spec is purely about
+ * how the mapping layer resolves {@code storedAs} from config. The plumbing into the
+ * encoder/decoder/query paths is exercised by {@link StringIdWithObjectIdStorageSpec}.
+ */
+class StringIdDefaultStoredAsConfigSpec extends Specification {
```
borinquenkid, line 37: The second standard lines up with spring 4.0 namespace if I remember correctly

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoCodecSession.groovy
borinquenkid, line 1: Is there a unit test for this?

borinquenkid (dismissed):
Please increase the test coverage on the noted parts. If the coverage was missed in the review, respond inline

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoCodecSession.groovy
codeconsole: done

grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/bugs/StringIdDefaultStoredAsConfigSpec.groovy
codeconsole: done

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/query/MongoQuery.java
codeconsole: done

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/MongoCodecEntityPersister.groovy
codeconsole: done

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/config/MongoMappingContext.java
```diff
@@ -133,6 +137,20 @@ public class MongoMappingContext extends DocumentMappingContext {
 
     private CodecRegistry codecRegistry;
     private Map<Class, Boolean> hasCodecCache = new HashMap<>();
+    /**
```
jdaugherty, line 141: Add a new line between the comment and the other variable?

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/codecs/PersistentEntityCodec.groovy
```diff
@@ -361,7 +361,40 @@ class PersistentEntityCodec extends BsonPersistentEntityCodec {
             }
         }
         else {
-            // TODO: Support non-dirty checkable objects?
+            // Non-DirtyCheckable values: no per-property change history available,
+            // so when the caller is encoding this as an embedded update (null→non-null
+            // transition on a single-valued embedded field), encode every persistent
+            // property. Without this, the parent's $set on the embedded path stays
+            // empty and the sub-document is silently dropped.
+            if (embedded) {
```
jdaugherty, line 369: This is from AI: 
(1) non-DirtyCheckable embedded with a discriminator (subclass) — the new branch writes `_class` only at the embedded root, but nested embedded objects via `Embedded`/`EmbeddedCollection` recurse through the same `else` branch and may need the same treatment

(2) confirm that the existing DirtyCheckable path is genuinely never hit for POGOs that have been touched by `markDirty` upstream — otherwise this branch can fire double-encodes.

grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
```diff
@@ -41,6 +41,25 @@ class IdentityEncoder implements PropertyEncoder<Identity> {
...
 
+        Class<?> storedAs = resolveStoredAs(property)
+        if (storedAs != null && id != null) {
+            if (ObjectId.isAssignableFrom(storedAs) && !(id instanceof ObjectId)) {
+                String hex = id.toString()
+                // Guard against natural-key strings accidentally paired with storedAs: ObjectId.
+                // new ObjectId(<non-hex>) throws IllegalArgumentException, which would surface
+                // deep inside the BSON write pipeline. Fall through to writeString for consistency
+                // with the converter-based paths (MongoCodecSession, MongoCodecEntityPersister).
+                if (ObjectId.isValid(hex)) {
```
jdaugherty, line 62: From AI: 

When `storedAs: ObjectId` is set but the value is not valid hex, the encoder silently writes a BSON String. The doc caveat acknowledges this and points users at `storedAs: String`, but at runtime there is no signal — a misconfigured `id generator: assigned, storedAs: ObjectId` domain just quietly persists Strings. Consider a one-shot `log.warn` (e.g. cached per `(entity, id-shape)` to avoid log spam) so this is debuggable in the field. Today the only way to discover the misconfiguration is to peek at raw BSON.

grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
```diff
@@ -51,6 +70,14 @@ class IdentityEncoder implements PropertyEncoder<Identity> {
 
     }
 
+    private static Class<?> resolveStoredAs(Identity property) {
+        try {
+            return property?.owner?.mapping?.identifier?.storedAs
+        } catch (Throwable ignored) {
```
jdaugherty: `catch (Throwable ignored)` is broader than it needs to be: it swallows `OutOfMemoryError`, `StackOverflowError`, etc. Why can't you use Exception ignored? 

Same pattern repeats in `MongoCodecSession#coerceIdToStoredType`, `MongoCodecEntityPersister#coerceIdToStoredType`, and the two `try` blocks in `MongoQuery` (`IdEquals` and the new `In` handler)

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoCodecSession.groovy
```diff
@@ -286,6 +286,44 @@ class MongoCodecSession extends AbstractMongoSession {
...
+     *
+     * <p>Exercised end-to-end by {@code StringIdWithObjectIdStorageSpec}:
+     * <ul>
+     *   <li>"with storedAs ObjectId, updates persist (no phantom OptimisticLockingException)" — happy path on update filter</li>
+     *   <li>"with storedAs ObjectId, update of a non-hex id document lands on the right row" — null-return fallback on update filter</li>
+     *   <li>"with storedAs ObjectId, delete of a non-hex id document removes the row" — null-return fallback on delete filter</li>
+     *   <li>"with storedAs ObjectId, legacy documents written directly as BSON ObjectId are fully accessible" — update path against legacy BSON ObjectId _id</li>
+     * </ul>
+     */
+    protected Object coerceIdToStoredType(Object nativeKey, PersistentEntity entity) {
```
jdaugherty, line 305: This looks identical to `MongoCodecEntityPersister#coerceIdToStoredType` & the same logic seems to exist in `MongoQuery`'s `IdEquals`/`In` handlers. Lets extract it into a static helper so the logic doesn't diverge?

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoCodecSession.groovy
```diff
@@ -286,6 +286,44 @@ class MongoCodecSession extends AbstractMongoSession {
...
     }
 
+    /**
+     * If the entity's id mapping declares {@code storedAs} and it differs from the in-memory
+     * native key type, coerce the key so that update/delete filters target BSON values of
+     * the correct type (otherwise {@code {_id: "<hex>"}} sent as a BSON String would never
+     * match an {@code _id: ObjectId(...)} document on disk, and the write would silently miss,
+     * surfacing as a misleading {@link OptimisticLockingException}).
+     *
+     * <p>Exercised end-to-end by {@code StringIdWithObjectIdStorageSpec}:
```
jdaugherty: Listing Spock test method names verbatim in javadoc (`"with storedAs ObjectId, updates persist…"`, etc.) is brittle: those names tend to get renamed during PR cycles, and there is no compile-time link to flag the doc as stale when that happens. A short semantic pointer (e.g. `see StringIdWithObjectIdStorageSpec for end-to-end coverage of update/delete filters under storedAs`) carries the same value without the rot risk. Same comment applies to the parallel javadoc in `MongoCodecEntityPersister#coerceIdToStoredType` and the inline comment block above the new `In` handler in `MongoQuery`.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/config/MongoMappingContext.java
```diff
@@ -133,6 +137,20 @@ public class MongoMappingContext extends DocumentMappingContext {
...
+     * Global default storage type for {@code String id} fields that don't declare an explicit
+     * {@code id storedAs: ...} in their mapping. Null means "no default — use the declared
+     * Java type" (current GORM behavior). See {@link MongoSettings#SETTING_STRING_IDS_DEFAULT_STORED_AS}.
+     */
+    private Class<?> stringIdDefaultStoredAs;
```
jdaugherty: `stringIdDefaultStoredAs` has a public setter but is neither `final` nor `volatile`. In practice the field is set in the constructor before `initialize(classes)`, but `MongoDocumentMappingFactory.createIdentity` reads it later for every entity — and any caller that uses the public setter from a different thread than the constructor thread has no guaranteed visibility. Either make it `final` and drop the setter (constructor-only assignment is JMM-safe via the publication rules), or mark it `volatile`. Same applies to the accessor pair if you keep the setter.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/config/MongoMappingContext.java
```diff
@@ -176,6 +215,10 @@ public MongoMappingContext(PropertyResolver configuration, Class... classes) {
      */
     public MongoMappingContext(AbstractMongoConnectionSourceSettings settings, Class... classes) {
         super(settings.getDatabase(), settings);
+        // Must run BEFORE initialize(classes) so that MongoDocumentMappingFactory.createIdentity
+        // (invoked during entity registration) can read the global default.
+        String storedAsDefault = settings.getStringIds() != null ? settings.getStringIds().getDefaultStoredAs() : null;
+        this.stringIdDefaultStoredAs = parseStoredAs(storedAsDefault);
```
jdaugherty: Asymmetry to flag: the deprecated `(PropertyResolver, Class...)` constructor and this `(AbstractMongoConnectionSourceSettings, Class...)` constructor both read `stringIdDefaultStoredAs` from settings, but the public `MongoMappingContext(String defaultDatabaseName, Closure defaultMapping, Class... classes)` constructor does not — so anyone wiring the context manually with that signature silently loses the global default. Either factor a private `applyStoredAsFromConfig(...)` helper and invoke it from all constructors that have the relevant inputs, or document on the bare constructor that it does not honor `grails.mongodb.stringIdsDefaultStoredAs`. Today this is the kind of inconsistency that gets discovered only when a test using the bare constructor passes locally and the production datastore — using the settings constructor — diverges.

grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/bugs/StringIdWithObjectIdStorageSpec.groovy
```diff
@@ -0,0 +1,388 @@
...
+}
+
+@Entity
+class StoredAsVideo {
+    String id
+    String title
+
+    static mapping = {
+        id storedAs: ObjectId
+        version false  // orthogonal to storedAs; avoids version-field noise in raw-insert tests
```
jdaugherty, line 405: This is from AI: 

Both `StoredAsVideo` and `AssignedNonHexVideo` declare `version false`. The headline regression for this PR — `with storedAs ObjectId, updates persist (no phantom OptimisticLockingException)` — therefore cannot actually exercise the optimistic-locking path it claims to cover, because there is no `version` field. Worth adding a parallel domain that keeps versioning on (the default) so the original failure mode (where a missed update filter throws `OptimisticLockingException`) is genuinely reproduced and proven fixed. Otherwise this regression test will pass even if a future change accidentally reintroduces the bug under versioned domains, which is the more common case in real apps.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/config/MongoMappingContext.java
codeconsole: Fixed

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/codecs/PersistentEntityCodec.groovy
codeconsole: (1) Nested embedded recurses through encodeEmbeddedUpdate -> encodeUpdate(embedded=true) and re-enters the same else-if  branch, writing _class at each non-root level. (2) Branches are mutually exclusive: POGOs that have markDirty() called on them must implement DirtyCheckable, so cannot reach this branch. No double-encode possible. No fix needed.

grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
codeconsole: Fixed: narrowed every catch to Exception. The four duplicate core sites collapsed into the new helper encoder's  resolveStoredAs uses catch (Exception ignored).

grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/IdentityEncoder.groovy
codeconsole: Fixed: added SLF4J logger + ConcurrentHashMap.newKeySet() warn-once cache. Logs one warn per owning entity class, pointing users at storedAs:String for natural-key domains.

grails-data-mongodb/core/src/test/groovy/org/grails/datastore/gorm/mongo/bugs/StringIdWithObjectIdStorageSpec.groovy
codeconsole: Fixed: added VersionedStoredAsVideo domain (default versioning, id storedAs:ObjectId) and a parallel test that saves,  reloads, mutates, saves again, asserts no OLE and version increment in the raw doc.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoCodecSession.groovy
codeconsole: Fixed: added MongoIdCoercion (Java, in org.grails.datastore.mapping.mongo.engine) with resolveStoredAs(entity) and coerceIdToStoredType(key, entity). Two Groovy methods delegate; MongoQuery IdEquals/In call directly. Null-return fallback lives in one place.

jdaugherty (approved):
@borinquenkid Can you confirm you are ok with this so it can be merged?

borinquenkid (approved):

matrei (commented):
Can we target `7.2.x`?

jdaugherty:
@codeconsole is it ok to target 7.2 with this change?

matrei (approved):

PR #15584, opened 2026-04-18
fix: prompt before removing old SVN release folder during promotion

### LINKED ISSUE #15551: Promotion Scripts need to support optional deleting
### Issue description

We released 7.0.10 & 7.1.0-RC1 in parallel.  The promotion scripts currently remove the "old" version automatically, which in this case was 7.0.10.  We need to change the svn promotion scripts to prompt if the old folder should be removed.

chloeb-smith:
**Description**

This PR addresses issue #15551. During parallel releases (e.g. 7.0.10 and 7.1.0-RC1), the promotion script was automatically deleting the old release folder, which could remove a still-current release. This change adds an interactive prompt before deleting, so the operator can choose whether to remove the old folder.

**Changes**
A [y/N] interactive prompt was added so the operator must explicitly confirm before the deletion occurs. If they answer anything other than y/Y/yes/Yes, the deletion is skipped and the promotion continues as normal.

**Verification**
 Verified using a local mock of the `svn` command, testing three scenarios:
  - Answering `y` — old folder was deleted and promotion completed
  - Answering `n` — deletion was skipped and promotion completed
  - Pressing Enter (default `N`) — deletion was skipped and promotion completed
  - Answering `y , y` — both folders deleted, then promoted                                                                                                            
  - Answering `n , n` — both folders skipped, then promoted                                                                                                            
  - Answering `y , n` — first folder deleted, second skipped, then promoted                                                                                            
  - Answering `yes/Yes` — both full-word confirmations accepted, both deleted, then promoted

### DIFF
--- a/.github/scripts/releaseDistributions.sh
+++ b/.github/scripts/releaseDistributions.sh
@@ -58,20 +58,18 @@ svn_exists() {
   svn ls "${svn_flags[@]}" --depth=empty "${url}" >/dev/null 2>&1
 }
 
-old_release_folder="$(svn ls "${svn_flags[@]}" "${RELEASE_ROOT}" | awk -F/ 'NF{print $1; exit}')"
-if [[ -n "${old_release_folder}" ]]; then
-  PRIOR_RELEASE_URL="${RELEASE_ROOT}/${old_release_folder}"
-  read -r -p "Remove old release folder '${old_release_folder}' at ${PRIOR_RELEASE_URL}? [y/N] " confirm
+for folder in $(svn ls "${svn_flags[@]}" "${RELEASE_ROOT}"); do
+  folder=$(echo "$folder" | sed 's|/$||')
+  PRIOR_RELEASE_URL="${RELEASE_ROOT}/${folder}"
+  read -r -p "Remove old release folder '${folder}' at ${PRIOR_RELEASE_URL}? [y/N] " confirm
   if [[ "${confirm}" =~ ^[Yy]$ ]]; then
     echo "🗑️ Deleting old release folder: ${PRIOR_RELEASE_URL}"
-    svn rm "${svn_flags[@]}" -m "Remove previous release ${old_release_folder}" "${PRIOR_RELEASE_URL}"
+    svn rm "${svn_flags[@]}" -m "Remove previous release ${folder}" "${PRIOR_RELEASE_URL}"
     echo "✅ Deleted old release folder"
   else
     echo "⏭️ Skipping removal of old release folder: ${PRIOR_RELEASE_URL}"
   fi
-else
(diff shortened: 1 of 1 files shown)


.github/scripts/releaseDistributions.sh
```diff
@@ -61,9 +61,14 @@ svn_exists() {
 old_release_folder="$(svn ls "${svn_flags[@]}" "${RELEASE_ROOT}" | awk -F/ 'NF{print $1; exit}')"
```
sanjana2505006: Using `awk ... exit` only identifies the first folder in the SVN directory. In a multi-branch environment (e.g., having both 6.x and 7.x releases in the same folder), this might prompt to delete a folder from a different branch while missing the actual previous version of the current branch.

I guess it would be safer to loop through all folders found by `svn ls` and prompt for each?

.github/scripts/releaseDistributions.sh
```diff
@@ -61,9 +61,14 @@ svn_exists() {
...
 if [[ -n "${old_release_folder}" ]]; then
   PRIOR_RELEASE_URL="${RELEASE_ROOT}/${old_release_folder}"
-  echo "🗑️ Deleting old release folder: ${PRIOR_RELEASE_URL}"
-  svn rm "${svn_flags[@]}" -m "Remove previous release ${old_release_folder}" "${PRIOR_RELEASE_URL}"
-  echo "✅ Deleted old release folder"
+  read -r -p "Remove old release folder '${old_release_folder}' at ${PRIOR_RELEASE_URL}? [y/N] " confirm
```
sanjana2505006: Since the script already reads a password from stdin, there's a risk that this second read call will consume unintended data if the script is run in a way that redirects stdin.

To be safe, we should probably use `read ... < /dev/tty` for this confirmation prompt to ensure it always interacts with the terminal.

.github/scripts/releaseDistributions.sh
```diff
@@ -61,9 +61,14 @@ svn_exists() {
...
+  if [[ "${confirm}" =~ ^[Yy]$ ]]; then
```
sanjana2505006: Minor suggestion: The regex `^[Yy]$` is quite strict. It might be more user-friendly to support `yes` as well by using something like `[[ "${confirm}" =~ ^[Yy](es)?$ ]]`.

.github/scripts/releaseDistributions.sh
chloeb-smith: Thanks! This makes sense

sanjana2505006:
@chloeb-smith LGTM :)

jdaugherty (approved):
@matrei are you ok with this too?

jamesfredley (approved):

PR #15585, opened 2026-04-19
Filterless Sitemesh 3

codeconsole:
##  Sitemesh 3 Filterless implementation

- performance updates
- filterless
- prep for Grails 8
- based on 3.2.3-SNAPSHOT which will be release to final after this merge.  Once testing is complete, a subsequent PR with 3.2.3 final will be released


jamesfredley:
Let's do this against 8.0.x, 7.1.x is shipped and we have to get 8.0.x shipping ASAP to beat the Spring Boot 3.5.x June 30th EOL.

codeconsole:
> Let's do this against 8.0.x, 7.1.x is shipped and we have to get 8.0.x shipping ASAP to beat the Spring Boot 3.5.x June 30th EOL.

@jamesfredley this won't work in 8.0.x  This is an incremental pathway to 8.0.x and it isn't enabled by default. 
As some as we get this out, I can get 8.0.x updated quickly.

codeconsole:
> Let's do this against 8.0.x, 7.1.x is shipped and we have to get 8.0.x shipping ASAP to beat the Spring Boot 3.5.x June 30th EOL.

@jamesfredley could also do a 7.2.x branch, merge and ship 7.2 after 8.0

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
```diff
@@ -36,7 +36,7 @@ class GrailsVersionSpec extends Specification {
         "3.2.0"         | true
         "3.1.0"         | true
         "3.3.0"         | true
-        "7.1.0"         | false
+        "7.1.0"         | true
```
jdaugherty: I think there was an intentional false case here. We should probably update 7.1.0 to some arbitrary large value instead.

jdaugherty:
@jamesfredley given that sitemesh3 has known problems, I view this as a bug fix.  @codeconsole does this fix the async & forwarding workflows? (there are ignored tests in the associated test-examples project).  

I'm ok doing a 7.2 and having this change, especially because this is the non-default path.

grails-datamapping-support/src/test/groovy/org/grails/datastore/mapping/core/grailsversion/GrailsVersionSpec.groovy
jdaugherty: I've updated this in the 7.1 branch after seeing it fail on multiple reviews.  You should be able to update your base branch now.

jdaugherty:
I've created a 7.2.x branch after discussion with the team.  Can you please set the base of this PR to 7.2 & update your branch?

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3CapturedPage.java
```diff
@@ -0,0 +1,359 @@
...
+    private StreamCharBuffer headBuffer;
+    private StreamCharBuffer bodyBuffer;
+    private StreamCharBuffer titleBuffer;
+    private StreamCharBuffer pageBuffer;
+    private CharSequence renderedContent;
+
+    private final Map<String, StreamCharBuffer> contentBuffers = new LinkedHashMap<>();
+    private final Map<String, String> pageProperties = new HashMap<>();
+
+    private boolean used;
```
jdaugherty: `used` (and `titleCaptured` below) should also be `volatile` for the same reason `propertiesMaterialized` is, per the comment on line 63. `isUsed()` is read by `CaptureAwareContentProcessor.build(...)` to pick the short-circuit vs. fallback path — and in the async-dispatch scenario called out in the javadoc, the capture taglibs may run on one thread and `CaptureAwareContentProcessor` may run on another. A plain boolean read there has no happens-before with the write in `markUsed()`, so the short-circuit could be missed (or worse, taken with a partially-written state in the rest of the object).

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3CapturedPage.java
```diff
@@ -0,0 +1,359 @@
...
+    @Override
+    public CharSequenceBuffer createDataOnlyBuffer() {
+        return delegate.createDataOnlyBuffer();
+    }
+
+    private void materializeProperties() {
+        if (propertiesMaterialized) {
+            return;
+        }
+        propertiesMaterialized = true;
```
jdaugherty: The `propertiesMaterialized` flag protects against *redundant* materialization but not *concurrent* materialization: two async-dispatch threads can both see `false`, both set it to `true`, and both walk the delegate's property tree, which is not synchronized. In the worst case the `InMemoryContent` delegate ends up with partially-initialized children. Consider wrapping the body in `synchronized(this)` with a double-checked pattern, or pre-materializing once on write rather than on read.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3LayoutFinder.java
```diff
@@ -0,0 +1,255 @@
...
+ *   <li>Configured default (e.g. {@code grails.sitemesh.default.layout})</li>
+ * </ol>
+ *
+ * <p>Results are cached by (controllerName, actionUri) outside of the
+ * DEVELOPMENT environment.</p>
+ */
+public class Sitemesh3LayoutFinder implements DecoratorSelector<SiteMeshContext> {
+
+    private static final String LAYOUTS_PATH = "/layouts";
+    private static final long LAYOUT_CACHE_EXPIRATION_MILLIS = Long.getLong("grails.gsp.reload.interval", 5000);
```
jdaugherty: Two concerns here:

1. Reading via `Long.getLong("grails.gsp.reload.interval", ...)` reaches into JVM system properties, which is unusual for Grails — the rest of the plugin reads from `grailsApplication.config`. This key won't show up when a user greps application.yml/config.groovy for `grails.gsp`.

2. The name `grails.gsp.reload.interval` suggests *GSP* reload timing, not *layout-cache* expiration. Using the same property for two unrelated concepts will be confusing if someone tunes one expecting the other. Consider a dedicated key like `grails.sitemesh.layout.cache.interval`, read through `grailsApplication.config`, with the value passed in via a setter (same pattern as `gspReloadEnabled` / `defaultDecoratorName`).

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3LayoutFinder.java
```diff
@@ -0,0 +1,255 @@
...
+
+    private final GrailsConventionGroovyPageLocator groovyPageLocator;
+
+    private String defaultDecoratorName;
+    private boolean gspReloadEnabled;
+    private boolean cacheEnabled = (Environment.getCurrent() != Environment.DEVELOPMENT);
+
+    private final Map<String, LayoutCacheValue> namedDecoratorCache = new ConcurrentHashMap<>();
+    private final Map<LayoutCacheKey, LayoutCacheValue> layoutDecoratorCache = new ConcurrentHashMap<>();
```
jdaugherty, line 80: Both caches are unbounded. In practice they're bounded by the number of distinct (controllerName, actionUri) pairs, which is fine for typical apps — but worth adding a note in the javadoc, or a bounded cache (e.g. an LRU via `LinkedHashMap` + `removeEldestEntry`) to be safe for apps with many wildcard-matched actions. Non-blocking, but worth thinking about.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
```diff
@@ -148,40 +163,44 @@ class RenderSitemeshTagLib implements TagLibrary {
...
-        out << tag.toString()
     }
 
     Closure content = { attrs, body ->
-        StringBuilder tag = new StringBuilder("""<content tag="${attrs.tag}">""")
-        tag.append(body())
-        tag.append('</content>')
-        out << tag.toString()
+        out << '<content tag="'
+        out << attrs.tag
```
jdaugherty: `attrs.tag` is written directly into an HTML attribute value without encoding. If a tag name ever contains `"`, `<`, or `&` the produced markup is malformed. `Sitemesh3LayoutTagLib.captureTagContent` already does the right thing via `codecLookup.lookupEncoder('HTML')` — worth mirroring here for consistency. Likely harmless in practice (tag names are usually static strings) but cheap to fix.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
```diff
@@ -48,55 +55,73 @@ class Sitemesh3GrailsPlugin extends Plugin {
...
 
-            grailsLayoutHandlerMapping(GrailsLayoutHandlerMapping)
+            Config config = grailsApplication.getConfig()
+            boolean developmentMode = Metadata.getCurrent().isDevelopmentEnvironmentAvailable()
+            Environment env = Environment.current
+            boolean enableReload = env.isReloadEnabled() ||
+                    config.getProperty('grails.gsp.enable.reload', Boolean, false) ||
+                    (developmentMode && env == Environment.DEVELOPMENT)
+            String resolvedDefaultLayout = config.getProperty('grails.sitemesh.default.layout') ?:
+                    config.getProperty('sitemesh.decorator.default')
```
jdaugherty: Nit: `defaultLayout` was already resolved on line 83 from the same config key. The fallback to `sitemesh.decorator.default` here is redundant because line 84 just added a property source that writes `sitemesh.decorator.default` = `defaultLayout` whenever `defaultLayout` is truthy. You can collapse this to `String resolvedDefaultLayout = defaultLayout`.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
```diff
@@ -48,55 +55,73 @@ class Sitemesh3GrailsPlugin extends Plugin {
...
+                defaultDecoratorName = resolvedDefaultLayout ?: null
+            }
+
+            // Replace the filter registration from
+            // org.sitemesh.autoconfigure.SiteMeshAutoConfiguration with a no-op
+            // filter bean under the same name. SiteMeshAutoConfiguration is
+            // @ConditionalOnMissingBean(name = "sitemesh") so registering this
+            // bean disables the upstream filter-based integration entirely.
+            // Decoration is done by the Spring MVC view resolver chain.
+            sitemesh(FilterRegistrationBean) { bean ->
```
jdaugherty, line 112: The `NoopSitemeshFilter` is never actually invoked (the registration has `enabled = false`), so the whole no-op filter class feels like scaffolding required only because the upstream bean type is `FilterRegistrationBean`. This works, but a one-line comment on `NoopSitemeshFilter` pointing to this explanation would help the next person touching it understand why an apparently-dead class exists.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3LayoutTagLib.groovy
```diff
@@ -0,0 +1,208 @@
...
+ */
+@CompileStatic
+@TagLib
+class Sitemesh3LayoutTagLib implements TagLibrary {
+
+    static String namespace = 'grailsLayout'
+
+    CodecLookup codecLookup
+
+    def captureTagContent(GrailsPrintWriter writer, String tagname, Map attrs, Object body, boolean noEndTagForEmpty = false) {
```
jdaugherty, line 52: Let's add documentation to these new tags?

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
```diff
@@ -16,78 +16,105 @@
...
-     *
-     * &lt;g:pageProperty default="defaultValue" name="body.onload" /&gt;<br/>
-     *
-     * @emptyTag
-     *
-     * @attr REQUIRED name the property name
-     * @attr default the default value to use if the property is null
-     * @attr writeEntireProperty if true, writes the property in the form 'foo = "bar"', otherwise renders 'bar'
-     */
     Closure pageProperty = { attrs ->
```
jdaugherty, line 134: Why remove the documentation?

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
```diff
@@ -109,18 +136,6 @@ class RenderSitemeshTagLib implements TagLibrary {
...
-     *
-     * &lt;g:ifPageProperty name="meta.index"&gt;body to invoke&lt;/g:ifPageProperty&gt;<br/>
-     *
-     * or it equals a certain value:<br/>
-     *
-     * &lt;g:ifPageProperty name="meta.index" equals="blah"&gt;body to invoke&lt;/g:ifPageProperty&gt;
-     *
-     * @attr name REQUIRED the property name
-     * @attr equals optional value to test against
-     */
```
jdaugherty, line 123: Why remove the documentation?

jdaugherty (dismissed):
It's great that the async, etc seems solved now.  I think it's very likely we can switch to sitemesh3 for Grails 8 as a defualt with these changes.  In addition to my other comments, we should enhance the grails-docs for this change: 
1. add upgrading72x.adoc to for 7.2.x instructions and document the sitemesh changes
2. update any other grails docs specific to sitemesh3

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3CapturedPage.java
codeconsole: Done — both are now volatile for the same reason as propertiesMaterialized.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3CapturedPage.java
codeconsole: Fixed — extracted body into doMaterializeProperties() and wrapped the entry in synchronized(this) with a double-check.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3LayoutFinder.java
codeconsole: Fixed — removed the system-property read; added a layoutCacheExpirationMillis setter injected from grails.sitemesh.layout.cache.interval via grailsApplication.config with default 5000 ms.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
codeconsole: Fixed — added @Autowired CodecLookup and HTML-encoded via codecLookup.lookupEncoder('HTML'), matching captureTagContent.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
codeconsole: Fixed — simplified to String resolvedDefaultLayout = defaultLayout.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
codeconsole: Added a comment explaining it satisfies SiteMeshAutoConfiguration's @ConditionalOnMissingBean(name = "sitemesh") guard and is intentionally never invoked.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3LayoutTagLib.groovy
codeconsole: Done — added Grails-style javadoc with @attr annotations to all capture closures (captureHead, captureBody, captureTitle, wrapTitleTag, captureMeta, captureContent, parameter).

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
codeconsole: Restored

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
codeconsole: Restored

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3LayoutFinder.java
codeconsole: Added a javadoc note on the class. In practice they're bounded by the number of distinct (controllerName, actionUri) pairs

jdaugherty:
@codeconsole is this ready to re-review?

codeconsole:
> @codeconsole is this ready to re-review?

@jdaugherty sure, unless you want me to bound the cache.  I updated the doc, just didn't bound the cache yet.

grails-doc/src/en/guide/theWebLayer/gsp/layouts.adoc
```diff
@@ -20,7 +20,9 @@ under the License.
 
 ==== Creating Layouts
 
-This product includes software developed by the OpenSymphony Group (https://github.com/sitemesh). The Sitemesh framework is used to provide a decorator engine, to support view layouts. Layouts are located in the `grails-app/views/layouts` directory. A typical layout can be seen below:
+Grails uses the https://github.com/sitemesh/sitemesh3[SiteMesh 3] framework as its decorator engine to support view layouts. In Grails 7.2 and later, decoration is performed by a Spring MVC `ViewResolver` rather than a Servlet filter, which improves performance and enables correct support for async controller return types.
```
jdaugherty, line 23: "This product includes software developed by the OpenSymphony Group (https://github.com/sitemesh)." is a requirement of the sitemesh 2 license.  Unless we're going to drop sitemesh 2 in Grails 7.2, we have to leave that line to be compatible with the license.  Note: we can move it to a subsection explaining that this is only if grails-layout is used.

grails-doc/src/en/guide/upgrading/upgrading72x.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+* The `SiteMeshFilter` servlet filter is no longer active. A `NoopSitemeshFilter` placeholder bean is registered under the `sitemesh` name to prevent SiteMesh's auto-configuration from activating the upstream filter-based integration.
+* Layout resolution and decoration are now handled by `Sitemesh3LayoutView` (a Spring `View`) and `CaptureAwareContentProcessor`.
+* GSP pages populate a `Sitemesh3CapturedPage` at render time via the `grailsLayout` capture tag library. SiteMesh no longer parses the raw HTML response to extract head/body/title.
+
+===== Required Actions
+
+For most applications **no changes are required**. The filter-less integration is a drop-in replacement: layouts, `<g:layoutHead>`, `<g:layoutTitle>`, `<g:layoutBody>`, `<g:pageProperty>`, `<g:applyLayout>`, and layout-by-convention all behave identically to the previous release.
+
+The following SiteMesh 2-era configuration properties are no longer consulted and can be removed from your `application.yml` if present:
```
jdaugherty: We still have to support sitemesh 2 so we need to clarify that this is only if sitemesh3 is being used.

jdaugherty (commented):
I took another pass, I think other than the license & sitemesh 2 clarifications, this looks good.

grails-doc/src/en/guide/upgrading/upgrading72x.adoc
codeconsole: done

grails-doc/src/en/guide/theWebLayer/gsp/layouts.adoc
codeconsole: done

sbglasius (approved):

jdaugherty:
I'm ok to merge this once we get one more person to sign off.  @jamesfredley @matrei I assume you're ok with the documentation updates specifically since they suggest sitemesh 3 as the default.

sbglasius:
@jdaugherty I think, that with your thorough review and mine we should merge.

jdaugherty:
I missed your review @sbglasius .  I think we can merge.

PR #15597, opened 2026-04-24
build: bootstrap Gradle plugin repositories

### LINKED ISSUE #15596: Settings/plugin bootstrap fails to resolve `grails-publish` from ASF snapshots
When updating `org.apache.grails.gradle:grails-publish` to `1.0.0-SNAPSHOT`, the Gradle build fails during initialization with an error like:
```console
Error resolving plugin [id: 'com.gradle.develocity', version: '4.3.2']
> Could not resolve all artifacts for configuration 'classpath'.
   > Could not find org.apache.grails.gradle:grails-publish:1.0.0-SNAPSHOT.
     Searched in the following locations:
       - https://plugins.gradle.org/m2/org/apache/grails/gradle/grails-publish/1.0.0-SNAPSHOT/maven-metadata.xml
       - https://plugins.gradle.org/m2/org/apache/grails/gradle/grails-publish/1.0.0-SNAPSHOT/grails-publish-1.0.0-SNAPSHOT.pom
     Required by:
         settings file 'grails-gradle/settings.gradle' > project :build-logic-root:build-logic

```

The problem is not that the artifact is missing, but that the repository is configured too late in the bootstrap sequence.

matrei:
Closes gh-15596

### DIFF
--- a/gradle/plugin-repositories.gradle
+++ b/gradle/plugin-repositories.gradle
@@ -1,0 +1,54 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


gradle/plugin-repositories.gradle
```diff
@@ -0,0 +1,54 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+ext.configurePluginRepositories = { RepositoryHandler repositories ->
```
jdaugherty, line 20: We have a settings repo plugin.  I'm assuming we can't take a similar solution for the plugin repos?  If not, can we at least move this configuration adjacent to that plugin (build-logic) so they're in the same location?

jdaugherty:
I think we should merge to help @jamesfredley get the micronaut change merged quicker.  Then we can revisit on the repo settings change.

jamesfredley (approved):
We can revisit that one item later, if needed.

jdaugherty:
@matrei would you be ok to relocate the config with the build-logic?  James & I wanted to get this merged so we can get the micronaut changes merged

PR #15598, opened 2026-04-24
fix(wrapper): honor PREFERRED_GRAILS_VERSION and repair publish CI

jamesfredley:
## Summary

The `publish` job's **Verify grails-wrapper** step has failed on every push to `7.0.x` since 8.0.0-SNAPSHOT began being published to the Apache Maven snapshot repo. This PR fixes that, fixes the downstream **Upload Wrapper Zip** path bug it exposes, makes the wrapper's documented `PREFERRED_GRAILS_VERSION` env var actually work (currently dead code), and tidies the wrapper's configuration surface.

## Root cause

With `GRAILS_WRAPPER_ALLOWED_TYPES='SNAPSHOT'` and no pinned version, the wrapper reads `grails-cli`'s root `maven-metadata.xml`, sorts by `GrailsVersion` order, and picks the globally latest SNAPSHOT. Because 8.0.x SNAPSHOTs are now being published from the main development branch, the resolver picks that jar:

```
java.lang.UnsupportedClassVersionError: org/apache/grails/cli/DelegatingShellApplication
  has been compiled by a more recent version of the Java Runtime (class file version 65.0),
  this version of the Java Runtime only recognizes class file versions up to 61.0
```

The wrapper exposes `PREFERRED_GRAILS_VERSION` as the documented way to pin the version (see the `README` "Creating a Grails Application" section), but the env var is dead code in `grails.init.GrailsVersion#getPreferredGrailsVersion` - it is only consulted when `gradle.properties` contains a `grailsVersion` key whose value is literally `null`, which cannot happen with a well-formed properties file. Users following the documented README flow therefore hit the same bytecode-compatibility failure the CI is hitting.

### 1. `grails-wrapper/.../GrailsVersion.java` - make `PREFERRED_GRAILS_VERSION` work

Rewrite `getPreferredGrailsVersion` so the precedence matches the documented behaviour:

1. **Inside a Grails project** (a `gradle.properties` containing a `grailsVersion` key) - use that and ignore the env var.
2. **Outside a project** (no `gradle.properties`, or one without `grailsVersion`) - honour `PREFERRED_GRAILS_VERSION`.
3. Otherwise - `null`, let the wrapper resolve the latest from Maven metadata.

Extract a package-private overload that accepts the properties file so the logic is testable without mutating the JVM working directory.

### 2. `grails-wrapper/.../GrailsVersion.java` - named constants for the wrapper config surface

Pin the user-facing configuration strings as public constants so they aren't duplicated as literals across production and tests:

- `PREFERRED_GRAILS_VERSION_ENV` = `"PREFERRED_GRAILS_VERSION"`
- `GRAILS_WRAPPER_ALLOWED_TYPES_ENV` = `"GRAILS_WRAPPER_ALLOWED_TYPES"`
- `GRAILS_VERSION_PROPERTY` = `"grailsVersion"`

Referenced from the production code and all tests. The validation error message now reads `"Invalid Value in " + GRAILS_WRAPPER_ALLOWED_TYPES_ENV + ": ..."` so it stays in sync if the env var is ever renamed.

### 3. `grails-wrapper/.../GrailsVersionSpec.groovy` - coverage

Five new Spock specs exercising the precedence matrix:

- `gradle.properties` with `grailsVersion` wins over env var
- env var used when `gradle.properties` is missing
- env var used when `gradle.properties` has no `grailsVersion` key
- whitespace-only env var treated as unset
- `null` when neither is set

All 20 `:grails-wrapper:test` specs pass (15 existing + 5 new).

### 4. `.github/workflows/gradle.yml` - pin wrapper verification, fix upload path

- Export `PREFERRED_GRAILS_VERSION` (now honoured by the wrapper) derived from the repo's `projectVersion` so the wrapper resolves against this branch's snapshot - branch-portable: 7.0.x → 7.0.x, 7.1.x → 7.1.x, 8.0.x → 8.0.x.
- Promote `GRAILS_WRAPPER_ALLOWED_TYPES` from an inline `export` to step-level `env`.
- Fix the pre-existing Upload Wrapper Zip path bug: the step used `path: tmp/wrapper` relative to the repo root but the zip lands under `build/tmp/wrapper`. Verified on the last successful `publish` (Apr 2):
  ```
  ##[warning]No files were found with the provided path: tmp/wrapper. No artifacts will be uploaded.
  ```
  The artifact has been silently empty for the entire life of this workflow. Fixed to `build/tmp/wrapper`.

## Verification

- `:grails-wrapper:test` passes locally (15 existing + 5 new specs, 20 total).
- CI failure pattern reproducible in every recent push run:
  - https://github.com/apache/grails-core/actions/runs/24496413864/job/71598069598 (Apr 16)
  - https://github.com/apache/grails-core/actions/runs/24487358870/job/71568765789 (Apr 16)
  - https://github.com/apache/grails-core/actions/runs/24227501271/job/70735767232 (Apr 10)
- Last successful `publish` on `7.0.x` was Apr 2 - that run also logged the empty-artifact warning, confirming the upload bug is pre-existing and not introduced by this PR.

## Commits

1. `fix(ci): pin wrapper verification to this branch's snapshot version` - initial CI-only fix writing `gradle.properties`.
2. `Honor PREFERRED_GRAILS_VERSION env var and tidy wrapper CI` - wrapper code fix, tests, and simplify CI to use the env var instead of writing `gradle.properties`. Also fixes matrei's and Copilot's review comments.
3. `Extract wrapper configuration strings as public constants on GrailsVersion` - pure refactor, no behaviour change.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -398,10 +398,19 @@ jobs:
         if: success()
         run: |
           export GRAILS_WRAPPER_ALLOWED_TYPES='SNAPSHOT'
+          # Pin the wrapper to this branch's snapshot; otherwise it resolves the
+          # globally latest SNAPSHOT in the Apache repo (e.g. 8.0.0-SNAPSHOT from
+          # the main development branch), which may be compiled for a newer Java
+          # runtime than this job's JDK and fail with UnsupportedClassVersionError.
+          PROJECT_VERSION=$(grep '^projectVersion=' gradle.properties | cut -d'=' -f2)
           cp grails-wrapper/build/distributions/apache-grails-wrapper-*.zip build/wrapper.zip
           cd build
           unzip wrapper -d tmp
           mv tmp/apache-grails-wrapper-* tmp/wrapper
+          # The wrapper reads `grailsVersion` from gradle.properties in its CWD
+          # (see grails.init.GrailsVersion#getPreferredGrailsVersion) to select
+          # which Grails CLI snapshot to download.
+          echo "grailsVersion=${PROJECT_VERSION}" > gradle.properties
           ./tmp/wrapper/grailsw --version
       - name: "📤 Upload Wrapper Zip to Workflow Summary Page"
         uses: actions/upload-artifact@v7.0.0
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -396,27 +396,23 @@ jobs:
           ./gradlew :grails-wrapper:distZip
       - name: "✅ Verify grails-wrapper"
         if: success()
+        env:
+          GRAILS_WRAPPER_ALLOWED_TYPES: 'SNAPSHOT'
         run: |
-          export GRAILS_WRAPPER_ALLOWED_TYPES='SNAPSHOT'
-          # Pin the wrapper to this branch's snapshot; otherwise it resolves the
-          # globally latest SNAPSHOT in the Apache repo (e.g. 8.0.0-SNAPSHOT from
-          # the main development branch), which may be compiled for a newer Java
-          # runtime than this job's JDK and fail with UnsupportedClassVersionError.
-          PROJECT_VERSION=$(grep '^projectVersion=' gradle.properties | cut -d'=' -f2)
+          # Pin the wrapper to this branch's snapshot so it doesn't resolve the
+          # globally latest SNAPSHOT in the Apache repo, which may be compiled
+          # for a newer Java runtime than this job's JDK.
+          export PREFERRED_GRAILS_VERSION=$(grep '^projectVersion=' gradle.properties | cut -d'=' -f2)
           cp grails-wrapper/build/distributions/apache-grails-wrapper-*.zip build/wrapper.zip
           cd build
           unzip wrapper -d tmp
           mv tmp/apache-grails-wrapper-* tmp/wrapper
-          # The wrapper reads `grailsVersion` from gradle.properties in its CWD
-          # (see grails.init.GrailsVersion#getPreferredGrailsVersion) to select
-          # which Grails CLI snapshot to download.
-          echo "grailsVersion=${PROJECT_VERSION}" > gradle.properties
           ./tmp/wrapper/grailsw --version
       - name: "📤 Upload Wrapper Zip to Workflow Summary Page"
         uses: actions/upload-artifact@v7.0.0
         with:
           name: apache-grails-wrapper-SNAPSHOT-bin
-          path: tmp/wrapper
+          path: build/tmp/wrapper
   publishForge:
     if: github.repository_owner == 'apache' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
     needs: [ buildForge, publishGradle, publish ]
--- a/grails-wrapper/src/main/java/grails/init/GrailsVersion.java
+++ b/grails-wrapper/src/main/java/grails/init/GrailsVersion.java
@@ -166,14 +166,16 @@ public class GrailsVersion implements Comparable<GrailsVersion> {
         }
 
         String overrideGrailsVersion = System.getenv(PREFERRED_GRAILS_VERSION_ENV);
-        if (overrideGrailsVersion != null && !overrideGrailsVersion.trim().isEmpty()) {
-            try {
-                return new GrailsVersion(overrideGrailsVersion.trim());
-            } catch (Exception e) {
-                System.out.println("An invalid Grails Version [" + overrideGrailsVersion + "] was specified in " + PREFERRED_GRAILS_VERSION_ENV);
-                e.printStackTrace();
-                System.exit(1);
-            }
+        if (overrideGrailsVersion == null || (overrideGrailsVersion = overrideGrailsVersion.trim()).isEmpty()) {
+            return null;
+        }
+
+        try {
+            return new GrailsVersion(overrideGrailsVersion);
+        } catch (Exception e) {
+            System.out.println("An invalid Grails Version [" + overrideGrailsVersion + "] was specified in " + PREFERRED_GRAILS_VERSION_ENV);
+            e.printStackTrace();
+            System.exit(1);
         }
 
         return null;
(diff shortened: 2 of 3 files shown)


.github/workflows/gradle.yml
```diff
@@ -398,10 +398,19 @@ jobs:
         if: success()
         run: |
           export GRAILS_WRAPPER_ALLOWED_TYPES='SNAPSHOT'
+          # Pin the wrapper to this branch's snapshot; otherwise it resolves the
+          # globally latest SNAPSHOT in the Apache repo (e.g. 8.0.0-SNAPSHOT from
```
matrei: Should we avoid specifying 8.0.0-SNAPSHOT since it will become stale?

matrei (approved):

.github/workflows/gradle.yml
jamesfredley: I'll remove this part of comment.

jamesfredley:
I am checking if a change is also needed for the end application grails wrapper, when using a snapshot version.

jamesfredley:
@matrei @copilot-pull-request-reviewer new commit `5037a40` - heads up, the PR has expanded in scope, your prior approval is stale:

- ✅ @matrei: reworked the inline comment to describe the root cause instead of naming `8.0.0-SNAPSHOT`.
- ✅ @copilot: `Upload Wrapper Zip` step's `path: tmp/wrapper` is repo-root relative but the zip actually lands under `build/tmp/wrapper`. Apr 2's successful run confirmed it: `##[warning]No files were found with the provided path: tmp/wrapper. No artifacts will be uploaded.` Fixed to `build/tmp/wrapper`.
- 🆕 End-user scenario: the wrapper README ("Creating a Grails Application" - extract to empty dir, `export PREFERRED_GRAILS_VERSION=...`, `grailsw -t forge create-app`) hits the same latent bug because `PREFERRED_GRAILS_VERSION` is dead code in `GrailsVersion#getPreferredGrailsVersion` - only consulted inside an unreachable branch. Reworked the precedence to match the README (gradle.properties inside a project wins; env var outside a project is honoured), extracted a package-private overload for testability, and added 5 Spock tests covering the precedence matrix.
- 🧹 With the env var now functional, the CI step `export PREFERRED_GRAILS_VERSION=...` is cleaner than writing a `gradle.properties` into `build/`.

All 12 `:grails-wrapper:test` specs pass locally (7 existing + 5 new).

grails-wrapper/src/main/java/grails/init/GrailsVersion.java
```diff
@@ -138,29 +190,17 @@ public static GrailsVersion getPreferredGrailsVersion() {
...
-                    System.out.println("An invalid Grails Version [" + overrideGrailsVersion + "] was specified in PREFERRED_GRAILS_VERSION");
-                    e.printStackTrace();
-                    System.exit(1);
-                }
-            }
-
-            System.out.println("gradle.properties does not contain grailsVersion; assuming latest Grails Version");
+        String grailsVersion = properties.getProperty(GRAILS_VERSION_PROPERTY);
+        if (grailsVersion == null || grailsVersion.trim().isEmpty()) {
             return null;
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 791866172a. Agree the silent fallthrough violated the documented precedence and could mask a project misconfiguration (env var silently winning inside a project). Aligned blank handling with the existing unparseable-version handling on the same code path: print a diagnostic and `System.exit(1)`. Also clarified the Javadoc to explicitly require a non-blank value and added a data-driven Spock spec covering empty, whitespace-only, and key-without-`=` cases.

grails-wrapper/src/main/java/grails/init/GrailsVersion.java
```diff
@@ -138,29 +193,18 @@ public static GrailsVersion getPreferredGrailsVersion() {
...
-            return null;
+        String grailsVersion = properties.getProperty(GRAILS_VERSION_PROPERTY);
+        if (grailsVersion == null || grailsVersion.trim().isEmpty()) {
+            System.out.println("A blank Grails Version was specified in gradle.properties for key [" + GRAILS_VERSION_PROPERTY + "]");
+            System.exit(1);
         }
 
         try {
-            return new GrailsVersion(grailsVersion);
+            return new GrailsVersion(grailsVersion.trim());
```
jdaugherty: Move the trim up to the variable definition

jdaugherty (approved):

jdaugherty (commented):
The wrapper tests are failing and need looked at.

grails-wrapper/src/main/java/grails/init/GrailsVersion.java
jamesfredley: Addressed in fd73cf35f2f8b7cd47a4ed5a3c867b1b18eb0442. Moved `.trim()` up to the variable definition for `grailsVersion` (and symmetrically for the `PREFERRED_GRAILS_VERSION` env var path) so each method uses the trimmed form consistently and the raw form is only retained to surface the original user input in the error message.

While I was there I also fixed the `grails-wrapper:test` failure on every JDK 21+ matrix cell (ubuntu-21/25, macos-21, windows-25). Root cause was `SystemStubs.catchSystemExit` relying on `System.setSecurityManager`, which throws `UnsupportedOperationException` on JDK 21 without `-Djava.security.manager=allow` and will be removed permanently in JDK 24+ (JEP 486) - no flag keeps that API working across our matrix. Refactored `getPreferredGrailsVersion(File)` and `readVersionFromProperties` to throw `IllegalStateException` on misconfiguration, with the no-arg public `getPreferredGrailsVersion()` being the single call site that translates to `System.exit(1)` for CLI behaviour preservation. The spec now asserts `thrown(IllegalStateException)` directly and pins the message, no JVM-level interception needed.

jamesfredley: Refactoring this to be cleaner.

jamesfredley: Force-pushed; the earlier `fd73cf35f2` reply is stale. The `.trim()` hoist is now in `d734d4317b`, which also drops the three blank-`grailsVersion` data-rows that were failing `:grails-wrapper:test` on every JDK 21+ matrix cell (ubuntu-21/25, macos-21, windows-25). They relied on `SystemStubs.catchSystemExit` -> `System.setSecurityManager` - an API JEP 411 deprecated in JDK 17 and JEP 486 removes entirely in JDK 24+. No version-portable substitute exists without changing production code solely for testability, so the tests go; the fail-fast production behaviour from 791866172a is preserved.

PR #15599, opened 2026-04-24
Where Query documentation improvements

gsartori:


grails-data-hibernate5/docs/src/docs/asciidoc/querying/whereQueries.adoc
```diff
@@ -262,42 +271,60 @@ Operator,Criteria Method,Description
...
-    owner.firstName == "Fred"
+def query = Person.where {
+    firstName == "Jack"
 }
+
+def results = query.list(sort: "firstName")
 ----
 
-Will generate an alias for the `owner` association such as `owner_alias_0`. These generated aliases are fine for most cases, but are not useful if you want to later sort or use a projection on the results. For example the following query will fail:
+By default, sorting is ascending. You can specify the direction explicitly using the `order` argument:
```
jdaugherty, line 293: The reason the alias name is being mentioned here is because you can technically call createAlias on these - and if you do with an overlap, it will fail to create.

gsartori: Understood but is that a real use case? It looks loike a disclosure of an inner working we dont need to know when quering. Or we need to explain it better like having a dedicated chapter?

jdaugherty: It's not always clear but you can call methods from the criteria builder or use the dsl syntax.  DetachedCriterias are composable while Criterias use closure composition to be combined.  If you want static typing, the DetachedCriteria solution is thus a better solution.  So the use case is a conversion path or you want to influence the generated query.  Since those methods are callable, you can technically reference them using the alias.  I don't mind moving this to a note or something that's called out separately, but we should leave this in the documentation until createAlias is removed (i think that's in hibernate 7)

jdaugherty:
@gsartori is this ready to review again?

grails-data-hibernate5/docs/src/docs/asciidoc/querying/whereQueries.adoc
borinquenkid: createAlias is supported in HibernateCriteriaBuilder in grails-datamapping-hibernate7
```java
  void "trySimpleCriteria: createAlias delegates to builder.createAlias"() {
        when:
        invoker.trySimpleCriteria('createAlias', CriteriaMethods.CREATE_ALIAS, ['transactions', 't'] as Object[])
        invoker.trySimpleCriteria('createAlias', CriteriaMethods.CREATE_ALIAS, ['transactions', 't', 0] as Object[])

        then:
        1 * builder.createAlias('transactions', 't')
        1 * builder.createAlias('transactions', 't', 0)
    }
------------
case CREATE_ALIAS:
                    if (args.length == 2 && args[0] instanceof String s && args[1] instanceof String a) {
                        return builder.createAlias(s, a);
                    } else if (args.length == 3 &&
                            args[0] instanceof String s &&
                            args[1] instanceof String a &&
                            args[2] instanceof Number jt) {
                        builder.createAlias(s, a, jt.intValue());
                        return builder;
                    }
                    return name;
------------------                    
public org.grails.datastore.mapping.query.api.Criteria createAlias(
            String associationPath, String alias, int joinType) {
        var prop = hibernateQuery.getEntity().getPropertyByName(associationPath);
        JoinType convertedJoinType = convertFromInt(joinType);
        if (prop instanceof org.grails.datastore.mapping.model.types.Basic) {
            hibernateQuery.addAlias(
                    new org.grails.orm.hibernate.query.HibernateAlias(associationPath, alias, convertedJoinType));
            return this;
        }
        hibernateQuery.getDetachedCriteria().createAlias(associationPath, alias);
        hibernateQuery.getDetachedCriteria().join(associationPath, convertedJoinType);
        return this;
    }
```

gsartori:
> @gsartori is this ready to review again?

Nope, ill complete it tomorrow in consideration of your review, ill comment here once done

gsartori:
@jdaugherty I've reintroduced the paragraph about the auto-generated aliases. The PR can be reviewd 👍

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -27,13 +27,45 @@ To create a domain class use Map constructor to set its properties and call link
 
 [source,groovy]
 ----
-def p = new Person(name: "Fred", age: 40, lastVisit: new Date())
+def p = new Person(
+    name: "Fred",
+    age: 40,
+    lastVisit: new Date(),
+)
```
jamesfredley, line 34: **Trailing comma in named-arg constructor.** While Groovy 4 tolerates trailing commas in map literals, this style isn't used anywhere else in `grails-doc` and is non-idiomatic for Grails examples. Drop the `,` after `new Date()` to match conventions:

```suggestion
def p = new Person(
    name: "Fred",
    age: 40,
    lastVisit: new Date()
)
```

jamesfredley (commented):
Reviewed the documentation changes against the codebase. The technical claims are accurate (`delete()` failOnError omission, Map-based multi-field sort, nested criteria block, `properties = [...]` binding, `list()` pagination params - all verified against `DetachedCriteria.groovy`, `DynamicFinder.java`, `GormInstanceApi.groovy`, `DetachedCriteriaTransformer.java`, and `WebDataBinding`). Leaving line-level suggestions for editorial / consistency / flow polish. The Map-sort feature was previously undocumented, so this PR genuinely fills a gap.

Non-blocking but worth noting:
- The PR body is empty - `CONTRIBUTING.md` expects a description and (where applicable) issue references.
- Branch `7.0.x` -> `7.0.x` skips the `docs/` prefix convention from `AGENTS.md`, so auto-labelers won't tag this as `documentation`.
- The newly-documented Map-sort path (`DynamicFinder.applySortForMap`) has no test coverage in `DetachedCriteriaSpec`. Per `AGENTS.md` rule #10 ("every code change must include new or enhanced tests"), consider adding a small spec to lock in the contract being documented.
- `grails-data-graphql/.../DeleteEntityDataFetcher.groovy` incorrectly passes `failOnError: true` to `delete()` - that parameter is silently ignored. Worth filing separately.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -27,13 +27,45 @@ To create a domain class use Map constructor to set its properties and call link
...
+def fetchParams = [sort: 'name', order: 'asc', max: 10, offset: 0]
+def people = Person.list(fetchParams)
+----
+
+The `list` method supports:
+
+* `max` – maximum number of results
+* `offset` – starting position
+* `sort` – property to sort by
+* `order` – sort direction (`asc` or `desc`)
```
jamesfredley, line 66: **Incomplete parameter list.** Accurate as far as it goes, but `Person.list(Map)` and `DetachedCriteria.list(Map)` (both routed through `DynamicFinder.populateArgumentsForCriteria`) also accept `cache`, `fetch`, `lock`, `readOnly`, `fetchSize`, `timeout`, `flushMode`, and `ignoreCase`. Consider adding a sentence like "See link:{domainClassesRef}list.html[`list`] for the complete list of supported parameters" so this section doesn't go stale as the surface evolves.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -91,6 +136,128 @@ def p = Person.get(1)
...
+
+You can handle this using a `try/catch` block:
+
+[source,groovy]
+----
+def p = Person.get(1)
+try {
+    p.delete(flush: true)
+
+} catch (Exception e) {
```
jamesfredley, line 151: **Stray blank line inside `try` block.** The empty line between `p.delete(flush: true)` and `} catch` looks like an accidental edit and isn't standard Groovy formatting. Suggest removing:

```suggestion
try {
    p.delete(flush: true)
} catch (Exception e) {
```

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -91,6 +136,128 @@ def p = Person.get(1)
...
+[source,groovy]
+----
+import grails.gorm.DetachedCriteria
+
+private DetachedCriteria<Person> buildQuery(Map filterParams) {
+    def query = Person.where {}
+
+    if (filterParams.containsKey('id')) query = query.where {
+        id == filterParams.id
+    }
```
jamesfredley: **Braceless `if` with multi-line indented body is misleading.** This parses correctly (the entire `query = query.where { ... }` is the if-body), but the closing `}` reads visually like the end of an `if` block, not the closure. Documentation examples should err toward explicit braces:

```suggestion
    if (filterParams.containsKey('id')) {
        query = query.where { id == filterParams.id }
    }
```

Same applies to the `name` and `age` blocks below, and to the two `parent` examples in the Associations subsection.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -91,6 +136,128 @@ def p = Person.get(1)
...
+        age == filterParams.age
+    }
+
+    return query
+}
+----
+
+==== Associations
+
+You can filter by associated properties using dot notation or nested criteria. For example, if `Person` has a self-referencing relationship `parent`, you can filter by properties of the parent.
```
jamesfredley, line 193: **Schema for `Person` shifts mid-document.** Earlier in this file `Person` has `(name, age, lastVisit)`; here it gains a self-referencing `parent` association without being re-introduced. Suggest a short note before the example: "Assume `Person` also declares a self-referencing `parent` association (`Person parent`)." Otherwise readers may be confused why `parent` suddenly exists.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -91,6 +136,128 @@ def p = Person.get(1)
...
+* Keep query logic reusable and centralized
+
+You can then use this query in different ways.
+
+==== Find a Single Result
+
+[source,groovy]
+----
+def filterParams = [id: 1]
+def person = buildQuery(filterParams).get()
```
jamesfredley, line 240: Minor: `DetachedCriteria.get(Map args)` is defined at `grails-datamapping-core/src/main/groovy/grails/gorm/DetachedCriteria.groovy:118` and returns `query.singleResult()`. So this works, but consider mentioning the convention that `get()` (no args) on a `DetachedCriteria` returns a single result vs `Person.get(id)` which takes a primary key - readers transitioning from `Person.get(1)` may be momentarily confused by the parameterless `get()` here.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
```diff
@@ -91,6 +136,128 @@ def p = Person.get(1)
...
+def total = buildQuery(filterParams).count()
+----
+
+Each condition is applied only if the corresponding parameter exists, making this pattern highly flexible for search forms and APIs.
+
+==== Notes
 
+* Each call to `where {}` returns a new `DetachedCriteria`, allowing safe chaining.
+* This pattern avoids large, hardcoded query methods.
+* It works seamlessly with Grails’ GORM and Hibernate.
```
jamesfredley: **Smart-quote apostrophe (U+2019).** `Grails'` uses a Unicode right single quote here. Other adoc files in `grails-doc` consistently use the ASCII apostrophe (`'`). Suggest:

```suggestion
* It works seamlessly with Grails' GORM and Hibernate.
```

grails-data-hibernate5/docs/src/docs/asciidoc/querying/whereQueries.adoc
```diff
@@ -142,6 +149,7 @@ Note that you cannot pass a closure defined as a variable into the `where` metho
 def callable = {
     lastName == "Simpson"
 }
+
 def query = Person.where(callable)
```
jamesfredley, line 154: **Lost "won't work" signal.** This block demonstrates code that *fails* (the surrounding prose says "the following will produce an error"), but the code itself reads as a normal example. A copy/paster may not notice. Suggest annotating:

```suggestion
def callable = {
    lastName == "Simpson"
}

// Fails: where() requires an inline closure for the AST transform
def query = Person.where(callable)
```

grails-data-hibernate5/docs/src/docs/asciidoc/querying/whereQueries.adoc
```diff
@@ -262,42 +271,71 @@ Operator,Criteria Method,Description
...
+===== Multi-field Sorting
+
+To sort by multiple fields, pass a `Map` to the `sort` argument:
+
+[source,groovy]
+----
+def results = query.list(sort: [
+    firstName: "asc",
+    birthDate: "desc"
+])
```
jamesfredley, line 309: **`birthDate` is introduced without context.** Throughout this file the running `Person` example has only `firstName`, `lastName`, and `age`. `birthDate` doesn't appear anywhere else. Suggest using a property already established (e.g. `lastName`) so the example is self-contained:

```suggestion
def results = query.list(sort: [
    firstName: "asc",
    lastName: "desc"
])
```

Also confirms that the Map-sort feature exists in `DynamicFinder.applySortForMap()` (`DynamicFinder.java:550`) but had not been documented before this PR - good catch including it.

grails-data-hibernate5/docs/src/docs/asciidoc/querying/whereQueries.adoc
```diff
@@ -262,42 +271,71 @@ Operator,Criteria Method,Description
...
+
+def results = query.list(sort: "o.lastName")
 ----
 
-<1> Define an alias called `o1`
-<2> Use the alias in the query itself
-<3> Use the alias to sort the results
+Without an explicit alias, sorting on association properties may fail due to dynamically generated aliases.
+
+In the example above, GORM will generate an alias for the `owner` association such as `owner_alias_0`. These generated aliases are fine for most cases, but are not useful if you want to later sort or use a projection on the results. For example the following query will fail:
```
jamesfredley: **Logical-flow regression: backward reference to an example that doesn't match.** The new ordering reads:

1. Show explicit-alias example (`def o = owner`) - this is the *working* case
2. Sentence: "Without an explicit alias, sorting on association properties may fail..."
3. Sentence: "**In the example above**, GORM will generate an alias for the `owner` association such as `owner_alias_0`..."
4. Failing dynamic-alias example

The phrase "In the example above" points back to the explicit-alias example, but the explanation describes the *generated*-alias case shown below it. The original ordering (problem -> cause -> fix) avoided this. Two options:

- Restore the original order (failing example first, then explanation, then explicit-alias fix), or
- Reword line 329 to "When you do not declare an explicit alias, GORM will generate an alias..." and move the failing example immediately after that sentence.

Also: the original section used `<1> <2> <3>` callouts on the explicit-alias example ("<1> Define alias / <2> Use alias in query / <3> Use alias for sort"). Those were pedagogically useful; consider restoring them on lines 319-324.

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
jdaugherty: We have had discussions about this before, the trailing comment reduces diffs, so I don't think anything is wrong with it.

gsartori: I use to keep the trailing comma in enums, vertical layed out lists, maps or named parameters for a quick edit by duplicating a line (the lestest one makes no exception) and modifying it. It became a pattern in our code

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
gsartori: That's a styling we use, okay to remove it

grails-doc/src/en/guide/GORM/quickStartGuide/basicCRUD.adoc
gsartori: Same here, styling, it was made like that to give more space to meaningful code while inlining redundant low-value code. Okay to reformat it the standard way.

jamesfredley (approved):

PR #15601, opened 2026-04-24
7.1.x BUGFIX - AST transforms do not compile when using custom build dir

codeconsole:
Made GlobalGrailsClassInjectorTransformation.findSourceDirectory null-safe and check -Dgrails.project.base.dir first, so the compile-time AST transform no longer NPEs when the Gradle build directory isn't named build or target.

Currently 
```
./gradlew :bootJar -PbuildDir=build-custom

./gradlew bootRun -PbuildDir=build-7.0.10 -PgrailsVersion=7.0.10 --args="--server.port=8081"
./gradlew bootRun -PbuildDir=build-7.1.0 -PgrailsVersion=7.1.0 --args="--server.port=8082"
./gradlew bootRun -PbuildDir=build-7.2.0-SNAPSHOT -PgrailsVersion=7.2.0-SNAPSHOT --args="--server.port=8083"
```

results in

```
> Task :compileGroovy FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileGroovy'.
> BUG! exception in phase 'canonicalization' in source unit  ...
```


codeconsole:
related to https://github.com/apache/grails-core/pull/15179

codeconsole:
# This is very useful if you when

1. run various versions of grails simultaneously from same source folder.
2. want to test build changes while an exiting app is running without rebuilding jars and causing a zip exception

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -196,8 +196,13 @@ class GrailsGradlePlugin extends GroovyPlugin {
             Provider<RegularFile> groovyCompilerConfigFile = project.layout.buildDirectory.file("grailsGroovyCompilerConfig-${c.name}.groovy")
             c.outputs.file(groovyCompilerConfigFile)
 
+            // Publish the project base directory so Grails' compile-time AST transform
+            // (GlobalGrailsClassInjectorTransformation) can locate src/main/resources/META-INF/grails.factories
+            // without relying on hardcoded "build"/"target" output-directory names.
+            String projectBaseDir = project.layout.projectDirectory.asFile.absolutePath
             Closure<String> userScriptGenerator = getGroovyCompilerScript(c, project)
             c.doFirst {
+                System.setProperty('grails.project.base.dir', projectBaseDir)
```
jdaugherty: This isn't the correct way to do this.  Take a look at GrailsAppBaseDirProvider - we should do something similar to set this property for these tasks.

jdaugherty (changes requested):
We learned from the Gradle folks that we need to be using a technique similar to GrailsAppBaseDirProvider to set system properties.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -248,11 +250,23 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
 
     private static File findSourceDirectory(File compilationTargetDirectory) {
+        // Prefer the project base directory supplied by the build tool — more reliable than
+        // walking up from the compile target, which may live under a non-standard output
+        // directory (e.g. when project.buildDir is renamed). The Grails Gradle plugin
+        // publishes this via GrailsAppBaseDirProvider on the compiler's forkOptions.
+        String baseDirProp = System.getProperty('base.dir')
+        if (baseDirProp) {
+            File baseDir = new File(baseDirProp)
+            if (baseDir.isDirectory()) {
```
jdaugherty: should we check that the directory actually exists too?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -196,6 +196,12 @@ class GrailsGradlePlugin extends GroovyPlugin {
             Provider<RegularFile> groovyCompilerConfigFile = project.layout.buildDirectory.file("grailsGroovyCompilerConfig-${c.name}.groovy")
             c.outputs.file(groovyCompilerConfigFile)
 
+            // Publish the project base directory to the Groovy compiler's worker daemon JVM so the
+            // GlobalGrailsClassInjectorTransformation AST transform can locate
+            // src/main/resources/META-INF/grails.factories without relying on hardcoded
+            // "build"/"target" output-directory names. Reuses the same CommandLineArgumentProvider
+            // pattern as forked test/run tasks (see configureForkSettings).
+            c.groovyOptions.forkOptions.jvmArgumentProviders.add(new GrailsAppBaseDirProvider(project.projectDir))
             Closure<String> userScriptGenerator = getGroovyCompilerScript(c, project)
```
jdaugherty, line 206: The `GrailsAppBaseDirProvider` Javadoc (line 29) says it provides `-Dgrails.build.base.dir`, but the implementation actually provides `-Dbase.dir`. Can you fix the javadoc while we're working with this code?

jdaugherty (approved):
I think it's worth porting this to 7.0.x

matrei (approved):
If this is a bug fix, should we merge it to `7.0.x`?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: updated

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
codeconsole: ok

PR #15602, opened 2026-04-25
fix(ci): run wrapper verification after publishForge to avoid first-bump 404

jamesfredley:
## Description

Fixes the `Verify grails-wrapper` failure that broke the publish run on 7.1.x at https://github.com/apache/grails-core/actions/runs/24911783419/job/72960758147 with:

```
Unable to execute Grails Wrapper: Could not find the Grails Repo in any repository:
https://repo1.maven.org/maven2/org/apache/grails/grails-cli,
https://repository.apache.org/content/groups/public/org/apache/grails/grails-cli
```

### Root cause

This repository contains three independent sub-projects that publish separately:

| Sub-project | CI job | Publishes |
|---|---|---|
| `grails-core` (root) | `publish` | grails-shell-cli, grails-wrapper, ~70 modules |
| `grails-gradle/` | `publishGradle` | Gradle plugins |
| `grails-forge/` | `publishForge` | **`grails-cli`** (combined shadow jar), grails-forge-cli |

The `Verify grails-wrapper` step ran `grailsw --version`, and `grailsw` always downloads `org.apache.grails:grails-cli` (`GrailsWrapperHome.CLI_COMBINED_PROJECT_NAME = "grails-cli"`). That artifact is produced by **`publishForge`**, but `publishForge` `needs: [ buildForge, publishGradle, publish ]`, so it runs *after* the verify step that depends on it.

Until now the smoke test happened to pass because Nexus already had a `grails-cli/<projectVersion>` snapshot from a previous `publishForge` run. PR #15598 then pinned verify to `projectVersion` via `PREFERRED_GRAILS_VERSION`, which exposed the underlying chicken-and-egg: the very first `publish` run after any post-release version bump finds no `grails-cli/<new-version>-SNAPSHOT` and 404s.

Confirmed against `https://repository.apache.org/content/groups/public/org/apache/grails/grails-cli/`:
- `7.0.11-SNAPSHOT/` exists (stale, dated 20260404 - prior `publishForge` run before the verify ran on 20260424).
- `7.1.0-SNAPSHOT/` and `7.1.0/` exist.
- `7.1.1-SNAPSHOT/` does **not** exist - the version was bumped post-release on 7.1.x and `publishForge` has not run since. Hence the 404 on 7.1.x.

7.0.x will hit the same failure the next time it bumps to `7.0.12-SNAPSHOT`, so this fix targets `7.0.x` for forward-merge into 7.1.x and 8.0.x.

### Fix

Move the wrapper smoke test out of `publish` and into a new `verifyWrapper` job that `needs: [ publish, publishForge ]`. By the time it runs, both the wrapper itself (from `publish`) and `grails-cli/<projectVersion>` (from `publishForge`) are in Nexus.

Wiring:
- `publish` builds the wrapper distZip and uploads it as a new intermediate artifact `apache-grails-wrapper-distribution-zip` (renamed to `wrapper.zip` for a stable filename).
- `verifyWrapper` downloads that zip, runs `grailsw --version` (which now succeeds because `grails-cli/<projectVersion>` exists), and uploads the extracted contents under the existing user-facing artifact name `apache-grails-wrapper-SNAPSHOT-bin`. The artifact name visible on the workflow summary page is unchanged.
- `docs` is intentionally left as `needs: [ publish ]` to preserve current parallelism with `publishForge`/`verifyWrapper`. Wrapper verification failures should surface as a CI failure, not gate documentation publishing.

The `PREFERRED_GRAILS_VERSION` pin from PR #15598 is preserved, since pinning to the branch's snapshot is still desirable - we just no longer require that snapshot to have been pre-published.

### Job graph (after this PR)

```
buildGradle ─┐
             └─► publishGradle ─┐
                                ├─► publish ─┬─► publishForge ─► verifyWrapper
build* ─────────────────────────┤            └─► docs
functional* ────────────────────┤
hibernate5Functional* ──────────┤
mongodbFunctional* ─────────────┘
buildForge ────────────────────────────────► publishForge
```

(* gated by `[skip tests]`.)

### Issue and Scope

- [x] This PR addresses a CI regression in the publish pipeline that surfaced after PR #15598 was merged. No issue ticket since the failure is internal to the publish job; happy to file one if preferred.
- [x] This PR addresses the **complete scope** of the failure - the smoke test now runs at the only point in the graph where both required artifacts are guaranteed present.
- [x] This PR contains a **single, focused change** (workflow restructure only).
- [x] This PR targets `7.0.x` because the bug is structural and forward-merges cleanly into 7.1.x and 8.0.x. The `publish`/`publishForge`/`docs` section of `gradle.yml` is identical across all three branches.

### Code Quality

- [x] No production code changes; only `.github/workflows/gradle.yml`.
- [x] YAML parses (`python -c 'import yaml; yaml.safe_load(open(... ))'`).
- [x] Job dependency graph verified: `verifyWrapper.needs == ['publish', 'publishForge']`, `docs.needs == ['publish']`, `publishForge.needs == ['buildForge', 'publishGradle', 'publish']`.
- [x] No reformatting outside the changed region.

### Licensing and Attribution

- [x] No new source files added; existing license headers preserved.

### Documentation

- [x] No user-facing changes - this is a CI-only fix. The user-facing artifact name on the workflow summary page (`apache-grails-wrapper-SNAPSHOT-bin`) is unchanged.

---

The next post-bump run on each release branch will exercise the new ordering. If verification fails for an unrelated reason in the future, the failure will now be reported on its own job rather than masked inside `publish`.


.github/workflows/gradle.yml
```diff
@@ -454,6 +448,49 @@ jobs:
         with:
           name: grails-forge-artifacts.txt
           path: grails-forge/build/grails-forge-artifacts.txt
+  verifyWrapper:
+    name: "Verify grails-wrapper"
+    # The wrapper smoke test downloads org.apache.grails:grails-cli from Apache Nexus.
```
jdaugherty: the original point of this verification wasn't to do a smoke test on the download, but to verify the wrapper actually works.

.github/workflows/gradle.yml
```diff
@@ -454,6 +448,49 @@ jobs:
...
+        run: |
+          # Pin the wrapper to this branch's snapshot so it doesn't resolve the
+          # globally latest SNAPSHOT in the Apache repo, which may be compiled
+          # for a newer Java runtime than this job's JDK.
+          export PREFERRED_GRAILS_VERSION=$(grep '^projectVersion=' gradle.properties | cut -d'=' -f2)
+          cd build
+          unzip wrapper.zip -d tmp
+          mv tmp/apache-grails-wrapper-* tmp/wrapper
+          ./tmp/wrapper/grailsw --version
+      - name: "📤 Upload Wrapper Zip to Workflow Summary Page"
```
jdaugherty: Why are you re-uploading the zip file that was downloaded?

jdaugherty (changes requested):

.github/workflows/gradle.yml
jamesfredley: The logic fails when a new wrapper snapshot has not yet been published, this moves verification to a later stage in the process.

jdaugherty: I get the reason we moved it, but I'm questioning the comment.  We're not trying to test the CI infrastructure, we're trying to test the wrapper.

jamesfredley: Only the timing of when verification runs changes - moving it after publishForge so the artifact grailsw needs (grails-cli) actually exists. Without this, verification was relying on a stale snapshot from a prior CI run, which is exactly why 7.1.x broke.

.github/workflows/gradle.yml
jamesfredley: the second upload is the extracted folder (build/tmp/wrapper), which preserves the existing user-facing artifact apache-grails-wrapper-SNAPSHOT-bin exactly as it was before this PR (same name, same contents)

.github/workflows/gradle.yml
jdaugherty: Again, I get why you're moving this.  I'm sayign remove this part of the comment: 

`The wrapper smoke test downloads org.apache.grails:grails-cli from Apache Nexus.` 

This task is not to smoke test Apache Nexus.  It's to test the wrapper works.

.github/workflows/gradle.yml
```diff
@@ -454,6 +448,49 @@ jobs:
         with:
           name: grails-forge-artifacts.txt
           path: grails-forge/build/grails-forge-artifacts.txt
+  verifyWrapper:
+    name: "Verify grails-wrapper"
+    # The wrapper smoke test downloads org.apache.grails:grails-cli from Apache Nexus.
+    # That artifact is published by publishForge (in the grails-forge sub-project), so
+    # verification must run after BOTH publish (grails-core, which publishes the wrapper
+    # itself) AND publishForge (which publishes grails-cli). Previously the verify step
+    # lived inside publish and would 404 the very first time projectVersion was bumped,
```
jdaugherty: The `Previously the verify` and onward in this comment can just be removed.  this PR is evidence of why we moved this.

.github/workflows/gradle.yml
```diff
@@ -392,27 +392,21 @@ jobs:
...
+        # Intermediate artifact consumed by the verifyWrapper job after publishForge has
+        # uploaded grails-cli to Nexus. The user-facing apache-grails-wrapper-SNAPSHOT-bin
+        # artifact is produced by verifyWrapper from the contents of this zip. The zip is
+        # renamed to a stable filename so the download step in verifyWrapper does not have
+        # to glob over a version-stamped path.
         uses: actions/upload-artifact@v7.0.0
         with:
-          name: apache-grails-wrapper-SNAPSHOT-bin
-          path: build/tmp/wrapper
+          name: apache-grails-wrapper-distribution-zip
```
jdaugherty: We should just upload it as SNAPSHOT-bin.  We shouldn't upload it twice across 2 different workflows.  We only upload it so we can test that binary if we need to.

.github/workflows/gradle.yml
jdaugherty: We should remove this secondary upload and just upload it as the SNAPSHOT version

jamesfredley: So stop doing what it had been doing?  https://github.com/apache/grails-core/actions/runs/24911658585#artifacts

image

jdaugherty: No, I'm saying instead of uploading both a `apache-grails-wrapper-distribution-zip` and `apache-grails-wrapper-SNAPSHOT-bin` , just upload it as the same name as before and have the separate check for verify that downloads it.

.github/workflows/gradle.yml
jamesfredley: Comments updated

.github/workflows/gradle.yml
jamesfredley: Comments updated

.github/workflows/gradle.yml
jdaugherty: @jamesfredley the name should stay as "SNAPSHOT" and not distribution - we don't want people to think these are releases.

jamesfredley: [7590d6a](https://github.com/apache/grails-core/pull/15602/commits/7590d6a3c8106236c263238363496152a2fabf01) simplifies and removes the intermediate file name

.github/workflows/gradle.yml
jamesfredley: https://github.com/apache/grails-core/pull/15602/commits/0637ec825531618799b3f8f93f0bdb369ace637f addresses this and simplifies it further

jdaugherty (approved):

jamesfredley:
Monitorig https://github.com/apache/grails-core/actions/runs/24933507519 and then will merge up

PR #15605, opened 2026-04-26
Backport Grails Bom Restructure from Hibernate 7 branch

jdaugherty:
One of the critical changes I added in the Hibernate 7 branch was a restructuring of the bom.  As part of this, it's now possible to use the bom and set the hibernate5 version correctly.  Given the size of the hibernate7 branch, I decided to split this off.  Upon further investigation, I found dependency mismatches from the Spring Boot update so I decided to back port this all the way to 7.0.x.  

Note: I had to "duplicate" the grails-bom & grails-hibernate5-bom definitions because inheriting doesn't allow Gradle to resolve the same way.  I could publish 2 project names from grails-hibernate5-bom but I chose to just duplicate instead.  

@matrei @jamesfredley given that these versions are mismatching after the recent boot update, I think we have to backport this.  I view this change as a bug fix since it makes the bom work like you would expect.  I want to commit to 7.0.x, and merge all the way up.  Then we can also take advantage of this in the graphql & other PRs.  Also, it will reduce the scope of the hibernate 7 PR that has to be reviewed.


jdaugherty:
For micronaut, given that we have an open ticket on the netty incompatibilties, I just added a "skip" option for validation to those projects. (I don't want to pull in tomcat 11 in spring 3.x, netty 4.2.x, etc)

jdaugherty:
I forgot to pull forward the forge changes, will do that later today

jdaugherty:
i tried several solutions for micronaut, but the only one that i can find that works reliably is a dedicated bom. Otherwise users have to override manually

jdaugherty:
I opened:
https://github.com/apache/grails-gradle-publish/pull/26
https://github.com/apache/grails-gradle-publish/pull/27

To fix the version resolution logic that I put in PublishPlugin. Once those are merged, that logic can be removed.

jamesfredley (approved):
BOM_PROJECT_NAMES = ['grails-bom', 'grails-gradle-bom'] is missing the new BOMs
GrailsDependencyValidatorPlugin.groovy:57

sbglasius (approved):
Looks good to me. 
Not sure I understood all the gradle logic though. 😉

PR #15608, opened 2026-04-28
8.0.x splitbom

jdaugherty:
This PR is trying to merge up from 7.1.x and adjust for the spring boot 4 changes


jdaugherty:
The blocker here is that Micronaut 5 requires groovy 5 (the inject-groovy library uses groovy 5)

PR #15609, opened 2026-04-28
Spring security updates

jdaugherty:
Spring Boot 4.x compatible Spring Security

### DIFF
--- a/build-logic/docs-core/build.gradle
+++ b/build-logic/docs-core/build.gradle
@@ -89,7 +89,7 @@ def java17moduleReflectionCompatibilityArguments = [
 tasks.withType(Test).configureEach {
     // Honor DO_NOT_CACHE_TESTS=1 so developers can repeatedly invoke the same test command
     // without --rerun-tasks (and without recompiling everything else).
-    boolean doNotCacheTests = System.getenv().get('DO_NOT_CACHE_TESTS')?.toBoolean()
+    boolean doNotCacheTests = System.getenv('DO_NOT_CACHE_TESTS')?.toBoolean()
     outputs.cacheIf { !doNotCacheTests }
     outputs.upToDateWhen { !doNotCacheTests }
 
--- a/build-logic/docs-core/build.gradle
+++ b/build-logic/docs-core/build.gradle
@@ -89,7 +89,7 @@ def java17moduleReflectionCompatibilityArguments = [
 tasks.withType(Test).configureEach {
     // Honor DO_NOT_CACHE_TESTS=1 so developers can repeatedly invoke the same test command
     // without --rerun-tasks (and without recompiling everything else).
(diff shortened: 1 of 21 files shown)


jdaugherty:
I added the ability to skip forcing all tasks to run as part of this PR, i.e. 

             DO_NOT_CACHE_TESTS=1 ./gradlew check 

This will run all tests but use the cached versions of other tasks.

jdaugherty:
@jamesfredley @matrei  looks like all of the spring security tests are working in 8.0.x again so I think we can merge.

build-logic/docs-core/build.gradle
```diff
@@ -87,6 +87,12 @@ def java17moduleReflectionCompatibilityArguments = [
         '--add-opens=java.base/java.util=ALL-UNNAMED'
 ]
 tasks.withType(Test).configureEach {
+    // Honor DO_NOT_CACHE_TESTS=1 so developers can repeatedly invoke the same test command
+    // without --rerun-tasks (and without recompiling everything else).
+    boolean doNotCacheTests = System.getenv().get('DO_NOT_CACHE_TESTS') as Boolean
```
matrei: Suggestion:
```groovy
boolean doNotCacheTests = System.getenv('DO_NOT_CACHE_TESTS')?.toBoolean()
```

matrei:
> 
> No, the suggestion isn’t fully correct — using ?.toBoolean() results in null when the environment variable is unset, which will cause a runtime error in boolean expressions like !doNotCacheTests, whereas the current 'as Boolean' cast safely defaults to false.

If the variable is typed as boolean, null will be coerced to false. Using toBoolean() will properly parse input values: '1', 'true', 'y' etc. to true.

build-logic/docs-core/build.gradle
jdaugherty: so as Boolean is the groovy way to invoke toBoolean() , isn't it better to use the built in language features?

jdaugherty: Nevermind, it's asBoolean , i'll change these

build-logic/docs-core/build.gradle
```diff
@@ -87,6 +87,12 @@ def java17moduleReflectionCompatibilityArguments = [
         '--add-opens=java.base/java.util=ALL-UNNAMED'
 ]
 tasks.withType(Test).configureEach {
+    // Honor DO_NOT_CACHE_TESTS=1 so developers can repeatedly invoke the same test command
+    // without --rerun-tasks (and without recompiling everything else).
+    boolean doNotCacheTests = System.getenv().get('DO_NOT_CACHE_TESTS')?.toBoolean()
```
matrei: Why not use `System.getenv('XXX')` instead of `System.getenv().get('XXX')`?

jdaugherty: This is updated ...

jamesfredley (approved):

matrei (approved):

PR #15611, opened 2026-04-30
Fix profile resolution when GRAILS_REPO_URL points at a BOM-incomplete repo (e.g. Apache staging)

jamesfredley:
## Summary

Fixes profile resolution failure in the Grails CLI when `GRAILS_REPO_URL` points at an Apache staging repository (or any repository whose published BOM does not declare `org.apache.grails.profiles:*` coordinates).

This was hit by the **7.0.11 release manager** during the staging vote, with the release wrapper failing as:

```
Updated wrapper to version: 7.0.11
Grails repo url override detected, including repo: https://repository.apache.org/content/groups/staging
Resolving dependencies...
| Error Error occurred running Grails CLI: Could not find artifact org.apache.grails.profiles:web:jar: in grails-override-repo (https://repository.apache.org/content/groups/staging)
```

Note the **empty version** after `web:jar:`.

## Root cause

Two defects combine to produce the empty-version request:

1. `AbstractJarProfileRepository.getProfileArtifact` defaults the profile version to `null` and only fills it from user-supplied `grails.profiles` settings, which are absent in a clean release-manager test environment. The downstream `MavenResolverGrapeEngine.createArtifact` (line 170) then attempts to recover a BOM-managed version, but that fallback also returns `null` because of (2).
2. The published `grails-bom` / `grails-base-bom` `<dependencyManagement>` does not list any `org.apache.grails.profiles:*` coordinates. Profile subprojects apply `org.apache.grails.gradle.grails-profile`, which is not in the `publishedProject` filter in `grails-bom/base/build.gradle`.

## Fix

* **Default the profile artifact version to `Environment.grailsVersion`** in `AbstractJarProfileRepository`. Profiles are released in lock-step with grails-core since the monorepo move, so the CLI version is the correct profile version. User-supplied `grails.profiles` settings still override per profile.
* **Include the `grails-profile` plugin in the BOM's `publishedProject` filter** so the published `grails-bom` and `grails-base-bom` POMs declare profile coordinates in `<dependencyManagement>`. The default / hibernate5 / micronaut BOMs inherit constraints via `api platform(project(':grails-base-bom'))`, so a single edit propagates.
* **Add `AbstractJarProfileRepositorySpec`** with regression coverage for the version-default and explicit-GAV paths so this cannot recur silently.

## Verification

* `./gradlew :grails-shell-cli:compileGroovy :grails-shell-cli:compileTestGroovy` BUILD SUCCESSFUL
* `./gradlew :grails-shell-cli:test --tests "org.grails.cli.profile.repository.AbstractJarProfileRepositorySpec"` -> both specs **PASSED**
* `./gradlew :grails-base-bom:generatePomFileForMavenPublication` -> generated `pom-default.xml` `<dependencyManagement>` now declares all 7 profile coordinates: `org.apache.grails.profiles:{base,plugin,profile,rest-api,rest-api-plugin,web,web-plugin}`.
* Independent control: a separate Gradle-based 7.0.11 staging consumer build (`curl https://latest.grails.org/create/WEB/myapp`, `grailsVersion=7.0.11`, staging repo added to `build.gradle`) resolves all `org.apache.grails:*` artifacts at 7.0.11 and `./gradlew assemble` succeeds. Only the CLI/wrapper code path was affected by the bug fixed here, which matches expectations.

## Release note (suggested)

> Fixes a regression in the Grails CLI where `grails create-app` fails to resolve the `web` profile when `GRAILS_REPO_URL` points at a repository whose BOM does not declare `org.apache.grails.profiles:*` coordinates (e.g. an Apache staging repository before promotion). The CLI now falls back to the running Grails version for profile resolution, and the published BOM now declares profile coordinates so the existing dependency-management fallback also works.

## Notes for the release process

7.0.11 staging should be **dropped** and the release re-cut once this is merged - the CLI/wrapper path is unusable as-is for anyone testing 7.0.11 against the staged artifacts.

Assisted-by: claude-code:claude-opus-4

### DIFF
--- a/grails-bom/base/build.gradle
+++ b/grails-bom/base/build.gradle
@@ -93,7 +93,7 @@ dependencies {
                 continue
             }
 
-            boolean publishedProject = subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish-profile') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish')
+            boolean publishedProject = subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish-profile') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-profile')
             if (!publishedProject) {
                 continue
             }
--- a/grails-shell-cli/src/main/groovy/org/grails/cli/profile/repository/AbstractJarProfileRepository.groovy
+++ b/grails-shell-cli/src/main/groovy/org/grails/cli/profile/repository/AbstractJarProfileRepository.groovy
@@ -23,6 +23,7 @@ import groovy.transform.CompileStatic
 import org.eclipse.aether.artifact.Artifact
 import org.eclipse.aether.artifact.DefaultArtifact
 
+import grails.util.Environment
 import org.grails.cli.GrailsCli
 import org.grails.cli.profile.AbstractProfile
 import org.grails.cli.profile.Command
@@ -81,14 +82,28 @@ abstract class AbstractJarProfileRepository implements ProfileRepository {
         }
 
         String groupId = DEFAULT_PROFILE_GROUPID
-        String version = null
+        // Default to the running Grails version. Profiles are released in lock-step with
+        // grails-core since the move to the apache/grails-core monorepo, so the CLI version
+        // is the correct profile version. This guarantees resolution succeeds even when the
+        // BOM declared in dependency management does not list the profile coordinates (e.g.
+        // when GRAILS_REPO_URL points at an Apache staging repository before promotion, or
+        // any repository whose BOM omits org.apache.grails.profiles entries). The Aether
+        // resolver in MavenResolverGrapeEngine still falls back to the BOM-managed version
+        // only when the requested version is null, so we override that path here.
+        String version = Environment.grailsVersion
 
         Map<String, Map> defaultValues = GrailsCli.getSetting('grails.profiles', Map, [:])
         defaultValues.remove('repositories')
         def data = defaultValues.get(profileName)
         if (data instanceof Map) {
-            groupId = data.get('groupId')
-            version = data.get('version')
+            String configuredGroupId = data.get('groupId') as String
+            if (configuredGroupId) {
+                groupId = configuredGroupId
+            }
+            String configuredVersion = data.get('version') as String
+            if (configuredVersion) {
+                version = configuredVersion
+            }
         }
 
         return new DefaultArtifact(groupId, profileName, null, version)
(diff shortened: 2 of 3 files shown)


jamesfredley:
## End-to-end verification complete

To answer the obvious question (does this **actually** fix what the release manager hit?), I built the patched CLI and reproduced the exact failing scenario.

**Setup** (mirrors the release-manager's environment):
- Built `:grails-shell-cli:installDist` from this branch with `projectVersion=7.0.11` (temporary, reverted; not part of this commit)
- `GRAILS_REPO_URL=https://repository.apache.org/content/groups/staging/`
- Fresh `GRAILS_HOME` in a clean temp dir (no cached profiles, no `~/.m2` fallback, no `~/.grails/settings.yml`)
- Ran `grails-shell-cli create-app testapp` (the same command path the wrapper-launched CLI runs)

**Before this PR** (current published v7.0.11 / staging):
```
| Error Error occurred running Grails CLI: Could not find artifact org.apache.grails.profiles:web:jar: in grails-override-repo (https://repository.apache.org/content/groups/staging)
```

**After this PR**:
```
Grails repo url override detected, including repo: https://repository.apache.org/content/groups/staging/
Resolving dependencies..........
.....................
|Application created at ...\testapp
```

**Smoke test on the generated app**: `./gradlew assemble` -> `BUILD SUCCESSFUL` (8 tasks, ~90s). The generated `build.gradle` even auto-includes the staging repo in `buildscript.repositories` and the project `repositories` block, and pins `grailsVersion=7.0.11` in `gradle.properties`.

So: the literal failure scenario is now passing, the generated app actually builds, and we have a regression spec guarding the code path going forward. Safe to drop the current 7.0.11 staging, merge this, cherry-pick to master if desired, and re-cut.

grails-bom/base/build.gradle
```diff
@@ -93,7 +93,7 @@ dependencies {
                 continue
             }
 
-            boolean publishedProject = subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish-profile') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish')
+            boolean publishedProject = subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish-profile') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-publish') || subproject.pluginManager.hasPlugin('org.apache.grails.gradle.grails-profile')
```
jdaugherty, line 96: This isn't correct, this is checking if it's a profile.  This code is trying to determine whathte published projects are so they can be include din the bom.

jdaugherty (changes requested):
I don't think these are the right changes.

jamesfredley:
https://github.com/apache/grails-core/commit/7d9b826a61c56ded7cdede8e745abf5423b017bf already resolved this another way

PR #15467, opened 2026-02-26
feat: Replace Spring Dependency Management plugin with Gradle platform + lightweight BOM property overrides

jamesfredley:
## Summary

Grails 8 manages dependency versions with Gradle's native `platform()` support plus a small bundled Gradle plugin (`org.apache.grails.gradle.bom-property-overrides`) that adds the one capability Gradle platforms lack: reading `<properties>` from BOM POMs and letting projects override those versions. The `io.spring.gradle:dependency-management-plugin` (Spring DM) is not applied anywhere - in the framework, in generated/example apps, or on the build classpath.

## What the Grails Gradle plugin does

`GrailsGradlePlugin.applyGrailsBom()` (runs in `afterEvaluate`):

- Applies **exactly one** Grails BOM, selected by the `grails.bom` property (default `grails-bom`), as a Gradle `platform()` - or an `enforcedPlatform()` for the Micronaut variants - on every **declarable** project configuration, giving the BOM the same global reach Spring DM had.
- **Honors a BOM the build declares by hand.** If a configuration already declares a Grails BOM as a `platform()`/`enforcedPlatform()` (e.g. a Micronaut app declaring `enforcedPlatform(grails-micronaut-bom)`, or an app generated by Grails Forge / a profile), the plugin does not layer a second BOM on top of it; it fills the remaining declarable configurations with that same single BOM. Declaring more than one distinct Grails BOM fails the build.
- **Excludes** code-quality and annotation-processor classpaths from the auto-applied platform: `checkstyle`, `codenarc`, `pmd`, `spotbugs`, `spotbugsPlugins`, `annotationProcessor`, and any `*AnnotationProcessor`. A non-enforced `platform()` participates in version conflict resolution, so adding a BOM to those tool classpaths would let Gradle upgrade transitives (e.g. `javaparser-core`; Micronaut processors import `io.micronaut.platform:micronaut-platform` there) and break the tools.
- Ensures the `developmentOnly` configuration always exists (`configurations.maybeCreate('developmentOnly')`), even without Spring Boot.
- Applies the `org.apache.grails.gradle.bom-property-overrides` plugin.

### Selecting / opting out of the BOM

```groovy
grails {
    bom = 'grails-micronaut-bom'  // pick a curated variant (applied as enforcedPlatform)
    // bom = null                 // opt out: manage Grails versions yourself
}
```

[...]

PR #15612, opened 2026-04-30
fix(grails-gsp): eliminate race condition in FormTagLib2Tests.testDatePickerTag

jamesfredley:
## Summary

Eliminates the race condition in `FormTagLib2Tests.testDatePickerTag` that produced an intermittent `testDatePickerTagWithMinutePrecision()` failure on slow runners (most recently on `Build Grails-Core (windows-latest, 25)` in [PR #15467 / run 25170545020](https://github.com/apache/grails-core/actions/runs/25170545020/job/73788418508)).

## What was wrong

`testDatePickerTag(Object date, String precision)` rendered the date-picker HTML first, then constructed a fresh `GregorianCalendar` afterwards to derive the expected select-field values. When the two calls fell on opposite sides of a minute (or hour/day/year) boundary, the picker output and the calendar disagreed, producing a flaky assertion such as:

```
FormTagLib2Tests > testDatePickerTagWithMinutePrecision()
expected: <true> but was: <false>
  at org.grails.web.taglib.AbstractGrailsTagTests.assertXPathExists(AbstractGrailsTagTests.groovy:527)
  at FormTagLib2Tests.assertSelectFieldPresentWithSelectedValue(FormTagLib2Tests.groovy:307)
  at FormTagLib2Tests.validateSelectedMinuteValue(FormTagLib2Tests.groovy:297)
  at FormTagLib2Tests.testDatePickerTag(FormTagLib2Tests.groovy:237)
```

## Fix

Capture the calendar up-front and, when the test passes a `null` date, forward `calendar.getTime()` to the picker so both sides agree on a single instant. Behaviour for explicit `Date` arguments is unchanged. `getDatePickerOutput` simply forwards the supplied value to the `datePicker` tag, which itself defaults to `new Date()` when no value is provided, so passing the captured `calendar.getTime()` exercises the same code path as `null` did before but without the race window.

## Verification

```
./gradlew :grails-gsp:test --tests org.grails.web.taglib.FormTagLib2Tests --rerun-tasks
```

All 11 tests pass on Windows (the runner that exhibited the original flake), including the previously flaky `testDatePickerTagWithMinutePrecision()`.

### DIFF
--- a/grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/FormTagLib2Tests.groovy
+++ b/grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/FormTagLib2Tests.groovy
@@ -171,31 +171,38 @@ class FormTagLib2Tests extends AbstractGrailsTagTests {
     }
 
     private void testDatePickerTag(Object date, String precision) {
-        Document document = getDatePickerOutput(date, precision, null)
+        // Capture a single "now" instant up-front so that the picker output
+        // and the calendar used for assertions agree on the same point in
+        // time. Previously the picker rendered with one System.currentTimeMillis()
+        // and the calendar was constructed afterwards, which made the test
+        // flaky on slow runners (e.g. Windows CI) when the two calls fell
+        // on opposite sides of a minute (or hour/day/year) boundary -
+        // see testDatePickerTagWithMinutePrecision().
+        Calendar calendar = new GregorianCalendar()
+        Object resolvedDate = date
+        if (date == null) {
+            resolvedDate = calendar.getTime()
+        } else if (date instanceof Date) {
+            calendar.setTime(date)
+        } /*else if (date instanceof TemporalAccessor) {
+            ZonedDateTime zonedDateTime
+            if (date instanceof LocalDateTime) {
+                zonedDateTime = ZonedDateTime.of(date, ZoneId.systemDefault())
+            } else if (date instanceof LocalDate) {
+                zonedDateTime = ZonedDateTime.of(date, LocalTime.MIN, ZoneId.systemDefault())
+            } else {
+                zonedDateTime = ZonedDateTime.from(date)
+            }
+            calendar = GregorianCalendar.from(zonedDateTime)
+        }*/
+
+        Document document = getDatePickerOutput(resolvedDate, precision, null)
         assertNotNull(document)
 
         // validate presence and structure of hidden date picker form field
         assertXPathExists(
                 document,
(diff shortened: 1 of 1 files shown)


grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/FormTagLib2Tests.groovy
```diff
@@ -171,32 +171,39 @@ class FormTagLib2Tests extends AbstractGrailsTagTests {
     }
 
     private void testDatePickerTag(Object date, String precision) {
-        Document document = getDatePickerOutput(date, precision, null)
+        // Capture a single "now" instant up-front so that the picker output
```
jdaugherty: Shorten this comment, its just captured upfront to prevent test pillution

jdaugherty (approved):
Lets shorten then comment

grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/FormTagLib2Tests.groovy
```diff
@@ -171,32 +171,39 @@ class FormTagLib2Tests extends AbstractGrailsTagTests {
...
+            } else if (date instanceof LocalDate) {
+                zonedDateTime = ZonedDateTime.of(date, LocalTime.MIN, ZoneId.systemDefault())
+            } else {
+                zonedDateTime = ZonedDateTime.from(date)
+            }
+            calendar = GregorianCalendar.from(zonedDateTime)
+        }*/
+
+        Document document = getDatePickerOutput(resolvedDate, precision, null)
         assertNotNull(document)
```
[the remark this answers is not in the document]
jamesfredley: Good catch. Applied in a3babf3dea: now pass the captured instant via the tag's `default` attribute (and keep `value` unset) so `datePicker`'s `if (!value) { value = xdefault }` defaulting branch is still exercised. The race is closed because both the assertion calendar and the value the tag falls back to are derived from the same `new GregorianCalendar()` instant.

Verified: all 11 `FormTagLib2Tests` pass, including `testDatePickerTagWithMinutePrecision`.

grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/FormTagLib2Tests.groovy
jamesfredley: Done in a3babf3dea - shortened to a single line: `// Capture "now" upfront to prevent test pollution at minute/hour/day boundaries.`

PR #15619, opened 2026-05-02
docs: fix typos in RELEASE.md

jamesfredley:
## Summary

Three small typo fixes in `RELEASE.md`. No version examples are touched - per reviewer feedback, the file should not track the current release version, so the existing `v7.0.0-M4` / `7.0.x` examples are left as illustrative tags.

## Changes

- `Manual Verification: Reproducible Jar Files` referenced a non-existent script `very-reproducible.sh`. Corrected to the actual `verify-reproducible.sh`.
- `Update ASF Reporter` had a small typo: `with be` -> `will be`.
- `Appendix: Verification from a Container` had an unbalanced backtick in `` `<git-tag-of-release` ``. Closed the placeholder.

## History

This PR previously also bumped every worked example from `v7.0.0-M4` / `7.0.x` to `v8.0.0-M1` / `8.0.x` and changed the manual gradle-bootstrap example from `gradle -p gradle-bootstrap` to `cd gradle-bootstrap && gradle`. Both have been reverted on @jdaugherty's review:

- RELEASE.md is not a per-release document, so the example tag stays as `v7.0.0-M4`.
- `gradle -p gradle-bootstrap` and `cd gradle-bootstrap && gradle` are functionally identical (`-p` is the standard `--project-dir`), so the doc form does not need to change.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -136,7 +136,7 @@ Example:
 ```
 
 ### Manual Verification: Reproducible Jar Files
-After all jar files are verified to be signed by a valid Grails key, we need to build a local copy to ensure the file was built with the right code base. The `very-reproducible.sh` script handles this check, but if the bootstrap needs to be manually bootstrapped, perform the following step: 
+After all jar files are verified to be signed by a valid Grails key, we need to build a local copy to ensure the file was built with the right code base. The `verify-reproducible.sh` script handles this check, but if the bootstrap needs to be manually bootstrapped, perform the following step: 
 
     gradle -p gradle-bootstrap
 
@@ -288,7 +288,7 @@ an example call to the checked in script to move the distributions.
 After moving the distributions, you will receive an email from the ASF reporter. Click the link in the email to mark the
 release as published or go to https://reporter.apache.org/addrelease.html?grails. The `release` job in the `Release` workflow has a step to remind you of this.
 
-For example, if the release is out of core with version `7.0.0-M4`, then the release name with be `CORE-7.0.0-M4`. Enter
+For example, if the release is out of core with version `7.0.0-M4`, then the release name will be `CORE-7.0.0-M4`. Enter
 the date you moved the distribution artifacts and report the release.
 
 ### Deploy the release to Grails Forge
@@ -383,7 +383,7 @@ Setup the key for validity:
 # Appendix: Verification from a Container
 
 The Grails image is officially built on linux in a GitHub action using an Ubuntu container. To run a linux container
-locally, you can use the following command (substitute `<git-tag-of-release` with the tag name):
+locally, you can use the following command (substitute `<git-tag-of-release>` with the tag name):
 
 **macOS/Linux**
 ```bash


RELEASE.md
```diff
@@ -34,8 +34,8 @@ During the staging step, we must create a source distribution & stage any binary
 1. Create a release in `grails-core`:
    * Click "Draft a new release" here: https://github.com/apache/grails-core/releases
    * On the draft new release screen, we execute the following steps:
-     * The tag will have the prefix 'v', so for our release it would be 'v7.0.0-M4'
-     * The "Target" will be the branch we build out of (this case it's the default, 7.0.x)
+     * The tag will have the prefix 'v', so for our release it would be 'v8.0.0-M1'
```
jdaugherty: Why even bother changing this?  The point of this is to just give an example release tag, not keep
it in sync with our version

RELEASE.md
```diff
@@ -132,13 +132,13 @@ To ensure checksums match the server & signatures match, run the script `etc/bin
...
-    ./etc/bin/verify-jar-artifacts.sh v7.0.0-M4 <grailsdownloadlocation>
+    ./etc/bin/verify-jar-artifacts.sh v8.0.0-M1 <grailsdownloadlocation>
 ```
 
 ### Manual Verification: Reproducible Jar Files
-After all jar files are verified to be signed by a valid Grails key, we need to build a local copy to ensure the file was built with the right code base. The `very-reproducible.sh` script handles this check, but if the bootstrap needs to be manually bootstrapped, perform the following step: 
+After all jar files are verified to be signed by a valid Grails key, we need to build a local copy to ensure the file was built with the right code base. The `verify-reproducible.sh` script handles this check, but if the bootstrap needs to be manually bootstrapped, perform the following step from inside the downloaded source distribution:
 
-    gradle -p gradle-bootstrap
+    cd gradle-bootstrap && gradle
```
jdaugherty: Why are you changing this?

jdaugherty (commented):
I don't think we should be updating release.md every major version.  the document is meant to be examples - its not supposed to track our versions.

RELEASE.md
jamesfredley: Agreed, reverted. Restored the original `v7.0.0-M4` / `7.0.x` example.

RELEASE.md
jamesfredley: Reverted. `gradle -p gradle-bootstrap` and `cd gradle-bootstrap && gradle` are functionally identical (`-p` is `--project-dir`), so no need to change the doc - kept only the `very-` -> `verify-` filename typo fix.

jamesfredley:
Agreed - reverted the version bumps. RELEASE.md should not track our current version; the existing `v7.0.0-M4` / `7.0.x` examples are fine as illustrative tags. PR now contains only three pure typo fixes: `very-reproducible.sh` -> `verify-reproducible.sh`, `with be` -> `will be`, and closing the unbalanced backtick on `<git-tag-of-release>`.

PR #15621, opened 2026-05-02
feat: add verify-branch.sh for pre-tag release readiness checks

jamesfredley:
## Summary

Add `etc/bin/verify-branch.sh`, a local-only pre-tag release readiness checker that runs every gate `release.yml`'s `publish` job runs (and a couple more) against the working tree, without requiring a tag, GitHub release, or staged artifacts. Closes the documented gap that the existing `verify.sh` only works *after* staging.

## Why

`verify.sh` answers the question *"Were the artifacts that release.yml staged equal to the artifacts I would build from this tag?"*. It needs three things that only exist after the tag is published: a GitHub Release, jars at repository.apache.org, and source/wrapper/cli zips at dist.apache.org. Until the tag is published, none of those exist.

That means today the only way to discover that a branch is not actually ready for release is to tag it, watch the publish job fail, run `Release - Abort Release`, fix, and re-tag. We need a local-only "is this branch ready to be tagged?" check.

## What it checks

| Step | Check | Time | Equivalent to |
|---|---|---|---|
| 1 | `dependencies.gradle` does not contain `-SNAPSHOT` | < 1 s | RELEASE.md Prerequisites |
| 2 | `KEYS` content matches canonical copy at `dist.apache.org/repos/dist/release/grails/KEYS` | ~ 1 s | content-aware version of `verify-keys.sh` |
| 3 | Apache RAT license audit (`./gradlew rat`) | ~ 5 min | `release.yml` publish (and CI) |
| 4 | Code style (`./gradlew codeStyle` - Checkstyle + CodeNarc) | ~ 3-5 min | `codestyle.yml` |
| 5 | grails-core assemble (`./gradlew assemble -PgithubBranch=...`) | ~ 10 min | `release.yml` publish |
| 6 | grails-forge assemble | ~ 5 min | `release.yml` publish |
| 7 | grails-doc build | ~ 3 min | `release.yml` publish |

Optional flags add slower checks:

- `--include-reproducibility` runs `etc/bin/test-reproducible-builds.sh` (30-60+ min)
- `--include-tests` runs `./gradlew test`
- `--include-all` adds both
- `--skip-build` cuts to steps 1-4 only (fast triage, ~5-10 min)

Defaults to steps 1-7 (typically 15-30 min).

## Other behaviors worth knowing

- **`SOURCE_DATE_EPOCH`** is set from `git log -1 --pretty=%ct` so any reproducibility-sensitive task that runs here uses the same epoch `release.yml`'s publish job will use.
- **Unclean-tree warning**: RAT scans every file regardless of git tracking. If the working tree has untracked or modified files (AI tool configs, IDE state, temporary work directories), they would falsely fail RAT even though the tracked branch is clean. The script warns explicitly and points you at `git worktree add` if you want a one-command path to a clean check.
- **CRLF resilience**: on a Windows working tree checked out under `core.autocrlf=true` before `.gitattributes` (apache/grails-core#15620) was in place, the on-disk `gradlew` and sibling scripts have CRLF and are unusable on Linux. The script transparently sed-strips CR-only copies under `build/branch-verify-tmp` and uses those, so it remains runnable from a Windows checkout. On Linux/macOS the sed step is a no-op rewrite.
- **Container parity**: runs unchanged inside the existing verification container documented in `RELEASE.md` Appendix - just `cd /home/groovy/project && etc/bin/verify-branch.sh`.

## RELEASE.md update

Add a "Pre-tag Branch Verification (Optional but Recommended)" subsection under Prerequisites that walks through `verify-branch.sh` invocations on host and inside the container, with a clear pointer that after staging completes you still run `verify.sh <tag> .`.

## Local validation

Tested on a Windows machine with `core.autocrlf=true` (so the worst-case path) inside the container documented in the existing Dockerfile:

- `etc/bin/verify-branch.sh --help` -> usage text
- `etc/bin/verify-branch.sh --skip-build --branch 8.0.x` -> SNAPSHOT and KEYS gates pass, RAT runs and correctly fails on local untracked debris (AI tool configs and a `cyclonedx-plugin-temp/` work directory). On a clean checkout RAT will pass cleanly. The script's unclean-tree warning fires before RAT and tells the user exactly why.

## Related

- Companion to apache/grails-core#15619 (`docs: update RELEASE.md examples for 8.0.0-M1`)
- Builds on apache/grails-core#15620 (`fix: enforce LF line endings for shell scripts, gradlew, KEYS`); this PR works correctly even before #15620 is merged thanks to the on-the-fly CR-strip resolver.

## Out of scope

- The verification container does not currently include `git`. The unclean-tree pre-flight is therefore a no-op when the script runs inside the container today (it explicitly says so). A follow-up adding `git` to `etc/bin/Dockerfile` would let the cleanliness warning fire inside the container too. Not bundled here to keep this PR focused.

### DIFF
--- a/etc/bin/verify-branch.sh
+++ b/etc/bin/verify-branch.sh
@@ -1,0 +1,349 @@
+#!/usr/bin/env bash
+#
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+#
+# verify-branch.sh - pre-tag release readiness check
+#
+# Runs all release-readiness checks that can be done LOCALLY against the
+# current working tree, with no tag, no GitHub release, and no staged
+# artifacts. This is the pre-tag counterpart to verify.sh.
+#
+# Use this BEFORE publishing the release tag (e.g. v8.0.0-M1) to gain
+# confidence that the publish job in .github/workflows/release.yml will
+# succeed. After staging completes, use the existing verify.sh to verify
+# the staged artifacts themselves.
+#
+# Recommended usage on the 8.0.x branch:
+#
+#   etc/bin/verify-branch.sh
+#
+# Or inside the verification container (recommended for parity with CI):
+#
+#   docker build -t grails:testing -f etc/bin/Dockerfile .
+#   docker run -it --rm -v "$(pwd):/home/groovy/project" grails:testing bash
+#   cd /home/groovy/project
+#   etc/bin/verify-branch.sh
+
+set -euo pipefail
+
+usage() {
+  cat <<'EOF'
+Usage: verify-branch.sh [OPTIONS]
+
+Runs the release-readiness checks that can be performed locally against the
+current working tree, without requiring a tag, GitHub release, or any staged
+artifacts. Mirrors the locally-runnable surface of release.yml's publish job.
+
+Default checks (typically 15-30 minutes):
+  1. dependencies.gradle does not contain SNAPSHOT versions
+  2. KEYS file matches the canonical copy at dist.apache.org
+  3. Apache RAT license audit (./gradlew rat)
+  4. Code style (./gradlew codeStyle)
+  5. grails-core assemble (./gradlew assemble)
+  6. grails-forge assemble (cd grails-forge && ./gradlew assemble)
+  7. grails-doc build (./gradlew grails-doc:build)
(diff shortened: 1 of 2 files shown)


etc/bin/verify-branch.sh
```diff
@@ -0,0 +1,349 @@
...
+
+usage() {
+  cat <<'EOF'
+Usage: verify-branch.sh [OPTIONS]
+
+Runs the release-readiness checks that can be performed locally against the
+current working tree, without requiring a tag, GitHub release, or any staged
+artifacts. Mirrors the locally-runnable surface of release.yml's publish job.
+
+Default checks (typically 15-30 minutes):
```
jdaugherty, line 52: We should add the valid dependency versions task to this and release.yml since that makes sure our versions resolve to the bom versions

jdaugherty (commented):
I am ok with this script addition, but by adding this are you suggesting we run this before doing a release? The point of the workflows is so we can run from a common env but this script would be by definition run locally. Why not just stage the release?  We basically have 2 code paths to maintain now and thats why we have CI

etc/bin/verify-branch.sh
jamesfredley: image

mostly to pre smoke test it before waiting for this to complete.  And this is possible now, this just tries to wrap it into a single script.

jdaugherty: I think you responded to the parent comment instead of the suggestion to add the validation dependency task. 

As for smoke testing, isn't this only an issue once we create major branches?  I'm trying to point out that the benefit to having another script isn't greater than the cost to maintain 2 separate pieces of logic.  Plus, you ultimately still can have the workflow fail b/c of environmental configuration.

PR #15624, opened 2026-05-02
fix: make SBOM and groovydoc output reproducible across CI and local builds

jamesfredley:
## Summary

The container-based release-vote verification described in `RELEASE.md` fails on **v8.0.0-M1** because nine jars are not byte-identical between the CI-published copies and a fresh local rebuild from the source distribution. The SbomPlugin's existing deterministic `serialNumber` logic and the project's existing reproducibility infrastructure are correct, but three independent root causes upstream of them break reproducibility for these specific files.

This PR addresses all three.

### Affected jars (verified via `verify.sh v8.0.0-M1`)

| Jar | What differs | Fix in this PR |
|---|---|---|
| `grails-async`, `grails-async-core`, `grails-async-gpars`, `grails-async-rxjava{,2,3}`, `grails-bootstrap` | `META-INF/sbom.json` | Fix #1 |
| `grails-cache` | `META-INF/sbom.json` and `META-INF/sbom/application.cdx.json` | Fix #1 + Fix #2 |
| `grails-bootstrap-javadoc` | 4 HTML files for `FieldDefinition`/`PropertyDefinition` | Fix #3 |

### 1. `SbomPlugin` direct task: strip the auto-injected `build-system` externalReference

`cyclonedx-gradle-plugin` v3.0.0 auto-injects a `build-system` externalReference pointing at `https://github.com/apache/grails-core/actions/runs/<run_id>` when `GITHUB_ACTIONS` env vars are present. This URL is unknowable from source, so its presence breaks the deterministic `serialNumber` hash that `SbomPlugin` already computes from the BOM content. Stripping this reference before the hash recompute makes `META-INF/sbom.json` byte-identical between CI and local rebuilds.

### 2. `SbomPlugin` aggregate task: apply the same reproducibility transforms

Spring Boot 4 (`CycloneDxPluginAction`) wires the cyclonedx-gradle-plugin aggregate task (`CyclonedxAggregateTask`, default name `cyclonedxBom`) into Grails-plugin jars, packaging its output at `META-INF/sbom/application.cdx.json`. `SbomPlugin` previously only configured the direct task (`CyclonedxDirectTask`), so the aggregate SBOM had:
- a random `serialNumber` UUID
- an `Instant.now()` timestamp that ignored `SOURCE_DATE_EPOCH`
- CI auto-detected externalReferences (`build-system` and `vcs`)

The rewrite logic is extracted into a shared `rewriteSbomFile` helper applied to both task types via `configureSbomTask` and a new `configureAggregateSbomReproducibility` method. The aggregate strip set covers both `build-system` AND `vcs` because the aggregate task has no explicit `externalReferences` configuration to fall back on (unlike the direct task which sets four explicit refs).

### 3. `FieldDefinition` / `PropertyDefinition`: qualify `Builder` return types for groovydoc

Both classes declare a static inner `Builder` class with the same simple name, and both extend `AbstractMemberDefinition`. The `static Builder builder()` factory method previously used the unqualified return type `Builder`. When **groovydoc** walks the class graph, it resolves that simple name based on file system iteration order. Different filesystems produce different (and incorrect) HTML:
- One ordering: every `Builder` reference renders as `PropertyDefinition.Builder` (so `FieldDefinition.html` is wrong)
- Other ordering: every `Builder` reference renders as `FieldDefinition.Builder` (so `PropertyDefinition.html` is wrong)

This is **not a Gradle 9 issue** (Gradle 9's reproducibility fixes target `AbstractArchiveTask`, not the groovydoc tool's name resolution), and `noTimestamp = true` is already configured in `GroovydocEnhancerPlugin` so the difference is content, not metadata. Qualifying the return types as `static FieldDefinition.Builder builder()` and `static PropertyDefinition.Builder builder()` forces groovydoc to render the correct (and consistent) types regardless of iteration order.

**Compiled bytecode is unchanged** because the nested `Builder` classes already resolved correctly via lexical scope at the `groovyc` stage.

## Verification

Built `grails-async-core`, `grails-bootstrap`, and `grails-cache` twice with the same `SOURCE_DATE_EPOCH` and confirmed:

```
✅ grails-async-core: byte-identical between runs
✅ grails-bootstrap : byte-identical between runs
✅ grails-cache     : byte-identical between runs (sbom.json AND application.cdx.json)
✅ no build-system externalReference in any sbom.json
✅ no build-system or vcs externalReference in application.cdx.json
✅ deterministic urn:uuid serialNumber format
```

Existing `FieldDefinitionSpec` and `PropertyDefinitionSpec` test suites continue to pass because the bytecode-level `builder()` return type is unchanged.

End-to-end CI-vs-local validation will happen on this PR's CI build: once merged and a release is cut, `verify.sh v<next_release>` should pass on the previously-failing 9 jars.

## Test plan

- [x] `./gradlew :grails-bootstrap:test --tests "grails.codegen.model.FieldDefinitionSpec" --tests "grails.codegen.model.PropertyDefinitionSpec"` passes
- [x] `./gradlew :grails-bootstrap:codeStyle` passes
- [x] `./gradlew :build-logic:compileGroovy` (in `build-logic/`) passes
- [x] `:grails-async-core:cyclonedxDirectBom`, `:grails-bootstrap:cyclonedxDirectBom`, `:grails-cache:cyclonedxDirectBom` produce byte-identical output across two consecutive runs with the same `SOURCE_DATE_EPOCH`
- [x] `:grails-cache:cyclonedxBom` (aggregate) produces byte-identical output across two runs

## Notes for reviewers

- The aggregate task strip set (`['build-system', 'vcs']`) is intentionally broader than the direct task's (`['build-system']`). The direct task explicitly sets a known-good `vcs` reference at `SbomPlugin.groovy:191-193`, so we only need to strip the auto-injected `build-system`. The aggregate task has no such explicit config, so any externalReferences it has come from CI/`.git` auto-detection that won't reproduce locally.

- Build-logic does not currently have a `codeStyle` task or unit-test infrastructure for plugins; the verification was done via behavioral integration build. If you'd like to add `SbomPluginSpec` coverage in this PR, happy to extend - just say the word.

- For the groovydoc fix, an alternative approach would be renaming one of the inner `Builder` classes (`FieldDefinitionBuilder` / `PropertyDefinitionBuilder`) to remove the simple-name collision entirely. That's an API-breaking change so I went with the non-breaking qualification approach. If you'd prefer the rename, happy to switch.

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
@@ -25,8 +25,9 @@ import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
-import org.cyclonedx.gradle.CyclonedxPlugin
+import org.cyclonedx.gradle.CyclonedxAggregateTask
 import org.cyclonedx.gradle.CyclonedxDirectTask
+import org.cyclonedx.gradle.CyclonedxPlugin
 import org.cyclonedx.model.Component
 import org.cyclonedx.model.ExternalReference
 import org.cyclonedx.model.License
@@ -147,6 +148,7 @@ class SbomPlugin implements Plugin<Project> {
         )
 
         configureSbomTask(project, sbomOutputLocation)
+        configureAggregateSbomReproducibility(project)
         configureNormalization(project)
         ensureLicensesValidated(project)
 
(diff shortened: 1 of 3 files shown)


jamesfredley:
### Clarification: relationship to PR #15614

To preempt a likely review question: **PR #15614's `projectPath`-into-UUID-seed fix was already merged and active when v8.0.0-M1 was tagged**, yet the 9 jars still failed reproducibility. Verified by checking the `SbomPlugin.groovy` content at the `v8.0.0-M1` tag (commit `6b7c015`):

```groovy
def uuidSeed = "${projectPath}\n${withoutSerial}"
def uuid = UUID.nameUUIDFromBytes(uuidSeed.getBytes(StandardCharsets.UTF_8))
```

**The two fixes address different problems:**

| | What it fixes |
|---|---|
| #15614 | `serialNumber` *collisions* between distinct modules whose post-processed BOM content happens to be identical |
| This PR | The *same module* having different BOM content between CI and local rebuilds, because `cyclonedx-gradle-plugin` v3.0.0 auto-injects an `externalReferences[type=build-system]` entry only when CI env vars are set |

Concretely, for v8.0.0-M1's `grails-async-8.0.0-M1.jar`:

- CI build: BOM had 5 externalRefs including `build-system: https://github.com/apache/grails-core/actions/runs/25254902009` → UUID `c4fdb31...`
- Local rebuild: BOM had 4 externalRefs (no `build-system`) → UUID `2ce9328...`

`projectPath` was the same in both. The deterministic hash worked correctly. The inputs were genuinely different, so the outputs were correctly different. Stripping `build-system` from the BOM *before* the hash recompute (this PR) is what makes the inputs match.

I've kept #15614's `projectPath`-mixing intact in the new shared `rewriteSbomFile` helper so both improvements are active.

jamesfredley:
Closing - wrong base branch. The `grails8-groovy6-canary` branch has unrelated pre-existing CI failures (Groovy 6 SNAPSHOT migration issues such as `@Delegate` not generating abstract method implementations in `TemplateRenderer`) that obscure validation of this fix. Reopening against `8.0.x` where the fix is more directly useful and where CI is stable. Same commit will appear in the new PR.

PR #15625, opened 2026-05-02
fix(sbom): make per-module SBOM and groovydoc output byte-reproducible

jamesfredley:
## Summary

Container-based release-vote verification (per `RELEASE.md`) on **v8.0.0-M1** found nine jars not byte-identical between CI-published copies and a fresh local rebuild:

- `grails-async`, `grails-async-core`, `grails-async-gpars`, `grails-async-rxjava{,2,3}`, `grails-bootstrap`, `grails-cache` - differed in their SBOM file(s)
- `grails-bootstrap-javadoc` - differed in 4 HTML files for `FieldDefinition` / `PropertyDefinition`

This PR fixes both root causes at source.

### 1. `SbomPlugin`: register `CyclonedxDirectTask` ourselves and set `includeBuildSystem = false`

Two related observations:

- **`cyclonedx-gradle-plugin`'s `apply()` always registers BOTH tasks.** Looking at `CyclonedxPlugin.apply()` ([source](https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/cyclonedx-gradle-plugin-3.0.0/src/main/java/org/cyclonedx/gradle/CyclonedxPlugin.java)), it unconditionally registers the per-module `cyclonedxDirectBom` and the aggregate `cyclonedxBom` (a `CyclonedxAggregateTask`). When Spring Boot 4 sees the plugin, its `CycloneDxPluginAction` then wires that aggregate's output into the jar at `META-INF/sbom/application.cdx.json` - misleading for library jars and an extra non-reproducible artifact to chase. Grails ships only per-module SBOMs at `META-INF/sbom.json`, so we register `CyclonedxDirectTask` directly and never apply `CyclonedxPlugin`. The aggregate task is never created in the first place; Spring Boot's reaction never fires either since it triggers on plugin-application of `CyclonedxPlugin`. There is nothing to disable, exclude, or filter.

- **`cyclonedx-gradle-plugin` 3.0.0 added a CI-only `build-system` externalReference** populated from GitHub Actions env vars (the run URL), which is unknowable from a local rebuild. The plugin exposes a first-class `includeBuildSystem` property (default `true`); setting it to `false` suppresses the reference at the plugin level. No JSON post-processing for that field, no externalReference filter set.

PR #15614's `projectPath`-into-UUID-seed fix is preserved unchanged.

### 2. `FieldDefinition` / `PropertyDefinition`: qualify the inner `Builder` return types

Both classes declare a static inner `Builder` of the same simple name. Groovydoc's `SimpleGroovyClassDoc` resolves unqualified `Builder` references via a shared short-name cache populated in filesystem iteration order, so the rendered HTML for `FieldDefinition.Builder` and `PropertyDefinition.Builder` differs between filesystems. Qualifying the return types as `static FieldDefinition.Builder builder()` / `static PropertyDefinition.Builder builder()` forces groovydoc to render the correct types reproducibly. Bytecode is unchanged - `groovyc` resolves nested `Builder` via lexical scope at compile time, which is why this was a docs-only bug.

This is the same root cause filed upstream as **[GROOVY-11954](https://issues.apache.org/jira/browse/GROOVY-11954)** and fixed in Groovy 4.0.32, 5.0.6, and 6.0.0-alpha-1 ([apache/groovy#2484](https://github.com/apache/groovy/pull/2484)). Grails pins `groovy.version=4.0.31` and 4.0.32 has not yet been published to Maven Central, so the qualification stays in place for v8.0.0-M1. Each `builder()` carries a `TODO(GROOVY-11954)` note so a future Groovy bump can drop the qualification trivially.

## Diff size

**`+58 / -6` across 3 files.** Everything else from earlier iterations of this PR (the aggregate-task disable, the `META-INF/sbom/**` jar exclude, the `META-INF/sbom/**` runtime-classpath ignore, the `rewriteSbomFile` extraction, the `externalRefTypesToStrip` parameter, the build-system stripping inside the JSON post-processor) became unnecessary once the cyclonedx plugin was no longer being applied and `includeBuildSystem` was used at the plugin level.

## Verification

```
./gradlew :grails-bootstrap:tasks --all | grep -i cyclonedx
# only cyclonedxDirectBom; no cyclonedxBom

./gradlew :grails-bootstrap:jar --rerun-tasks      # build 1
./gradlew :grails-bootstrap:jar --rerun-tasks      # build 2
# both jars: SHA-256 4FE425486AF7E77CC16117D899BFB7042B8292685237FEAAF17259E68FF87620

# SBOM in jar:
#   - present at META-INF/sbom.json
#   - NO META-INF/sbom/application.cdx.json
#   - NO build-system externalReference
#   - deterministic urn:uuid serialNumber
#   - dependencies[*].dependsOn sorted
```

## Test plan

- [x] `./gradlew :build-logic:compileGroovy --rerun-tasks` passes
- [x] `./gradlew :grails-bootstrap:jar --rerun-tasks` passes; jar is byte-reproducible across two runs
- [x] `cyclonedxBom` aggregate task is **not registered** on grails-plugin modules (confirmed via `:grails-bootstrap:tasks --all`)
- [x] Produced jars contain only `META-INF/sbom.json`, no `META-INF/sbom/**`
- [x] No `build-system` externalReference in any sbom.json
- [x] `:grails-bootstrap:test --tests "grails.codegen.model.FieldDefinitionSpec" --tests "grails.codegen.model.PropertyDefinitionSpec"` passes

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
@@ -25,7 +25,6 @@ import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
-import org.cyclonedx.gradle.CyclonedxPlugin
 import org.cyclonedx.gradle.CyclonedxDirectTask
 import org.cyclonedx.model.Component
 import org.cyclonedx.model.ExternalReference
@@ -37,6 +36,7 @@ import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.CopySpec
+import org.gradle.api.file.Directory
 import org.gradle.api.file.RegularFile
 import org.gradle.api.java.archives.Manifest
 import org.gradle.api.provider.Provider
@@ -126,7 +126,7 @@ class SbomPlugin implements Plugin<Project> {
 
     @Override
     void apply(Project project) {
-        project.pluginManager.apply(CyclonedxPlugin)
+        registerCyclonedxDirectBomTask(project)
 
         def sbomOutputLocation = project.layout.buildDirectory.file(
                 project.provider {
@@ -142,6 +142,25 @@ class SbomPlugin implements Plugin<Project> {
 
         // sboms are only published to Grails jar files at this time
         publishSbomForJarProjects(project, sbomOutputLocation)
+    }
+
+    /**
+     * Register only the per-module {@link CyclonedxDirectTask}. We deliberately do not apply
+     * {@code CyclonedxPlugin} because doing so unconditionally also registers the
+     * {@code cyclonedxBom} aggregate task ({@code CyclonedxAggregateTask}), and triggers
+     * Spring Boot 4's {@code CycloneDxPluginAction} to wire that aggregate's output into the
+     * jar at {@code META-INF/sbom/application.cdx.json}. Grails ships per-module SBOMs only
+     * (at {@code META-INF/sbom.json}, see {@link #publishSbomForJarProjects}); the aggregate
+     * SBOM is misleading for library jars and would also need its own reproducibility
+     * post-processing. Registering the direct task ourselves means the aggregate task is
+     * never created in the first place, so there is nothing to disable, exclude, or filter.
+     */
+    private static void registerCyclonedxDirectBomTask(Project project) {
+        project.tasks.register('cyclonedxDirectBom', CyclonedxDirectTask) { CyclonedxDirectTask task ->
+            Provider<Directory> reportDir = project.layout.buildDirectory.dir('reports/cyclonedx-direct')
+            task.xmlOutput.convention(reportDir.map { it.file('bom.xml') })
+            task.jsonOutput.convention(reportDir.map { it.file('bom.json') })
+        }
     }
 
     private static void configureSbomTask(Project project, Provider<RegularFile> sbomOutputLocation) {
@@ -202,6 +221,13 @@ class SbomPlugin implements Plugin<Project> {
 
                 // turn off license text since it's base64 encoded & will inflate the jar sizes
                 includeLicenseText.set(false)
+
+                // Don't auto-inject the build-system externalReference. cyclonedx-gradle-plugin 3.0.0
+                // populates this from CI env vars (e.g. GitHub Actions run URL), which is unknowable
+                // from a local rebuild and would break byte-identical reproducibility between CI and
+                // local builds. Per ASF policy we ship reproducible artifacts and intentionally leave
+                // build info out of the sbom; this is the supported plugin-level off switch for it.
+                includeBuildSystem.set(false)
 
                 // disable xml output
                 xmlOutput.unsetConvention()
--- a/grails-bootstrap/src/main/groovy/grails/codegen/model/FieldDefinition.groovy
+++ b/grails-bootstrap/src/main/groovy/grails/codegen/model/FieldDefinition.groovy
@@ -150,7 +150,20 @@ class FieldDefinition extends AbstractMemberDefinition {
         }
     }
 
-    static Builder builder() {
-        new Builder()
+    /**
+     * Factory for the {@link FieldDefinition.Builder}. The return type is intentionally fully qualified:
+     * {@code PropertyDefinition} declares an inner Builder of the same simple name, and groovydoc resolves
+     * unqualified {@code Builder} references inconsistently between builds (depends on filesystem iteration
+     * order). Keep this qualified to ensure groovydoc renders {@code FieldDefinition.Builder} reproducibly.
+     *
+     * TODO(GROOVY-11954): Revisit and simplify the return type to {@code Builder} once Grails depends on
+     * Groovy 4.0.32+ (also fixed in 5.0.6 and 6.0.0-alpha-1). The upstream fix corrects the shared
+     * short-name resolution cache in {@code SimpleGroovyClassDoc} that causes this non-reproducibility,
+     * so the workaround here will no longer be needed. As of writing, Grails pins {@code groovy.version=4.0.31}
+     * and 4.0.32 has not yet been released to Maven Central. See:
+     * https://issues.apache.org/jira/browse/GROOVY-11954 and https://github.com/apache/groovy/pull/2484.
+     */
+    static FieldDefinition.Builder builder() {
+        new FieldDefinition.Builder()
     }
 }
--- a/grails-bootstrap/src/main/groovy/grails/codegen/model/PropertyDefinition.groovy
+++ b/grails-bootstrap/src/main/groovy/grails/codegen/model/PropertyDefinition.groovy
@@ -113,7 +113,20 @@ class PropertyDefinition extends AbstractMemberDefinition {
         }
     }
 
-    static Builder builder() {
-        new Builder()
+    /**
+     * Factory for the {@link PropertyDefinition.Builder}. The return type is intentionally fully qualified:
+     * {@code FieldDefinition} declares an inner Builder of the same simple name, and groovydoc resolves
+     * unqualified {@code Builder} references inconsistently between builds (depends on filesystem iteration
+     * order). Keep this qualified to ensure groovydoc renders {@code PropertyDefinition.Builder} reproducibly.
+     *
+     * TODO(GROOVY-11954): Revisit and simplify the return type to {@code Builder} once Grails depends on
+     * Groovy 4.0.32+ (also fixed in 5.0.6 and 6.0.0-alpha-1). The upstream fix corrects the shared
+     * short-name resolution cache in {@code SimpleGroovyClassDoc} that causes this non-reproducibility,
+     * so the workaround here will no longer be needed. As of writing, Grails pins {@code groovy.version=4.0.31}
+     * and 4.0.32 has not yet been released to Maven Central. See:
+     * https://issues.apache.org/jira/browse/GROOVY-11954 and https://github.com/apache/groovy/pull/2484.
+     */
+    static PropertyDefinition.Builder builder() {
+        new PropertyDefinition.Builder()
     }
 }


jamesfredley:
### Self-review pass: applied targeted feedback

I ran a multi-agent self-review against codebase patterns, prior reviewer history (PR #15614 / jdaugherty feedback), upstream `cyclonedx-gradle-plugin` v3.0.0 source, and architectural correctness. Force-pushed three small refinements:

1. **Defensive file-existence guard** in `rewriteSbomFile`. The aggregate task's `task.jsonOutput.isPresent()` check confirms the property is set, but doesn't guarantee the file exists on disk. Added `if (!f.exists() || f.length() == 0) { logger.warn(...); return }` to fail soft if the upstream task produced no output.
2. **Method-level Javadoc on `configureAggregateSbomReproducibility`** explaining when this hook fires (Spring Boot 4's `CycloneDxPluginAction` registers `cyclonedxBom` and packages output at `META-INF/sbom/application.cdx.json`) and why the strip set is broader than the direct task's (no explicit externalReferences fallback).
3. **Method-level Javadoc on `FieldDefinition.builder()` and `PropertyDefinition.builder()`** explaining that the qualified `FieldDefinition.Builder` / `PropertyDefinition.Builder` return type is intentional for groovydoc reproducibility. Without this, a future "simplify return type" refactor would silently re-introduce the bug.

### Empirical groovydoc verification

I noticed I'd never actually generated groovydoc to confirm the qualified-Builder fix produces correct output. Did so now (build twice with `SOURCE_DATE_EPOCH` set):

```
✅ FieldDefinition.html builder() correctly links to .../FieldDefinition.Builder.html
✅ PropertyDefinition.html builder() correctly links to .../PropertyDefinition.Builder.html
✅ Both HTMLs byte-identical across two consecutive runs
```

Compared to v8.0.0-M1's published javadoc (where every Builder reference linked to PropertyDefinition.Builder regardless of which class was being documented), this is the correct, deterministic output.

### Review findings I considered but rejected

- **Reorganize all imports in SbomPlugin.groovy**: scope creep. My only addition (`CyclonedxAggregateTask`) was alphabetically correct in its position relative to existing `org.cyclonedx.*` imports. The pre-existing `java.*` ordering quirk (with `AtomicBoolean` separated from other `java.*` imports) is the project's existing convention.
- **Revert qualified return types in FieldDefinition/PropertyDefinition**: that was the bug. Empirically verified via the groovydoc rebuild above.
- **Add SbomPluginSpec unit tests**: no existing test infrastructure for build-logic plugins. Behavioral verification via the integration build (run twice, byte-identical) is the established pattern for this code per CONTRIBUTING.md.

### Pre-existing tech debt noted (NOT addressed in this PR)

- Both `configureSbomTask` (existing) and `configureAggregateSbomReproducibility` (new) capture `project` in `Provider` closures, which would break Gradle's configuration cache if it were enabled. The new method matches the existing pattern. Should be a follow-up issue covering both call sites - happy to file one if desired.

matrei:
Related: https://issues.apache.org/jira/browse/GROOVY-11954

grails-bootstrap/src/main/groovy/grails/codegen/model/FieldDefinition.groovy
```diff
@@ -150,7 +150,13 @@ class FieldDefinition extends AbstractMemberDefinition {
         }
     }
 
-    static Builder builder() {
-        new Builder()
+    /**
+     * Factory for the {@link FieldDefinition.Builder}. The return type is intentionally fully qualified:
+     * {@code PropertyDefinition} declares an inner Builder of the same simple name, and groovydoc resolves
+     * unqualified {@code Builder} references inconsistently between builds (depends on filesystem iteration
```
jdaugherty, line 156: This sounds like a groovy bug.  shouldn't we open a ticket on this?

jdaugherty (commented):
why are we wiring the aggregate even?  We intentionally left build info out of our sbom because this process wasn't ready.

jamesfredley:
### Pivot: drop the aggregate SBOM rather than make it reproducible

Pushed [`95b7d83`](https://github.com/apache/grails-core/pull/15625/commits/95b7d83968) which changes the approach to Fix #2 after looking at this from a different angle. The PR title and description have been updated to reflect the new direction.

**Original Fix #2**: port the direct-task reproducibility transforms (timestamp pinning, `build-system` externalReference strip, deterministic `serialNumber`) onto the aggregate task so `META-INF/sbom/application.cdx.json` becomes byte-identical between CI and local.

**New Fix #2**: drop the aggregate task (`cyclonedxBom`) entirely from Grails-plugin jars. Grails plugins are libraries, not applications, and the `application.cdx.json` filename Spring Boot 4 packages is misleading. The per-module `META-INF/sbom.json` already provides the same data with a more accurate name and a much simpler reproducibility path - and aligns with how comparable frameworks (Spring Boot itself, Quarkus, Micronaut) and other Apache projects ship SBOMs (per-module / per-component, not aggregated framework-wide).

**What @matrei's GROOVY-11954 link surfaced**: that ticket is the upstream fix for the same shared-short-name-cache root cause that Fix #3 works around (Date/Builder/etc. resolving inconsistently in groovydoc based on filesystem iteration order). It's resolved in `4.0.32`, `5.0.6`, and `6.0.0-alpha-1` - none yet published to Maven Central. Rather than rush a Groovy bump, both `FieldDefinition.builder()` and `PropertyDefinition.builder()` now carry a `TODO(GROOVY-11954)` Javadoc note, so once Grails bumps to `4.0.32+` the workaround can be removed in a one-line cleanup with no behavior change (bytecode is identical either way).

**What stayed the same**: Fix #1 (direct-task `build-system` strip) and Fix #3 (qualified `Builder` return type) are unchanged from the previous push - they remain necessary for v8.0.0-M1 reproducibility today.

**What changed in the diff**:

- `configureAggregateSbomReproducibility(...)` removed, replaced by `disableAggregateSbomGeneration(...)` which sets `cyclonedxBom.enabled = false`.
- `rewriteSbomFile(...)` simplified: dropped the `Set<String> externalRefTypesToStrip` parameter (no second caller now) and inlined the `build-system` strip set.
- Defense in depth: `jar.exclude('META-INF/sbom/**')` on every Grails-plugin jar plus a matching `META-INF/sbom/**` entry in `configureNormalization`'s `runtimeClasspath` ignore list, so a future Spring Boot 4 update cannot silently re-introduce an aggregate SBOM from a different source.

**Verification** (build twice with the same `SOURCE_DATE_EPOCH`, hash and inspect):

```
✅ :grails-cache:jar              sha256 98fa318a... matches between runs
✅ :grails-cache META-INF/sbom.json   sha256 6ac48de0... matches between runs
✅ :grails-cache:cyclonedxBom        SKIPPED (disabled)
✅ :grails-cache:cyclonedxDirectBom  produces the per-module SBOM
✅ Produced jars contain only META-INF/sbom.json - no entries under META-INF/sbom/
✅ FieldDefinitionSpec, PropertyDefinitionSpec, :grails-bootstrap:codeStyle all pass
```

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
```diff
@@ -383,6 +436,13 @@ class SbomPlugin implements Plugin<Project> {
                                         'sbom.json'
                                     }
                                 }
+                                // Defense-in-depth against Spring Boot 4's CycloneDxPluginAction
+                                // (or any future Gradle plugin) re-introducing an aggregate SBOM at
+                                // META-INF/sbom/application.cdx.json. Grails-plugin jars are libraries
+                                // and ship only the per-module META-INF/sbom.json above; the aggregate
+                                // path under META-INF/sbom/ is intentionally excluded. See
+                                // disableAggregateSbomGeneration for the primary disable.
+                                jar.exclude('META-INF/sbom/**')
```
jdaugherty: This isn't the right solution, it masks the problem

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/SbomPlugin.groovy
```diff
@@ -218,81 +251,101 @@ class SbomPlugin implements Plugin<Project> {
...
         project.normalization { handler ->
             handler.runtimeClasspath {
+                // The per-module SBOM lives at META-INF/sbom.json; ignore it so that SBOM-only
+                // changes (e.g. dependency version bumps that flip a license id) do not invalidate
+                // the runtime classpath fingerprint of downstream projects.
                 it.ignore("META-INF/sbom.json")
+                // META-INF/sbom/** would only appear if Spring Boot 4 or a future plugin re-injected
+                // an aggregate SBOM despite disableAggregateSbomGeneration; keep this entry as a
+                // safety net so such a file would not silently start invalidating consumer caches.
+                it.ignore("META-INF/sbom/**")
```
jdaugherty: This isn't the right solution

jdaugherty (changes requested):
we shouldn't mask problems with global excludes - it makes it harder to discover these type of problems

I also think the method changes can be reverted for an easier diff

jamesfredley:
@jdaugherty - rewritten and force-pushed as a single squashed commit ([7f753f8](https://github.com/jamesfredley/grails-core/commit/7f753f8be59701bd4b4b51d6b300ef36958e22cb)) addressing all of your feedback. Diff is now **+58 / -6** across 3 files (was +152 / -66).

The earlier iterations were chasing the wrong problem. Once I went and read [`CyclonedxPlugin.apply()`](https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/cyclonedx-gradle-plugin-3.0.0/src/main/java/org/cyclonedx/gradle/CyclonedxPlugin.java) and [`BaseCyclonedxTask`](https://github.com/CycloneDX/cyclonedx-gradle-plugin/blob/cyclonedx-gradle-plugin-3.0.0/src/main/java/org/cyclonedx/gradle/BaseCyclonedxTask.java) directly, two things became obvious:

### 1. "why are we wiring the aggregate even?"

We weren't, deliberately - `CyclonedxPlugin.apply()` registers it for us, unconditionally. Its body is literally:

```java
public void apply(final Project project) {
    getProjectAndSubprojects(project).forEach(this::configureProject); // direct task per project
    ...
    registerCyclonedxAggregateBomTask(project, ...);                   // ALWAYS registers aggregate
    ...
}
```

So **the only way to not wire the aggregate is to not apply `CyclonedxPlugin`**. The fix is one line: replace `project.pluginManager.apply(CyclonedxPlugin)` with a direct `project.tasks.register('cyclonedxDirectBom', CyclonedxDirectTask) { ... }`. The aggregate task is then never registered, Spring Boot 4's `CycloneDxPluginAction` never fires (it triggers via `withType(CyclonedxPlugin)`), and there is nothing to disable, exclude, or filter.

That deletes `disableAggregateSbomGeneration`, the `jar.exclude('META-INF/sbom/**')`, and the `META-INF/sbom/**` runtime-classpath ignore in one stroke.

### 2. "we intentionally left build info out of our sbom because this process wasn't ready"

`BaseCyclonedxTask` exposes `includeBuildSystem` as a first-class `Property<Boolean>` with default `true`. The plugin's CI auto-injection of the `build-system` externalReference (the GitHub Actions run URL) is gated on it. **Setting `includeBuildSystem.set(false)` is the supported plugin-level off switch**, and it matches your stated intent for the SBOM. The JSON post-processor doesn't need to know about it at all, so the `Set<String> externalRefTypesToStrip` parameter and the externalReferences filter block are gone.

Net effect on `rewriteSbom`: it goes back to being the single inline closure it was before this PR, with no signature change and no extracted helper. That should resolve "the method changes can be reverted for an easier diff."

### 3. "This sounds like a groovy bug. shouldn't we open a ticket on this?"

Yes - it's filed as **[GROOVY-11954](https://issues.apache.org/jira/browse/GROOVY-11954)** and already fixed upstream in Groovy 4.0.32, 5.0.6, and 6.0.0-alpha-1 ([apache/groovy#2484](https://github.com/apache/groovy/pull/2484)). Grails currently pins `groovy.version=4.0.31` and 4.0.32 has not yet been published to Maven Central, so the qualification stays in place for v8.0.0-M1; each `builder()` factory has a `TODO(GROOVY-11954)` note so a future Groovy bump can drop it trivially. Bytecode is unchanged either way.

### Verification

- `:grails-bootstrap:tasks --all` shows `cyclonedxDirectBom` only - no `cyclonedxBom` aggregate
- Two consecutive `:grails-bootstrap:jar --rerun-tasks` runs produce byte-identical jars (same SHA-256: `4FE425486AF7E77CC16117D899BFB7042B8292685237FEAAF17259E68FF87620`)
- Produced jar contains only `META-INF/sbom.json`; no `META-INF/sbom/application.cdx.json`
- No `build-system` externalReference anywhere in `metadata.component.externalReferences`
- `serialNumber` is a deterministic `urn:uuid:...` (PR #15614's `projectPath`-into-UUID-seed mixing preserved unchanged)

PTAL.

jdaugherty (approved):

PR #15628, opened 2026-05-03
Set minimum Java Version to 21 for Grails 8.x

jdaugherty:
8.0.0-SNAPSHOT is currently showing Java 17 as an option, this removes it.


jamesfredley (approved):
We could add JDK_26 since it is supported, we did this with Grails 6 and 7, chased the latest until LTS was released.

PR #15631, opened 2026-05-05
Vulnerability scanning

jdaugherty:
Adds a vulnerability scanning workflow for grails-core


build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/VulnerabilityScanPlugin.groovy
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    void apply(Project project) {
+        project.pluginManager.apply('org.sonatype.gradle.plugins.scan')
+
+        OssIndexPluginExtension extension = project.extensions.getByType(OssIndexPluginExtension)
+        extension.username = System.getenv('SONATYPE_GUIDE_USERNAME') ?: ''
+        extension.password = System.getenv('SONATYPE_GUIDE_TOKEN') ?: ''
+        extension.excludeCoordinates = [
+            'org.hibernate:hibernate-core:5.6.11.Final',
```
jdaugherty, line 64: @jamesfredley @matrei We'll have to manage exceptions here.

jdaugherty:
Looks like the find main task is failing with these changes for some reason.  It's likely due to the plexus update.

jdaugherty:
Root cause: Adding the Sonatype scan plugin to build-logic/plugins/build.gradle changed the build-logic's content hash, which invalidated the Gradle build cache for all subproject tasks. When findMainClass runs fresh (not from       
  cache), MainClassFinder tries to read compiled Java 17 class files (major version 61) using groovyjarjarasm.asm.ClassReader — the ASM bundled with Groovy 3.x — which doesn't support class file version 61. This is a pre-existing bug  
  that was hidden by the build cache.                                                                                                                                                                                                      
                                                                                                                                                                                                                                           
  The fix is to switch MainClassFinder from groovyjarjarasm.asm (Groovy's bundled, potentially old ASM) to org.objectweb.asm (the explicit ASM library, version-managed by Spring Framework BOM at 9.7+, which supports Java 17+).

jdaugherty:
Converting this to draft, the dependency bumps have caused gradle issues

jdaugherty:
Alright, the problem is that the ASM classes are shaded inside scan-gradle-plugin-3.1.5.jar itself. They're not a separate transitive dependency. Gradle exclusions only work on transitive dependencies, not on classes bundled inside a fat/shaded jar. So we can't exclude them with a standard exclude rule.

The find main class is picking the shaded version, which is frustrating.

jdaugherty:
I've opened https://github.com/sonatype-nexus-community/scan-gradle-plugin/issues/206

jdaugherty:
https://github.com/sonatype-nexus-community/scan-gradle-plugin/pull/207 is the fix for this. We can't adopt due to the classpath conflict.  I'm hopeful Sonatype will accept the change.

jdaugherty:
Sonatype fixed the classpath issue so this is ready for review.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/VulnerabilityScanPlugin.groovy
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Convention plugin that configures Sonatype OSS Index / Guide vulnerability scanning.
+ *
+ * <p>Credentials are supplied via environment variables:
+ * <ul>
+ *   <li>{@code SONATYPE_GUIDE_USERNAME} - Guide account email / user token username</li>
+ *   <li>{@code SONATYPE_GUIDE_TOKEN} - Guide Personal Access Token (PAT)</li>
+ * </ul>
+ * The {@code ossIndexAudit} task is skipped silently when credentials are absent.</p>
```
jamesfredley: **#1 - Doc contradicts behavior.** This says the task is *"skipped silently when credentials are absent"*, but `doFirst` below (lines 63-67) actually **throws** `GradleException` when `SONATYPE_GUIDE_TOKEN` is missing. Since the workflow runs it with `continue-on-error: true`, throwing is probably the intent - so please update this sentence to say the task *fails* (rather than silently skips) when credentials are absent.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/VulnerabilityScanPlugin.groovy
```diff
@@ -0,0 +1,75 @@
...
+        extension.username = System.getenv('SONATYPE_GUIDE_USERNAME') ?: ''
+        extension.password = System.getenv('SONATYPE_GUIDE_TOKEN') ?: ''
+        extension.excludeCoordinates = [
+            'org.hibernate:hibernate-core:5.6.11.Final',
+            // CVE-2018-14335: Sonatype flags this against all H2 versions; no upstream fix exists.
+            'com.h2database:h2:2.4.240',
+            // CVE-2026-22747: spring-security-web 6.5.10 is the latest available release; no fix exists yet upstream.
+            // Remove this exclusion once Spring Security releases a patched version and spring-boot.version is bumped.
+            'org.springframework.security:spring-security-web:6.5.10',
+        ] as Set
```
jamesfredley, line 81: **#2 - Version-pinned exclusions are brittle.** These coordinates are pinned to exact versions (`hibernate-core:5.6.11.Final`, `h2:2.4.240`, `spring-security-web:6.5.10`). When any of those deps bump - e.g. a future `spring-boot.version` bump can pull a different `spring-security-web` - the exclusion **stops matching**, the suppressed CVE silently reappears, and the scan starts failing with no obvious cause. Prefer excluding by coordinate without the version if the plugin supports it, or add a check tying each exclusion to the version the BOM manages. Also: the `hibernate-core` line (52) has no explaining comment while the other two do - please add one (which CVE, why excluded).

.github/workflows/vulnerability-scan.yml
```diff
@@ -0,0 +1,85 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+jobs:
+  scan-grails-core:
+    name: "OSS Index Scan - grails-core"
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      security-events: write
```
jamesfredley: **#3 - Unused permission.** `security-events: write` is granted but nothing ever uploads SARIF - the scan output only goes to `$GITHUB_STEP_SUMMARY`. Either drop this permission, or add the `upload-sarif` step it implies.

.github/workflows/vulnerability-scan.yml
```diff
@@ -0,0 +1,85 @@
...
+          java-version: 17
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
+        with:
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+      - name: "🔍 Run OSS Index Vulnerability Scan"
+        env:
+          SONATYPE_GUIDE_USERNAME: ${{ secrets.SONATYPE_GUIDE_USERNAME }}
+          SONATYPE_GUIDE_TOKEN: ${{ secrets.SONATYPE_GUIDE_TOKEN }}
+        continue-on-error: true
```
jamesfredley, line 53: **#4 - Scan never fails CI.** With `continue-on-error: true`, a detected vulnerability produces a red ❌ in the job summary but the workflow still reports success. If report-only is intended (weekly informational scan), great - just confirming this is deliberate and not meant to gate.

jamesfredley (approved):
Approving - the scanning setup is solid and the ASM un-shading + BOM version consistency are correct (verified locally with `validateDependencyVersions`). I left 4 inline notes (#1-#4) for quick follow-ups James can knock out; none are blockers.

1. Javadoc contradicts behavior (skips vs throws).
2. `excludeCoordinates` pinned to exact versions are brittle - suppressed CVEs silently reappear on the next bump, and the hibernate entry lacks a justifying comment.
3. `security-events: write` permission is granted but no SARIF is ever uploaded.
4. `continue-on-error: true` means the scan never fails CI - confirm that's intentional (report-only).

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/VulnerabilityScanPlugin.groovy
jdaugherty: This was actually intentional - we have to review these and fix as soon as we can.

PR #15645, opened 2026-05-08
7.1.x FIX: Rank URL mapping specificity by literal token count before wildcard captures

codeconsole:
When wildcard validation is enabled, AbstractGrailsControllerUrlMappings now sorts validated wildcard matches first by literal URL-token count (more literals = more specific) and only then by non-routing capture count. Previously a generic pattern like /$id/$action/$imageId could win over a more specific route such as /list/$stage/$sort because both had the same wildcard-capture status, causing the wrong controller/action to be resolved regardless of declaration order. Adds two WildcardActionValidationSpec cases covering both declaration orders. 

https://github.com/apache/grails-core/pull/15525
https://github.com/apache/grails-core/pull/15542

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -59,7 +59,7 @@ micronautPlatformVersion=4.9.4
 
 # Libraries only specific to test apps, these should not be exposed
 ersatzVersion=4.0.1
-grailsSpringSecurityVersion=7.0.2-SNAPSHOT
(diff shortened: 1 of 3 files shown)


jdaugherty:
@codeconsole looks like this broke some tests. if you can fix them, I'm happy to follow-up with a review.

jdaugherty (approved):

sbglasius (approved):

PR #15653, opened 2026-05-11
Gh 14503 unique discriminator null spec

jagiro:
Follow-up to #14503 per @borinquenkid's request:
*"add a spec to H5 and H7 core modules and mark them as pending and then we
will follow up on the fix"*.

This PR adds the failing scenario as a `@PendingFeature` Spock test in both
`grails-data-hibernate5/core` and `grails-data-hibernate7/core` test modules.
It does **not** propose a fix. The fix approach is still open per
@jdaugherty's "we should discuss this as a team before proposing a fix"
(option 1 — skip-on-null SQL-standard, option 2 — explicit `IS NULL` check,
or configurable).

### What this PR adds

In both modules, `MultiColumnUniqueConstraintSpec.groovy` gains:

- A new `@Entity DomainTwo` with `name unique: ['discriminator']` and
  `discriminator nullable: true`.
- A new `@PendingFeature` test case:
  *"save with null discriminator succeeds when only non-null discriminator
  rows exist for the same main field"*.
- The new domain registered in `manager.addAllDomainClasses(...)`.

### Verification done locally

- `./gradlew :grails-data-hibernate5-core:test --tests "*MultiColumnUniqueConstraintSpec*"`
  → 4 tests, 3 passed, 1 skipped (the new `@PendingFeature`, as expected).
- `./gradlew :grails-data-hibernate7-core:test --tests "*MultiColumnUniqueConstraintSpec*"`
  → same result.
- Sanity check: temporarily removed `@PendingFeature` from the H5 file and
  re-ran. The test FAILED with `Condition not satisfied: saved != null`
  (the `save()` call returns null because the validator rejects the new row
  as a duplicate even though the existing row has a non-null discriminator).
  Restored `@PendingFeature` after confirming.
- Full module test runs (`./gradlew :grails-data-hibernate5-core:test
  :grails-data-hibernate7-core:test`): 2892 tests, 5 failed, 30 skipped.
  All 5 failures are pre-existing environmental issues (testcontainers
  cannot find a Docker environment on this machine):
  `RLikeHibernate7Spec` (3 tests), `GrailsSequenceGeneratorEnumSpec`,
  `HibernateDatastoreIntegrationSpec`. None of the failures touch the
  files modified in this PR. CI with Docker available should report
  these as passing.

When the fix lands and the test starts passing, Spock will throw
`PendingFeatureNotFinished` as a reminder to remove the annotation.

### Note on AI tooling

This contribution was prepared with assistance from Claude (Anthropic).
The reproducer was first verified in an external Grails 7.x application
and the spec was authored, run, and sanity-checked locally before
submission, per the ASF Generative Tooling policy.

### DIFF
--- a/grails-data-hibernate5/core/src/test/groovy/grails/gorm/specs/MultiColumnUniqueConstraintSpec.groovy
+++ b/grails-data-hibernate5/core/src/test/groovy/grails/gorm/specs/MultiColumnUniqueConstraintSpec.groovy
@@ -23,11 +23,12 @@ import org.apache.grails.data.hibernate5.core.GrailsDataHibernate5TckManager
 import org.apache.grails.data.testing.tck.base.GrailsDataTckSpec
 import org.springframework.dao.DataIntegrityViolationException
 import spock.lang.Issue
+import spock.lang.PendingFeature
 
 @Issue('https://github.com/apache/grails-data-mapping/issues/617')
 class MultiColumnUniqueConstraintSpec extends GrailsDataTckSpec<GrailsDataHibernate5TckManager> {
     void setupSpec() {
-        manager.addAllDomainClasses([DomainOne, Task1, TaskLink])
+        manager.addAllDomainClasses([DomainOne, Task1, TaskLink, DomainTwo])
     }
 
     void "test generated unique constraints"() {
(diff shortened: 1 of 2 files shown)


jdaugherty:
@jagiro I'm guessing you tried to merge 8.x into 7.x?  Should the base of this be 8.x?

jdaugherty:
@jagiro I see in your issue comment this was meant to be merged against 8.0.x-hibernate7 so I've updated the base branch.

jagiro:
Thanks @jdaugherty — 8.0.x-hibernate7 is the right target, my mistake on the base in the PR form. Status & scope cross-posted on #14503 for @borinquenkid.

jdaugherty:
@jagiro we're trying to get the hibernate changes merged.  We haven't forgotten about you but there's a significant amount of reviews going on so we'll have to get this merged once that code line is merged.

borinquenkid (approved):
@jagiro Good Spec
@jdaugherty You need to merge this ASAP into the hibernate-7 branch

PR #15666, opened 2026-05-16
GrailsUtil: honor stackTraceFiltererClass and logFullStackTraceOnFilter

codeconsole:
`grails.logging.stackTraceFiltererClass` config value is not honored by `GrailsUtil.deepSanitize` 

Let's say you customize the `stackTraceFiltererClass` via yml or groovy
`application.groovy`
```groovy
grails.logging.stackTraceFiltererClass=com.pixoto.grails.NonLoggingStackTraceFilterer.class
```

 GroovyPageView.java:103:
`GrailsUtil.deepSanitize`  dumps a massive exception to your console and there is no way around it even if you defined a custom `stackTraceFiltererClass`. Any gap error results in massive spam.

```groovy
  protected void handleException(Exception exception, GroovyPagesTemplateEngine engine) {
      GrailsUtil.deepSanitize(exception);   // ← uses the hardcoded static filterer
      ...
      if (exception instanceof GroovyPagesException) {
          throw (GroovyPagesException) exception;
      }
      ...
  }
```

This PR uses Holders to route to the appropriate `stackTraceFiltererClass`

## Summary

Follow-up to #15564. `GrailsUtil` held a hardcoded `private static final DefaultStackTraceFilterer` that ignored both `grails.logging.stackTraceFiltererClass` (the class-swap key that `GrailsExceptionResolver` honors) and the new `grails.exceptionresolver.logFullStackTraceOnFilter` flag introduced by #15564.

Non-resolver callers of the filterer — most visibly `GroovyPageView.handleException` → `GrailsUtil.deepSanitize` on GSP view-render exceptions, plus scheduled jobs and custom code calling `sanitizeRootCause` / `deepSanitize` directly — produced `StackTrace` logger emissions that no config key could suppress. The only workaround was silencing the `StackTrace` logger in logback, which the user guide already calls out as a fallback rather than the intended control surface.

This PR closes that gap: `GrailsUtil` now resolves its filterer lazily on first use from `Holders.findApplication().getConfig()`, honoring the same two keys the resolver honors and propagating `logFullStackTraceOnFilter` to `DefaultStackTraceFilterer` instances exactly the way `GrailsExceptionResolver.applyLogFullStackTraceOnFilter()` does.

## Why this matters

Pre-7.1, applications could set `grails.logging.stackTraceFiltererClass` to a custom filterer and reach every framework caller — including `GrailsUtil`. The transition through #15564 and the years prior left `GrailsUtil` as the one path the key never reached, which only becomes visible when GSP rendering throws (because that's the path that calls `GrailsUtil.deepSanitize` directly, bypassing the resolver). Apps hitting this today see 2 unfilterable `StackTrace` records per render-time exception in 7.1.x (3 in 7.0.x, before #15564 dropped the redundant trailing `filter(source)`), with no Grails-level knob.

After this PR, setting `grails.exceptionresolver.logFullStackTraceOnFilter: false` silences every caller of the filterer — resolver and `GrailsUtil` alike — without logback intervention.

## Design

- **Lazy + cached.** Pre-context callers (early init, plain `main`, tests that don't wire an application) get a fresh uncached `DefaultStackTraceFilterer` so a later call after the context boots can still populate the cache. Post-context callers resolve from config and cache for JVM lifetime, matching the historical semantics of the previous `static final` field.
- **Backwards compatible.** Unset config yields `DefaultStackTraceFilterer` with `logFullStackTraceOnFilter=true` — identical to the previous hardcoded value. No behavior change for apps that don't set the new keys.
- **Defensive.** Every config-read and instantiation path swallows `Throwable` and falls back to `DefaultStackTraceFilterer` with a logged warning. A bad config value can't break `GrailsUtil` callers.
- **Respects custom filterers.** `setLogFullStackTraceOnFilter` only applied when the resolved instance is a `DefaultStackTraceFilterer` (or subclass) — matching `GrailsExceptionResolver.applyLogFullStackTraceOnFilter()` exactly. Custom `StackTraceFilterer` implementations remain responsible for their own logging policy.

## Tests

New `GrailsUtilStackFiltererSpec` covers three branches:

- Falls back to a `DefaultStackTraceFilterer` when no `GrailsApplication` is discoverable
- Honors `grails.logging.stackTraceFiltererClass` (verifies a custom filterer is instantiated and invoked)
- Propagates `logFullStackTraceOnFilter=false` to `DefaultStackTraceFilterer` instances (verifies no `StackTrace` emission)

Existing `GrailsUtilTests` and `StackTraceFiltererSpec` are unchanged and continue to pass. Full `:grails-core:test` and `:grails-web-mvc:test` suites green locally.

## Docs

- `loggingFullStackTraces.adoc`: NOTE block clarifying that `GrailsUtil` now participates in the same emission policy as the resolver, so the matrix applies to `GrailsUtil`-driven paths too (including the GSP render path).
- `upgrading71x.adoc` §2.13: short paragraph noting `GrailsUtil` now honors both keys and that apps previously silencing the `StackTrace` logger in logback purely to suppress GSP-render noise can now use `logFullStackTraceOnFilter: false` instead.

## Test plan

- [ ] CI passes (grails-core + grails-web-mvc test suites)
- [ ] Existing apps with no new-key config: identical log output to pre-PR (verified via default-path test)
- [ ] App with `grails.exceptionresolver.logFullStackTraceOnFilter: false`: GSP-render exceptions produce zero `StackTrace` records (manually verified against a repro app on 7.2.0-SNAPSHOT after applying the patch)
- [ ] App with custom `grails.logging.stackTraceFiltererClass`: `GrailsUtil.deepSanitize` routes through the custom class (verified via custom-class test)

### DIFF
--- a/grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
+++ b/grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
@@ -36,6 +36,16 @@ public interface StackTraceFilterer {
     String SYS_PROP_DISPLAY_FULL_STACKTRACE = "grails.full.stacktrace";
 
     /**
+     * Name under which the {@link org.apache.grails.core.GrailsBootstrapRegistryInitializer}
+     * promotes the config-resolved filterer as an {@code ApplicationContext} singleton bean,
+     * so later-lifecycle consumers (e.g. {@code GrailsExceptionResolver}) reuse the same
+     * instance instead of instantiating a second copy from config.
+     *
+     * @since 8.0
+     */
+    String BEAN_NAME = "stackTraceFilterer";
+
+    /**
      * Adds a package name that should be filtered
      *
      * @param name The name of the package
--- a/grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
+++ b/grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
@@ -36,16 +36,6 @@ public interface StackTraceFilterer {
     String SYS_PROP_DISPLAY_FULL_STACKTRACE = "grails.full.stacktrace";
 
     /**
-     * Name under which the {@link org.apache.grails.core.GrailsBootstrapRegistryInitializer}
-     * promotes the config-resolved filterer as an {@code ApplicationContext} singleton bean,
-     * so later-lifecycle consumers (e.g. {@code GrailsExceptionResolver}) reuse the same
-     * instance instead of instantiating a second copy from config.
-     *
-     * @since 8.0
-     */
-    String BEAN_NAME = "stackTraceFilterer";
-
-    /**
      * Adds a package name that should be filtered
      *
      * @param name The name of the package
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -43,15 +43,47 @@ public class GrailsUtil {
     private static final boolean LOG_DEPRECATED = Boolean.valueOf(System.getProperty("grails.log.deprecated", String.valueOf(Environment.isDevelopmentMode())));
 
     /**
-     * Lazily-resolved filterer used by {@link #printSanitizedStackTrace}, {@link #sanitizeRootCause}
-     * and {@link #deepSanitize}. Cached once a {@link GrailsApplication} is discoverable via
-     * {@link Holders#findApplication()} so the config-driven class and emission flag are read
-     * exactly once. Volatile to publish the cached value safely; double-checked init in
-     * {@link #resolveStackFilterer()}.
+     * Default filterer used before {@link #initializeStackFilterer(GrailsApplication)} runs (CLI,
+     * tests that don't boot a context, plain {@code main()} usage). Preserves the pre-PR behaviour
+     * of a single hardcoded {@link DefaultStackTraceFilterer} instance for the JVM lifetime when no
+     * application is wired.
      */
-    private static volatile StackTraceFilterer stackFilterer;
+    private static final StackTraceFilterer FALLBACK_FILTERER = new DefaultStackTraceFilterer();
+
+    /**
+     * Active filterer for {@link #printSanitizedStackTrace}, {@link #sanitizeRootCause} and
+     * {@link #deepSanitize}. Starts as {@link #FALLBACK_FILTERER} and is replaced with a
+     * config-driven instance when {@link #initializeStackFilterer(GrailsApplication)} runs during
+     * Grails bootstrap. Volatile so the bootstrap-time write publishes safely to the request
+     * threads that read it later.
+     */
+    private static volatile StackTraceFilterer stackFilterer = FALLBACK_FILTERER;
 
     private GrailsUtil() {
+    }
+
+    /**
+     * Installs a {@link StackTraceFilterer} resolved from the given application's config, replacing
+     * the default fallback. Reads {@link Settings#SETTING_LOGGING_STACKTRACE_FILTER_CLASS} for the
+     * filterer class and propagates {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER} to
+     * instances of {@link DefaultStackTraceFilterer}. Called by {@code GrailsExceptionResolver}
+     * during Spring bean wiring (which is the same point the resolver consults these keys for its
+     * own filterer), so request-time callers of the static {@code sanitize}/{@code deepSanitize}
+     * methods see the configured instance.
+     *
+     * <p>No-ops when {@code application} is null. Safe to call more than once — the last successful
+     * invocation wins.
+     *
+     * @since 7.1.2
+     */
+    public static void initializeStackFilterer(GrailsApplication application) {
+        if (application == null) {
+            return;
+        }
+        StackTraceFilterer instance = createConfiguredFilterer(application);
+        if (instance != null) {
+            stackFilterer = instance;
+        }
     }
 
     /**
@@ -119,7 +151,7 @@ public class GrailsUtil {
     }
 
     public static void printSanitizedStackTrace(Throwable t, PrintWriter p) {
-        printSanitizedStackTrace(t, p, resolveStackFilterer());
+        printSanitizedStackTrace(t, p, stackFilterer);
     }
 
     public static void printSanitizedStackTrace(Throwable t, PrintWriter p, StackTraceFilterer stackTraceFilterer) {
@@ -157,7 +189,7 @@ public class GrailsUtil {
      * @return The root cause exception instance, with its stace trace modified to filter out grails runtime classes
      */
     public static Throwable sanitizeRootCause(Throwable t) {
-        return resolveStackFilterer().filter(extractRootCause(t));
+        return stackFilterer.filter(extractRootCause(t));
     }
 
     /**
@@ -167,70 +199,26 @@ public class GrailsUtil {
      * @return The root cause exception instances, with stack trace modified to filter out grails runtime classes
      */
     public static Throwable deepSanitize(Throwable t) {
-        return resolveStackFilterer().filter(t, true);
-    }
-
-    /**
-     * Returns the {@link StackTraceFilterer} used by this class, lazily initialised from the
-     * Grails application config when one is discoverable. Honours
-     * {@link Settings#SETTING_LOGGING_STACKTRACE_FILTER_CLASS} (the filterer class — same key
-     * the exception resolver consults) and propagates
-     * {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER} to instances of
-     * {@link DefaultStackTraceFilterer}.
-     *
-     * <p>While no {@link GrailsApplication} is available (early-init paths, plain {@code main}
-     * usage, tests that don't wire one up) a fresh {@link DefaultStackTraceFilterer} is returned
-     * and <em>not</em> cached — so once the application context boots, the next call resolves
-     * the configured filterer for real. After that the value is cached for the lifetime of the
-     * JVM, matching the historical behaviour of the previous {@code static final} field.
-     */
-    private static StackTraceFilterer resolveStackFilterer() {
-        StackTraceFilterer cached = stackFilterer;
-        if (cached != null) {
-            return cached;
-        }
-        GrailsApplication application = findApplicationQuietly();
-        if (application == null) {
-            // No application discoverable yet — return an uncached default. A later call,
-            // once the context is up, will run through the configured-resolution branch
-            // and populate the cache.
-            return new DefaultStackTraceFilterer();
-        }
-        synchronized (GrailsUtil.class) {
-            cached = stackFilterer;
-            if (cached != null) {
-                return cached;
-            }
-            stackFilterer = createConfiguredFilterer(application);
-            return stackFilterer;
-        }
-    }
-
-    private static GrailsApplication findApplicationQuietly() {
-        try {
-            return Holders.findApplication();
-        }
-        catch (Throwable ignored) {
-            return null;
-        }
+        return stackFilterer.filter(t, true);
     }
 
     private static StackTraceFilterer createConfiguredFilterer(GrailsApplication application) {
         Class<? extends StackTraceFilterer> filtererClass = DefaultStackTraceFilterer.class;
         boolean logOnFilter = true;
-        try {
-            Config config = application.getConfig();
-            if (config != null) {
-                filtererClass = config.getProperty(
-                        Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
-                        Class.class, DefaultStackTraceFilterer.class);
-                logOnFilter = config.getProperty(
-                        Settings.SETTING_LOG_FULL_STACKTRACE_ON_FILTER,
-                        Boolean.class, true);
+        Config config = application.getConfig();
+        if (config != null) {
+            Class<? extends StackTraceFilterer> configured = config.getProperty(
+                    Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
+                    Class.class, DefaultStackTraceFilterer.class);
+            if (configured != null) {
+                filtererClass = configured;
             }
-        }
-        catch (Throwable t) {
-            LOG.warn("Unable to resolve StackTraceFilterer config; using default: " + t.getMessage());
+            Boolean configuredLogOnFilter = config.getProperty(
+                    Settings.SETTING_LOG_FULL_STACKTRACE_ON_FILTER,
+                    Boolean.class, Boolean.TRUE);
+            if (configuredLogOnFilter != null) {
+                logOnFilter = configuredLogOnFilter;
+            }
         }
         StackTraceFilterer instance;
         try {
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -22,6 +22,7 @@ import java.io.PrintWriter;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.BeanUtils;
 
 import grails.config.Config;
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -22,7 +22,11 @@ import java.io.PrintWriter;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.BeanUtils;
 
+import grails.config.Config;
+import grails.config.Settings;
+import grails.core.GrailsApplication;
 import org.grails.exceptions.reporting.DefaultStackTraceFilterer;
 import org.grails.exceptions.reporting.StackTraceFilterer;
 
@@ -36,7 +40,15 @@ public class GrailsUtil {
 
     private static final Log LOG = LogFactory.getLog(GrailsUtil.class);
     private static final boolean LOG_DEPRECATED = Boolean.valueOf(System.getProperty("grails.log.deprecated", String.valueOf(Environment.isDevelopmentMode())));
-    private static final StackTraceFilterer stackFilterer = new DefaultStackTraceFilterer();
+
+    /**
+     * Lazily-resolved filterer used by {@link #printSanitizedStackTrace}, {@link #sanitizeRootCause}
+     * and {@link #deepSanitize}. Cached once a {@link GrailsApplication} is discoverable via
+     * {@link Holders#findApplication()} so the config-driven class and emission flag are read
+     * exactly once. Volatile to publish the cached value safely; double-checked init in
+     * {@link #resolveStackFilterer()}.
+     */
+    private static volatile StackTraceFilterer stackFilterer;
 
     private GrailsUtil() {
     }
@@ -106,7 +118,7 @@ public class GrailsUtil {
     }
 
     public static void printSanitizedStackTrace(Throwable t, PrintWriter p) {
-        printSanitizedStackTrace(t, p, stackFilterer);
+        printSanitizedStackTrace(t, p, resolveStackFilterer());
     }
 
     public static void printSanitizedStackTrace(Throwable t, PrintWriter p, StackTraceFilterer stackTraceFilterer) {
@@ -144,7 +156,7 @@ public class GrailsUtil {
      * @return The root cause exception instance, with its stace trace modified to filter out grails runtime classes
      */
     public static Throwable sanitizeRootCause(Throwable t) {
-        return stackFilterer.filter(extractRootCause(t));
+        return resolveStackFilterer().filter(extractRootCause(t));
     }
 
     /**
@@ -154,7 +166,84 @@ public class GrailsUtil {
      * @return The root cause exception instances, with stack trace modified to filter out grails runtime classes
      */
     public static Throwable deepSanitize(Throwable t) {
-        return stackFilterer.filter(t, true);
+        return resolveStackFilterer().filter(t, true);
+    }
+
+    /**
+     * Returns the {@link StackTraceFilterer} used by this class, lazily initialised from the
+     * Grails application config when one is discoverable. Honours
+     * {@link Settings#SETTING_LOGGING_STACKTRACE_FILTER_CLASS} (the filterer class — same key
+     * the exception resolver consults) and propagates
+     * {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER} to instances of
+     * {@link DefaultStackTraceFilterer}.
+     *
+     * <p>While no {@link GrailsApplication} is available (early-init paths, plain {@code main}
+     * usage, tests that don't wire one up) a fresh {@link DefaultStackTraceFilterer} is returned
+     * and <em>not</em> cached — so once the application context boots, the next call resolves
+     * the configured filterer for real. After that the value is cached for the lifetime of the
+     * JVM, matching the historical behaviour of the previous {@code static final} field.
+     */
+    private static StackTraceFilterer resolveStackFilterer() {
+        StackTraceFilterer cached = stackFilterer;
+        if (cached != null) {
+            return cached;
+        }
+        GrailsApplication application = findApplicationQuietly();
+        if (application == null) {
+            // No application discoverable yet — return an uncached default. A later call,
+            // once the context is up, will run through the configured-resolution branch
+            // and populate the cache.
+            return new DefaultStackTraceFilterer();
+        }
+        synchronized (GrailsUtil.class) {
+            cached = stackFilterer;
+            if (cached != null) {
+                return cached;
+            }
+            stackFilterer = createConfiguredFilterer(application);
+            return stackFilterer;
+        }
+    }
+
+    private static GrailsApplication findApplicationQuietly() {
+        try {
+            return Holders.findApplication();
+        }
+        catch (Throwable ignored) {
+            return null;
+        }
+    }
+
+    private static StackTraceFilterer createConfiguredFilterer(GrailsApplication application) {
+        Class<? extends StackTraceFilterer> filtererClass = DefaultStackTraceFilterer.class;
+        boolean logOnFilter = true;
+        try {
+            Config config = application.getConfig();
+            if (config != null) {
+                filtererClass = config.getProperty(
+                        Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
+                        Class.class, DefaultStackTraceFilterer.class);
+                logOnFilter = config.getProperty(
+                        Settings.SETTING_LOG_FULL_STACKTRACE_ON_FILTER,
+                        Boolean.class, true);
+            }
+        }
+        catch (Throwable t) {
+            LOG.warn("Unable to resolve StackTraceFilterer config; using default: " + t.getMessage());
+        }
+        StackTraceFilterer instance;
+        try {
+            instance = BeanUtils.instantiateClass(filtererClass, StackTraceFilterer.class);
+        }
+        catch (Throwable t) {
+            LOG.warn("Problem instantiating configured StackTraceFilterer [" + filtererClass.getName() +
+                    "], falling back to default: " + t.getMessage());
+            instance = new DefaultStackTraceFilterer();
+        }
+        if (instance instanceof DefaultStackTraceFilterer) {
+            ((DefaultStackTraceFilterer) instance).setLogFullStackTraceOnFilter(logOnFilter);
+        }
+        return instance;
     }
 
 }
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -23,11 +23,6 @@ import java.io.PrintWriter;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import org.springframework.beans.BeanUtils;
-
-import grails.config.Config;
-import grails.config.Settings;
-import grails.core.GrailsApplication;
 import org.grails.exceptions.reporting.DefaultStackTraceFilterer;
 import org.grails.exceptions.reporting.StackTraceFilterer;
 
@@ -43,7 +38,7 @@ public class GrailsUtil {
     private static final boolean LOG_DEPRECATED = Boolean.valueOf(System.getProperty("grails.log.deprecated", String.valueOf(Environment.isDevelopmentMode())));
 
     /**
-     * Default filterer used before {@link #initializeStackFilterer(GrailsApplication)} runs (CLI,
+     * Default filterer used before {@link #initializeStackFilterer(StackTraceFilterer)} runs (CLI,
      * tests that don't boot a context, plain {@code main()} usage). Preserves the pre-PR behaviour
      * of a single hardcoded {@link DefaultStackTraceFilterer} instance for the JVM lifetime when no
      * application is wired.
@@ -53,7 +48,7 @@ public class GrailsUtil {
     /**
      * Active filterer for {@link #printSanitizedStackTrace}, {@link #sanitizeRootCause} and
      * {@link #deepSanitize}. Starts as {@link #FALLBACK_FILTERER} and is replaced with a
-     * config-driven instance when {@link #initializeStackFilterer(GrailsApplication)} runs during
+     * config-driven instance when {@link #initializeStackFilterer(StackTraceFilterer)} runs during
      * Grails bootstrap. Volatile so the bootstrap-time write publishes safely to the request
      * threads that read it later.
      */
@@ -63,26 +58,24 @@ public class GrailsUtil {
     }
 
     /**
-     * Installs a {@link StackTraceFilterer} resolved from the given application's config, replacing
-     * the default fallback. Reads {@link Settings#SETTING_LOGGING_STACKTRACE_FILTER_CLASS} for the
-     * filterer class and propagates {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER} to
-     * instances of {@link DefaultStackTraceFilterer}. Called by {@code GrailsExceptionResolver}
-     * during Spring bean wiring (which is the same point the resolver consults these keys for its
-     * own filterer), so request-time callers of the static {@code sanitize}/{@code deepSanitize}
-     * methods see the configured instance.
+     * Installs the given {@link StackTraceFilterer}, replacing the default fallback, so that
+     * request-time callers of the static {@code sanitize}/{@code deepSanitize} methods see the
+     * configured instance. Called by {@link org.apache.grails.core.GrailsBootstrapRegistryInitializer}
+     * once the config-resolved filterer is available — before the {@code ApplicationContext}
+     * refreshes, so every app type (web or not) is covered, not just apps that wire a
+     * {@code GrailsExceptionResolver} bean.
      *
-     * <p>No-ops when {@code application} is null. Safe to call more than once — the last successful
-     * invocation wins.
+     * <p>{@code GrailsUtil} intentionally has no dependency on {@code GrailsApplication} or Spring's
+     * {@code Environment} here — the caller is responsible for resolving the configured class and
+     * {@code logFullStackTraceOnFilter} flag and handing over a ready-to-use instance.
      *
-     * @since 7.1.5
+     * <p>No-ops when {@code filterer} is null. Safe to call more than once — the last call wins.
+     *
+     * @since 8.0
      */
-    public static void initializeStackFilterer(GrailsApplication application) {
-        if (application == null) {
-            return;
-        }
-        StackTraceFilterer instance = createConfiguredFilterer(application);
-        if (instance != null) {
-            stackFilterer = instance;
+    public static void initializeStackFilterer(StackTraceFilterer filterer) {
+        if (filterer != null) {
+            stackFilterer = filterer;
         }
     }
 
@@ -200,39 +193,6 @@ public class GrailsUtil {
      */
     public static Throwable deepSanitize(Throwable t) {
         return stackFilterer.filter(t, true);
-    }
-
-    private static StackTraceFilterer createConfiguredFilterer(GrailsApplication application) {
-        Class<? extends StackTraceFilterer> filtererClass = DefaultStackTraceFilterer.class;
-        boolean logOnFilter = true;
-        Config config = application.getConfig();
-        if (config != null) {
-            Class<? extends StackTraceFilterer> configuredFiltererClass = config.getProperty(
-                    Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
-                    Class.class, DefaultStackTraceFilterer.class);
-            if (configuredFiltererClass != null) {
-                filtererClass = configuredFiltererClass;
-            }
-            Boolean configuredLogOnFilter = config.getProperty(
-                    Settings.SETTING_LOG_FULL_STACKTRACE_ON_FILTER,
-                    Boolean.class, Boolean.TRUE);
-            if (configuredLogOnFilter != null) {
-                logOnFilter = configuredLogOnFilter;
-            }
-        }
-        StackTraceFilterer filterer;
-        try {
-            filterer = BeanUtils.instantiateClass(filtererClass, StackTraceFilterer.class);
-        }
-        catch (Throwable t) {
-            LOG.warn("Problem instantiating configured StackTraceFilterer [" + filtererClass.getName() +
-                    "], falling back to default: " + t.getMessage());
-            filterer = new DefaultStackTraceFilterer();
-        }
-        if (filterer instanceof DefaultStackTraceFilterer) {
-            ((DefaultStackTraceFilterer) filterer).setLogFullStackTraceOnFilter(logOnFilter);
-        }
-        return filterer;
     }
 
 }
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -38,10 +38,10 @@ public class GrailsUtil {
     private static final boolean LOG_DEPRECATED = Boolean.valueOf(System.getProperty("grails.log.deprecated", String.valueOf(Environment.isDevelopmentMode())));
 
     /**
-     * Default filterer used before {@link #initializeStackFilterer(StackTraceFilterer)} runs (CLI,
-     * tests that don't boot a context, plain {@code main()} usage). Preserves the pre-PR behaviour
-     * of a single hardcoded {@link DefaultStackTraceFilterer} instance for the JVM lifetime when no
-     * application is wired.
+     * Default filterer used before {@link #initializeStackFilterer(StackTraceFilterer)} runs — that
+     * is, in any context not started through {@code SpringApplication} (CLI, plain {@code main()}
+     * usage, Grails unit tests). Preserves the pre-PR behaviour of a single hardcoded
+     * {@link DefaultStackTraceFilterer} instance for the JVM lifetime when no application is wired.
      */
     private static final StackTraceFilterer FALLBACK_FILTERER = new DefaultStackTraceFilterer();
 
@@ -70,6 +70,8 @@ public class GrailsUtil {
      * {@code logFullStackTraceOnFilter} flag and handing over a ready-to-use instance.
      *
      * <p>No-ops when {@code filterer} is null. Safe to call more than once — the last call wins.
+     * The installed filterer is JVM-global rather than per-context, so in a JVM hosting more than
+     * one application context the last one booted supplies the filterer for all of them.
      *
      * @since 8.0
      */
--- a/grails-core/src/main/groovy/grails/util/GrailsUtil.java
+++ b/grails-core/src/main/groovy/grails/util/GrailsUtil.java
@@ -74,7 +74,7 @@ public class GrailsUtil {
      * <p>No-ops when {@code application} is null. Safe to call more than once — the last successful
      * invocation wins.
      *
-     * @since 7.1.2
+     * @since 7.1.5
      */
     public static void initializeStackFilterer(GrailsApplication application) {
         if (application == null) {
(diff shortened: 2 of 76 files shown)


grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -154,7 +167,84 @@ public static Throwable sanitizeRootCause(Throwable t) {
...
+                return cached;
+            }
+            stackFilterer = createConfiguredFilterer(application);
+            return stackFilterer;
+        }
+    }
+
+    private static GrailsApplication findApplicationQuietly() {
+        try {
+            return Holders.findApplication();
```
jdaugherty: If this throws it's likely trying to fetch this prior to the application being initialized.  It seems like this would be a major issue because before this functioned regardless of the application being initialized.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Verifies that {@link GrailsUtil#deepSanitize}, {@link GrailsUtil#sanitizeRootCause} and
+ * {@link GrailsUtil#printSanitizedStackTrace} honour the same config keys as
+ * {@code GrailsExceptionResolver} — {@code grails.logging.stackTraceFiltererClass} and
+ * {@code grails.exceptionresolver.logFullStackTraceOnFilter}.
+ *
+ * The cached filterer is reset between scenarios via reflection so each test sees a
+ * fresh lookup against its own {@link GrailsApplication}.
+ */
+class GrailsUtilStackFiltererSpec extends Specification {
```
jdaugherty, line 36: Can you point me to an integration test that exercises this logic?

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -154,7 +167,84 @@ public static Throwable sanitizeRootCause(Throwable t) {
...
+     * and <em>not</em> cached — so once the application context boots, the next call resolves
+     * the configured filterer for real. After that the value is cached for the lifetime of the
+     * JVM, matching the historical behaviour of the previous {@code static final} field.
+     */
+    private static StackTraceFilterer resolveStackFilterer() {
+        StackTraceFilterer cached = stackFilterer;
+        if (cached != null) {
+            return cached;
+        }
+        GrailsApplication application = findApplicationQuietly();
```
jdaugherty: Before it always used one instance, but now it's using multiple potentially.  This seems backwards.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -154,7 +167,84 @@ public static Throwable sanitizeRootCause(Throwable t) {
...
+        if (application == null) {
+            // No application discoverable yet — return an uncached default. A later call,
+            // once the context is up, will run through the configured-resolution branch
+            // and populate the cache.
+            return new DefaultStackTraceFilterer();
```
jdaugherty: What scenario is this occuring?  Now we're initializing this every access until the app is online.  Why aren't we just setting this value earlier in the bootstrap context and then referencing outside of the bean scope?

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Agreed — `Holders.findApplication()` doesn't actually throw (Holders.java:124-132 iterates strategies and returns null), and "regardless of the application being initialized" is the right invariant. Switched the whole approach in cdffb8f3 to a bootstrap hook: `GrailsExceptionResolver.setGrailsApplication` calls a new public `GrailsUtil.initializeStackFilterer(application)` during Spring bean wiring, so `GrailsUtil` no longer reaches into `Holders` at all. Hot path is now a single volatile read.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
codeconsole: Currently only the unit spec; same coverage level the parent PR #15564 shipped with for the resolver side of these keys. Happy to add an `@Integration` spec in `grails-test-examples/app2` that boots a real context, sets `grails.logging.stackTraceFiltererClass` in `application.yml`, calls `GrailsUtil.deepSanitize` from inside the running app, and asserts the configured class was used. Prefer `app2` (it already exercises exception handling) or a dedicated minimal app like `config-report`?

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Right, that was a real regression of the prior invariant. Fixed in cdffb8f3 with a `static final FALLBACK_FILTERER` sentinel: the volatile `stackFilterer` field is initialised to the sentinel so CLI/test/`main()` paths see one shared instance for the JVM lifetime, matching the pre-PR `static final` shape. The bootstrap hook (C4 thread) then overwrites it once during Spring bean wiring.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Good point — switched to the bootstrap-hook approach in cdffb8f3. `GrailsExceptionResolver.setGrailsApplication` now calls a new public `GrailsUtil.initializeStackFilterer(application)` immediately after `createStackFilterer()`, mirroring how the resolver consumes the same two config keys. The `static final FALLBACK_FILTERER` sentinel is the field's initial value so CLI/test/`main()` paths work unchanged, but in any Grails web app the configured filterer is pinned during bean wiring — no per-access initialization, no `Holders` lookup at runtime, hot path is one volatile read.

borinquenkid:
*Posted with assistance from Claude Code (Anthropic's CLI agent), doing a pass over open PRs.*

Hi @jdaugherty — flagging this one since it's been quiet since @codeconsole's 2026-05-29 replies to your four review comments. From what I can tell each point got a substantive response: the `Holders` lookup was dropped in favor of a real bootstrap hook (`GrailsExceptionResolver.setGrailsApplication` → `GrailsUtil.initializeStackFilterer`), and the single-shared-instance semantics were restored via a `FALLBACK_FILTERER` sentinel. No pressure, just didn't want it to fall through the cracks — worth a look when you have time.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -157,4 +202,37 @@ public static Throwable deepSanitize(Throwable t) {
...
+    private static StackTraceFilterer createConfiguredFilterer(GrailsApplication application) {
+        Class<? extends StackTraceFilterer> filtererClass = DefaultStackTraceFilterer.class;
+        boolean logOnFilter = true;
+        Config config = application.getConfig();
+        if (config != null) {
+            Class<? extends StackTraceFilterer> configured = config.getProperty(
+                    Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
+                    Class.class, DefaultStackTraceFilterer.class);
+            if (configured != null) {
+                filtererClass = configured;
```
matrei:
```suggestion
            Class<? extends StackTraceFilterer> configuredFiltererClass = config.getProperty(
                    Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS,
                    Class.class, DefaultStackTraceFilterer.class);
            if (configuredFiltererClass != null) {
                filtererClass = configuredFiltererClass;
```

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -157,4 +202,37 @@ public static Throwable deepSanitize(Throwable t) {
...
+        }
+        catch (Throwable t) {
+            LOG.warn("Problem instantiating configured StackTraceFilterer [" + filtererClass.getName() +
+                    "], falling back to default: " + t.getMessage());
+            instance = new DefaultStackTraceFilterer();
+        }
+        if (instance instanceof DefaultStackTraceFilterer) {
+            ((DefaultStackTraceFilterer) instance).setLogFullStackTraceOnFilter(logOnFilter);
+        }
+        return instance;
```
matrei:
```suggestion
        StackTraceFilterer filterer;
        try {
            filterer = BeanUtils.instantiateClass(filtererClass, StackTraceFilterer.class);
        }
        catch (Throwable t) {
            LOG.warn("Problem instantiating configured StackTraceFilterer [" + filtererClass.getName() +
                    "], falling back to default: " + t.getMessage());
            filterer = new DefaultStackTraceFilterer();
        }
        if (filterer instanceof DefaultStackTraceFilterer) {
            ((DefaultStackTraceFilterer) filterer).setLogFullStackTraceOnFilter(logOnFilter);
        }
        return filterer;
```

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -36,11 +41,51 @@ public class GrailsUtil {
...
+     * filterer class and propagates {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER} to
+     * instances of {@link DefaultStackTraceFilterer}. Called by {@code GrailsExceptionResolver}
+     * during Spring bean wiring (which is the same point the resolver consults these keys for its
+     * own filterer), so request-time callers of the static {@code sanitize}/{@code deepSanitize}
+     * methods see the configured instance.
+     *
+     * <p>No-ops when {@code application} is null. Safe to call more than once — the last successful
+     * invocation wins.
+     *
+     * @since 7.1.2
```
matrei:
```suggestion
     * @since 7.1.5
```

matrei (dismissed):
The `@since` has to be updated. The other suggestions would be nice for readability.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Applied, thanks — pushed in 3eed2896.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Applied, thanks — pushed in 3eed2896.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: pushed in 3eed2896

matrei (approved):

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
```diff
@@ -131,6 +132,7 @@ public void setServletContext(ServletContext servletContext) {
     public void setGrailsApplication(GrailsApplication grailsApplication) {
         this.grailsApplication = grailsApplication;
         createStackFilterer();
+        GrailsUtil.initializeStackFilterer(grailsApplication);
```
jdaugherty: This still couples `grails-core` static state to a `grails-web-mvc` bean: any context where this resolver isn't wired (a non-web app, or an app that replaces the `exceptionHandler` bean with a different resolver) silently stays on the fallback with no warning.

Isn't this what `GrailsBootstrapRegistryInitializer` is meant to solve? Its close listener fires at the end of `prepareContext()` — environment fully bound, before `refresh()` — so `grails-core` could initialize its own utility there for every app type, and the configured filterer would be active before any bean instantiates (startup failures would honour the config too). It could also promote the filterer as a singleton (the way `PluginDiscovery.BEAN_NAME` is promoted) so this resolver consumes the same instance instead of instantiating a second copy of a possibly custom filterer class.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
```diff
@@ -36,11 +41,51 @@ public class GrailsUtil {
...
+     * during Spring bean wiring (which is the same point the resolver consults these keys for its
+     * own filterer), so request-time callers of the static {@code sanitize}/{@code deepSanitize}
+     * methods see the configured instance.
+     *
+     * <p>No-ops when {@code application} is null. Safe to call more than once — the last successful
+     * invocation wins.
+     *
+     * @since 7.1.5
+     */
+    public static void initializeStackFilterer(GrailsApplication application) {
```
jdaugherty: If initialization moves to the bootstrap-registry phase, this should take the `Environment` (or simply the resolved class + boolean) rather than `GrailsApplication` — the `grailsApplication` bean doesn't exist until refresh, while `Environment.getProperty(key, Class.class, default)` performs the same String→Class conversion `Config` does. That keeps `GrailsUtil` free of any dependency on the application object and lets the filterer be configured before the context refreshes.

grails-doc/src/en/guide/upgrading/upgrading71x.adoc
```diff
@@ -849,3 +849,14 @@ Set to `false` to disable the side-effect emission and rely solely on `logFullSt
 output. The two flags interact — if both are enabled, a request exception with N causes produces N+1 `StackTrace`
 records (one resolver-driven plus one per throwable visited by the recursive filter walk). The Logging Full
 Stack Traces section of the user guide includes a matrix of behaviours for the four flag combinations.
+
+`GrailsUtil` honours both `grails.logging.stackTraceFiltererClass` and
```
jdaugherty: Same doc concern as the user-guide NOTE: this names `GrailsExceptionResolver.setGrailsApplication`, `GroovyPageView.handleException`, and `GrailsUtil.initializeStackFilterer` wiring details. For the upgrade notes, what a user needs is: (1) the two keys now reach `GrailsUtil`-driven paths including GSP render errors, (2) before the application context is initialized `GrailsUtil` falls back to the default filterer (unchanged behaviour), and (3) apps that silenced the `StackTrace` logger in logback purely to suppress GSP-render noise can now use `logFullStackTraceOnFilter: false` instead. The internal method references should be dropped.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+import java.lang.reflect.Field
+
+/**
+ * Verifies that {@link GrailsUtil#initializeStackFilterer} resolves the configured filterer class
+ * from the application's config and propagates {@code grails.exceptionresolver.logFullStackTraceOnFilter}
+ * to {@link DefaultStackTraceFilterer} instances. Before initialization the FALLBACK_FILTERER
+ * (a {@link DefaultStackTraceFilterer} singleton) is used so CLI/test/main paths work unchanged.
+ */
+class GrailsUtilStackFiltererSpec extends Specification {
```
jdaugherty, line 36: Following up on the earlier integration-test thread: yes, please add the `@Integration` spec — `app2` is the right home since it already exercises exception handling. Boot with `grails.logging.stackTraceFiltererClass` set in config, call `GrailsUtil.deepSanitize` in the running app, and assert the configured class is used. That's the piece the unit spec can't prove: that the bootstrap wiring actually fires in a real app.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
```diff
@@ -0,0 +1,172 @@
...
+    def setup() {
+        previous = currentFilterer()
+        setFilterer(fallbackFilterer())
+    }
+
+    def cleanup() {
+        setFilterer(previous)
+    }
+
+    def 'deepSanitize uses the fallback filterer before initializeStackFilterer is called'() {
```
jdaugherty: `setup()` installs `FALLBACK_FILTERER` via reflection, so this feature asserts the state the test itself just created rather than the class's initial state. As written it can't fail. Either drop it or assert only the behaviour (`deepSanitize` succeeds with no application wired).

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
```diff
@@ -0,0 +1,172 @@
...
+        def baos = new ByteArrayOutputStream()
+        System.setErr(new PrintStream(baos, true))
+
+        when:
+        GrailsUtil.initializeStackFilterer(application)
+        GrailsUtil.deepSanitize(new RuntimeException('boom'))
+
+        then:
+        System.err.flush()
+        !baos.toString().contains('ERROR StackTrace')
```
jdaugherty: `StackTraceFiltererSpec` asserts on the `'Full Stack Trace:'` marker; `'ERROR StackTrace'` depends on the logging pattern, so this negative assertion can pass vacuously even when emission occurs. Please use the same marker, and consider adding the positive control (flag `true` + initialized application → emission does happen through `GrailsUtil.deepSanitize`) so the negative case is proven meaningful.

jdaugherty (dismissed):
I think this really should target 8.x and we should move up the initialization

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Superseded — this comment was on the old lazy `resolveStackFilterer()` design (per-call `Holders.findApplication()` lookup), which no longer exists. The rework moves resolution to `GrailsBootstrapRegistryInitializer`, which sets `GrailsUtil`'s filterer exactly once, before the `ApplicationContext` refreshes. No per-request/per-access resolution happens anymore.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: This is exactly what the rework does now — moved to `GrailsBootstrapRegistryInitializer`, which resolves the configured filterer from the environment and calls `GrailsUtil.initializeStackFilterer(...)` once, before `refresh()`, instead of lazily on each access. See the reply on the CHANGES_REQUESTED review comment for the full design.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Superseded along with the above -- same lazy-resolution design that is gone now. With the bootstrap-phase init, there is a single stackFilterer static field written once (or left at the FALLBACK_FILTERER default), matching the pre-PR single-instance semantics.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
codeconsole: Added grails-test-examples/app2/src/integration-test/groovy/app2/GrailsUtilStackFiltererIntegrationSpec.groovy. Boots app2 with grails.logging.stackTraceFiltererClass=app2.RecordingStackTraceFilterer (via a distinct @TestPropertySource-scoped context so ErrorsControllerSpec/NotFoundHandlerSpec keep the default context), hits a new TestController action that calls GrailsUtil.deepSanitize directly, and asserts the configured class -- not the default -- handled it. That is the one thing the unit specs cannot prove: that the bootstrap wiring actually fires in a real app.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
codeconsole: Fixed -- dropped the assertion on the state the test itself just installed; the case now only asserts noExceptionThrown() (deepSanitize succeeds with no application wired).

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
codeconsole: Fixed -- both tests now check for StackTraceFilterer.FULL_STACK_TRACE_MESSAGE ('Full Stack Trace:'), the actual marker DefaultStackTraceFilterer emits, and I added the positive-control case (default/enabled emission) alongside the disabled one so the negative assertion is proven meaningful.

grails-doc/src/en/guide/upgrading/upgrading71x.adoc
codeconsole: Reworded both the user-guide NOTE and the upgrade-notes paragraph to drop the internal method references (setGrailsApplication, initializeStackFilterer, handleException) and state only the three user-facing points you listed: the two keys now reach GrailsUtil-driven paths including GSP render errors, the pre-context fallback is unchanged, and logFullStackTraceOnFilter: false is now the supported alternative to silencing the StackTrace logger in logback.

grails-core/src/main/groovy/grails/util/GrailsUtil.java
codeconsole: Went with a variant of your second option: GrailsUtil.initializeStackFilterer now takes the already-resolved StackTraceFilterer instance rather than an Environment, so GrailsUtil has zero dependency on Spring/Grails config types at all -- GrailsBootstrapRegistryInitializer does the resolution and hands over a ready instance. One thing worth flagging: environment.getProperty(key, Class.class, default) won't actually do the String->Class conversion -- neither Spring's DefaultConversionService nor Boot's ApplicationConversionService register a String-to-Class converter (checked both jars), so that call throws ConverterNotFoundException for a real class name. The resolver in GrailsBootstrapRegistryInitializer reads the property as a String and resolves it via ClassUtils.forName instead.

codeconsole:
Replying to the CHANGES_REQUESTED review: done both. Retargeted the PR to 8.0.x, and moved stack-filterer initialization into GrailsBootstrapRegistryInitializer's bootstrap-phase close listener (the same point PluginDiscovery is promoted to a bean), so the configured filterer is active before the ApplicationContext refreshes and every app type is covered -- not just apps that wire a GrailsExceptionResolver bean. Replied inline on each of the specific threads with the details. Also picked up the outstanding integration-test ask and the two test-quality issues from the same review while I was in there.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
codeconsole: Moved init into GrailsBootstrapRegistryInitializer, following the PluginDiscovery pattern: it resolves the configured filterer from the Environment in a close listener firing after prepareContext() but before refresh(), installs it into GrailsUtil there (covering every app type, not only web apps with an exception resolver bean), and promotes the same instance as a singleton bean under StackTraceFilterer.BEAN_NAME. createStackFilterer in the resolver now looks up that bean first, falling back to its prior config-based construction only when the bean is absent -- no more duplicate instantiation. The direct call from setGrailsApplication is removed.

sbglasius (approved):

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
```diff
@@ -62,5 +78,60 @@ public void initialize(BootstrapRegistry registry) {
...
+     * {@code environment.getProperty(key, Class.class)} — neither Spring's default conversion
+     * service nor Spring Boot's {@code ApplicationConversionService} register a String-to-Class
+     * converter, so that call would throw {@code ConverterNotFoundException} for a real class name.
+     *
+     * <p>Defensive: any failure reading config or instantiating the configured class falls back to
+     * a plain {@link DefaultStackTraceFilterer}, matching the resolver's own fallback behaviour.
+     */
+    private StackTraceFilterer resolveConfiguredStackTraceFilterer(Environment environment) {
+        Class<? extends StackTraceFilterer> filtererClass = DefaultStackTraceFilterer.class;
+        String configuredClassName = environment.getProperty(Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS);
```
jdaugherty: Reading this key as a `String` drops the configuration form this PR was opened to fix.

`application.groovy` is loaded by `GroovyConfigPropertySourceLoader` into a `NavigableMapPropertySource`, which returns values with their original types. So the form in the description --

```groovy
grails.logging.stackTraceFiltererClass = com.pixoto.grails.NonLoggingStackTraceFilterer.class
```

-- reaches the environment as a `java.lang.Class`. `environment.getProperty(key)` asks for `String.class`, Spring falls back to `Object#toString()`, and the value becomes `"class com.pixoto.grails.NonLoggingStackTraceFilterer"`. `ClassUtils.forName` then fails, the code warns, and installs the default.

That is a regression rather than a no-op, because `createStackFilterer()` now short-circuits on the promoted bean: the resolver loses the custom filterer too. Before this PR it resolved correctly -- `PropertySourcesConfig` keeps the raw value and `NavigableMapConfig.convertValueIfNecessary` returns it through the `targetType.isInstance(originalValue)` branch.

Reading the raw value and accepting both shapes covers it:

```java
Object configured = environment.getProperty(Settings.SETTING_LOGGING_STACKTRACE_FILTER_CLASS, Object.class);
if (configured instanceof Class<?> configuredClass) {
    filtererClass = configuredClass.asSubclass(StackTraceFilterer.class);
}
else if (configured instanceof CharSequence configuredName && StringUtils.hasText(configuredName)) {
    filtererClass = ClassUtils.forName(configuredName.toString(), classLoader)
            .asSubclass(StackTraceFilterer.class);
}
```

Worth noting the flip side, since it argues for keeping both paths: the YAML string form never worked before this PR either -- `Config.getProperty(key, Class.class, ...)` swallows the `ConverterNotFoundException` and silently returns the default -- so the string handling here is a real fix. It just cannot come at the cost of the `Class` form.

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
```diff
@@ -62,5 +78,60 @@ public void initialize(BootstrapRegistry registry) {
...
+        if (StringUtils.hasText(configuredClassName)) {
+            try {
+                filtererClass = ClassUtils.forName(configuredClassName, getClass().getClassLoader())
```
jdaugherty: `getClass()` is `GrailsBootstrapRegistryInitializer`, loaded from the grails-core jar, so this resolves against the jar's ClassLoader.

Under `spring-boot-devtools` -- which grails-forge adds to generated applications as a `developmentOnly` dependency -- application classes are loaded by `RestartClassLoader` while jar classes stay on the base loader. A custom filterer living in `grails-app` or `src/main/groovy` is therefore invisible from here, and every dev-mode restart warns and silently falls back to the default. That is precisely the environment where someone is tuning stack-trace noise.

`postProcessApplicationContext()` runs before `bootstrapContext.close(context)` in `SpringApplication.prepareContext`, so the context's own loader is already set at this point:

```java
ClassUtils.forName(configuredClassName, applicationContext.getClassLoader())
```

That would need the listener to pass the context (or its ClassLoader) into this method alongside the environment.

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
```diff
@@ -62,5 +78,60 @@ public void initialize(BootstrapRegistry registry) {
...
+     * Resolves a {@link StackTraceFilterer} from the given environment, honouring
+     * {@link Settings#SETTING_LOGGING_STACKTRACE_FILTER_CLASS} and
+     * {@link Settings#SETTING_LOG_FULL_STACKTRACE_ON_FILTER}.
+     *
+     * <p>Resolves the class name manually via {@link ClassUtils#forName} rather than
+     * {@code environment.getProperty(key, Class.class)} — neither Spring's default conversion
+     * service nor Spring Boot's {@code ApplicationConversionService} register a String-to-Class
+     * converter, so that call would throw {@code ConverterNotFoundException} for a real class name.
+     *
+     * <p>Defensive: any failure reading config or instantiating the configured class falls back to
```
jdaugherty: The code is not as defensive as this paragraph claims. Only `ClassUtils.forName` and `BeanUtils.instantiateClass` sit inside a `try`; both `environment.getProperty(...)` calls -- and the close-listener body itself -- are unguarded.

So a value that fails conversion propagates out of `bootstrapContext.close()` and fails application startup. `logFullStackTraceOnFilter: yes-please` is enough to trigger it via the `Boolean.class` read on line 120, and a `Class`-valued property would do it on line 109 if the environment's conversion service has no path to `String`. The old resolver path degraded to the default in both cases, because `NavigableMapConfig.convertValueIfNecessary` catches `ConversionException`.

Please wrap the property reads (or the whole listener body) so a bad config value cannot take the application down -- a filterer misconfiguration should never be fatal.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
```diff
@@ -35,6 +35,16 @@ public interface StackTraceFilterer {
...
 
+    /**
+     * Name under which the {@link org.apache.grails.core.GrailsBootstrapRegistryInitializer}
+     * promotes the config-resolved filterer as an {@code ApplicationContext} singleton bean,
+     * so later-lifecycle consumers (e.g. {@code GrailsExceptionResolver}) reuse the same
+     * instance instead of instantiating a second copy from config.
+     *
+     * @since 8.0
+     */
+    String BEAN_NAME = "stackTraceFilterer";
```
jdaugherty: This constant is in the wrong module. grails-bootstrap has no dependency on grails-core -- the dependency runs the other way -- so the `{@link org.apache.grails.core.GrailsBootstrapRegistryInitializer}` on line 39 cannot resolve, and a Spring bean-name constant sitting in the module that deliberately stays out of Spring wiring inverts the layering.

`PluginDiscovery.BEAN_NAME`, the pattern this follows, lives in grails-core right next to the registrar that promotes it. Suggest moving `BEAN_NAME` to grails-core (`GrailsBootstrapRegistryInitializer` itself is a reasonable home) so the constant sits with the code that registers the bean; at minimum the cross-module `{@link}` should become `{@code}`.

Unrelated drive-by while you are in this area: `GrailsConsole.java:429` still says "can't use `StackTraceFilterer#SYS_PROP_DISPLAY_FULL_STACKTRACE` as it is in grails-core", which is stale now that the interface lives here.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
```diff
@@ -453,6 +460,28 @@ protected void createStackFilterer() {
...
+     */
+    protected StackTraceFilterer resolvePromotedStackTraceFilterer() {
+        ApplicationContext context = grailsApplication.getMainContext();
+        if (context == null) {
+            return null;
+        }
+        try {
+            return context.getBean(StackTraceFilterer.BEAN_NAME, StackTraceFilterer.class);
+        }
+        catch (NoSuchBeanDefinitionException e) {
```
jdaugherty: Catching only `NoSuchBeanDefinitionException` narrows what this path used to tolerate. This call happens *before* `createStackFilterer()`'s own `try`, so an application that registers its own bean named `stackTraceFilterer` of an unrelated type gets a `BeanNotOfRequiredTypeException` straight out of `setGrailsApplication`, failing the context. Previously every `Throwable` in filterer construction was contained and degraded to the default.

Either widen the catch to `BeansException` or look the bean up by type instead of by name.

Related question on the same name collision: if an application does define its own `stackTraceFilterer` bean, its definition is registered after the manual singleton, so the resolver and `GrailsUtil` can end up holding different instances. Worth a sentence in the javadoc about which one wins.

grails-doc/src/en/guide/upgrading/upgrading71x.adoc
```diff
@@ -849,3 +849,11 @@ Set to `false` to disable the side-effect emission and rely solely on `logFullSt
 output. The two flags interact — if both are enabled, a request exception with N causes produces N+1 `StackTrace`
 records (one resolver-driven plus one per throwable visited by the recursive filter walk). The Logging Full
 Stack Traces section of the user guide includes a matrix of behaviours for the four flag combinations.
+
+Both `grails.logging.stackTraceFiltererClass` and `grails.exceptionresolver.logFullStackTraceOnFilter` now
```
jdaugherty: This paragraph is in the wrong chapter now that the PR targets `8.0.x`. As written, 7.1 users are told the two keys reach `GrailsUtil`-driven paths, which is not true on 7.1, and anyone upgrading 7.x -> 8 never sees the note at all.

`grails-doc/src/en/guide/upgrading/upgrading80x.adoc` already exists -- the paragraph belongs there. The rest of section 2.13 describes what genuinely shipped in 7.1 and should stay put.

grails-doc/src/en/guide/conf/config/logging/loggingFullStackTraces.adoc
```diff
@@ -97,6 +97,14 @@ log record. It means non-resolver code paths (for example, a scheduled job that
 `GrailsUtil.sanitizeRootCause(ex)` before logging via its own logger) continue to populate the `StackTrace`
 appender without an explicit emission call.
 
+NOTE: Both `grails.logging.stackTraceFiltererClass` and `grails.exceptionresolver.logFullStackTraceOnFilter`
+apply outside the exception resolver too — including GSP view-render exceptions — so this property controls
+emission everywhere the filterer is used, not just for resolver-driven requests. Before the application
+context is initialized (CLI usage, tests that don't boot a context, plain `main()`), the pre-7.1 fallback
```
jdaugherty: "tests that don't boot a context" is not quite the boundary -- Grails unit tests (`ServiceUnitTest` and friends) do boot a context, they just do not go through `SpringApplication`, so they get the fallback too. "any context not started through `SpringApplication`" describes the actual condition without naming internals.

Worth adding one more sentence here as well: the installed filterer is JVM-global static state, so in a JVM hosting more than one application context the last one booted wins. See my comment on the integration spec.

grails-test-examples/app2/src/integration-test/groovy/app2/GrailsUtilStackFiltererIntegrationSpec.groovy
```diff
@@ -0,0 +1,49 @@
...
+ * Boots the full application with {@code grails.logging.stackTraceFiltererClass} set, then drives a
+ * non-resolver code path ({@link grails.util.GrailsUtil#deepSanitize}) to prove the bootstrap wiring in
+ * {@code org.apache.grails.core.GrailsBootstrapRegistryInitializer} actually installs the configured
+ * filterer in a real running app -- the piece {@code GrailsUtilStackFiltererSpec} and
+ * {@code GrailsBootstrapRegistryInitializerSpec} can't prove on their own, since they exercise the
+ * classes directly rather than through a real Spring Boot bootstrap.
+ *
+ * <p>The {@link TestPropertySource} on this class gives it a merged context configuration distinct from
+ * {@link ErrorsControllerSpec} and {@link NotFoundHandlerSpec}, so Spring boots a separate application
+ * context for it and the other specs in this module are unaffected by the custom filterer class.
```
jdaugherty: This claim does not hold. The separate merged context isolates the *Spring* contexts, but `GrailsUtil.stackFilterer` is a JVM-global static -- once this spec's context boots, `RecordingStackTraceFilterer` stays installed for the rest of the fork. `ErrorsControllerSpec` and `NotFoundHandlerSpec` have their contexts cached by the TestContext framework and will not re-boot to overwrite it, so depending on class ordering they run against this filterer.

Nothing fails today because neither of those specs asserts on filterer behaviour, but the comment states the opposite of what actually happens and will mislead whoever touches this next.

The underlying design point is worth deciding explicitly: static installation is last-context-booted-wins, and the static also pins an instance of an application-loaded class for the JVM lifetime (devtools restart loops retain the old `RestartClassLoader`). Resetting to `FALLBACK_FILTERER` on `ContextClosedEvent` would address both; otherwise the single-context assumption should be stated in the docs. Either way please correct this paragraph.

grails-test-examples/app2/src/integration-test/groovy/app2/GrailsUtilStackFiltererIntegrationSpec.groovy
```diff
@@ -0,0 +1,49 @@
...
+ */
+@Integration(applicationClass = Application)
+@TestPropertySource(properties = ['grails.logging.stackTraceFiltererClass=app2.RecordingStackTraceFilterer'])
```
jdaugherty, line 45: This exercises the string form of the key only. Given the `Class`-valued case I flagged in `GrailsBootstrapRegistryInitializer`, please add coverage for a filterer configured as a class literal in `application.groovy` -- that is the shape in the original bug report and the one that currently regresses. A unit-level case in `GrailsBootstrapRegistryInitializerSpec` (a `MapPropertySource` holding the `Class` object rather than its name) would catch it cheaply, without a second application context.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
```diff
@@ -0,0 +1,173 @@
...
+    private static StackTraceFilterer currentFilterer() {
+        filtererField().get(null) as StackTraceFilterer
+    }
+
+    private static void setFilterer(StackTraceFilterer filterer) {
+        filtererField().set(null, filterer)
+    }
+
+    private static StackTraceFilterer fallbackFilterer() {
+        Field field = GrailsUtil.getDeclaredField('FALLBACK_FILTERER')
```
jdaugherty: These specs reach into private statics by reflection (`stackFilterer`, `FALLBACK_FILTERER`) and then assert on them -- `currentFilterer().is(bean)` and friends. `CLAUDE.md` rule 9 asks tests to exercise behaviour through the same surface an end user calls, and every assertion here has a behavioural equivalent: install a recording filterer and assert it was invoked via `GrailsUtil.deepSanitize`, which `'initializeStackFilterer installs the given filterer'` already does well.

Save/restore of the static across features is fair enough given there is no public reset, but the *assertions* should not be white-box -- as written they also couple the specs to field names that are otherwise free to change. Same comment applies to `GrailsBootstrapRegistryInitializerSpec`.

jdaugherty (commented):
Re-reviewing after the bootstrap-phase rework. Everything I raised in the previous two rounds is addressed -- `Holders` is gone, resolution happens once in `GrailsBootstrapRegistryInitializer` before refresh, single-instance semantics are restored, the resolver consumes the promoted bean instead of building its own, `initializeStackFilterer` no longer takes `GrailsApplication`, the docs dropped the internal method references, the self-asserting test case is gone, and the emission assertions now use `FULL_STACK_TRACE_MESSAGE` with a positive control. The integration spec is a genuine improvement over what the parent PR shipped with. Thanks for working through all of it.

Moving the resolution to the environment did introduce a few new problems, one of which I think blocks merge:

- **Blocking:** the environment is read as a `String`, so a filterer configured as a class literal in `application.groovy` -- the exact form in this PR's description -- now stringifies, fails to load, and silently falls back to the default. Because the resolver reuses the promoted bean, it loses the custom filterer as well, where previously it resolved correctly. The string form, conversely, never worked before, so both shapes need handling.
- The class is loaded with grails-core's own ClassLoader, which cannot see application classes under `spring-boot-devtools`.
- The property reads are outside the `try`, so a bad config value now fails startup rather than degrading to the default, contrary to what the javadoc promises.
- The upgrade note landed in `upgrading71x.adoc` although this ships in 8.0.
- `StackTraceFilterer.BEAN_NAME` is in grails-bootstrap with a javadoc link to a grails-core class that module cannot see.
- The integration spec's isolation claim does not hold for JVM-global static state.

Details inline.

jdaugherty:
@codeconsole I posted an AI review on this, I don't think all of these are issues, but I felt some of them would be helpful to get this closed up.

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
codeconsole: Fixed — reads the raw value and accepts both `Class` and `CharSequence`. Worse than you thought: `getProperty(key)` on a `Class` throws `ConverterNotFoundException` out of `bootstrapContext.close()`, so it fails startup rather than falling back. Both forms now covered by tests.

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
codeconsole: Fixed — uses `applicationContext.getClassLoader()`. Test defines a filterer in a loader grails-core can't see, standing in for the devtools split.

grails-core/src/main/groovy/org/apache/grails/core/GrailsBootstrapRegistryInitializer.java
codeconsole: Fixed — both reads guarded independently. `logFullStackTraceOnFilter: yes-please` is now a regression test.

grails-bootstrap/src/main/groovy/org/grails/exceptions/reporting/StackTraceFilterer.java
codeconsole: Moved to `GrailsBootstrapRegistryInitializer.STACK_TRACE_FILTERER_BEAN_NAME`, same as `PluginDiscovery.BEAN_NAME`. On the drive-by: there's no `GrailsConsole` class in the repo any more, so no stale comment to fix.

grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java
codeconsole: Widened to `BeansException`, with a test. On your question: a later bean definition wins — `registerBeanDefinition` → `resetBeanDefinition` → `destroySingleton` kills the manual singleton — so the resolver gets the app's bean and `GrailsUtil` keeps the bootstrap one. Documented on `resolvePromotedStackTraceFilterer`.

grails-doc/src/en/guide/upgrading/upgrading71x.adoc
codeconsole: Moved to `upgrading80x.adoc` §38. 7.1 §2.13 is back to 7.1-only content.

grails-doc/src/en/guide/conf/config/logging/loggingFullStackTraces.adoc
codeconsole: Both fixed — now "any context not started through `SpringApplication`", plus a sentence that the filterer is JVM-global and last-booted wins.

grails-test-examples/app2/src/integration-test/groovy/app2/GrailsUtilStackFiltererIntegrationSpec.groovy
codeconsole: Paragraph rewritten — it said the opposite of what happens. Went with documenting the single-context assumption rather than a `ContextClosedEvent` reset, since resetting on close would break shutdown-time logging. `cleanupSpec` restores the default so it doesn't leak into the fork.

grails-test-examples/app2/src/integration-test/groovy/app2/GrailsUtilStackFiltererIntegrationSpec.groovy
codeconsole: Added in `GrailsBootstrapRegistryInitializerSpec` as you suggested — `MapPropertySource` holding the `Class`, no second context. Positive case plus a `Class` that doesn't implement the interface.

grails-core/src/test/groovy/grails/util/GrailsUtilStackFiltererSpec.groovy
codeconsole: Reflection gone from both specs. Assertions go through `deepSanitize` / `sanitizeRootCause` / `printSanitizedStackTrace`; `initializeStackFilterer` is its own reset. Added coverage for the latter two, which weren't exercised.

codeconsole:
> @codeconsole I posted an AI review on this, I don't think all of these are issues, but I felt some of them would be helpful to get this closed up.

@jdaugherty thanks, please take another pass and see if all your suggestions are met.

borinquenkid:
*Posted with assistance from Claude Code (Anthropic's CLI agent), verifying the fixes in cbfe2fc8.*

@codeconsole I checked cbfe2fc8 against each point from @jdaugherty's 2026-07-26 review by reading the diff and running the affected test suites (new specs, checkstyle, full grails-core + grails-web-mvc). All seven are genuinely resolved, not just asserted in the reply comments:

- Class-literal config now resolves correctly (verified: a raw `Class` value survives `environment.getProperty(key, Object.class)` unconverted)
- ClassLoader fixed to `applicationContext.getClassLoader()` (test proves grails-core's own loader can't see the class, the context's loader can)
- Both property reads are now independently try/caught — a bad value degrades instead of failing `bootstrapContext.close()`
- `STACK_TRACE_FILTERER_BEAN_NAME` moved out of `grails-bootstrap` (zero net diff on `StackTraceFilterer.java` vs `8.0.x`)
- `GrailsExceptionResolver`'s promoted-bean lookup now catches `BeansException` broadly — a same-name bean of the wrong type degrades cleanly
- Doc note moved to `upgrading80x.adoc` §38, sequential numbering, accurate boundary wording
- Reflection is gone from both unit specs

Two things I don't think were caught yet:

1. **`GrailsBootstrapRegistryInitializer.java:155-178` (`resolveFiltererClass`) — silent fallback on a wrong-typed config value.** If `grails.logging.stackTraceFiltererClass` is set to something that's neither a `Class` nor a `CharSequence` (a YAML list, boolean, int), it falls through both `instanceof` checks and returns `DefaultStackTraceFilterer.class` at line 178 with no `LOG.warn`. Every other failure path here logs; this one silently downgrades, contradicting the method's own javadoc ("every config read... degrades... with a logged warning").
2. **Class-literal test coverage doesn't exercise the real `application.groovy` path.** The new `GrailsBootstrapRegistryInitializerSpec` test uses a synthetic `MapPropertySource` holding a raw `Class`, not the actual `GroovyConfigPropertySourceLoader` → `NavigableMapPropertySource` pipeline that parses `application.groovy` — the exact shape from the original bug report. Both classes are available in `grails-core/src/main` and could back a more faithful test.

Everything else I traced (bean-promotion ordering vs. `refresh()`, `volatile` field visibility, the `registerSingleton` → later `registerBeanDefinition` override semantics, test isolation via `cleanup()`/`cleanupSpec()`) checked out under test.

codeconsole:
fixed in 4000654c8e:

1. **Silent wrong-type fallback**: confirmed, real bug. `resolveFiltererClass` now logs a warning when the value is neither `Class` nor `String` before falling back, plus a regression test with a List value.
2. **Test fidelity**: added a test that loads the Class-literal case through the real `GroovyConfigPropertySourceLoader` -> `ConfigSlurper` pipeline (temp `application.groovy` file) instead of a synthetic `MapPropertySource`. For what it's worth, I traced `NavigableMapPropertySource.getProperty` and confirmed it behaves identically to `MapPropertySource` for a raw `Class` value (no extra conversion), so the synthetic test wasn't masking a real behavioral gap -- but the new test is the more direct proof.

Issue #14023, opened 2025-02-16
@GrailsCompileStatic & taglib namespace support in controllers


jdaugherty:
### Issue description

`@GrailsCompileStatic` on a controller will cause compile failures if taglibs are used directly from the controller.

jeffscottbrown:
It may be that an additional extension like the ones at https://github.com/grails/grails-core/tree/2fa846400063c5d3349cb6677244d43e47c91cb0/grails-core/src/main/groovy/org/grails/compiler will help.

jeffscottbrown:
Is there a reliable way at compile time to look at an expression like `foo.bar(...)` and know for sure that `foo` is going to be a valid taglib namespace at runtime?

jdaugherty:
@davydotcom if you want to take on converting taglibs from closures, this is the ticket to do it in =)

jdaugherty:
Turns out that dynamic dispatch can be used already with CompileStatic.  So this doesn't need design and can take an easier approach initially

jamesfredley:
Thanks for the report. Closing as resolved by PR #15669, merged June 18, 2026 and released in Grails 8.0.0-M2. Grails 8 supports direct and namespaced tag-library calls from `@GrailsCompileStatic` controllers, with unit, integration, and documentation coverage.

PR #15669, opened 2026-05-18
Add `@GrailsCompileStatic` support for controllers that call tag libs

jdaugherty:
Fixes https://github.com/apache/grails-core/issues/14023


jamesfredley (commented):
Nice work on this - it tackles a real long-standing pain point and the architectural choice (a `TypeCheckingExtension` hooking the `methodMissing` / `propertyMissing` weave from `TagLibraryInvoker`) looks right. CI is green across all 31 jobs.

Left some inline notes below - mostly minor style/consistency things, plus a couple of behavior cases that might be worth a second look (the broadness of `unresolvedVariable` and the tag-as-property closure form). None are dealbreakers, just thoughts.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+            scopeExit()
+        }
+
+        unresolvedVariable { VariableExpression ve ->
+            if (currentScope?.isController) {
+                currentScope.dynamicNamespaceProperties << ve
+                return makeDynamic(ve)
+            }
+            null
+        }
```
jamesfredley, line 108: This part might be a bit broader than necessary - marking *every* unresolved variable in a controller as dynamic could end up silencing genuine typos. For example:

```groovy
@GrailsCompileStatic
class BookController {
    BookService bookSvc
    def index() {
        bookSvce.list()  // typo - 'bookSvce' is unresolved
    }
}
```

Here `bookSvce` would be silently made dynamic (and added to `dynamicNamespaceProperties`, so the subsequent `.list()` call is also silenced) instead of producing the compile error `@GrailsCompileStatic` users probably expect.

Could it maybe be worth narrowing this? One option would be to defer the dynamic mark until the variable is actually used as the receiver of a method call (i.e. only silence the namespace-dispatcher access pattern `<ident>.<method>(...)`), so standalone typos still surface. Just a thought - happy to be wrong if there's a reason you've gone broader.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+        unresolvedProperty { PropertyExpression pe ->
+            if (currentScope?.isController && isThisReceiver(pe)) {
+                currentScope.dynamicNamespaceProperties << pe
+                return makeDynamic(pe)
+            }
+            null
+        }
```
jamesfredley, line 116: Just flagging a subtle case here: this hook silences `def t = link` at compile time, but at runtime `TagLibraryInvoker.propertyMissing` only returns a `NamespacedTagDispatcher` for namespace names - it doesn't return a `Closure` for tag-property access the way `TagLibrary.propertyMissing` does inside a tag library itself. So a controller with `def t = link` would compile cleanly but throw `MissingPropertyException` at runtime.

Probably fine as a known limitation, but maybe worth a Javadoc note here (or a line in the user docs) so the boundary is explicit?

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * <p>Two calling patterns are supported:
+ * <ul>
+ *   <li>Direct calls on {@code this}: {@code link(controller: 'home')},
+ *       {@code message(code: 'key')}</li>
+ *   <li>Namespaced calls via a namespace dispatcher property:
+ *       {@code g.message(code: 'key')}, {@code my.customTag(attr: 'val')}</li>
+ * </ul>
+ *
+ * @since 7.0
```
jamesfredley: Tiny thing - since this PR targets `8.0.0-SNAPSHOT` and `whatsNew.adoc` describes the change under "introduced in Grails 8", `@since 8.0` would maybe be more accurate here?

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+        beforeVisitClass { ClassNode classNode ->
+            newScope {
+                isController = classNode.name.endsWith(ControllerArtefactHandler.TYPE)
+                dynamicNamespaceProperties = [] as Set
+            }
+        }
+
+        afterVisitClass { ClassNode classNode ->
+            scopeExit()
+        }
```
jamesfredley, line 100: Minor consistency note: the other class-scoped extensions in this package (`ValidateableTypeCheckingExtension`, `DomainMappingTypeCheckingExtension`, `NamedQueryTypeCheckingExtension`) pair `beforeVisitClass` / `afterVisitClass` with an outer `setup { newScope() }` / `finish { scopeExit() }` guard:

```groovy
setup { newScope() }
finish { scopeExit() }
```

The `currentScope?` null-safe accesses below make this functionally safe without them, but could you maybe add the outer pair to keep this in lockstep with the rest of the package?

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+        unresolvedProperty { PropertyExpression pe ->
+            if (currentScope?.isController && isThisReceiver(pe)) {
+                currentScope.dynamicNamespaceProperties << pe
+                return makeDynamic(pe)
+            }
+            null
+        }
+
+        methodNotFound { receiver, name, argList, argTypes, call ->
```
jamesfredley: Could you maybe add parameter types here for consistency with the other extensions in this package? They all use the fully typed form:

```groovy
methodNotFound { ClassNode receiver, String name, ArgumentListExpression argList, ClassNode[] argTypes, MethodCall call ->
```

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+            if (!currentScope?.isController) return null
+            if (isThisReceiver(call)) return makeDynamic(call)
+            if (call instanceof MethodCallExpression && call.objectExpression in currentScope.dynamicNamespaceProperties) return makeDynamic(call)
```
jamesfredley, line 122: Cosmetic only - line 84 uses `currentScope?.isController` but this line dereferences `currentScope.dynamicNamespaceProperties` directly. They're functionally equivalent given the guard above, but `currentScope?.dynamicNamespaceProperties` here might read a touch more uniformly?

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+            null
+        }
+    }
```
jamesfredley, line 127: The other extensions in this package all end `run()` with an explicit `null`. Could you maybe add one here for consistency? Just a style nit.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+    private boolean isThisReceiver(expr) {
```
jamesfredley: Could you maybe type this parameter (perhaps `Expression expr`)? Helps line up with the rest of the codebase's static-typing lean.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @since 7.0
+ */
+class ControllerTagLibTypeCheckingExtension extends GroovyTypeCheckingExtensionSupport.TypeCheckingDSL {
+
+    @Override
+    Object run() {
+        beforeVisitClass { ClassNode classNode ->
+            newScope {
+                isController = classNode.name.endsWith(ControllerArtefactHandler.TYPE)
```
jamesfredley, line 93: Since detection is purely by name suffix, an inner class called something like `FooController` declared inside, say, a service would also receive the silencing treatment. Probably rare in practice - but might be worth a Javadoc note so the behavior is documented?

grails-test-examples/demo33/src/integration-test/groovy/demo/CompileStaticControllerIntegrationSpec.groovy
```diff
@@ -0,0 +1,45 @@
...
+import grails.testing.mixin.integration.Integration
+import org.apache.grails.testing.http.client.HttpClientSupport
+import spock.lang.Specification
+import spock.lang.Tag
+
+@Integration
+@Tag('http-client')
+class CompileStaticControllerSpec extends Specification implements HttpClientSupport {
+
+    void 'controller with @GrailsCompileStatic can call a default-namespace tag directly'() {
```
jamesfredley, line 30: Tiny convenience nit: this integration spec and the unit spec at `src/test/groovy/demo/CompileStaticControllerSpec.groovy` share the same fully qualified name `demo.CompileStaticControllerSpec`. Different source sets so the build is fine, but `--tests "demo.CompileStaticControllerSpec"` becomes ambiguous when running them individually.

Could you maybe rename one to something like `CompileStaticControllerIntegrationSpec` (or `CompileStaticControllerHttpSpec`)?

grails-gsp/plugin/src/test/groovy/org/grails/web/taglib/ControllerCompileStaticTagLibSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+ */
+package org.grails.web.taglib
+
+import grails.artefact.Artefact
+import grails.compiler.GrailsCompileStatic
+import grails.testing.web.controllers.ControllerUnitTest
+import spock.lang.Specification
+
+class ControllerCompileStaticTagLibSpec extends Specification implements ControllerUnitTest<CompileStaticTagController> {
+
```
jamesfredley, line 27: Could it maybe be worth adding a negative test alongside these positive ones? Something pinning down the boundary - e.g. a spec asserting that a genuinely undefined reference inside an `@GrailsCompileStatic` controller still fails to compile - would help guard against the extension drifting to silence more than intended in future changes.

Not a blocker, just a thought.

jdaugherty: Negative tests added.

grails-test-examples/demo33/src/integration-test/groovy/demo/CompileStaticControllerIntegrationSpec.groovy
jdaugherty: renamed

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: Updated

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: Updated

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: With the return currentScope can't be null now.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: Updated the javadoc.  This issue exists across the Grails code base though, so we may want to consider annotating controllers so the compiler could then reliably detect a controller instead of being name driven.  This probably worth a ticket.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: Updated the javadoc.  Using the method() version would fix this.

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
jdaugherty: I'm still investigating this one.

jdaugherty: I tested removing dynamicNamespaceProperties entirely: the tests fail because makeDynamic(ve) alone is insufficient; methodNotFound does fire on dynamic receivers and must
   explicitly make the call dynamic. So the tracking is structurally required for namespace dispatch to work.

  The only genuine fixes are:

  1. Whitelist known namespace names — scan all TagLib classes at compile time, collect their static namespace values, and only make variables dynamic if the name matches a
  known namespace. This is architecturally sound but requires significant infrastructure.
  2. Accept the limitation — document it (done in the Javadoc "Scope note") and rely on the fact that bookSvce.list() fails immediately at runtime with
  MissingPropertyException.

  The negative tests I added do verify a meaningful guarantee: method calls on declared types still fail — bookService.nonExistentMethod() (where bookService is a declared
  field) produces a compile error. The silencing only affects undeclared identifiers.

jdaugherty: The only other solution would be to somehow notify the compiler of namespaces on the classpath, which would require saving the known namespaces and accessing them in the compiler.  This is probably possible, but out of scope of this change.  Ithink this is a net benefit, but we probably should ticket this as an enhancement

codeconsole (commented):
`ControllerTagLibTypeCheckingExtension` doesn't currently compose with other type-checking extensions that also resolve unrecognised method calls.

  With both it and a `CriteriaTypeCheckingExtension` active (#15720), a controller that uses a criteria closure fails:

  ```
[Static type checking] - Reference to method is ambiguous.                                                                            
  Cannot choose between [Object#eq(), Object#eq()]   // and order()
  ```

  ```groovy
// inside a @GrailsCompileStatic controller action
  ScheduledJob.createCriteria().list {
      eq 'scope', params.string('scope')   // eq/order claimed by BOTH extensions → ambiguous
      order 'nextRunAt', 'asc'
  }
  ```

  Both extensions register a candidate method node for the same unresolved call. It looks like the catch-all tag extension should defer (leave the call unhandled) when another extension has already resolved it, rather than both producing a node. Not a blocker for #15669's intended cases, just surfaces when it's combined with a DSL-handling extension.

jdaugherty:
@codeconsole I believe I've addressed your finding in the latest commit.  Concerning the overall change, do you think we should require the registry solution to move forward with this or is this solution "good enough" that we can proceed with it?  Obviously the registry solution is the "correct" solution, but it's a much larger (and impactful change) since every taglib would have to be recompiled.

codeconsole:
> @codeconsole I believe I've addressed your finding in the latest commit. Concerning the overall change, do you think we should require the registry solution to move forward with this or is this solution "good enough" that we can proceed with it? Obviously the registry solution is the "correct" solution, but it's a much larger (and impactful change) since every taglib would have to be recompiled.

Confirmed — tested the latest commit against a real CriteriaTypeCheckingExtension (not the stub) with the catch-all registered last and actual createCriteria{} closures in a @GrailsCompileStatic controller. The DYNAMIC_RESOLUTION deferral resolves it: no more ambiguous-method error.

Good enough to proceed, but the registry would be ideal.  The question is whether to tackle it now in 8.x or later at another major version. This is fine for an intermediary solution. 

At runtime each tag call from a controller stays a dynamic call site (the same methodMissing → tag-lookup →
captureTagOutput path used today in non-statically-compiled Groovy). 

The registry solution would generate real methods, so the call becomes a direct invokevirtual — measurably faster per call.

codeconsole (approved):

jamesfredley (approved):

PR #15673, opened 2026-05-21
Fix 8.0.x build break introduced by 7.2.x merge (6ebc60c6d0)

jamesfredley:
## What this PR fixes

Merge commit [`6ebc60c6d0`](https://github.com/apache/grails-core/commit/6ebc60c6d06b41505598ddb0c44ed129b98ffa65) ("Merge branch '7.2.x' into 8.0.x") plus pre-existing version drift turned every CI job on `8.0.x` red. Reproducing the failures locally surfaced six distinct problems that need to be addressed for `8.0.x` to build again on the Grails 8 / Spring Boot 4 / Spring 7 / Hibernate Jakarta stack. A seventh commit aligns the new file with house Gradle conventions discovered during a post-fix audit.

## CI failures and their fixes

| CI job (representative run id) | Root cause | Fixed by |
|---|---|---|
| `Validate Dependency Versions` - root project | `graphql-java` pinned to `24.3` in `dependencies.gradle` while Spring Boot 4.0.5's BOM ships `25.0`; SB constraint wins and the validator flags the drift | Commit 2 |
| `Validate Dependency Versions` - `cd grails-gradle && validateDependencyVersions` | `gradle-groovy.version=4.0.31` drifted from main `groovy.version=4.0.32`, transitive resolution upgraded all Groovy artifacts to 4.0.32 while grails-gradle BOM still expected 4.0.31 | Commit 6 |
| `Build Grails-Core (ubuntu/macos/windows, JDK 21+25)` | `:grails-test-examples-graphql-spring-boot-app:compileGroovy` couldn't resolve `org.springframework.boot.hibernate.autoconfigure.HibernateJpaAutoConfiguration` because SB4 split it into a separate module that the new example's `build.gradle` didn't depend on | Commit 3 |
| `Functional Tests`, `Hibernate5 Functional Tests`, `Mongodb Functional Tests` | All cascade from the same `:grails-test-examples-graphql-spring-boot-app:compileGroovy` failure | Commit 3 |
| `Hibernate5 Functional Tests` (after compile fix) | `:grails-data-hibernate5-dbmigration:integrationTest` was failing with `LoggerFactory is not a Logback LoggerContext but Logback is on the classpath` because the merge brought back a stray `testRuntimeOnly 'org.slf4j:slf4j-simple'` that competes with SB4's transitive Logback binding | Commit 4 |
| `CodeQL` (`./gradlew testClasses`) | Same compile failure as the main build | Commit 3 |
| `CI - Groovy Joint Validation Build` | Two failures: graphql spring-boot-app compile (already fixed by Commit 3) and dbmigration slf4j conflict (already fixed by Commit 4) | Commits 3 + 4 |

## Latent bugs caught while reproducing locally

These are not in any failed CI log yet because earlier failures stopped the build before reaching them, but they would surface immediately once the compile-blocking failures were fixed.

| Bug | Root cause | Fixed by |
|---|---|---|
| Three string-literal `@EnableAutoConfiguration` exclusion paths still pointed at Spring Boot 3 packages, silently failing to exclude anything in user apps | `ApplicationClassInjector` (Liquibase), `ApplicationClassInjectorSpec` (matching test), `GrailsApplicationCompilerAutoConfiguration` (DataSource) | Commit 1 |
| The forward-ported `grails-test-examples-graphql-grails-multi-datastore-app` didn't compile/run cleanly on the 8.0.x stack | integration specs missing `extends Specification`, deprecated Logback Groovy DSL config, no MongoDB testcontainer, deprecated `session.datastore` GORM API, graphql-java 25's stricter scalar `Coercing` signature | Commit 5 |

## Commits

1. **Fix Spring Boot 4 autoconfigure package paths in exclude strings** - updates three stale Spring Boot 3 package names in `@EnableAutoConfiguration(excludeName=...)` style string literals and the matching unit test assertion.
2. **Stop pinning graphql-java; let spring-boot-dependencies own its version** - drops the redundant 24.3 pin (and matching `bomDependencies` alias) so Spring Boot's 25.0 constraint is the single source of truth.
3. **Add spring-boot-hibernate dep to graphql spring-boot-app for SB4 module split** - adds the runtime dep that SB4's split makes mandatory; uses `implementation` not `compileOnly` because GORM's `@AutoConfigureBefore([HibernateJpaAutoConfiguration])` needs the class loaded.
4. **Remove slf4j-simple from dbmigration; Logback is now the only binding** - removes the duplicate SLF4J binding that aborted Logback initialisation under SB4.
5. **Repair graphql multi-datastore-app integration tests on Grails 8 stack** - five interrelated fixes to make this newly-included module's integration tests pass: Spock `extends Specification`, `logback.xml` (Groovy DSL gone in Logback 1.5), MongoDB testcontainer via `grails-testing-support-mongodb`, modern `GormEnhancer.findStaticApi(...)` access, and `Coercing<ObjectId, String>` for graphql-java 25.
6. **Bump gradle-groovy version to 4.0.32 to match groovy.version** - aligns the grails-gradle subproject's BOM with the main BOM so its `validateDependencyVersions` task stops failing on the Groovy artifact drift.
7. **Use testImplementation for grails-testing-support-mongodb to match house style** - small follow-up to commit 5: every other usage of `grails-testing-support-mongodb` in this repo declares it on `testImplementation`, so this module follows suit (the `integrationTest` source set extends `test` via the `grails-extension-gradle-config` so the testcontainer support still reaches integration tests).

## Local verification

All testing on the merge commit with these patches applied. JDK 21 (Corretto), Gradle 9.4.1, `GRADLE_OPTS=-Xms2G -Xmx5G`.

| Task (forced `--rerun-tasks`) | Result |
|---|---|
| `./gradlew testClasses` (CodeQL target) | exit 0, 693 tasks |
| `./gradlew validateDependencyVersions --continue` (root) | exit 0, 186 tasks |
| `cd grails-gradle && ./gradlew validateDependencyVersions --continue` | exit 0, 10 tasks |
| `./gradlew :grails-data-graphql-core:test` | 254/254 PASSED on graphql-java 25.0 |
| `./gradlew :grails-test-examples-graphql-spring-boot-app:check` | 2/2 PASSED |
| `./gradlew :grails-data-hibernate5-dbmigration:integrationTest` | 3/3 PASSED |
| `./gradlew :grails-core:test --tests ...ApplicationClassInjectorSpec` | 8/8 PASSED |
| `./gradlew :grails-shell-cli:test` | 105/105 PASSED |
| `./gradlew :grails-test-examples-graphql-{grails-test,grails-tenant,grails-docs,grails-multi-datastore,spring-boot}-app` integration tests | All green (multi-datastore-app re-verified after commit 7) |
| `./gradlew build :grails-shell-cli:installDist groovydoc --continue` (matches CI Build Grails-Core command) | 5004 tasks executed, the lone failure was a flaky `HttpClientSupportSpec.httpPatchXml` (`server.verify()` returned `false`) that passes consistently in isolation; not introduced by this PR |

## Convention audit

After the initial six commits, I audited every change in this PR against existing patterns in the repo:

- **Gradle deps**: bare GAV strings are universal in `build.gradle` files (the `bomDependencies['key']` shortcut in `dependencies.gradle` is for BOM generation, not for `build.gradle` consumption). All my additions use bare GAVs. **Verdict: consistent.**
- **Comment style**: `//`-prefixed multi-line for inline explanations in `dependencies.gradle` and `build.gradle`. **Verdict: consistent.**
- **`logback.xml`**: matches the 37-line template used by all three sibling graphql apps. **Verdict: consistent.**
- **Integration specs**: `extends Specification implements GraphQLSpec`, no `@Stepwise` since both feature methods are independent. **Verdict: consistent with sibling specs.**
- **`Coercing<Internal, External>` generic signature**: the in-repo built-ins use `<T, T>` because their Java type is already a valid wire value; `<ObjectId, String>` is the correct divergent-type form for a Java object that does not JSON-serialise to the wire form on its own (graphql-java 25 no longer toString's scalar return values). **Verdict: consistent with the generic contract.**
- **Spring Boot 4 package renames**: `org.springframework.boot.autoconfigure.{jdbc,orm,liquibase,batch,flyway,amqp,cache,jms,mail,data,security,web.reactive}.*` produce zero remaining matches in the repo. The only remaining old-path reference is in `grails-data-neo4j/boot-plugin` which is not included in the main build (the Neo4j module is not used). **Verdict: complete.**
- **`testImplementation` vs `integrationTestImplementation`**: every other use of `grails-testing-support-mongodb` in the repo is on `testImplementation`; commit 7 brings the multi-datastore-app in line.

## Known non-blocking notes

- `HttpClientSupportSpec.httpPatchXml` (custom client variant) intermittently fails under heavy parallel load because Ersatz's `server.verify()` races with the JDK `HttpClient` request completion. The test predates the merge and passes in isolation. It is not a regression from this PR.
- `org.grails.datastore.mapping.mongo.ProjectionsSpec` (3 feature methods) is flaky when the full `./gradlew build` is run on a machine where many tasks share the Docker daemon; root cause is a `ServiceLoader` singleton + `@Shared` manager interaction in `GrailsDataTckSpec` that predates this PR and did not block CI on `fc747cc76a`. Out of scope for the merge-fix PR but worth tracking separately.
- `grails-data-neo4j` (separate Gradle build, not in main settings.gradle) has one remaining Spring Boot 3 import in `Neo4jAutoConfiguration.groovy`. The module is not used and not built by CI, so it is left for a separate Neo4j-focused PR.

Fixes the CI breakage on 8.0.x introduced by [6ebc60c6d0](https://github.com/apache/grails-core/commit/6ebc60c6d06b41505598ddb0c44ed129b98ffa65) plus the gradle-groovy drift that surfaced the same week.


dependencies.gradle
```diff
@@ -78,7 +78,9 @@ ext {
             'commons-codec.version'         : '1.21.0',
             'commons-lang3.version'         : '3.20.0',
             'geb-spock.version'             : '8.0.1',
-            'graphql-java.version'          : '24.3',
+            // graphql-java is managed by spring-boot-dependencies (the SB version always wins),
```
jdaugherty: We should still pin the version so we know when to update the scalars version.  Otherwise they will drift.  The validator plugin will tell us when they mismatch.

grails-data-hibernate5/dbmigration/build.gradle
```diff
@@ -75,9 +75,12 @@ dependencies {
     testImplementation project(':grails-testing-support-web')
     testImplementation 'com.h2database:h2'
 
-    // Liquibase uses JUL for logging -> redirect it to SLF4J to reliably capture its output
+    // Liquibase uses JUL for logging -> redirect it to SLF4J to reliably capture its output.
+    // Spring Boot's starter (transitively via spring-boot-starter-tomcat) already supplies
+    // Logback as the SLF4J binding; adding slf4j-simple here would cause two competing
```
jdaugherty: We can remove the explanation for the transitive include.  I already did this in the hibernate pr b/c this is the right solution

grails-test-examples/graphql/grails-multi-datastore-app/build.gradle
```diff
@@ -70,6 +70,10 @@ dependencies {
 
     testImplementation 'org.apache.grails:grails-testing-support-datamapping'
     testImplementation 'org.apache.grails:grails-testing-support-web'
+
+    // Auto-start a MongoDB testcontainer before integration tests (overrides the
```
jdaugherty: Remove the comments, they're unnecessary

grails-test-examples/graphql/grails-multi-datastore-app/src/integration-test/groovy/myapp/BarIntegrationSpec.groovy
```diff
@@ -43,8 +44,10 @@ class BarIntegrationSpec implements GraphQLSpec {
         """)
         Map obj = resp.body().data.barCreate
 
-        then:
+        then: 'bar is created in the Mongo datastore with a valid ObjectId'
         new ObjectId((String) obj.id)
-        Bar.withNewSession { session -> session.datastore instanceof MongoDatastore }
+        // GORM no longer surfaces `datastore` as a property on the Mongo session;
```
jdaugherty: Remove the comment

grails-test-examples/graphql/grails-multi-datastore-app/src/integration-test/groovy/myapp/FooIntegrationSpec.groovy
```diff
@@ -41,8 +43,10 @@ class FooIntegrationSpec implements GraphQLSpec {
         """)
         Map obj = resp.body().data.fooCreate
 
-        then:
+        then: 'foo is created in the Hibernate datastore'
         obj.id == 1
-        Foo.withNewSession { session -> session.datastore instanceof HibernateDatastore }
+        // GORM no longer surfaces `datastore` as a property on the Hibernate
```
jdaugherty: Remove the comment

grails-test-examples/graphql/spring-boot-app/build.gradle
```diff
@@ -37,6 +37,13 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter'
     implementation 'org.springframework.boot:spring-boot-starter-web'
 
+    // Spring Boot 4 moved HibernateJpaAutoConfiguration to the spring-boot-hibernate module.
+    // It must be on the runtime classpath: GORM's HibernateGormAutoConfiguration
```
jdaugherty: You can remove teh comments they're self evident by update

jdaugherty (approved):

dependencies.gradle
jamesfredley: I know 0% about GraphQL, the weird thing is that https://mvnrepository.com/artifact/com.graphql-java/graphql-java-extended-scalars/versions is still on 24 and the rest of GraphQL is on 25+.  spring-boot-dependendencies does not include java-extended-scalars.  So they are already divergent.

grails-data-hibernate5/dbmigration/build.gradle
jamesfredley: Done in 8275577. Dropped the extended explanation.

grails-test-examples/graphql/grails-multi-datastore-app/build.gradle
jamesfredley: Done in 8275577. Removed the testcontainer-host comment.

grails-test-examples/graphql/grails-multi-datastore-app/src/integration-test/groovy/myapp/BarIntegrationSpec.groovy
jamesfredley: Done in 8275577.

grails-test-examples/graphql/grails-multi-datastore-app/src/integration-test/groovy/myapp/FooIntegrationSpec.groovy
jamesfredley: Done in 8275577.

grails-test-examples/graphql/spring-boot-app/build.gradle
jamesfredley: Done in 8275577. Removed both comment blocks.

dependencies.gradle
jamesfredley: graphql-java 25 and graphql-java-extended-scalars 24 combo does pass the tests.

No breaking changes required (yet)
The extended-scalars library mostly just provides implementations of the Coercing interface.
In the 24.0 release of extended-scalars, they did a big cleanup: they removed all the deprecated Coercing methods and updated to the modern versions (with GraphQLContext + Locale).
Since then, graphql-java 25 and 26 apparently haven't introduced further breaking changes to the scalar/Coercing APIs that would force a new release of extended-scalars.

jdaugherty: I'm just saying we have to review these versions and make sure they're compatible with each other.  The way to do that is to track the versions together.  If the library isn't needed we should remove it and then like you said track the spring boot bom.

jamesfredley: Putting it back for now to move forward, I do not agree given they are diverged versions and I don't love having to manage versions that spring-boot-dependencies advances for us for free, from security standpoint.  If they matched and we could have one variable, I would be on board.  I'll create an issue separately.

Issue #15613, opened 2026-04-30
CI - Groovy Joint Validation Build is broken: spock-core:2.4-groovy-5.0 incompatible with Groovy 4.x snapshot


jamesfredley:
## Summary

The `CI - Groovy Joint Validation Build` workflow (`build_grails` job) has been **failing on every push to `8.0.x` since 2026-04-29** with a Spock vs. Groovy major-version mismatch:

```
> Task :grails-test-examples-issue-11767:compileIntegrationTestGroovy
startup failed:

Could not instantiate global transform class
org.spockframework.compiler.SpockTransform specified at
jar:file:.../spock-core/2.4-groovy-5.0/.../spock-core-2.4-groovy-5.0.jar!/META-INF/services/...
because of exception
org.spockframework.util.IncompatibleGroovyVersionException:
The Spock compiler plugin cannot execute because Spock 2.4.0-groovy-5.0
is not compatible with Groovy 4.0.32-SNAPSHOT.

Spock artifact: file:.../spock-core-2.4-groovy-5.0.jar
Groovy artifact: file:.../groovy-4.0.32-SNAPSHOT.jar

> Task :grails-test-examples-issue-11767:compileIntegrationTestGroovy FAILED
> Task :grails-test-examples-micronaut:compileIntegrationTestGroovy FAILED
```

## Affected runs (10 most recent `8.0.x` runs of the workflow)

| Run | Commit | Result |
|---|---|---|
| [25144143636](https://github.com/apache/grails-core/actions/runs/25144143636) | 7cf1d19 | failure |
| [25134285481](https://github.com/apache/grails-core/actions/runs/25134285481) | 1b33cf8 | failure |
| [25086923954](https://github.com/apache/grails-core/actions/runs/25086923954) | 7174dd2 | failure |
| [25086897319](https://github.com/apache/grails-core/actions/runs/25086897319) | 919802e | failure |
| [25082518134](https://github.com/apache/grails-core/actions/runs/25082518134) | 6e4c6a3 | failure |

The same failure is reproduced on every PR that merges latest `8.0.x` (e.g. [PR #15467 / run 25170545059](https://github.com/apache/grails-core/actions/runs/25170545059/job/73789008877)).

## Root cause

[...]

Issue #11767, opened 2021-03-19
Grails 4: property inside plugin has no effect on environment configuration (extension to #11519)


davidkron:
This is an extension to the already closed issue https://github.com/grails/grails-core/issues/11519

The example which I described in my initial issue where a call to `environment.getProperty('grails.databinding.dateFormats')` inside `BootStrap` doesn't return the values from plugin.yml does work now.

But I discovered another use-case where it still doesn't work:
My example of configuring some default framework parameters like `grails.databinding.dateFormats` inside a "core"-plugin still doesn't work. The concrete problem in the databinding case seems to be, that the configuration is now implemented as a Micronaut `@ConfigurationProperties` bean:
```
@ConfigurationProperties("grails.databinding")
public class DataBindingConfigurationProperties {
```

And to me it seems that the configuration from the plugins gets added to the Micronaut context very late:
https://github.com/grails/grails-core/blob/77843bd857d3718d439e264b013eb566fc3afd32/grails-core/src/main/groovy/grails/boot/GrailsApp.groovy#L99-L107

At this point the application context is already created.

### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. create a grails plugin
2. override a default grails property (e.g. grails.databinding.dateFormats) inside the plugin's grails-app/config/plugin.yml file with a non-default value of something like "dd.MM.yyyy"
3. create a grails application
4. add plugin dependency to app
5. try databinding of a "LocalDate" property with a value of "19.03.2021"

### Expected Behaviour

Configuration properties from plugins should be loaded already in the Micronaut context.

### Actual Behaviour

Configuration is loaded when Micronaut context is already setup.

### Environment Information

- **Operating System**: Windows 10
- **Grails Version:** 4.0.9
- **JDK Version:** 11

davidkron:
So as I am currently trying to implement a workaround to make our app work, I experimented with the following extension to `MicronautEnvironment`, so the Spring context also receives the property sources from Micronaut:

```
@Replaces(MicronautEnvironment.class)
@Singleton
@Primary
@CompileStatic
class ExtendedMicronautEnvironment extends MicronautEnvironment {

    private static final Set<String> EXCLUDED_SOURCES_NAMES = Set.of(
        SystemPropertiesPropertySource.NAME,
        EnvironmentPropertySource.NAME,
[...]

PR #15677, opened 2026-05-22
feat(deps): Bump Micronaut platform to 5.0.0 (GA)

jamesfredley:
**Draft - stacked on `deps/spring-boot-4.0.6` (#15676)** so the Spring Boot 4.0.6 dependency fallout does not muddy this one.

Bumps `micronautPlatformVersion` from `5.0.0-M2` to `5.0.0` (GA).

Release notes: https://github.com/micronaut-projects/micronaut-platform/releases/tag/v5.0.0

## The core problem

Micronaut 5.0.0 GA publishes JARs targeting JVM 25 bytecode and declares `org.gradle.jvm.version=25`. Grails 8 keeps its baseline at JDK 21. A sub-25 JDK cannot resolve or compile the Micronaut platform, so the Grails-Micronaut "island" - `grails-micronaut`, `grails-micronaut-bom`, and the micronaut-tied `grails-test-examples-*` - cannot be part of a JDK 21 build.

### 1. Platform bump

- `gradle.properties`: `micronautPlatformVersion` `5.0.0-M2` -> `5.0.0`.
- `dependencies.gradle`: bump the `grails-micronaut-bom` Groovy override to `5.0.6` to match Micronaut 5 GA's managed `groovy.version`. Spock stays at `2.4-groovy-5.0`.
- `grails-bom/micronaut/build.gradle`: the `tools.jackson` exclude stays so `spring-boot-dependencies` remains the single source of truth for Jackson 3 (Micronaut 5 GA ships jackson-bom 3.1.3, Spring Boot 4.0.6 ships 3.1.2).

### 2. The island is toggled automatically by the build JDK

`settings.gradle` now decides island membership from the JDK running the build:

- **sub-25 JDK -> island auto-excluded.** A plain `./gradlew build` works on the Grails 8 baseline (JDK 21) with no flags.
- **JDK 25+ -> island auto-included.** The same command transparently builds the Micronaut island too.

Two presence-based overrides always win over the auto-detection:

- `-PskipMicronautProjects` - force-exclude on any JDK.
- `-PincludeMicronautProjects` - force-include on a sub-25 JDK (escape hatch; the island still cannot compile there).
- `-PskipMicronautProjects` wins if both are passed.

### 3. CI (`gradle.yml`)

- Java 21 jobs build/test everything except the island; Java 25 jobs build the full graph. This now follows purely from the JDK each job pins via `setup-java` - the old `matrix.java == 21 && -PskipMicronautProjects` conditionals are gone.
- Snapshot publish is split: the `publish` job (JDK 21) publishes everything except the island; a `publishMicronaut` job (JDK 25) publishes `grails-micronaut` + `grails-micronaut-bom`.

### 4. Apache release flow (`release.yml`) + reproducibility

ASF releases must be byte-for-byte reproducible against a pinned JDK. The flow keeps `JAVA_VERSION=21.0.7` for everything outside the island and adds a second pin `JAVA_VERSION_MICRONAUT=25.0.3` for the island:

- JDK 21 staging-publish steps build everything except the island (auto-pruned on JDK 21); a dedicated step switches to JDK 25 to sign + stage the two island artifacts against the same staging repo, then restores JDK 21 for the close + checksum-combine steps.
- `etc/bin/Dockerfile`: the verification container installs a second Liberica JDK 25 (pinned version + multi-arch SHA-512) exposed as `$JDK_25_HOME`; JDK 21 stays the default.
- `etc/bin/verify-reproducible.sh` + `etc/bin/test-reproducible-builds.sh`: each runs a JDK 21 pass for everything outside the island, then a `JAVA_HOME=$JDK_25_HOME` pass for the two island artifacts. Both fail fast when `$JDK_25_HOME` is unset.
- `RELEASE.md` documents the dual-JDK requirement.

The explicit `-PskipMicronautProjects` flag is retained only where it is **not** a JDK proxy:

- `groovy-joint-workflow.yml` - the island's Groovy 5 / Spock 2.4-groovy-5.0 pin clashes with the Groovy 4.x snapshot that build swaps in, a reason independent of the JDK (#15613).
- the standalone verify scripts - their first pass must exclude the island regardless of the ambient default JDK on `PATH`.

### 5. Micronaut HTTP client removed from `grails-data-graphql` tests

`GraphQLSpec` previously pulled `micronaut-rxjava2-http-client` + `micronaut-serde-jackson` just to call the running app from integration tests. Micronaut 5 drops RxJava 2 from the BOM, so it now uses Spring Boot 4's `RestClient` with Groovy `JsonOutput` / `JsonSlurper` (no Jackson or Gson on the test classpath). `gradle.properties` loses the now-unused `micronautRxjava2Version` and `micronautSerdeJacksonVersion` pins. Nothing in `grails-core` now depends on Micronaut artifacts outside the island.

## Consumer impact

The published `grails-micronaut:8.0.0` artifact carries `org.gradle.jvm.version=25`. Grails 8 apps on JDK 21 will have Gradle variant resolution reject it - it is consumable only on JDK 25+. JDK 21 users get the rest of Grails 8 without the Micronaut integration. The release ships both audiences; choosing the right artifact set is on the consumer's build config.

## Key versions in Micronaut 5.0.0 GA

| Property | Version |
|---|---|
| `micronaut.core.version` | `5.0.0` |
| `micronaut.data.version` | `5.0.1` |
| `micronaut.serde.version` | `3.0.0` |
| `jackson.version` | `3.1.3` |
| `groovy.version` | `5.0.6` |
| `spring.boot.version` | `4.0.6` |
| `spring.version` | `7.0.7` |

Platform-level breaking changes in v5.0.0 that are landmines for downstream Grails-on-Micronaut apps (not consumed by `grails-core` itself): RxJava 2 and MicroStream removed from the BOM, `jaxrs.api.version` 3 -> 4, `kafka.version` 3.9 -> 4.2.

`grails-forge` stays on `micronautVersion=4.10.10` (the Forge is itself a Micronaut 4.x app) - out of scope for this platform bump.

## Verification

```
# JDK 21 (Grails baseline): builds the full graph minus the island, no flags needed
./gradlew validateDependencyVersions compileGroovy compileTestGroovy

# JDK 25: builds the full graph including the island
./gradlew build
```

Project-graph toggle verified on JDK 21 via `./gradlew projects`: island absent by default, present with `-PincludeMicronautProjects`, absent with both override flags (skip wins).


jamesfredley:
## Micronaut-leak audit (post-`228f9a52ce`)

Did a full grep for `io.micronaut` across every `build.gradle` in the main `grails-core` multi-project to confirm nothing JDK-25 lives outside the `-PskipMicronautProjects` gating. Clean result.

### Every `build.gradle` referencing `io.micronaut` in the main build

| `build.gradle` | Project | In the island gating? |
|---|---|---|
| `grails-bom/micronaut/build.gradle` | `:grails-micronaut-bom` | ✅ |
| `grails-micronaut/build.gradle` | `:grails-micronaut` | ✅ |
| `grails-test-examples/micronaut/build.gradle` | `:grails-test-examples-micronaut` | ✅ |
| `grails-test-examples/plugins/issue-11767/build.gradle` | `:grails-test-examples-plugins-issue-11767` | ✅ |
| `grails-test-examples/plugins/micronaut-singleton/build.gradle` | `:grails-test-examples-plugins-micronaut-singleton` | ✅ |

Plus the two transitive consumers that depend on `:grails-micronaut` without naming `io.micronaut` directly:

- `:grails-test-examples-issue-11767` (`grails-test-examples/issue-11767/`)
- `:grails-test-examples-micronaut-groovy-only` (`grails-test-examples/micronaut-groovy-only/`)

All seven live inside the `if (!skipMicronautProjects) { ... }` block at `settings.gradle:511-520`.

### Heads up: island is 7 projects, not 5

I'd been referring to "5 island projects" in earlier commits and the PR description. The correct count from `settings.gradle` is **7 build-graph projects** in the island (the two missing in my count were `:grails-test-examples-issue-11767` and `:grails-test-examples-plugins-issue-11767`). The two `issue-11767` projects exist to repro `apache/grails-core#11767` and depend on `:grails-micronaut` + `:grails-micronaut-bom`, so they correctly belong in the island gating.

The PUBLISHED subset is still just 2 (`grails-micronaut` plugin + `grails-micronaut-bom`), which is what `publishMicronaut` in `gradle.yml` and the new JDK 25 publish step in `release.yml` stage. The three test-examples + two issue-11767 projects build inside the island but aren't published.

### `grails-data-graphql` was the only real leak; it's fixed

`grails-data-graphql/plugin` was the only non-island main-build project with a direct `io.micronaut.*` dependency, via the rxjava2 HTTP client compileOnly in `GraphQLSpec.groovy`. Removed in `228f9a52ce` (replaced with Spring `RestClient` + Groovy `JsonSlurper`).

### `grails-forge/` is intentionally separate

The 50+ `io.micronaut.*` references in `grails-forge/*/build.gradle` are all in the **separate composite build** which is itself a Micronaut 4.10.10 app (`grails-forge/gradle.properties:27`). Out of scope - the Forge stays on Micronaut 4.x regardless of what the main grails-core multi-project does with Micronaut 5.

### Net

Nothing in `grails-core` outside the explicitly-gated 7-project Micronaut island pulls Micronaut 5 / JDK 25 onto the JDK 21 classpath. The Forge composite is its own Micronaut 4.x universe. Audit clean.

jdaugherty:
Is this supposed to still be draft?

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
```diff
@@ -56,59 +65,90 @@ trait GraphQLSpec {
...
     static class GraphQLRequestHelper {
 
-        RxHttpClient rest
+        private static final MediaType APPLICATION_GRAPHQL = MediaType.parseMediaType('application/graphql')
+        private static final JsonSlurper SLURPER = new JsonSlurper()
 
-        HttpResponse<Map> graphql(String requestBody) {
-            rest.exchange(HttpRequest.POST('/graphql', requestBody).contentType('application/graphql'), Map)
-                    .firstOrError().blockingGet()
+        RestClient rest
```
jdaugherty, line 62: I think this is a fine change, but the object mapper will determine the default behavior here and it could be problematic.  We should update the grails-docs to warn users so they know about this.

etc/bin/verify-reproducible.sh
```diff
@@ -58,13 +58,36 @@ else
...
+# (Micronaut island skipped), grails-forge composite (transitively pulls in
+# the root build via includeBuild('..'), island skipped there too).
 cd grails-gradle
 ./gradlew publishToMavenLocal --rerun-tasks -PskipTests --no-build-cache
 cd ..
-./gradlew publishToMavenLocal --rerun-tasks -PskipTests --no-build-cache
+./gradlew publishToMavenLocal --rerun-tasks -PskipTests --no-build-cache -PskipMicronautProjects
 cd grails-forge
-./gradlew publishToMavenLocal --rerun-tasks -PskipTests --no-build-cache
+./gradlew publishToMavenLocal --rerun-tasks -PskipTests --no-build-cache -PskipMicronautProjects
```
jdaugherty: Forge doesn't have any micronaut projects?

etc/bin/verify-reproducible.sh
```diff
@@ -58,13 +58,36 @@ else
...
+# container provides ${JDK_25_HOME}; for local verification outside the
+# container, install Liberica JDK matching $JAVA_VERSION_MICRONAUT in
+# release.yml and export JDK_25_HOME before running this script.
+if [[ -z "${JDK_25_HOME:-}" ]]; then
+  echo "❌ JDK_25_HOME is not set; the Grails-Micronaut island requires a separate Liberica JDK 25 install."
+  echo "   In the verification container this is set automatically. Outside the container, install Liberica JDK"
+  echo "   matching JAVA_VERSION_MICRONAUT in .github/workflows/release.yml and export JDK_25_HOME=/path/to/jdk."
+  exit 1
+fi
+killall -e java || true
```
jdaugherty, line 85: We probably should add --no-daemon arguments to our gradle calls too

.github/workflows/release.yml
```diff
@@ -203,6 +219,40 @@ jobs:
...
+          NEXUS_PUBLISH_URL: ${{ vars.STAGING_URL }}
+          NEXUS_PUBLISH_STAGING_PROFILE_ID: ${{ secrets.STAGING_PROFILE_ID }}
+          NEXUS_PUBLISH_DESCRIPTION: '${{ env.REPO_NAME }}:${{ env.VERSION }}'
+          SIGNING_KEY: ${{ secrets.GPG_KEY_ID }}
+        run: >
+          ./gradlew
+          -x initializeSonatypeStagingRepository
+          findSonatypeStagingRepository
+          :grails-micronaut:publishToSonatype
+          :grails-micronaut-bom:publishToSonatype
```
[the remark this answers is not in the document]
jamesfredley: Good catch - fixed in 92f8cf0ad3.

The two island artifacts (`grails-micronaut`, `grails-micronaut-bom`) are now appended to the combined `CHECKSUMS.txt` / `PUBLISHED_ARTIFACTS.txt` in the "Combine checksums" and "Combine published artifacts" steps, reading the per-project files that `publishedChecksums` / `savePublishedArtifacts` emit as finalizers of the JDK 25 `publishToSonatype` step.

I deliberately did *not* add `aggregateChecksums` / `aggregatePublishedArtifacts` to the JDK 25 step as suggested. The build compiles with `options.release` and no Gradle toolchain (see `build-logic/.../CompilePlugin.groovy`), so re-running those tasks on the JDK 25 runner would invalidate and recompile the non-island projects against a JDK 25 compiler, producing different (non-reproducible) bytecode than the JDK 21 artifacts that were already signed and staged. Appending the island's already-generated checksum/artifact files avoids touching the JDK 21 outputs entirely while still making the final manifests complete.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jamesfredley: Added a "Testing and JSON Serialization" warning to the grails-data-graphql guide (`grails-data-graphql/docs/src/main/docs/guide/otherNotes.adoc`) in 92f8cf0ad3.

It documents that `GraphQLSpec` serializes request bodies and parses responses with Groovy's `JsonOutput` / `JsonSlurper` rather than the application's configured object mapper, so a customized object mapper (non-default date formats, property naming strategies, custom serializers) can render production responses differently from what the trait asserts on. The note recommends asserting against the raw response body and verifying the object mapper configuration separately when serialization details matter.

etc/bin/verify-reproducible.sh
jamesfredley: Correct - the Forge build itself has no Micronaut island projects. The flag is needed because `grails-forge/settings.gradle:75` does `includeBuild('..')`, which pulls the root grails-core build (which *does* contain the island) into the Forge composite. Project properties passed on the command line propagate into included builds, so without `-PskipMicronautProjects` the island would be configured/resolved on JDK 21 and fail (Micronaut 5 GA targets JVM 25 bytecode). The inline comment directly above this block documents the rationale. Leaving the flag in place.

etc/bin/verify-reproducible.sh
jamesfredley: Done in 92f8cf0ad3 - added `--no-daemon` to every gradle invocation in both `etc/bin/verify-reproducible.sh` and `etc/bin/test-reproducible-builds.sh`.

jamesfredley:
## Forge behavior: the `grails-micronaut` feature and Java versions

Documenting the current Forge behavior (asked about earlier - it was already in place on this branch, committed in `5f94d86f79`).

### What happens

When the `grails-micronaut` feature is selected in Grails Forge (CLI, web UI, or API) and the requested target JDK is **below 25**, app generation is **rejected** - no project is produced. The feature's `processSelectedFeatures` hook throws:

```
grails-micronaut requires JDK 25 or later (selected: JDK <NN>).
```

Selecting the feature with JDK 25+ generates the app normally.

### Where it is enforced

`org.grails.forge.feature.micronaut.GrailsMicronaut#processSelectedFeatures`:

```java
JdkVersion jdk = featureContext.getJavaVersion();
if (jdk.majorVersion() < JdkVersion.JDK_25.majorVersion()) {
    throw new IllegalArgumentException(
        getName() + " requires JDK 25 or later (selected: JDK " + jdk.majorVersion() + ").");
}
```

### Why JDK 25

`micronaut-core`'s `ScopedValues` references `java.lang.ScopedValue.CallableOp`, which only exists on JDK 25+ (JEP 506). More broadly, the Micronaut 5.0.0 GA platform bumped in this PR publishes JARs with `org.gradle.jvm.version=25` and JVM 25 bytecode. This is the same constraint that drives the consumer-side note in the PR description: the published `grails-micronaut:8.0.0` artifact carries `org.gradle.jvm.version=25`, so Grails 8 apps on JDK 21 cannot resolve it. Forge fails fast at generation time rather than letting users discover the incompatibility at build time.

### Related Forge validation

`GrailsMicronautValidator#validatePreProcessing` additionally rejects combining `grails-micronaut` with Spring Boot DevTools (see micronaut-projects/micronaut-spring#769):

```
Spring Boot DevTools are not supported with Grails Micronaut
```

### Note on test coverage

The JDK 25 guard itself currently has no dedicated negative test - existing specs that exercise `grails-micronaut` simply supply `JdkVersion.JDK_25` so they pass. If desired, I can add a Forge spec asserting that selecting `grails-micronaut` with a sub-25 JDK throws the expected `IllegalArgumentException`. Let me know and I'll include it.

jamesfredley:
Added the negative test in cb750435a9 (`GrailsMicronautSpec`). It covers both directions of the guard:

- `grails-micronaut` + JDK 25 -> the `org.apache.grails:grails-micronaut` dependency is rendered.
- `grails-micronaut` + JDK 21 -> generation is rejected with `IllegalArgumentException: grails-micronaut requires JDK 25 or later (selected: JDK 21).`

Verified locally (run with `-PskipMicronautProjects`, which the Forge composite needs on JDK 21 because of `includeBuild('..')`):

```
GrailsMicronautSpec > test grails-micronaut adds the dependency when JDK 25 is selected PASSED
GrailsMicronautSpec > test grails-micronaut is rejected when the selected JDK is below 25 PASSED
BUILD SUCCESSFUL
```

jamesfredley:
### Local build now keys off the build JDK

Pushed a change to `settings.gradle` so the Grails-Micronaut island is selected by whatever JDK runs the build, instead of having to remember a flag:

- **`./gradlew build` on JDK 21 (the Grails 8 baseline) just works.** The island is auto-excluded because Micronaut 5 GA bytecode targets JVM 25 and a sub-25 JDK cannot resolve or compile it. No `-PskipMicronautProjects` needed anymore.
- **`./gradlew build` on JDK 25+ also builds Micronaut locally.** When the build JDK is 25 or newer the island is auto-included, so the same command builds the full graph - no flag, no separate task list.

Explicit flags still win over the detection:

- `-PskipMicronautProjects` force-excludes the island on **any** JDK (this is what `groovy-joint-workflow.yml` uses - that exclusion is a Groovy-version reason, not a JDK reason, so it must hold even on JDK 25).
- `-PincludeMicronautProjects` force-includes the island on a sub-25 JDK (escape hatch; it still won't compile there, but the tasks become addressable).
- If both are passed, `-PskipMicronautProjects` wins.

Because each CI job pins its JDK via `setup-java`, the now-redundant `matrix.java == 21 && -PskipMicronautProjects` conditionals and the unconditional JDK-21 skip flags were removed from `gradle.yml`, `release.yml`, and `release-publish-docs.yml` - the JDK the job runs on decides it. The explicit flag is kept only where it is not a JDK proxy: `groovy-joint-workflow.yml` (Groovy reason) and the standalone `etc/bin` verify scripts (their first pass must exclude the island regardless of the ambient default JDK).

Verified on JDK 21 with `./gradlew projects`: island absent by default, present with `-PincludeMicronautProjects`, absent with both flags.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
```diff
@@ -37,7 +38,15 @@ trait GraphQLSpec {
 
     GraphQLRequestHelper getGraphQL() {
         if (_graphql == null) {
-            _graphql = new GraphQLRequestHelper(rest: RxHttpClient.create(new URL(getServerUrl())))
+            StringHttpMessageConverter stringConverter = new StringHttpMessageConverter()
+            stringConverter.supportedMediaTypes = [MediaType.ALL]
+            _graphql = new GraphQLRequestHelper(rest: RestClient.builder()
+                    .baseUrl(getServerUrl())
+                    .messageConverters({ converters ->
+                        converters.clear()
```
jdaugherty: I'm assuming these changes were due to the upgrade?  The .clear() caught me off guard since you're creating a new helper, I assumed there isn't cached state somewhere?

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jdaugherty: I don't believe this is correct.  The client itself will encode the json and that encoding is driven by Spring's configured ObjectMapper.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
```diff
@@ -56,59 +65,90 @@ trait GraphQLSpec {
...
+                    .toEntity(String)
         }
 
-        private HttpResponse<Map> buildJsonRequest(Map<String, Object> data) {
-            rest.exchange(HttpRequest.POST('/graphql', data), Map).firstOrError().blockingGet()
+        private ResponseEntity<Map> buildJsonRequest(Map<String, Object> data) {
+            wrapJson(rest.post()
+                    .uri('/graphql')
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .body(JsonOutput.toJson(data))
```
jdaugherty: Why are you reimplementing json encoding and not using spring's built in mechanism?

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
```diff
@@ -56,59 +65,90 @@ trait GraphQLSpec {
...
         }
 
-        def <T> HttpResponse<T> graphql(String requestBody, Class<T> bodyType) {
-            rest.exchange(HttpRequest.POST('/graphql', requestBody).contentType('application/graphql'), bodyType)
-                    .firstOrError().blockingGet()
+        private ResponseEntity<String> exchangeGraphql(String requestBody) {
+            rest.post()
+                    .uri('/graphql')
+                    .contentType(APPLICATION_GRAPHQL)
+                    .body(requestBody)
```
jdaugherty, line 73: Where are the functional tests using a mock http server to test this?  I don't think this will work

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jamesfredley: Removed the converters.clear() / StringHttpMessageConverter workaround. The helper now uses RestClient's default message converters, so there is no cached or custom converter state - the Jackson converter handles JSON. Fixed in c54d35e.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jamesfredley: Done. The JSON request body is now passed as a Map and serialized by RestClient's Jackson message converter, and the response is read back with toEntity(Map). JsonOutput/JsonSlurper are gone from the POST paths. Fixed in c54d35e.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jamesfredley: You were right that it didn't work. I added a mock-HTTP-server (Ersatz) spec, GraphQLSpecSpec, covering the application/graphql POST, the JSON POST (asserting the Jackson-encoded request body), the GET helper, and Map response parsing. Writing it surfaced two real bugs: RestClient had no JSON converter because the example apps don't pull Jackson onto the classpath (now declared integrationTestRuntimeOnly), and the GET helper targeted the app root and expanded the GraphQL query braces as URI templates (now targets /graphql with values bound as URI variables). All four graphql integration suites plus the new spec pass on JDK 21. Fixed in c54d35e.

grails-data-graphql/plugin/src/main/groovy/org/grails/gorm/graphql/plugin/testing/GraphQLSpec.groovy
jamesfredley: Agreed. The trait now lets RestClient encode and decode JSON through its Jackson message converter, so the encoding is driven by the message converter as you said. I corrected the guide note accordingly - the previous 'serializes using JsonOutput/JsonSlurper' wording is gone. Fixed in c54d35e.

jdaugherty:
@jamesfredley FYI: I've been waiting to review this because the tests are all failing still.  Can you please take a look?

jamesfredley:
## Fixed the failing JDK 25 CI jobs

All JDK 25 jobs (`Build Grails-Core 25`, `Functional Tests 25`, `Mongodb / Hibernate5 / Hibernate7 Functional Tests 25`) were failing at **configuration time** with the same error:

```
Could not determine the dependencies of task ':grails-test-examples-micronaut-hibernate7:bootJar'.
> Could not resolve org.apache.groovy:groovy:{strictly 5.0.5}.
  ConflictingConstraintsException: Component is the target of multiple version constraints with conflicting requirements:
    5.0.5 - directly in 'project :grails-hibernate7-micronaut-bom' (enforcedRuntimeElements)
    5.0.6 - transitively via 'project :grails-hibernate7-micronaut-bom' (enforcedRuntimeElements)
```

### Root cause

In `dependencies.gradle`, the Micronaut-island BOM overrides live in two adjacent blocks. The Micronaut 5.0.0 GA bump set the Groovy override to `5.0.6` in the `grails-micronaut-bom` / `grails-hibernate5-micronaut-bom` block, but the `grails-hibernate7-micronaut-bom` block was left at `5.0.5`.

Because `grails-hibernate7-micronaut-bom` enforces `5.0.5` **directly** while also inheriting `5.0.6` **transitively** from `grails-micronaut-bom`, the two strict/enforced constraints conflicted and dependency resolution failed - blocking every JDK 25 job before any task could run. JDK 21 jobs were unaffected since the island is auto-excluded there.

### Fix

One-line change: align the stale override to Micronaut 5 GA's managed `groovy.version`.

```diff
  else if (project.name in (['grails-hibernate7-micronaut-bom'])) {
          ...
-                 'groovy.version'  : '5.0.5',
+                 'groovy.version'  : '5.0.6',
```

All three island BOMs now agree on `5.0.6`. `spock.version` stays `2.4-groovy-5.0` (Groovy 5 line, patch-independent).

### Verified locally on JDK 25 (Corretto 25)

- `:grails-test-examples-micronaut-hibernate7:dependencies --configuration runtimeClasspath` → now resolves to `{strictly 5.0.6} -> 5.0.6`, no conflict
- `:grails-test-examples-micronaut-hibernate7:bootJar` (the exact previously-failing task) → **BUILD SUCCESSFUL**
- `validateDependencyVersions` → **BUILD SUCCESSFUL**

jdaugherty (approved):

PR #15656, opened 2026-05-12
GORM: Shared Mapping Registry O(M+N) Scaling

borinquenkid:
## Description
This PR addresses a scalability bottleneck in the GORM datamapping layer. Previously, the system instantiated a full set of mapping objects for every tenant, resulting in O(M x N) memory complexity. By refactoring the architecture to use a stateless, shared-registry approach, we have reduced this to O(M + N), significantly lowering the memory overhead in multi-tenant environments.

### Issue and Scope
- [X] This PR addresses the **complete scope** of the linked issue.

### Code Quality
- [X] I have **added or updated tests** that cover the changes introduced in this PR.
- [X] I have verified that all existing tests pass by running `./gradlew build --rerun-tasks`.
- [X] My code follows the project's **code style** guidelines. I have run `./gradlew codeStyle` and resolved any violations.
- [X] If generative AI tooling was used in preparing this contribution, a quality model was used to ensure contributions are consistent with the project's quality standards.

### Licensing and Attribution
- [X] All contributed code is provided under the Apache License 2.0, and new source files include the appropriate Apache license header.
- [X] I have the necessary rights to submit this contribution and confirm it is my own original work.
- [X] If generative AI tooling was used in preparing this contribution, I have followed the Apache Software Foundation's policy on generative tooling and have properly attributed its use.

PR #15678, opened 2026-05-23
GORM: Shared Mapping Registry O(M+N) Scaling (clean rebuild)

borinquenkid:
## Description

This PR replaces #15656, which was rebuilt on a clean branch to remove a corruption commit from the history.

**Problem:** In multi-tenant GORM environments with many tenants (M) and domain classes (N), the previous implementation instantiated a full set of static API, instance API, and validation API objects per tenant per entity, producing O(M × N) object allocations. This caused excessive memory consumption and degraded startup performance as tenant counts scaled.

**Solution:** Refactor `GormRegistry` to use a single shared registry keyed by entity class name and qualifier (datasource/tenant), with the GORM API objects created once per entity per qualifier and reused across tenants. This reduces the allocation profile to O(M + N).

**This rebuild also fixes two regression classes exposed by the new registry:**

1. **Multi-tenancy resolution regressions** — `GormApiResolver` and `GormRegistry.registerEntityDatastores` were routing DISCRIMINATOR/SCHEMA tenant IDs through the datasource connection lookup path, causing child datastores to be overwritten by the parent and `PartitionedMultiTenancySpec.count()` to NPE. Fixed by detecting multi-tenancy mode before delegating to `getDatastoreForConnection`, and by skipping non-DEFAULT qualifier registration when the qualifier resolves back to the parent (i.e., it's a runtime tenant ID, not a datasource name).

2. **Child datastore initialization order** — `HibernateDatastore` (H5) and `ChildHibernateDatastore` (H7) were throwing `ConfigurationException` when `getDatastoreForConnection` was called for a sibling during initialization before all children were registered. Fixed to return `null` during the initialization phase so `GormRegistry` falls back gracefully and re-registers once initialization completes.

**Test infrastructure:** Added `forkEvery = 1` to `gradle/hibernate5-test-config.gradle` and `gradle/hibernate7-test-config.gradle`. The root config uses `forkEvery = 50/100` for speed, but with a shared `GormRegistry` singleton, TCK specs running in the same JVM before `PartitionedMultiTenancySpec` were clearing `datastoresByQualifier["default"]` and causing the NPE described above. Each test class now gets its own JVM.

**Verified:** H5 — 669 tests / 0 failures. H7 — 2960 tests / 0 failures.

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. *(Replaces #15656, which tracks the approved O(M+N) scaling work.)*
- [x] This PR targets the **correct branch** (`8.0.x-hibernate7` — major release branch; breaking API changes permitted).

### Code Quality

- [x] I have verified that all existing tests pass (H5: 669/0 failures, H7: 2960/0 failures).
- [x] My code follows the project's **code style** guidelines. `./gradlew codeStyle` has been run and violations resolved.
- [x] This PR does **not** include unsolicited reformatting or unrelated refactoring.
- [x] Generative AI tooling was used in preparing this contribution with a quality model, consistent with the project's quality standards.

### Licensing and Attribution

- [x] All contributed code is provided under the Apache License 2.0, and new source files include the appropriate Apache license header.
- [x] Generative AI tooling use follows the [ASF policy on generative tooling](https://www.apache.org/legal/generative-tooling.html) and is properly attributed.

### Documentation

- [x] No new user-facing APIs are introduced; this is a performance/correctness fix to the internal registry.
- [x] The PR description explains what was changed and why.


jamesfredley:
taking a look now, I use multi-tenant in a few apps

jamesfredley:
@borinquenkid I am seeing a bit more than expected in the PR including routing fixes, transaction changes, test isolation changes, Mongo/Neo4j/simple datastore edits, TCK rewrites, debug specs, and deleted cross-layer tests.  Can you catch me up on how these relate to Multi-tenancy resolution regressions and Child datastore initialization order ?

borinquenkid:
Thanks for the detailed read — happy to connect the dots.

The single root cause is commit 8f1500dd03, which replaced GormEnhancer's per-entity static maps (one (entityClass, qualifier) → GormApi entry per entity × per connection — O(M×N) memory) with a single process-wide GormRegistry that resolves APIs on demand using a selector chain. Every other change in this PR is a downstream ripple from that one architectural swap. The two threads you named map to the two hardest ripples:

---
Thread 1 — Multi-tenancy resolution regressions

The old static maps gave each entity a direct pointer to its API for a given connection at enhancement time. The registry replaces that with a runtime selector chain (ActiveSession → Preferred → Qualified → Default). Several call sites assumed the old direct-lookup contract and started resolving to the wrong (usually default) datastore for non-default connections or tenants:

- Routing fixes (DatastoreResolver, GormApiResolver, ActiveSessionDatastoreSelector, PreferredDatastoreSelector): correct the selector chain for non-default connection/tenant resolution paths that the static maps used to short-circuit.
- Transaction changes (GormStaticApi.withTransaction(Map), GrailsHibernateTransactionManager): the withTransaction(Map) overload called definition.setProperty(k, v) — no such method on DefaultTransactionDefinition (a Java bean), which was silently tolerated before but now hits a hot path under registry routing. Also fixed GrailsHibernateTransactionManager to see child-datastore sessions when participating in a parent transaction.
- Mongo/Neo4j/Simple edits: each datastore previously managed its own static enhancement maps. This PR registers each with GormRegistry (registerDatastore() / registerApiFactory()) and adds the getDatastoreForTenantId fallback to MongoDatastore (the multi-tenancy tenant resolver was calling into the registry before the child datastore was registered). SimpleMapSession had a latent rollbackOnly bug (a rolled-back tx set a session-level flag that was never cleared) that the registry's new session-per-call paths surfaced.
- Test isolation changes (DataTestSetupInterceptor / DataTestCleanupInterceptor): the harness only bound a session for the default datastore. A non-default-connection entity now resolves to a dedicated child datastore — without a bound session, save() opened a throwaway per-call session that was discarded on return. The fix binds and unbinds a session per connection source (no-op for single-datasource specs).

---
Thread 2 — Child datastore initialization order

ChildHibernateDatastore.initialize() returned null in the 8.0.x-hibernate7 base — the child was intentionally not enhanced. Under the registry, each child must register its own APIs (so multi-datasource entities resolve to the right child). The new initialize() calls new HibernateGormEnhancer(this, txManager, settings, datastoresByConnectionSource). The critical ordering constraint: datastoresByConnectionSource in the parent must be populated before the child is constructed (the parent builds its connection map during its own initialize(), then calls new ChildHibernateDatastore(parent, ...)). The bindParent() / PARENT_HOLDER thread-local in ChildHibernateDatastore handles the case where getPrimaryDatastore() is called during the super-constructor chain before this.parent is assigned.

- TCK rewrites (GrailsDataTckManager): the manager now calls setTargetDatastore(childDatastore) after the parent initializes, so the TCK spec's manager.session and service lookups go to the right child. The domain-class list was changed from Set to List because entity registration order now matters for the registry's MappingContext traversal.

---
Two items that need explicit acknowledgment:

- Debug specs (DebugGetSpec, DebugGeoJSONDecodeSpec, etc. in grails-data-mongodb): these were written to isolate get() failures during root-cause investigation. Their underlying test logic is sound, but the Debug prefix and println calls shouldn't be merged. I'll rename them to production spec names and remove the debug output before this lands.
- Deleted cross-layer tests (CrossLayerMultiDataSourceSpec, CrossLayerMultiTenantMultiDataSourceSpec): these called manager.getServiceForConnection(SomeService, 'secondary') to fetch a service bean from the child datastore's Spring context — internal wiring that the registry no longer exposes that way. Rather than patching around a removed internal, the contract is now covered in two purpose-built specs: DomainMultiDataSourceSpec (domain-API routing to secondary) and DataServiceConnectionRoutingSpec (service routing via @Service(datasource='secondary')). The net line count for multi-datasource coverage went up, not down. That said, I can explicitly document the equivalence in the spec Javadoc if that helps reviewers verify the coverage isn't lost.

---
tl;dr: The PR is wider than a typical bug-fix because the root change was architectural — every datastore (H5, H7, Mongo, Neo4j, SimpleMap), the test harness, and the TCK had local copies of the enhancement pattern that all needed updating. The scope tracks the scope of the original GormRegistry rewrite.

jamesfredley:
@borinquenkid I think something laser targeted to `O(entityCount * qualifierCount)` like https://github.com/apache/grails-core/pull/15771, will be much easier to get the required reviews and merged.

jamesfredley:
And here is a targeted stab at Child datastore initialization order: https://github.com/apache/grails-core/pull/15776

Having 456 files in this PR makes it challenging for the team to review in a reasonable timeframe.

borinquenkid:
Superseded by the restructured GormRegistry PR stack (feat/gorm-registry-core-impl, feat/gorm-registry-core-tests, and per-adapter PRs).

PR #15654, opened 2026-05-11
Hibernate 7 - Step 1

jdaugherty:
creating this request by cloning the existing hibernate five modules so that we can easily review the hibernate seven changes between hibernate five and seven

Copy all source, test examples, BOMs, and build config from the
hibernate5 namespace to hibernate7 so that the real hibernate7 PR
can be reviewed as a true delta rather than a sea of new files.

Modules added:
- grails-hibernate7-bom (copy of grails-hibernate5-bom)
- grails-data-hibernate7-core, spring-orm, grails-plugin, dbmigration, spring-boot, docs
- grails-test-examples/hibernate7 (12 projects mirroring hibernate5)
- gradle/hibernate7-test-config.gradle (skipHibernate7Tests flag)

Build infrastructure:
- publish-root-config.gradle: register hibernate7 modules for publishing
- SbomPlugin.groovy: add LGPL exemptions for hibernate5 artifacts used by hibernate7 staging modules
- settings.gradle: include all hibernate7 projects

PR #15683, opened 2026-05-27
Step 1 (PR-C): Add addAllDomainClasses helper to GrailsDataTckManager

jamesfredley:
**Step 1 (PR-C) prerequisite for Hibernate 7 work** - extracted from the staging branch so it can be reviewed on its own merits.

## Context

This change was originally pulled forward into the hibernate7 staging branch as commit `ed0916dd93` on PR #15654. Reviewers (@matrei, @sbglasius) correctly pointed out it is unrelated to the Hibernate 7 clone and should be a standalone PR. Extracting it here against `8.0.x` so it can land on its own and then flow naturally into both the staging branch (PR #15654) and Step 2 (PR #15568).

## Scope

Adds a small helper method to [`GrailsDataTckManager`](grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy) and mechanically replaces all callers:

```groovy
// before
manager.domainClasses.addAll([Foo, Bar])

// after
manager.addAllDomainClasses([Foo, Bar])
```

The helper hides the implementation detail that `domainClasses` is a mutable list field, and reads more naturally at the call site.

### Files changed

182 files total - the method definition + 181 call sites:

| Location | Count |
|----------|-------|
| `grails-datamapping-tck/src/main/...` | 13 specs + 1 manager (with new method) |
| `grails-datamapping-core-test/src/test/...` | 49 core test specs |
| `grails-data-hibernate5/core/src/test/...` | 27 hibernate5 specs |
| `grails-data-mongodb/core/src/test/...` | 92 mongo specs |

The change is purely mechanical - I verified a sample of files diff identically to commit `ed0916dd93` on the staging branch.

## Why a separate PR

PR #15654 (Step 1) is meant to be a near-pure clone of `hibernate5` → `hibernate7`. PR #15568 (Step 2) is the actual Hibernate 7 logic. Both reviewers asked for cleanup of this nature to be split out so the review effort on Step 1 and Step 2 can stay focused on hibernate-related diffs.

Once this PR is merged, the corresponding "Pull forward manager.addAllDomainClasses change" commit on the hibernate7 staging branch should be removed since the change will arrive through the next merge of `8.0.x`.

## Related

- Step 1 PR-A: #15654 (the hibernate7 clone, already updated to remove `logback.groovy` and the unrelated `stepByStep.adoc` change)
- Step 1 PR-B: #15682 (async defensive coding cleanup - in review)
- Step 1 PR-D: DetachedCriteriaSpec TCK style cleanup (forthcoming)
- Step 2: #15568 (the actual Hibernate 7 logic - blocked on Step 1 prerequisites)

## Note on MongoDatastoreSpec base class

The companion change on the staging branch (commit `c8cb43f2ec` - introducing `MongoDatastoreSpec` as a thin base class to reduce generic noise across 108 mongo specs) was **not** included in this PR because it depends on a separate `tests/` → `specs/` package rename that hasn't been applied to `8.0.x`. That refactor can be a follow-up PR or stay in the staging branch since @sbglasius's review comment on it was conditional ("if needed, ok").


matrei (commented):
I am against this change in its current form.
It changes/adds nothing of value and you can still access `manager.domainClasses.addAll()` or `manager.domainClasses += [Book]`.

For reference, see `HibernateValidationSpec`.

borinquenkid:
We need to update the code to make the domain classes list immutable. Domain binding should only be done once in a lifecycle of an application.

sbglasius (approved):
Clear and easy to review. If it has value for the downstream implementation of Hibernate 7, go for it.

jdaugherty:
I believe the original point of this change was to ensure a consistent list of classes (centralize into one location), although we didn't make the base list access private.  I think that would resolve the concern @matrei has?

jamesfredley:
@borinquenkid 

> We need to update the code to make the domain classes list immutable. Domain binding should only be done once in a lifecycle of an application.

Can that be done in this PR to demonstrate the need for these changes?

jdaugherty:
Take a look at GrailsDataTckManager in the end branch.   The following changes would also need to be done as part of this: 

            private List<Class> domainClasses = [
            ]

            /**
             * Returns an unmodifiable view of the domain classes list.
             * @return An unmodifiable list of domain classes
             */
            List<Class> getDomainClasses() {
                return Collections.unmodifiableList(domainClasses)
            }

            /**
             * Adds all the specified classes to the domain classes list.
             * @param classes The classes to add
             */
            void addAllDomainClasses(Collection<Class> classes) {
                domainClasses.addAll(classes)
            }


There is also a lot more missing from this PR - all of the tests should add the domains they want at the start in setupSpec().  The point of this change was to have 1 spot where we pick the domains that are required for tests.

borinquenkid:
> ake a look at GrailsDataTckManager in the end branch. The following changes would also need to be done as part of this:
> 
> ```
>         private List<Class> domainClasses = [
>         ]
> 
>         /**
>          * Returns an unmodifiable view of the domain classes list.
>          * @return An unmodifiable list of domain classes
>          */
>         List<Class> getDomainClasses() {
>             return Collections.unmodifiableList(domainClasses)
>         }
> 
>         /**
>          * Adds all the specified classes to the domain classes list.
>          * @param classes The classes to add
>          */
>         void addAllDomainClasses(Collection<Class> classes) {
>             domainClasses.addAll(classes)
>         }
> ```
> 
> There is also a lot more missing from this PR - all of the tests should add the domains they want at the start in setupSpec(). The point of this change was to have 1 spot where we pick the domains that are required for tests.

@jdaugherty Done

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +56,25 @@ abstract class GrailsDataTckManager {
...
-            Person,
-            PersonEvent,
-            Pet,
-            PetType,
-            Plant,
-            PlantCategory,
-            Publication,
-            Task,
-            TestEntity
+    private List<Class> domainClasses = [
```
matrei: Would it not be better to store this or at least return a `Class[]` from `getDomainClasses()` as that is what is consumed downstream to not have to do a cast?

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +56,25 @@ abstract class GrailsDataTckManager {
...
+     */
+    List<Class> getDomainClasses() {
+        return Collections.unmodifiableList(domainClasses)
+    }
+
+    /**
+     * Adds all the specified classes to the domain classes list.
+     * @param classes The classes to add
+     */
+    void addAllDomainClasses(Collection<Class> classes) {
```
matrei: Could we name this method `addDomainClasses` or `registerDomainClasses`?
Could we make the parameter a vararg: `void addDomainClasses(Class... domainClasses)`?

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid: Done

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid: I added your preferred name and deprecated the old one . Done

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
+/**
+ * Returns a defensive copy of the registered domain classes.
+ * Mutating this array will not affect the manager state.
+ */
+    Class[] getDomainClasses() {
+        return domainClasses as Class[]
+    }
+
+    @Deprecated
+    void addAllDomainClasses(Collection<Class> classes) {
```
matrei: Do we need to deprecate a method that was never added?

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
-            Plant,
-            PlantCategory,
-            Publication,
-            Task,
-            TestEntity
+    private List<Class> domainClasses = [
     ]
 
+/**
+ * Returns a defensive copy of the registered domain classes.
```
matrei: Javadoc has wrong indentation.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
+
+    @Deprecated
+    void addAllDomainClasses(Collection<Class> classes) {
+        if (classes) {
+            domainClasses.addAll(classes)
+        }
+    }
+
+    /**
+     * Adds all the specified classes to the domain classes list.
```
matrei: Suggestion: `Registers the domain classes that will be available when testing.` We can drop the "Supports clean vararg..." line.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
-            TestEntity
+    private List<Class> domainClasses = [
     ]
 
+/**
+ * Returns a defensive copy of the registered domain classes.
+ * Mutating this array will not affect the manager state.
+ */
+    Class[] getDomainClasses() {
+        return domainClasses as Class[]
```
matrei: Remove `return` keyword?

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
+        }
+    }
+
+    /**
+     * Adds all the specified classes to the domain classes list.
+     * Supports clean vararg syntax: manager.registerDomainClasses(Foo, Bar)
+     *
+     * @param classes The classes to register
+     */
+    void registerDomainClasses(Class... classes) {
```
matrei, line 49: This method is currently unused. Why are we not using it?

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
```diff
@@ -56,34 +31,36 @@ abstract class GrailsDataTckManager {
...
-            Person,
-            PersonEvent,
-            Pet,
-            PetType,
-            Plant,
-            PlantCategory,
-            Publication,
-            Task,
-            TestEntity
+    private List<Class> domainClasses = [
```
matrei: Should this be a `Set` to prevent duplicates?

borinquenkid: Yes

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid: If we rename  addAllDomainClasses  in this repository, downstream callers will not automatically update, and it will cause compile-time failures for several reasons:                                  
                                                                                                                                                                                                          
  1. External Repository Breakages: GORM implementations and plugins that reside in external repositories (such as Neo4j, Cassandra, or custom third-party datastores) depend on  grails-datamapping-tck  
  as a published library. If  addAllDomainClasses  is renamed or deleted, their builds will break immediately upon upgrading because they still call the old method.                                      
  2. IntelliJ Groovy Refactoring Limits: Even within the same monorepo, IntelliJ's "Rename Refactoring" tool is notoriously unreliable for Groovy. Because of Groovy's dynamic nature, the IDE frequently 
  misses dynamic call sites or uses of the method in dynamic test specs, requiring a lot of manual search-and-replace anyway.                                                                             
                                                                                                                                                                                                          
  By keeping  addAllDomainClasses  as a  @Deprecated  wrapper delegation to  registerDomainClasses , we preserve backward compatibility for downstream/external TCK consumers while indicating that they  
  should move to the new API.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid: It is used everywhere now.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid:
```groovy
    /**
     * Returns an unmodifiable view of the domain classes list.
     * @return An unmodifiable list of domain classes
     */
    List<Class> getDomainClasses() {
        Collections.unmodifiableList(domainClasses)
    }
    ```
    It has to be a defensive copy

File: grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy (hunk 7 above)
Comment (matrei) [reply grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy:49]: @borinquenkid 
1. How can external repositories break by removing a method that was added in this PR, and thus does not exist outside this PR?
2. If that is the case, was it not same when changing from `manager.domainClasses.addAll()` to `manager.addAllDomainClasses()`?

File: grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy (hunk 6 above)
Comment (matrei) [reply grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy]: I don't understand this comment. It seems to show an old version of the `getDomainClasses()` method.

**Comment by jdaugherty** [comment]
@borinquenkid why are we force pushing to these changes?  It makes it harder to review them.  Can you let us know what was changed?

File: grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy (hunk 6 above)
Comment (borinquenkid) [reply grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy]: I dont understand why Github behaves like it does, but that is the current state.

File: grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy (hunk 7 above)
Comment (borinquenkid) [reply grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy:49]: Not external repositoriess you are right, we would have to change rename the methods over 250 times and IDE does not do that well

**Comment by borinquenkid** [comment]
> @borinquenkid why are we force pushing to these changes? It makes it harder to review them. Can you let us know what was changed?

@jdaugherty     Essentially   
``` java
manager.domainClasses.addAll` --> manager.addAllDomainClasses
```
over 250 times

I forced pushed because when one does a change that break 100s of tests on a private branch and you  ar enot sure which change made it break and you know that before the change everyhting was working and it is a private branch you reset hard and add them one at a time until you find the culprit and then you force push.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
matrei: > Not external repositoriess you are right, we would have to change rename the methods over 250 times and IDE does not do that well

This is not a problem, I'd be happy to do it.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
jdaugherty: This happens when there is a three way merge.  That's why you want to create pull requests that have ancestors instead of multiple sources.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
jdaugherty: We should just rename the method.

grails-data-mongodb/core/src/test/groovy/org/apache/grails/data/mongo/core/GrailsDataMongoTckManager.groovy
```diff
@@ -90,7 +90,7 @@ class GrailsDataMongoTckManager extends GrailsDataTckManager {
 
     @Override
     Session createSession() {
-        def allClasses = getDomainClasses() as Class[]
+        def allClasses = getDomainClasses()
```
jdaugherty: Since this is groovy, you can use the groovy property reference syntax: `domainClasses`

grails-datamapping-core-test/src/test/groovy/grails/gorm/tests/WhereMethodSpec.groovy
```diff
@@ -90,7 +90,7 @@ class Project {
     }
 }
 ''')
-        manager.domainClasses.addAll(list)
+        manager.addAllDomainClasses(list)
```
jdaugherty: This still looks incorrect.

jdaugherty (approved):
I think there are some usages of the old method still, once fixed I think we can merge if @matrei agrees.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/base/GrailsDataTckManager.groovy
borinquenkid: I used AI to do it , change the name and change the args and change all the callers

grails-data-mongodb/core/src/test/groovy/org/apache/grails/data/mongo/core/GrailsDataMongoTckManager.groovy
borinquenkid: done

grails-datamapping-core-test/src/test/groovy/grails/gorm/tests/WhereMethodSpec.groovy
borinquenkid: fixed

grails-data-mongodb/core/src/test/groovy/org/apache/grails/data/mongo/core/GrailsDataMongoTckManager.groovy
```diff
@@ -126,10 +125,10 @@ class GrailsDataMongoTckManager extends GrailsDataTckManager {
                 return null
             }
         })
-        mappingContext.addPersistentEntities(allClasses as Class[])
+        mappingContext.addPersistentEntities(domainClasses as Class[])
```
jdaugherty: I think you still want to store a reference to the classes prior to the creation of the context.

borinquenkid: this is really a codesmell from subclassing breaking encapsulation. the correct structure passed around should be an immutable collection. But I really want to get out of this branch, the amount of churn here and what we got out of it is too big IMHO.

matrei (approved):
We don't need to cast the result from `getDomainClasses()` to `Class[]` anymore as that is what is returned.

Thank you @borinquenkid!

PR #15684, opened 2026-05-27
Step 1 (PR-D): DetachedCriteriaSpec command-chain syntax cleanup

jamesfredley:
**Step 1 (PR-D) prerequisite for Hibernate 7 work** - extracted from the staging branch so it can be reviewed on its own merits.

## Context

This change was originally pulled forward into the hibernate7 staging branch as part of PR #15654. @sbglasius flagged it on [`DetachedCriteriaSpec.groovy`](grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/tests/DetachedCriteriaSpec.groovy):

> "This is an unneeded change for this PR. Removing parentheses should have been done in a clean-up PR."

Extracting it here against `8.0.x` so it can land on its own and then flow naturally into both the staging branch (PR #15654) and Step 2 (PR #15568).

## Scope

Single file. 18 line changes. Pure style cleanup:

```groovy
// before
criteria.with {
    eq('lastName', 'Simpson')
}

// after
criteria.with {
    eq 'lastName', 'Simpson'
}
```

| Call | Count |
|------|-------|
| `eq('lastName', 'Simpson')` → `eq 'lastName', 'Simpson'` | 15 |
| `like('firstName', 'B%')` → `like 'firstName', 'B%'` | 3 |
| **Total** | **18** |

This matches the idiomatic Groovy DSL convention used elsewhere in GORM criteria builders and is the form adopted on the hibernate7 staging branch.

## Why a separate PR

PR #15654 (Step 1) is meant to be a near-pure clone of `hibernate5` → `hibernate7`. PR #15568 (Step 2) is the actual Hibernate 7 logic. @sbglasius's review of the staging PR asked for cleanup of this nature to be split out so the review effort on Step 1 and Step 2 can stay focused on hibernate-related diffs.

Once this PR is merged, the corresponding style change on the hibernate7 staging branch should be removed since the change will arrive through the next merge of `8.0.x`.

## Related

- Step 1 PR-A: #15654 (the hibernate7 clone, already updated to remove `logback.groovy` and the unrelated `stepByStep.adoc` change)
- Step 1 PR-B: #15682 (async defensive coding cleanup - in review)
- Step 1 PR-C: #15683 (addAllDomainClasses helper - in review)
- Step 2: #15568 (the actual Hibernate 7 logic - blocked on Step 1 prerequisites)


matrei (commented):
I am against this change and I think we should always use parentheses for method calls as it is much easier to read/skim.

sbglasius (commented):
I agree with @matrei - removing parentheses does not add value.

jamesfredley:
I agree with standardizing on parentheses for method calls.  It is the majority, but separate from this we will need to create a issue to go back to cleanup.  

on `origin/8.0.x` (the PR's actual base = historical truth)

Combined `eq` + `like`:

| Scope | paren | no-paren | split |
|-------|------:|------:|:-----:|
| All `.groovy` | **109** (92 + 17) | 97 (77 + 20) | ~53 / 47 |
| `*Spec.groovy` only | **94** (79 + 15) | 78 (69 + 9) | ~55 / 45 |

Per-method breakdown:

| Method / scope | paren | no-paren |
|---|------:|------:|
| `eq` - all groovy | **92** | 77 |
| `eq` - specs only | **79** | 69 |
| `like` - all groovy | 17 | **20** |
| `like` - specs only | **15** | 9 |

## Staging vs base (specs only)

| Branch | paren | no-paren |
|---|------:|------:|
| `8.0.x` (base, historical truth) | **94** | 78 |
| `8.0.x-stage-hibernate7` (staging) | 95 | 88 |

The staging branch added ~10 no-paren spec lines on top of base. The migration toward no-paren is something the hibernate7 staging work itself introduced, not a pre-existing house style.


## Verdict
- Historically on `8.0.x`, the **parenthesized form is the plurality** - roughly 55/45 in specs, 53/47 overall.
- `eq(...)` specifically leads in every slice (92 vs 77 overall, 79 vs 69 in specs).
- The only place no-paren wins is `like` across all groovy (20 vs 17), and that flips back to paren when restricted to specs (15 vs 9).
- The `HibernateCriteriaBuilder` javadoc examples and most example/integration-test apps use the parenthesized form.

jdaugherty:
We never previously formatted tests, and I think that's what's made this merge so difficult.  I actually agree with @matrei here, so if we wanted to add a commit reverting this explicit instance, I'm fine with that.  We would still need to merge this so we correctly undo in the later branches though.

sbglasius:
I can be convinced to approve, given a ticket is made to revert it back to using parentheses at a later date. This in the name of getting Hibernate 7 merged asap.

jamesfredley:
I created https://github.com/apache/grails-core/issues/15701 to later standardize on parentheses for method calls in tests, closing this PR since it is the opposite.

PR #15686, opened 2026-05-28
Step 1 (PR-G): Add grails-code-analysis convention plugin + violation aggregation

jamesfredley:
## Step 1 (PR-G): grails-code-analysis convention plugin + violation aggregation

Carved out of the Hibernate 7 Step 1 PR (#15654) per review feedback from @matrei that the static-analysis tooling should be introduced in its own focused PR.

### What this adds

- **`grails-code-analysis`** convention plugin (PMD + SpotBugs) applied to 96 subprojects, with `GrailsCodeAnalysisExtension`.
- **`grails-violation-aggregation`** root-only plugin that consolidates per-module Checkstyle/CodeNarc/PMD XML reports into Markdown summaries under `build/reports/violations/` via an `aggregateStyleViolations` task.
- **`grails-code-style`** plugin improvements split cleanly from analysis: enable/disable properties, a `codenarcFix` task, a test-styling toggle, and a resource-path rename to `grails-code-style` (PMD config under `grails-code-analysis`).
- `codeanalysis.yml` workflow; `codestyle.yml` switched to the aggregated report format.
- `pmdVersion` / `spotbugsPluginVersion` properties and the `violation-fixer` skill.

### Notes

- JaCoCo coverage is intentionally **not** in this PR. It layers on top of the `grails-violation-aggregation` plugin introduced here and is carved out separately (PR-H).
- Verified with `cd build-logic && ../gradlew :build-logic:test` (plugin Spock specs pass).

### Why

Splitting this from #15654 keeps the Hibernate 7 baseline clone focused on the actual h5 -> h7 work and lets the project review/decide on the analysis tooling independently.

Assisted-by: claude-code:claude-4.7-opus


build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -90,72 +100,65 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
...
-        if (!Files.exists(expectedPath) || expectedPath.size() == 0) {
+    private static void createOrLoad(Path expectedPath, String defaultResource, Project project) {
+        boolean defaultPath = expectedPath.startsWith(project.rootProject.buildDir.toPath())
+        if (!Files.exists(expectedPath) || expectedPath.size() == 0 || defaultPath) {
             def defaultValue = GrailsCodeStylePlugin.getResourceAsStream(defaultResource)
             if (!defaultValue) {
                 throw new IllegalStateException("Could not locate default configuration file: ${defaultResource}")
             }
+            // TODO: This really need to use gradle caching instead
+            project.logger.info("Replacing code style configuration")
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. createOrLoad now only rewrites the config when the target is missing or its on-disk content differs from the bundled resource, and logs at debug level, so repeated builds across subprojects no longer churn the file or spam the log.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,467 @@
...
+            }
+        }
+    }
+
+    @CompileDynamic
+    private static void parseStyleViolations(Directory styleXmlDir, Directory violationsDir,
+            boolean checkStyleTests, boolean codenarcEnabled, boolean checkstyleEnabled) {
+        def slurper = new XmlSlurper()
+        slurper.setFeature('http://apache.org/xml/features/nonvalidating/load-external-dtd', false)
+        slurper.setFeature('http://xml.org/sax/features/namespaces', false)
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. The XmlSlurper now disallows DOCTYPE declarations and disables external general and parameter entities (in addition to the existing external-DTD and namespace settings) before parsing any report XML.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,467 @@
...
+        }
+        writeReport('CHECKSTYLE_VIOLATIONS.md', checkstyleViolations, 'Checkstyle Violations Summary')
+    }
+
+    @CompileDynamic
+    private static void parseAnalysisViolations(Directory analysisXmlDir, Directory violationsDir,
+            boolean checkAnalysisTests, boolean pmdEnabled, boolean spotbugsEnabled) {
+        def slurper = new XmlSlurper()
+        slurper.setFeature('http://apache.org/xml/features/nonvalidating/load-external-dtd', false)
+        slurper.setFeature('http://xml.org/sax/features/namespaces', false)
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. The XmlSlurper now disallows DOCTYPE declarations and disables external general and parameter entities (in addition to the existing external-DTD and namespace settings) before parsing any report XML.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,467 @@
...
+            }
+        }
+
+        if (jacocoCoverage.isEmpty()) {
+            LOGGER.info('No JaCoCo coverage reports found to aggregate')
+            return
+        }
+
+        jacocoCoverage.removeIf { it.className.startsWith('org.grails.orm.hibernate.support.hibernate7.') }
+
```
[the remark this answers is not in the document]
jamesfredley: This filter is intentional. The hibernate5 and hibernate7 variants compile classes with identical fully-qualified names, and JaCoCo cannot aggregate coverage for two different classes that share a name (it fails with 'Can't add different class with same name'). Dropping the hibernate7 support classes keeps the aggregate valid. I agree it should not be hard-coded in a generic plugin long-term and will make it configurable via an extension property in a follow-up, so leaving this thread open to track that.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStyleExtension.groovy
```diff
@@ -25,18 +25,19 @@ import groovy.transform.CompileStatic
 import org.gradle.api.Project
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1, removed the unused org.gradle.api.provider.Property import.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisExtension.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+import org.gradle.api.Project
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.model.ObjectFactory
+
+@CompileStatic
+class GrailsCodeAnalysisExtension {
+
+    /**
+     * Defaults to project.rootProject.buildDir/codestyle/pmd.
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1, the javadoc now reads build/codeanalysis/pmd to match the actual convention.

.agents/skills/violation-fixer/SKILL.md
```diff
@@ -0,0 +1,231 @@
...
+- Preparing a commit — the plugin output must be clean before merging.
+
+---
+
+## Plugin Overview
+
+| Plugin | Applied to | Responsibility |
+|--------|-----------|----------------|
+| `org.apache.grails.gradle.grails-code-style` | Every subproject | Applies Checkstyle and CodeNarc; registers per-project `codeStyle` task; redirects XML reports to root `build/reports/codestyle/` |
+| `org.apache.grails.gradle.grails-code-analysis` | Every subproject | Applies PMD and SpotBugs (both opt-in); registers per-project `codeAnalysis` task; redirects XML reports to root `build/reports/codestyle/` |
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1, the grails-code-analysis row now points to build/reports/codeanalysis/.

.agents/skills/violation-fixer/SKILL.md
```diff
@@ -0,0 +1,231 @@
...
+├── checkstyle/
+│   ├── grails-core-checkstyleMain.xml
+│   ├── grails-web-mvc-checkstyleMain.xml
+│   └── ...
+├── codenarc/
+│   ├── grails-core-codenarcMain.xml
+│   └── ...
+├── pmd/       (if enabled)
+└── spotbugs/  (if enabled)
+
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1, the directory layout now shows the separate build/reports/codeanalysis/{pmd,spotbugs} inputs.

.github/workflows/codestyle.yml
```diff
@@ -117,24 +110,20 @@ jobs:
         uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
         with:
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
-      - name: "🔎 Check Forge Projects"
-        working-directory: grails-forge
-        run: ./gradlew codeStyle
-      - name: "📤 Upload Failure Reports"
+      - name: "🔎 Check Gradle Plugin Projects"
+        working-directory: grails-gradle
+        run: ./gradlew aggregateStyleViolations --continue
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. grails-gradle/build.gradle now applies grails-violation-aggregation, and codeanalysis.yml passes -Pgrails.codeanalysis.enabled.pmd=true -Pgrails.codeanalysis.enabled.spotbugs=true so the analysis tasks both exist and actually run in that build.

.github/workflows/codeanalysis.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: liberica
+          java-version: 21
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
+        with:
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+      - name: "🔎 Check Core Projects"
+        run: ./gradlew aggregateAnalysisViolations --continue
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. grails-gradle/build.gradle now applies grails-violation-aggregation, and codeanalysis.yml passes -Pgrails.codeanalysis.enabled.pmd=true -Pgrails.codeanalysis.enabled.spotbugs=true so the analysis tasks both exist and actually run in that build.

.github/workflows/codeanalysis.yml
```diff
@@ -0,0 +1,95 @@
...
+          name: gradle-plugin-reports
+          path: grails-gradle/build/reports/violations/
+      - name: "📋 Publish Code Analysis Report in Job Summary"
+        if: always()
+        run: |
+          echo "## 🔎 Code Analysis Report - Gradle Plugin Projects" >> $GITHUB_STEP_SUMMARY
+          for report in PMD_VIOLATIONS.md SPOTBUGS_VIOLATIONS.md; do
+            file="grails-gradle/build/reports/violations/$report"
+            [ -f "$file" ] && cat "$file" >> $GITHUB_STEP_SUMMARY || true
+          done
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 0ca655d3f1. grails-gradle/build.gradle now applies grails-violation-aggregation, and codeanalysis.yml passes -Pgrails.codeanalysis.enabled.pmd=true -Pgrails.codeanalysis.enabled.spotbugs=true so the analysis tasks both exist and actually run in that build.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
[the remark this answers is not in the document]
jamesfredley: Done in b18a523848. The filter is now configurable via the `grails.jacoco.aggregation.excludedClassPrefixes` Gradle property (comma-separated), defaulting to `org.grails.orm.hibernate.support.hibernate7.`. The default is still needed because the Hibernate 5 and Hibernate 7 variants compile classes with identical fully-qualified names and JaCoCo cannot aggregate two different classes that share a name. The property and its rationale are documented on the plugin, and specs cover both the default exclusion and a configurable override.

sbglasius (approved):

jamesfredley:
## Code-analysis status: configuration now surfaces the violations (converted to draft)

I pushed a config change and converted this PR to **draft**, because once the analysis actually runs it reveals a far larger problem than expected.

### Root cause of the previous "red with no report"

`pmdVersion` was `6.55.0`. PMD 6.x **cannot parse Java 21+ source**, so `pmdMain` threw `PMDException: Error while processing` on essentially every `.java` file and never produced findings. Because the PMD/SpotBugs tasks failed, `aggregateAnalysisViolations` (which only generates the Markdown report, it does not gate) never ran, so the job was red with `No files were found ... build/reports/violations/`.

Two changes in the latest commit:

1. **`pmdVersion` 6.55.0 → 7.25.0** - supports Java 21 through 26 (the Micronaut/Forge modules compile on Java 25).
2. **Code-analysis CI runs report-only** (`-Pgrails.codeanalysis.ignoreFailures=true`) so PMD + SpotBugs complete, the aggregate report is generated, and the full violation list is published to the job summary instead of aborting.

### The scope problem

With PMD parsing correctly, the **Core build alone** reports:

- **5,016 PMD violations**
- **187 SpotBugs violations**

The Forge and Gradle-plugin builds are not measured yet - realistic total is **~8,000-12,000 violations**.

Top PMD rules:

| Count | Rule | Nature |
| ---: | :--- | :--- |
| 1,758 | MissingOverride | trivial/safe (`@Override`) |
| 282 | LooseCoupling | mechanical |
| 264 | AvoidCatchingGenericException | behavior-risky |
| 247 | AvoidLiteralsInIfCondition | stylistic |
| 239 | AvoidReassigningParameters | behavior-risky |
| 203 | LiteralsFirstInComparisons | mechanical |
| 171 | ConstructorCallsOverridableMethod | behavior-risky |
| 150 | CloseResource | behavior-risky |
| 147 | AssignmentInOperand | mixed |
| 140 | UseVarargs | mixed |

Top SpotBugs bug patterns:

| Count | Pattern |
| ---: | :--- |
| 46 | EC_UNRELATED_TYPES_USING_POINTER_EQUALITY |
| 21 | DM_DEFAULT_ENCODING |
| 20 | RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE |
| 16 | MS_SHOULD_BE_FINAL |
| 12 | RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT |

The ruleset currently enables the **entire** `bestpractices` + `errorprone` + `security` categories, which is why the count is so high - a large share are pure style preferences, and many of the rest require behavior-changing edits to framework internals.

### Recommendation

Fixing 8,000-12,000 locations across the framework in one PR is neither reviewable nor safely mergeable. Suggested path before this can gate:

1. Curate the PMD ruleset down to a high-signal subset (and review the SpotBugs `reportLevel`/`effort`).
2. Bulk-fix the safe mechanical rules (e.g. the 1,758 `MissingOverride`).
3. Baseline or defer the behavior-risky rules to follow-up PRs.

> Note: PR #15687 (grails-jacoco) is stacked on top of this branch and inherits these jobs, so it is affected by the same decision.

jamesfredley:
## Code-style (Checkstyle) status: forge backlog left red on purpose

Separate from the PMD/SpotBugs report above, this stack also turns on **Checkstyle gating for the `grails-forge` modules** (the code-style config + `grails-forge/gradle/code-style-config.gradle` are part of this branch's diff vs `8.0.x`). The Core and Gradle-plugin code-style jobs are clean and pass; only **Forge Projects** fails.

I briefly made the code-style jobs report-only for consistency with the analysis jobs, then reverted it - we are **not** masking style errors. The `Forge Projects` code-style job is intentionally **left red** for now rather than fixed, because it's the same "huge backlog" situation.

### The forge Checkstyle backlog: 1,088 errors across 259 files

By module:

| Module | Files | Errors |
| :--- | ---: | ---: |
| grails-forge-core | 165 | 731 |
| grails-forge-cli | 41 | 206 |
| grails-forge-api | 46 | 134 |
| grails-forge-analytics-postgres | 6 | 15 |
| grails-forge-web-netty | 1 | 2 |

By rule:

| Count | Rule |
| ---: | :--- |
| 566 | ImportOrder |
| 206 | UnusedImports |
| 200 | EmptyLineSeparator |
| 37 | Indentation |
| 36 | SingleSpaceSeparator |
| 25 | AvoidStarImport |
| 17 | WhitespaceAfter |
| 1 | SeparatorWrap |

### Note

Unlike the PMD/SpotBugs findings, every one of these is a **purely mechanical formatting rule** - no behavior change. The top three (ImportOrder + UnusedImports + EmptyLineSeparator = ~89%) are exactly what an IDE "Optimize Imports" + reformat, or a Spotless/`importOrder` pass, would fix automatically. So this backlog is safe to clear in bulk whenever we choose to - it's just out of scope for this draft right now.

> PR #15687 is stacked on this branch and inherits the same code-style jobs / red Forge result.

jamesfredley:
This will have to wait for Grails 8.1 or 9, there is too much work left.

borinquenkid:
@jamesfredley For the SpotBugs and PMD, have two skill features one after running report does not break build but updates a predetermined ISSUE with the report. Have another skill called fix SpotBugsOrPmd that picks one and fixes it and creates a PR. If anyone has  tokens to spare could create quite a few focused PRS. It would move issues from under the radar to very visible and we can tackle them whenever we have the bandwidth

jdaugherty:
I don't think we can merge hibernate 7 without this change.  This is a critical component of that merge.  We only have it configured to run in hibernate 7.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -90,37 +100,26 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
...
-                throw new IllegalStateException("Could not locate default configuration file: ${defaultResource}")
-            }
-            expectedPath.text = defaultValue.text
-        }
-    }
-
-    private static void doNotApplyStylingToTests(Project project) {
-        project.tasks.named('checkstyleTest') {
-            it.enabled = false // Do not check test sources at this time
+    private static void createOrLoad(Path expectedPath, String defaultResource, Project project) {
```
jdaugherty: The changes to this method should be reverted.

jdaugherty:
This PR expanded scope: `grails-code-analysis convention plugin (PMD + SpotBugs) applied to 96 subprojects, with GrailsCodeAnalysisExtension.`  We do not need to do this.  The original change was to help implement hibernate 7.

jamesfredley:
I now understand that PMD is for 3 subprojects right now and not the entire project.  That changes scope and I'll update and move back from draft.

It appeared there was a large amount of work across the whole codebase and I now see that being possible in our short timeline.

borinquenkid (approved):
@jamesfredley @jdaugherty Since this is totally optional in a regular build it should be ok to add. Once we get past 8.0.x I want to add agentic solutions of PMD and Spotbugs one at a time

borinquenkid:
1. JDK 25 Compatibility Fix:                                                                                                                                                                            
      • Removed  google-java-format  from the  grails-forge  Spotless configuration, resolving runtime compilation crashes ( NoSuchMethodError ) caused by internal compiler API changes in JDK 25.       
  2. Unified Styling & Layout Conventions:                                                                                                                                                                
      • Modified Spotless configurations in  grails-forge  ( gradle/code-style-config.gradle ) to use custom conventions ( importOrder  +  removeUnusedImports ) to mirror Checkstyle's layout            
      requirements.                                                                                                                                                                                       
      • Restricted Spotless scans strictly to production Java source files to prevent conflict with tasks generating resources (e.g.  copyGrailsWrapper ).                                                
  3. Enhanced Checkstyle Rules:                                                                                                                                                                           
      • Configured Checkstyle in the  grails-forge  project to run layout validation on all production Java source files.                                                                                 
      • Cleared all formatting and import violations across the  grails-forge-core ,  grails-forge-api , and  grails-forge-cli  modules.                                                                  
  4. Build Cleanup:                                                                                                                                                                                       
      • Removed obsolete formatting task declarations from the build configurations.

sbglasius (approved):

borinquenkid:
Update — Spotless removed from grails-forge (supersedes the Spotless changes described above)
  
  While clearing the remaining Checkstyle violations, the importOrder + removeUnusedImports Spotless config turned out to be unable to fully mirror Checkstyle:

  - Spotless's prefix-based importOrder treats grails.* and org.grails.* as separate groups (inserts a blank line between them), while Checkstyle's regex merges them into one group (forbids that blank
  line). No Spotless configuration can express that merged group, so any file importing both (e.g. AddPropertyCommand) was an unsatisfiable conflict — Spotless and Checkstyle could never both pass.
  - Because checkstyleMain depended on spotlessCheck, that conflict (and pre-existing violations elsewhere) were gated/hidden rather than surfaced.

  Resolution: removed the bespoke Spotless config from grails-forge entirely. Layout is now owned solely by Checkstyle (verification) + the shared project .idea code-style, which can express the merged
  grails group and matches Checkstyle exactly. This puts grails-forge on the same Checkstyle + CodeNarc model as the root build.

  - JDK 25 fix still holds — there's no Java formatter left to hit the NoSuchMethodError.
  - Dropped the checkstyleMain dependsOn spotlessCheck gate; Checkstyle now runs on all production Java unconditionally.
  - Cleared all surfaced violations across grails-forge-core, grails-forge-api, and grails-forge-cli (whitespace, import order, annotation-array indentation).

  ⚠️ Heads-up: a surprising transitive dependency we hit and deliberately left in place

  We could not delete the spotless-plugin-gradle line from grails-forge/buildSrc/build.gradle, even though Spotless is no longer applied. It's load-bearing for dependency resolution, not for formatting:

  io.micronaut.build.internal:micronaut-gradle-plugins
    └─ com.diffplug.spotless:spotless-plugin-gradle:6.8.0   (old, transitive)
         └─ com.diffplug.spotless:spotless-lib-extra:2.27.0
              └─ org.codehaus.groovy:groovy-xml:3.0.9        ← Groovy 3 (codehaus)

  micronaut-gradle-plugins drags in an old Spotless (6.8.0) that pulls Groovy 3 (org.codehaus.groovy), which collides with the buildSrc's Groovy 4 (org.apache.groovy) and fails resolution. The explicit
  spotless-plugin-gradle:6.25.0 we keep forces Spotless up to a Groovy-4-compatible version, overriding the transitive 6.8.0. So the dependency stays purely as a version pin — it is never applied as a
  plugin.
  
  Follow-up candidate (out of scope here): replace this implicit pin with an explicit dependency constraint, or exclude the old Spotless from micronaut-gradle-plugins, so the intent is obvious rather than
  relying on a leftover implementation line.

  Commit: 4149f44554

matrei:
> I don't think we can merge hibernate 7 without this change. This is a critical component of that merge. We only have it configured to run in hibernate 7.

@jdaugherty Why is it a critical component of merging an implementation of Hibernate 7?

Also, a general observation, there seems to be a lot of commits with functional and formatting changes combined.
It is helpful if we separate these into different commits.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GradleUtils.groovy
```diff
@@ -37,6 +38,12 @@ class GradleUtils {
         asfFile.exists() ? currentDirectory : findAsfRootDir(currentDirectory.dir('../'))
     }
 
+    static Provider<Boolean> booleanProvider(Project project, String name, boolean defaultValue = false) {
+        project.providers.gradleProperty(name)
+                .map { Boolean.parseBoolean(it as String) }
```
matrei: Should we use: `it.trim().toBoolean()`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisExtension.groovy
```diff
@@ -0,0 +1,53 @@
...
+class GrailsCodeAnalysisExtension {
+
+    /**
+     * Defaults to project.rootProject.buildDir/codeanalysis/pmd.
+     * Directory for PMD configuration files (e.g. pmd.xml).
+     */
+    final DirectoryProperty pmdDirectory
+
+    /**
+     * Defaults to rootProject.buildDir/reports/codeanalysis.
```
matrei: Should this be `project.rootProject`, like the one above, or can we remove the `project.` from both javadoc snippets?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisExtension.groovy
```diff
@@ -0,0 +1,53 @@
...
+     * PMD and SpotBugs XML reports will be written here.
+     */
+    final DirectoryProperty reportsDirectory
+
+    @Inject
+    GrailsCodeAnalysisExtension(ObjectFactory objects, Project project) {
+        pmdDirectory = objects.directoryProperty().convention(
+                project.rootProject.layout.buildDirectory.dir('codeanalysis/pmd')
```
matrei: Could we use directory name `code-analysis`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisExtension.groovy
```diff
@@ -0,0 +1,53 @@
...
+        )
+        reportsDirectory = objects.directoryProperty().convention(
+                project.rootProject.layout.buildDirectory.dir('reports/codeanalysis')
```
matrei: `code-analysis`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+import com.github.spotbugs.snom.SpotBugsTask
+
+/**
+ * Convention plugin for Grails byte code analysis (PMD and SpotBugs).
+ * Both tools are opt-in; enable via Gradle properties.
+ */
+@CompileStatic
+class GrailsCodeAnalysisPlugin implements Plugin<Project> {
+
+    static String PMD_DIR_PROPERTY = 'grails.codeanalysis.dir.pmd'
```
matrei: Can we use hyphened `code-analysis` here and below?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    private static void initExtension(Project project) {
+        def gca = project.extensions.create('grailsCodeAnalysis', GrailsCodeAnalysisExtension)
+
+        gca.pmdDirectory.set(project.provider {
+            def directory = project.hasProperty(PMD_DIR_PROPERTY) ?
+                    project.rootProject.layout.projectDirectory.dir(project.property(PMD_DIR_PROPERTY) as String) :
+                    project.rootProject.layout.buildDirectory.get().dir('codeanalysis').dir('pmd')
```
matrei: `code-analysis`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+                    "${BASE_RESOURCE_PATH}/pmd/${PMD_CONFIG_FILE_NAME}",
+                    project
+            )
+
+            directory
+        })
+    }
+
+    private static void createOrLoad(Path expectedPath, String defaultResource, Project project) {
+        boolean defaultPath = expectedPath.startsWith(project.rootProject.buildDir.toPath())
```
matrei: `buildDir` deprecated?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+            if (!defaultValue) {
+                throw new IllegalStateException("Could not locate default configuration file: ${defaultResource}")
+            }
+            project.logger.info('Replacing code analysis configuration')
+            expectedPath.text = defaultValue.text
+        }
+    }
+
+    static void configurePmd(Project project) {
+        Provider<Boolean> pmdEnabled = GradleUtils.booleanProvider(project, PMD_ENABLED_PROPERTY)
```
matrei: `def` here and below where the type can be inferred?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+                    project.extensions.getByType(GrailsCodeAnalysisExtension)
+                            .reportsDirectory.get()
+                            .dir('pmd')
+                            .file("${project.name}-${task.name}.xml")
+            )
+        }
+    }
+
+    static void configureSpotbugs(Project project) {
+        Provider<Boolean> spotbugsEnabled = GradleUtils.booleanProvider(project, SPOTBUGS_ENABLED_PROPERTY)
```
matrei: `def` where the type can be inferred here and below?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+
+        project.extensions.configure(SpotBugsExtension) {
+            it.effort.set(Effort.valueOf('MAX'))
+            it.reportLevel.set(Confidence.valueOf('HIGH'))
+            it.ignoreFailures.set(ignoreFailures)
+        }
+
+        project.tasks.withType(SpotBugsTask).configureEach { SpotBugsTask task ->
+            task.group = 'verification'
+            NamedDomainObjectContainer<SpotBugsReport> spotBugsReports = task.reports
```
matrei: `def`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+            SpotBugsReport htmlReport = spotBugsReports.maybeCreate('html')
```
matrei: `def`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
```diff
@@ -0,0 +1,183 @@
...
+            htmlReport.required.set(true)
+            SpotBugsReport xmlReport = spotBugsReports.maybeCreate('xml')
```
matrei: `def`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStyleExtension.groovy
```diff
@@ -50,10 +50,10 @@ class GrailsCodeStyleExtension {
...
         checkstyleDirectory = objects.directoryProperty().convention(
-                project.rootProject.layout.buildDirectory.dir('checkstyle')
+                project.rootProject.layout.buildDirectory.dir('codestyle/checkstyle')
         )
         codenarcDirectory = objects.directoryProperty().convention(
-                project.rootProject.layout.buildDirectory.dir('codenarc')
+                project.rootProject.layout.buildDirectory.dir('codestyle/codenarc')
         )
         reportsDirectory = objects.directoryProperty().convention(
                 project.rootProject.layout.buildDirectory.dir('reports/codestyle')
```
matrei: This was not added in this PR but should probably be `code-style`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -21,44 +21,52 @@ package org.apache.grails.buildsrc
...
 import org.gradle.api.plugins.quality.CodeNarcPlugin
+import org.gradle.api.provider.Provider
 
+/**
+ * Convention plugin for Grails code style enforcement (Checkstyle and CodeNarc).
+ */
 @CompileStatic
 class GrailsCodeStylePlugin implements Plugin<Project> {
 
     static String CHECKSTYLE_DIR_PROPERTY = 'grails.codestyle.dir.checkstyle'
```
matrei, line 45: Camel or kebab for `codestyle` here and below?

jdaugherty:
> > I don't think we can merge hibernate 7 without this change. This is a critical component of that merge. We only have it configured to run in hibernate 7.
> 
> 
> 
> @jdaugherty Why is it a critical component of merging an implementation of Hibernate 7?
> 
> 
> 
> Also, a general observation, there seems to be a lot of commits with functional and formatting changes combined.
> 
> It is helpful if we separate these into different commits.

We used this plugin only in the hibernate 7 dev to ensure security best practices.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -131,31 +126,40 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
...
-            it.dependsOn(project.tasks.withType(Checkstyle))
             it.dependsOn(project.tasks.withType(CodeNarc))
+            it.dependsOn(project.tasks.withType(Checkstyle))
         }
     }
 
     static void configureCheckstyle(Project project) {
         project.pluginManager.apply(CheckstylePlugin)
 
+        Provider<Boolean> ignoreFailures = GradleUtils.booleanProvider(project, IGNORE_FAILURES_PROPERTY)
```
matrei: `def`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -131,31 +126,40 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
...
+
         project.extensions.configure(CheckstyleExtension) {
             // Explicit `it` is required in extension configuration
             it.getConfigDirectory().set(project.extensions.getByType(GrailsCodeStyleExtension).checkstyleDirectory)
```
matrei: Not changed in this PR but could use groovy property accessor.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -131,31 +126,40 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
...
             task.group = 'verification'
             task.onlyIf { !project.hasProperty('skipCodeStyle') }
+            task.ignoreFailures = ignoreFailures.get()
+
+            if (task.name.toLowerCase().contains('test')) {
+                task.enabled = false
+            }
+
+            task.exclude { org.gradle.api.file.FileTreeElement element ->
+                element.getFile().getAbsolutePath().contains("/build/")
```
matrei: - Add comment why this is necessary and for which projects?
- Will this work on Windows?
- Use Groovy property accessors?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
```diff
@@ -168,20 +172,33 @@ class GrailsCodeStylePlugin implements Plugin<Project> {
     static void configureCodenarc(Project project) {
         project.pluginManager.apply(CodeNarcPlugin)
 
+        registerCodenarcFixTask(project)
+
+        Provider<Boolean> ignoreFailures = GradleUtils.booleanProvider(project, IGNORE_FAILURES_PROPERTY)
```
matrei: `def` here and below?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStyleExtension.groovy
```diff
@@ -30,13 +30,13 @@ import org.gradle.api.model.ObjectFactory
 class GrailsCodeStyleExtension {
 
     /**
-     * Defaults to project.buildDir/checkstyle.
+     * Defaults to project.rootProject.buildDir/codestyle/checkstyle.
```
matrei: - Align javadoc path?
- `code-style`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+import org.gradle.api.plugins.AppliedPlugin
+import org.gradle.api.plugins.quality.Checkstyle
+import org.gradle.api.plugins.quality.CodeNarc
+import org.gradle.api.plugins.quality.Pmd
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.testing.jacoco.tasks.JacocoReport
+
+import com.github.spotbugs.snom.SpotBugsTask
+import groovy.xml.XmlSlurper
```
matrei: Order imports?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+    @Override
+    void apply(Project project) {
+        if (project != project.rootProject) {
+            throw new GradleException(
+                'GrailsViolationAggregationPlugin must be applied to the root project only. ' +
+                'Apply grails-code-style and grails-jacoco to subprojects instead.'
+            )
+        }
+
+        Provider<Directory> violationsDir = project.layout.buildDirectory.dir('reports/violations')
```
matrei: `def` here and below?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+        Provider<Directory> styleXmlDir = project.layout.buildDirectory.dir('reports/codestyle')
```
matrei: `code-style`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+        Provider<Directory> analysisXmlDir = project.layout.buildDirectory.dir('reports/codeanalysis')
```
matrei: `code-analysis`?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+                    task.dependsOn(sub.tasks.withType(JacocoReport))
+                }
+            }
+        }
+    }
+
+    @CompileDynamic
+    private static void parseStyleViolations(Directory styleXmlDir, Directory violationsDir,
+            boolean checkStyleTests, boolean codenarcEnabled, boolean checkstyleEnabled) {
+        def slurper = new XmlSlurper()
```
matrei: Extract duplicate slurper creation into helper?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,501 @@
...
+            if (includeTests) {
+                return false
+            }
+            if (filePath && (filePath.contains('src/test/') || filePath.contains('src/integrationTest/'))) {
+                return true
+            }
+            !filePath && (className.contains('Spec') || className.contains('Test') || className.contains('Tests'))
+        }
+
+        def writeReport = { String fileName, List violations, String title ->
```
matrei: Looks like there is some duplication here?

matrei (commented):
This is a first iteration, I did not get through it all.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GradleUtils.groovy
```diff
@@ -37,6 +38,12 @@ class GradleUtils {
         asfFile.exists() ? currentDirectory : findAsfRootDir(currentDirectory.dir('../'))
     }
 
+    static Provider<Boolean> booleanProvider(Project project, String name, boolean defaultValue = false) {
+        project.providers.gradleProperty(name)
+                .map { String it -> it.trim().toBoolean() }
```
matrei: No need for explicit `it` parameter?

matrei:
> > > I don't think we can merge hibernate 7 without this change. This is a critical component of that merge. We only have it configured to run in hibernate 7.
> > 
> > 
> > @jdaugherty Why is it a critical component of merging an implementation of Hibernate 7?
> > Also, a general observation, there seems to be a lot of commits with functional and formatting changes combined.
> > It is helpful if we separate these into different commits.
> 
> We used this plugin only in the hibernate 7 dev to ensure security best practices.

That is interesting. I have not used PMD before. Looking at the [security](https://docs.pmd-code.org/latest/pmd_rules_java.html#security) rule set, there only seems to be two rules, and none of them seems to apply to Hibernate 7 code. Are there other rules that helps ensure security best practices?

build-logic/plugins/src/test/groovy/org/apache/grails/buildsrc/GrailsCodeStylePluginSpec.groovy
```diff
@@ -0,0 +1,223 @@
...
+"""
+        when: "running codenarcFix"
+        def result = GradleRunner.create()
+                .withProjectDir(testProjectDir.toFile())
+                .withArguments('codenarcFix', '--stacktrace')
+                .withPluginClasspath()
+                .build()
+
+        then: "task finished successfully"
+        result.task(':codenarcFix').outcome == org.gradle.testkit.runner.TaskOutcome.SUCCESS
```
matrei: Unnecessary qualified reference? (here and below)

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,25 +1,5 @@
...
-     "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-     KIND, either express or implied.  See the License for the
-     specific language governing permissions and limitations
-     under the License.
-
--->
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default copy" />
```
matrei: What happened here?

.idea/codeStyles/Project.xml
```diff
@@ -1,105 +1,97 @@
...
-            <option name="BLANK_LINES_AFTER_CLASS_HEADER" value="1"/>
-            <option name="ARRAY_INITIALIZER_WRAP" value="1"/>
-            <option name="ARRAY_INITIALIZER_LBRACE_ON_NEXT_LINE" value="true"/>
-            <indentOptions>
-                <option name="CONTINUATION_INDENT_SIZE" value="4"/>
-            </indentOptions>
-        </codeStyleSettings>
-    </code_scheme>
+  <code_scheme name="Project" version="173">
+    <option name="AUTODETECT_INDENTS" value="false" />
```
matrei: Refrain from functional and whitespace changes in the same commit?

matrei:
Is it a good idea to change Forge styling in this PR?

borinquenkid:
> Is it a good idea to change Forge styling in this PR?

It was failing CodeStyle!!!

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GradleUtils.groovy
borinquenkid: Done

build-logic/plugins/src/test/groovy/org/apache/grails/buildsrc/GrailsCodeStylePluginSpec.groovy
borinquenkid: Done

.idea/codeStyles/codeStyleConfig.xml
borinquenkid: Done

.idea/codeStyles/Project.xml
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeAnalysisPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStyleExtension.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStylePlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsCodeStyleExtension.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
borinquenkid: Done

jamesfredley:
@matrei I am trimming this back to @borinquenkid's original intent, which was PMD for a small number of projects.  I misunderstood the scope and thought it was standardize and apply to entire codebase.  It was more prove tool in small place and then expand later.

jamesfredley:
Dialed this PR back to its intended scope based on the review feedback. Pushed three focused commits:

**1. Limit `grails-code-analysis` to its original scope** (`6d79c8addc`)
Removed the `grails-code-analysis` (PMD + SpotBugs) plugin application from all 96 subprojects. The convention plugin stays registered in `build-logic` as available infrastructure - it is just applied nowhere on this branch.

On why it is zero and not "3 subprojects": tracing the Hibernate 7 branch history, the analysis plugin originally ran on a **single** module, `grails-data-hibernate7/core`, before a later commit expanded it project-wide. That module does not exist on `8.0.x` yet (it lives in the Hibernate 7 work this was carved out of), so there is nothing here to apply it to. **When the Hibernate 7 work is merged back in, the analysis plugin will need to be re-hooked up to the relevant `grails-data-hibernate7` module(s).**

**2. Restore code style configuration to the 8.0.x baseline** (`b08441587d`)
`grails-forge` is now byte-identical to `8.0.x`: its own (more lenient) Checkstyle ruleset and Spotless config are back, and the reformatting of 217 source files is reverted. Changing Forge's styling here was accidental - Forge keeps its separate rule set and its code style job passes as before (verified `:grails-forge-core:spotlessCheck checkstyleMain` green on JDK 21). This commit also:
- reverts `createOrLoad()` to the original create-only-if-missing behavior (per @jdaugherty's review),
- keeps the legacy `grails.codestyle.dir.*` property-override keys so Forge's custom config dir still resolves (otherwise Forge falls back to the bundled strict ruleset),
- points the Forge code-style CI job back at `./gradlew codeStyle` (Forge no longer applies the violation-aggregation plugin).

**3. Fix code-style report path in the violation-aggregation spec** (`3b808a9bb8`)
A test was still reading the pre-rename `build/reports/codestyle` location; aligned it with the plugin's `build/reports/code-style` output.

Verification: `build-logic:test` passes, the root build configures (JDK 21), and Forge code style passes.

Note: `codeanalysis.yml` is intentionally dormant for now (analysis is applied nowhere) and will start producing real reports again once analysis is re-hooked to the Hibernate 7 module(s) on merge-back.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
matrei: It seems `com.github.spotbugs.snom.SpotBugsTask` is in the wrong place. How do you execute the import formatting? I press `Ctrl+Alt+O` in Intellij.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
matrei: There are more explicit types below.
Lines 90-91, 104-106, 108, 134-136, 138, 165, 171, 178

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+
+    private static XmlSlurper createSecureSlurper() {
+        def slurper = new XmlSlurper()
+        slurper.setFeature('http://apache.org/xml/features/disallow-doctype-decl', true)
+        slurper.setFeature('http://apache.org/xml/features/nonvalidating/load-external-dtd', false)
+        slurper.setFeature('http://xml.org/sax/features/external-general-entities', false)
+        slurper.setFeature('http://xml.org/sax/features/external-parameter-entities', false)
+        slurper.setFeature('http://xml.org/sax/features/namespaces', false)
+        return slurper
+    }
```
matrei, line 205: Simplify?
```suggestion
    private static XmlSlurper createSecureSlurper() {
        new XmlSlurper().tap {
            setFeature('http://apache.org/xml/features/disallow-doctype-decl', true)
            setFeature('http://apache.org/xml/features/nonvalidating/load-external-dtd', false)
            setFeature('http://xml.org/sax/features/external-general-entities', false)
            setFeature('http://xml.org/sax/features/external-parameter-entities', false)
            setFeature('http://xml.org/sax/features/namespaces', false)
        }
    }
```

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+            root.allprojects.collect { Project p -> p.file('build/reports/jacoco/test/jacocoTestReport.csv') }
+        )
+
+        // Resolve the excluded class-name prefixes as a Provider so the value is captured
+        // configuration-cache-safely and read at task execution time.
+        Provider<List<String>> excludedClassPrefixes = root.providers
+            .gradleProperty(JACOCO_EXCLUDED_CLASS_PREFIXES_PROPERTY)
+            .orElse(DEFAULT_JACOCO_EXCLUDED_CLASS_PREFIXES)
+            .map { String value ->
+                value.split(',').collect { it.trim() }.findAll { !it.isEmpty() }
```
matrei: Simplify?
```suggestion
                value.split(',').collect { it.trim() }.findAll()
```

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+        def checkstyleDir = styleXmlDir.dir('checkstyle').asFile
+        if (checkstyleDir.exists() && checkstyleEnabled) {
+            checkstyleDir.eachFileMatch(~/.*\.xml/) { file ->
+                if (file.size() == 0 || (!checkStyleTests && isTestFile(file.name))) {
+                    return
+                }
+                def module = getModule(file.name)
+                def xml = slurper.parse(file)
+                xml.file.each { f ->
+                    def filePath = f.@name.text()
```
matrei: Use explicit type `String filePath`. Cannot be inferred.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+                out.append('| Class | Tool | Violation | Line | Message |\n')
+                out.append('| :--- | :--- | :--- | :--- | :--- |\n')
+                modViolations.each { v ->
+                    out.append("| ${v.className} | ${v.tool} | ${v.type} | ${v.line} | ${v.message.replaceAll(/\|/, '\\|')} |\n")
+                }
+                out.append('\n')
+            }
+        }
+        reportFile.text = out.toString()
+        LOGGER.lifecycle("Aggregated report generated: ${reportFile.absolutePath}")
```
matrei: Use lazily evaluated log statements:
```suggestion
        LOGGER.lifecycle('Aggregated report generated: {}', reportFile.absolutePath)
```

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+        }
+        writeReport(violationsDir, 'SPOTBUGS_VIOLATIONS.md', spotbugsViolations, 'SpotBugs Violations Summary')
+    }
+
+    @CompileDynamic
+    private static void parseJacocoCoverage(FileCollection csvFiles, Directory violationsDir, List<String> excludedClassPrefixes) {
+        def jacocoCoverage = []
+        csvFiles.each { File csvReport ->
+            if (csvReport.exists()) {
+                LOGGER.debug("Processing JaCoCo report: ${csvReport.absolutePath}")
```
matrei: Use lazily evaluated log statements:
```suggestion
                LOGGER.debug('Processing JaCoCo report: {}', csvReport.absolutePath)
```

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsViolationAggregationPlugin.groovy
```diff
@@ -0,0 +1,465 @@
...
+            out.append("## Module: ${module}\n")
+            out.append('| Class | % Instructions Covered |\n')
+            out.append('| :--- | :--- |\n')
+            coverageList.sort { it.percent }.each { c ->
+                out.append("| ${c.className} | ${c.percent}% |\n")
+            }
+            out.append('\n')
+        }
+        reportFile.text = out.toString()
+        LOGGER.lifecycle("Aggregated JaCoCo report generated: ${reportFile.absolutePath}")
```
matrei: Use lazily evaluated log statements:
```suggestion
        LOGGER.lifecycle('Aggregated JaCoCo report generated: {}', reportFile.absolutePath)
```

.agents/skills/violation-fixer/SKILL.md
matrei, line 1: This file is not up to date with which projects are affected (Every subproject), but I'm not sure if that is a problem or not.

matrei (approved):

PR #15692, opened 2026-05-29
Remove no-op failOnError argument from DeleteEntityDataFetcher

jamesfredley:
## Summary

`DeleteEntityDataFetcher.deleteInstance()` called `instance.delete(failOnError: true)`, but GORM's `delete(Map)` only honors the `flush` key - the `failOnError` argument is silently ignored. The call therefore already behaved as a plain, non-flushing `delete()`. The `failOnError` argument gave a false impression of doing something when it did nothing.

This removes that misleading argument by switching to a plain `instance.delete()`. **Behavior is unchanged**: the delete still executes without forcing a flush, exactly as before. This is a pure cleanup with no functional change - it just drops a parameter that was a no-op.

> Note: an earlier revision of this PR switched the call to `delete(flush: true)`. That would have *changed* behavior by forcing an immediate flush, so it was reverted in favor of the no-op `delete()` cleanup.

## Changes

- `DeleteEntityDataFetcher`: `delete(failOnError: true)` to `delete()`.
- `DeleteEntityDataFetcherSpec`: updated the interaction test to assert `deleteInstance` calls `delete()` with no arguments and never passes the ignored `failOnError` argument.

## Testing

```
./gradlew :grails-data-graphql-core:test --tests "org.grails.gorm.graphql.fetcher.impl.DeleteEntityDataFetcherSpec" --tests "org.grails.gorm.graphql.fetcher.impl.SoftDeleteEntityDataFetcherSpec"
```

All pass, including the existing happy-path / invalid-id cases and the `SoftDeleteEntityDataFetcher` (which overrides `deleteInstance`) regression check.

Flagged during review of #15599.


grails-data-graphql/core/src/test/groovy/org/grails/gorm/graphql/fetcher/impl/DeleteEntityDataFetcherSpec.groovy
```diff
@@ -76,6 +77,19 @@ class DeleteEntityDataFetcherSpec extends HibernateSpec {
...
+        given:
+        DeleteEntityDataFetcher fetcher = new DeleteEntityDataFetcher<>(mappingContext.getPersistentEntity(One.name))
+        GormEntity instance = Mock(GormEntity)
+
+        when:
+        fetcher.deleteInstance(instance)
+
+        then: 'the delete is flushed so database failures surface, and failOnError (ignored by delete) is not used'
+        1 * instance.delete([flush: true])
+        0 * instance.delete([failOnError: true])
```
jdaugherty: Mock tests generally are fragile since we're not actually testing the result.  Shouldn't we instead check the persisted state?

jdaugherty (commented):
I had no idea that delete doesn't honor failOnError.  Seems like a gap in the API.  It makes sense to remove the failOnError, but forcing a flush may have unexpected behavior.  Why are you flushing by default now?  Wouldn't it be better to let the transaction finish and let the transaction manager flush instead?

jamesfredley:
Updated this PR to avoid changing any behavior.

The previous revision switched the call to `instance.delete(flush: true)`. While that fixed the misleading `failOnError` argument, it also **changed runtime behavior** by forcing an immediate flush on every delete - which is more than this cleanup should do.

The latest commit instead changes the call to a plain `instance.delete()`:

- The original `delete(failOnError: true)` was effectively a non-flushing `delete()`, since GORM's `delete(Map)` only honors the `flush` key and silently ignores `failOnError`.
- Plain `delete()` keeps that exact behavior - **no flush change, no functional change**.
- It simply removes the `failOnError` parameter, which created a false impression of doing something when it did nothing.

In short: this is now a pure no-op cleanup that drops a dead parameter, not a behavior change. The interaction test was updated to assert `delete()` is called with no arguments and the ignored `failOnError` argument is never passed.

jdaugherty (approved):

PR #15693, opened 2026-05-29
fix: Order milestone and release candidate versions in VersionComparator

### LINKED ISSUE #14058: Test Plugin Compatibility Check with Milestone and Snapshot Versions
### Issue description

https://github.com/grails/grails-core/blob/80792f89dc73f349f029c544786a7422f962859d/grails-core/src/main/groovy/grails/plugins/DefaultGrailsPluginManager.java#L398-L465

Test with a combination of plugins requiring a `milestone` or `snapshot` version and the current Grails version being a `milestone` or `snapshot` version.

The check produces a warning and not an error.

`Plugin [myplugin:0.1.0-rc3] may not be compatible with this application as the application Grails version is less than the plugin requires. Plugin is compatible with Grails version 3.3.10 > * but app is 4.0.0`

jamesfredley:
Fixes #14058

## Summary

`grails.plugins.VersionComparator` did not understand milestone (`M`) or release candidate (`RC`) version qualifiers. The plugin compatibility check in `DefaultGrailsPluginManager` delegates to this comparator, so plugins targeting (or applications running on) a milestone or release candidate produced incorrect "may not be compatible with this application" warnings. This PR makes the comparator qualifier-aware and adds the missing test coverage requested by the issue.

## The bug

`VersionComparator` tokenised a version by splitting on `.` and keeping only purely numeric tokens (`==~ /\d+/`). `-SNAPSHOT` / `.BUILD-SNAPSHOT` were handled as a special case, but `M`/`RC` qualifiers were not, so the qualifier - and often the patch number - was silently discarded:

| Input | Parsed as (before) | Problem |
|-------|--------------------|---------|
| `7.0.0-M1` | `[7, 0]` | patch `0` and `-M1` both dropped |
| `7.0.5-M1` | `[7, 0]` | patch `5` lost; compares as `7.0` |
| `7.0.0.M1` | `[7, 0, 0]` | `M1` dropped; compares equal to final `7.0.0` |

This led to two concrete defects in the compatibility check:

1. **Lost patch number.** A plugin requiring `7.0.3 > *` running on `7.0.5-M1` was wrongly flagged incompatible, because `7.0.5-M1` was parsed as `7.0` (< `7.0.3`).
2. **Milestones/RCs of a base version all compared equal**, and equal to the final release. A plugin requiring the final `7.0.0` was treated as compatible with the pre-release `7.0.0-RC1`, and milestone ordering carried no meaning.

### Why this looked like it already had coverage

The codebase has three independent version representations. The well-tested milestone/RC ordering lives on classes the compatibility check never calls:

| Class | Handles `M`/`RC`? | Tested for `M`/`RC`? | Used by the compatibility check? |
|-------|-------------------|----------------------|----------------------------------|
| `grails.plugins.VersionComparator` | No (fixed here) | No (added here) | **Yes** |
| `org.grails.datastore.mapping.core.grailsversion.GrailsVersion` / `Snapshot` | Yes | Yes | No |
| `grails.init.GrailsVersion` (wrapper) | Yes | Yes | No |

## The fix

`VersionComparator` now parses a version into its numeric components plus an optional qualifier:

- Numeric components are compared first, zero padding the shorter side so `7.0` == `7.0.0` and `7.0.1` > `7.0.0-RC9`.
- When the numeric components are equal, the qualifier breaks the tie using the same ordering as `GrailsVersion` / `Snapshot`:

  ```
  7.0.0-M1 < 7.0.0-M2 < 7.0.0-RC1 < 7.0.0-RC2 < 7.0.0-SNAPSHOT < 7.0.0
  ```

- Dotted (`7.0.0.M1`) and hyphenated (`7.0.0-M1`) forms are treated as equivalent, and qualifier matching is case insensitive (`7.0.0-rc3` == `7.0.0-RC3`).
- Milestone and release candidate numbers are compared numerically, so `7.0.0-RC10` > `7.0.0-RC2`.
- Unrecognised qualifiers are still treated as a final release, preserving the previous "ignore unknown suffix" behaviour.

The protected `deSnapshot` / `isSnapshot` methods are retained for backwards compatibility.

## Behaviour preserved

All pre-existing `VersionComparatorSpec` and `DefaultGrailsPluginManagerSpec` cases are unchanged and still pass (numeric comparisons, `BUILD-SNAPSHOT < release`, the existing range checks, etc.).

## Blast radius

`VersionComparator` has four callers, all verified safe:

- `GrailsVersionUtils.isValidVersion` - already strips qualifiers via `trimTag()` before comparing, so its behaviour is unchanged.
- `GrailsVersionUtils.isVersionGreaterThan` - no callers.
- `RegexUrlMapping` - compares user-defined API versions (plain numeric, no `M`/`RC`).
- `DefaultGrailsPluginManager.isCompatiblePlugin` - the path this PR fixes.

## Tests

- `VersionComparatorSpec`: pre-release vs final ordering, patch preservation (`7.0.5-M1` > `7.0.0`), milestone/RC numeric ordering (`7.0.0-RC10` > `7.0.0-RC2`), the full tier order, dotted/hyphenated and snapshot-form equivalence, numeric-base dominance, plus a sort assertion over a mixed list.
- `DefaultGrailsPluginManagerSpec`: `isCompatiblePlugin` rows for milestone, release candidate and snapshot versions on both the application and the plugin.

## Verification

- `./gradlew :grails-core:test --tests "grails.plugins.VersionComparatorSpec" --tests "grails.plugins.DefaultGrailsPluginManagerSpec"` - passing.
- `./gradlew :grails-bootstrap:codenarcMain` - passing.

### DIFF
--- a/grails-bootstrap/src/main/groovy/grails/plugins/VersionComparator.groovy
+++ b/grails-bootstrap/src/main/groovy/grails/plugins/VersionComparator.groovy
@@ -18,79 +18,66 @@
  */
 package grails.plugins
 
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+
 import groovy.transform.CompileStatic
 
 /**
- * A comparator capable of sorting versions from from newest to oldest
+ * A comparator capable of sorting versions from newest to oldest.
+ *
+ * <p>Versions are compared by their numeric components first (major, minor, patch, ...),
+ * padding the shorter side with zeros so that {@code 7.0} and {@code 7.0.0} are equal.
+ * When the numeric components are equal the version qualifier is used as a tie-breaker
+ * following the same ordering as {@code org.grails.datastore.mapping.core.grailsversion.GrailsVersion}:</p>
+ *
+ * <pre>
+ * 7.0.0-M1 &lt; 7.0.0-M2 &lt; 7.0.0-RC1 &lt; 7.0.0-RC2 &lt; 7.0.0-SNAPSHOT &lt; 7.0.0
+ * </pre>
+ *
+ * <p>In other words a milestone is older than a release candidate, a release candidate is
+ * older than a snapshot, and any qualified (pre-release) version is older than the final
+ * release of the same number. Both the dotted ({@code 7.0.0.M1}) and hyphenated
+ * ({@code 7.0.0-M1}) qualifier forms are treated as equivalent. Unrecognised qualifiers are
+ * treated as a final release to preserve backwards compatible behaviour.</p>
  */
 @CompileStatic
 class VersionComparator implements Comparator<String> {
 
+    private static final String SNAPSHOT = 'SNAPSHOT'
+    private static final String RELEASE_CANDIDATE = 'RC'
+    private static final String MILESTONE = 'M'
+
     static private final List<String> SNAPSHOT_SUFFIXES = ['-SNAPSHOT', '.BUILD-SNAPSHOT'].asImmutable()
 
+    private static final Pattern DIGITS = ~/\d+/
+    private static final Pattern NUMERIC_PREFIX = ~/(\d+)-(.+)/
+    private static final Pattern TRAILING_DIGITS = ~/(\d+)$/
+
+    private static final int TIER_FINAL = 4
+    private static final int TIER_SNAPSHOT = 3
+    private static final int TIER_RELEASE_CANDIDATE = 2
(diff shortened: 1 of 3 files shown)


jamesfredley:
Recreating this PR from a branch on the apache/grails-core repository instead of my fork. See the replacement PR below.

jamesfredley:
Replaced by #15696 (same change, branch hosted on apache/grails-core).

PR #15694, opened 2026-05-29
Disable Jansi in generated logback-spring.xml to fix console logging after DevTools restart

### LINKED ISSUE #15663: `<withJansi>true</withJansi>` breaks console logging after first reload in Grails 7.1.0
In Grails 7.1.0, the default `logback-spring.xml` includes:

`<withJansi>true</withJansi>`

When this setting is enabled, console logging behaves incorrectly after the first hot-reload / recompilation event in development mode.

### Observed behavior

On application startup: Logs are displayed correctly in the console

After the first code change (triggering recompilation): Only this message is shown:

`File .../UserConfigService.groovy changed, recompiling...`

After that point:
- No further console logs are displayed for subsequent changes
- The application continues running normally
- Logging output appears effectively stuck or disabled

### Expected behavior

Console logging should remain consistent across:
- application startup
- subsequent file changes

regardless of whether Jansi is enabled.

### Steps to reproduce

1. Create a plain Grails 7.1.0 app with default logback-spring.xml
2. Start the application in development mode
3. Modify a source file to trigger recompilation
4. Observe logs during the first reload
5. Modify another file
6. Observe that no further logs appear in the console

### Workaround

Disabling Jansi restores correct behavior:

`<withJansi>false</withJansi>`

or removing the setting entirely.

Notes
- The issue appears after the first recompilation event
- Startup logging is initially correct
- Logging breaks only after hot reload activity begins
- Likely interaction between Jansi ANSI console rendering and Grails dev-mode logging/reloading pipeline

### Runtime
Tested on `macOS` and `Windows 11` with `Java 25`

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

7.1.0

jamesfredley:
## Summary

This PR disables Jansi in Forge-generated `logback-spring.xml` files to prevent console logging from dying after a Spring Boot DevTools restart. It preserves development console colors without Jansi by configuring `bootRun` in the Grails Gradle plugin, rather than generating that configuration into each end-user `build.gradle`.

Generated `build.gradle` files no longer include a `bootRun` ANSI snippet; the Grails Gradle plugin now applies the `spring.output.ansi.console-available=true` system property lazily to `BootRun` tasks.

## Root cause

- Jansi's `AnsiConsole.systemInstall()` globally replaces `System.out`/`System.err` and tracks installation with a static reference counter. That state lives in the Spring Boot DevTools base classloader, so it survives restarts.
- Logback's `ConsoleAppender` calls `systemInstall()` on `start()` but never calls `systemUninstall()` on `stop()`. After the first DevTools restart, the old appender closes the shared Jansi stream while the counter stays above zero, so `systemInstall()` is a no-op and the new appender writes to a closed, stale stream. Console logging then silently stops.
- Spring Boot renders ANSI colors via its own `AnsiOutput` and `%clr` converter, which does not require Jansi or global `System.out` replacement.

## Change

- Always generate `<withJansi>false</withJansi>` and remove the OS-based Jansi toggle in `Logback`, matching the base profile skeleton.
- Configure `spring.output.ansi.console-available=true` on `BootRun` tasks in `GrailsGradlePlugin` with the existing lazy `project.tasks.withType(BootRun).configureEach` path.
- Remove the previous generated `tasks.matching { it.name == 'bootRun' }` block from Forge's `build.gradle` template.
- Add regression coverage that generated Logback config disables Jansi, generated build files do not inline the `bootRun` ANSI config, and the Grails Gradle plugin configures `BootRun` with the Spring Boot console availability property.

## Verification

- `./gradlew :grails-gradle-plugins:test --tests org.grails.gradle.plugin.core.GrailsGradlePluginToolchainSpec`
- `./gradlew :grails-forge-core:test --tests org.grails.forge.feature.logging.LogbackSpec`

Fixes #15663


grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/Logback.java
```diff
@@ -61,17 +60,10 @@ public boolean shouldApply(ApplicationType applicationType, Options options, Set
 
     @Override
     public void apply(GeneratorContext generatorContext) {
-        OperatingSystem operatingSystem = generatorContext.getOperatingSystem();
-        boolean jansi = false;
-
-        if (operatingSystem != OperatingSystem.WINDOWS) {
```
jdaugherty, line 67: This breaks color logging on other OSs.  I don't agree with this solution.

jdaugherty (changes requested):
I disagree with this PR.  It breaks the color logging.  We should only be disabling jansi if devtools is selected as the reloading, we shouldn't remove it by default.

jamesfredley:
@jdaugherty we should look into why Spring Boot deliberately does **not** enable Jansi in its own logback defaults.  This has been a landmine for Grails for multiple versions.  Color is nice, but not at the cost that has been paid, and this time this is Mac/Linux too.

The PR that updates Jansi for CLI, might improve this situation.  No sure.

jdaugherty:
> @jdaugherty we should look into why Spring Boot deliberately does **not** enable Jansi in its own logback defaults. This has been a landmine for Grails for multiple versions. Color is nice, but not at the cost that has been paid, and this time this is Mac/Linux too.
> 
> The PR that updates Jansi for CLI, might improve this situation. No sure.

Windows is horrible with utf-8 / coloring.  It's the only platform that has this issue.  For why Spring Boot doesn't use that default, it doesn't have a lot of defaults.  That doesn't stop us from having convention over configuration.  This should be a requirement of Linux/Mac OS

jamesfredley:
The `withJansi=false` change is correct and doesn't cost us colors on Mac/Linux - I checked the Spring Boot 4 source to be sure.

Colors come from Spring Boot's `%clr` converter (`AnsiOutput`/`ColorConverter`), which the template already uses via `${CONSOLE_LOG_PATTERN}`. That path is completely independent of Jansi:

- In a real terminal, `%clr` colors render on Mac/Linux whether `withJansi` is true or false - the terminal interprets ANSI natively. Jansi only ever mattered for legacy Windows consoles.
- Under `./gradlew bootRun`, `System.console()` is null, so `AnsiOutput`'s DETECT mode emits no colors *regardless of Jansi*. So the old `withJansi=true` gave us no colors under `bootRun` anyway - it only added the DevTools landmine.

So we lose nothing by dropping Jansi, and we can keep the Mac/Linux colors - by convention, even under `bootRun` - the proper way. Pushed a follow-up commit that adds this to the generated `build.gradle`:

```groovy
tasks.matching { it.name == 'bootRun' }.configureEach {
    systemProperty 'spring.output.ansi.console-available', 'true'
}
```

`spring.output.ansi.console-available=true` tells Spring Boot's `AnsiOutput` to make the per-OS decision at runtime: colors on Linux/macOS and modern Windows terminals, and automatically off on legacy Windows consoles (no raw escape codes). It's scoped to `bootRun` so it never leaks ANSI into prod/CI logs, and it has none of Jansi's global `System.out` state - so no DevTools restart breakage on any platform. Net effect: better colors on Mac/Linux than the old Jansi toggle ever delivered, with the Windows landmine gone.

jdaugherty:
fyi: matching is non-lazy gradle. We should only configure the task.  Ideally this would be done the gradle plugin and not in end user apps.

jdaugherty:
I confirmed jansi only benefits windows and only added color support for older versions.  I am inclinded to remove it since @jamesfredley is the requester.  Lets fix the color config in the gradle plugin and then we can merge.

jamesfredley:
Updated for the latest review feedback:

- Removed the generated `tasks.matching { it.name == 'bootRun' }` block from Forge's `build.gradle` template.
- Moved `spring.output.ansi.console-available=true` into `GrailsGradlePlugin` using the existing lazy `project.tasks.withType(BootRun).configureEach` configuration path.
- Updated `LogbackSpec` to assert generated apps no longer inline the `bootRun` ANSI config.
- Added a Grails Gradle plugin TestKit fixture/spec assertion that `BootRun` receives the Spring Boot console availability system property from the plugin.

Verified with:

- `./gradlew :grails-gradle-plugins:test --tests org.grails.gradle.plugin.core.GrailsGradlePluginToolchainSpec`
- `./gradlew :grails-forge-core:test --tests org.grails.forge.feature.logging.LogbackSpec`

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
```diff
@@ -17,14 +17,14 @@ specific language governing permissions and limitations
...
-@args (String projectName, String packageName, boolean jansi)
+@args (String projectName, String packageName)
 
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
     <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
 
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
-        <withJansi>@jansi</withJansi>
+        <withJansi>false</withJansi>
```
jdaugherty, line 31: Should we add a comment that this is incompatible with some reloading tools?

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -846,6 +846,7 @@ class GrailsGradlePlugin extends GroovyPlugin {
             project.tasks.withType(BootRun).configureEach { BootRun it ->
                 it.dependsOn(findMainClassTask)
                 it.mainClass.convention(GrailsGradlePlugin.getMainClassProvider(project))
+                it.systemProperty('spring.output.ansi.console-available', 'true')
```
jdaugherty, line 854: Shouldn't this be where  OS != Windows?

jdaugherty (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
jamesfredley: Good call - added an explanatory comment above `<withJansi>false</withJansi>` in the generated `logback-spring.xml`. It notes that Jansi is intentionally disabled because it replaces `System.out` globally, which is incompatible with reloading tools (Spring Boot DevTools) and breaks console logging after a restart, and that colors come from Spring Boot's `%clr` converter via `CONSOLE_LOG_PATTERN` so Jansi isn't needed. Done in a392fc918c.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jamesfredley: It doesn't need the `OS != Windows` guard - setting `spring.output.ansi.console-available=true` is not a force-on. Spring Boot's `AnsiOutput` in DETECT mode (the default) still gates Windows out internally: `detectIfAnsiCapable()` ends with `return !(OPERATING_SYSTEM_NAME.contains(\win\))` (verified in Spring Boot 3.5.x `AnsiOutput`). So on legacy Windows consoles no raw ANSI escapes are emitted regardless of this property; the only thing it does is tell AnsiOutput a console is available under `bootRun` (where `System.console()` is null), enabling colors on macOS/Linux and modern terminals. Setting it unconditionally is therefore correct, and an OS check here would just duplicate Spring Boot's own logic. I've added a code comment explaining this above the line in a392fc918c.

PR #15699, opened 2026-05-29
Fix bindData binding id/version when domain extends a @DirtyCheck base (#15681)

### LINKED ISSUE #15681: GORM bindData incorrectly binds id when domain class extends a @DirtyCheck abstract base class
# Bug Report — `bindData` incorrectly binds `id` when domain class extends a `@DirtyCheck` abstract base class

## Expected Behavior

When calling `bindData(domainObject, params)` on a GORM domain class, the `id` property should **not** be bound from the input, as documented in the Grails Data Binding guide:

> The "id" and "version" properties of a domain class are never bound by default.

---

## Actual Behaviour

When the domain class extends an abstract class located in `src/main/groovy/` that is annotated with `@DirtyCheck`, the `id` property **is** bound from the input.

The failing test in this project demonstrates it:

```
Condition not satisfied:

response.text == 'id=null'
|        |    |
|        |    false
|        id=99
```

`params.id = '99'` was silently applied to the new domain object.

---

## Steps To Reproduce

1. Create an abstract class in `src/main/groovy/` annotated with `@DirtyCheck`:

```groovy
// src/main/groovy/databinding/bug/demo/AbstractBase.groovy
package databinding.bug.demo

import grails.gorm.dirty.checking.DirtyCheck

@DirtyCheck
@CompileStatic
abstract class AbstractBase {
    String name
}
```

2. Create a domain class in `grails-app/domain/` extending that abstract class:

```groovy
// grails-app/domain/databinding/bug/demo/MyDomain.groovy
package databinding.bug.demo

class MyDomain extends AbstractBase {
    static constraints = {
        name nullable: false
    }
}
```

3. In a controller action call `bindData` with params that include an `id`:

```groovy
def domain = new MyDomain()
bindData(domain, params)   // params contains id='99'
assert domain.id == null   // FAILS — domain.id is 99
```

4. Run `grails test-app` — the single test in this project fails, proving the bug.

---

## Root Cause

`@DirtyCheck` on `AbstractBase` causes GORM's `DirtyCheckingTransform` to implement the
`DirtyCheckable` trait on `AbstractBase`, which injects `id` and `version` as concrete fields into the compiled class — even though `AbstractBase` is not itself a domain class.

When `DefaultASTDatabindingHelper` builds the `$defaultDatabindingWhiteList` for `AbstractBase`, `GrailsASTUtils.isDomainClass()` returns `false` (the file is in `src/main/groovy/`, not `grails-app/domain/`). Because `isDomainClass` is `false`, `shouldFieldBeInWhiteList` skips the `DOMAIN_CLASS_PROPERTIES_TO_EXCLUDE_BY_DEFAULT` check entirely (the `else if (isDomainClass)` branch). As a result, `id` and `version` are included in `AbstractBase`'s whitelist.

When `DefaultASTDatabindingHelper` subsequently processes `MyDomain`, it inherits `AbstractBase`'s whitelist into `bindablePropertyNames`. The field loop then checks:

```java
if (bindablePropertyNames.contains(fieldName)
        || shouldFieldBeInWhiteList(fieldNode, fieldsInTransientsList, isDomainClass)) {}
```

Because `id` is already in `bindablePropertyNames`, the `||` short-circuits and
`shouldFieldBeInWhiteList` — which would correctly return `false` for `id` on a domain class —
is never consulted. `id` and `version` are therefore included in `MyDomain.$defaultDatabindingWhiteList`.

---

## Workaround

Exclude `id` explicitly until the bug is fixed:

```groovy
bindData(domainObject, params, [exclude: ['id']])
```

---

## Environment Information

- Operating System: Linux
- JDK Version: 17

---

## Example Application

[This repository](https://github.com/sbglasius/grails-gorm-data-bind-bug) — run `grails test-app` to see the single failing test.

---

## Version

7.1.1

jamesfredley:
Fixes #15681

## Problem

The Grails data binding guide states:

> The "id" and "version" properties of a domain class are never bound by default.

However, when a domain class extends an abstract base class that is **not** itself recognised as a domain class - for example an abstract `@DirtyCheck` class in `src/main/groovy` onto which GORM injects `id` and `version` (via `getFurthestUnresolvedParent()` during joint compilation) - `bindData` (and the map constructor) would bind `id` and `version` from the input.

## Root cause

In `DefaultASTDatabindingHelper.getPropertyNamesToIncludeInWhiteList()`:

1. The abstract base's whitelist is computed with `isDomainClass = false` (it has no `@Entity`/`@jakarta.persistence.Entity` annotation and is not under `grails-app/domain`). Because of that, `shouldFieldBeInWhiteList()` skips the `DOMAIN_CLASS_PROPERTIES_TO_EXCLUDE_BY_DEFAULT` check, so `id` and `version` are included in the base class' whitelist.
2. When the domain subclass is processed, it inherits the parent whitelist into `bindablePropertyNames`.
3. The field loop then short-circuits on `bindablePropertyNames.contains(fieldName)`, so `shouldFieldBeInWhiteList()` - which would correctly return `false` for `id`/`version` on a domain class - is never consulted. As a result `id` and `version` end up in the subclass' `$defaultDatabindingWhiteList`.

## Fix

When the current class is a domain class, the default-excluded properties (`id`, `version`, `dateCreated`, `lastUpdated`) are no longer propagated from the inherited parent whitelist. An explicit `bindable: true` constraint declared on the domain class itself still re-enables binding, because constraints are processed after the inherited names are filtered.

## Tests

Added a regression test to `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec` that binds `id`, `version` and a regular property to a domain class extending a `@DirtyCheck` abstract base, asserting that `id`/`version` are not bound while the regular property is. The test fails before the fix and passes after.

Verified no regressions across the existing data-binding suites (`grails-test-suite-web` binding/command-object/bindData specs and the persistence `GrailsWebDataBinderSpec`).


sbglasius (approved):

grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java
```diff
@@ -145,9 +145,20 @@ private Set<String> getPropertyNamesToIncludeInWhiteList(final SourceUnit source
         final Set<String> propertyNamesToIncludeInWhiteList = new HashSet<>();
         final Set<String> unbindablePropertyNames = new HashSet<>();
         final Set<String> bindablePropertyNames = new HashSet<>();
+        final boolean isDomainClass = GrailsASTUtils.isDomainClass(classNode, sourceUnit);
         if (!classNode.getSuperClass().equals(new ClassNode(Object.class))) {
             final Set<String> parentClassPropertyNames = getPropertyNamesToIncludeInWhiteListForParentClass(sourceUnit, classNode.getSuperClass());
-            bindablePropertyNames.addAll(parentClassPropertyNames);
+            for (final String parentPropertyName : parentClassPropertyNames) {
+                // The id, version, dateCreated and lastUpdated properties of a domain class are never bound by default.
```
jdaugherty: We should simplify the comment or just remove it.  I think the code is self-explanatory

grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java
```diff
@@ -145,9 +145,20 @@ private Set<String> getPropertyNamesToIncludeInWhiteList(final SourceUnit source
...
+                // A parent class may legitimately include these in its own whitelist when it is not itself recognised as
+                // a domain class (for example an abstract @DirtyCheck base class in src/main/groovy onto which GORM injects
+                // id and version). Such properties must not be inherited into a domain class' whitelist, otherwise the
+                // exclusion performed by shouldFieldBeInWhiteList is bypassed via the inherited bindable property names.
+                if (isDomainClass && DOMAIN_CLASS_PROPERTIES_TO_EXCLUDE_BY_DEFAULT.contains(parentPropertyName)) {
```
jdaugherty, line 154: What about command objects?  Isn't this used for this too?

jdaugherty (commented):
The comment is minor clean up and I'm pretty sure the changed class is specific to domains only, but we should confirm that.

matrei (approved):

grails-test-examples/gorm/src/integration-test/groovy/gorm/DirtyCheckBindingSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+import spock.lang.Tag
+
+import org.apache.grails.testing.http.client.HttpClientSupport
+
+/**
+ * Functional test reproducing issue 15681 end-to-end: a real Grails application binds request parameters
+ * (including {@code id} and {@code version}) to a domain class that extends an abstract {@code @DirtyCheck}
+ * base. The framework must not bind {@code id} or {@code version} by default.
+ */
+@Integration(applicationClass = Application)
```
matrei: I think we can remove the `(applicationClass = Application)`?

grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java
jamesfredley: Good call - simplified to a concise two-liner that states only the *why*: a domain class never binds its special properties by default, so they shouldn't be inherited from a non-domain parent such as a `@DirtyCheck` base. Done in bca8bdcc48.

grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java
jamesfredley: Good question - yes, command objects do flow through `DefaultASTDatabindingHelper`. It's invoked from two transformers: `ControllerActionTransformer` (for command objects, i.e. controller action parameters) and `ControllerDomainTransformer` (for domain classes).

The change is safe for command objects because the new filter is guarded by `isDomainClass`, and `GrailsASTUtils.isDomainClass()` returns `false` for them - a command object has neither `@Entity`/`@jakarta.persistence.Entity` nor a source path under `grails-app/domain/`.

This is consistent with the existing design: the `DOMAIN_CLASS_PROPERTIES_TO_EXCLUDE_BY_DEFAULT` exclusion (`id`/`version`/`dateCreated`/`lastUpdated`) was already gated on `isDomainClass` everywhere it is applied - in `shouldFieldBeInWhiteList` (the field loop) and in the setter-derived branch. So a command object with a field named `id` or `version` still gets it included in its whitelist, exactly as before. No behavior change for command objects; the fix only affects actual GORM domain classes.

grails-test-examples/gorm/src/integration-test/groovy/gorm/DirtyCheckBindingSpec.groovy
jamesfredley: Removed - it's bare `@Integration` now (bca8bdcc48). `gorm.Application` is auto-discovered (same package, extends `GrailsAutoConfiguration`), and 11 of the existing integration tests in this module already use bare `@Integration`, so this matches the module's convention.

grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java
jdaugherty: Apologies, I wasn't asking if this would skip command objects. I was asking if this same issue exists with them.

sbglasius: @jdaugherty In a command object, you want to be able to bind `id` etc... so it should not skip those (and doesn't)

jdaugherty: @sbglasius  I couldn't recall if there were existing restrictions on it.  I agree with you, but was asking that we confirm that behavior and ensure no regressions here.

jdaugherty (approved):

PR #15700, opened 2026-05-29
Fix constraints overwriting mapping index configuration (#15680)

### LINKED ISSUE #15680: GORM-MongoDB: constraints unique:true suppresses mapping index creation with indexAttributes
## Summary

When a domain class field has both `unique: true` in `static constraints` and `index: true, indexAttributes: [unique: true]` in `static mapping`, the MongoDB index is silently not created. Removing the constraint allows the index to be created normally.

## Steps to Reproduce

1. Create a domain class with the following:

```groovy
class Category {
    String name

    static constraints = {
        name unique: true
    }

    static mapWith = 'mongo'
    static mapping = {
        name index: true, indexAttributes: [unique: true]
    }
}
```

2. Start the application and insert data via BootStrap
3. Query MongoDB for the collection's indexes:

```javascript
db.category.getIndexes()
```

**Result:** Only `_id_` index exists. The `name_1` unique index is missing.

4. Remove the constraint:

```groovy
static constraints = {
    name()  // removed unique: true
}
```

5. Restart and check indexes again.

**Result:** `name_1` unique index is now created correctly.

## Expected Behavior

The `index: true, indexAttributes: [unique: true]` declaration in the mapping block should create the MongoDB index regardless of whether the constraints block also declares `unique: true`. The constraint and the mapping serve different purposes — constraints validate at the GORM layer, while `indexAttributes` creates the database-level index.

## Actual Behavior

The presence of `unique: true` in the constraints block silently prevents `initializeIndices()` from creating the MongoDB index declared in the mapping block. No error or warning is logged.

## Additional Context

- Fields with plain `index: true` (no `indexAttributes`) are created correctly even with constraint declarations
- Fields with `index: true, indexAttributes: [unique: true]` but WITHOUT a constraint `unique: true` are created correctly
- The `compoundIndex([name: 1, indexAttributes: [unique: true]])` syntax is not affected — it always creates the index regardless of constraints
- Other domain classes (Country, Type) with the same `index: true, indexAttributes: [unique: true]` mapping but WITHOUT `unique: true` in constraints create their indexes correctly

## Environment

- Grails: 7.2.0-SNAPSHOT
- GORM MongoDB: 7.2.0-SNAPSHOT (`grails-data-mongodb-core`)
- MongoDB: embedded (de.flapdoodle)
- Java: 25

codeconsole:
## Summary

Fixes #15680.

When a domain property has both `index: true, indexAttributes: [unique: true]` in `static mapping` and `unique: true` in `static constraints`, the constraint evaluation overwrites the mapping-configured property, so the database-level unique index is silently never created.

## Root Cause

`AbstractGormMappingFactory.createMappedForm()` evaluates both closures through the same `DefaultMappingConfigurationBuilder`:
1. The mapping closure (`name index: true, indexAttributes: [unique: true]`) goes through `builder.invokeMethod`, which stores a `MongoAttribute` with `index=true, indexAttributes={unique:true}` in `builder.properties['name']`.
2. The constraint closure (`name unique: true`) goes through `Entity.methodMissing`, which creates a SEPARATE instance in `target.propertyConfigs['name']` containing only `unique=true`.
3. `builder.getProperties()` then does `properties.putAll(target.propertyConfigs)`, overwriting the mapping-set property with the constraint-only version. `index` reverts to `false` and `indexAttributes` becomes `null`.

`MongoDatastore.initializeIndices()` reads `property.getMapping().getMappedForm().isIndex()`, which now returns `false`, so the MongoDB index is silently never created.

## Fix

Change `getProperties()` to use `putIfAbsent` semantics — properties already configured by the mapping closure are preserved when the constraint closure provides a conflicting entry.

## Test

`DefaultMappingConfigurationBuilderSpec` covers the scenario directly: a property simulating GORM's `Entity.methodMissing` path against a builder that already has a mapping-configured property. The test FAILS before the fix and PASSES after.

## Verified

Tested against a real Pixoto3 Grails 7.2 app with embedded MongoDB:
- Before: `Category.name` declared `index: true, indexAttributes: [unique: true]` in mapping + `unique: true` in constraints → only `_id_` index in MongoDB
- After: same declarations → `name_1` unique index correctly created in MongoDB

## Backward Compatibility

Behavior change: if a user previously relied on constraints overwriting mapping (unlikely — that's the bug), they would observe their mapping-configured property surviving instead. The opposite direction (constraint values being applied on top of an existing mapping property) was already happening through `builder.invokeMethod` reusing the existing instance, so the typical "add a constraint to a mapped property" flow is unaffected.


grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/groovy/DefaultMappingConfigurationBuilder.groovy
```diff
@@ -47,7 +47,11 @@ class DefaultMappingConfigurationBuilder implements MappingConfigurationBuilder
 
     Map<String, Property> getProperties() {
         if (!target.propertyConfigs.isEmpty()) {
-            properties.putAll(target.propertyConfigs)
+            for (Map.Entry entry : target.propertyConfigs.entrySet()) {
+                if (!properties.containsKey(entry.key)) {
+                    properties.put(entry.key, entry.value)
+                }
```
sbglasius, line 80: Would it make sense to make a log statement if the `properties` actually contains the `entry` and that way give the developer a chance to know that there is a conflicting property?

codeconsole:
## Updated based on review

Pushed an amended commit with the following tweaks (verified by an independent multi-agent review of the original PR):

### Test coverage improved

The original test (now renamed *"constraints closure does not overwrite mapping properties (standard DSL flow)"*) passed even with the fix reverted — both closures route through `builder.invokeMethod` and reuse the same instance, so `target.propertyConfigs` stays empty for that flow and the changed merge path is never exercised.

The new spec adds **"mapping property in builder.properties is preserved when target.propertyConfigs has a different instance for the same key"** which:
- Populates `builder.properties['name']` via the mapping closure (the normal `invokeMethod` path)
- Independently populates `entity.propertyConfigs['name']` with a separate instance (simulating the constraint-evaluator / `Entity.property(name, Map)` direct-API path that actually triggers the bug in a full Grails app)
- Asserts the builder's mapping-configured instance is the one returned by `getProperties()`

**Verified:** this new test FAILS against the pre-fix `properties.putAll(target.propertyConfigs)` code and PASSES with the `putIfAbsent` fix. The TCK-based MongoDB integration spec was attempted but does not exercise the bug — the TCK uses `DefaultGrailsApplication` directly and skips the constraint-evaluator wiring that populates `propertyConfigs` in a real Grails-app boot, so it cannot reproduce the failure.

Also added **"target.propertyConfigs entries fill in for keys the mapping closure never touched"** to lock in that constraint-only configuration for properties the mapping never mentions still propagates correctly.

### Code comment added

Added javadoc to `getProperties()` documenting:
- The precedence rule (mapping-side instance wins when both maps have the same key)
- Why the asymmetry exists (dual-store legacy)
- A link back to this issue for the structural cleanup that would let the method go away

### Known follow-up

The dual-store design (`builder.properties` vs `Entity.propertyConfigs`) is the underlying root cause — this PR papers over a symptom. A future major-release cleanup should collapse the two into a single canonical store. Tracking would happen in a separate issue.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/groovy/DefaultMappingConfigurationBuilder.groovy
codeconsole: Good call — added a DEBUG-level log statement when a key collision is detected. Kept it at debug because in a real Grails-app boot this fires for every property that appears in both a `static mapping` and `static constraints` block (which is most domains with any field validation alongside any explicit mapping), so info-level would be too noisy for a steady-state condition. Anyone investigating an unexpected precedence outcome can flip the logger for `org.grails.datastore.mapping.config.groovy.DefaultMappingConfigurationBuilder` to DEBUG and see exactly which keys collided.

sbglasius (approved):

jdaugherty:
The code style & licensing issues need addressed

jdaugherty (approved):

Issue #9837, opened 2016-03-30
[3.1.4] - Test - VerifyError occours when using variable of an data table inside a closure


marcioaug:
| Grails Version: 3.1.4
| Groovy Version: 2.4.6
| JVM Version: 1.8.0_77

``` groovy
    void "Test the homepage"() {
        expect:
	
        def stuffs = [[code: 1, letter: 'a'], [code: 2, letter: 'b'], [code: 3, letter: 'c']]

        stuffs.find { it.code == code }?.letter == letter

        where:

        code | letter
        1    | 'a'
        2    | 'b'
        3    | 'c'
    }
```
It works in Grails 3.1.1, but in 3.1.4:
```
initializationError

java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    functional/FunctionalTestSpecSpec.$tt__$spock_feature_1_0(Ljava/lang/Object;Ljava/lang/Object;Lorg/springframework/transaction/TransactionStatus;)V @302: invokespecial
  Reason:
    Type 'java/lang/Object' (current frame, stack[24]) is not assignable to 'groovy/lang/Reference'
  Current Frame:
    bci: @302
    flags: { }
    locals: { 'functional/FunctionalTestSpecSpec', 'java/lang/Object', 'java/lang/Object', 'org/springframework/transaction/TransactionStatus', '[Lorg/codehaus/groovy/runtime/callsite/CallSite;', 'java/lang/Object', 'java/util/List' }
    stack: { 'org/spockframework/runtime/ValueRecorder', 'java/lang/String', integer, integer, null, 'org/spockframework/runtime/ValueRecorder', integer, 'org/spockframework/runtime/ValueRecorder', integer, 'org/codehaus/groovy/runtime/callsite/CallSite', 'org/spockframework/runtime/ValueRecorder', integer, 'java/lang/Class', 'java/lang/Object', 'java/lang/String', '[Ljava/lang/Object;', '[Ljava/lang/Object;', integer, 'org/spockframework/runtime/ValueRecorder', integer, uninitialized 295, uninitialized 295, 'functional/FunctionalTestSpecSpec', 'functional/FunctionalTestSpecSpec', 'java/lang/Object' }
  Bytecode:
    0x0000000: b800 233a 0419 0412 ee32 12f0 b900 3002
...
at java.lang.Class.getDeclaredFields0(Native Method)
	at java.lang.Class.privateGetDeclaredFields(Class.java:2583)
[...]

Issue #9646, opened 2016-02-06
FunctionalSpec with @Rollback and assert in closure causes compilation error "The current scope already contains a variable of the name $spock_valueRecorder"


droggo:
Having simple functional spec like:
```
package error

import grails.test.mixin.integration.Integration
import grails.transaction.Rollback
import geb.spock.GebSpec

@Integration
@Rollback
class WithAssertFunctionalSpec extends GebSpec {
	void 'any spec method'(){
		when:
		List a = [1, 2, 3]

		then:
		a.each {
			assert it < 4
		}
	}
}
```
You will get compilation error:
```
:compileIntegrationTestGroovy
startup failed:
path\to\app\src\integration-test\groovy\error\WithAssertFunctionalSpec.groovy: -1: The current scope already contains a variable of the name $spock_valueRecorder
 @ line -1, column -1.
1 error
```

Error will disappear if either `@Rollback` annotation or `assert` inside of closure is not used. This is similar issue to https://github.com/spockframework/spock/issues/491

We have it on Grails 3.1.1

droggo:
This happens also for integration spec which extends standard `spock.lang.Specification` instead of Geb

hbogaards:
I got the same compiler error using Grails 3.1.1

Is this also an issue in the master (3.1.x) branch?

graemerocher:
it is fixed in the 3.0.x branch and will be merged into 3.1.x and available in 3.0.15 and 3.1.2

PR #15703, opened 2026-05-29
Update Spock to 2.4-groovy-4.0 and align test dependencies

jamesfredley:
## What

Upgrades Spock from `2.3-groovy-4.0` to the latest stable **`2.4-groovy-4.0`** on the `8.0.x` branch, pulls in the new dependency Spock 2.4 requires, and adapts the framework tests to Spock 2.4's stricter behavior.

| Dependency | Before | After | Notes |
|---|---|---|---|
| `spock.version` | `2.3-groovy-4.0` | `2.4-groovy-4.0` | latest stable for Groovy 4.0 (there is no 2.5) |
| `gradle-spock.version` | `2.3-groovy-4.0` | `2.4-groovy-4.0` | build-logic + grails-gradle |
| `grails-forge` `spockVersion` | `2.3-groovy-4.0` | `2.4-groovy-4.0` | grails-forge's own tests |
| `geantyref.version` | - | `1.3.16` | **new** - required by Spock 2.4 mocking (see below) |

Evaluated but deliberately left unchanged:

- **Selenium** stays at `4.38.0`. Bumping it drags OpenTelemetry above Spring Boot 4.0.6's managed `1.55.0` (selenium `4.38.0` → otel `1.55.0`, but `4.44.0` → `1.62.0`), which fails the `validateDependencyVersions` gate. `4.38.0` is the release whose OpenTelemetry matches Boot, so it is the correct pin.
- **geb-spock** (`8.0.1`) is already the latest release.
- **JUnit** (`6.0.3`) already matches Boot's managed `junit-jupiter`; **byte-buddy** (`1.18.8`) keeps its existing override; **objenesis** (`3.4`) is already latest.

## New `geantyref` dependency

Spock 2.4 replaced its bundled gentyref code with `io.leangen.geantyref:geantyref:1.3.16`, declared in `spock-core` as a required `runtime` dependency and used by the mocking engine (`GenericTypeReflectorUtil`). Modules that declare `spock-core` with `transitive = false` (and therefore re-add `byte-buddy`/`objenesis` for mocking) must also add `geantyref`, otherwise `Mock()`/`Stub()`/`Spy()` fail at runtime with:

```
java.lang.NoClassDefFoundError: io/leangen/geantyref/GenericTypeReflector
```

This PR adds `geantyref` to the BOM and declares it next to `byte-buddy`/`objenesis` in every module that strips Spock's transitives.

**Generated applications need no change.** grails-forge and grails-profiles declare `spock-core` transitively, so `geantyref` arrives automatically at the version managed by `grails-bom` - no generator or skeleton template change is required.

## Spock 2.4 test compatibility

Spock 2.4 tightened two behaviors that existing tests relied on:

- `JsonViewTemplateResolverSpec` stubbed the **final** methods `getRequest()`/`getResponse()` on a `GrailsWebRequest` mock. Spock 2.3 silently ignored final-method stubs; Spock 2.4's byte-buddy mock maker fails fast with `InvalidSpecException`. The stubs were dead (the test passes the request and response to `resolveView()` explicitly), so they were removed.
- `TransactionalTransformSpec` reflectively invoked the `@Rollback`-transformed feature methods directly, outside Spock's runner. Spock 2.4 throws `IllegalStateException` ("Cannot request current iteration") when a closure that references data variables or contains conditions runs without an active iteration. The direct-invocation blocks were removed; the `GroovyShell` compilation plus the transformed method-signature assertions still cover the #9837 and #9646 regressions.
- `IOUtilsSpec` (grails-gradle-model) asserts the resolved `spock-core` jar name; updated `2.3-groovy-4.0` → `2.4-groovy-4.0`.

Stale Spock version references in `AGENTS.md` and a `settings.gradle` comment were also refreshed.

## Verification

`grails-bootstrap`, `grails-web-mvc`, `grails-views-gson`, and `grails-datamapping-core` tests pass locally, along with `:grails-gradle-model:test` and `validateDependencyVersions`. CI exercises the full module matrix, the separate `grails-gradle` build, and `grails-forge`.


jdaugherty:
Shouldn't we update 7.0.x since this is targeting Groovy 4?

jdaugherty (approved):

jamesfredley:
**Re: the new `geantyref` dependency**

Spock 2.4 externalized its bundled gentyref into `io.leangen.geantyref:geantyref:1.3.16` and declares it as a **required `runtime` dependency** of `spock-core` (used by the mock engine's `GenericTypeReflectorUtil`). Without it, `Mock()`/`Stub()`/`Spy()` fail at runtime with `NoClassDefFoundError: io/leangen/geantyref/GenericTypeReflector`.

The reason it has to be added in ~34 modules is our existing `spock-core { transitive = false }`, which strips *all* of Spock's transitives (groovy/junit/hamcrest/geantyref) on purpose - to keep Spock from pinning the Groovy/JUnit versions, which matters for the Groovy joint-validation build. So `geantyref` is just the first *required* Spock dep that isolation removes.

Note it's different from `byte-buddy`/`objenesis`: those aren't in Spock's POM at all (optional, delivered via Gradle module metadata), so they'd never come transitively regardless - that's why they were always declared by hand. `geantyref` *is* a required POM dep, so it only needs re-adding because of `transitive = false`. Modules that take Spock transitively (e.g. grails-forge, which only excludes `groovy-all`) get it automatically and need no change.

Alternative considered: swapping `transitive = false` for a targeted `exclude group: 'org.apache.groovy'` would let `geantyref` flow in by itself, but that changes a long-standing pattern across 34 modules and pulls Spock's junit/hamcrest back in - better as a separate cleanup than part of this Spock bump.

matrei:
> **Re: the new `geantyref` dependency**
> 
> Spock 2.4 externalized its bundled gentyref into `io.leangen.geantyref:geantyref:1.3.16` and declares it as a **required `runtime` dependency** of `spock-core` (used by the mock engine's `GenericTypeReflectorUtil`). Without it, `Mock()`/`Stub()`/`Spy()` fail at runtime with `NoClassDefFoundError: io/leangen/geantyref/GenericTypeReflector`.
> 
> The reason it has to be added in ~34 modules is our existing `spock-core { transitive = false }`, which strips _all_ of Spock's transitives (groovy/junit/hamcrest/geantyref) on purpose - to keep Spock from pinning the Groovy/JUnit versions, which matters for the Groovy joint-validation build. So `geantyref` is just the first _required_ Spock dep that isolation removes.
> 
> Note it's different from `byte-buddy`/`objenesis`: those aren't in Spock's POM at all (optional, delivered via Gradle module metadata), so they'd never come transitively regardless - that's why they were always declared by hand. `geantyref` _is_ a required POM dep, so it only needs re-adding because of `transitive = false`. Modules that take Spock transitively (e.g. grails-forge, which only excludes `groovy-all`) get it automatically and need no change.
> 
> Alternative considered: swapping `transitive = false` for a targeted `exclude group: 'org.apache.groovy'` would let `geantyref` flow in by itself, but that changes a long-standing pattern across 34 modules and pulls Spock's junit/hamcrest back in - better as a separate cleanup than part of this Spock bump.

Yes, it seems GeantyRef code was previously vendored by Spock, but is now using the upstream library instead:
https://github.com/spockframework/spock/pull/1743

It seems we are not consistent in excluding transitive dependencies from Spock (see `grails-async-core` for example).
I actually don't think it's needed to exclude Spock transitives as we are declaring our own dependency on Groovy and get the version from the bom.
On the joint workflow we switch the version in the bom and it is picked up as it should even without excluding Spock transitives: https://develocity.apache.org/s/iasnjafc4pxlo/dependencies?toggled=W1s0XSxbNCwxNF0sWzQsMTQsWzE3NjddXV0

matrei:
I have tested locally with removing all `transitive = false` from Spock and removing explicit dependency on  'io.leangen.geantyref:geantyref' in this branch. The tests run just fine. I think we should clean this up.

matrei (approved):

PR #15705, opened 2026-05-30
Pin SiteMesh 3 to 3.2.3 release and drop snapshot repo resolution

codeconsole:
## What

SiteMesh `3.2.3` is now published to Maven Central, so the SiteMesh 3 dependencies no longer need to resolve from the Sonatype `maven-snapshots` repository.

## Changes

- **`dependencies.gradle`**: bump `starter-sitemesh.version` and `spring-webmvc-sitemesh.version` from `3.2.3-SNAPSHOT` to `3.2.3`.
- **`GrailsRepoSettingsPlugin`**: remove the `org.sitemesh` snapshot `includeVersionByRegex` from the `central.sonatype.com/repository/maven-snapshots` repository (both the `pluginManagement` and `dependencyResolutionManagement` blocks). The `cloud.wondrify` (asset-pipeline) snapshot include is retained.

## Notes

- SiteMesh 2 (`opensymphony:sitemesh:2.6.0`, used by `grails-layout`) is unaffected.
- Verified both release artifacts exist on Maven Central:
  - `org.sitemesh:spring-boot-starter-sitemesh:3.2.3`
  - `org.sitemesh:spring-webmvc-sitemesh:3.2.3`
- This also unblocks the `validateNoSnapshotDependencies` release checks for SiteMesh.

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsRepoSettingsPlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsRepoSettingsPlugin.groovy
@@ -47,7 +47,6 @@ class GrailsRepoSettingsPlugin implements Plugin<Settings> {
                     url = 'https://central.sonatype.com/repository/maven-snapshots'
                     content {
                         it.includeVersionByRegex('cloud[.]wondrify.*', '.*', '.*-SNAPSHOT')
-                        it.includeVersionByRegex('org[.]sitemesh.*', '.*', '.*-SNAPSHOT')
                     }
                     mavenContent {
                         it.snapshotsOnly()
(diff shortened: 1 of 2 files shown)


sbglasius (approved):

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsRepoSettingsPlugin.groovy
```diff
@@ -47,7 +47,6 @@ class GrailsRepoSettingsPlugin implements Plugin<Settings> {
                     url = 'https://central.sonatype.com/repository/maven-snapshots'
                     content {
                         it.includeVersionByRegex('cloud[.]wondrify.*', '.*', '.*-SNAPSHOT')
-                        it.includeVersionByRegex('org[.]sitemesh.*', '.*', '.*-SNAPSHOT')
```
jdaugherty, line 50: If we think we need to run with snapshots in the future, we should just leave this.  It won't use it because we don't have a dependency as snapshot after this change.

jdaugherty (approved):

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsRepoSettingsPlugin.groovy
```diff
@@ -92,7 +91,6 @@ class GrailsRepoSettingsPlugin implements Plugin<Settings> {
...
```
jdaugherty, line 95: Same as the other comment

PR #15710, opened 2026-06-01
Enable SiteMesh 3 on the 8.x (Spring Boot 4) line

codeconsole:
## What

Re-enables **SiteMesh 3** on the 8.x (Spring Boot 4) line and makes it the default GSP layout decorator for generated applications, while keeping the legacy SiteMesh 2 `grails-layout` available as a mutually exclusive alternative.

| Area | Change |
|------|--------|
| `dependencies.gradle` | `starter-sitemesh` / `spring-webmvc-sitemesh` `3.2.2` → `3.3.0-SNAPSHOT` (Boot 4 line) |
| `settings.gradle` | Re-include `grails-sitemesh3` and `grails-gsp-spring-boot` |
| `grails-gsp/grails-sitemesh3/build.gradle` | Re-enable the publish plugin |
| `gradle/publish-root-config.gradle` | Restore `grails-sitemesh3` to the published projects so it is BOM-managed |
| `grails-forge` | New `GspLayout` one-of group: `Sitemesh3` (default) + `GrailsLayout` (opt-in), mutually exclusive; `GrailsGsp` no longer hardcodes the layout dep |
| `grails-profiles/web/profile.yml` | Web profile uses `grails-sitemesh3` instead of `grails-layout` |
| `grails-forge` + `grails-shell-cli` | Generated SNAPSHOT apps add the Sonatype `org.sitemesh` snapshot repo so `3.3.0-SNAPSHOT` resolves |

## Why

SiteMesh `3.2.x` targets **Spring Boot 3**; the 8.x line runs on **Spring Boot 4**, which is why these modules were commented out ("does not support boot 4 at this time"). SiteMesh `3.3.0` is the Boot 4 line (published as `3.3.0-SNAPSHOT`). The two `org.sitemesh.*:*-SNAPSHOT` exemptions already exist in `GrailsRepoSettingsPlugin`, so grails-core itself resolves it; the only missing piece for **generated** apps was a snapshot repo, added here (SNAPSHOT-gated, filtered to `org.sitemesh`).

## Layout choice (both available, never both)

`Sitemesh3` and `GrailsLayout` share a `GspLayout` one-of feature class, so `OneOfFeatureValidator` rejects selecting both. `Sitemesh3` is a `DefaultFeature` (auto-applied to web apps unless `grails-layout` is explicitly chosen). In the **profiles** generator the web profile simply defaults to `grails-sitemesh3` (one decorator, never both).

## Verification (JDK 21 / Boot 4)

- `:grails-sitemesh3:test`, `:grails-sitemesh3:classes`/`:jar`, `:grails-gsp-spring-boot:classes` — ✅
- `:grails-shell-cli:compileGroovy` — ✅
- forge `GspLayoutSpec` (sitemesh3 default for WEB/WEB_PLUGIN; grails-layout replaces it; both → throws; generated SNAPSHOT app exposes the org.sitemesh snapshot repo) — ✅
- forge `GrailsGspSpec` — ✅ no regression

## Notes / follow-ups

- The Sonatype snapshot repo is added to generated apps **only for SNAPSHOT builds**; a released `org.sitemesh:3.3.0` must be pinned before 8.x GA, otherwise released apps cannot resolve it.
- `grails-gsp-spring-boot` is re-enabled as a buildable module but remains unpublished (only consumer is the still-disabled `gsp-spring-boot` test example).
- New user-facing forge feature `grails-layout` — a docs note in `grails-doc` is a likely follow-up.
- ⚠️ Merge-up: #15705 pinned 7.2.x to the `3.2.3` release (Boot 3); when 7.2.x merges into 8.0.x it will conflict with `3.3.0-SNAPSHOT` — the 8.x Boot 4 value should win.


grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/sitemesh3/Sitemesh3.java
```diff
@@ -22,14 +22,20 @@
...
 
-    public Sitemesh3() {
-    }
+/**
+ * Default GSP layout decorator, backed by SiteMesh 3 ({@code grails-sitemesh3}).
+ * Applied automatically to web applications unless another {@link GspLayout}
+ * (e.g. {@code grails-layout}) is explicitly selected.
+ */
+@Singleton
+public class Sitemesh3 extends GspLayout implements DefaultFeature {
```
jdaugherty: I think making this the default & updating to sitemesh 3 should be separate PRs

jdaugherty (changes requested):
I don't see a removal of `boot4-disabled-integration-test-config.gradle` - which I expected with this PR.  I believe the only remaining use of this was sitemesh related.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/GrailsGsp.java
```diff
@@ -113,7 +113,7 @@ public void apply(GeneratorContext generatorContext) {
         if (!generatorContext.isFeaturePresent(Sitemesh3.class)) {
             generatorContext.addDependency(Dependency.builder()
                     .groupId("org.apache.grails")
-                    .artifactId("grails-layout")
+                    .artifactId("grails-sitemesh3")
```
jdaugherty: This would make sitemesh3 still a default, yes?  Wouldn't this need to remain layout?

jdaugherty: @codeconsole if we can get this change reverted, I'm good to merge this.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/view/GrailsGsp.java
```diff
@@ -110,12 +109,6 @@ public void apply(GeneratorContext generatorContext) {
                 .groupId("org.apache.grails")
                 .artifactId("grails-gsp")
                 .implementation());
-        if (!generatorContext.isFeaturePresent(Sitemesh3.class)) {
```
jdaugherty, line 113: It looks like this is still being removed so layout is not remaining as the default.

jdaugherty (changes requested):
@codeconsole I think the changes to GrailsGsp need reverted until we do the other PR promoting sitemesh 3 as the default.

jdaugherty (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
```diff
@@ -66,6 +66,17 @@ static Set<GradleRepository> getDefaultRepositories(String grailsVersion) {
                 ),
                 List.of(VersionType.SNAPSHOT)
             ));
+            repositories.add(new DefaultGradleRepository(
+                repositories.size(),
+                "https://central.sonatype.com/repository/maven-snapshots",
+                null,
+                List.of(
+                    new VersionRegexRepoFilter(
+                        "org[.]sitemesh.*", ".*", ".*-SNAPSHOT"
```
matrei, line 75: We don't need to include `-SNAPSHOT` in the regex here as that is explicitly set by the `VersionType.SNAPSHOT` below. (I know you probably followed suit here)

grails-profiles/web/profile.yml
```diff
@@ -44,7 +44,7 @@ dependencies:
     - scope: implementation
       coords: "org.apache.grails:grails-url-mappings"
     - scope: implementation
-      coords: "org.apache.grails:grails-layout"
+      coords: "org.apache.grails:grails-sitemesh3"
```
matrei: Shouldn't this be changed in gh-15713?

sbglasius (approved):

grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
```diff
@@ -517,6 +517,10 @@ class CreateAppCommand extends ArgumentCompletingCommand implements ProfileRepos
             GrailsGradleRepository repository = new GrailsGradleRepository(url: 'https://repository.apache.org/content/groups/snapshots', snapshotsOnly: true)
             repository.includeOnly('org[.]apache[.](grails|groovy).*', '.*', '.*-SNAPSHOT')
             configuredRepositories.add(repository)
+
+            GrailsGradleRepository sitemeshSnapshots = new GrailsGradleRepository(url: 'https://central.sonatype.com/repository/maven-snapshots', snapshotsOnly: true)
+            sitemeshSnapshots.includeOnly('org[.]sitemesh.*', '.*', '.*-SNAPSHOT')
```
matrei, line 522: The regex does not have to include `-SNAPSHOT` as `snapshotsOnly: true` is set above. (I know you probably followed suit here).

matrei (approved):

grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
jdaugherty: FYI: We have the filter on all of the other examples in this file.

grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
jdaugherty: FYI: We have the filter on all of the other examples in this file.

jdaugherty:
@matrei I'm going to go ahead and merge this - we can take a second pass at fixing the snapshot regex (this matches our existing pattern)

PR #15713, opened 2026-06-02
8.x Make SiteMesh 3 the default GSP layout

codeconsole:
Makes SiteMesh 3 the default GSP layout in generated applications, mutually exclusive with the legacy SiteMesh 2 `grails-layout` feature.

Introduces a `GspLayout` one-of feature group (enforced by `OneOfFeatureValidator`):
- **`GspLayout`** — abstract one-of parent (`Category.VIEW`, WEB/WEB_PLUGIN)
- **`Sitemesh3`** — default member; auto-applied unless another `GspLayout` is selected; adds `grails-sitemesh3`
- **`GrailsLayout`** — opt-in member; adds `grails-layout` (SiteMesh 2)

`GrailsGsp` is intentionally **not** modified: its existing `if (!isFeaturePresent(Sitemesh3))` guard already skips `grails-layout` when SiteMesh 3 applies, so SiteMesh 2 remains available via `GrailsLayout`. Selecting both `sitemesh3` and `grails-layout` now fails fast.

### Depends on #15710
Split out from #15710 per review feedback ("making this the default & updating to sitemesh 3 should be separate PRs"). This branch is stacked on #15710, so **until #15710 merges this PR's diff also shows the enable-SiteMesh-3 commits**; it will reduce to just the make-default change once #15710 lands. Please merge #15710 first.


jdaugherty (approved):

jamesfredley:
Gaps to review and potentially address before SM3 becomes default

#

| Gap / Open item | SM2 / Current state | SM3 / Needed |
| --- | --- | --- |
| `g:applyLayout` attributes (`template`,`url`,`action`,`controller`,`params`,`model`,`contentType`,`encoding`,`parse`) | Full support - `RenderGrailsLayoutTagLib.groovy:100-173` | `name` + body only; other attrs silently ignored - `RenderSitemeshTagLib.groovy:77-106` |
| Default-layout config + implicit `application` fallback | Reads `grails.views.layout.default` (`LayoutGrailsPlugin:37`); falls back to layout `application` (`GroovyPageLayoutFinder:180`) | Reads only `grails.sitemesh.default.layout` (`Sitemesh3GrailsPlugin:83`); no `application` fallback |
| Partial/template render suppression (`GrailsRenderViewMutator` + `GrailsLayoutSelector`) | Registers both (`LayoutGrailsPlugin:74-75`); consumed by core `ResponseRenderer:340-341,563` | Registers neither → beans null → behavior lost |
| `grails.views.layout.enable.nongsp` (non-GSP/JSP layouts) | `LayoutGrailsPlugin:38` + `GroovyPageLayoutFinder` `enableNonGspViews`/`viewMustExist` | `Sitemesh3LayoutFinder:192-194` GSP locator only; flag unsupported |
| Captured-page isolation across nested/sibling `applyLayout` | N/A (different mechanism) | `GrailsSiteMeshViewContext.dispatch:81` sets fresh page, no restore; taglib saves only `LAYOUT_ATTRIBUTE` (`RenderSitemeshTagLib:78,100-104`) |
| No-body captured page + `layoutBody` | Handled | `CaptureAwareContentProcessor:82` returns used page without data; `layoutBody` reads only `body` (`RenderSitemeshTagLib:218-225`) |
| `<title>` strip robustness | Regex-based, well-formed | `extractHead()` matches `"<title"` as prefix - mis-slices `<titlebar>`/`<title-x>` (`Sitemesh3CapturedPage.java:232`) |
| Aggregate web starter still defaults to SM2 | `starter-web/build.gradle:53` → `:grails-layout` unless `SITEMESH3_TESTING_ENABLED=true` | Default branch must be `:grails-sitemesh3` |
| ~35 test examples default to SM2 | `else → grails-layout` in every example (e.g. `app1/build.gradle:42-47`) | Default = SM3; SM2 via forced lane |
| Main unit suite defaults to SM2 | `grails-test-suite-uber/build.gradle:53-58` → `:grails-layout` | Default = SM3 |
| SM2-only examples with **no** SM3 coverage | Hardcoded `grails-layout`: `test-phases:42`, `scaffolding:49`, `scaffolding-fields:38`, `jetty:41`, `database-cleanup:37`, `gsp-layout:50` | Keep as SM2 anchors; add SM3 variants where parity matters |
| Default path ≠ tested path | `SITEMESH3_TESTING_ENABLED` gates SM3; default build never runs SM3 | Retire/invert flag so default == tested |
| No CI lane exercising SM3 as default | Only opt-in env runs SM3 | Add SM3-default lane + SM2 lane |
| Fragile upstream auto-config suppression | Disabled `NoopSitemeshFilter` named `sitemesh` suppresses upstream `SiteMeshAutoConfiguration` (`Sitemesh3GrailsPlugin:106-124`) | Robust suppression + guard test |
| Servlet 6.1 / Tomcat 11 forward→include dispatch | SM3 bypasses `RequestDispatcher.forward()` for absolute layout paths (`GrailsSiteMeshViewContext.java:88-93`) | Functional coverage on Tomcat 11 |
| `grails-gsp-spring-boot` unpublished | In `settings.gradle:174`, depends on SM3 (`spring-boot/build.gradle:37`), but absent from `publish-root-config.gradle` (only `grails-sitemesh3` at `:75`) | Publish or don't rely on it |

## Not gaps (verified parity)

Resolution order (6 steps), `NONE_LAYOUT`, `layoutTitle`/`layoutHead`, `pageProperty`/`ifPageProperty`, meta extraction, `body.*` attrs, `grailsLayout:parameter`, comma/chained decoration (happy path), unbounded layout cache (SM2 identical).

## Minor / cosmetic

Cache-interval key differs: SM3 `grails.sitemesh.layout.cache.interval` vs SM2 `grails.gsp.reload.interval`.

sbglasius (approved):

grails-forge/grails-forge-core/src/test/groovy/org/grails/forge/feature/view/GspLayoutSpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+        final String build = new BuildBuilder(beanContext)
+            .features(["gsp"])
+            .applicationType(applicationType)
+            .render()
+
+        then:
+        build.contains('implementation "org.apache.grails:grails-sitemesh3"')
+        !build.contains('implementation "org.apache.grails:grails-layout"')
+
+        and: "a SNAPSHOT build exposes the SiteMesh 3 snapshot repo so org.sitemesh:*-SNAPSHOT resolves"
```
matrei: These assertions will be short-lived

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/sitemesh3/Sitemesh3.java
```diff
@@ -38,12 +44,18 @@ public String getName() {
 
     @Override
     public String getTitle() {
-        return "Sitemesh 3";
+        return "SiteMesh 3";
```
matrei: Should this not align with the SiteMesh 2 title: "GSP SiteMesh 3 Layouts"?

grails-forge/grails-forge-core/src/test/groovy/org/grails/forge/feature/view/GspLayoutSpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+import org.grails.forge.application.ApplicationType
+import spock.lang.Unroll
+
+class GspLayoutSpec extends ApplicationContextSpec {
+
+    @Unroll
+    void "web #applicationType apps use SiteMesh 3 (grails-sitemesh3) by default"() {
+        when:
+        final String build = new BuildBuilder(beanContext)
+            .features(["gsp"])
```
matrei: Can we align on using single quotes for strings (here and below)?

matrei (commented):
There seems to be a glitch in the UI when a default feature is replaced:

Image

jdaugherty (dismissed):
I forgot about the test apps pointing at grails-layout by default.  We'll need to update that as part of this switch and ensure no regressions.

jamesfredley:
I wish github didn't scroll on tables like that in comments, but there are a number of key features that will need to be added to the sitemesh 3 plugin to match sitemesh 2's historical feature set.  When the tests are all run against sitemesh 3 it will surface most of these, but the comment above should be close to comprehensive.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/sitemesh3/Sitemesh3.java
codeconsole: Yes — renamed to `GSP SiteMesh 3 Layouts` to match the SiteMesh 2 title.

grails-forge/grails-forge-core/src/test/groovy/org/grails/forge/feature/view/GspLayoutSpec.groovy
codeconsole: Done — converted the spec to single-quoted strings throughout.

grails-forge/grails-forge-core/src/test/groovy/org/grails/forge/feature/view/GspLayoutSpec.groovy
codeconsole: Agreed — dropped the SNAPSHOT-repo assertions; the spec now only verifies the resolved layout dependency.

sbglasius (approved):

codeconsole:
https://github.com/apache/grails-core/pull/15585

.github/workflows/gradle.yml
```diff
@@ -24,6 +24,11 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+# Exercise SiteMesh 3 by default in the test apps. The example/suite builds read this
+# flag to choose grails-sitemesh3 over the legacy grails-layout; unset it (or set it to
+# anything other than 'true') to fall back to SiteMesh 2.
+env:
+  SITEMESH3_TESTING_ENABLED: 'true'
```
matrei: We should turn this around to default to SiteMesh 3. Otherwise we will run SiteMesh 2 by default locally.

matrei (commented):
There is a commented out `grails-test-examples-gsp-sitemesh3` project.
I enabled it and ran the tests. They fail. Is this expected?

matrei:
I think the Forge change actually made it worse, because now I cannot select `grails-layout` at all:

Screenshot from 2026-06-09 17-37-55

.github/workflows/gradle.yml
codeconsole: flipped

jamesfredley:
I re-checked the current PR head (`bb94c8c27d`) against the earlier SM2 parity list. A few items look addressed now, but these still look partial or unresolved:

| Item | Status | Notes |
| --- | --- | --- |
| `g:applyLayout` attribute parity | Partial | `RenderSitemeshTagLib.groovy:78` now handles body, `name`, and `params` (`:112`, `:122`), but still does not support the SM2 attribute surface: `template`, `url`, `action`, `controller`, `model`, `contentType`, `encoding`, or `parse`. |
| Default layout config + implicit `application` fallback | Unresolved | SM3 still reads only `grails.sitemesh.default.layout` (`Sitemesh3GrailsPlugin.groovy:83`) and passes `null` when unset (`:102`). It does not consult `grails.views.layout.default`, and does not match SM2's implicit `application` fallback behavior (`GroovyPageLayoutFinder.groovy:180`). |
| `grails.views.layout.enable.nongsp` | Unresolved | SM2 wires this through `LayoutGrailsPlugin.groovy:38` and `GroovyPageLayoutFinder.groovy:87`. I do not see an SM3 equivalent in `Sitemesh3LayoutFinder.java`, which still only resolves via the GSP locator. |
| `<title>` stripping robustness | Unresolved | `Sitemesh3CapturedPage.java:232` still searches for `"<title"` without checking the tag boundary, so cases like `<titlebar>` / `<title-x>` can still be misidentified before the real `</title>` is found. |
| Upstream auto-config suppression guard | Partial | There is improved view-resolver auto-config (`Sitemesh3AutoConfiguration.java:38`), but the plugin still relies on a disabled bean named `sitemesh` with `NoopSitemeshFilter` (`Sitemesh3GrailsPlugin.groovy:112`, `:125`). I do not see a guard test proving the upstream filter auto-config stays suppressed. |
| Test examples default path | Partial | Most examples now default to SM3 via `SITEMESH2_TESTING_ENABLED`, but the Hibernate 7 examples still use the old/inverted `SITEMESH3_TESTING_ENABLED` toggle and therefore default to SM2 when unset, e.g. `grails-test-examples/hibernate7/grails-hibernate/build.gradle:41`. |
| SM2-only examples without SM3 coverage | Partial | `database-cleanup`, `scaffolding-fields`, and `test-phases` were fixed, but `scaffolding` and `jetty` remain hardcoded to `grails-layout` only (`scaffolding/build.gradle:49`, `jetty/build.gradle:41`). `gsp-layout` can probably remain the intentional SM2 anchor. |
| CI coverage for SM2 lane | Unresolved | I do not see `SITEMESH2_TESTING_ENABLED` or `SITEMESH3_TESTING_ENABLED` in `.github/workflows`. The default SM3 path is exercised, but there is no explicit SM2 compatibility lane. |
| `grails-gsp-spring-boot` publication | Unresolved | The project is included in `settings.gradle:175` and depends on `grails-sitemesh3` (`grails-gsp/spring-boot/build.gradle:37`), but it is still absent from the published projects list in `gradle/publish-root-config.gradle` while `grails-sitemesh3` is listed at `:76`. |

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/web/taglib/RenderSitemeshTagLib.groovy
```diff
@@ -76,27 +110,130 @@ class RenderSitemeshTagLib implements TagLibrary {
     // causing "request is not active anymore" errors.
     Closure applyLayout = { Map attrs, body ->
```
jdaugherty, line 111: To make this faster, you should use the method version instead of closure.

jdaugherty:
I confirmed with forge locally that selecting sitemesh2 does what's expected: 

image

Otherwise, it uses sitemesh3 by default.

jdaugherty (commented):
The only change I would suggest is a separate workflow to run the sitesmesh2 work since legacy apps may still use it and we don't have those tests running by default.  I'm ok to merge without this or address later if we make gsp changes though.  Will defer to @jamesfredley on if we should do this.  Assuming he approves, I approve

jamesfredley:
I re-verified my earlier SM2 parity list against the current PR head and addressed the outstanding CI item. Most items have been resolved by the commits pushed since my last review; status below.

## SM2 parity list - current status

| Item | Status | Notes |
| --- | --- | --- |
| `g:applyLayout` attribute parity | ✅ Addressed | `RenderSitemeshTagLib` now handles `name`, `template`/`view`, `url`, `action`+`controller`, `params`, `model`, `contentType`, and `parse` (commit "Support the full applyLayout attribute surface under SiteMesh 3"). |
| Default layout config + implicit `application` fallback | ✅ Addressed | `Sitemesh3GrailsPlugin` reads `grails.sitemesh.default.layout` then falls back to `grails.views.layout.default`; `Sitemesh3LayoutFinder.resolveDefaultDecorator()` falls back to the implicit `application` layout when none is configured (commit "Match SiteMesh 2's default layout config and application fallback"). |
| `<title>` stripping robustness | ✅ Addressed | `Sitemesh3CapturedPage.indexOfTitleOpenTag()` now requires the char after `<title` to be `>` or whitespace, so `<titlebar>`/`<title-x>` are no longer mis-sliced (commit "Don't mis-slice head content around title-prefixed elements"). |
| Test examples default path (Hibernate 7) | ✅ Addressed | The hibernate7 examples now use the `SITEMESH2_TESTING_ENABLED` toggle and default to SM3 when unset, consistent with the rest of the examples. |
| SM2-only examples without SM3 coverage | ✅ Addressed | `scaffolding` and `jetty` now use the `SITEMESH2_TESTING_ENABLED` toggle (default SM3) rather than being hardcoded to `grails-layout`. |
| `grails-gsp-spring-boot` publication | ✅ Addressed | Now listed in `gradle/publish-root-config.gradle` alongside `grails-sitemesh3`. |
| Upstream auto-config suppression guard | ⚠️ Mechanism replaced | The disabled `NoopSitemeshFilter` workaround was dropped. Suppression is now done in `Sitemesh3AutoConfiguration` via `@AutoConfigureBefore(SiteMeshViewResolverAutoConfiguration)` + `@ConditionalOnMissingBean(SiteMeshViewResolverBeanPostProcessor)`, covered by `GrailsSiteMeshViewResolverBeanPostProcessorSpec`. There is still no dedicated test asserting the upstream *filter* auto-config stays suppressed, but the original fragile bean is gone. |
| `grails.views.layout.enable.nongsp` | ❌ Not addressed | `Sitemesh3LayoutFinder` still resolves only through the GSP locator; the SM2 `enableNonGspViews` / non-GSP (JSP) layout flag has no SM3 equivalent. Low priority - can be a follow-up if non-GSP layouts need to be supported under SM3. |
| CI coverage for SM2 lane | ✅ Addressed here | See below. |

## SM2 CI lane

Per @jdaugherty's suggestion (a separate workflow to run the SiteMesh 2 work, since legacy apps may still use it and it isn't exercised by default), I added `.github/workflows/sitemesh2.yml`. It sets `SITEMESH2_TESTING_ENABLED=true` - which flips the starter, the uber unit suite and the functional test examples back onto `:grails-layout` - and runs the core (`-PonlyCoreTests`) and functional (`-PonlyFunctionalTests`) lanes against it. `--rerun-tasks` ensures the tests execute against the grails-layout classpath rather than reusing SM3-cached outputs.

Triggers: `workflow_dispatch`, weekly `schedule`, and `push`/`pull_request` scoped to the GSP / test-suite / test-example / dependency paths where SM2-vs-SM3 behavior lives.

One caveat worth noting: GitHub only runs `schedule` and `workflow_dispatch` from the repository default branch, so the weekly/manual lane will not fire from an 8.x branch until 8.x is the default (PR and push coverage is unaffected). Remaining open items are `grails.views.layout.enable.nongsp` and an explicit upstream-filter-suppression guard test, both of which can be deferred as follow-ups.

jamesfredley (approved):

PR #15715, opened 2026-06-05
Add typed attribute converters for session, request, flash and servletContext

codeconsole:
## Summary

Adds null-safe, typed conversion methods to the `session`, `request`, `flash` and `servletContext` objects, mirroring the existing `params.int('id', 1)` API. Also adds a missing `string`/`getString` converter to `TypeConvertingMap` itself.

```groovy
Integer page    = request.int('page', 1)
String  tz      = session.string('userTimeZoneId', 'America/Los_Angeles')
Boolean active  = session.boolean('active', false)
String  notice  = flash.string('notice')
int     uploads = servletContext.int('maxUploads', 10)
```

The full converter set is available — `string`, `byte`, `char`, `short`, `int`, `long`, `double`, `float`, `boolean`, `list`, `date` — each with an optional default-value overload, exactly as on `params`.

## Motivation

Under `@CompileStatic` / `@GrailsCompileStatic`, dynamic attribute access such as `session.userTimeZoneId` does not compile, and the explicit form requires a cast at every call site:

```groovy
String tz = (session.getAttribute('userTimeZoneId') as String) ?: 'America/Los_Angeles'
```

These converters give the same null-safe, typed, default-aware ergonomics as `params`, and they compile under static compilation. They are equally useful in dynamic code for the parsing/defaulting behaviour (just like `params.int(...)`).

## Design

- Conversion logic is extracted into an internal helper, `org.grails.util.TypeConverters`, which converts a raw `Object` value to the target type (with an optional default). Both `TypeConvertingMap`'s instance `getX` methods and the new attribute extensions delegate to it, so the logic lives in a single place and no new conversion methods are added to the public `TypeConvertingMap` API.
- The extensions call the converters directly on the attribute value, avoiding any per-access allocation.
- The instance `getX(name, default)` semantics are preserved, so `params` behaviour is unchanged (verified by `GrailsParameterMapBindingSpec`).
- The one deliberate exception is the `boolean` default: "is the value present" differs by holder (map `containsKey` vs. attribute set), which cannot be expressed from a value alone, so it stays per-site and is documented inline.
- The `string` converter returns the first element when the attribute is an array, mirroring the single-value semantics of the other converters; use `list` for all values.

## Tests & docs

- Unit tests for the internal converters (incl. default-value coercion) and for `getString`/`string`.
- A Spock spec per holder (`HttpSessionExtensionSpec`, `HttpServletRequestExtensionSpec`, `ServletContextExtensionSpec`, `FlashScopeExtensionSpec`) covering conversion, null-safety, defaults, and a `@CompileStatic` static-resolution guard.
- `GrailsParameterMapBindingSpec` passes unchanged (regression guard for the core refactor).
- New "Type Conversion of Attributes" section in the controllers guide.

### DIFF
--- a/grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
+++ b/grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
@@ -16,7 +16,7 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-package org.grails.util;
+package org.apache.grails.core.internal.util;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
--- a/grails-web-core/src/main/groovy/org/grails/web/servlet/FlashScopeExtension.groovy
+++ b/grails-web-core/src/main/groovy/org/grails/web/servlet/FlashScopeExtension.groovy
@@ -1,0 +1,127 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.web.servlet
+
+import groovy.transform.CompileStatic
+
+import org.grails.util.TypeConverters
+
+import grails.web.mvc.FlashScope
+
+/**
+ * An extension that adds type conversion methods to the {@link FlashScope} interface
+ *
+ * @since 8.0
+ */
+@CompileStatic
+class FlashScopeExtension {
+
+    static Byte 'byte'(FlashScope flash, String name) {
+        TypeConverters.toByte(flash.get(name))
+    }
+
+    static Byte 'byte'(FlashScope flash, String name, Integer defaultValue) {
+        TypeConverters.toByte(flash.get(name), defaultValue)
+    }
+
+    static Character 'char'(FlashScope flash, String name) {
+        TypeConverters.toCharacter(flash.get(name))
+    }
+
+    static Character 'char'(FlashScope flash, String name, Character defaultValue) {
+        TypeConverters.toCharacter(flash.get(name), defaultValue)
+    }
+
+    static Character 'char'(FlashScope flash, String name, Integer defaultValue) {
+        TypeConverters.toCharacter(flash.get(name), defaultValue)
+    }
+
+    static Short 'short'(FlashScope flash, String name) {
+        TypeConverters.toShort(flash.get(name))
+    }
+
+    static Short 'short'(FlashScope flash, String name, Integer defaultValue) {
+        TypeConverters.toShort(flash.get(name), defaultValue)
+    }
+
+    static Integer 'int'(FlashScope flash, String name) {
+        TypeConverters.toInteger(flash.get(name))
+    }
+
+    static Integer 'int'(FlashScope flash, String name, Integer defaultValue) {
+        TypeConverters.toInteger(flash.get(name), defaultValue)
+    }
+
+    static Long 'long'(FlashScope flash, String name) {
+        TypeConverters.toLong(flash.get(name))
+    }
+
+    static Long 'long'(FlashScope flash, String name, Long defaultValue) {
+        TypeConverters.toLong(flash.get(name), defaultValue)
+    }
+
+    static Double 'double'(FlashScope flash, String name) {
+        TypeConverters.toDouble(flash.get(name))
+    }
+
+    static Double 'double'(FlashScope flash, String name, Double defaultValue) {
+        TypeConverters.toDouble(flash.get(name), defaultValue)
+    }
+
+    static Float 'float'(FlashScope flash, String name) {
+        TypeConverters.toFloat(flash.get(name))
+    }
+
+    static Float 'float'(FlashScope flash, String name, Float defaultValue) {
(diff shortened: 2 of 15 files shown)


matrei:
@codeconsole Can this target 8?
Decided in the weekly meeting May 27th, point 1, 7.x.x is not accepting new features.
Focus should now be on 8+.

codeconsole:
> @codeconsole Can this target 8? Decided in the weekly meeting May 27th, point 1, 7.x.x is not accepting new features. Focus should now be on 8+.

@matrei done, how about a review ;P

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    public static Byte toByte(Object o) {
+        if (o instanceof Number) {
+            return ((Number) o).byteValue();
+        }
+        if (o != null) {
+            try {
+                String string = o.toString();
+                if (string != null && string.length() > 0) {
```
sbglasius, line 55: it would be a weird case, if `o.toString()` would return `null` - but ok, if someone implements `String toString() { return null };` that would be a real edge case

grails-data-graphql/core/src/main/groovy/org/grails/gorm/graphql/fetcher/impl/DeleteEntityDataFetcher.groovy
```diff
@@ -48,7 +48,7 @@ class DeleteEntityDataFetcher<T> extends DefaultGormDataFetcher<T> implements De
     }
 
     protected void deleteInstance(GormEntity instance) {
-        instance.delete(failOnError: true)
```
sbglasius, line 51: This was implemented in https://github.com/apache/grails-core/pull/15716 - should it be here?

grails-data-graphql/core/src/test/groovy/org/grails/gorm/graphql/fetcher/impl/DeleteEntityDataFetcherSpec.groovy
```diff
@@ -76,6 +77,19 @@ class DeleteEntityDataFetcherSpec extends HibernateSpec {
...
+        GormEntity instance = Mock(GormEntity)
+
+        when:
+        fetcher.deleteInstance(instance)
+
+        then: 'delete is called with no arguments; failOnError was silently ignored by GORM, so removing it preserves the original behavior'
+        1 * instance.delete()
+        0 * instance.delete(_)
+    }
+
```
sbglasius: This was implemented in https://github.com/apache/grails-core/pull/15716 - should it be here too?

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.List;
+
+import grails.util.GrailsStringUtils;
+
+/**
+ * Internal helpers that convert a raw {@code Object} value to a target type without allocating
+ * an intermediate map. These power the type conversion methods on {@code TypeConvertingMap} and
+ * the servlet attribute extensions, keeping the conversion logic in a single place.
+ *
+ * <p>This is an internal class and is not part of the public Grails API.
```
jdaugherty, line 37: If this is an internal class, we shouldn't place this in org.grails.util.  The `grails.*` package was previously reserved for these type of classes.  I'd suggest we take this opportunity to move it under the apache namespace, so I would put this under `org.apache.grails.core.internal.util`

Reasoning being: 
* Base package across the entire project should represent our project - i.e. `org.apache.grails`
* Package names should be unique per gradle project (JMS compliance) - hence, the `.core` addition
* `internal` to indicate it shouldn't be used externally, this mirrors the old `grails.*` namespace
* `util` since it was extracted from existing `util` package code

grails-data-graphql/core/src/main/groovy/org/grails/gorm/graphql/fetcher/impl/DeleteEntityDataFetcher.groovy
jdaugherty: We should set the merge base to https://github.com/apache/grails-core/pull/15716 to remove this.

grails-web-core/src/main/groovy/org/grails/web/servlet/FlashScopeExtension.groovy
```diff
@@ -0,0 +1,127 @@
...
+import groovy.transform.CompileStatic
+
+import org.grails.util.TypeConverters
+
+import grails.web.mvc.FlashScope
+
+/**
+ * An extension that adds type conversion methods to the {@link FlashScope} interface
+ *
+ * @since 7.2
```
jdaugherty: 8.0

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * @since 7.2
+ */
+@SuppressWarnings({ "rawtypes", "unchecked" })
+public final class TypeConverters {
```
jdaugherty, line 41: I'm assuming this logic was a straight copy from the original methods?

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
codeconsole: agree, but still follows the logic checking from TypeConvertingMap where this is based off of.

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
codeconsole: yeah, took it from `AbstractTypeConvertingMap`, but genericizing toList got rid of it.

grails-core/src/main/groovy/org/apache/grails/core/internal/util/TypeConverters.java
codeconsole: ok

codeconsole:
@sbglasius @jdaugherty @matrei   annoying that I got bonus changes from other peoples work when I switched the base to 8.0.x, but I replayed everything on top of a fresh 8.0.x then force pushed this PR.  Should be clean now.

sbglasius (approved):

jdaugherty (approved):

PR #15716, opened 2026-06-05
Add MongoClientSettingsBuilderCustomizer hook for the Mongo client

codeconsole:
## What

Adds a `MongoClientSettingsBuilderCustomizer` extension point to GORM for MongoDB. Any beans of this type in the application context are applied, in order, to the `MongoClientSettings.Builder` of the default connection source before the `MongoClient` is created.

## Why

There is currently no supported way to customize `MongoClientSettings` for things that have no `grails.mongodb.*` configuration equivalent — most notably registering a driver `com.mongodb.event.CommandListener` for metrics/tracing (e.g. Micrometer's `MongoMetricsCommandListener`), but also programmatic connection-pool tuning, read/write concerns, etc. Today that requires replacing the datastore wiring.

Mirrors Spring Boot's `MongoClientSettingsBuilderCustomizer` of the same name, giving GORM-MongoDB users the same ergonomics.

## How

- New functional interface `org.grails.datastore.mapping.mongo.connections.MongoClientSettingsBuilderCustomizer`.
- `MongoConnectionSourceFactory` gains an `@Autowired(required = false) List<MongoClientSettingsBuilderCustomizer>` applied in `create()` just before `MongoClients.create(...)`. The factory is already registered with `bean.autowire = true` in `MongoDbDataStoreSpringInitializer`, so customizer beans are injected automatically. No behavior change when none are defined.

## Example

```groovy
@Bean
MongoClientSettingsBuilderCustomizer mongoMetrics(MeterRegistry registry) {
    return { builder -> builder.addCommandListener(new MongoMetricsCommandListener(registry)) }
}
```

## Tests

Added to `MongoConnectionSourceFactorySpec`:
- a registered customizer is applied, and its command listener lands in the built settings
- multiple customizers run in registration order

### DIFF
--- a/grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/connections/MongoClientSettingsBuilderCustomizer.groovy
+++ b/grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/connections/MongoClientSettingsBuilderCustomizer.groovy
@@ -1,0 +1,57 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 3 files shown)


matrei:
@codeconsole Can this target 8?
Decided in the weekly meeting May 27th, point 1, 7.x.x is not accepting new features.
Focus should now be on 8+.

codeconsole:
> @codeconsole Can this target 8? Decided in the weekly meeting May 27th, point 1, 7.x.x is not accepting new features. Focus should now be on 8+.

@matrei done, how about a review ;P

sbglasius (approved):

jdaugherty (approved):

PR #15717, opened 2026-06-05
Parameterize the @Scaffold service superclass so scaffolded service methods aren't type-erased

codeconsole:
## Problem

A scaffolded service declared with a parameterized base, e.g. `@Scaffold(GormMongoService<Book>)`, is given the **raw** type `GormService`/`GormMongoService` as its superclass — the `<Book>` type argument is dropped. So every inherited generic method (`get():T`, `list():List<T>`, `save(T):T`, …) resolves to the `GormEntity` upper bound rather than the concrete domain type.

Under `@GrailsCompileStatic`/`@CompileStatic` this forces a cast at every call site:

```groovy
Book b         = bookService.get(id)      // Cannot assign value of type GormEntity to variable of type Book
List<Book> all = bookService.list(params) // List<GormEntity>
```

## Root cause

In `ScaffoldingServiceInjector.performInjectionOnAnnotatedClass`:

```groovy
superClassNode = valueClassNode.getPlainNodeReference()                    // strips the <Book> generic
...
classNode.setSuperClass(GrailsASTUtils.nonGeneric(superClassNode, domainClass))
```

`getPlainNodeReference()` removes the generics, so the following `nonGeneric(raw, domain)` is a no-op: `replaceGenericsPlaceholders` early-returns the plain node because the type is no longer using generics. The domain type is known but never applied to the superclass.

## Fix

Set the superclass to a parameterized `GormService<Domain>`:

```groovy
ClassNode parameterizedSuper = superClassNode.getPlainNodeReference()
parameterizedSuper.setGenericsTypes(
    [new GenericsType(GrailsASTUtils.nonGeneric(domainClass))] as GenericsType[])
classNode.setSuperClass(parameterizedSuper)
```

## Impact

Inherited scaffold-service methods now resolve to the domain type with no casts under static compilation:

```groovy
Book b         = bookService.get(id)      // ✅ Book
List<Book> all = bookService.list(params) // ✅ List<Book>
```

Generics are compile-time only, so there is no runtime behavior change. Verified against a real app's `@Scaffold(GormMongoService<T>)` services — `get()`/`list()` compile to the domain type under `@GrailsCompileStatic` with zero casts (previously every such call required a cast or `@CompileDynamic`).

### DIFF
--- a/grails-scaffolding/src/main/groovy/org/grails/compiler/scaffolding/ScaffoldingServiceInjector.groovy
+++ b/grails-scaffolding/src/main/groovy/org/grails/compiler/scaffolding/ScaffoldingServiceInjector.groovy
@@ -92,11 +92,15 @@ class ScaffoldingServiceInjector implements GrailsArtefactClassInjector {
                 if (!domainClass) {
                     GrailsASTUtils.error(source, classNode, "Scaffolded service (${classNode.name}) with @Scaffold does not have domain class set.", true)
                 }
-                // Parameterize the superclass (GormService<Domain>) so inherited get()/list()/save()
+                // Parameterize the superclass (e.g. GormService<Domain>) so inherited get()/list()/save()
                 // resolve to the domain type under static compilation, not the GormEntity upper bound.
                 ClassNode parameterizedSuper = superClassNode.getPlainNodeReference()
                 parameterizedSuper.setGenericsTypes(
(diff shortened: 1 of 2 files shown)


jdaugherty (commented):
Let's add a test that makes use of this.

codeconsole:
> Let's add a test that makes use of this.

@jdaugherty done

sbglasius (approved):

jdaugherty (approved):

PR #15719, opened 2026-06-05
Register CriteriaTypeCheckingExtension in @GrailsCompileStatic and cover createCriteria().<terminal>{} closures

codeconsole:
## What

`CriteriaTypeCheckingExtension` exists in grails-core but is **not registered** in `@GrailsCompileStatic`'s extension list, and it only recognizes the `Domain.withCriteria { … }` form (closure as the direct argument). The far more common `Domain.createCriteria().list/get/count { … }` chain — where the closure is the argument to the *terminal* call on the builder — is never entered, so its `eq`/`order`/`projections`/etc. calls fail static type checking.

The result: every `createCriteria()…{}` site under static compilation has to be marked `@CompileDynamic`, even when the only dynamic part is the criteria builder DSL.

## Change

1. Extend `isCriteriaCall` to also open a criteria scope when the call is a terminal (`list`, `listDistinct`, `get`, `count`, `scroll`) chained directly on `Domain.createCriteria()`.
2. Add `org.grails.compiler.CriteriaTypeCheckingExtension` to the `@CompileStatic(extensions=[…])` list on `@GrailsCompileStatic`.

With both in place, criteria-builder closures type-check under `@GrailsCompileStatic` and no longer require `@CompileDynamic`.

## Scope / limits

The extension makes the *direct children* of the criteria closure dynamic. Deeply nested builder closures (e.g. `projections { property '…' }`) and post-processing that indexes the projection result (`withCriteria { … }[0]`) still need `@CompileDynamic` — those are out of scope here.

## Testing

Verified against a multi-module Grails 7.2 app: `createCriteria().list(params){ eq(...) }` and `createCriteria().get { ilike(...); maxResults(1) }` compile cleanly under `@GrailsCompileStatic` with the extension registered, where they previously required `@CompileDynamic`.

### DIFF
--- a/grails-core/src/main/groovy/grails/compiler/GrailsCompileStatic.groovy
+++ b/grails-core/src/main/groovy/grails/compiler/GrailsCompileStatic.groovy
@@ -33,5 +33,6 @@ import groovy.transform.CompileStatic
                            'org.grails.compiler.WhereQueryTypeCheckingExtension',
                            'org.grails.compiler.DynamicFinderTypeCheckingExtension',
                            'org.grails.compiler.DomainMappingTypeCheckingExtension',
+                           'org.grails.compiler.CriteriaTypeCheckingExtension',
                            'org.grails.compiler.RelationshipManagementMethodTypeCheckingExtension'])
 @interface GrailsCompileStatic {}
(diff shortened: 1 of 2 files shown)


codeconsole:
Superseded by the 8.0.x-based PR — 7.2.x is no longer accepting new features.

PR #15720, opened 2026-06-05
8.x Register CriteriaTypeCheckingExtension and cover createCriteria().<terminal>{} closures

codeconsole:
## What

`CriteriaTypeCheckingExtension` exists in grails-core but is **not registered** in `@GrailsCompileStatic`'s extension list, and `isCriteriaCall` only recognizes the `Domain.withCriteria { … }` / `Domain.createCriteria()` forms (closure as the direct argument). The chained `Domain.createCriteria().<terminal> { … }` form — where the closure is the argument to a terminal called *on the builder* — is never matched, so a criteria scope is never opened for it.

## Why it matters

Most chained terminals (`list`, `listDistinct`, `get`, `scroll`) still type-check today because `BuildableCriteria` declares them with a `Closure` parameter and `@DelegatesTo` resolves the closure body. But `count(Closure)` is **not** declared on `BuildableCriteria`, so:

```groovy
@GrailsCompileStatic
class Foo {
    def c() { Person.createCriteria().count { eq 'name', 'Anakin' } }
}
```

fails static compilation on stock 8.0.x:

```
[Static type checking] - Cannot find matching method
org.grails.datastore.mapping.query.api.BuildableCriteria#count(groovy.lang.Closure)
```

forcing a `@CompileDynamic` escape hatch.

## Change

1. Extend `isCriteriaCall` to also open a criteria scope when the call is a terminal (`list`, `listDistinct`, `get`, `count`, `scroll`) chained directly on `Domain.createCriteria()`. Registration alone is not enough — the `createCriteria()` scope is exited (via `afterMethodCall`) before the terminal call is visited, so the terminal needs to be recognized in its own right.
2. Add `org.grails.compiler.CriteriaTypeCheckingExtension` to the `@CompileStatic(extensions=[…])` list on `@GrailsCompileStatic`.

## Test

Adds `'Test compiling a class which invokes a chained createCriteria() terminal on a domain class'` to `GrailsCompileStaticCompilationErrorsSpec`, covering all six chained terminals. It **fails on stock 8.0.x** (on the `count` terminal) and **passes with this change**. The existing `GRAILS-11255` criteria spec continues to pass.

## Scope / limits

The extension makes the *direct children* of the criteria closure dynamic. Deeply nested builder closures (e.g. `projections { property '…' }`) and post-processing that indexes the result (`withCriteria { … }[0]`) still require `@CompileDynamic` — out of scope here.


grails-core/src/main/groovy/grails/compiler/GrailsCompileStatic.groovy
```diff
@@ -33,5 +33,6 @@ import groovy.transform.CompileStatic
                            'org.grails.compiler.WhereQueryTypeCheckingExtension',
                            'org.grails.compiler.DynamicFinderTypeCheckingExtension',
                            'org.grails.compiler.DomainMappingTypeCheckingExtension',
+                           'org.grails.compiler.CriteriaTypeCheckingExtension',
```
jdaugherty: To change the default, we need to expand the test coverage so that there are tests in the TCK - this way we know we haven't broken it in hibernate 7.

codeconsole: Added in b10f36fae9. The TCK now executes criteria queries that were compiled with `@GrailsCompileStatic`, covering both dispatch paths the extension produces: statically-resolved `Criteria` API calls and the dynamic fallback (`projections`, `min`/`countDistinct`/`property`, `maxResults`, and the chained `count { }` terminal, which is not declared on `BuildableCriteria`). Verified green against the in-memory, Hibernate 5, Hibernate 7, and MongoDB implementations locally.

codeconsole:
Wired `StaticCompiledCriteriaSpec` into `HibernateSuite` and `Hibernate5Suite` (`34854cbfc1`). The spec had been added to the TCK but wasn't in any `@SelectClasses` suite, so it never ran against a GORM implementation. It now executes against both Hibernate 7 and Hibernate 5 — 9 tests each, asserting real query results — covering both dispatch paths the extension produces: static delegate dispatch (`eq`/`like`/`or`/`order`) and the dynamic fallback (projections, `maxResults`, `count` chained on `createCriteria()`). A regression in either ORM now fails CI.

jdaugherty (approved):

sbglasius (approved):

PR #15721, opened 2026-06-08
Make domain properties nullable by default (Grails 8)

codeconsole:
## Summary

Change GORM's default constraint so an **unconstrained persistent (domain) property is nullable** (optional) rather than required. Today Grails applies an implicit `nullable: false` to every persistent property; this PR flips that default to `nullable: true`.

The default lives in `DefaultConstraintEvaluator` and is controlled by a new **YAML‑friendly boolean** that defaults to nullable‑by‑default:

```yaml
# application.yml — set false to restore the legacy required-by-default behaviour
grails:
  gorm:
    default:
      nullable: false
```

**Scope is the validation layer only.** Command‑object validation (`Validateable.defaultNullable()`) is intentionally left **required‑by‑default and unchanged** in this PR.

## Why this should be the Grails 8 default

**Grails is the only mainstream JVM data framework that is required‑by‑default.** Every comparable layer treats an unconstrained property as nullable and makes you opt *into* "required":

| Framework | Default for an unconstrained reference property |
|---|---|
| JPA / Hibernate | nullable (`@Column(nullable=false)` / `@NotNull` to require) |
| Spring Data JPA | nullable |
| Spring Data MongoDB | nullable (schemaless) |
| Micronaut Data | nullable (`@NonNull` / Kotlin non‑null type to require) |
| Jakarta Bean Validation (JSR‑380) | nullable — *no constraint = valid*; `@NotNull` is the opt‑in |
| **Grails / GORM (today)** | **required** (`nullable: true` to allow null) |

The case for flipping it:

1. **Principle of least surprise / ecosystem consistency.** Developers arriving from Spring Boot, Micronaut, JPA, or plain Bean Validation expect "nullable unless I say otherwise." Grails inverts that — one of the most common first‑week surprises ("why does saving fail when I didn't mark anything required?").
2. **It contradicts the storage layer.** SQL columns are nullable by default; MongoDB is schemaless. Grails' validation default is *stricter than the database it maps to*, for no structural reason.
3. **It inverts the JSR‑380 contract.** Bean Validation defines "absence of a constraint ⇒ the value is valid (including null)"; `@NotNull` is the explicit opt‑in. Grails' implicit `nullable:false` is a silent, framework‑specific reversal of the spec it otherwise embraces.
4. **It breaks composition.** Fields contributed by traits, base classes, or shared modules — or simply a new field added to an existing domain — silently become **required**, surfacing only as a runtime `ValidationException` on the first save (often in an unrelated path, e.g. a BootStrap seed). Mixing in reusable field sets is exactly the modular design Grails 8 should encourage; required‑by‑default punishes it.
5. **A major version is the right time.** Defaults that no longer match the ecosystem should be corrected at a major boundary. Grails 8 is that boundary.

## Backward compatibility

Deliberate behavior change, with a **one‑line opt‑out** to restore the legacy behavior per application. Either set the boolean flag (YAML or groovy):

```yaml
# grails-app/conf/application.yml
grails:
  gorm:
    default:
      nullable: false
```
```groovy
// grails-app/conf/application.groovy
grails.gorm.default.nullable = false
```

…or use the existing wildcard‑constraint form, which also still works:

```groovy
grails.gorm.default.constraints = {
    '*'(nullable: false)
}
```

The `grails.gorm.default.constraints` machinery applies `'*'` constraints *before* the framework default and skips the default when one is set (`canApplyNullableConstraint`), so either opt‑out composes cleanly with per‑property overrides.

Command objects are unaffected (still required‑by‑default). If you additionally want command objects to be nullable‑by‑default, that remains an explicit per‑class opt‑in via `static boolean defaultNullable() { true }`.

## Test suite

The affected test suite has been swept and the **full `./gradlew test` is green**. Rather than disabling the new default wholesale, each spec that asserts required‑by‑default semantics now declares the specific field(s) it depends on explicitly (`nullable: false`), reproducing the prior baseline for just those fields. A module‑wide `grails.gorm.default.constraints` opt‑out was deliberately **not** used as the general mechanism because that closure is also evaluated against the GORM mapping builder and can perturb per‑property mapping (e.g. `formula:`/derived‑property detection used by `f:fields`/`f:all`), and because several specs define their own `grails.gorm.default.constraints` that a module‑wide file would clobber.

A new `NullableByDefaultSpec` demonstrates the change **without any opt‑out** (an unconstrained property validates while null, while a property with an explicit `nullable: false` is still required), and verifies that `grails.gorm.default.nullable = false` restores required‑by‑default through `DefaultValidatorRegistry`.

## Notes / scope

- **Validation layer only.** Column/DDL nullability is governed separately by the mapping layer (`Property.nullable` / `GrailsDomainBinder`) and is **not** changed here; aligning the DDL default is a documented follow‑up for maintainers.
- Opening against `8.0.x`.

### DIFF
--- a/grails-core/src/main/resources/META-INF/spring-configuration-metadata.json
+++ b/grails-core/src/main/resources/META-INF/spring-configuration-metadata.json
@@ -223,6 +223,12 @@
             "defaultValue": {}
         },
         {
+            "name": "grails.gorm.default.nullable",
+            "type": "java.lang.Boolean",
+            "description": "Whether an unconstrained persistent property is nullable by default (Grails 8). Set to false to restore the legacy required-by-default behaviour.",
+            "defaultValue": true
+        },
+        {
             "name": "grails.gorm.custom.types",
             "type": "java.util.Map<java.lang.String,java.lang.Object>",
             "description": "Map of custom GORM types.",
--- a/grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
+++ b/grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
@@ -29,6 +29,7 @@ import org.grails.datastore.gorm.validation.constraints.eval.ConstraintsEvaluato
 import org.grails.datastore.gorm.validation.constraints.eval.DefaultConstraintEvaluator
 import org.grails.datastore.gorm.validation.constraints.registry.ConstraintRegistry
 import org.grails.datastore.gorm.validation.constraints.registry.DefaultConstraintRegistry
+import org.grails.datastore.mapping.config.Settings
 import org.grails.datastore.mapping.model.MappingContext
 import org.grails.validation.ConstraintEvalUtils
 
@@ -50,7 +51,10 @@ class DefaultConstraintEvaluatorFactoryBean implements FactoryBean<ConstraintsEv
     ConstraintsEvaluator getObject() throws Exception {
         ConstraintRegistry registry = new DefaultConstraintRegistry(messageSource)
 
-        new DefaultConstraintEvaluator(registry, grailsDomainClassMappingContext, ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config))
+        boolean cacheAutoTimestampAnnotations = grailsApplication.config.getProperty(Settings.SETTING_AUTO_TIMESTAMP_CACHE_ANNOTATIONS, Boolean, true)
+        boolean defaultNullable = grailsApplication.config.getProperty(Settings.SETTING_DEFAULT_NULLABLE, Boolean, true)
+        new DefaultConstraintEvaluator(registry, grailsDomainClassMappingContext,
+                ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config), cacheAutoTimestampAnnotations, defaultNullable)
     }
 
     @Override
--- a/grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
+++ b/grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
@@ -29,7 +29,6 @@ import org.grails.datastore.gorm.validation.constraints.eval.ConstraintsEvaluato
 import org.grails.datastore.gorm.validation.constraints.eval.DefaultConstraintEvaluator
 import org.grails.datastore.gorm.validation.constraints.registry.ConstraintRegistry
 import org.grails.datastore.gorm.validation.constraints.registry.DefaultConstraintRegistry
-import org.grails.datastore.mapping.config.Settings
 import org.grails.datastore.mapping.model.MappingContext
 import org.grails.validation.ConstraintEvalUtils
 
@@ -51,10 +50,9 @@ class DefaultConstraintEvaluatorFactoryBean implements FactoryBean<ConstraintsEv
     ConstraintsEvaluator getObject() throws Exception {
         ConstraintRegistry registry = new DefaultConstraintRegistry(messageSource)
 
-        boolean cacheAutoTimestampAnnotations = grailsApplication.config.getProperty(Settings.SETTING_AUTO_TIMESTAMP_CACHE_ANNOTATIONS, Boolean, true)
-        boolean defaultNullable = grailsApplication.config.getProperty(Settings.SETTING_DEFAULT_NULLABLE, Boolean, true)
         new DefaultConstraintEvaluator(registry, grailsDomainClassMappingContext,
-                ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config), cacheAutoTimestampAnnotations, defaultNullable)
+                ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config), true,
+                ConstraintEvalUtils.getDefaultNullable(grailsApplication.config))
     }
 
     @Override
--- a/grails-test-examples/hibernate5/grails-multiple-datasources/grails-app/conf/application.yml
+++ b/grails-test-examples/hibernate5/grails-multiple-datasources/grails-app/conf/application.yml
@@ -67,3 +67,10 @@ dataSources:
     driverClassName: org.h2.Driver
     dbCreate: create-drop
     url: jdbc:h2:mem:devDb2;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
+
+---
+# Grails 8 is nullable-by-default; this example asserts required-by-default validation, so opt out.
+grails:
+    gorm:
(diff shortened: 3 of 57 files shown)


sbglasius (approved):

jdaugherty (dismissed):
@jamesfredley created a documentation file for our default values and you're now changing that default, but not changing those files

jamesfredley:
Updated the default-values metadata JSON: `grails-core/src/main/resources/META-INF/spring-configuration-metadata.json` now includes `grails.gorm.default.nullable` with `defaultValue: true`, matching the Grails 8 nullable-by-default behavior.

grails-core/src/main/resources/META-INF/spring-configuration-metadata.json
```diff
@@ -222,6 +222,12 @@
             "description": "A closure applied as the default constraints for all domain classes.",
             "defaultValue": {}
         },
+        {
```
jdaugherty, line 225: You added this to grails-core instead of grails-datastore-core.

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
```diff
@@ -50,7 +51,10 @@ class DefaultConstraintEvaluatorFactoryBean implements FactoryBean<ConstraintsEv
     ConstraintsEvaluator getObject() throws Exception {
         ConstraintRegistry registry = new DefaultConstraintRegistry(messageSource)
 
-        new DefaultConstraintEvaluator(registry, grailsDomainClassMappingContext, ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config))
+        boolean cacheAutoTimestampAnnotations = grailsApplication.config.getProperty(Settings.SETTING_AUTO_TIMESTAMP_CACHE_ANNOTATIONS, Boolean, true)
```
jdaugherty: Why is this here? This doesn't seem related to the PR's change.

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
```diff
@@ -50,7 +51,10 @@ class DefaultConstraintEvaluatorFactoryBean implements FactoryBean<ConstraintsEv
...
+        boolean defaultNullable = grailsApplication.config.getProperty(Settings.SETTING_DEFAULT_NULLABLE, Boolean, true)
+        new DefaultConstraintEvaluator(registry, grailsDomainClassMappingContext,
```
jdaugherty, line 53: Let's move the defaultNullable assignment below the original code?

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
```diff
@@ -50,7 +51,10 @@ class DefaultConstraintEvaluatorFactoryBean implements FactoryBean<ConstraintsEv
...
+                ConstraintEvalUtils.getDefaultConstraints(grailsApplication.config), cacheAutoTimestampAnnotations, defaultNullable)
```
jdaugherty: Why aren't we setting these defaults inside of getDefaultConstraints?  You already have the config, it seems odd to pass them into it.

grails-test-examples/hibernate5/grails-multiple-datasources/grails-app/conf/application.yml
```diff
@@ -66,4 +66,11 @@ dataSources:
     pooled: true
     driverClassName: org.h2.Driver
     dbCreate: create-drop
-    url: jdbc:h2:mem:devDb2;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
\ No newline at end of file
+    url: jdbc:h2:mem:devDb2;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
+
+---
+# Grails 8 is nullable-by-default; this example asserts required-by-default validation, so opt out.
```
jdaugherty: Let's remove the redundant comments on the config in the test apps?  I'll only make this statement once

grails-core/src/main/resources/META-INF/spring-configuration-metadata.json
jdaugherty: @jamesfredley why are we not adding a specific configuration file for the datastore project?  That's where the constraint is actually used.

codeconsole: @jdaugherty there is no similar file in `grails-datastore-core`.   I put it where all the other `grails.gorm.*` keys are.  It wouldn't make sense to put it a location different from where the other keys are.

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
codeconsole: Reverted to passing true literally in b112a97d43

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
codeconsole: it's now resolved inline via ConstraintEvalUtils.getDefaultNullable(config) b112a97d43

grails-domain-class/src/main/groovy/org/grails/plugins/domain/support/DefaultConstraintEvaluatorFactoryBean.groovy
codeconsole: b112a97d43

grails-test-examples/hibernate5/grails-multiple-datasources/grails-app/conf/application.yml
codeconsole: Done

grails-core/src/main/resources/META-INF/spring-configuration-metadata.json
jdaugherty: I agree, and we shouldn't let this block this change.  But I think we've made a larger mistake here - the config keys were supposed to exist in the projects where they're used/defined.  It's my understanding that hasn't happened based on this.

jdaugherty:
I think this can be merged after we discuss in the weekly.  This PR identified a configuration issue that I believe we need to discuss before merging.  I'd suggest we merge this after Wednesday's meeting.

codeconsole:
@jdaugherty but if the hold up is just the json file, why not just approve and I can do a separate PR to address all the data mapping json changes. Unless you want just my 1 change by itself in a json file which I can do here if you want.

jdaugherty (approved):

PR #15722, opened 2026-06-08
Replace usages of deprecated external APIs

matrei:
Hibernate deprecations are excluded.


jdaugherty:
This will cause a merge conflict with the step 2 PR - since it evolves on top of the hibernate 7 changes.  I think we should merge that before this for that reason.

grails-shell-cli/src/main/groovy/org/grails/cli/util/ResourceUtils.java
```diff
@@ -159,7 +160,7 @@ public Resource getResource(String location) {
             }
             try {
                 // Try to parse the location as a URL...
-                URL url = new URL(location);
+                URL url = URI.create(location).toURL();
                 return new UrlResource(url);
             }
             catch (MalformedURLException ex) {
```
codeconsole: because of the switch to URI.create: 
```
catch (MalformedURLException | IllegalArgumentException ex) {
```

codeconsole (approved):
nice work

grails-shell-cli/src/main/groovy/org/grails/cli/util/ResourceUtils.java
matrei: Thanks!

jdaugherty (approved):

PR #15726, opened 2026-06-10
fix: parse ASCII minus in locale-aware number binding

### LINKED ISSUE #15714: Negative number bug with gsp and fields and Norwegian locale in browser.
### Expected Behavior

The solution provided by this issue https://github.com/apache/grails-core/issues/15178 did not quite do the trick. 

Still not easy to entering negative numbers for both Long and BigDecimal fields.

### Actual Behaviour

I did locate the issue of the issue to LocaleAwareNumberConverter. The convert method uses NumberFormat(Locale), and there are a few locale's that does not use '-' as minusSign. So then numberFormat.parse fails when there is a minusSign other than the one given by by the numberFormat.

If you find that the solution described in the example application is the way to solve the issue, I can try to create a pull request.

### Steps To Reproduce

_No response_

### Environment Information

- Ubuntu 24.04
- java 17.0.17
- Grails 7.1.1

### Example Application

https://github.com/yggdrasildy/negativenumbers_7_1_1

### Version

7.1.1

jamesfredley:
## Summary

Fixes #15714 by accepting browser-submitted ASCII hyphen-minus values in locale-aware number binding for locales whose DecimalFormat minus sign is U+2212, such as nb_NO.

The earlier PR #15475 correctly scoped the rendering fix to form field input output, but #15714 shows the remaining failure is on the request binding side. The sample application's proposed direction was right: convert a leading ASCII minus to the active locale minus before calling NumberFormat.parse, instead of changing display tags.

## Changes

- Normalize only a leading ASCII minus in LocaleAwareNumberConverter before parsing.
- Leave g:formatNumber, g:fieldValue, and form field rendering behavior unchanged.
- Add Spock coverage for Long, BigDecimal, and BigInteger converters in nb_NO and en_US.

## Verification

- ./gradlew :grails-databinding:test --tests "org.grails.databinding.converters.web.LocaleAwareNumberConverterSpec"
- ./gradlew :grails-databinding:test
- Oracle review-gate verdict: GREEN

### DIFF
--- a/grails-databinding/src/main/groovy/org/grails/databinding/converters/web/LocaleAwareNumberConverter.groovy
+++ b/grails-databinding/src/main/groovy/org/grails/databinding/converters/web/LocaleAwareNumberConverter.groovy
@@ -18,6 +18,7 @@
  */
 package org.grails.databinding.converters.web
 
+import java.text.DecimalFormat
 import java.text.NumberFormat
 import java.text.ParsePosition
 
@@ -56,12 +57,24 @@ class LocaleAwareNumberConverter implements ValueConverter {
     @Override
     Object convert(Object value) {
         def trimmedValue = value.toString().trim()
+        def formatter = numberFormatter
+        def valueToParse = replaceAsciiMinusWithLocaleMinus(formatter, trimmedValue)
         def parsePosition = new ParsePosition(0)
-        def result = numberFormatter.parse((String) value, parsePosition).asType(getTargetType())
-        if (parsePosition.index != trimmedValue.size()) {
+        def result = formatter.parse(valueToParse, parsePosition).asType(getTargetType())
+        if (parsePosition.index != valueToParse.size()) {
             throw new NumberFormatException("Unable to parse number [${value}]")
         }
         result
+    }
+
+    protected String replaceAsciiMinusWithLocaleMinus(NumberFormat formatter, String value) {
+        if (formatter instanceof DecimalFormat && value.startsWith('-')) {
+            char minusSign = formatter.decimalFormatSymbols.minusSign
+            if (minusSign != '-' as char) {
+                return String.valueOf(minusSign) + value.substring(1)
+            }
+        }
+        value
     }
 
     protected NumberFormat getNumberFormatter() {
--- a/grails-databinding/src/test/groovy/org/grails/databinding/converters/web/LocaleAwareNumberConverterSpec.groovy
+++ b/grails-databinding/src/test/groovy/org/grails/databinding/converters/web/LocaleAwareNumberConverterSpec.groovy
@@ -1,0 +1,54 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


grails-databinding/src/main/groovy/org/grails/databinding/converters/web/LocaleAwareNumberConverter.groovy
```diff
@@ -56,14 +57,26 @@ class LocaleAwareNumberConverter implements ValueConverter {
     @Override
     Object convert(Object value) {
         def trimmedValue = value.toString().trim()
+        def formatter = numberFormatter
+        def valueToParse = replaceAsciiMinusWithLocaleMinus(formatter, trimmedValue)
```
jdaugherty, line 61: So the JVM has built in locale handling and yet we still have to work around it?

jdaugherty (approved):
I'm ok with the change, but I don't think I fully understand why we have to handle this locale issue directly

jamesfredley:
> I'm ok with the change, but I don't think I fully understand why we have to handle this locale issue directly

Short version: this isn't Grails working around itself - it's a hard limitation of `java.text.DecimalFormat` once the JVM uses CLDR locale data (the default since JDK 9). I dug into the JDK internals to be sure we weren't just doing something dumb on our side, and we aren't.

**Root cause**

For `nb_NO` (and `nn_NO`, `sv_SE`, `fi_FI`, ...) CLDR defines the minus sign as **U+2212 MINUS SIGN**, not **U+002D ASCII HYPHEN-MINUS**. `DecimalFormat` builds its negative prefix straight from `DecimalFormatSymbols.getMinusSign()` and matches that prefix *literally* during `parse()`. So an ASCII `-` is simply not recognized as a sign - the parser stops at index 0 and the bind fails.

Probe on JDK 21 (Corretto 21.0.11):

```
=== Locale nb_NO ===
  minusSign char = U+2212
  negativePrefix = U+2212
  parse("-1"  [U+002D U+0031]) -> null   consumedIndex=0/2   // ASCII minus: FAILS
  parse("...1" [U+2212 U+0031]) -> -1    consumedIndex=2/2   // locale minus: ok
  format(-1)  -> [U+2212 U+0031]

=== Locale en_US ===
  minusSign char = U+002D
  parse("-1"  [U+002D U+0031]) -> -1     consumedIndex=2/2   // ASCII minus: ok
```

So `NumberFormat.getInstance(new Locale("nb","NO")).parse("-1")` returns `null`. `en_US` only works because its CLDR minus *happens* to be ASCII - which is exactly why this stayed hidden until a Norwegian-locale browser posted a form.

**Why Java won't "just pass it through"**

There is no leniency knob for this. `DecimalFormat` has no "accept either minus" mode, and it has no idea the string arrived from an HTML form. The request side, meanwhile, is always ASCII: per the WHATWG "valid floating-point number" rule an `<input type=number>` submits ASCII `-`, and on a `text` field the user types the `-` on their keyboard, which is also ASCII. So the wire is ASCII while the locale parser only accepts U+2212, and nothing in the JDK reconciles that gap for us.

**The "just use Java's default" options I ruled out**

- `-Djava.locale.providers=COMPAT` restores the pre-9 ASCII minus, but it's a global, deprecated JVM flag (COMPAT data is being removed) and a wrong fix for one parse path.
- Forcing `symbols.minusSign = '-'` makes ASCII parse but then *rejects* a genuine U+2212 - strictly less lenient than what we ship now.

That's why #15475 fixed the render side (emit ASCII minus into inputs) and this PR fixes the matching bind side. Normalizing only a leading ASCII `-` to the active locale's minus before `parse()` is the minimal change that keeps full locale-aware parsing (grouping/decimal separators) intact while accepting the sign every browser actually sends - and the locale's native minus still parses too, so it's strictly more lenient, not less.

PR #15729, opened 2026-06-11
7.2.x (grails-scaffolding) - Parameterize the @Scaffold controller superclass so scaffolded controller members aren't type-erased

codeconsole:
## Summary

Controller-side counterpart of #15717. `ScaffoldingControllerInjector` set the injected superclass via `GrailsASTUtils.nonGeneric(superClassNode, domainClass)`, so a scaffolded controller's superclass was written **raw** and every inherited generic member erased to its `GormEntity` upper bound under static compilation:

```groovy
@Scaffold(RestfulController<Widget>)
@GrailsCompileStatic
class WidgetController {
    @Override
    protected Widget queryForResource(Serializable id) {
        def widget = super.queryForResource(id)   // before: typed GormEntity, not Widget
        (widget?.owner == currentOwner) ? widget : null
    }
}
```

fails with:

```
[Static type checking] - No such property: owner for class: org.grails.datastore.gorm.GormEntity
[Static type checking] - Cannot return value of type org.grails.datastore.gorm.GormEntity for method returning com.example.Widget
```

The common workaround in real apps is sprinkling `@CompileDynamic` over every `queryForResource`/`createResource`/`listAllResources` override, giving up static compilation exactly where project-scoping security checks live.

## Design

Same pattern #15717 applied to `ScaffoldingServiceInjector`: take a `getPlainNodeReference()` copy of the resolved superclass, set its generics to the domain type, mark the class node with `setUsingGenerics(true)` — injection runs at CANONICALIZATION (after generics resolution), so the generic superclass signature is only emitted when the class node itself reports `usesGenerics`; otherwise it is written raw — then `setSuperClass(parameterizedSuper)`. Applies to all three forms: `@Scaffold(Widget)`, `@Scaffold(RestfulController<Widget>)`, and custom bases `@Scaffold(ApiController<Widget>)` (the declared base is preserved, not collapsed to `RestfulController`).

## Tests

New `ScaffoldingControllerInjectorSpec` mirroring `ScaffoldingServiceInjectorSpec`:

- simple form `@Scaffold(Widget)` → `genericSuperclass` is `RestfulController<Widget>`, not raw
- generic form `@Scaffold(RestfulController<Widget>)` → parameterized
- custom scaffold base preserved and parameterized as `CustomBase<Widget>`
- a `@CompileStatic` scaffolded controller overriding `queryForResource` / `createResource` / `listAllResources` and narrowing the `super.*` results to the domain type compiles without casts

All four tests fail against the previous injector and pass with the fix; `:grails-scaffolding:test` passes in full.

### DIFF
--- a/grails-scaffolding/src/main/groovy/org/grails/compiler/scaffolding/ScaffoldingControllerInjector.groovy
+++ b/grails-scaffolding/src/main/groovy/org/grails/compiler/scaffolding/ScaffoldingControllerInjector.groovy
@@ -22,6 +22,7 @@ package org.grails.compiler.scaffolding
 import groovy.transform.CompileStatic
 import org.codehaus.groovy.ast.ClassHelper
 import org.codehaus.groovy.ast.ClassNode
+import org.codehaus.groovy.ast.GenericsType
 import org.codehaus.groovy.ast.expr.ClassExpression
 import org.codehaus.groovy.ast.expr.ConstantExpression
 import org.codehaus.groovy.classgen.GeneratorContext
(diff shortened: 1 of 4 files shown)


codeconsole:
Added a small robustness guard (both injectors, since the merged #15717 service-injector code shares the shape): the superclass is only parameterized when the base declares exactly one type parameter. A base with zero or multiple type parameters — e.g. `@Scaffold(MultiParamBase<Widget, String>)` — previously would have received a malformed single-argument generic signature; it now keeps the prior raw form. Covered by a new spec case in each injector spec (both fail without the guard).

sbglasius (approved):

jdaugherty (commented):
In 8.x, I 100% agree with this change.  In 7.2.x, I'm not sure.  @matrei would you consider this a breaking change?

codeconsole:
@jdaugherty it matches https://github.com/apache/grails-core/pull/15717 which is already in 7.2.x

jdaugherty:
I'm good to merge then.

jdaugherty (approved):

codeconsole:
@jdaugherty thanks

PR #15730, opened 2026-06-11
Inherit Spring Boot managed dependency versions in grails-bom

### LINKED ISSUE #15674: Audit dependencies.gradle overrides against Spring Boot 4.0.6 BOM
# Audit `dependencies.gradle` overrides against Spring Boot 4.0.6 BOM

## Goal

`dependencies.gradle` (root) declares two version maps that feed `grails-bom`: `gradleBomDependencyVersions` (build deps) and `bomDependencyVersions` (app deps). Many of those pins overlap with versions Spring Boot's `spring-boot-dependencies` BOM already manages.

- **Where our version equals SB's**: the pin is dead weight - it just prevents us from receiving free patches and security updates when SB bumps the dep. Each such pin should be deleted unless there is a specific reason to keep it.
- **Where our version diverges from SB's**: we need a documented, defensible justification or we should align with SB.

This issue audits every pin in the root `dependencies.gradle` against Spring Boot 4.0.6's BOM (the version we are about to upgrade to) and proposes concrete removals and re-alignments.

> Note: the audit was originally drafted against SB 4.0.5 and updated to SB 4.0.6 ahead of that upgrade. The only material change SB 4.0.6 introduces for our pinned deps is `mongodb.version` 5.6.4 → 5.6.5; everything else of ours stays in the same SAME / DIVERGENT bucket it was in.

## Methodology

- Source of truth for SB's managed versions: [`spring-boot-dependencies-4.0.6.pom`](https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/4.0.6/spring-boot-dependencies-4.0.6.pom) and the BOMs it imports (notably `junit-bom:6.0.3`).
- Cross-referenced every key in `bomDependencyVersions` and `gradleBomDependencyVersions` against SB's `<properties>` block.
- Located the introducing/bumping commit for each override via `git log -S/-G -- dependencies.gradle`.

## Summary table

| Pin | Ours | SB 4.0.6 | Status | Recommendation |
|---|---|---|---|---|
| `mongodb.version` | 5.6.4 | 5.6.5 | **DIVERGENT** -1 patch (was SAME in SB 4.0.5) | Remove |
| `rxjava3.version` | 3.1.12 | 3.1.12 | **SAME** | Remove |
| `jakarta-servlet-api.version` | 6.1.0 | 6.1.0 (`jakarta-servlet.version`) | **SAME** | Remove |
| `jakarta-validation.version` | 3.1.1 | 3.1.1 | **SAME** | Remove |
| `junit.version` | 6.0.3 | 6.0.3 (via imported `junit-bom`) | **SAME** | Remove or rewrite comment - see notes |
| `graphql-java.version` | 25.0 | 25.0 | **SAME** (deliberate tripwire, just added in #15673) | Keep with rationale comment |
| `commons-codec.version` | 1.21.0 | 1.19.0 | **DIVERGENT** +2 minor | Investigate - no documented reason |
| `selenium.version` | 4.38.0 | 4.37.0 | **DIVERGENT** +1 minor | Investigate - stale conflict reason |
| `commons-lang3.version` | 3.20.0 | 3.19.0 | **DIVERGENT** +1 minor | Keep - CVE-2025-48924 + LANG-1786 |
| `byte-buddy.version` (gradleBom) | 1.18.8 | 1.17.8 | **DIVERGENT** +1 minor | Keep - Mockito conflict |
| `groovy.version` | 4.0.32 | 5.0.5 | **DIVERGENT** major | Keep - Grails 8 baseline, Spock classifier lock |

## SAME as / behind SB - remove candidates

These pins either match SB's managed version exactly or have just fallen behind by a patch. They prevent automatic uptake of SB's patches and serve no resolution purpose today.

### `mongodb.version = 5.6.4` (SB 4.0.6 ships 5.6.5)

SB 4.0.5 shipped 5.6.4 (we were aligned); SB 4.0.6 bumps to 5.6.5. After the SB upgrade, our pin would force `5.6.4` while SB's transitively imported BOM would prefer `5.6.5`. The dependency validator plugin will flag this as drift on every grails-data-mongodb consumer, exactly the friction this whole audit is intended to eliminate.

**Action**: drop the property and the four `mongodb-*` entries from `bomDependencies` (`mongodb-bson`, `mongodb-driver-core`, `mongodb-driver-sync`, `mongodb-record-codec`). Let SB own mongodb going forward.

### `rxjava3.version = 3.1.12`

SB 4.0.6 manages `rxjava3.version = 3.1.12`. Identical. Drop the property and the `rxjava3` entry from `bomDependencies`. (RxJava 1 and RxJava 2 are not managed by SB; keep those pins.)

### `jakarta-servlet-api.version = 6.1.0`

SB 4.0.6 manages `jakarta-servlet.version = 6.1.0` (note property name differs, artifact `jakarta.servlet:jakarta.servlet-api` is the same). Identical. Drop the property and the `jakarta-servlet-api` entry from `bomDependencies`.

Historical context: an older revision of `dependencies.gradle` had this in a `// start - boot & selenium conflict, so pin the version we want (newest)` block (commit `d1e8460ca0`, May 2026). Selenium previously pulled an older `jakarta.servlet-api`, but both now agree on 6.1.0. The conflict is resolved.

### `jakarta-validation.version = 3.1.1`

SB 4.0.6 manages `jakarta-validation.version = 3.1.1`. Identical. Same history as `jakarta-servlet-api`. Drop the property and the `jakarta-validation` entry.

### `junit.version = 6.0.3` (most complex case)

SB 4.0.6 manages `junit-jupiter.version = 6.0.3` AND imports `org.junit:junit-bom:6.0.3`, which manages every `junit-jupiter-*` and `junit-platform-*` artifact at 6.0.3.

`grails-bom` ALSO imports `junit-bom:6.0.3` directly via `bomPlatformDependencies` (line 108). The 12 explicit `junit-jupiter-*` / `junit-platform-*` entries in `bomDependencies` (lines 157-168), all annotated `// boot & spock conflict`, are now redundant for version resolution - both SB and our own `junit-bom` import already agree on 6.0.3.

History (two commits):

- `1e0026c2eb` (2026-04-26) introduced the pin at JUnit **5.13.3** / platform 1.13.3 with the verbose `// start - boot & spock conflict, so pin the version we want (newest)` block. At the time, SB was managing an older JUnit 5.x and Spock 2.3's transitive graph pulled a different `junit-platform` version. The pin forced the highest.
- `ed097eccc6` (2026-04-28) upgraded the pin to **6.0.3** (collapsing the separate `junit-platform.version` property into `junit.version`) and shortened the comment to today's terse `// boot & spock conflict`.

Spock 2.3 BOM contributes **zero junit version constraints** (verified by reading `spock-bom-2.3-groovy-4.0.pom`). Spock 2.3 transitively requires `junit-platform >= 1.8`; version 6.0.3 satisfies that. The original three-way conflict (Boot vs Spock-transitive vs grails) is gone.

Two options:

- **(a) Remove the property and the 12 entries entirely.** `enforcedPlatform(:grails-bom)` consumers would lose forced versions for junit-* artefacts - they would still resolve to 6.0.3 via SB's transitively-imported BOM, but Gradle would not strictly enforce it.
- **(b) Keep the 12 entries but rewrite the comment.** If we want grails-bom's published POM to provide managed junit versions to downstream consumers using `enforcedPlatform`, the entries are still load-bearing for that secondary purpose. Replace `// boot & spock conflict` with `// expose junit versions in grails-bom for enforcedPlatform consumers` or similar.

Recommendation: pick (a) unless we have evidence downstream Grails apps depend on grails-bom strictly enforcing junit versions when they import as `enforcedPlatform`.

### `graphql-java.version = 25.0` (keep, deliberate tripwire)

SB 4.0.6 still ships `graphql-java = 25.0` (unchanged from 4.0.5). This pin was just RE-ADDED in PR #15673 specifically to act as a drift tripwire for `graphql-java-extended-scalars`. The dependency validator will flag drift when SB bumps graphql-java, prompting a check of whether extended-scalars has shipped a matching release. Should remain pinned even though the value equals SB's.

Recommend adding a comment above the line documenting this rationale, so a future audit doesn't blindly delete it.

### `commons-lang3.version = 3.20.0` (SB 4.0.6 ships 3.19.0)

Bumped from SB-managed 3.17.0 → 3.20.0 in commit [`c0d58dfda8`](../commit/c0d58dfda8). Commit message:

> Override commons-lang3 from 3.17.0 (Spring Boot managed) to 3.20.0 in the Grails BOM to fix LANG-1786 timezone warnings and **CVE-2025-48924**.

SB went 3.17.0 → 3.19.0 between releases but has not yet picked up 3.20.0 in 4.0.6 either. Pin should remain until SB catches up.

### `groovy.version = 4.0.32` (SB 4.0.6 ships 5.0.5)

SB went 5.0.4 → 5.0.5 in 4.0.6 but the gap remains a major version. Three independent, repository-documented reasons to keep this override:

1. **Spock classifier incompatibility.** `spock.version = 2.3-groovy-4.0` is a binary artefact classifier - no `2.3-groovy-5.0` artefact exists. Accepting SB's Groovy 5 without simultaneously migrating to Spock 2.4-groovy-5.0 breaks test compilation for every module that uses Spock.
2. **Spring Dependency Management Plugin resolution bug.** Commit [`44bbef5bad`](../commit/44bbef5bad) (issue #15324) documents that without an explicit `groovy-bom` platform pin and explicit `groovy-*` artefact entries, the Spring DM Plugin silently resolves SB's higher Groovy version, overriding Grails's intended version. The fix was to add the explicit pins at lines 107, 123-152.
3. **Documented Grails 8 baseline.** [`AGENTS.md`](../blob/8.0.x/AGENTS.md) (root) line 75 and [`.agents/skills/grails-developer/SKILL.md`](../blob/8.0.x/.agents/skills/grails-developer/SKILL.md) line 51 both state Groovy 4.0.x as the supported version. Groovy 5 support requires a coordinated Spock 2.4 migration and a 60+ module compatibility audit.

The repository deliberately isolates Groovy 5 to the `grails-micronaut-bom` block (`dependencies.gradle` lines 220-268). [`settings.gradle`](../blob/8.0.x/settings.gradle) lines 505-510 document the split verbatim:

> Grails-Micronaut 'island' - kept on its own dependency graph because the Micronaut platform pins Groovy 5 / Spock 2.4-groovy-5.0, which is incompatible with the Groovy 4 / Spock 2.3-groovy-4.0 stack the rest of grails-core targets.

### `byte-buddy.version = 1.18.8` (SB 4.0.6 ships 1.17.8)

Pin lives in `gradleBomDependencyVersions` (line 29). SB 4.0.6 did not bump byte-buddy from 4.0.5 (still 1.17.8). Bumped from SB-managed 1.17.8 → 1.18.8 in commit [`49d6418f88`](../commit/49d6418f88) alongside a Mockito bump (5.20.0 → 5.23.0). Historical justification commit [`5e8ef281e6`](../commit/5e8ef281e6):

> Mockito & boot boms conflict, so pin the newest version of byte-buddy to match.

Mockito's own dependency graph pulls a newer byte-buddy than SB's; pinning to newest keeps both happy. Should remain until SB and Mockito converge. Note that SB 4.0.6 still ships `mockito.version = 5.20.0` (unchanged), so the underlying Mockito-vs-SB conflict is unchanged from 4.0.5.

### `commons-codec.version = 1.21.0` (SB 4.0.6 ships 1.19.0)

SB 4.0.6 did not bump commons-codec from 4.0.5. Bumped from 1.19.0 → 1.21.0 in commit [`49d6418f88`](../commit/49d6418f88) as part of a generic "patch-level dependency versions" bump. No CVE or specific issue cited in the commit message:

> Bump patch-level dependencies:
> - byte-buddy 1.17.8 -> 1.18.8
> - commons-text 1.14.0 -> 1.15.0
> - ant 1.10.15 -> 1.10.17
> - commons-codec 1.19.0 -> 1.21.0
> - selenium-bom 4.38.0 -> 4.41.0
> - mockito-core 5.20.0 -> 5.23.0

**Action needed**: Decide whether to keep `1.21.0` (and document a reason) or downgrade to match SB's `1.19.0`. There's no current evidence the 1.21.0 bump is load-bearing; if it's just dependabot-style currency, removing the pin and inheriting from SB is preferable for security update flow.

### `selenium.version = 4.38.0` (SB 4.0.6 ships 4.37.0)

SB 4.0.6 did not bump selenium from 4.0.5. Bumped in commit [`471af067fc`](../commit/471af067fc) ("fix(deps): update selenium to 4.38.0") - a Renovate-style auto-bump. The original justification for pinning selenium AT ALL was commit [`d758bf6524`](../commit/d758bf6524):

> Spring boot's bom brings in opentelemetry & so does selenium, pin to a consistent version.

That earlier opentelemetry conflict has been resolved (the opentelemetry pin was removed in a later commit). The version is now just being kept current by automation.

**Action needed**: Decide whether to keep selenium pinned. If we drop the pin, we follow SB (currently 4.37.0). If we keep it, we should document the current reason (e.g., "stay on newest stable Selenium for Geb compatibility") or accept that Renovate keeps it 0-1 minor ahead.

## Pins NOT managed by Spring Boot - correctly retained

These are not managed by SB's BOM and must remain pinned. No action required:

`asset-pipeline-bom`, `bootstrap-icons`, `bootstrap`, `geb-spock`, `graphql-java-extended-scalars`, `hibernate-groovy-proxy`, `jquery`, `rxjava` (RxJava 1), `rxjava2`, `sitemesh` (Sitemesh 2), `spock`, `starter-sitemesh`, `spring-webmvc-sitemesh`, `spring-retry` (note: SB 4 explicitly dropped management of spring-retry - documented inline), and the entire `gradleBomDependencyVersions` map except `byte-buddy.version` (ant, jline, jansi, jna, asciidoctorj, asciidoctor-gradle-jvm, asset-pipeline-gradle, commons-text, directory-watcher, gradle-groovy, gradle-spock, grails-publish-plugin, javaparser-core, jline2, objenesis - none are in SB's BOM).

## Intentionally divergent custom-BOM overrides - no action

The `customBomVersions` blocks in `dependencies.gradle` (lines 198-269) are intentionally divergent and documented inline:

- **`grails-hibernate5-bom` block** (lines 198-214): pins `hibernate.version = 5.6.15.Final` (SB 4.0.6 ships 7.2.12.Final) and `liquibase.version = 4.27.0` (SB ships 5.0.2). These are the entire reason `grails-hibernate5-bom` exists - to provide Hibernate 5 instead of SB's Hibernate 7.
- **`grails-micronaut-bom` block** (lines 220-268): pins `groovy.version = 5.0.5`, `spock.version = 2.4-groovy-5.0`, `protobuf.version = 4.30.2`. Layers Micronaut-compatible versions on top of grails-bom for Micronaut consumers. Documented as the "island" in `settings.gradle` lines 505-510. Note: this block's Groovy 5.0.5 now happens to match SB 4.0.6's `groovy.version = 5.0.5`, which is incidental rather than designed.

## Action items

- [ ] Remove `mongodb.version` and its 4 `bomDependencies` entries (the SB 4.0.5→4.0.6 mongodb bump from 5.6.4→5.6.5 makes this the most urgent removal candidate)
- [ ] Remove `rxjava3.version` and its `bomDependencies` entry
- [ ] Remove `jakarta-servlet-api.version` and its `bomDependencies` entry
- [ ] Remove `jakarta-validation.version` and its `bomDependencies` entry
- [ ] Decide on `junit.version`: remove entirely OR rewrite the 12 inline comments to reflect the actual current purpose (expose in published BOM for enforcedPlatform consumers)
- [ ] Add a rationale comment above `graphql-java.version` documenting it as a drift tripwire so future audits don't blindly delete it
- [ ] Decide on `commons-codec.version`: downgrade to SB's 1.19.0 or document a justification for 1.21.0
- [ ] Decide on `selenium.version`: drop the pin (and follow SB) or document why we stay 0-1 minor ahead
- [ ] When SB next ships a release that bumps `commons-lang3` to >= 3.20.0, drop our override
- [ ] When SB next ships a release that bumps `byte-buddy` to match Mockito's needs, drop our override
- [ ] Verify with downstream Grails app builds that removing these pins does not break `enforcedPlatform(:grails-bom)` consumers

## Notes on `enforcedPlatform` semantics

`grails-bom/default/build.gradle` lines 60-63 contain this comment:

```groovy
// Re-declare base BOM constraints directly so enforcedPlatform() consumers
// get forced versions. Constraints inherited via platform() are not enforced
// by enforcedPlatform - only direct constraints are.
```

Removing an entry from `bomDependencies` therefore changes behaviour for consumers who use `enforcedPlatform(:grails-bom)`: the dep is still managed transitively via SB's imported BOM, but is no longer **forced** by the enforcedPlatform mechanism. For deps where transitive higher versions could exist in the consumer's dependency graph, this could change resolution.

In practice this is unlikely to bite for `mongodb`, `rxjava3`, `jakarta-*` (no common deps pull higher versions transitively), but the audit should confirm against `grails-test-examples/*` builds before removing.

## What SB 4.0.6 changes vs 4.0.5

For completeness, here are the SB version bumps between 4.0.5 and 4.0.6 that touch any of our pinned modules:

| Module | SB 4.0.5 | SB 4.0.6 | Our pin | Effect |
|---|---|---|---|---|
| `mongodb` | 5.6.4 | **5.6.5** | 5.6.4 | Was SAME, now -1 patch behind |
| `groovy` | 5.0.4 | 5.0.5 | 4.0.32 | Still major-divergent (intentional) |
| `graphql-java` | 25.0 | 25.0 | 25.0 | Still SAME - tripwire holds |
| `byte-buddy` | 1.17.8 | 1.17.8 | 1.18.8 | Still divergent (Mockito) |
| `mockito` | 5.20.0 | 5.20.0 | (not pinned) | No change to byte-buddy rationale |
| `commons-codec` | 1.19.0 | 1.19.0 | 1.21.0 | Still divergent |
| `commons-lang3` | 3.19.0 | 3.19.0 | 3.20.0 | Still divergent (CVE) |
| `selenium` | 4.37.0 | 4.37.0 | 4.38.0 | Still divergent |
| `jakarta-servlet` | 6.1.0 | 6.1.0 | 6.1.0 | Still SAME |
| `jakarta-validation` | 3.1.1 | 3.1.1 | 3.1.1 | Still SAME |
| `junit-jupiter` | 6.0.3 | 6.0.3 | 6.0.3 | Still SAME |
| `rxjava3` | 3.1.12 | 3.1.12 | 3.1.12 | Still SAME |

The mongodb shift is the only material change; every other recommendation in this issue is unaffected by the 4.0.5 → 4.0.6 upgrade.

## References

- Spring Boot 4.0.6 dependencies POM: https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/4.0.6/spring-boot-dependencies-4.0.6.pom
- Spring Boot 4.0.5 dependencies POM (for comparison): https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/4.0.5/spring-boot-dependencies-4.0.5.pom
- JUnit BOM 6.0.3: https://repo1.maven.org/maven2/org/junit/junit-bom/6.0.3/junit-bom-6.0.3.pom
- Spock BOM 2.3-groovy-4.0: https://repo1.maven.org/maven2/org/spockframework/spock-bom/2.3-groovy-4.0/spock-bom-2.3-groovy-4.0.pom
- PR #15673 (re-added the graphql-java tripwire pin)
- Issue #15324 (root cause of explicit Groovy 4 pins)
- CVE-2025-48924 (commons-lang3 3.18.0 fix)

jamesfredley:
## Current summary

This PR removes `grails-bom` version pins that duplicate versions already managed by the Spring Boot 4.1.0 `spring-boot-dependencies` BOM, while keeping intentional divergences pinned. It also updates the legacy `grails-shell-cli` BOM reader so versions that now live only behind imported BOMs, especially Spring Boot-managed versions, are still visible to CLI dependency management and app creation.

After review feedback, imported BOM resolution now fails fast with the failing BOM coordinates instead of silently dropping all managed versions behind a failed import. The shell CLI coverage now exercises third-party BOM recursion, duplicate/cycle protection, nested imported BOMs, first-writer precedence for Grails-owned version properties, and the fail-fast path.

Fixes #15674

> Note: the issue was drafted against Spring Boot 4.0.6. This PR re-audits against the **4.1.0** BOM actually on `8.0.x`, so the SAME/DIVERGENT buckets differ from the issue. For example, `commons-lang3` and `byte-buddy` have since converged with Spring Boot and are now removable.

## Removed (each equal to Spring Boot 4.1.0's managed version)

| Pin | Version | How it's still managed |
|-----|---------|------------------------|
| `byte-buddy` (+ `byte-buddy-agent`) | 1.18.10 | spring-boot-dependencies |
| `commons-codec` | 1.21.0 | spring-boot-dependencies |
| `commons-lang3` | 3.20.0 | spring-boot-dependencies |
| `jakarta.servlet-api` | 6.1.0 | spring-boot-dependencies |
| `jakarta.validation-api` | 3.1.1 | spring-boot-dependencies |
| `junit` / `junit-jupiter` (12 `junit-*` entries + `junit-bom` platform import) | 6.0.3 | Spring Boot already imports `junit-bom:6.0.3` |
| `mongodb` (bson, driver-core, driver-sync, bson-record-codec) | 5.8.0 | spring-boot-dependencies |
| `rxjava3` | 3.1.12 | spring-boot-dependencies |

Also removed the `jakarta.servlet` / `jakarta.validation` excludes from `grails-bom/base/build.gradle`. They only existed to win a since-resolved selenium-vs-Spring-Boot conflict, and both versions now match Spring Boot.

## Kept intentionally

- **`graphql-java` 25.0** - retained as a deliberate drift tripwire (rationale comment added), because `graphql-java-extended-scalars` is **not** managed by Spring Boot and must move in lockstep.
- **`groovy` 4.0.32** - intentional major divergence vs Spring Boot's 5.0.6 (Grails 8 baseline / Spock 2.4-groovy-4.0).
- **`selenium` 4.38.0** - left pinned. It is now behind Spring Boot's 4.43.0, so dropping it would be a version bump (Geb/functional impact), not a pure inheritance. Worth a separate follow-up decision.
- **`spring-retry` 2.0.11** - Spring Boot 4 dropped management.

## Legacy CLI fix (required so this does not regress)

`grails-shell-cli` builds its own dependency management by parsing the published `grails-bom` POM, and previously **skipped** third-party imports such as `spring-boot-dependencies` (there was a test enforcing this). After this cleanup those versions only exist behind the Spring Boot import, so `GrailsDependencyVersions` now follows third-party imported BOMs too, with:

- **cycle/duplicate protection** (`spring-boot-dependencies` is imported by both `grails-bom` and `grails-base-bom`),
- **first-writer-wins precedence** so Grails' own pinned versions and Grails-owned version properties (for example, Groovy 4.0.32) still win over Spring Boot's,
- **per-POM `${...}` property resolution** (fixes a latent property-table clobbering bug during recursion), and
- **fail-fast imported BOM resolution** so a broken import cannot silently remove every managed version behind that BOM.

The spec test that asserted third-party BOMs are never resolved was inverted; precedence, cycle-protection, nested-import, version-property precedence, and fail-fast coverage were added.

## Verification

- Regenerated the `grails-bom` and `grails-base-bom` POMs and diffed against baseline: **only the removed constraints disappear**; the `spring-boot-dependencies` import remains, so consumers still get the versions.
- Resolved compile/runtime/test classpaths of representative modules (`grails-data-mongodb-core`, `grails-codecs-core`, `grails-async-rxjava3`, `grails-converters`, `grails-web-core`): every removed dependency resolves to Spring Boot's identical version.
- `./gradlew validateDependencyVersions` passes across **all** modules.
- `./gradlew :grails-shell-cli:test --tests "org.grails.cli.boot.GrailsDependencyVersionsSpec"` passes.
- `./gradlew :grails-shell-cli:test --no-build-cache` passes.


jamesfredley:
We need to figure out if selenium 4.38.0 should still be behind Spring Boot.

codeconsole:
Doe the spec have to hard code the spring bom version or can it resolve whatever the current one used is? 4.1.x?

jamesfredley:
@codeconsole Good call - the spec no longer hard-codes the Spring Boot BOM version (or any other version).

These specs drive a mocked `GrapeEngine` that matches imported BOMs by module name (and sometimes type), never by version, so every version literal in the fixtures (`4.1.0`, `5.8.0`, `4.0.32`, etc.) was just a synthetic placeholder that happened to look like a real release - not a value resolved from the real BOM. They were never asserting against actual managed versions, which is exactly why they'd be a maintenance burden to keep "current."

I've reworked `GrailsDependencyVersionsSpec` so each feature declares arbitrary placeholder version variables, injects them into the synthetic POMs via GStrings, and asserts against those same variables. The specs now verify the parse/import-resolution mechanism and precedence (e.g. a Grails-pinned Groovy still wins over the Spring Boot-managed one) rather than any specific release, so they stay green no matter how the real managed versions move (4.1.x and beyond).

Pushed in 635df22.

codeconsole (approved):
Reworking the spec to use placeholder versions instead of hard-coded releases is the right move. Mergeable as-is; the notes below are mostly hardening.

Thanks for the spec rework in `635df22` — that resolves my earlier question about hard-coded BOM versions.

### Recommendations (prioritized)

**1 — Highest value: missing test for `versionProperties` precedence (the `create-app` path).**
`CreateAppCommand` reads `profileDependencyVersions.versionProperties.get('groovy.version')` (and `gorm.version`, `grails-gradle-plugins.version`) to stamp a **newly created app's** build. This PR changes `getVersionProperties()` from "replaced per POM" to an accumulated `putIfAbsent` merge that now also recurses into `spring-boot-dependencies`, which declares a conflicting `groovy.version` (5.0.x). It works today only because the root `grails-bom` declares `groovy.version` (4.0.32) in its own `<properties>` and is merged before the Spring Boot recursion runs — first-writer-wins protects it. But that invariant has **no test coverage**; every new spec asserts on `find(...)`, none on the `versionProperties` map. If it ever broke, `grails create-app` would silently stamp Grails 8 apps with Groovy 5. Suggest a spec where both the Grails BOM and a recursed third-party BOM declare the same `*.version` property, asserting the Grails value wins in `versions.versionProperties`.

**2 — Silent-failure blast radius in `resolveImportedBom`.**
The empty `catch (Exception e) { }` ("skip it") predates this PR, but its impact is now much larger: Spring Boot's versions exist **only** behind the recursed import, so a silently-swallowed resolution failure drops *all* Spring Boot-managed versions at once, with no signal. Logging isn't idiomatic in this module, so this is a judgment call — at minimum worth a comment acknowledging the wider blast radius, or a debug log if a logger is acceptable here.

**3 — Document the precedence-ordering assumption.**
"Grails imports before third-party imports" only separates `org.apache.grails` from everything else. Among third-party imports (`groovy-bom`, `selenium-bom`, `spock-bom`, `spring-boot-dependencies`), precedence is just POM declaration order. It's safe *today* for the conflict-prone artifacts (Groovy, Spock) only because those are **direct** constraints in `dependencies.gradle`, so first-writer-wins pre-empts every import. A one-line comment on the resolution loop noting "direct constraints above protect us; ordering among third-party imports is otherwise undefined" would stop a future maintainer from moving a Groovy pin into a BOM import and silently regressing it.

**4 — Minor: `find(String artifactId)` single-arg now resolves first-writer for ambiguous artifactIds.**
`io.reactivex.rxjava2:rxjava` and `io.reactivex.rxjava3:rxjava` share artifactId `rxjava`; the new first-writer guard flips single-arg `find('rxjava')` from last- to first-writer. The two-arg `find` (used everywhere internally) is unaffected, so impact is near-zero — just flagging it.

**5 — Nit: the `!version.startsWith('${')` guard is nearly dead.**
After `versionLookup`, `version` is either resolved, a literal, or `null`. The meaningful check is the `null` one (skip unresolved-property deps); `startsWith('${')` only triggers on a malformed `${foo` with no closing brace. Harmless; the comment could just say "skip deps whose version property didn't resolve."

### Verified and fine
- Cycle/duplicate protection (`resolvedBoms`, keyed `group:artifact:version`) correctly handles `spring-boot-dependencies` being imported by both `grails-bom` and `grails-base-bom` — covered by the "resolved exactly once" spec.
- The per-POM `localProperties` change genuinely fixes the recursion property-clobbering bug and is more Maven-accurate (properties resolve in their declaring POM).
- Removing the `jakarta.servlet` / `jakarta.validation` excludes is consistent with removing the matching direct pins — both now inherit `6.1.0` / `3.1.1` from Spring Boot.
- Groovy `4.0.32` precedence holds because all groovy artifacts are direct constraints.

### Open item the PR already flags
`selenium` is left pinned at `4.38.0`, now *behind* Spring Boot's `4.43.0`. Dropping it would be a version bump with Geb/functional impact rather than pure inheritance, so deferring to a separate follow-up is reasonable.

jamesfredley:
Thanks for the thorough review @codeconsole. Pushed `791565333e` addressing the hardening notes:

**1 - `versionProperties` precedence (the `create-app` path).** Added a `GrailsDependencyVersionsSpec` feature that wires a Grails BOM declaring `groovy.version` and importing a Spring Boot BOM that declares a *conflicting* `groovy.version`, then asserts `versionProperties['groovy.version']` resolves to the Grails value (first-writer-wins). This is exactly the invariant `CreateAppCommand` relies on when stamping a new app, and it now has direct coverage so a regression would fail the build instead of silently emitting the wrong Groovy.

**2 - Silent-failure blast radius in `resolveImportedBom`.** Expanded the `catch` comment to call out that the swallowed-failure impact is now much wider than when only Grails BOMs were followed - Spring Boot's managed versions exist *only* behind the recursed import, so one silent failure drops them all. Kept it as a comment since logging isn't idiomatic in this module, as you noted.

**3 - Precedence-ordering assumption.** Added a comment on the resolution loop noting that ordering among third-party imports is just POM declaration order and otherwise undefined, and that conflict-prone artifacts (Groovy, Spock) stay correct only because they're direct constraints that first-writer-wins pre-empts - with a warning not to move such a pin into a BOM import without restoring an equivalent direct constraint.

**5 - Near-dead `startsWith('${')` guard.** Clarified the comment to say the `null` check is the meaningful guard (`versionLookup` returns `null` for an unresolved `${...}` reference); `startsWith('${')` only catches a malformed reference with no closing brace.

**4 - Single-arg `find(String)` first-writer flip.** Left as-is - as flagged, the two-arg `find` used internally is unaffected and the impact is near-zero, so no behavior change seemed warranted.

The change is comments plus the one new spec (no production behavior change). `./gradlew :grails-shell-cli:test --tests "org.grails.cli.boot.GrailsDependencyVersionsSpec"` is green (8/8, including the new feature).

The `selenium` pin remains deferred as a separate follow-up, per the open item already noted on this PR.

jamesfredley:
@jdaugherty the extra stuff in here is the fix shell-cli, which is required when inheriting some of these versions from spring-boot-dependencies.

codeconsole:
1 New spec: a Grails BOM declaring groovy.version + importing a Spring Boot BOM that declares a conflicting groovy.version, asserting versionProperties['groovy.version'] stays the Grails value -  verified

2  Expanded the catch comment: Spring Boot's versions exist only behind the recursed import, so one swallowed failure drops them all; logging left out as non-idiomatic  -  verified

3 Added a comment on the resolution loop: third-party import order is undefined; Groovy/Spock stay correct only as direct constraints — with a "don't move a pin into a BOM import" warning    -  verified      

4 two-arg find unaffected, near-zero impact -  verified

5  null check is the meaningful guard  -  verified

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
```diff
@@ -85,27 +86,62 @@ class GrailsDependencyVersions implements DependencyManagement {
...
+
+        List<Map<String, String>> grailsImports = []
+        List<Map<String, String>> thirdPartyImports = []
+
         pom.dependencyManagement.dependencies.dependency.each { dep ->
             String groupId = dep.groupId.text()
             String artifactId = dep.artifactId.text()
-            String version = versionLookup(dep.version.text())
+            String version = versionLookup(dep.version.text(), localProperties)
             String scope = dep.scope.text()
```
jdaugherty: This is a lot of logic for resolving what should be recursive. It was claimed this was to fix the shell cli.  Can you please elaborate how this was broken?

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
```diff
@@ -115,7 +151,11 @@ class GrailsDependencyVersions implements DependencyManagement {
                 addDependencyManagement(importedPom)
             }
         } catch (Exception e) {
-            // If the imported BOM cannot be resolved, skip it
```
jdaugherty, line 118: Does this actually occur in the build?  I assume this exception should fail the build

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
```diff
@@ -85,27 +86,62 @@ class GrailsDependencyVersions implements DependencyManagement {
 
     @CompileDynamic
     void addDependencyManagement(GPathResult pom) {
-        versionProperties = pom.properties.'*'.collectEntries { [(it.name()): it.text()] }
```
jdaugherty, line 88: Isn't the core issue that we are trying to manual extract the pom with xml parsing?  The maven model solves all of this, no?

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
jamesfredley: Agreed. In the healthy build this should not occur, and if it does occur the build/CLI setup should fail rather than continue with incomplete dependency management. Pushed a5a70ec to make imported BOM resolution fail fast with the failing coordinates, and updated GrailsDependencyVersionsSpec to assert that behavior and preserve the original cause.

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
jamesfredley: The Maven model builder is the broader cleanup direction, and the shell CLI already uses it for @DependencyManagementBom. I kept this PR narrower because GrailsDependencyVersions also exposes legacy versionProperties used by create-app plus the existing group/artifact lookup contract, so replacing it with the model path would be a larger refactor than the BOM-pin cleanup. This patch keeps the existing contract, adds recursive imports and precedence protection needed for the removed pins, and now matches the model-builder behavior by failing fast on unresolved imported BOMs.

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
jamesfredley: Thanks, yes. The shell CLI breakage is that GrailsDependencyVersions is the dependency-management source used by MavenProfileRepository/CreateAppCommand and the CLI DependencyResolutionContext. Once this PR removes direct pins, versions such as mongodb, junit, rxjava3, byte-buddy, and the Jakarta APIs live only behind spring-boot-dependencies; without recursing into that import, the legacy CLI no longer exposes those managed versions. The added specs cover a Spring Boot-only managed dependency, duplicate imports, nested BOM imports, and Grails-first precedence.

jamesfredley:
Addressed the remaining feedback from @jdaugherty and pushed `a5a70ec91b`.

What changed:

- `GrailsDependencyVersions` now fails fast when an imported BOM cannot be resolved, including the failing BOM coordinates and preserving the original cause.
- `GrailsDependencyVersionsSpec` now asserts that fail-fast behavior, so a broken imported BOM cannot silently produce incomplete CLI dependency management.
- The PR description has been updated at the top to reflect the current behavior, including the fail-fast imported BOM handling.
- Replied to the three inline review comments and marked those threads resolved.

Verification:

- `./gradlew :grails-shell-cli:test --tests "org.grails.cli.boot.GrailsDependencyVersionsSpec"` passed.
- `./gradlew :grails-shell-cli:test --no-build-cache` passed.
- Oracle review gate returned GREEN for the exact pushed diff.

Note: the first full `:grails-shell-cli:test` run executed the tests successfully but failed afterward in `jacocoTestReport` because another Java process held the Gradle journal cache lock. Rerunning with `--no-build-cache` completed successfully.

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
jdaugherty: I think we need to make this change.  If you are too busy, I'm happy to offer assistance. @jamesfredley

jdaugherty: I'm not sure why this resolved.  The change needing to be made here is to move away from custom xml parsing.  This is not yet done in this PR.

jdaugherty:
@jamesfredley  I don't think the feedback / solution we discussed has been completed on this PR.  it needs another pass.

grails-shell-cli/src/main/groovy/org/grails/cli/boot/GrailsDependencyVersions.groovy
jamesfredley: Addressed in b18a036167. GrailsDependencyVersions now resolves BOMs through Maven DefaultModelBuilder instead of custom XML traversal, while preserving versionProperties from imported BOM models with first-writer-wins semantics for create-app.

jamesfredley:
Pushed b18a036167 addressing the remaining feedback from @jdaugherty. GrailsDependencyVersions now uses Maven DefaultModelBuilder effective-model resolution instead of custom XML BOM traversal, preserves first-writer-wins versionProperties including properties declared only by imported BOMs, keeps fail-fast imported BOM resolution, and refreshes GrailsDependencyVersionsSpec coverage for Spring Boot-managed dependencies, duplicate imports, transitive imports, imported-only properties, and failure propagation. Verified with :grails-shell-cli:test --no-build-cache --max-workers=2 -PmaxTestParallel=2, :grails-shell-cli:installDist --max-workers=2 -PmaxTestParallel=2, and the installed grails-shell-cli list-profiles command.

jdaugherty (approved):

codeconsole (approved):

PR #15731, opened 2026-06-11
Expand Hibernate 7 functional coverage and query parity

jamesfredley:
## Summary

This PR expands the Hibernate 7 functional-test path and closes several parity gaps found while running the functional applications against Hibernate 7. The branch targets `8.0.x-hibernate7` and includes the functional matrix work plus follow-up fixes, documentation, Hibernate ORM 7.4 alignment, and review-gate cleanup.

The main outcome is that the GORM behavior gaps found while exercising Hibernate 7 are now covered by tests that run on **both** Hibernate 5.6 and Hibernate 7.4. Per review feedback (see the discussion below), the earlier runtime BOM-substitution lane (`-PhibernateVersion=7`) has been **removed** - it broke Gradle build caching, dependency debugging, and coverage attribution. Shared, version-agnostic GORM behavior now lives in the shared `grails-datamapping-tck`, which both `grails-data-hibernate5` and `grails-data-hibernate7` already execute via the `GrailsDataTckManager` ServiceLoader, so the same specs run natively against both Hibernate lines with no dependency switching. The existing `hibernate7Functional` job (which runs the H7-specific example apps) now gates `publish`. Coverage that genuinely cannot move to the TCK will be added as dedicated H7 example specs as a tracked follow-up.

Spring Boot 4.1 now manages Hibernate ORM 7.4, and Hibernate ORM 7.4 is the current stable Hibernate 7 line. Hibernate ORM 7.3 and 7.2 are limited-support lines, so this PR updates the Hibernate 7 path to Hibernate ORM `7.4.1.Final` while keeping Hibernate Tools on `7.3.8.Final` because no Hibernate Tools 7.4 artifact is published.

## Detailed issue coverage

| Area | Issue | Why it is an issue | How it manifests | Fix in this PR |
|---|---|---|---|---|
| Functional CI publish gate | Hibernate 7 functional results were not part of the publish dependency chain. | CI could publish artifacts even when the H7 functional lane had not passed. | The `publish` job did not wait on Hibernate 7 functional results. | Adds the existing `hibernate7Functional` job (which runs the `grails-test-examples/hibernate7/*` apps) to the `publish` `needs`/`if` gate so Hibernate 7 functional results block publishing. |
| H7 GORM behavior coverage | Hibernate 7 regressions were hidden because the broad suite mostly validated the Hibernate 5 path. | The version-agnostic GORM contracts (finders, getAll, query settings) had no test that ran against the Hibernate 7 line. | H7-only regressions slipped through until app boot. | Moves shared GORM behavior into `grails-datamapping-tck`, which both Hibernate modules run natively, so each contract is verified on **both** H5.6 and H7.4 with no runtime dependency substitution. (Replaces the removed `-PhibernateVersion=7` BOM-substitution lane that reviewers `-1`'d.) |
| Hibernate 7 support line | The branch previously targeted Hibernate ORM 7.2 even though Spring Boot 4.1 manages Hibernate ORM 7.4. | Hibernate ORM 7.3 and 7.2 are limited-support lines, so new H7 work should target the actively maintained Hibernate 7 line. | Grails could appear to certify a limited-support Hibernate line while Boot-based applications resolve a newer ORM line. | Moves the Hibernate 7 path to Hibernate ORM `7.4.1.Final` and documents that Hibernate 7.4 is the target line. |
| Hibernate Tools alignment | Hibernate Tools does not currently publish a 7.4 artifact matching Hibernate ORM 7.4. | Reusing the ORM version for Tools would resolve a non-existent artifact. | Dependency resolution can fail for `org.hibernate.tool:hibernate-tools-orm:7.4.1.Final` or `hibernate-tools-utils:7.4.1.Final`. | Adds a separate `hibernate-tools.version` and pins Hibernate Tools to `7.3.8.Final` while ORM uses `7.4.1.Final`. |
| Hibernate 7.4 metadata API | Hibernate 7.4 adds an `isView` argument to `InFlightMetadataCollector.addTable(...)`. | Existing Grails domain binding code and tests called the older six-argument method. | Compilation fails in collection, root, joined-subclass, and second-pass binding tests. | Updates production binding calls and Spock mocks to pass `isView=false`. |
| Database migration bootstrap API | Hibernate 7.4 eliminates raw or string-only map typing in JPA bootstrap contracts. | The database migration integration used `Map<String, String>` for settings that Hibernate now accepts as `Map<String, Object>`. | `grails-data-hibernate7-dbmigration-core` can fail Java compilation against Hibernate 7.4. | Widens the bootstrap settings map to `Map<String, Object>`. |
| Hibernate proxy interception | The H7 proxy interceptor carried a fallback workaround for older ByteBuddy proxy behavior. | The workaround duplicated Hibernate 7.4 proxy invocation behavior and touched proxy internals that are no longer needed. | Proxy method interception could diverge from the Hibernate 7.4 `ByteBuddyInterceptor` path. | Removes the obsolete fallback and delegates to Hibernate's current invocation path after the Grails uninitialized-proxy guards run. |
| Simple identifier primary keys | `SimpleIdBinder` installed an empty `PrimaryKey` before Hibernate created the table primary key. | Hibernate 7.4 expects `RootClass.createPrimaryKey()` to build the populated primary key from the bound identifier columns. | Schema metadata can end up with an empty or incorrect primary key. | Stops pre-installing the empty primary key and adds coverage that Hibernate creates a one-column table primary key after identifier binding. |
| Composite identifier associations | Composite-id many-to-one and dependent collection keys needed stricter column ordering under Hibernate 7.4. | Hibernate 7.4 validates composite key and foreign-key column ordering more strictly. | Composite-id specs can fail with mismatched FK columns, sequence/composite interactions, or unstable dependent key ordering. | Sorts referenced component identifiers, sorts or indexes local FK columns consistently, creates FK metadata with matching referenced columns, sorts bidirectional one-to-many keys, and avoids marking dependent component keys sorted too early. |
| Identity generator tests | Hibernate 7.4 generator creation needs type information from the generator context. | The spec mocks did not provide the type metadata used by the current generator path. | `GrailsIdentityGeneratorSpec` fails even though production generator behavior is valid. | Stubs `GeneratorCreationContext.getType().getReturnedClass()` in the affected specs. |
| Event listener tests | Hibernate 7.4 event/session access calls `EventSource.asEventSource()` and `getFactory()` in this path. | The event listener spec mock did not behave like a current Hibernate event source. | `HibernateEventListenerSpec` fails during listener invocation. | Updates the mock event source to return itself from `asEventSource()` and the mocked session factory from `getFactory()`. |
| Chained transaction manager (multi-datasource) | With multiple data sources and `chainedTransactionManager.enabled`, every `@Rollback` transaction failed under Hibernate 7 with `IllegalTransactionStateException: Transaction is already completed`. | Under Hibernate 7 a second bean name aliases the same primary transaction-manager instance, so it appears twice in the chained manager. `MultiTransactionStatus` is keyed by instance, so it was committed/rolled back twice. | An identical multi-datasource app passes on H5 but fails 30/30 on H7; a real blocker for users migrating multi-datasource apps. | `ChainedTransactionManager` now dedupes delegate managers by identity (no-op when there is no duplicate, e.g. H5). Locked with a unit test and verified by the new H7 `datasources` functional app. |
| Datasources functional parity | The general `datasources` app (datasource switching, cross-datasource transactions, OSIV GSP rendering) ran only on Hibernate 5. | No Hibernate 7 functional coverage for multi-datasource/OSIV behavior. | H7 multi-datasource regressions slipped through. | Adds `grails-test-examples/hibernate7/datasources`, a near-identical clone of the H5 app (only `build.gradle` differs). All four integration specs pass on Hibernate 7, matching H5. |
| Eager fetch joins (`fetch:[assoc:'join']`) | Under Hibernate 7, `findById`/dynamic finders with `fetch:[assoc:'join']` did not eagerly initialize the association, while Hibernate 5 did. | GORM's `DynamicFinder` turns an EAGER fetch into `query.join(property)`; the rewritten H7 query layer materialized it as a plain JPA `root.join` only when referenced in a predicate (and a fetch-only association never is), so it was never fetched. JPA needs `root.fetch(...)`. | JSON APIs silently drop hasMany collections after migration. | `HibernateQuery` records fetch-join paths; `JpaCriteriaQueryCreator` materializes them as `root.fetch(path, joinType)` for full-entity selects (count/projection queries skipped). Locked with `FetchJoinSpec` in both modules; full H7 core suite green (2944 tests). |
| Proxy / duplicate-entity id rendering in JSON views | Under Hibernate 7, a JSON view rendered an entity loaded via a Hibernate proxy, or appearing more than once in the object graph, WITHOUT its id (and de-duplicated a repeat to an empty `{}`), while Hibernate 5 rendered it. | `DefaultJsonViewHelper.findEntity` used `object.getClass()`, which for a proxy is the generated subclass and is not a registered entity, so the proxy rendered as a plain POGO (losing its id). Separately, `preProcessedOutput` reused a cached, context-bound `JsonWritable` for repeated occurrences. | REST/JSON responses could silently omit identifiers for proxy-loaded or repeated entities. | `findEntity` now unwraps the proxy before resolving the entity class; `preProcessedOutput` renders each occurrence fresh (cycles are still broken by the circular-association handling in `process()`). Verified by 175 grails-views-gson unit tests and the H7 views app (52/52). |
| Non-transactional BootStrap saves not persisted | On Hibernate 7, GORM saves in a non-transactional `BootStrap` `init` closure were silently discarded; Hibernate 5 persisted them. | `executeGrailsBootstraps` binds a session via the `PersistenceContextInterceptor`, runs the closures, then calls `interceptor.flush()`. The H7 interceptor's `flush()` was a no-op without active transaction synchronization (which a non-transactional BootStrap lacks), so the owned session was never flushed/committed and `destroy()` discarded the inserts. | Apps that seed data in a non-transactional BootStrap start empty on H7. | The H7 `HibernatePersistenceContextInterceptor.flush()` now flushes and commits its owned session when it is not participating in an outer one and no synchronization is active. Locked with a `HibernatePersistenceContextInterceptorSpec` regression test; resolved the scaffolding bootstrap-data failures (123/136 -> 134/136) with `views-functional-tests` still 52/52. |
| H7 general-app functional coverage | Many functional apps exercise GORM/Hibernate behavior (queries, associations, proxy/scaffolding rendering, bootstrap seeding, transactions) but ran only on Hibernate 5. | No H7 functional coverage for those areas. | H7 regressions in queries, rendering, or app boot could slip through. | Adds near-identical Hibernate 7 clones of every GORM-bearing example app and verifies them green on Hibernate 7: `gorm` (255), `app1` (688), `async-events-pubsub-demo` (17), `database-cleanup` (8), `demo33` (3), `issue-698-domain-save-npe` (3), and `scaffolding-fields` (**136/136**), alongside the previously-landed `datasources` (4/4) and `views-functional-tests` (52/52). Cloning these surfaced and fixed three H7 regressions (rows above): non-transactional BootStrap flush, bidirectional one-to-many List duplicate-column boot, and lazy belongsTo display rendering. |
| Forge analytics dependencies | Forge analytics Postgres was using the old Micronaut BOM coordinate. | The old coordinate does not match the platform coordinate used by current Micronaut dependency management. | H7 functional/Forge dependency resolution can fail or pull inconsistent dependency constraints. | Switches to `io.micronaut.platform:micronaut-platform` and centralizes the Testcontainers version used by the Forge H7 path. |
| H7 application boot | Some functional test apps still carried Hibernate 5 specific configuration. | H5-specific cache/autoconfiguration prevents the same app from proving H7 behavior. | H7 app boot fails before specs can exercise GORM behavior. | Registers H7 autoconfiguration, removes H5-specific Ehcache assumptions, and updates H7 functional app configuration. |
| Plain String HQL in single-argument overloads | Earlier in this PR the single-argument `find`/`findAll`/`executeQuery`/`executeUpdate(CharSequence)` overloads rejected plain Strings (a GString-only guard), diverging from the Hibernate 5 calling convention. | The guard threw `UnsupportedOperationException` for plain-String HQL. | `executeQuery("from Book where inStock = true")` and similar plain-String calls failed where they worked on Hibernate 5. | Removes the guard so plain Strings work exactly as on Hibernate 5; injection safety is preserved by binding every GString `${value}` as a named parameter in `HqlQueryContext` (so `executeQuery("... = ${userInput}")` is bound, not interpolated). Reverts the spec workarounds and updates the HQL/upgrade docs. |
| Bidirectional one-to-many List duplicate-column boot failure | An entity with `belongsTo` to a parent that has a `hasMany` `List` (e.g. `TeamMember`/`Team`) failed to boot on Hibernate 7 with `MappingException: Column 'team_id' is duplicated`. | `ListSecondPassBinder` left the synthetic `Backref`/`IndexBackref` `insertable=true` for one-to-many Lists (only many-to-many got `insertable=false`); Hibernate 7's `checkColumnDuplication` then treats the back-reference column as a real mapping and rejects the duplicate. | Apps with this common mapping fail to boot on H7 (app1 was 0/688). | Sets `insertable=false` unconditionally on `Backref`/`IndexBackref`; the owning many-to-one manages the FK column. app1 now boots and passes 688/688 on H7. |
| Lazy belongsTo display in scaffolded views | A scaffolded show view rendered a lazy `belongsTo` association as `entityName#id` (e.g. `scaffoldingfields.Department:1`) instead of its display value on Hibernate 7. | Hibernate 7's `AbstractLazyInitializer.toString()` returns `entityName#id` without initializing the proxy (Hibernate 5's proxy initialized on `toString()`), and the Fields plugin rendered that raw value. | Scaffolded show pages display a proxy identifier instead of the associated object on H7. | `BeanPropertyAccessorFactory` now unwraps the proxy via the existing GORM `ProxyHandler` when resolving the property value, so the real display value renders. scaffolding-fields now passes 136/136 on H7. |
| DetachedCriteria single-result behavior | Hibernate 7 throws when `getSingleResult()` returns more than one row. | GORM's historical behavior is to return the first matching row for single-result helpers where multiple rows can match. | `DetachedCriteria.get()` can throw `NonUniqueResultException` in H7 where H5 returned one result. | Adjusts H7 query handling so single-result paths limit to one row or otherwise match the GORM contract. |
| Managed collection identity | Hibernate 7 is stricter about multiple representations of the same collection. | GORM add-to/save flows must not leave duplicate collection wrappers in the session. | Cascade/add-to functional specs can fail with `Found two representations of same collection`. | Fixes the H7 managed-entity add-to/save path so the bidirectional collection state remains consistent. |
| Aggregate query result typing | H7 enforces selected result type more strictly than H5. | Aggregate HQL like `avg`, `max`, `min`, and `sum` does not return the domain entity type. | Data service aggregate queries fail with incorrect result type errors. | Routes aggregate/projection HQL through query creation that lets Hibernate return the correct scalar type. |
| Cross-property arithmetic | H7 SQM typing is stricter for numeric property comparisons. | GORM where queries comparing differently typed numeric expressions need explicit type handling. | Expressions such as `pageCount > price * 10` can fail with coercion errors. | Updates the H7 where/query handling so cross-property arithmetic comparisons work under H7. |
| `findWhere` result limiting | H7 `findWhere` did not reliably limit duplicate matches to a single row. | `findWhere` is a single-result convenience API and should not allow duplicate matches to produce multiple-row failures. | Duplicate rows matching the property map can throw or behave differently from H5. | Forces `findWhere` to set `max: 1` internally and adds SQL-capture coverage proving the generated query includes the row limit. |
| Null values in `findWhere` and `findAllWhere` | The generated HQL treated null map values as named parameters. | In SQL/HQL, `= null` does not match null rows; it must be `is null`. | `findWhere(nullableProperty: null)` and `findAllWhere(nullableProperty: null)` fail to find rows with null values. | Generates `is null` for null values and removes those values from named parameters. |
| Property-map query safety | Property names from `findWhere` and `findAllWhere` were interpolated into HQL without validation. | Property-map keys are caller input at the API boundary and should not become arbitrary HQL fragments. | Invalid or malicious keys could produce malformed HQL or unsafe generated query text. | Validates every key against the persistent entity before building HQL and tests invalid property names. |
| `getAll` ordering with convertible ids | H7 converted ids before querying but reconstructed results with the original id values. | When callers pass string ids for numeric identifiers, the returned row map is keyed by converted ids. | `getAll(["3", "1", "2"])` can return nulls or the wrong order even though those rows exist. | Reconstructs the result list using converted ids and documents that existing ids preserve caller order. |
| HQL query settings | H7 HQL helpers only partially applied common GORM query settings. | Users expect settings such as `fetchSize`, `timeout`, `readOnly`, `cache`, `max`, `offset`, `flushMode`, and `lock` to behave consistently. | String values for `max` or `offset` can fail casts, `cache` might not be applied, and `lock` was not filtered from named parameters. | Converts string/number/boolean setting values, applies all supported settings, maps `lock: true` to pessimistic write locking, disables query cache for locked queries, and filters `lock` from query parameters. |
| TCK spec migration | Some migrated H7 TCK specs had missing imports or stale registration code. | The H7 TCK specs must compile and exercise the same public behavior as the source TCK specs. | Specs fail at compile time or register the wrong domain set. | Restores required `HibernateGormDatastoreSpec` imports, removes duplicate setup, and narrows Spock imports. |
| H7 has-many support class | `Something.groovy` referenced `Book` without the required imported package. | Test support classes must compile in isolation under the H7 module. | Compilation can fail because `Book` is in `grails.gorm.tests.hasmany`, not the support class package. | Adds the explicit `Book` import. |
| Multitenancy package scanning | The multitenancy spec scanned the spec package instead of the package containing the support domain and services. | The datastore only discovers classes in the package it scans. | Services or domain classes can be missing when the spec starts the datastore. | Imports the support classes explicitly and scans `Department.getPackage()`. |
| Public documentation | The H7 behavior changes were not reflected in the user-facing docs. | Users need documented contracts for query settings, finder null handling, property validation, lock behavior, `getAll` ordering, and remaining Hibernate 7.4 migration differences. | The reference pages understate supported settings or omit the new null/order behavior and remaining Hibernate 7.4 migration checks. | Updates the Hibernate 7 guide and Grails reference docs for HQL settings, `lock`, `findWhere`, `findAllWhere`, `getAll`, and the remaining Hibernate 7.4 migration checklist. |
| Historical bug report | `H7_GORM_BUG_REPORT.md` was useful triage context but lacked an Apache header and read like current expected failure status. | New source files need ASF licensing, and stale present-tense failure statements mislead reviewers. | RAT/compliance review would fail, and readers could think fixed failures are still expected. | Adds the Apache/SPDX Markdown header, frames the report as historical triage rationale, and normalizes dash typography. |

## Verification

- `git diff --check origin/8.0.x-hibernate7...HEAD` completed with no output.
- Dash typography search in `H7_GORM_BUG_REPORT.md` completed with no output.
- Markdown LSP diagnostics for `H7_GORM_BUG_REPORT.md` found no diagnostics.
- Java LSP diagnostics for `HqlQueryMethods.java` and `SelectHqlQuery.java` found no diagnostics.
- Groovy/AsciiDoc LSP diagnostics are not authoritative in this workspace because the Groovy server does not have the Gradle classpath and there is no AsciiDoc LSP configured.
- `./gradlew --no-daemon --no-parallel :grails-data-hibernate7-core:test --tests "org.grails.orm.hibernate.HibernateGormStaticApiFindWhereSpec" --tests "org.grails.orm.hibernate.HibernateGormStaticApiSpec" --tests "org.grails.orm.hibernate.query.HqlQueryMethodsSpec" --tests "org.grails.orm.hibernate.query.SelectHqlQuerySpec"` passed after cleaning `:grails-core`: 130 tests, 130 successes, 0 failures.
- `./gradlew :grails-data-hibernate7-core:test --rerun-tasks --stacktrace` passed: 2921 tests, 0 failures.
- `./gradlew :grails-data-hibernate7-dbmigration-core:compileJava` passed.
- `./gradlew :grails-data-hibernate7-dbmigration-core:test --rerun-tasks --stacktrace` passed.
- `./gradlew aggregateViolations --continue` passed.
- `./gradlew :grails-data-hibernate7-docs:asciidoctor` passed after adding the Hibernate 7.4 migration checklist. Existing AsciiDoctor warnings remain in unrelated pre-existing docs.
- `./gradlew --no-daemon --no-parallel :grails-data-hibernate7-docs:asciidoctor :grails-doc:publishGuide -x aggregateGroovydoc` passed and built the user manual before the final Hibernate 7.4 checklist expansion.

## Known follow-up work not included here

| Project | Current state | Follow-up needed |
|---|---|---|
| `grails-test-examples-datasources` | **Done.** A near-identical H7 clone now exists at `grails-test-examples/hibernate7/datasources`; all four integration specs (datasource switching, cross-datasource transactions, multiple datasources, OSIV GSP rendering) pass on Hibernate 7, matching H5. Required the `ChainedTransactionManager` dedupe fix. | None - covered. |
| `grails-test-examples-views-functional-tests` | **Done - 52/52 on H7.** The eager-fetch-join fix and the proxy/duplicate-entity id-rendering fix (both in this PR) resolved all failures; a near-identical H7 clone now lives at `grails-test-examples/hibernate7/views-functional-tests` and passes all 52 integration specs (association/proxy/inheritance/embedded/circular rendering, deep rendering, HAL, composite ids). | None - covered. |
| `grails-test-examples-scaffolding-fields` | **Done - 136/136 on H7 (Geb).** A near-identical H7 clone now lives at `grails-test-examples/hibernate7/scaffolding-fields`. Cloning it surfaced two H7 framework regressions, both fixed in this PR: the non-transactional BootStrap flush and the lazy belongsTo display-value rendering in scaffolded show views. The two previously-flaky Geb specs were stabilised with page-object waits (wait-for-link-before-click and poll-rendered-content). | None - covered. |

## Hibernate 7.4 vs Hibernate 5.6 end-application breaking-change checklist

This table is the migration checklist for application code. It calls out the places where Hibernate 7.4 is stricter or behaves differently than Hibernate 5.6, what can break in real applications, whether this PR resolves the Grails-facing issue, and what application owners should change or verify.

| Status | Area | Hibernate 5.6 behavior | Hibernate 7.4 behavior | What can break in applications | Application migration action |
|---|---|---|---|---|---|
| Remaining app migration item | Hibernate module dependencies | Applications commonly depend on Hibernate 5 aligned Grails artifacts and constraints. | Applications must resolve the Hibernate 7 aligned artifacts and constraints together. | Mixing `grails-data-hibernate5`, H5 BOM constraints, or H5-only transitive dependencies with the H7 runtime can fail dependency resolution or application boot. | Use the Hibernate 7 BOM/artifacts consistently. Do not mix H5 and H7 GORM/Hibernate artifacts in the same application runtime. |
| Remaining app migration item | H5-specific cache setup | H5 applications may rely on Ehcache/Hibernate 5 cache integration and related configuration. | H7 no longer supports the same H5-specific cache integration path. | Boot can fail before application code runs when H5 cache classes or region-factory settings are still configured. | Remove H5-specific cache dependencies and configuration. Reintroduce caching only with H7-compatible cache providers and settings. |
| **Resolved by this PR** | **Plain String HQL in `executeQuery` and `executeUpdate`** | **H5 accepts plain `String` HQL and binds `GString` interpolations as parameters.** | **This PR restores the same on H7: a plain `String` runs as written, and every `GString` `${value}` is bound as a named parameter rather than interpolated into the query text.** | **Nothing - `Book.executeQuery("from Book where inStock = true")` works again, and `executeQuery("... = ${userInput}")` is injection-safe by binding.** | **No app change is needed. As in any ORM, do not hand-concatenate untrusted input into a plain `String`; prefer the `GString` form or an explicit params map/list.** |
| **Resolved by this PR** | Single-result queries with multiple rows | H5 paths often returned the first row for GORM single-result helpers when multiple rows matched. | Hibernate 7.4 inherits the Hibernate 7.3 `getSingleResult()` strictness and throws when more than one row matches. | Direct Hibernate/JPA calls can throw `NonUniqueResultException` where H5 returned one row. | GORM helpers fixed by this PR preserve expected first-row behavior. Direct Hibernate/JPA usage should add `maxResults(1)` or use a list query and take the first result intentionally. |
| **Resolved by this PR** | **`findWhere` duplicate matches** | **H5 `findWhere` behaves as a first-match helper.** | **The H7 implementation needs an explicit one-row limit to avoid strict single-result failures.** | **Duplicate property-map matches can fail instead of returning one instance.** | **No app change is needed when using the fixed GORM API from this PR. Direct custom single-result HQL should still add an explicit limit.** |
| **Resolved by this PR** | **Null property-map values** | **H5 GORM property-map queries treat `findWhere(prop: null)` and `findAllWhere(prop: null)` as null checks.** | **HQL requires `is null`; passing null as a named value does not match null rows.** | **Null-valued property-map lookups can return no rows or behave differently if generated as `= :param` with a null value.** | **Use the fixed GORM APIs from this PR. In custom HQL, write `is null` or `is not null` explicitly rather than `= :param` with a null parameter.** |
| Intentional behavior change | Property-map query keys | H5-era code could accidentally rely on permissive interpolation of property-map keys. | The H7 path validates keys against real domain properties before generating HQL. | Code that passes computed, misspelled, or HQL-fragment-like keys to `findWhere` or `findAllWhere` now fails fast. | Pass only real domain property names as map keys. Move dynamic predicates to criteria, where queries, or explicit HQL with bound parameters. |
| **Resolved by this PR** | **`getAll` order with converted ids** | **H5 GORM preserves caller id order, including common string-to-number id conversion cases.** | **H7 must compare results against converted identifiers, not the original raw id values.** | **`getAll(["3", "1", "2"])` can return null entries or wrong ordering if id conversion is not considered.** | **Use the fixed GORM API from this PR. For custom `in (:ids)` queries, remember SQL does not preserve input order unless you reorder results yourself.** |
| **Resolved by this PR** | Aggregate HQL result typing | H5 was more permissive about query result type mismatch. | H7 validates the selected result type strictly. | Queries like `select avg(price)` or `select max(pageCount)` can fail if they are created as typed entity queries or mapped to the wrong return type. | GORM aggregate/projection HQL paths fixed by this PR avoid entity-typed query creation where appropriate. Direct Hibernate queries should use scalar-compatible return types. |
| **Resolved by this PR** | Data service aggregate methods | H5 often tolerated loose aggregate method return typing. | H7's stricter result typing makes mismatches visible. | Data service methods returning a type that does not match the aggregate result can fail at runtime. | Review `@Query` aggregate return types. `avg` generally returns `Double`; `max` and `min` follow the selected expression type; `count` returns `Long`. |
| **Resolved by this PR** | Cross-property numeric arithmetic | H5 allowed more implicit numeric coercion in where/HQL predicates. | H7 SQM typing is stricter about comparing different numeric expression types. | Queries such as `where { pageCount > price * 10 }` can fail when operands resolve to incompatible numeric types. | GORM where-query comparison handling is fixed by this PR. Direct HQL should align domain numeric types, cast explicitly where needed, or adjust expressions so both sides have compatible types. |
| **Resolved by this PR** | Managed collection identity | H5 tolerated some duplicate representations of the same collection in a session. | H7 detects duplicate managed collection wrappers more aggressively. | `addTo*`, cascade saves, detached graph merges, or replacing collection instances can throw `Found two representations of same collection`. | Prefer mutating the managed collection instance instead of replacing it. Be careful when merging detached graphs; reload managed parents and apply changes to the managed association. |
| **Resolved by this PR** | Bidirectional association synchronization | H5 could be more forgiving when only one side of an association was updated. | H7 is less tolerant of inconsistent managed association state during flush. | Saves involving parent/child associations can fail or leave stale collection snapshots. | Keep both sides of bidirectional associations synchronized. Use GORM `addTo*` and `removeFrom*` helpers where possible. |
| Documented behavior change | Locking and query cache | H5-era code may combine cacheable query settings with lock settings. | H7 locked queries are not query-cacheable; this PR disables query caching when `lock: true` is used. | Code expecting a locked query to use the query cache will behave differently. | Do not rely on query cache for pessimistically locked queries. Treat locked queries as database reads requiring fresh row state. |
| **Resolved by this PR** | **Query setting value types** | **H5/GORM paths were often permissive about string values for settings.** | **H7 query APIs expect typed values, and the GORM adapter must coerce supported strings.** | **Settings such as `max: "10"`, `offset: "5"`, `readOnly: "true"`, `fetchSize: "50"`, or `timeout: "30"` can fail if passed directly to strict Hibernate APIs.** | **With the fixed GORM APIs, common string settings are coerced. In direct Hibernate usage, pass typed `int` and `boolean` values.** |
| **Resolved by this PR** | **Query setting names as parameters** | **H5 paths may have been more permissive when internal setting names appeared beside named params.** | **H7 must filter internal GORM settings before binding named parameters.** | **Settings like `lock`, `max`, or `offset` can be mistaken for named HQL parameters if not filtered.** | **Use the documented GORM settings map position. The fixed GORM APIs filter execution settings before parameter binding.** |
| Remaining app migration item | Read-only entity collections | H5 could allow collection mutation even when the owning entity was loaded read-only. | H7.3 and later mark collections owned by read-only entities as read-only too. | Mutating a collection on a read-only entity can now fail. | Reload the entity in a writable session before changing its collections. Do not mutate associations on read-only entities. |
| Remaining app migration item | Timeout exception type | H5 and earlier H7 lines commonly exposed `QueryTimeoutException` or `LockTimeoutException` for query and lock timeout paths. | H7.3 and later may throw `PersistenceException` when the database marks the transaction for rollback. | Timeout handling that catches only the narrower exception types can miss rollback-producing timeout failures. | Catch `PersistenceException` around direct Hibernate/JPA timeout-sensitive code and inspect the cause when necessary. |
| Remaining app migration item | Native SQL temporal values | H5 native queries commonly returned `java.sql` temporal values. | Hibernate 7 native SQL queries return `java.time` temporal values by default. | Native SQL result handling can fail if it casts to `java.sql.Date`, `Time`, or `Timestamp`. | Update result handling to `java.time` types, or set `hibernate.query.native.prefer_jdbc_datetime_types=true` during migration. |
| Remaining app migration item | Version-column DDL | H5 schema export did not necessarily declare version columns `not null` by default. | H7.3 declares `@Version` columns `not null` by default. | Schema validation or generated migration diffs can show new not-null constraints on version columns. | Validate generated DDL against existing schemas before using `dbCreate=update`, especially for legacy nullable version columns. |
| Remaining app migration item | General DDL generation | H5 generated different column definitions for several mappings. | H7 changes generated DDL for mappings including `char`/`Character`, Oracle floating point, `@ElementCollection` sets, `@CreationTimestamp`, `@UpdateTimestamp`, and Oracle 23c LOB columns. | Automatic schema update can propose or apply unexpected column, constraint, or LOB storage changes. | Prefer explicit database migrations for production schemas. Compare generated DDL in staging before using schema update tooling. |
| Remaining app migration item | Schema actions and `import.sql` | Schema actions could be skipped when no entities were mapped. | H7.3 and later process schema actions even when no entities are mapped. | An accidental `import.sql` on the classpath can run where it was previously ignored. | Remove accidental `import.sql` files from the runtime classpath or make schema-generation settings explicit. |
| Remaining app migration item | MySQL default fetch depth | H5 and earlier Hibernate lines could apply a MySQL dialect-level `hibernate.max_fetch_depth=2` default. | H7.4 removes that MySQL-specific default. | Applications relying on the implicit MySQL fetch depth can see different association fetch plans. | Set `hibernate.max_fetch_depth` explicitly if the application relies on that behavior. |
| Remaining app migration item | Fetch joins with limits | H5 commonly handled collection fetch-join limits in memory. | H7.4 applies pagination or limits with collection fetch joins in SQL. | Queries combining `max`/`offset` or limits with collection fetch joins can return a different row/window shape. | Review those queries and set the `org.hibernate.limitInMemory` query hint only when the previous in-memory behavior is required. |
| Remaining app migration item | Oracle HQL date expressions | Oracle `current date` and `local date` could preserve a time component because Oracle has no true date-only function. | H7.4 translates them to `trunc(current_date)`. | Oracle HQL predicates depending on the time part can behave differently. | Review Oracle queries that use `current date` or `local date` and change the expression if a timestamp is required. |
| Remaining app migration item | Eager `@Any` mappings | Eager `@Any` mappings were loaded by a separate select. | H7.4 join-fetches eager `@Any` associations when loading an entity by id. | Direct Hibernate `@Any` mappings can produce different SQL shape and row width. | Review direct Hibernate `@Any` mappings for performance and row duplication implications. |
| Remaining app migration item | Spanner PostgreSQL dialect | Spanner PostgreSQL dialect lived in `hibernate-community-dialects` under `org.hibernate.community.dialect`. | H7.4 moves it into `hibernate-core` under `org.hibernate.dialect`. | Explicit dialect configuration can reference a class that is no longer correct. | Use `org.hibernate.dialect.SpannerPostgreSQLDialect` or rely on automatic dialect resolution. |
| Remaining app migration item | Envers `NOT_AUDITED` associations | H5-era Envers behavior could read the associated target from its audit table even when `NOT_AUDITED` was requested. | H7.3 respects `RelationTargetAuditMode.NOT_AUDITED`. | Historic audit reads can return the current associated entity instead of a historic row. | Verify Envers audit queries that use `RelationTargetAuditMode.NOT_AUDITED`. |
| Fixed | Multiple datasources and OSIV | H5 datasources and OSIV coverage is mature in the general functional app. | Multi-datasource `@Rollback` transactions failed outright on H7 (chained transaction manager double-completion). | Multi-datasource switching, transactions, or GSP rendering under OSIV could behave differently after migration. | Fixed the `ChainedTransactionManager` duplicate-manager bug and added a full H7 `datasources` app (switching, cross-datasource transactions, OSIV GSP rendering) that passes on H7, matching H5. |
| Fixed | hasMany association rendering via eager fetch joins | H5 functional apps render hasMany associations fetched with `fetch:[assoc:'join']`. | Hibernate 7 did not eagerly initialize the collection for `findById`/dynamic finders, so non-deep JSON views dropped it. | Apps that render hasMany associations as JSON could silently lose collections after migration. | Fixed the H7 query layer to emit `root.fetch(...)` for eager fetch joins (now 50/52 views specs pass on H7). |
| Fixed | Proxy / duplicate-entity id rendering in JSON views | H5 renders the `id` of associations loaded via a Hibernate proxy and of entities appearing more than once in the graph. | Under Hibernate 7 a proxy-loaded entity (e.g. `Player.load(id)`) or a duplicated entity rendered without its `id` in `grails-views-gson` (2 `TeamSpec` cases: deep association + composite ID). | REST/JSON responses could omit identifiers for proxy-loaded or repeated entities. | Fixed: `findEntity` unwraps the proxy before resolving the entity class, and `preProcessedOutput` renders each occurrence fresh. Verified by 175 grails-views-gson unit tests and the H7 views app (52/52). |
| Remaining app migration item | Programmatic datastore package scanning | H5 test/application setups may accidentally scan a package that still found needed classes. | H7 tests exposed that the datastore must scan the package containing the actual support domains/services. | Programmatic `HibernateDatastore` setup can miss domain or service classes if the scanned package is wrong. | When constructing datastores manually, pass the package that contains the domain/service classes, not merely the spec or caller package. |


grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateGormStaticApi.groovy
```diff
@@ -353,22 +354,42 @@ class HibernateGormStaticApi<D> extends GormStaticApi<D> {
         if (!queryMap) return null
         Map coercedMap = queryMap.collectEntries { k, v -> [k.toString(), v] }
         String hql = buildWhereHql(coercedMap)
-        doSingleInternal(hql, coercedMap, [], args, false)
+        doSingleInternal(hql, buildWhereParams(coercedMap), [], buildFindWhereArgs(args), false)
```
borinquenkid: Try to leverage HqlQueryContext, it is the HQL Query generator

borinquenkid (dismissed):
Try to use HqlQueryContext it has all the HQL Logic

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateGormStaticApi.groovy
```diff
@@ -353,22 +354,42 @@ class HibernateGormStaticApi<D> extends GormStaticApi<D> {
...
+        }.join(' and ')
         return "from ${persistentEntity.name} where $whereClause"
     }
 
+    private String validateWherePropertyName(String propertyName) {
+        if (persistentEntity.getPropertyByName(propertyName) == null) {
+            throw new IllegalArgumentException("Property [$propertyName] is not a valid property of ${persistentEntity.name}")
+        }
+        return propertyName
+    }
```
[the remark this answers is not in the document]
jamesfredley: Addressed in `00b1e05ce4`: `validateWherePropertyName` now rejects mapped target/column aliases by requiring the resolved persistent property name to exactly match the requested key. Added `HibernateGormStaticApiMappedPropertyEntity.findWhere(name_col: ...)` coverage to lock the behavior.

grails-test-examples/hyphenated/grails-app/conf/application.yml
```diff
@@ -62,9 +62,14 @@ grails:
 hibernate:
     cache:
         queries: false
-        use_second_level_cache: true
+        use_second_level_cache: false
         use_query_cache: false
         region.factory_class: 'org.hibernate.cache.ehcache.EhCacheRegionFactory'
```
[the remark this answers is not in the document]
jamesfredley: Addressed in `00b1e05ce4`: removed the stale `org.hibernate.cache.ehcache.EhCacheRegionFactory` setting from the hyphenated sample config since second-level and query caching are disabled and the Hibernate 7 lane does not carry that provider.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateGormStaticApi.groovy
jamesfredley: Thanks, applied in `00b1e05ce4`. The generated `findWhere`/`findAllWhere` HQL now flows through the shared `prepareHqlQuery` path so `HqlQueryContext.prepare(...)` remains responsible for HQL/query setup while preserving settings, supported JPA hints, query events, and the existing `max: 1` behavior. I also added coverage for mapped-column rejection and `AvailableHints.HINT_READ_ONLY` on `findWhere`.

jdaugherty:
I am highly against arbitrary substitution of hibernate modules.  We should be cloning apps to enhance the coverage.  Otherwise we will be dealing with unexpected dependency resolutions

jamesfredley:
@jdaugherty please explain why?  We were missing a ton of issues and breaking changes due to the missing tests on h7?  This is working and tests are not production code for end apps.  This is DRY

I think we should run the same tests against both.

jamesfredley:
Updated the Hibernate 7 path to Hibernate ORM 7.4 because Spring Boot 4.1 has moved its managed Hibernate version there. Hibernate 7.3 is already in limited-support mode, so staying there would leave Grails targeting a line that is not expected to receive the ongoing updates we need. Hibernate 7.4 is the actively maintained 7.x line, so aligning with Boot 4.1 and 7.4 keeps the Hibernate 7 integration on the version that will continue to get fixes.

jamesfredley:
Pushed a docs follow-up commit (a5cd01959a) and updated the PR description. The PR body now targets Hibernate ORM 7.4, expands the fixed-issue table for the 7.4 dependency/API, proxy, identifier, generator, and event-listener fixes, and marks resolved rows in the breaking-change checklist. The Hibernate 7 docs now document the remaining end-application migration items after the PR, including Hibernate 7.3/7.4 behavior and DDL changes that are not resolved by Grails API fixes.

jdaugherty:
> @jdaugherty please explain why?  We were missing a ton of issues and breaking changes due to the missing tests on h7?  This is working and tests are not production code for end apps.  This is DRY
> 
> 
> 
> I think we should run the same tests against both.

I previously have explained this on multiple PRs.  The issue is it complicates the setup, running locally, and troubleshooting dependency resolution.  Any missing coverage we should add tests for, you have already identified the gaps so it should be easy.

borinquenkid (approved):
I made changes to improve the complexity and testability

gradle/functional-test-config.gradle
```diff
@@ -32,7 +72,8 @@ configurations.configureEach {
                 //TODO: This does not handle libraries that are both test fixtures & a libraries like grails-data-mongodb,
                 // see grails-test-examples-mongodb-base, & grails-test-examples-mongodb-hibernate5 for project() workaround
                 if (possibleProject.name == 'grails-bom') {
-                    substitute module(substitutedArtifact) using platform(project(':grails-bom'))
+                    def targetBom = redirectBomToH7 ? ':grails-hibernate7-bom' : ':grails-bom'
+                    substitute module(substitutedArtifact) using platform(project(targetBom))
```
jdaugherty: @borinquenkid my issue is this file, specifically this substitution.

jdaugherty (changes requested):

gradle/functional-test-config.gradle
borinquenkid: "Hey [@jdaugherty](https://github.com/jdaugherty), that makes total sense. In an OSS context, hidden classpath manipulation like this easily rots if left unchecked, and I agree we don't want people troubleshooting brittle IDE syncs locally.

Here is  compromise so we can hit our end-of-June release target for Hibernate 7 parity without abandoning your architectural goal:

Fail-Safe Logging: I will add an explicit, loud deprecation warning or a hardcoded TODO block directly above this substitution in functional-test-config.gradle tracking its removal.

Immediate Post-Release Issue: I'll open a tracking issue right now to completely untangle this layer by cloning the apps into a clean directory structure.

Given that I want to completely drop Hibernate 5 support in the near future anyway, duplicating the entire H5 test matrix right this second might end up being throwaway work. Can we use this substitution as a temporary bridge to unblock the June release gate, with the explicit agreement that the cleanup issue is the immediate next priority?"

matrei:
> > @jdaugherty please explain why?  We were missing a ton of issues and breaking changes due to the missing tests on h7?  This is working and tests are not production code for end apps.  This is DRY
> > I think we should run the same tests against both.
> 
> I previously have explained this on multiple PRs. The issue is it complicates the setup, running locally, and troubleshooting dependency resolution. Any missing coverage we should add tests for, you have already identified the gaps so it should be easy.

I think we have to switch hibernate modules for functional tests. I don't think it is scalable to duplicate all test apps. It will create a lot of duplication and a heavy maintenance burden. Can we create shell scripts or other solution to easily run both/all versions locally?

gradle/functional-test-config.gradle
jdaugherty: I'm strongly a -1 on this.  People will just ignore the depreciation and it doesn't solve our actual problem.  The majority of the work I've done with Grails is dependency debugging - probably over 50% of my time has been spent on it.  This completely breaks that workflow.  By substituting arbitrarily, we now have to populate that option throughout our workflows to be able to debug.

jdaugherty:
> > > @jdaugherty please explain why?  We were missing a ton of issues and breaking changes due to the missing tests on h7?  This is working and tests are not production code for end apps.  This is DRY
> > > I think we should run the same tests against both.
> > 
> > 
> > I previously have explained this on multiple PRs. The issue is it complicates the setup, running locally, and troubleshooting dependency resolution. Any missing coverage we should add tests for, you have already identified the gaps so it should be easy.
> 
> I think we have to switch hibernate modules for functional tests. I don't think it is scalable to duplicate all test apps. It will create a lot of duplication and a heavy maintenance burden. Can we create shell scripts or other solution to easily run both/all versions locally?

We don't need to duplicate the functional apps themself, but rather the test coverage in the TCK or the example apps to ensure proper coverage.  The problem with having switches for functionality is the build tools behave differently based on those switches.  This means: 

1. Gradle caching is broken when dependencies change (this is huge and made worse by the mono project)
2. Breaks debugging / troubleshooting since switches have to be populated through every tool (from the CLI to intellij)
3. Tools like IntelliJ can't easily pass parameters so they will use what ever is the default.  If hibernate5 is the default now, and there's a problem with 7 it's not easy to debug.  Once we switch, it will be the opposite.  
4. Test coverage reports don't make it clear which version was used.  This means anytime we look at code coverage or other tools we can't determine which one was the problem.  

We effectively hide behavior behind a switch and Gradle is not designed for that.  I'm a strong no for this reason.  I believe we can add test coverage, but tests in a functional app for UrlMappings don't need hibernate and shouldn't need to be cloned.  We should clone the tests that are relevant or add scenarios for them.  We need to do this anyway to prevent regressions.  

I realize that substitution is a short cut to getting this, but I view it as a major blocker for regular development & proper project hygiene.

jdaugherty:
Since no one else is offering to expand the test coverage, I'll try to make a pass at this in the next few days.

jamesfredley:
Taking a swing at this again, to attempt to address the items listed.  Test coverage for h7.4 must match h5.6, except for breaking changes we have agreed to accept (some are unavoidable given hibernate changes) and new additional tests for h7.4.

gradle/functional-test-config.gradle
jamesfredley: Agreed, and done. Removed the BOM substitution entirely in 83e5ebd454.

`gradle/functional-test-config.gradle` is back to the base-branch version: no `redirectBomToH7`, no `grails-bom` -> `grails-hibernate7-bom` redirect, no `grails-data-hibernate5` -> `grails-data-hibernate7` swap, and no `hibernate-ehcache`/`jboss-transaction` excludes or H7 BOM platform attachment. The `hibernate-version` matrix is also gone from the general `functional` workflow job, so there's no hidden classpath manipulation and `dependencies`/`dependencyInsight` debugging behaves normally again. The default lane is just Hibernate 5 as before; the existing `hibernate7Functional` job still runs the H7-specific example apps and now gates `publish`.

To keep the H7 coverage we were missing without a switch, the version-agnostic GORM behavior now lives in the shared TCK (`grails-datamapping-tck`), which both `grails-data-hibernate5` and `grails-data-hibernate7` already run via the `GrailsDataTckManager` ServiceLoader. The same specs therefore execute natively on both lines, Gradle caching stays intact, and coverage reports clearly attribute to each module.

First batch landed in `FindWhereSpec` and `GormEnhancerSpec` (findWhere/findAllWhere null matching, findWhere single-result limiting, and getAll order preservation including convertible String ids). All six new tests pass on both Hibernate 5.6 and 7.4. Remaining relevant coverage will follow the same TCK-first pattern, with dedicated H7 example specs only where behavior genuinely cannot move to the TCK.

jamesfredley:
## Update: rebased on `8.0.x-hibernate7` and pivoted off BOM substitution

I merged the latest `8.0.x-hibernate7`, resolved the conflicts, and reworked the functional-coverage approach based on the review discussion. Pushed as `83e5ebd454` (plus the merge commit `798446f7a1`).

### Addressing the substitution objection

@jdaugherty - you're right, and the arbitrary BOM substitution is gone. No more `redirectBomToH7`, no `grails-bom` -> `grails-hibernate7-bom` redirect, no `grails-data-hibernate5` -> `grails-data-hibernate7` swap, and no `-PhibernateVersion` matrix in the general `functional` job. Default dependency resolution is back to exactly what it was, so `dependencies`/`dependencyInsight` debugging and Gradle build caching are unaffected, and coverage reports attribute cleanly to each module.

@matrei - this also avoids duplicating whole apps and needing shell scripts/switches to run both versions. The shared `grails-datamapping-tck` is already run by **both** `grails-data-hibernate5` and `grails-data-hibernate7` via the `GrailsDataTckManager` ServiceLoader, so a single spec written once executes natively against both Hibernate lines with no flags. That keeps it DRY without hiding behavior behind a switch.

### What changed

**Sync / merge (`798446f7a1`)**
- Merged `origin/8.0.x-hibernate7`.
- Resolved 3 conflicts by adopting the base conventions: `registerDomainClasses()` (replacing `addAllDomainClasses`), kept the base `setupSpec()` in `Hibernate7OptimisticLockingSpec`, kept the PR's `HibernateGormStaticApiMappedPropertyEntity` coverage + `AvailableHints` import in `HibernateGormStaticApiSpec`, and dropped the `@Nonnull` annotation in `CompositeIdWithJoinTableSpec` to match base.
- Fixed merge-induced stale `grails.gorm.specs.*` imports across 7 specs (the base relocated `HibernateGormDatastoreSpec` and the multitenancy support classes to `grails.gorm.tests`), so both core test modules compile again.

**Substitution removal + shared coverage (`83e5ebd454`)**
- `gradle/functional-test-config.gradle`: restored to the base version (all substitution machinery removed).
- `.github/workflows/gradle.yml`: dropped the `hibernate-version` matrix from the general `functional` job; **kept** the publish gate that requires the existing `hibernate7Functional` job (which runs `grails-test-examples/hibernate7/*`), so H7 functional results still block publishing.
- `grails-datamapping-tck`: added shared coverage for the version-agnostic GORM behavior this PR fixed:
  - `FindWhereSpec`: `findWhere`/`findAllWhere` null matching (`is null`, not `= null`), and `findWhere` returning a single instance when multiple rows match.
  - `GormEnhancerSpec`: `getAll` order preservation, including convertible `String` ids, and a `null` slot for a missing id.

### Verification

- Both core test modules compile after the merge.
- The 6 new TCK specs pass on **both** lanes: `grails-data-hibernate5-core` (full suite, 680 tests, 0 failures) and `grails-data-hibernate7-core` (22 tests, 0 failures). No behavioral divergence for these contracts between H5.6 and H7.4.
- CodeNarc clean on `grails-datamapping-tck`.
- Hibernate ORM stays on `7.4.1.Final` / Hibernate Tools `7.3.8.Final` (unchanged).

### Remaining follow-up (TCK-first)

The H7-only architecture tests (domain-binding, HQL infrastructure) stay module-local since they have no H5 equivalent class. The broader parity sweep continues with the same pattern:
- Port the remaining version-agnostic behavioral specs (e.g. the rest of `HibernateGormStaticApiSpec`'s finder/`getAll` contracts, query-setting coercion) into the TCK so they run on both lines, and add the matching H5 coverage where H5 currently lacks it.
- Add dedicated Hibernate 7 example specs/apps only for behavior that genuinely cannot move to the TCK: multi-datasource/OSIV (`datasources`), and association rendering / unique constraints (`views-functional-tests`, `scaffolding-fields`).
- Any intentional H5.6 -> H7.4 breaking change stays documented in the migration checklist above and gets a test locking the behavior on each side.

The "Detailed issue coverage" and follow-up tables at the top have been updated to match.

jamesfredley:
I'll have one more commit that bring testing as close to parity as possible, between h5.6 and h7.4.

jamesfredley:
## Bidirectional H5.6 <-> H7.4 test-parity pass

Following up on the coverage discussion, I did a full, TCK-aware parity audit across the three test layers and added what is genuinely shared. Pushed as `ec562deee4`.

### The three layers (where coverage belongs)

| Layer | Module | Runs on | Use for |
|---|---|---|---|
| Shared TCK | `grails-datamapping-tck` | H5 **and** H7 (and Mongo) via the `GrailsDataTckManager` ServiceLoader | Version-agnostic GORM contracts |
| Per-module | `grails-data-hibernate5/7/core/src/test` | one Hibernate line | Hibernate-implementation behavior (cannot live in the datastore-agnostic TCK) |
| Functional | `grails-test-examples` | app boot | End-to-end app behavior |

### What is already at parity (no action needed)

The shared TCK already runs **~60 GORM behavioral specs on both H5.6 and H7.4** (CRUD, dynamic finders, `findWhere`, criteria, detached criteria, projections, paged results, optimistic locking, proxies, validation, events, multi-datasource, multi-tenancy, etc.). The raw file-count gap (H5 ~109 vs H7 ~315 specs) is **not** a coverage gap: ~140 of the H7-only specs cover the Hibernate-7 internal rewrite (`cfg/domainbinding/**`, `query/**` HQL infrastructure, `support/hibernate7/**`) - code that does not exist in H5, so there is nothing to mirror.

### What I added this pass

- **TCK (both sides):** extended `FindWhereSpec` and `GormEnhancerSpec` with the finder behaviors this PR fixed - `findWhere`/`findAllWhere` null matching (`is null`), `findWhere` single-result limiting, `getAll` order preservation incl. convertible String ids and a null slot for a missing id. Verified passing on both lines.
- **H5 per-module (parity with H7):** ported 7 H7 specs whose classes exist identically in H5 - `HibernateGormEnhancerSpec`, `HibernateDatastoreMultiTenancySpec`, `HibernateEventListenersSpec`, `HibernateConnectionSourceSettingsBuilderSpec`, `GrailsQueryExceptionSpec`, `HibernateDatastoreConnectionSourcesRegistrarSpec`, `SoftKeySpec`. All pass on H5 (32 tests, 0 failures) and still pass on H7.

### Verified H7-only (correctly asymmetric - not gaps)

I copied every remaining shared-looking H7 behavioral spec into H5 and **ran them**; these failed because they genuinely test Hibernate-7-specific API/behavior, so they stay H7-only (documented as the new behavior where it is a breaking change):

| Spec | Why it is H7-only |
|---|---|
| `HibernateCriteriaBuilderSpec` / `...DirectSpec` | H7 adds a `HibernateCriteriaBuilder(Class, SessionFactory, HibernateDatastore)` ctor + criteria behaviors (isEmpty/scroll/fetchMode/singleResult) that differ on H5; criteria behavior itself is already covered both-sides by the TCK `CriteriaBuilderSpec` |
| `InstanceApiHelperSpec` | H7 `InstanceApiHelper.remove(Object, boolean)` signature differs from H5 |
| `EventListenerIntegratorSpec` | Hibernate 7 event SPI (`integrate(Metadata, BootstrapContext, SessionFactoryImplementor)`) differs from H5 |
| `HibernateGormValidationApiSpec` | H7 shares the datastore `GrailsHibernateTemplate`; H5 does not (intentional H7 behavior) |
| `TraitPropertyAccessStrategySpec`, `GrailsEntityDirtinessStrategySpec`, `MultiTenantEventListenerSpec`, `HibernateRuntimeUtilsSpec`, `HibernateDatastoreSchemaMultiTenancySpec` | Depend on Hibernate-7 SPI / metamodel types (`EntityMappingType`, etc.) or H7-only classes (`ChildHibernateDatastore`, `CloseSuppressingInvocationHandler`, `HibernatePagedResultList`, `SelectHqlQuery`, `HibernateDetachedCriteria`) absent from H5 |

### Tracked follow-up (next pass)

- Promote to the TCK (both sides) where fully generic: `GormEnhancerCleanupSpec`, `PersistentPropertySpec`, the `GroovyProxyFactory` proxy spec.
- Reverse gap: H5 `ByteBuddyProxySpec` (default ByteBuddy proxy: getId/`isInitialized` without initialization) has no H7 equivalent - add an H7 version using the H7 `HibernateProxyHandler`.
- De-dupe: H7 has a local `NullValueEqualSpec` that duplicates the TCK one.
- Functional layer: dedicated H7 apps for `datasources` (OSIV/multi-datasource), `views-functional-tests` and `scaffolding-fields` (association rendering, unique constraints), rather than the removed runtime substitution.

Net: every version-agnostic GORM behavior is exercised on both Hibernate lines (TCK + the new H5 specs), and every remaining asymmetry is a Hibernate-implementation difference that was verified by actually running the tests on both sides - not an untested gap.

jamesfredley:
## Module + TCK parity: tracked follow-ups completed

Continued the parity pass; the module/TCK follow-ups from the previous comment are now done and verified. Pushed as `ec562deee4` and `08ae34fa36`.

**H7 -> H5 (shared behavior added to H5, verified passing on H5):**
`HibernateGormEnhancerSpec`, `HibernateDatastoreMultiTenancySpec`, `HibernateEventListenersSpec`, `HibernateConnectionSourceSettingsBuilderSpec`, `GrailsQueryExceptionSpec`, `HibernateDatastoreConnectionSourcesRegistrarSpec`, `SoftKeySpec`, `AddToManagedEntitySpec`, `GormEnhancerCleanupSpec`, `PersistentPropertySpec`, `NullValueEqualSpec`.

**H5 -> H7 (reverse gap closed):**
`grails/gorm/tests/proxy/ByteBuddyProxySpec` added to H7 (default ByteBuddy proxy getId/id/truthy/association checks do not initialize the proxy).

**Documented H5 vs H7 behavioral differences (each side asserts its own real result, verified):**

| Behavior | Hibernate 5.6 | Hibernate 7.4 |
|---|---|---|
| `PersistentProperty.isLazyAble()` on a unidirectional one-to-many | `false` | `true` |
| `isDirty()` on a clean loaded proxy | initializes the proxy | does **not** initialize the proxy |
| `countByAgeNotEqual(11)` with null-aged rows present | `0` (SQL three-valued logic excludes nulls) | `2` (null-valued rows included) |

**Verified H7-only (not gaps):** the remaining H7 behavioral specs were each copied into H5 and run; they fail on H5 because they depend on Hibernate-7-specific API/behavior (`HibernateCriteriaBuilder`/`InstanceApiHelper`/`EventListenerIntegrator` signatures, H7 event SPI/metamodel types, `ChildHibernateDatastore`/`CloseSuppressingInvocationHandler`/`HibernatePagedResultList`/`SelectHqlQuery`/`HibernateDetachedCriteria`), so they correctly stay H7-only. The ~140 `cfg/domainbinding/**` + `query/**` specs cover the H7 internal rewrite that has no H5 code to mirror.

**State:** full `grails-data-hibernate5-core` suite green (722 tests, 0 failures, 33 skipped) with all additions; new H7 specs verified on H7. Every version-agnostic GORM contract now runs on both lines (shared TCK + the per-module specs), and there are no remaining true H5-only or H7-only behavioral gaps - only legitimate Hibernate-implementation asymmetry.

**Remaining (separate functional pass):** dedicated Hibernate 7 functional apps under `grails-test-examples` for the GORM-bearing general apps (`datasources` OSIV/multi-datasource, `views-functional-tests` and `scaffolding-fields` association rendering / unique constraints), replacing the removed runtime substitution. This is app-creation + Geb work rather than spec porting.

jamesfredley:
## Update: chained transaction manager fix + Hibernate 7 datasources app + base sync

Continued making Hibernate 7.4 a drop-in for Hibernate 5.6 by cloning the flagged general functional apps to H7 and fixing the real production bugs they surfaced. Synced with the latest `8.0.x-hibernate7`. New commits: `3d82ad2c63`, `a9b8a98f52` (+ merge `8e5735500c`).

### Fixed: `ChainedTransactionManager` double-completion under Hibernate 7 (`3d82ad2c63`)

A genuine H7 regression, found by cloning the `datasources` app and running it:

- **Symptom:** with multiple data sources + `chainedTransactionManager.enabled: true`, every `@Rollback` transaction failed under H7 with `IllegalTransactionStateException: Transaction is already completed - do not call commit or rollback more than once`. The byte-identical app passes on H5 (30/30) and failed on H7 (30/30).
- **Root cause (confirmed by instrumenting the transaction lifecycle):** `ChainedTransactionManagerPostProcessor` builds the delegate list from bean *names* (`$primaryTransactionManager` + every `*transactionManager(_.+)?` bean). Under H7 a second bean name aliases the **same** primary transaction-manager instance, so it appears twice. `MultiTransactionStatus` is keyed by instance, so the chained `commit` iterated that instance twice and committed the same `TransactionStatus` twice.
- **Fix:** `ChainedTransactionManager` now dedupes delegate managers by identity (order preserved). A duplicate instance is always meaningless for this class, so it is a no-op when there is no duplicate (e.g. H5). Locked with `ChainedTransactionManagerTests.shouldDeduplicateRepeatedTransactionManagerInstances`.

### Added: Hibernate 7 `datasources` functional app (`a9b8a98f52`)

`grails-test-examples/hibernate7/datasources` is a **near-identical clone** of the H5 `grails-test-examples/datasources` app - the only difference is `build.gradle` (`grails-bom` -> `grails-hibernate7-bom`, `grails-data-hibernate5` -> `grails-data-hibernate7`), so the two are trivially diffable. All four integration specs pass on H7, matching H5:

- `DatasourceSwitchingSpec` 18/18, `CrossDatasourceTransactionSpec` 12/12, `MultipleDataSourcesSpec` 1/1, `OsivGspRenderingSpec` 1/1.

It auto-registers as `grails-test-examples-hibernate7-datasources` and runs in the existing `hibernate7Functional` job.

### Found (documented, tracked): Hibernate 7 lazy hasMany not rendered in JSON views

Cloning `views-functional-tests` to H7 showed **48/52** integration specs already pass under H7 (association/proxy/inheritance/embedded/circular rendering all work). The 4 `TeamSpec` failures share one root cause, confirmed by instrumentation:

- Under H7, `Team.findById(id, [fetch:[players:'join']])` returns the `players` hasMany as an **uninitialized** `org.hibernate.collection.spi.PersistentList`. The JSON view helper by-design skips uninitialized lazy collections in non-deep mode, so `players` is dropped from the JSON (`captain`, a to-one, still renders). H5 renders the collection.
- This is a GORM query-layer difference (`fetch:[assoc:'join']` not eager-initializing the collection under H7), not a view-layer bug. It is a real migration risk (JSON APIs can silently lose hasMany collections) and is tracked in the description for a focused fix. The `views`/`scaffolding` H7 clones are deferred until it is resolved (I did not ship partial/`@PendingFeature` apps).

### Also

- Merged the latest `8.0.x-hibernate7` (typed attribute reads, scaffolding generics, http-client form post) - clean, no conflicts.
- Updated the "Detailed issue coverage", "Known follow-up", and migration-checklist tables above to reflect the fix, the new app, and the views finding.

Net since the parity pass: the multi-datasource chained-transaction regression is fixed, the `datasources` general app now has full verified H7 parity, and the remaining views/scaffolding gap is reduced to one precisely-characterized GORM query bug.

jamesfredley:
I am trying to work through the outstanding h7.4 vs h5.6 items before the call today.

jamesfredley:
## CI fixes + Hibernate 7 eager fetch-join fix

Fixed the failing CI and the underlying H7 query bug it pointed at. New commits: `81b0c52996`, `fab342ec2d`.

### CI failures fixed
- **`Core Projects` (checkstyle):** removed an unused `java.util.Arrays` import left over from the ChainedTransactionManager dedupe fix.
- **`MongoDB Functional Tests` + `Coverage - grails-core` (`grails-data-mongodb-core:test`):** the shared TCK test `getAll preserves order for convertible String ids` is Hibernate-specific (MongoDB ObjectId ids are not derived from arbitrary numeric strings), so it failed on Mongo. Moved it out of the datastore-agnostic TCK into `HibernateGetAllConvertibleIdSpec` in both Hibernate modules (H5<->H7 parity preserved; no longer run on Mongo). The version-agnostic getAll tests stay in the TCK.

### Fixed: Hibernate 7 eager fetch joins (`fab342ec2d`)
A real H7 regression, surfaced while cloning the views app: `Entity.findById(id, [fetch:[assoc:'join']])` (and dynamic finders) did not eagerly initialize the association on H7, while H5 did - so JSON views silently dropped hasMany collections.

Root cause: GORM's `DynamicFinder` turns an EAGER fetch into `query.join(property)`; the rewritten H7 query layer materialized that as a plain JPA `root.join` only when referenced in a predicate (a fetch-only association never is), so it was never fetched. JPA needs `root.fetch(...)`.

Fix (H7 only): `HibernateQuery` records fetch-join paths; `JpaCriteriaQueryCreator` materializes them as `root.fetch(path, joinType)` for full-entity selects (count/projection queries skipped). Locked with `FetchJoinSpec` in both modules. The **full `grails-data-hibernate7-core` suite is green (2944 tests, 0 failures)**, so criteria/finder/projection/count/pagination are not regressed; H5 is untouched.

### Views functional app: 48/52 -> 50/52 on H7
With the fetch-join fix, cloning `views-functional-tests` to H7 now passes 50/52 (association + HAL rendering fixed). The remaining 2 (`TeamSpec` deep association + composite ID) hit a **separate, narrower** grails-views-gson bug: an entity loaded via a Hibernate proxy (`Player.load(id)`), or appearing twice in the graph, renders without its `id` under H7 (and a duplicate is de-duplicated to an empty object). H5 renders the ids. That is documented in the tables above as the remaining blocker for landing the `views`/`scaffolding` H7 clones - it is a view-layer (`grails-views-gson`) id-extraction + de-dup issue, not the GORM query bug that is now fixed.

CI is re-running on these fixes.

H7_GORM_BUG_REPORT.md
```diff
@@ -0,0 +1,99 @@
+<!--
+SPDX-License-Identifier: Apache-2.0
```
jdaugherty, line 2: Why is this checked in?

grails-core/src/main/groovy/org/grails/transaction/ChainedTransactionManager.java
```diff
@@ -88,7 +89,18 @@ public ChainedTransactionManager(PlatformTransactionManager... transactionManage
 
         this.synchronizationManager = synchronizationManager;
         this.transactionManagers = new ArrayList<>();
-        this.transactionManagers.addAll(Arrays.asList(transactionManagers));
+        // A single PlatformTransactionManager instance can be exposed under more than one bean
```
jdaugherty: Let's simplify this comment?

jdaugherty (approved):

grails-test-examples/hyphenated/grails-app/conf/application.yml
```diff
@@ -62,9 +62,13 @@ grails:
 hibernate:
     cache:
         queries: false
-        use_second_level_cache: true
+        use_second_level_cache: false
```
jdaugherty, line 65: Can we make sure there is a test that still tests the second level cache being configured?

jdaugherty (approved):

H7_GORM_BUG_REPORT.md
jamesfredley: Let's address this in the step 2 PR, after this one merges

jamesfredley:
## H7 JSON-view proxy/duplicate-entity id rendering fixed; views app landed

Fixed the remaining grails-views-gson rendering bug and landed the Hibernate 7 views functional app. New commits: `cc6acc3804`, `33a3a1f349`.

### Fixed: proxy / duplicate-entity id rendering (`cc6acc3804`)
The two `TeamSpec` cases that were still failing (deep association + composite id) had two root causes, both now fixed:

- `DefaultJsonViewHelper.findEntity` used `object.getClass()`, which for a Hibernate proxy is the generated subclass (`Player$HibernateProxy$xxx`) and is **not** a registered persistent entity. `findEntity` returned null and the proxy was rendered as a plain POGO (via `processSimple`, dropping its id) instead of as a domain entity. Now it unwraps the proxy before resolving the entity class.
- `DefaultGrailsJsonViewHelper.preProcessedOutput` returned a cached `JsonWritable` bound to the first occurrence's delegate/path/deep state, so a repeated instance (e.g. the same player as both a team's captain and a member of its players collection) rendered the wrong context and dropped the id. Now each occurrence renders fresh; true cyclic associations are still broken earlier by the circular-association handling in `process()`.

Verified: full grails-views-gson unit suite **175/175**, and Hibernate 5 behaviour unchanged.

### Landed: Hibernate 7 views-functional-tests app (`33a3a1f349`)
Added `grails-test-examples/hibernate7/views-functional-tests`, a near-identical clone of the H5 app (only `build.gradle` differs). **All 52 integration specs pass on Hibernate 7**, matching H5 - association/proxy/inheritance/embedded/circular rendering, deep rendering, HAL, and composite ids.

### Scaffolding status
Cloning `scaffolding-fields` to H7 boots and passes **123/136** Geb specs (CRUD, field types, pagination, custom templates). The remaining 13 (10 relationship rendering, 2 validation, 1 CRUD) are a **separate** scaffolding/fields-vs-H7 issue: bootstrap-seeded association data does not appear in scaffolded belongsTo/hasMany/many-to-many forms under H7. This is unrelated to the JSON-view rendering path fixed here and is tracked as the remaining follow-up (the clone is not landed until it is resolved). The breaking-change/coverage tables above are updated accordingly.

grails-core/src/main/groovy/org/grails/transaction/ChainedTransactionManager.java
jamesfredley: Done in `6ae130adb5` - condensed the six-line comment to two lines that state the identity-dedup rationale (the same manager can be registered under several bean names, and committing one instance twice throws "Transaction is already completed").

grails-test-examples/hyphenated/grails-app/conf/application.yml
jamesfredley: Yes - the second-level cache being configured is covered on Hibernate 7 by `SecondLevelCacheSpec` (`grails-data-hibernate7/core/src/test/.../connections/SecondLevelCacheSpec.groovy`). It sets `hibernate.cache.use_second_level_cache: true` with the Hibernate-7-compatible `org.hibernate.cache.jcache.internal.JCacheRegionFactory`, maps a `@Cache` entity, and asserts the 2LC region is registered and exercises the miss-then-hit behaviour (load once = miss, load again = hit, served from the cache).

This `hyphenated` app sets `use_second_level_cache: false` only because the value it previously used, `org.hibernate.cache.ehcache.EhCacheRegionFactory`, is the Hibernate-5 EhCache 2 region factory and does not exist on Hibernate 7. The 2LC configuration behaviour itself is not lost - it remains covered by `SecondLevelCacheSpec`.

jamesfredley:
## Fixed: non-transactional BootStrap saves silently discarded on Hibernate 7

Investigated the scaffolding-fields-on-H7 failures and found a real Hibernate 7 framework regression. New commits: `7a6cc6375f` (fix + test), `6ae130adb5` (review-comment: simplified the `ChainedTransactionManager` dedupe comment).

### Root cause (runtime-traced)
A `@Integration` clone of `scaffolding-fields` showed bootstrap-seeded data missing from scaffolded relationship forms. Instrumenting `BootStrap` proved that GORM `.save()` calls in a **non-transactional** `init` closure did not persist on H7 (`Department.count() == 0` immediately after seeding), while a `withTransaction {}` probe committed them.

`GrailsConfigUtils.executeGrailsBootstraps` binds a session via the `PersistenceContextInterceptor` (`init()`), runs the BootStrap closures, then calls `interceptor.flush()` / `destroy()`. On H7, `HibernatePersistenceContextInterceptor.flush()` was a **no-op** for this path: its constructor hardcodes `transactionRequired = true`, and `flush()` only flushed when `TransactionSynchronizationManager.isSynchronizationActive()` - which is `false` during a non-transactional BootStrap. So the interceptor-owned session was never flushed/committed and `destroy()` discarded the pending inserts. (H5 happened to persist them.)

### Fix
`HibernatePersistenceContextInterceptor.flush()` now, when it is not participating in an outer session and there is no active synchronization (it opened the session itself, e.g. for a non-transactional BootStrap), flushes **and commits** the owned session's pending changes in a short transaction. The existing non-transactional and active-synchronization branches (the OSIV web-request path) are unchanged. Locked with a `HibernatePersistenceContextInterceptorSpec` regression test mirroring the BootStrap flow.

### Verification
- `HibernatePersistenceContextInterceptorSpec`: 4/4 (3 existing + the new test).
- H7 `scaffolding-fields` functional app: **123/136 -> 134/136** (all the bootstrap-data relationship failures resolved).
- H7 `views-functional-tests`: still **52/52** (OSIV not regressed).

The 2 residual scaffolding failures are Geb test-stability issues, not H7 regressions: a flaky sort/list interaction (a different pagination spec fails on each run) and a show-page belongsTo display assertion (body captured before content renders). Tables above updated; the second-level-cache question is answered inline (covered by `SecondLevelCacheSpec`).

PR #15732, opened 2026-06-11
8.x Add Class-typed getAttribute reads for session, request and servletContext

codeconsole:
## Summary

Follow-up to #15715. The named converters (`string`, `int`, `boolean`, …) cover scalar attribute values, but attributes frequently hold richer objects — a security principal, a domain instance, a CSRF token — where today the only statically-compilable read is a cast at every call site:

```groovy
CsrfToken token = (CsrfToken) session.getAttribute('_csrf')
UserDetails user = request.getAttribute(USER_ATTRIBUTE) as UserDetails
```

This adds a `Class`-typed overload of `getAttribute` (plus a default-value variant) to `session`, `request` and `servletContext`:

```groovy
CsrfToken token   = session.getAttribute('_csrf', CsrfToken)
UserDetails user  = request.getAttribute('apiUser', UserDetails)
Theme theme       = session.getAttribute('theme', Theme, Theme.DEFAULT)
```

The generic signature (`static <T> T getAttribute(holder, String name, Class<T> type)`) lets the static compiler infer the return type from the class literal, so the read compiles under `@CompileStatic` / `@GrailsCompileStatic` with no cast.

## Design

- **Typed read, not coercion.** The attribute is returned only when it is an instance of the requested type; absent and wrong-typed attributes read as `null` (so the default-value overload also applies on a type mismatch). No `TypeConverters` delegation — the named converters remain the coercing API, the `Class` overload is the type-safe retrieval API. Two simple contracts instead of one mixed one.
- **Overloads `getAttribute` rather than introducing a new verb.** Mirrors `grails.config.Config.getProperty(String key, Class<T> targetType, T defaultValue)` (and Spring's `Environment.getProperty(key, type, default)`), and shows up in IDE completion right next to the raw `getAttribute(String)`.
- **Scope: the three attribute holders only.** `flash` is skipped — it is a `Map`, and Groovy's DGM `Map.get(key, default)` (which *inserts* the default) makes a two-arg sibling on flash a semantic trap. `params` is skipped — its values are Strings, where a non-coercing typed read is not useful.

## Tests & docs

- `HttpSessionExtensionSpec`, `HttpServletRequestExtensionSpec`, `ServletContextExtensionSpec` each gain two features: instance-match / supertype-match / absent / wrong-type-reads-as-null semantics, default-value behavior, and a `@CompileStatic` static-resolution guard (the guard fails test compilation without the new extension methods).
- New "Typed Reads of Attributes" subsection in the controllers guide, following the "Type Conversion of Attributes" section from #15715.

`:grails-web-core:test` passes in full.

### DIFF
--- a/grails-web-core/build.gradle
+++ b/grails-web-core/build.gradle
@@ -71,6 +71,7 @@ dependencies {
     // Testing
     testImplementation 'org.slf4j:slf4j-simple'
     testImplementation 'org.spockframework:spock-core'
+    testImplementation 'net.bytebuddy:byte-buddy'
 }
 
 apply {
--- a/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
+++ b/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
@@ -21,6 +21,7 @@ package org.grails.web.servlet
 import groovy.transform.CompileStatic
 
 import org.apache.grails.core.internal.util.TypeConverters
+import org.springframework.util.ClassUtils
 
 import jakarta.servlet.http.HttpServletRequest
 
@@ -258,7 +259,8 @@ class HttpServletRequestExtension {
      */
     static <T> T getAttribute(HttpServletRequest request, String name, Class<T> type) {
         Object value = request.getAttribute(name)
-        type.isInstance(value) ? type.cast(value) : null
+        Class<T> resolvedType = (Class<T>) ClassUtils.resolvePrimitiveIfNecessary(type)
+        resolvedType.isInstance(value) ? resolvedType.cast(value) : null
     }
 
     /**
--- a/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
+++ b/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
@@ -258,6 +258,9 @@ class HttpServletRequestExtension {
      * use the named converters ({@code string}, {@code int}, ...) for type conversion.
      */
     static <T> T getAttribute(HttpServletRequest request, String name, Class<T> type) {
+        if (type == null) {
+            throw new IllegalArgumentException('type must not be null - use getAttribute(name) for an untyped read')
+        }
         Object value = request.getAttribute(name)
         Class<T> resolvedType = (Class<T>) ClassUtils.resolvePrimitiveIfNecessary(type)
         resolvedType.isInstance(value) ? resolvedType.cast(value) : null
--- a/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
+++ b/grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
@@ -251,4 +251,22 @@ class HttpServletRequestExtension {
     static Date date(HttpServletRequest request, String name, Collection<String> formats) {
         TypeConverters.toDate(request.getAttribute(name), formats)
     }
+    /**
+     * Null-safe, typed read of an attribute. Returns the attribute when it is an
+     * instance of {@code type}; otherwise {@code null}. No coercion is attempted —
+     * use the named converters ({@code string}, {@code int}, ...) for type conversion.
+     */
+    static <T> T getAttribute(HttpServletRequest request, String name, Class<T> type) {
+        Object value = request.getAttribute(name)
+        type.isInstance(value) ? type.cast(value) : null
+    }
+
+    /**
+     * Null-safe, typed read of an attribute with a default. Returns {@code defaultValue}
+     * when the attribute is absent or is not an instance of {@code type}.
+     */
+    static <T> T getAttribute(HttpServletRequest request, String name, Class<T> type, T defaultValue) {
+        T value = getAttribute(request, name, type)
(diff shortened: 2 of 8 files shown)


grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
```diff
@@ -251,4 +252,26 @@ class HttpServletRequestExtension {
...
+     * instance of {@code type}; otherwise {@code null}. No coercion is attempted —
+     * use the named converters ({@code string}, {@code int}, ...) for type conversion.
+     */
+    static <T> T getAttribute(HttpServletRequest request, String name, Class<T> type) {
+        if (type == null) {
+            throw new IllegalArgumentException('type must not be null - use getAttribute(name) for an untyped read')
+        }
+        Object value = request.getAttribute(name)
+        Class<T> resolvedType = (Class<T>) ClassUtils.resolvePrimitiveIfNecessary(type)
+        resolvedType.isInstance(value) ? resolvedType.cast(value) : null
```
jdaugherty, line 266: Shouldn't this be value?.class?.isAssignableFrom ?

codeconsole: `type.isInstance(value)` is true when `value` *is-a* `type`, including when `type` is a supertype/interface of the value's concrete class. That's the primary use case for these reads:

```groovy
// stored value is a concrete DefaultCsrfToken; caller requests the interface
CsrfToken token = request.getAttribute('_csrf', CsrfToken)
```

`value.class.isAssignableFrom(type)` inverts the relationship — it asks "is `type` the same as or a *subclass* of the value's class", i.e. `DefaultCsrfToken.isAssignableFrom(CsrfToken)`, which is `false` — so it would return `null` for every request-by-interface. The supertype-match case is covered by the spec (`getAttribute('principal', CharSequence)` for a `StringBuilder` value).

The `isAssignableFrom` equivalent of the current check would be `type.isAssignableFrom(value.class)` (with a null guard); `type.isInstance(value)` just expresses that directly and is null-safe.

sbglasius (approved):

grails-web-core/src/main/groovy/org/grails/web/servlet/HttpServletRequestExtension.groovy
jdaugherty: What about proxies then?  isInstance will return false for proxies.

codeconsole: Proxies are fine with `isInstance` — added a test covering both mechanisms: a JDK dynamic proxy (its runtime-generated class implements the interface, so `isInstance(interface)` is true) and a subclass (the relationship CGLIB / ByteBuddy / Hibernate-lazy proxies use — `isInstance` walks the class hierarchy the same way whether the subclass is generated at runtime or declared). Both resolve; a wrong-type request still reads as `null` rather than an unsafe cast.

`value.class.isAssignableFrom(type)` would actually break the proxy case — a proxy's class is a *subtype* of the proxied type, so it returns false even when the value is a valid instance.

codeconsole: @jdaugherty I added a test example

grails-web-core/src/test/groovy/org/grails/web/servlet/HttpServletRequestExtensionSpec.groovy
```diff
@@ -105,6 +105,38 @@ class HttpServletRequestExtensionSpec extends Specification {
             request.getAttribute('count', Integer, 7) == 42
     }
 
+    void "Class-typed getAttribute matches proxied values — both JVM proxy mechanisms"() {
+        given:
+            HttpServletRequest request = new MockHttpServletRequest()
+
+            // (1) Interface proxy: a JDK dynamic proxy's runtime-generated class *implements* the
+            // requested interface, so isInstance(interface) is true — the non-obvious runtime case.
+            CharSequence jdkProxy = (CharSequence) java.lang.reflect.Proxy.newProxyInstance(
```
jdaugherty: We don't use the JDK proxy though?  We use bytebuddy.

codeconsole: Fair point — updated the test to use a real ByteBuddy subclass proxy (`new ByteBuddy().subclass(...).make().load(...)`), the same machinery GORM/Hibernate use for lazy proxies. It asserts the stored value is a generated subclass (`proxy.class != Entity`) yet `getAttribute(name, Entity)` resolves it via `isInstance`, and still returns null for an unrelated type. Added `net.bytebuddy:byte-buddy` (BOM-managed) as a testImplementation.

grails-web-core/build.gradle
```diff
@@ -71,6 +71,7 @@ dependencies {
     // Testing
     testImplementation 'org.slf4j:slf4j-simple'
     testImplementation 'org.spockframework:spock-core'
+    testImplementation 'net.bytebuddy:byte-buddy'  // exercise the Class-typed getAttribute against real GORM/Hibernate-style proxies
```
jdaugherty: Remove the redundant comments

jdaugherty (approved):

PR #15733, opened 2026-06-11
8.x Add statically-resolvable message(Map) for controllers

codeconsole:
## Summary

`message(code: ..., args: [...])` is by far the most common tag invoked as a method from controllers — it is what scaffolding generates for every flash message:

```groovy
flash.message = message(code: 'default.created.message', args: [...])
redirect book
```

Today that call only resolves via `TagLibraryInvoker.methodMissing`, so under `@CompileStatic` / `@GrailsCompileStatic` it fails to compile and forces `@CompileDynamic` onto exactly the controller actions that are otherwise fully statically compilable (the same pain #15715 addressed for attribute access). This PR makes the canonical idiom compile statically, verbatim.

## Design

A new trait `grails.artefact.gsp.ControllerTagLibraryInvoker extends TagLibraryInvoker` declares a real `Object message(Map attrs)`, and `ControllerTagLibraryTraitInjector` now injects it (controllers keep the full inherited `TagLibraryInvoker` API).

- **Dispatch parity.** The method resolves the tag through the same lookup order as `methodMissing` (declared `taglibNamespace` first, then the default namespace) and invokes it through the same `TagOutput.captureTagOutput(..., OutputContextLookupHelper.lookupOutputContext())` machinery that GSP-compiled code and the `TagLibraryMetaUtils`-registered meta-methods use — so attribute handling (`code`, `args`, `default`, `error`, `message`, `locale`, `encodeAs`), encoding, and return semantics are identical to the dynamic path. When no tag library provides the tag, it throws `MissingMethodException`, matching `methodMissing`'s failure mode.
- **Why a subtrait instead of `TagLibraryInvoker` itself:** tag libraries also implement `TagLibraryInvoker` (via `grails.artefact.TagLibrary`), and a real `message` method inherited by every tag library is picked up by `TagMethodInvoker` as if the library declared a `message` tag — recursing infinitely for libraries that don't (caught by `grails-test-suite-uber` during development). Scoping the method to the controller-injected trait avoids leaking it into the tag-dispatch path. The javadoc documents this constraint.

## Tests

New `ControllerTagLibraryInvokerMessageSpec`:
- dispatches to the `message` tag of the default namespace with attrs passed through
- prefers the declared `taglibNamespace`, falling back to the default — mirroring `methodMissing` order
- throws `MissingMethodException` when no tag library provides the tag
- a `@CompileStatic` controller-style class invoking the canonical idiom compiles and resolves (this guard fails test compilation without the trait method)

`:grails-web-taglib:test`, `:grails-gsp:test`, `:grails-test-suite-web:test`, and `:grails-test-suite-uber:test` all pass.

### DIFF
--- a/grails-gsp/grails-web-taglib/src/main/groovy/grails/artefact/gsp/ControllerTagLibraryInvoker.groovy
+++ b/grails-gsp/grails-web-taglib/src/main/groovy/grails/artefact/gsp/ControllerTagLibraryInvoker.groovy
@@ -1,0 +1,70 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package grails.artefact.gsp
+
+import groovy.transform.CompileStatic
+
+import org.grails.taglib.TagLibraryLookup
+import org.grails.taglib.TagOutput
+import org.grails.taglib.encoder.OutputContextLookupHelper
+
+/**
+ * The {@link TagLibraryInvoker} flavor injected into controllers. On top of the dynamic
+ * tag dispatch inherited from {@link TagLibraryInvoker}, it declares a statically-resolvable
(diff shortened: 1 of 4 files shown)


grails-gsp/grails-web-taglib/src/main/groovy/grails/artefact/gsp/ControllerTagLibraryInvoker.groovy
```diff
@@ -0,0 +1,72 @@
...
+     * Statically-resolvable counterpart of the dynamic {@code message(...)} tag dispatch.
+     * Dispatches through the same tag-library lookup (declared namespace first, then the
+     * default namespace) and the same {@link TagOutput#captureTagOutput} machinery as the
+     * dynamic path, so behavior is identical.
+     *
+     * @param attrs the tag attributes ({@code code}, {@code args}, {@code default}, {@code error},
+     *              {@code message}, {@code locale}, {@code encodeAs})
+     * @return the resolved message
+     */
+    Object message(Map attrs) {
```
jdaugherty, line 56: This is the wrong solution: we basically are copying/moving the functionality for a specific tag.  You can inject the taglib directly to get static support.  We should consider the registry solution I proposed last week or the static solution that defers to dynamic dispatch.

jdaugherty (changes requested):

codeconsole:
in favor of #15669

PR #15743, opened 2026-06-17
Do not close an externally-supplied MongoClient in GORM for MongoDB

codeconsole:
## What

GORM's connection-source shutdown closes any `Closeable` source, and `com.mongodb.client.MongoClient` is `Closeable`, so a client handed to GORM — a Spring-managed bean, or one passed to `new MongoDatastore(mongoClient, ...)` — was closed when the datastore shut down. That double-closes a client whose lifecycle is owned elsewhere.

This makes GORM **not close a `MongoClient` it did not create**:

- `DefaultConnectionSource` gains a non-closeable mode (new 4-arg constructor + `isCloseable()`); the existing constructor is unchanged, so there is no behavior change for current callers.
- `MongoDatastore.createDefaultConnectionSources` wraps an externally-supplied client as non-closeable.
- `MongoDbGormAutoConfiguration` implements `DisposableBean` and closes only the client it created itself; a client supplied as a bean is left for the application context to close.

GORM still closes clients it creates from `grails.mongodb.*` configuration.

## Why

- Removes a redundant double-close in the common Spring Boot setup (Spring already owns and closes the `MongoClient` bean).
- Establishes "whoever creates the client owns closing it" — the safe basis for sharing one `MongoClient` across GORM and other consumers.

## Behavior change

The only externally-visible change: callers that relied on `MongoDatastore.close()` to also close a client they passed in must now close it themselves. Documented in the upgrade notes.

## Tests

- `DefaultConnectionSourceSpec` — closes by default; no-op when non-closeable; closes an `AutoCloseable` when closeable.
- `MongoDatastoreExternalClientSpec` — a supplied client is not closed on datastore shutdown.
- `MongoDbGormAutoConfigurationCloseSpec` — the auto-config closes only an internally-created client.

Targets `8.0.x`.

PR #15744, opened 2026-06-17
Add opt-in MongoDB multi-document transactions to GORM for MongoDB

codeconsole:
## What

Adds **opt-in MongoDB multi-document transactions** to GORM for MongoDB.

Previously GORM for MongoDB never used a `com.mongodb.client.ClientSession`: every driver write was issued session-less and auto-committed, so a GORM transaction was only a client-side flush boundary — writes already flushed within it were **not** rolled back on failure (no server-side atomicity).

With `grails.mongodb.transactional = true`, a GORM transaction now starts and drives a real `ClientSession` transaction, so all reads and writes commit or roll back atomically:

```groovy
Person.withTransaction {
    new Person(name: "Fred").save()
    new Person(name: "Wilma").save()
    // both commit together, or neither if an exception is thrown
}
```

## How

- New `MongoTransaction` (replaces the flush-only `SessionOnlyTransaction` when enabled): `commit()` flushes then `commitTransaction()` (with bounded retry on `UnknownTransactionCommitResult`), `rollback()` aborts; both close the session. On commit failure the GORM session cache is cleared.
- `AbstractMongoSession` holds the active `ClientSession`, starts it in `beginTransactionInternal()`, and routes every read/write through small helpers that pass the session when a transaction is active and stay session-less otherwise.
- The session is threaded through both session engines, both persisters, `MongoQuery`, and the `MongoStaticApi`/`MongoEntity` surface.
- Core `DatastoreTransactionManager` is unchanged — it already orchestrates flush/commit/rollback; this just supplies a `Transaction` that drives a server transaction.

## Opt-in and fallback

- Default is **off** (`grails.mongodb.transactional` defaults to `false`) — no behavior change for existing apps.
- Requires a replica set or sharded cluster. If a standalone topology is detected, the feature is disabled with a one-time warning and GORM falls back to the legacy flush behavior.

## Boundaries

- Identifier generation (the native `Long` counter) is intentionally left non-transactional, mirroring database sequence semantics.
- Like GORM's transaction manager generally, a single flat transaction (`PROPAGATION_REQUIRED`) is supported; `REQUIRES_NEW`/`NESTED` are not.

## Tests

- `MongoTransactionSpec` — commit persists multiple docs; rollback discards them on the server; read-your-writes within a transaction; cross-collection atomic rollback; `findOneAndDelete` participates in the transaction; nested GORM `REQUIRES_NEW`.
- `MongoTransactionDisabledSpec` — default-off keeps the legacy flush behavior.

Targets `8.0.x`. Independent of #15743; this is also the prerequisite for a follow-up Spring Data MongoDB interop module.


borinquenkid (commented):
Hi @codeconsole,

Please keep an eye on #15678 ([GORM: Shared Mapping Registry O(M+N) Scaling](https://github.com/apache/grails-core/pull/15678)), which introduces significant internal structural refactoring to how GormRegistry and MongoDatastore handle tenant routing and fallback resolution.

Since that optimization is targeting 8.0.x-hibernate7, your transaction changes here will be downstream from those modifications. It might be worth checking your diff against those updates to prevent initialization order regressions or multi-tenant signature mismatches when merging into the 8.0 release line.

jdaugherty (commented):
Reading about the MongoDB transactions and asking AI made me think there is a gap here.  Was whole-transaction retry on TransientTransactionError consciously deferred?

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoTransaction.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Override
+    public void setTimeout(int timeout) {
+        // The transaction is started before the manager applies a timeout, so a per-transaction
+        // timeout cannot be applied to the server-side transaction; the server enforces its own
+        // transactionLifetimeLimitSeconds instead. Warn once so a configured timeout is not silently
+        // ignored.
+        if (!warnedTimeoutIgnored) {
+            warnedTimeoutIgnored = true;
+            LOG.warn("A per-transaction timeout was requested but GORM for MongoDB does not apply it to the " +
```
jdaugherty: We should throw an error here, not silently fail.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/AbstractMongoSession.java
```diff
@@ -200,6 +215,120 @@ public MongoMappingContext getMappingContext() {
...
+    // behave identically (session-less) when no transaction is active.
+
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    public BulkWriteResult bulkWrite(com.mongodb.client.MongoCollection collection, List<? extends WriteModel> writes) {
+        return clientSession != null ? collection.bulkWrite(clientSession, writes) : collection.bulkWrite(writes);
+    }
+
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    public DeleteResult deleteMany(com.mongodb.client.MongoCollection collection, Bson filter) {
+        return clientSession != null ? collection.deleteMany(clientSession, filter) : collection.deleteMany(filter);
```
jdaugherty: Shouldn't all of these not null checks use the helper hasActiveTransaction() instead?  Isn't the session only required for the transaction case?

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoTransaction.java
codeconsole: Done — throws `TransactionUsageException` on a non-default timeout now. `doBegin` rolls back the started session and rethrows as `CannotCreateTransactionException`, so nothing leaks.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/AbstractMongoSession.java
codeconsole: Yes — all nine use `hasActiveTransaction()` now. Only the transactional case needs the session, and it falls back to the session-less overload if one ever outlives its transaction.

codeconsole:
Yes — deliberately, and it's a boundary rather than missing work.

Commit-retry is implemented: on an `UnknownTransactionCommitResult` the commit is retried automatically (it may already have succeeded, so retrying is safe).

Retrying the **whole transaction** on a `TransientTransactionError` is intentionally left to the application, because doing it automatically re-runs everything in the transaction body — including side effects like emails or HTTP calls — so it's only safe when that body is idempotent. Spring Data MongoDB's own `MongoTransactionManager` draws the same line: it surfaces transient errors and leaves retry to the application (e.g. Spring Retry). Code that needs it can wrap the call in its own retry.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/gorm/mongo/api/MongoStaticApi.groovy
```diff
@@ -71,9 +71,9 @@ class MongoStaticApi<D> extends GormStaticApi<D> implements MongoAllOperations<D
         withSession { AbstractMongoSession session ->
             def entity = session.mappingContext.getPersistentEntity(persistentClass.name)
             filter = wrapFilterWithMultiTenancy(filter)
-            return session.getCollection(entity)
+            MongoCollection<D> collection = session.getCollection(entity)
```
sbglasius: Here it's named `session` later in other files you name it `mongoSession`, if it's the same, perhaps the name should be the same?

sbglasius (approved):

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/gorm/mongo/api/MongoStaticApi.groovy
codeconsole: Fixed in 134146a4e6 — renamed to `mongoSession` throughout. Went with `mongoSession` since `MongoQuery` can't use `session` (the base `Query` class already declares a `Session session` field).

jdaugherty (approved):
I'm good to merge, but we should better document the retry semantics since the implementation was a conscious decision.

codeconsole:
Done in 973747bdff — added a "Retry Semantics" section to the Multi-Document Transactions docs in `advancedConfig.adoc`.

It documents both halves of the boundary explicitly:

- **`UnknownTransactionCommitResult`** — GORM retries the commit automatically (up to 3 times), since the commit may have already succeeded and MongoDB guarantees re-issuing it will not double-apply the writes.
- **`TransientTransactionError`** — deliberately *not* retried. Re-running the transaction body re-runs its side effects (email, HTTP calls, messaging), which is only safe when the body is idempotent — something only the application can know. This matches the line Spring Data MongoDB's `MongoTransactionManager` draws, and relational GORM behaves the same way for transient serialization/deadlock errors.

The docs also include a worked example of an application-level retry loop using `MongoException.TRANSIENT_TRANSACTION_ERROR_LABEL`, plus a pointer to Spring Retry's `@Retryable` as the declarative alternative. The `MongoTransaction` javadoc already covered the rationale at the code level, so this fills the user-facing gap.

jdaugherty:
You should be good to merge this, if you merge 8.0.x most of these failures will clean up.  Otherwise, I'm addressing them as part of the 7.2 -> 8.x merge that includes the plugin merges

PR #15745, opened 2026-06-17
Add GORM for MongoDB / Spring Data MongoDB interop module

codeconsole:
## Depends on #15744 (stacked PR)

This is stacked on top of #15744 (opt-in MongoDB transactions). Until #15744 merges, the diff here also includes its commit — please review/merge **#15744 first**, after which this PR reduces to just the new interop module. The unified transaction here relies on the `ClientSession` introduced by #15744.

## What

A new **opt-in** module, `grails-data-mongodb-spring-data`, that lets GORM for MongoDB and Spring Data MongoDB run side by side over the **same** `MongoClient`, database and codecs — and, within a single `@Transactional` method, the **same** MongoDB transaction.

When the module and `spring-data-mongodb` are on the classpath of a Spring Boot app that has a GORM `MongoDatastore`, it auto-configures over GORM's connection:

- a `MongoDatabaseFactory` bound to GORM's client + default database (it does **not** close GORM's client),
- a `MongoTemplate` (+ `MappingMongoConverter`) sharing the driver codec registry,
- a primary `transactionManager` (`GormSharedSessionMongoTransactionManager`) that binds GORM's `ClientSession` into Spring Data, so a GORM `save()` and a `MongoTemplate`/repository write in one `@Transactional` method commit or roll back together.

Unified transactions require GORM server-side transactions (`grails.mongodb.transactional = true`, from #15744). Spring Data repositories are enabled the usual way via `@EnableMongoRepositories`, on a package separate from GORM `@Entity` classes.

## Boundary

Only the connection, codecs and (within a transaction) the `ClientSession` are shared; the two object-mapping models stay separate (GORM maps `@Entity`; Spring Data maps its own documents). Sharing the session reaches Spring Data's package-private `MongoResourceHolder`, so a small helper lives in package `org.springframework.data.mongodb` — verified against the Spring Data MongoDB 5.x line in Spring Boot 4, class-path only (not JPMS module-path). A single flat transaction (`PROPAGATION_REQUIRED`) is supported; `REQUIRES_NEW`/`NESTED` are not.

## Tests

- `UnifiedMongoTransactionSpec` — GORM + `MongoTemplate` commit and roll back together on one session; coexistence read; sequential transactions without session/holder leaks; a Spring Data repository over the shared connection; auto-config wiring; the shared factory does not close GORM's client.
- `GormSpringDataSessionSupportSpec` — coupling smoke test that fails loudly if the Spring Data internal shape changes.

Targets `8.0.x`.


borinquenkid:
Hi @codeconsole, 

Please keep an eye on #15678, which modifies the way the MongoRegistry is being handled. It might be worth checking your diff against those updates, as this PR is downstream from 8.0.x-hibernate7.

jdaugherty (commented):
Whats the benefit for having Spring Data and GORM?

codeconsole:
This is for apps that already have or need both, where today the only option is two separate `MongoClient`s with no transactional consistency between them. The module makes that coexistence correct: GORM and Spring Data share one `MongoClient`, database and codec registry, and — within a single `@Transactional` — the same MongoDB `ClientSession`, so a GORM `save()` and a `MongoTemplate`/repository write commit or roll back together.

Concrete cases: Spring-ecosystem libraries built on `MongoTemplate`/Spring Data repositories (Spring Batch, Spring Session, Spring Integration, Spring Security ACL); incremental migration in either direction without doubling connections; and reaching actively-developed Spring Data features for specific operations while keeping GORM for domain modeling.

It's opt-in — a separate module that pulls in `spring-data-mongodb` only when an app adds it; core stays driver-only and the mapping models stay separate.

codeconsole:
Thanks @borinquenkid — I checked #15678 against this PR's diff. It's Hibernate-only: the registry-scaling changes are all in `grails-data-hibernate5`/`grails-data-hibernate7` (plus graphql and gradle config), with no changes to `grails-data-mongodb`, `grails-datastore-core`, or `grails-datamapping-core`. The only shared touchpoint is `gradle/mongodb-test-config.gradle`, which this PR doesn't modify, so it merges cleanly. I'll rebase onto `8.0.x` and re-run the Mongo suites once #15678 lands, but there's no code overlap with the interop module here.

jdaugherty:
Isn't the transaction support a requirement before this PR?

jdaugherty (approved):

jdaugherty (commented):
There's zero docs added for this, we should add docs to grails-doc instead of just the README.  Otherwise, it's approved.

PR #15746, opened 2026-06-18
Convert SiteMesh 3 taglibs to method handlers and use SiteMesh 3.3.0-M1

codeconsole:
## Summary

Follow-up to #15713, addressing review feedback on `RenderSitemeshTagLib`:

> To make this faster, you should use the method version instead of closure.

Converts **both** taglibs shipped by the SiteMesh 3 plugin from closure-based tags to the method handlers introduced in #15465, so GSP dispatch uses the faster reflective method-invocation path instead of cloning a closure on every tag invocation. This also clears the closure-based-tag deprecation warning the #15465 compiler check emits for these classes.

- **`RenderSitemeshTagLib`** (`sitemesh` namespace) — `applyLayout`, `pageProperty`, `ifPageProperty`, `layoutTitle`, `layoutHead`, `layoutBody`, `content`.
- **`Sitemesh3LayoutTagLib`** (`grailsLayout` namespace, `@CompileStatic`) — the compile-time capture tags `captureHead`, `captureBody`, `captureTitle`, `wrapTitleTag`, `captureMeta`, `captureContent`, `parameter`. These are emitted by `GrailsLayoutPreprocessor` and dispatched through `GroovyPage.invokeTag`, which already routes to method handlers (`invokeTagLibMethod` → `TagMethodInvoker`).

After this change, no taglib used by the SiteMesh 3 plugin remains closure-based.

## Behavior preservation

Only the tag declarations change; the tag bodies are untouched. Both the closure and method dispatch paths (`TagOutput.captureTagOutput` and `GroovyPage.invokeTag`) share the same output-stack setup and already pass `TagOutput.EMPTY_BODY_CLOSURE` (never `null`) as the body to two-argument tags, so the existing `body` / `body instanceof Closure` checks resolve identically whether each tag is a closure field or a method.

## SiteMesh 3.3.0-M1

Also bumps the SiteMesh 3 dependencies (`spring-boot-starter-sitemesh`, `spring-webmvc-sitemesh`, and the transitive `org.sitemesh:sitemesh` core) from `3.3.0-SNAPSHOT` to the released **3.3.0-M1** milestone, now on Maven Central. SiteMesh 2 (`opensymphony:sitemesh` 2.6.0, used by `grails-layout`) is unaffected.

## Tests

- Adds `RenderSitemeshTagLibSpec` and `Sitemesh3LayoutTagLibSpec`, which verify via the framework `TagMethodInvoker` API that all fourteen tags are discovered as invokable method handlers and no longer exist as closure fields.
- End-to-end coverage in `grails-test-examples/gsp-sitemesh3` (`EndToEndSpec`) continues to exercise the full capture/decoration pipeline at runtime.

## Validation

- `./gradlew :grails-sitemesh3:test` — green (incl. both new specs)
- `./gradlew :grails-sitemesh3:checkstyleMain :grails-sitemesh3:codenarcMain` — green
- `./gradlew :grails-test-examples-gsp-sitemesh3:integrationTest` — green (`EndToEndSpec` 14/14, `GrailsLayoutSpec` 7/7) against SiteMesh 3.3.0-M1

## Snapshot-repo regex cleanup (addresses #15710 review)

Addresses two review comments on #15710: the generated-app sitemesh snapshot repository's version filter does not need to match `-SNAPSHOT` explicitly, because the repository is already constrained to snapshots — `snapshotsOnly: true` in `CreateAppCommand` and `VersionType.SNAPSHOT` in `GradleRepository`. The version regex is simplified from `.*-SNAPSHOT` to `.*` in both places. (The pre-existing apache/groovy snapshot filters are left untouched.)

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
@@ -72,7 +72,7 @@ public interface GradleRepository extends Ordered {
                 null,
                 List.of(
                     new VersionRegexRepoFilter(
-                        "org[.]sitemesh.*", ".*", ".*-SNAPSHOT"
+                        "org[.]sitemesh.*", ".*", ".*"
                     )
                 ),
                 List.of(VersionType.SNAPSHOT)
--- a/grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
+++ b/grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
@@ -519,7 +519,7 @@ class CreateAppCommand extends ArgumentCompletingCommand implements ProfileRepos
             configuredRepositories.add(repository)
 
             GrailsGradleRepository sitemeshSnapshots = new GrailsGradleRepository(url: 'https://central.sonatype.com/repository/maven-snapshots', snapshotsOnly: true)
-            sitemeshSnapshots.includeOnly('org[.]sitemesh.*', '.*', '.*-SNAPSHOT')
+            sitemeshSnapshots.includeOnly('org[.]sitemesh.*', '.*', '.*')
             configuredRepositories.add(sitemeshSnapshots)
         }
         configuredRepositories.unique()
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -94,8 +94,8 @@ ext {
             'selenium.version'              : '4.38.0',
             'sitemesh.version'              : '2.6.0',
             'spock.version'                 : '2.4-groovy-4.0',
-            'starter-sitemesh.version'      : '3.3.0-SNAPSHOT',
-            'spring-webmvc-sitemesh.version': '3.3.0-SNAPSHOT',
+            'starter-sitemesh.version'      : '3.3.0-M1',
+            'spring-webmvc-sitemesh.version': '3.3.0-M1',
             // Spring Boot 4 no longer manages spring-retry; pin it here so the
             // grails-shell-cli SpringRetryCompilerAutoConfiguration's unversioned
             // reference resolves and consumer apps using @Retryable get a known version.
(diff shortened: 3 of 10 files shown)


grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
```diff
@@ -72,7 +72,7 @@ static Set<GradleRepository> getDefaultRepositories(String grailsVersion) {
                 null,
                 List.of(
                     new VersionRegexRepoFilter(
-                        "org[.]sitemesh.*", ".*", ".*-SNAPSHOT"
+                        "org[.]sitemesh.*", ".*", ".*"
```
jdaugherty, line 75: Is the M1 and actual release?  I don't think this code should change since the below filter will still filter to only snapshots

grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
```diff
@@ -519,7 +519,7 @@ class CreateAppCommand extends ArgumentCompletingCommand implements ProfileRepos
             configuredRepositories.add(repository)
 
             GrailsGradleRepository sitemeshSnapshots = new GrailsGradleRepository(url: 'https://central.sonatype.com/repository/maven-snapshots', snapshotsOnly: true)
-            sitemeshSnapshots.includeOnly('org[.]sitemesh.*', '.*', '.*-SNAPSHOT')
+            sitemeshSnapshots.includeOnly('org[.]sitemesh.*', '.*', '.*')
```
jdaugherty, line 522: Same
comment, I don't think this needs to change

grails-forge/grails-forge-core/src/main/java/org/grails/forge/build/gradle/GradleRepository.java
codeconsole: @jdaugherty  per @matrei it's redundant and not needed. https://github.com/apache/grails-core/pull/15710

grails-shell-cli/src/main/groovy/org/grails/cli/profile/commands/CreateAppCommand.groovy
codeconsole: @jdaugherty  per @matrei it's redundant and not needed. https://github.com/apache/grails-core/pull/15710

sbglasius (approved):

jdaugherty (approved):

PR #15747, opened 2026-06-18
Support TTL indexes in the GORM-Mongo mapping DSL and reconcile index option changes

codeconsole:
## What

Two related improvements to MongoDB index initialisation in `MongoDatastore.initializeIndices`.

### 1. TTL indexes can be declared in the mapping DSL

A single-field index declared with `expireAfterSeconds` now materialises a MongoDB TTL index:

```groovy
static mapping = {
    created index: true, indexAttributes: [expireAfterSeconds: 7776000]   // 90 days
}
```

Previously this was impossible. `MongoConstants.mapToObject` builds `IndexOptions` by reflectively invoking same-named **single-argument** setters, but the driver's only TTL setter is the **two-argument** `IndexOptions.expireAfter(Long, TimeUnit)`, which the mapper can't reach. So `expireAfterSeconds` is now pulled out of the index attributes and applied explicitly.

TTL is single-field only — MongoDB rejects `expireAfterSeconds` on a compound index with `CannotCreateIndex`; that error surfaces unchanged.

### 2. Index-option conflicts are reconciled, not just logged

All three index paths (declared `getIndices()`, `getCompoundIndices()`, per-property) now route through one helper. On `IndexOptionsConflict` (server error 85 — an index already exists on these keys with different options) it reconciles instead of only logging the stack trace:

- **TTL value change** (e.g. a configurable retention changed between restarts) is applied in place via `collMod` — no drop, no rebuild, no window without an index.
- **Any other conflicting change** is dropped and recreated only when `indexAttributes: [recreateOnConflict: true]` is declared; otherwise it logs a clear, actionable message. A unique index is never silently dropped.

## Why

Without DSL support, a TTL index has to be created imperatively (raw-driver `createIndex` in a bootstrap hook), and changing its retention needs hand-rolled drop-and-recreate logic. That is easy to get wrong: a stale plain index on the same key blocks the TTL create, and the TTL then silently never applies. Declaring it in the mapping, with in-place reconciliation handled by the framework, removes both the boilerplate and that footgun.

## Tests

`TtlIndexSpec` (runs against the Mongo testcontainer) covers:
- `expireAfterSeconds` in the mapping produces a TTL index with the declared value.
- Mutating the live TTL and re-initialising restores the declared value on the same index (one index, no conflict) — exercising the `collMod` reconcile path.

Existing `IndexAttributesAndCompoundKeySpec` (unique / compound) still passes.


jdaugherty (dismissed):
1. We need to add new features to the "whats new guide" 
2. let's make sure there is a functional test app with the old behavior + new behavior (expiring vs not)

codeconsole:
> 1. We need to add new features to the "whats new guide"
> 2. let's make sure there is a functional test app with the old behavior + new behavior (expiring vs not)

@jdaugherty done

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+        }
+
+        try {
+            collection.createIndex(keys, indexOptions);
+        } catch (MongoCommandException e) {
+            if (e.getErrorCode() == INDEX_OPTIONS_CONFLICT_CODE) {
+                reconcileIndexConflict(entity, collection, keys, indexOptions,
+                        expireAfterSeconds, recreateOnConflict, descriptor, e);
+            } else {
+                LOG.error("Failed to create index for entity [" + entity.getName() + "] " + descriptor + ": " + e.getMessage(), e);
```
matrei:
```suggestion
                LOG.error("Failed to create index for entity [{}] {}: {}",
                    entity.getName(), descriptor, e.getMessage(), e);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+    private void reconcileIndexConflict(PersistentEntity entity,
+                                        com.mongodb.client.MongoCollection<Document> collection,
+                                        Document keys, IndexOptions desired, Long expireAfterSeconds,
+                                        boolean recreateOnConflict, String descriptor, MongoCommandException original) {
+        Document existing;
+        try {
+            existing = findIndexByKeyPattern(collection, keys);
+        } catch (RuntimeException listError) {
+            LOG.error("Failed to create index for entity [" + entity.getName() + "] " + descriptor +
+                    " and could not inspect existing indexes: " + listError.getMessage(), original);
```
matrei:
```suggestion
            LOG.error("Failed to create index for entity [{}] {} and could not inspect existing indexes: {}",
                entity.getName(), descriptor, listError.getMessage(), original);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+            return;
+        }
+        if (existing == null) {
+            LOG.error("Failed to create index for entity [" + entity.getName() + "] " + descriptor + ": " + original.getMessage(), original);
```
matrei:
```suggestion
            LOG.error("Failed to create index for entity [{}] {}: {}",
                entity.getName(), descriptor, original.getMessage(), original);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+
+        // TTL change on an existing index — update in place, no rebuild, no gap.
+        boolean ttlChange = expireAfterSeconds != null && !expireAfterSeconds.equals(existingTtlSeconds);
+        if (ttlChange) {
+            try {
+                getMongoClient().getDatabase(getDatabaseName(entity))
+                        .runCommand(new Document("collMod", getCollectionName(entity))
+                                .append("index", new Document("name", existingName)
+                                        .append(INDEX_EXPIRE_AFTER_SECONDS, expireAfterSeconds)));
+                LOG.info("Updated TTL of index [" + existingName + "] on entity [" + entity.getName() + "] to " + expireAfterSeconds + "s");
```
matrei:
```suggestion
                LOG.info("Updated TTL of index [{}] on entity [{}] to {}s",
                    existingName, entity.getName(), expireAfterSeconds);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+                return;
+            } catch (MongoCommandException collModError) {
+                // collMod can't make every change (e.g. add a TTL to a non-TTL index on older
+                // servers) — fall through to recreate (if authorised) rather than fail outright.
+                LOG.warn("collMod TTL update failed for index [" + existingName + "] on entity [" + entity.getName() + "]: " +
+                        collModError.getMessage() + (recreateOnConflict ? " — recreating" : ""));
```
matrei:
```suggestion
                LOG.warn("collMod TTL update failed for index [{}] on entity [{}]: {}{}",
                    existingName, entity.getName(), collModError.getMessage(), recreateOnConflict ? " — recreating" : "");
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+            }
+        }
+
+        if (recreateOnConflict) {
+            try {
+                collection.dropIndex(existingName);
+                collection.createIndex(keys, desired);
+                LOG.info("Recreated index [" + existingName + "] on entity [" + entity.getName() + "] " + descriptor);
```
matrei:
```suggestion
                LOG.info("Recreated index [{}] on entity [{}] {}", existingName, entity.getName(), descriptor);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+            } catch (MongoCommandException recreateError) {
+                LOG.error("Failed to recreate index [" + existingName + "] on entity [" + entity.getName() + "] " + descriptor + ": " + recreateError.getMessage(), recreateError);
```
matrei:
```suggestion
                LOG.error("Failed to recreate index [{}] on entity [{}] {}: {}",
                    existingName, entity.getName(), descriptor, recreateError.getMessage(), recreateError);
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+            }
+            return;
+        }
+
+        LOG.error("Index conflict for entity [" + entity.getName() + "] " + descriptor + ": an index [" + existingName +
+                "] already exists on the same keys with different options. Declare indexAttributes:[recreateOnConflict:true]" +
+                " to drop and recreate it. Original error: " + original.getMessage());
```
matrei:
```suggestion
        LOG.error(
            "Index conflict for entity [{}] {}: an index [{}] already exists on the same keys with different options. " +
                "Declare indexAttributes:[recreateOnConflict:true] to drop and recreate it. Original error: {}",
            entity.getName(), descriptor, existingName, original.getMessage());
```

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/MongoDatastore.java
```diff
@@ -897,20 +904,180 @@ protected void initializeIndices(final PersistentEntity entity) {
...
+            if (!existingKey.containsKey(entry.getKey())) {
+                return false;
+            }
+            Object a = existingKey.get(entry.getKey());
+            Object b = entry.getValue();
+            if (a instanceof Number && b instanceof Number) {
+                if (((Number) a).doubleValue() != ((Number) b).doubleValue()) {
+                    return false;
+                }
+            } else if (a == null ? b != null : !a.equals(b)) {
```
matrei:
```suggestion
            } else if (!Objects.equals(a, b)) {
```

matrei (approved):

sbglasius (approved):

jdaugherty (approved):

PR #13863, opened 2024-11-20
Stop @EnableWebMvc from being automatically added to Grails Applications

codeconsole:
Fixes #13843

This produces unknown/unexpected behavior and should not be done.  @EnableWebMvc is a Spring Framework configuration annotation not to be used with Spring Boot.  It produces undesired behavior by automatically creating beans that are not wanted. WebMvcAutoConfiguration performs all the behavior of this configuration because it extends it and adds conditional logic. 

Fixes 
```
Overriding bean definition for bean 'localeResolver' with a different definition: replacing [Root bean: class=null; scope=; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=org.springframework.web.servlet.config.annotation.DelegatingWebMvcConfiguration; factoryMethodName=localeResolver; initMethodNames=null; destroyMethodNames=[(inferred)]; defined in org.springframework.web.servlet.config.annotation.DelegatingWebMvcConfiguration] with [Generic bean: class=org.springframework.web.servlet.i18n.SessionLocaleResolver; scope=singleton; abstract=false; lazyInit=false; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=null; factoryMethodName=null; initMethodNames=null; destroyMethodNames=null]
Overriding bean definition for bean 'org.springframework.aop.config.internalAutoProxyCreator' with a different definition: replacing [Root bean: class=org.springframework.aop.framework.autoproxy.InfrastructureAdvisorAutoProxyCreator; scope=; abstract=false; lazyInit=null; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=null; factoryMethodName=null; initMethodNames=null; destroyMethodNames=null] with [Generic bean: class=org.grails.spring.aop.autoproxy.GroovyAwareInfrastructureAdvisorAutoProxyCreator; scope=singleton; abstract=false; lazyInit=false; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=null; factoryMethodName=null; initMethodNames=null; destroyMethodNames=null]
```

```java
class WebMvcAutoConfiguration {
    static class EnableWebMvcConfiguration extends DelegatingWebMvcConfiguration { 
    } 
}

class DelegatingWebMvcConfiguration extends WebMvcConfigurationSupport {
}

[...]

Issue #14915, opened 2025-07-16
Spring 7 - XML configuration no longer supported


jdaugherty:
### Issue description

Spring has stated that xml driven configuration (at least for MVC) is not going to be supported in Spring Framework 7 (it will be there, but Spring won't support it).  However, they're separately adding a programmatic way to add beans.  We'll likely need to rework the bean builder as a result since it's ultimately backed by xml.

https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-7.0-Release-Notes#deprecations

jdaugherty:
FYI: the `BeanRegistrar` interface is how Spring decided to implement this, which has this sole method: `register(BeanRegistry registry, Environment environment)` 

Documentation for registry is here: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/BeanRegistry.html 

Someone put together an example repository here: https://github.com/danvega/sb4-bean-registrar for usage.

jdaugherty:
Desired approach: 
1. mark the dsl deprecated
2. update the docs to indicate the dsl is removed / replace with the BeanRegistrar examples
3. inject the BeanRegistrar where we wire beans today
4. Remove the bean dsl in Grails 9

jdaugherty:
This is not fixed yet.

jdaugherty:
I've requested we revert the changes merged by: https://github.com/apache/grails-core/pull/15824#issuecomment-5208595820 

This PR isn't doing what it suggested.

jdaugherty:
I revisited this and confirmed that the original change was not needed: 

- It fixed a problem that did not exist. The plain resources.groovy DSL never went through Spring XML. It registered programmatic bean definitions all along. The XML reader was constructed at startup and then only used by xmlns and .xml imports.
- It did not remove any XML dependency. The same four Spring XML classes are still required for xmlns and importBeans. They just moved into a lazily built holder. If Spring ever drops them, Grails is in the same position either way, and none of them are deprecated in Spring 7.0.9.
- It did not do what the issue asks for. Issue 14915 wants the DSL deprecated and re-implemented on BeanRegistrar. The PR did not touch that.


The actual bean dsl enhancements & moving to `@Configuration` is the real fix to this ticket.  We've moved the majority of Grails to use these internally and will be eventually removing `resources.groovy`.  There's no further change required for this so closing this and moving it to RC1.

PR #15755, opened 2026-06-23
Register plugin beans before Spring Boot auto-configuration (doWithSpringBeforeAutoConfiguration)

codeconsole:
## Summary

Adds a new plugin lifecycle phase, `doWithSpringBeforeAutoConfiguration()`, that lets a Grails plugin register Spring beans **before** Spring Boot auto-configuration runs. Boot beans guarded by `@ConditionalOnMissingBean` then **defer** to the plugin's bean, instead of the plugin having to **override or remove** the Boot bean afterwards (the only thing the existing `doWithSpring()` can do, because it runs *after* auto-configuration).

This is the root-cause fix for an entire class of friction: the "Overriding bean definition for bean 'localeResolver'" warnings, and the `removeBeanDefinition`/alias hacks Grails uses to bend Boot's beans into its own model.

> **Depends on #13863** ("Stop `@EnableWebMvc` from being automatically added"). This branch is stacked on it, so the first 4 commits here are from that PR — the diff will shrink to just this feature once #13863 merges. The feature *requires* #13863: Boot's `WebMvcAutoConfiguration` must be active for `@ConditionalOnMissingBean` to have anything to defer.

## Why

`GrailsApplicationPostProcessor` (which drains `doWithSpring`) is an un-ordered `BeanDefinitionRegistryPostProcessor`; Boot's `ConfigurationClassPostProcessor` is `PriorityOrdered` and expands every `@AutoConfiguration` first. So **every `doWithSpring` bean is registered after auto-config** — a plugin can only override or remove a Boot bean, never make `@ConditionalOnMissingBean` back off. That late override is also wasteful: the Boot bean is built, then thrown away.

## How

A second `BeanDefinitionRegistryPostProcessor` (`GrailsBeforeAutoConfigurationPostProcessor`) is added via `addBeanFactoryPostProcessor` (manually-registered BDRPPs run **before** registry-discovered ones such as `ConfigurationClassPostProcessor`). It drains each plugin's `doWithSpringBeforeAutoConfiguration` closure into the registry ahead of auto-config. A narrow, opt-out warning (`DeferrableOverrideWarner`) nudges the one anti-pattern this removes — a late `doWithSpring` bean overriding a name-guarded `@ConditionalOnMissingBean` auto-config bean that *would* have deferred.

## Two permanent phases (not modern-vs-legacy)

`doWithSpring` is **not** deprecated. It stays the correct phase for: decorating/wrapping a bean auto-config already created; aggregating/inspecting the full registry; artefact-driven beans (one per controller/service/taglib — those need the loaded `GrailsApplication` and stay late); and intentionally overriding an *unconditional* Boot bean. Only "register late solely to override a deferrable conditional" moves to the new phase.

## What's included

- **Mechanism**: the `doWithSpringBeforeAutoConfiguration()` lifecycle method (default no-op on `GrailsApplicationLifeCycle`/`Plugin`), the drain plumbing, the BDRPP + initializer (`spring.factories`).
- **Warning**: `DeferrableOverrideWarner`, gated on `grails.plugins.warnOnDeferrableOverride` (default on), precise to name-guarded `@ConditionalOnMissingBean` overrides.
- **First migrations** to the overridable pattern (each still wins by default; now overridable without a warning):
  - **grails-i18n**: `localeResolver`, `localeChangeInterceptor`, `messageSource` → `@ConditionalOnMissingBean`. This is exactly what #15751 attempted and could not achieve alone — it works here because #13863 removed `@EnableWebMvc`, so grails-i18n's `@AutoConfigureBefore(WebMvcAutoConfiguration)` wins the race.
  - **grails-url-mappings** / **grails-controllers**: the Grails-specific beans (`grailsCorsFilter`, `urlMappingsErrorPageCustomizer`, `urlMappingsInfoHandlerAdapter`, `grailsWebRequestFilter` registration, `webMvcConfig`). `dispatcherServlet`/`multipartConfigElement` are deliberately left as intentional Boot replacements (they'd need `@AutoConfigureBefore(DispatcherServlet/MultipartAutoConfiguration)` to stay safe).
- **Tests**: a unit test proving the ordering makes `@ConditionalOnMissingBean` defer (+ control); a `DeferrableOverrideWarner` precision test; and an `@Integration` test (app3) proving a real plugin's bean wins over an app's `@ConditionalOnMissingBean` default through a real boot.
- **Docs**: a "Registering Beans Before Auto-Configuration" section in the plugins guide.

## Design notes

- **The contract** (documented on the API): the closure must define bean definitions only and be side-effect-free — the phase runs an isolated lightweight plugin load, and the real lifecycle re-instantiates plugins, so side effects would run twice. The early phase deliberately does **not** propagate the application context to its throwaway plugin instances, so a plugin implementing `ApplicationListener` is registered only once.
- **Rejected: sharing one plugin manager/`GrailsApplication` across the two phases.** It's fragile — `setApplication()` doesn't update the field `DefaultGrailsPlugin` binds `doWithSpring`'s `application` from, so rebinding breaks artefact handling (`tagLibClasses`). The isolated lightweight load is the robust design; the second (cheap) plugin instantiation is benign.
- **Application classes don't need this phase**: an app class is a `@Configuration`, and Boot processes user config before auto-config, so a plain `@Bean` already registers ahead of auto-config. The phase is plugin-scoped.
- **AOT/native**: introduces no new risk — it mirrors the existing `GrailsApplicationPostProcessor` runtime-plugin-load + `BeanBuilder` mechanism.

## Validation

Full `:grails-core:test` green; app1 (GSP) and app3 (Hibernate) boot clean; the app3 integration test proves deferral end-to-end; every migration verified via the conditions report (bean still created by default, Boot's defers).

## Suggested rollout

Ship opt-in + the warning → migrate framework beans incrementally (i18n + the Tier-1 set done here) → make the modern phase the default registration timing → remove the legacy override path. Each conversion deletes one Boot-fighting site.


jdaugherty (changes requested):
I don't think we should be adding *another* lifecycle event like this PR does.  We have previously discussed this as part of the Groovy Bean DSL retirement.  Spring introduced a new interface for bean registration, and we should instead just move to initializing the beans at the same time as we do the config.  We didn't do this in 7 due to the impact, and we haven't done it in 8 yet due to higher priorities.  If you're willing to do this work, let's discuss and I can point you to the tickets / previous discussions on how this should work.

jdaugherty:
Adding this comment from slack: My concern is we're going to add this new way, then add yet another way over 2 major versions.

There are 2 components to this change:
1. the bean dsl needs deprecated. We discussed it in the weekly and we plan to add a method that just uses the BeanRegistry that spring added in 4 to directly wire - with the intent to remove the bean dsl. This doesn't have to be done as part of your PR. https://github.com/apache/grails-core/issues/14915 is the ticket for this part.

2. in 7.x I added the PluginDiscovery interface and shipped a default implementation. the purpose of this was to move the discovery process earlier in the spring lifecycle and integrate it so anything done in grails config is at the same point in the lifecycle that spring config would be. The class GrailsEnvironmentPostProcessor currently uses that to find the plugins and then add the configuration to the environment at the same point in the spring lifecycle that spring does (this means that the config is now usable in the spring annotations - ie the conditional annotations based on config). Prior to my change, spring would initialize, then as part of the application context creation, we would use the PluginManager to locate plugins, add their configuration, and then add their beans. I moved the find + config process to GrailsEnvironmentPostProcessor but never moved up the bean wiring. Your PR appears is the start at moving them up. My hope was we can rip out the legacy path and go directly to the beanRegistry API in the best case. Worst case we would simplify the grails api entry points to assume they would initialize earlier (which fixes the problem your PR identified)

codeconsole:
Agreed on all points — adopting your direction, and closing this PR in favor of a replacement rather than patching it. The independently useful pieces have already been extracted (#13863 got the hardening delta, #15916 has the i18n/URL-mappings `@ConditionalOnMissingBean` migrations, stacked on #13863).

Since we're close to the last 8.0.0 milestone and major changes shouldn't land in an RC, this goes directly to your best case in 8.0 rather than staging it across two majors. What the replacement PR does:

1. **Retimes the existing entry point — no new method.** `doWithSpring`/`doWithRuntimeConfiguration` runs before Spring Boot auto-configuration, via a `BeanDefinitionRegistryPostProcessor` added ahead of `ConfigurationClassPostProcessor`. The plugin manager and `GrailsApplication` are built exactly once from your `PluginDiscovery` — no second plugin manager, no double plugin instantiation. This is the follow-up you scoped out of #15409 ("we could move plugins completely to the autoconfiguration workflow"), completed.
2. **Artefact discovery moves into the same early phase**, so `doWithSpring` closures that iterate artefacts (controllers, services, interceptors) work unmodified.
3. **`BeanRegistrar` is the native, documented registration API for plugins.** Plugins expose a `BeanRegistrar` and we apply it pre-refresh through Spring's own `BeanRegistryAdapter` — the same mechanism `GenericApplicationContext.register(BeanRegistrar...)` uses.
4. **The bean DSL is deprecated in 8.0** (`@Deprecated(since = '8.0')` on `doWithSpring`) pointing at (3) — the visible half of #14915.

One deliberate exclusion: reimplementing the DSL's *internals* on top of `BeanRegistry` (the other half of #14915). The DSL supports constructs with no clean `BeanRegistry` equivalent (parent/abstract bean definitions, late-resolved bean references, `BeanConfiguration` post-processing), the plumbing is unobservable to users, and the translation work gets deleted anyway when the DSL is removed. Deprecated DSL closures keep draining through the existing `RuntimeSpringConfiguration` path — just earlier — until removal.

Things we verified rather than assumed, in case any of it contradicts your notes:

- **Dev reload is unaffected.** No reload path re-runs `doWithSpring` — `informOfClassChange` → `onChange` writes straight to the registry, bypassing `GrailsApplicationPostProcessor` — so the initial drain's position doesn't matter to reloading.
- **App-over-plugin overrides keep working.** Grails forces `allowBeanDefinitionOverriding=true` by default (`GrailsApp`), so `resources.groovy`/application beans still replace early-registered plugin beans. The one edge: an app that explicitly sets `spring.main.allow-bean-definition-overriding=false` now gets a `BeanDefinitionOverrideException` where the old single-drain merge was silent — arguably more correct for someone opting into strictness, and it's called out in the upgrade notes.
- **AOT is neutral-to-better.** `refreshForAotProcessing` executes all registry post-processors, so early registration participates in AOT identically — and build-time `@ConditionalOnMissingBean` evaluation now sees plugin beans, so baked condition outcomes match runtime semantics instead of baking in an auto-configured bean the late drain would override.
- **Closures touching `applicationContext` pre-refresh:** audited every `doWithSpring` in-repo; all uses are pre-refresh-safe surface (environment, classLoader, resources, the registry itself — nothing calls `getBean`). The verification sweep did catch two real regressions this way — OSIV detection gating on the auto-configured `dispatcherServlet` bean, and artefact scanning missing the injected `packageNames()` — both fixed in the replacement PR, with details there.

Net effect for 8.0: one lifecycle phase, `BeanRegistry` as the one true way, DSL deprecated but working — no transitional API ever exists, which I think addresses the "new way, then yet another way over 2 major versions" concern directly.

Superseded by #15934 — closing.

PR #15759, opened 2026-06-24
Add opt-in to compile all controllers, services and tag libraries with @GrailsCompileStatic

codeconsole:
### Summary

Adds an opt-in that compiles **every** controller, service and tag library with `@GrailsCompileStatic` automatically, without annotating each class. The opt-ins are exposed as a nested `compileStatic { }` block on the `grails` Gradle extension:

```groovy
grails {
    compileStatic {
        controllers = true   // every grails-app/controllers class
        services = true      // every grails-app/services class
        tagLibs = true       // every grails-app/taglib class
    }
}
```

…or the `all = true` shortcut to enable all three at once:

```groovy
grails {
    compileStatic {
        all = true
    }
}
```

Every flag is a lazy `Property<Boolean>` that defaults to `false` and is read when the Groovy compile task runs (not at configuration time), so the values reflect the `grails { }` block regardless of configuration ordering. This automates what the framework's own guidance already recommends doing per-class, while keeping it opt-in and reversible.

### Design

The configuration rides the same path PR #15118 established for the `grails { importJavaTime / starImports }` DSL, but the annotation is applied through an artefact-aware AST injector rather than a global `ImportCustomizer`, because static compilation is selective (controllers/services/taglibs only), semantic (can fail the build), and must yield to an explicit per-class choice — none of which a blanket customizer can do.

- **Config + transport (grails-gradle):** `GrailsExtension` gains a nested `compileStatic { }` block backed by `GrailsCompileStaticOptions` (modeled on the existing `ViewCompileOptions`), whose `all` / `controllers` / `services` / `tagLibs` opt-ins are lazy `Property<Boolean>` flags. A `GrailsCompileStaticArtefactsProvider` (a `CommandLineArgumentProvider`, modeled on the existing `GrailsAppBaseDirProvider`) reads those lazy options and publishes the enabled flags to the Groovy compiler worker JVM as `-D` system properties. The property names live in `grails.util.BuildSettings` alongside `APP_BASE_DIR`.
- **Application (grails-core):** `CompileStaticArtefactInjector` (an `@AstTransformer` `GrailsArtefactClassInjector` for `Controller`/`Service`/`TagLib`) reads the `@Artefact` type, checks the corresponding flag, and applies `@GrailsCompileStatic` via a new `GrailsASTUtils.addGrailsCompileStaticAnnotation(ClassNode)` helper. That helper reproduces `@GrailsCompileStatic` (i.e. `@CompileStatic` carrying the Grails type-checking extensions) and registers the static-compile transform via `ClassNode.addTransform`, the same mechanism the existing `addCompileStaticAnnotation` uses so a late-added annotation still fires. For the `tagLibs` opt-in the injector additionally skips tag libraries that still declare tags as closure fields (see below).

### Tag library type-checking

`@GrailsCompileStatic` already silences tag dispatch in controllers, but the type-checking extension that does it was controller-only and named `ControllerTagLibTypeCheckingExtension`. This PR generalizes it to `*TagLib` classes too (both carry the `TagLibraryInvoker` trait at runtime) and renames it to **`TagLibraryInvokerTypeCheckingExtension`** to reflect the broader scope. It also makes the extension defer — rather than throw — when there is no enclosing method, so a deprecated **closure-field** tag that dispatches to other tags fails with a normal type error instead of crashing the compiler with an NPE.

**Method-based tags** (`def myTag(Map attrs, Closure body)`) are the supported form. Closure-field tags are already deprecated in Grails 8, so the `tagLibs` build opt-in detects a tag library that still uses them — a public, non-static closure-valued property, the same signal Grails uses to register tags — and **skips it automatically**, leaving it dynamically compiled and emitting a build warning rather than failing the build. An explicit `@GrailsCompileStatic` on the class still forces static compilation (and still fails cleanly on closure-field tag dispatch via the extension above); the auto-skip applies only to the blanket build opt-in.

### Opt-out always wins

A class that declares its own `@CompileDynamic` (or `@GrailsCompileStatic` / `@CompileStatic` / `@GrailsTypeChecked` / `@TypeChecked`) keeps that setting; the opt-in never overrides an explicit choice:

```groovy
@CompileDynamic
class LegacyController {
    // compiled dynamically even when compileStatic { controllers = true } is enabled
}
```

When the flags are unset (the default), the injector is fully inert.

### Tests

- `CompileStaticArtefactInjectorSpec` — controller/service/taglib compiled statically when its flag is on (a type error fails compilation), dynamically when off, `@CompileDynamic` opted out even when on, and each type unaffected by the others' flags; a legacy closure-field tag library is skipped (left dynamic) under the `tagLibs` opt-in, while a static closure and a static non-closure field do not trip that detection.
- `TagLibraryInvokerTypeCheckingExtensionSpec` — a `@GrailsCompileStatic` taglib with method-based tags can invoke tags on `this` and via a namespace dispatcher; type errors are still caught; a deprecated closure-field tag fails cleanly (no compiler NPE); all existing controller cases still pass.
- `GrailsASTUtilsSpec` — `addGrailsCompileStaticAnnotation` applies `@CompileStatic` with the full Grails extension list; `hasStaticCompilationAnnotation` detects every opt-out annotation; the extension list stays in sync with `@GrailsCompileStatic`.
- `GrailsCompileStaticOptionsSpec` — the nested options' lazy flags default to false, are independent, and support both the `property = value` DSL convenience and `property.set(...)`.
- `GrailsCompileStaticArtefactsProviderSpec` + `GrailsExtensionSpec` — the `compileStatic { }` block (and the `all` shortcut) defaults to off, is configurable, is read lazily after construction, and is published as the expected `-D` system properties.

Full `grails-core` suite and the `grails-gradle` plugins/model suites pass; checkstyle + codenarc clean on the changed modules.

### Docs

Documented under the GrailsCompileStatic guide section and the Grails 8 "what's new" page (public DSL only), including the method-based-tags guidance for tag libraries.

### DIFF
--- a/grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
+++ b/grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
@@ -28,6 +28,7 @@ import org.codehaus.groovy.ast.expr.PropertyExpression
 import org.codehaus.groovy.ast.expr.VariableExpression
 import org.codehaus.groovy.transform.stc.GroovyTypeCheckingExtensionSupport
 import org.codehaus.groovy.transform.stc.StaticTypesMarker
+import org.grails.compiler.injection.CompileStaticArtefactInjector
 import org.grails.core.artefact.ControllerArtefactHandler
 
 /**
@@ -91,10 +92,8 @@ class ControllerTagLibTypeCheckingExtension extends GroovyTypeCheckingExtensionS
             newScope {
                 // Both controllers and tag libraries dispatch tags at runtime through the
                 // TagLibraryInvoker trait, so both receive the same dynamic-dispatch silencing.
-                // 'TagLib' is matched as a literal because grails-core has no dependency on the
-                // taglib module; it mirrors org.grails.core.artefact.gsp.TagLibArtefactHandler.TYPE.
                 isTagDispatcher = classNode.name.endsWith(ControllerArtefactHandler.TYPE) ||
-                        classNode.name.endsWith('TagLib')
+                        classNode.name.endsWith(CompileStaticArtefactInjector.TAGLIB_TYPE)
                 dynamicNamespaceProperties = [] as Set
             }
         }
--- a/grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
+++ b/grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
@@ -31,21 +31,20 @@ import org.codehaus.groovy.transform.stc.StaticTypesMarker
 import org.grails.core.artefact.ControllerArtefactHandler
 
 /**
- * A type-checking extension that allows {@code @GrailsCompileStatic} controllers
- * to invoke tag library methods without compile-time errors.
+ * A type-checking extension that allows {@code @GrailsCompileStatic} controllers and
+ * tag libraries to invoke tag library methods without compile-time errors.
  *
- * <p>Tag calls in controllers are dispatched at runtime through
+ * <p>Tag calls in controllers and tag libraries are dispatched at runtime through
  * {@code TagLibraryInvoker#methodMissing} and
  * {@code TagLibraryInvoker#propertyMissing}. These hooks are
  * invisible to the static type checker, so this extension marks the affected
  * expressions as dynamic, silencing the false-positive errors while preserving
- * full type checking for all other code in the controller.
+ * full type checking for all other code in the class.
  *
- * <p>Controller detection mirrors {@code ControllerActionTransformer}: a class is
- * treated as a controller when its qualified name ends with {@code "Controller"}.
- * Note that inner classes whose simple name ends with {@code "Controller"} (e.g.,
- * a {@code FooController} defined inside a service) will also receive this
- * silencing treatment.
+ * <p>Detection is by name suffix: a class is treated as tag-dispatching when its
+ * qualified name ends with {@code "Controller"} or {@code "TagLib"} (both carry the
+ * {@code TagLibraryInvoker} trait at runtime). Note that inner classes whose simple
+ * name ends with one of those suffixes will also receive this silencing treatment.
  *
  * <p>Two calling patterns are supported:
  * <ul>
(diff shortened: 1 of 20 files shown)


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
```diff
@@ -93,6 +93,40 @@ class GrailsExtension {
...
+     */
+    boolean compileStaticArtefacts = false
+
+    /**
+     * Whether every controller under {@code grails-app/controllers} should be compiled with
+     * {@code @GrailsCompileStatic} automatically. Disabled by default. Individual controllers can opt out
+     * by carrying their own {@code @CompileDynamic} (or {@code @CompileStatic}/{@code @GrailsCompileStatic})
+     * annotation, which is always honoured over this default.
+     */
+    boolean compileStaticControllers = false
```
jdaugherty: We should be adding lazy properties instead of booleans so we can configure at execution instead of configuration time.  I also think it's worth adding a nested object here so the config could look like: 

		grails {
			compileStatic {
				controllers = true
			}
		}

codeconsole: good idea, done

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
```diff
@@ -90,7 +89,12 @@ class ControllerTagLibTypeCheckingExtension extends GroovyTypeCheckingExtensionS
 
         beforeVisitClass { ClassNode classNode ->
             newScope {
-                isController = classNode.name.endsWith(ControllerArtefactHandler.TYPE)
+                // Both controllers and tag libraries dispatch tags at runtime through the
+                // TagLibraryInvoker trait, so both receive the same dynamic-dispatch silencing.
+                // 'TagLib' is matched as a literal because grails-core has no dependency on the
+                // taglib module; it mirrors org.grails.core.artefact.gsp.TagLibArtefactHandler.TYPE.
+                isTagDispatcher = classNode.name.endsWith(ControllerArtefactHandler.TYPE) ||
+                        classNode.name.endsWith('TagLib')
```
sbglasius: Instead use `CompileStaticArtefactInjector.TAGLIB_TYPE` as text constant.

sbglasius (approved):
One minor thing to fix, otherwise ok

grails-core/src/main/groovy/org/grails/compiler/ControllerTagLibTypeCheckingExtension.groovy
codeconsole: done, thanks

grails-core/src/main/groovy/org/grails/compiler/TagLibraryInvokerTypeCheckingExtension.groovy
```diff
@@ -90,7 +90,10 @@ class ControllerTagLibTypeCheckingExtension extends GroovyTypeCheckingExtensionS
 
         beforeVisitClass { ClassNode classNode ->
             newScope {
-                isController = classNode.name.endsWith(ControllerArtefactHandler.TYPE)
+                // Both controllers and tag libraries dispatch tags at runtime through the
```
jdaugherty, line 93: This class is now a generic TagLib compilestatic and no longer specific to controllers. We should rename it.

grails-core/src/main/groovy/org/grails/compiler/TagLibraryInvokerTypeCheckingExtension.groovy
```diff
@@ -126,6 +129,17 @@ class ControllerTagLibTypeCheckingExtension extends GroovyTypeCheckingExtensionS
...
 
+    /**
+     * {@code makeDynamic} stores its dynamic-resolution marker on the enclosing method, so it can only be
+     * used inside a method body. Tags defined the deprecated way - as closure fields rather than methods -
+     * are type-checked as field initializers with no enclosing method; silencing there would throw an NPE.
+     * In that case we defer, leaving the call to be reported as a normal type error rather than crashing
+     * the compiler. Method-based tags (the supported form) always have an enclosing method.
+     */
+    private boolean canMakeDynamic() {
+        getEnclosingMethod() != null
```
jdaugherty, line 140: If I undersatnd correctly, this means the legacy tag libs aren't supported now?

codeconsole: they required `@CompileDynamic`, but I just updated it to just not attempt static compilation and log a warning instead

grails-core/src/main/groovy/org/grails/compiler/TagLibraryInvokerTypeCheckingExtension.groovy
codeconsole: done

jdaugherty (approved):

PR #15765, opened 2026-06-24
8.0.x - make configurable: allow-bean-definition-overriding / allow-circular-references

codeconsole:
## Summary

We shouldn't force bad practices and start providing an exit strategy

### [WARNING: Bean overriding will be deprecated in a future release.](https://docs.spring.io/spring-framework/reference/core/beans/definition.html)

`GrailsApp` and the unit-test harness `GrailsApplicationBuilder` hardcoded
`allowBeanDefinitionOverriding(true)` and `allowCircularReferences(true)`. This
silently overrode any value an application set through the standard Spring Boot
properties `spring.main.allow-bean-definition-overriding` and
`spring.main.allow-circular-references` — so there was no way to turn either off.

Both now resolve those properties from the configured environment and **default
to `true`** to preserve Grails' historical behavior (a plain Spring Boot app
defaults both to `false`). Applications can now opt into stricter startup
semantics:

```yaml
spring:
  main:
    allow-bean-definition-overriding: false
    allow-circular-references: false
```

## Changes

- **grails-core** — `GrailsApp.createApplicationContext()` resolves both flags
  from the configured environment (defaulting to `true`).
- **grails-testing-support-core** — `GrailsApplicationBuilder.createMainContext()`
  resolves the same properties, read after configuration data is loaded and
  before `refresh()`.
- **docs** — new "Bean Definition Overriding and Circular References" section in
  the Application class configuration guide.
- **tests** — `GrailsAppContextOverridingSpec` (grails-core) and
  `GrailsApplicationBuilderContextOverridingSpec` (grails-databinding).

## Notes

- The default is intentionally `true`. Grails' own bootstrap relies on bean
  definition overriding (the harness re-registers infrastructure beans such as
  `internalConfigurationAnnotationProcessor`). The test suite documents that
  disabling overriding makes a Grails context fail fast with
  `BeanDefinitionOverrideException` — which is exactly why the default differs
  from Spring Boot's.
- `GrailsApplicationBuilder` lives in `grails-testing-support-core`, but its full
  bootstrap requires the servlet API, so its behavioral test lives in
  `grails-databinding` (which has the servlet API on its test classpath and
  already exercises the builder).

## Verification

- `./gradlew :grails-core:test :grails-databinding:test
  :grails-testing-support-core:test :grails-fields:test --rerun-tasks` — green.
- `checkstyleMain`/`codenarcMain` + `checkstyleTest`/`codenarcTest` on the
  changed modules — no violations.


sbglasius (approved):

sbglasius (approved):

grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsApplicationBuilder.groovy
```diff
@@ -166,11 +167,15 @@ class GrailsApplicationBuilder {
...
-            allowBeanDefinitionOverriding = true
-            allowCircularReferences = true
-        }
+        def beanFactory = context.beanFactory as DefaultListableBeanFactory
         prepareContext(context, beanFactory)
+        // Bean definition overriding and circular references default to true (the historical Grails
+        // behavior) but can be turned off via the standard spring.main.* properties. These are read
+        // after prepareContext so that application.yml (loaded by ConfigDataApplicationContextInitializer)
+        // and other property sources are available, and before refresh so the values take effect.
+        def environment = context.environment
```
jdaugherty: The environment can be initialized prior to the context.  As part of the context creation, can't beans be created?  This would this make the settings after the fact?

codeconsole: @jdaugherty  Reordered so the environment is initialized and the `spring.main.*` flags are set before any bean definitions are registered (and before `refresh()`) — they now apply to the whole context lifecycle instead of after the fact.

jdaugherty (approved):

PR #15791, opened 2026-06-29
Infer controller namespace for link generation when unambiguous

jamesfredley:
## The Problem

Grails controllers can declare a `namespace`, but URL generation only applied the request namespace when a link targeted the **current** controller. A link to any *other* controller silently dropped the namespace unless it was passed explicitly - so the same `controller`/`action` attributes produced namespaced URLs in some places and non-namespaced URLs in others.

This inconsistency spanned `g:link`, `g:createLink`, `g:form`, `g:paginate`, `g:sortableColumn`, `g:include`, `redirect()`, and `chain()`.

## The Fix

Link generation now infers the namespace from the **target controller**, so links "just work" from `controller` and `action` alone with no `namespace` attribute in the normal case.

When no `namespace` is supplied, a shared `LinkGenerator.getDefaultNamespace(controller, plugin)` resolver decides:

1. **Self-link** (target is the controller currently handling the request) -> the current request namespace.
2. **Exactly one controller has the name** -> that controller's namespace (namespaced or not). This is the normal case - nothing to specify.
3. **Same name in multiple namespaces** (a discouraged design) -> the non-namespaced controller is preferred (matching how dispatch already resolves the collision), then the current namespace; any remaining ambiguity needs an explicit `namespace`.

An explicit `namespace` always wins. To target a non-namespaced controller, use `namespace=""` in GSP markup or `namespace: null` in a Groovy call (a blank value normalizes to `null` so it matches non-namespaced reverse mappings).

## Dispatch alignment

This does **not** change request dispatch (which controller serves a URL); it makes *link generation* produce the URL dispatch already resolves to. Linking `controller="car" action="list"` with **no explicit namespace**:

| Controllers named `car` | `/car/list` dispatches to (unchanged) | Old link gen | New link gen |
|---|---|---|---|
| One, non-namespaced (root) | root `car` | `/car/list` | `/car/list` (unchanged) |
| One, namespaced (`admin`) | `admin` `car` (no-namespace fallback) | `/car/list` - namespace dropped, non-canonical | `/admin/car/list` - canonical |
| Self-link (page served by `admin` `car`) | current (`admin`) `car` | `/admin/car/list` (already worked) | `/admin/car/list` (unchanged) |
| Two: root **and** `admin` | root `car` (non-namespaced wins the null key) | `/car/list` -> root | `/car/list` -> root (matches dispatch); use `namespace="admin"` for the other |
| Two: `admin` **and** `sales` (no root) | non-deterministic (last registered wins) | whichever won the collision | current namespace if the target exists there, else `/car/list`; explicit `namespace` required to choose reliably |

Explicit selection always works: `namespace="admin"` -> `/admin/car/list`; `namespace=""` / `namespace: null` -> `/car/list`.

## Implementation

- New shared resolver on `LinkGenerator`, implemented in `DefaultLinkGenerator`, backed by a lazily-built controller-name -> namespaces index that rebuilds when registered controllers change (array-identity invalidation; reset on reload via `CachingLinkGenerator.clearCache()`).
- Wired into `DefaultLinkGenerator.link()`, `CachingLinkGenerator` cache-key construction (folding the resolved namespace for top-level and nested `url` shapes and the request context for resource links, so links from different request namespaces never collide), controller `chain()`, and the `g:include` / `g:sortableColumn` tags.
- `ForwardUrlMappingInfo` keeps an explicit-namespace-presence flag so `g:include` can tell an omitted namespace from a deliberate blank/null opt-out when `UrlMappingUtils` delegates include URL generation back through `LinkGenerator`.
- Plugin-targeted links are **not** inferred from the application's own controllers.

## Tests

- `LinkGeneratorNamespaceInferenceSpec` - exhaustive resolver + `link()` cases, cache-collision regressions (url-map and resource), plugin guard, blank-namespace normalization.
- `NamespaceInferenceTagLibSpec` - every URL-generating GSP tag.
- `RedirectMethodTests` - `chain()` namespace cases.
- `grails-test-examples/namespaces` Geb functional specs - every tag, a redirect, a chain, the opt-out, and a custom application tag library that builds links internally.

## Docs

`grails-doc` updated (guide + tag/controller references + What's New + 8.0 upgrade notes) describing the automatic resolution and the `namespace=""` / `namespace: null` opt-out.

### DIFF
--- a/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
+++ b/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
@@ -80,8 +80,20 @@ class UrlMappingTagLib implements TagLibrary {
                 id: attrs.id as String,
                 params: attrs.params as Map)
 
-            if (attrs.namespace != null) {
-                mapping.namespace = attrs.namespace as String
+            // Honor an explicit namespace so a blank one (namespace="" or namespace: null) opts out to
+            // the non-namespaced controller; otherwise infer the namespace for the target controller so
+            // g:include stays consistent with link generation.
+            if (attrs.containsKey('namespace')) {
+                String explicitNamespace = attrs.namespace as String
+                if (explicitNamespace?.trim()) {
+                    mapping.namespace = explicitNamespace
+                }
+            }
+            else if (attrs.controller) {
+                String inferredNamespace = linkGenerator.getDefaultNamespace(attrs.controller as String, attrs.plugin as String)
+                if (inferredNamespace != null) {
+                    mapping.namespace = inferredNamespace
+                }
             }
             if (attrs.plugin != null) {
                 mapping.pluginName = attrs.plugin as String
@@ -289,6 +301,9 @@ class UrlMappingTagLib implements TagLibrary {
 
         def property = attrs.remove('property')
         def action = attrs.action ? attrs.remove('action') : (actionName ?: 'list')
+        // Only forward namespace when the tag actually supplied one, so that an omitted namespace
+        // lets the link generator infer the current controller's namespace.
+        boolean hasNamespace = attrs.containsKey('namespace')
         def namespace = attrs.remove('namespace')
 
         def defaultOrder = attrs.remove('defaultOrder')
@@ -352,7 +367,9 @@ class UrlMappingTagLib implements TagLibrary {
         }
 
         linkAttrs.action = action
-        linkAttrs.namespace = namespace
+        if (hasNamespace) {
+            linkAttrs.namespace = namespace
+        }
 
         writer << callLink((Map) linkAttrs) {
             title
--- a/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
+++ b/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
@@ -80,20 +80,9 @@ class UrlMappingTagLib implements TagLibrary {
                 id: attrs.id as String,
                 params: attrs.params as Map)
 
-            // Honor an explicit namespace so a blank one (namespace="" or namespace: null) opts out to
-            // the non-namespaced controller; otherwise infer the namespace for the target controller so
-            // g:include stays consistent with link generation.
-            if (attrs.containsKey('namespace')) {
-                String explicitNamespace = attrs.namespace as String
-                if (explicitNamespace?.trim()) {
-                    mapping.namespace = explicitNamespace
-                }
-            }
-            else if (attrs.controller) {
-                String inferredNamespace = linkGenerator.getDefaultNamespace(attrs.controller as String, attrs.plugin as String)
-                if (inferredNamespace != null) {
-                    mapping.namespace = inferredNamespace
-                }
+            String namespace = linkGenerator.resolveNamespace(attrs.controller as String, attrs.plugin as String, attrs)
+            if (namespace != null) {
+                mapping.namespace = namespace
             }
             if (attrs.plugin != null) {
                 mapping.pluginName = attrs.plugin as String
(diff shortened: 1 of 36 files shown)


grails-web-url-mappings/src/test/groovy/org/grails/web/mapping/LinkGeneratorNamespaceInferenceSpec.groovy
```diff
@@ -0,0 +1,259 @@
...
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+/**
+ * Exhaustive tests for namespace inference performed by
+ * {@link org.grails.web.mapping.DefaultLinkGenerator#getDefaultNamespace(String)} and the way it
+ * feeds into {@link org.grails.web.mapping.DefaultLinkGenerator#link(java.util.Map, String)}.
+ *
+ * <p>The registered controllers establish these logical-name to namespace mappings:</p>
```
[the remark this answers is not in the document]
jamesfredley: Good catch - fixed in a52e874a1b. The @link now references getDefaultNamespace(String, String) to match the method signature with the pluginName parameter.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
```diff
@@ -80,8 +80,20 @@ class UrlMappingTagLib implements TagLibrary {
                 id: attrs.id as String,
                 params: attrs.params as Map)
 
-            if (attrs.namespace != null) {
-                mapping.namespace = attrs.namespace as String
+            // Honor an explicit namespace so a blank one (namespace="" or namespace: null) opts out to
+            // the non-namespaced controller; otherwise infer the namespace for the target controller so
+            // g:include stays consistent with link generation.
+            if (attrs.containsKey('namespace')) {
```
jdaugherty: The existing pattern is to actually fetch the key and only execute the code if it's set. The below controller namespace default code will be skipped if namespace = ''

grails-test-examples/namespaces/grails-app/controllers/UrlMappings.groovy
```diff
@@ -30,6 +30,10 @@ class UrlMappings {
             namespace = "admin"
         }
 
+        "/frontend/$controller/$action?/$id?(.$format)?"{
+            namespace = "frontend"
```
jdaugherty, line 37: namespaces in spring boot can be set by path / header / configured statically.  I'm guessing this support is static only?

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
```diff
@@ -247,12 +257,20 @@ class DefaultLinkGenerator implements LinkGenerator, PluginManagerAware {
                     params.put(ATTRIBUTE_ID, id)
                 }
                 def pluginName = attrs.get(UrlMapping.PLUGIN)?.toString()
-                def namespace = attrs.get(UrlMapping.NAMESPACE)?.toString()
-                if (namespace == null) {
-                    if (controller == requestStateLookupStrategy.controllerName) {
-                        namespace = requestStateLookupStrategy.controllerNamespace
+                // An explicit namespace attribute always wins so that a blank one (namespace="" or
```
jdaugherty: This code looks copied from another area.  Why are we implementing the same logic multiple times?

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/UrlMappingTagLib.groovy
jamesfredley: Addressed in 1a3ea791ba. g:include now tracks whether the namespace key is present before reading it, so omitted namespace can infer normally while namespace: null / blank remains an explicit opt-out through ForwardUrlMappingInfo.namespaceSpecified.

grails-test-examples/namespaces/grails-app/controllers/UrlMappings.groovy
jamesfredley: Addressed in 114e515503. The fixture comment now clarifies this coverage is for static controller namespace metadata only; path/header/request-derived namespace decisions still need an explicit namespace when ambiguous.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
jamesfredley: Addressed in 2a8727c577. The duplicated namespace fallback is now centralized in LinkGenerator.resolveNamespace and reused by DefaultLinkGenerator, ResponseRedirector, and taglib paths.

Issue #10604, opened 2017-04-25
NullPointerException when validating associated domain


bysouleater:
### Steps to Reproduce
Using Grails 3.2.8
1.- Create domain class A
```
@Resource(readOnly = false, formats = ['json', 'xml'], uri = "/a")
class A {
    B b
}
```
2.- Create domain class B
```
class B {
    String field1
    String field2
}
```
3.- Make POST call ignoring 1 property from associated class B on purpose
```
POST http://localhost:8080/a
{
    "b": {
        "field1": "test"
    }
}
```
### Expected Behaviour
```
400 Bad Request
{
     "message": "Property field2 can't be null"
}
```
### Actual Behaviour
```
500 Internal Server Error
2017-04-25 15:42:43.003 ERROR --- [nio-8080-exec-2] o.g.web.errors.GrailsExceptionResolver   : NullPointerException occurred when processing request: [POST] /a
Stacktrace follows:

java.lang.reflect.InvocationTargetException: null
	at org.grails.core.DefaultGrailsControllerClass$ReflectionInvoker.invoke(DefaultGrailsControllerClass.java:211)
	at org.grails.core.DefaultGrailsControllerClass.invoke(DefaultGrailsControllerClass.java:188)
	at org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter.handle(UrlMappingsInfoHandlerAdapter.groovy:90)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:963)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:897)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:970)
	at org.springframework.web.servlet.FrameworkServlet.doPost(FrameworkServlet.java:872)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:846)
	at org.springframework.boot.web.filter.ApplicationContextHeaderFilter.doFilterInternal(ApplicationContextHeaderFilter.java:55)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:77)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:67)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #15793, opened 2026-06-30
Stabilize recurring CI flaky tests in functional/example apps

jamesfredley:
## The Problem

CI runs of `gradle.yml` and `groovy-joint-workflow.yml` had recurring flaky failures. Mining ~100 recent failures surfaced a handful of environment/timing-sensitive tests in the functional/example apps that fail intermittently and erode confidence in the pipeline.

All changes here are confined to functional/example test apps under `grails-test-examples/`. **No framework runtime behavior is altered.**

## The Fix

Root-cause synchronization hardening (via the Page Object convention already used elsewhere) where feasible, plus the project's existing Spock `@Retry` convention as a backstop only where a root-cause fix was not available.

| Test | Symptom | Fix |
|---|---|---|
| Scaffolding `UserControllerSpec` / `UserCommunityControllerSpec` "User list" | `WaitTimeoutException` - navigation lands on the Spring Security login page because the session wasn't authenticated yet | Harden `LoginPage.login()` / `LogoutPage.logout()` to wait for a **definitive** signal (login form gone / present) instead of a transient title change |
| `AsyncPromiseSpec` "async task handles success without error" (app1 + hibernate7/app1) | Intermittent HTTP 500 instead of 200 (async timing); the JDK `HttpClient` doesn't throw on non-2xx, so a transient 500 surfaces as an `assertJson(200, ...)` failure | `@Retry(count = 2, delay = 500, exceptions = [AssertionError])` on the method |
| `GreetingControllerFunctionalSpec` "renders response" (test-phases) | Geb timing - `pageSource.contains('Hello')` evaluated before render finished | Introduce `GreetingPage` and move the render-timing wait into its `at` check (Page Object pattern), then switch the spec to `to(GreetingPage)` |

Note: the `@Retry` backstops originally added to the scaffolding and Greeting specs were later **removed** once the Page Object / definitive-signal fixes addressed the root cause - retries mask real regressions, so they were dropped where a genuine fix exists (verified green under Docker).

## Investigated, intentionally not changed

- **`DirtyCheckingAfterListenerSpec`** stale `@PendingFeatureIf` - already resolved on `8.0.x` by `05368b662d`.
- **`SaveWithInvalidEntitySpec`** - Hibernate 5 spec already carries the root-cause fix from `05368b662d`; the Hibernate 7 mirror is intentionally `@Ignore`d (documented NPE, #10604), so neither can be the CI failure.
- **`JsonViewTemplateResolverSpec`** - no shared mutable state and fresh instances per feature; a fast unit test, so `@Retry` would risk masking real regressions. Left unchanged.

## Verification note

These are `ContainerGebSpec` / functional / integration tests that require Docker and a full app boot and are intermittent by nature, so confirmation is **CI-bound** - a single green local run can't prove a flake is fixed. Changes were kept conservative and lean on the established `@Retry` convention plus Page Object root-cause fixes.


borinquenkid (approved):
Focused LGTM

jdaugherty (changes requested):
We should not be fixing tests with retry.  They burn cpu cycles instead of fixing the underlying issue.

borinquenkid:
> Stabilizes recurring CI flaky tests identified by mining ~100 recent failures of the `gradle.yml` and `groovy-joint-workflow.yml` workflows. Uses the project's existing Spock `@Retry` convention (already used in `GitHubCreateControllerSpec` and `MicronautErsatzRoundtripSpec`) plus root-cause synchronization hardening where feasible.
> 
> All changes are confined to functional/example test apps under `grails-test-examples/`; no framework runtime behavior is altered.

@jdaugherty Fredley did address the deterministic issues from the non-deterministic.

jdaugherty:
> > Stabilizes recurring CI flaky tests identified by mining ~100 recent failures of the `gradle.yml` and `groovy-joint-workflow.yml` workflows. Uses the project's existing Spock `@Retry` convention (already used in `GitHubCreateControllerSpec` and `MicronautErsatzRoundtripSpec`) plus root-cause synchronization hardening where feasible.
> > All changes are confined to functional/example test apps under `grails-test-examples/`; no framework runtime behavior is altered.
> 
> @jdaugherty Fredley did address the deterministic issues from the non-deterministic.

@borinquenkid  I'm ok with the wait's being added, I am not ok with the retry.  We need to address the tests to make them non-flaky by adding waits.  There are still `@Retry` changes in this PR.

jamesfredley:
@jdaugherty even the smartest combination of AI models could not find a fix for these flaky tests, if you have one please commit it.  Or maybe we remove them.

jdaugherty:
We just need to turn on recording and run locally to find the cause.  It's likely a wait or slow performance issue.  @jamesfredley the tests *do* test functionality and every time we've added retries, we've just caused our runs to be longer.  We can open a ticket as future work or mute the tests in the mean time and they won't be a problem on the PRs.

borinquenkid:
> We just need to turn on recording and run locally to find the cause. It's likely a wait or slow performance issue. @jamesfredley the tests _do_ test functionality and every time we've added retries, we've just caused our runs to be longer. We can open a ticket as future work or mute the tests in the mean time and they won't be a problem on the PRs.

lets @Pending and add an ISSUE. thanks @jdaugherty

jdaugherty:
@borinquenkid I disagree with pending since they're not always failing.

borinquenkid:
> @borinquenkid I disagree with pending since they're not always failing.

You are correct 
`@Ignore("Muted due to Geb timing flake. See ISSUE #10604")`

jdaugherty:
> > @borinquenkid I disagree with pending since they're not always failing.
> 
> You are correct `@Ignore("Muted due to Geb timing flake. See ISSUE #10604")`

I get that we want to take the quick solution here and I understand that AI hasn't been able to figure this out, but AI can't figure out a lot of things.  This is one of those issues that's a true technical debt and not good for the project long term.  We need to fix these issues.  If someone wants to dig into these issues and fix them, that's great.  But I am a hard -1 on masking these problems. I've been burned in the past by this on too many projects and it leads to a longer running test run with more problems longer term.

borinquenkid (commented):
We added PageObject to the test @jdaugherty

jamesfredley:
https://github.com/apache/grails-core/actions/runs/28542759715/job/84626412824 best example of the need for these flaky tests to be fixed, since one test failed the snapshot is not published.

jamesfredley:
@jdaugherty I would say getting past these is a high priority.  I have been trying to get a snapshot published with the Groovy 5 changes since yesterday on 8.0.x and these or others keep getting in the way.  There was also the need to extend the time to publish to nexus, but these flaky tests are failing most snapshot publishes.

jdaugherty:
@jamesfredley can you show me where these runs failed?  The only failure I see is with https://develocity.apache.org/s/si4mliaeifl7e and it looks like a valid failure to me.

jdaugherty:
I believe all of the waitFor changes have been merged independently of this except GreetingControllerFunctionalSpec.

grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LoginPage.groovy
```diff
@@ -33,6 +33,9 @@ class LoginPage extends NavigationPage {
     void login(String username = 'test@grails.org', String password = 'letmein') {
         this.username = username
         this.password = password
-        clickAndWaitForNavigation(loginButton)
+        loginButton.click()
+        // Wait for a definitive authenticated signal: the login page must be fully replaced
+        // (title changed AND the login form is gone), not merely a transient title change.
+        waitFor(30) { title != pageTitle && $('input', name: 'username').empty }
```
jdaugherty: We need to remove the hard coded 30, we set these on the test run and increase them as necessary

grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LogoutPage.groovy
```diff
@@ -29,6 +29,9 @@ class LogoutPage extends NavigationPage {
     }
 
     void logout() {
-        clickAndWaitForNavigation(logoutButton)
+        logoutButton.click()
+        // Wait for a definitive logged-out signal: the confirm page is replaced and the login
+        // form is present again, not merely a transient title change.
+        waitFor(30) { title != pageTitle && !$('input', name: 'username').empty }
```
jdaugherty: We need to remove the hard coded 30, we set these on the test run and increase them as necessary

jdaugherty (approved):
let's replace waitFor(30) with waitFor

grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LoginPage.groovy
borinquenkid: Agreed on dropping the hardcoded `30`. One option, in case it's useful: rather than falling back to a single global timeout, make the escalation opt-in and scoped to just this call site, so it doesn't change the default for every other test using this Page Object, but still gives the flaky path more patience without re-running the interaction the way `@Retry` would:

```groovy
void login(String username, String password, List<Number> waitTimeouts = [5]) {
    usernameField = username
    passwordField = password
    loginButton.click()

    waitForEscalating(waitTimeouts) {
        title != pageTitle && $('input', name: 'username').empty
    }
}

private <T> T waitForEscalating(List<Number> timeouts, Closure<T> condition) {
    def remaining = timeouts.iterator()
    while (true) {
        Number t = remaining.next()
        try {
            return waitFor(t, condition)
        } catch (geb.waiting.WaitTimeoutException e) {
            if (!remaining.hasNext()) throw e
        }
    }
}
```

Default callers get plain `waitFor(5)` (matches `grails.geb.timeouts.timeout`'s default). Only the flaky spec would opt in with `login(username, password, [5, 15, 45])`. Same idea would apply to `LogoutPage.groovy`. Just floating this as an option — happy to go with plain `waitFor()` too if that's preferred.

jamesfredley: Done - removed the hard-coded 30 so it now uses the configured grails.geb.timeouts.timeout (set to 30 in gradle/functional-test-config.gradle and increasable per test run). Pushed in 21670d3d86.

grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LogoutPage.groovy
jamesfredley: Done - same fix here, dropped the hard-coded 30 so logout() uses the configured grails.geb.timeouts.timeout. Pushed in 21670d3d86.

PR #15796, opened 2026-06-30
Fix inheritance of an explicit bindable:true constraint on id/version declared in an abstract @DirtyCheck base

### LINKED ISSUE #15795: Explicit bindable: true constraint on id/version declared in an abstract @DirtyCheck base is not inherited by domain subclasses
### Issue description

**Summary**

When an abstract base class (e.g. a `@DirtyCheck` class in src/main/groovy) declares an explicit bindable: true constraint on one of the default-excluded special properties (id, version, dateCreated, lastUpdated), a concrete domain subclass that does not redeclare the constraint fails to inherit it. The property is silently excluded from data binding even though the parent explicitly opted it in.

This is a regression introduced by the fix for #15681 (PR #15699).

**Affected version**

- 7.0.x (current), starting with the merge of PR #15699 (#15681 fix).

**Root cause**

PR #15699 added a filter in DefaultASTDatabindingHelper.getPropertyNamesToIncludeInWhiteList that strips the default-excluded special properties from the inherited parent whitelist for any domain class:

            if (isDomainClass && DOMAIN_CLASS_PROPERTIES_TO_EXCLUDE_BY_DEFAULT.contains(parentPropertyName)) {
                continue;
            }

This correctly prevents GORM-injected id/version from leaking out of a non-domain `@DirtyCheck` base into the subclass whitelist. However, the filter is unconditional: it cannot tell the difference between a special property that landed in the parent whitelist by default (the case #15681 fixed) and one the parent made bindable via an explicit bindable: true constraint (which should be inherited). As a result, the explicit constraint is discarded.

**Steps to reproduce**

1. Define an abstract @DirtyCheck base in src/main/groovy that opts id into binding:

        @DirtyCheck
        abstract class AbstractBindableIdRecord {
            String description
            static constraints = {
                id bindable: true
            }
        }

2. Define a concrete domain subclass that declares no constraint for id:

            class BindableIdRecord extends AbstractBindableIdRecord {
                static constraints = { description nullable: true }
            }

3. Bind a request that includes id:

        def record = new BindableIdRecord()
        bindData(record, params)   // params: id=99&description=...

**Expected behavior**

id is bound (the inherited bindable: true constraint is honored): record.id == 99.

**Actual behavior**

id is not bound: record.id == null. The inherited explicit constraint is ignored.

Notes

- A subclass override (id bindable: false) must continue to win over an inherited bindable: true.
- The original #15681 behavior must be preserved: special properties that only landed in a non-domain parent's whitelist by default (no explicit constraint) must still not be inherited.

jdaugherty:
## Description

  When an abstract base class — for example a `@DirtyCheck` class in `src/main/groovy` — declares an
  explicit `bindable: true` constraint on one of the default-excluded special properties
  (`id`, `version`, `dateCreated`, `lastUpdated`), a concrete domain subclass that does not redeclare
  the constraint failed to inherit it. The property was silently excluded from data binding even though
  the parent explicitly opted it in.

  This is a regression introduced by the fix for #15681 (PR #15699).

  ### Root cause

  PR #15699 added a filter in `DefaultASTDatabindingHelper.getPropertyNamesToIncludeInWhiteList` that
  strips the default-excluded special properties from the *inherited* parent whitelist for any domain
  class. That correctly stops GORM-injected `id`/`version` from leaking out of a non-domain `@DirtyCheck`
  base, but the filter was unconditional: it could not distinguish a special property that landed in the
  parent whitelist *by default* from one the parent made bindable via an *explicit* `bindable: true`
  constraint. As a result the explicit constraint was discarded.

  ### Fix

  The helper now tracks which excluded-by-default special properties were made bindable via an explicit
  constraint, propagates that set down the inheritance hierarchy, and only skips an inherited special
  property when it was **not** explicitly opted in:

  - The #15681 behavior is preserved — special properties that only landed in a non-domain parent's
    whitelist by default are still not inherited.
  - A subclass `bindable: false` override still wins over an inherited `bindable: true`.

  ### Tests

  - **Unit** (`DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec`): an explicit `bindable: true`
    on `id` declared in a `@DirtyCheck` abstract base is inherited by the domain subclass (this case fails
    without the fix), and a subclass `bindable: false` override beats the inherited `bindable: true`.
  - **Functional** (`grails-test-examples-gorm` / `DirtyCheckBindingSpec`): a new `@DirtyCheck` base
    (`AbstractBindableIdRecord`) opting `id` into binding, an inheriting domain class (`BindableIdRecord`),
    a controller action, and an end-to-end HTTP assertion that the inherited `id` is bound while `version`
  remains unbound.

Fixes #15795

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. 

### Code Quality

- [x] I have verified that all existing tests pass by running `./gradlew build --rerun-tasks`. 
- [x] My code follows the project's **code style** guidelines. I have run `./gradlew codeStyle` and resolved any violations.
- [x] This PR does **not** include mass reformatting, style-only changes, or large-scale refactoring.

### Licensing and Attribution

- [x] All contributed code is provided under the Apache License 2.0, and new source files include the appropriate **Apache license header**.
- [x] If generative AI tooling was used in preparing this contribution, I have followed the ASF's policy on generative tooling and have properly attributed its use.

### Documentation

- [x] If this PR introduces breaking changes or changes that require user action during an upgrade, I have updated the **Upgrade Notes**. 

### DIFF
--- a/grails-test-examples/gorm/grails-app/controllers/gorm/DirtyCheckBindingController.groovy
+++ b/grails-test-examples/gorm/grails-app/controllers/gorm/DirtyCheckBindingController.groovy
@@ -20,7 +20,7 @@
 package gorm
 
 /**
- * Controller used by the functional tests for issue 15681 and PR 15699. It binds the incoming request
+ * Controller used by the functional tests for issues 15681 and 15795. It binds the incoming request
  * parameters to a new domain instance and renders the resulting {@code id}, {@code version} and
  * {@code description} so the tests can assert over HTTP which properties were bound.
  */
--- a/grails-test-examples/gorm/grails-app/domain/gorm/BindableIdRecord.groovy
+++ b/grails-test-examples/gorm/grails-app/domain/gorm/BindableIdRecord.groovy
@@ -20,21 +20,14 @@
 package gorm
 
 /**
- * Controller used by the functional tests for issue 15681 and PR 15699. It binds the incoming request
- * parameters to a new domain instance and renders the resulting {@code id}, {@code version} and
- * {@code description} so the tests can assert over HTTP which properties were bound.
+ * Concrete GORM domain class (located in {@code grails-app/domain}) that extends an abstract
+ * {@code @DirtyCheck} base which declares an explicit {@code id bindable: true} constraint. The subclass
+ * declares no constraint of its own, so the bindable constraint must be inherited and {@code id} must be
+ * bound by data binding. See PR 15699.
  */
-class DirtyCheckBindingController {
+class BindableIdRecord extends AbstractBindableIdRecord {
 
-    def bind() {
-        def record = new DirtyCheckedRecord()
-        bindData(record, params)
-        render "id=${record.id}|version=${record.version}|description=${record.description}"
-    }
-
-    def bindInheritedBindableId() {
-        def record = new BindableIdRecord()
-        bindData(record, params)
-        render "id=${record.id}|version=${record.version}|description=${record.description}"
+    static constraints = {
+        description nullable: true
     }
 }
--- a/grails-test-examples/gorm/grails-app/domain/gorm/BindableIdRecord.groovy
+++ b/grails-test-examples/gorm/grails-app/domain/gorm/BindableIdRecord.groovy
@@ -23,7 +23,7 @@ package gorm
  * Concrete GORM domain class (located in {@code grails-app/domain}) that extends an abstract
  * {@code @DirtyCheck} base which declares an explicit {@code id bindable: true} constraint. The subclass
  * declares no constraint of its own, so the bindable constraint must be inherited and {@code id} must be
- * bound by data binding. See PR 15699.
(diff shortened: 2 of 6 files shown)


jamesfredley (approved):
there's no positive multi-level case - a @DirtyCheck grandparent with id bindable:true, a plain abstract class in between, and the domain subclass binding id while still excluding version. Existing tests cover direct-inherited-true, direct-false-override, and multi-level-without-explicit, but not explicit opt-in propagating through an intermediate level. The code handles it correctly; adding the test would just lock that path. Purely optional - not a blocker for merge.

@sbglasius I think you should get final review since the original issue was one you submitted.

jdaugherty:
@jamesfredley I've added the test you suggested (and also updated the issue number in comments)

sbglasius (approved):

PR #15799, opened 2026-06-30
Ensure that the GlobalGrailsClassInjectorTransformation fails in the grails-core build if the wrong directory is passed

jdaugherty:
After trying to build the 8.0.x build, it was not reproducible because the grails.factories file in grails-core was merged with grails-core & grails-gsp projects somehow.  The only way I can see this happening is if there is an incorrect base directory or a missing directory.  This change is meant to be a "catch all" to find these cases.  The goal is to merge this and then retry the 8.0.0-M2 build and see if this helps.

### DIFF
--- a/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
+++ b/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
@@ -187,7 +187,7 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
      * @return {@code true} when the {@code grails.isolated.build} system property is {@code true}.
      */
     static boolean isIsolatedBuild() {
-        return Boolean.parseBoolean(System.getProperty(ISOLATED_BUILD_PROPERTY))
+        System.getProperty(ISOLATED_BUILD_PROPERTY, 'false').toBoolean()
     }
 
     static File resolveCompilationTargetDirectory(SourceUnit source) {
--- a/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
+++ b/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
@@ -191,7 +191,7 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
     }
 
     static File resolveCompilationTargetDirectory(SourceUnit source) {
-        resolveCompilationTargetDirectory(source, isIsolatedBuild())
+        resolveCompilationTargetDirectory(source, isolatedBuild)
     }
 
     static File resolveCompilationTargetDirectory(SourceUnit source, boolean isolatedBuild) {
--- a/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
+++ b/grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
@@ -208,9 +208,9 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
             // grails-plugin.xml can leak into another. In an isolated build, fail loudly instead.
             if (isolatedBuild) {
                 throw new IllegalStateException(
-                        "Unable to resolve the compilation target directory for '${source?.name}' while " +
-                        'GRAILS_ISOLATED_BUILD=true. Refusing to fall back to the shared relative ' +
-                        "'build/classes/main' path, which would leak generated metadata between modules. " +
+                        "Unable to resolve the compilation target directory for '${source?.name}' while the " +
+                        "'${ISOLATED_BUILD_PROPERTY}' system property is set. Refusing to fall back to the shared " +
+                        "relative 'build/classes/main' path, which would leak generated metadata between modules. " +
                         'Ensure the Groovy compiler supplies CompilerConfiguration.targetDirectory.')
             }
             targetDirectory = new File('build/classes/main')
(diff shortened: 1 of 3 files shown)


grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -122,4 +123,104 @@ class FooGrailsPlugin {
...
+                System.clearProperty('grails.isolated.build')
+            }
+
+        where:
+            value   || expected
+            'true'  || true
+            'false' || false
+            'TRUE'  || true
+            'yes'   || false
+    }
```
matrei, line 147:
```suggestion
    @RestoreSystemProperties
    void "isIsolatedBuild reflects the grails.isolated.build system property"() {
        when:
            System.setProperty('grails.isolated.build', value)

        then:
            GlobalGrailsClassInjectorTransformation.isIsolatedBuild() == expected

        where:
            value   || expected
            'true'  || true
            'false' || false
            'TRUE'  || true
            'y'     || true
            'null'  || false
            '1'     || true
            '0'     || false
            '-1'    || false
            ''      || false
    }
```

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -176,14 +176,43 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+     * directory. When set, the transform must never fall back to a shared/guessed location, which can
+     * leak one module's generated metadata into another.
+     */
+    public static final String ISOLATED_BUILD_PROPERTY = 'grails.isolated.build'
+
+    /**
+     * @return {@code true} when the {@code grails.isolated.build} system property is {@code true}.
+     */
+    static boolean isIsolatedBuild() {
+        return Boolean.parseBoolean(System.getProperty(ISOLATED_BUILD_PROPERTY))
```
matrei:
```suggestion
        System.getProperty(ISOLATED_BUILD_PROPERTY, 'false').toBoolean()
```

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -122,4 +123,104 @@ class FooGrailsPlugin {
...
+            resolved == baseDir
+
+        cleanup:
+            System.clearProperty('base.dir')
+            baseDir.deleteDir()
+    }
+
+    void "findSourceDirectory walks up to the project directory when base.dir is not set"() {
+        given: "no base.dir and a standard per-project compile target"
+            System.clearProperty('base.dir')
```
matrei, line 211:
```suggestion
```

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -122,4 +123,104 @@ class FooGrailsPlugin {
...
+            File target = new File(baseDir, 'build/classes/groovy/main')
+
+        when:
+            File resolved = GlobalGrailsClassInjectorTransformation.findSourceDirectory(target)
+
+        then: "the build-tool supplied base.dir wins"
+            resolved == baseDir
+
+        cleanup:
+            System.clearProperty('base.dir')
```
matrei:
```suggestion
```

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -122,4 +123,104 @@ class FooGrailsPlugin {
...
+
+        when: "the target directory cannot be resolved in an isolated build"
+            GlobalGrailsClassInjectorTransformation.resolveCompilationTargetDirectory(source, true)
+
+        then: "the build fails loudly rather than writing to a shared location"
+            IllegalStateException e = thrown()
+            e.message.contains('GRAILS_ISOLATED_BUILD=true')
+    }
+
+    void "findSourceDirectory prefers the per-project base.dir system property when set"() {
```
matrei, line 192:
```suggestion
    @RestoreSystemProperties
    void "findSourceDirectory prefers the per-project base.dir system property when set"() {
```

matrei (approved):
Strange that Gradle would have such a bug.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -176,14 +176,43 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+    }
+
     static File resolveCompilationTargetDirectory(SourceUnit source) {
+        resolveCompilationTargetDirectory(source, isIsolatedBuild())
```
sbglasius:
```suggestion
        resolveCompilationTargetDirectory(source, isolatedBuild)
```
Property access instead

sbglasius (approved):

jamesfredley:
Not surprising to me, unfortunately.  But is is nice we can usually wrap it with a little duct tape and move forward thanks to Groovy.

jamesfredley (approved):

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
jdaugherty: For consistent state I think we should always clear before running the test. I've added a RestoreSystemProperties to the top though.

PR #15800, opened 2026-07-01
Reduce cached codec metaclass registration overhead

### LINKED ISSUE #15374: Grails 8 grails-core: Invoke Dynamic (Indy) Optimization Opportunities
### Issue description

Grails 7 exhibits a **~4x performance regression** compared to Grails 6 in GORM-based operations when compiled with Invoke Dynamic (Indy), compiling with indy off resolves this currently.  

### Root Cause

Grails frameworks frequently modify metaclasses during request processing, GORM operations, and testing - triggering mass invalidation cascades that prevent JIT optimization.

---

### How Groovy 4 Indy Works

1. **Bootstrap**: When a Groovy method is first called, `IndyInterface.bootstrap()` creates a `CacheableCallSite`
2. **Cache Lookup**: `fromCache()` checks if the method handle is already cached for the receiver's class
3. **Fallback**: If not cached, `selectMethod()` resolves the method and caches it
4. **Optimization**: After `INDY_OPTIMIZE_THRESHOLD` (10,000) hits, the call site target is updated to the cached method handle

**Problem**: When Grails modifies ANY metaclass (e.g., adding a method to a domain class), this invalidates ALL cached call sites in the entire application, forcing re-resolution.

### 1. Framework-Level: Reduce Metaclass Modifications

**Goal**: Minimize the frequency of metaclass changes during request processing.

#### 1.1 Lazy/Deferred Metaclass Registration

Instead of modifying metaclasses during request processing, apply all metaclass modifications to application startup:

```groovy
// grails-core/src/main/groovy/grails/boot/GrailsApp.groovy
class GrailsApp extends SpringApplication {
    
    @Override
    protected void afterRefresh(ConfigurableApplicationContext context, ApplicationArguments args) {
        super.afterRefresh(context, args)
        
        // Trigger all lazy metaclass registrations BEFORE request processing begins
        GrailsMetaClassRegistry.instance.finalizeMetaClasses()
    }
}
```

#### 1.2 Batch Metaclass Modifications

When multiple metaclass changes are needed, batch them to trigger only ONE invalidation:

```groovy
// grails-core/src/main/groovy/org/grails/core/metaclass/MetaClassBatcher.groovy
class MetaClassBatcher {
    private static final ThreadLocal<Boolean> batchMode = new ThreadLocal<>()
    private static final ThreadLocal<List<Runnable>> pendingChanges = new ThreadLocal<>()
    
    static void batch(Closure block) {
        batchMode.set(true)
        pendingChanges.set([])
        try {
            block()
            // Apply all changes at once
            pendingChanges.get().each { it.run() }
        } finally {
            batchMode.set(false)
            pendingChanges.remove()
        }
        // Single invalidation for all batched changes
    }
    
    static void scheduleMetaClassChange(Runnable change) {
        if (batchMode.get()) {
            pendingChanges.get().add(change)
        } else {
            change.run()
        }
    }
}
```

### 2. GORM-Level: Optimize Dynamic Finders

**Goal**: Reduce metaclass modifications from GORM dynamic methods.

#### 2.1 Pre-Register Common Dynamic Finders at Startup

```groovy
// grails-datastore-gorm/src/main/groovy/org/grails/datastore/gorm/GormEnhancer.groovy
class GormEnhancer {
    
    void enhanceAll() {
        // Register ALL potential dynamic finders at startup
        domainClasses.each { domainClass ->
            domainClass.persistentProperties.each { property ->
                preRegisterDynamicFinder(domainClass, property)
            }
        }
    }
    
    private void preRegisterDynamicFinder(Class domainClass, PersistentProperty property) {
        def metaClass = domainClass.metaClass
        def propertyName = property.capitilizedName
        
        // Pre-register findBy, findAllBy, countBy, etc.
        ['findBy', 'findAllBy', 'countBy', 'listOrderBy'].each { prefix ->
            def methodName = "${prefix}${propertyName}"
            if (!metaClass.respondsTo(domainClass, methodName)) {
                metaClass."${methodName}" = { Object[] args ->
                    // Delegate to GORM
                }
            }
        }
    }
}
```

#### 2.2 Use Method Missing Cache

Instead of adding methods to metaclass dynamically, cache method resolutions:

```groovy
// Implement per-class method resolution cache that doesn't modify metaclass
class DynamicMethodCache {
    private static final Map<String, Closure> methodCache = new ConcurrentHashMap<>()
    
    static Closure getMethod(Class clazz, String methodName) {
        String key = "${clazz.name}#${methodName}"
        methodCache.computeIfAbsent(key) { 
            resolveDynamicMethod(clazz, methodName) 
        }
    }
}
```

### 3. Code Generation: AST Transformations

**Goal**: Generate static method handles at compile time instead of runtime metaclass modifications.

#### 3.1 @StaticDynamicFinders AST Transform

```groovy
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.SOURCE)
@GroovyASTTransformationClass("org.grails.compiler.StaticDynamicFindersTransformation")
@interface StaticDynamicFinders {
}

// Generates static finder methods at compile time
@StaticDynamicFinders
class Book {
    String title
    String author
    
    // AST generates:
    // static Book findByTitle(String title) { ... }
    // static Book findByAuthor(String author) { ... }
    // static List<Book> findAllByTitleLike(String titlePattern) { ... }
}
```

#### 3.2 @PrecompiledGormMethods

```groovy
// grails-compiler/src/main/groovy/org/grails/compiler/PrecompiledGormMethodsTransformation.groovy
class PrecompiledGormMethodsTransformation extends AbstractASTTransformation {
    
    void visit(ASTNode[] nodes, SourceUnit source) {
        ClassNode classNode = nodes[1]
        
        if (isDomainClass(classNode)) {
            // Generate all GORM methods as static compiled methods
            generateSaveMethod(classNode)
            generateDeleteMethod(classNode)
            generateGetMethod(classNode)
            generateListMethod(classNode)
            generateCountMethod(classNode)
            // etc.
        }
    }
}
```

#### 4.1 @CompileStatic for Hot Paths

Identify and annotate performance-critical code:

```groovy
// Services that handle many requests
@CompileStatic
class BookService {
    
    Book findByIsbn(String isbn) {
        Book.findByIsbn(isbn)  // Compiles to static dispatch
    }
    
    List<Book> search(String query) {
        // Static compilation avoids invokedynamic overhead
    }
}
```

jamesfredley:
## The Problem

Fixes apache/grails-core#15374.

Grails codec registration dynamically adds `encodeAs*` and `decode*` methods onto globally hot metaclasses (`String`, `GStringImpl`, `StringBuffer`, `StringBuilder`, `Object`). On Groovy Indy, repeatedly mutating those `ExpandoMetaClass` types invalidates call sites and creates avoidable startup/runtime overhead. The clearest, fully compatible win is to stop re-adding the **exact same cached codec method** when the same `CodecFactory` is already registered for the same target class.

## The Fix (reshaped after review)

Two layers, deliberately simple:

### 1. Stop duplicate registration at the callers

- `CodecsConfiguration.codecLookup` no longer calls `reInitialize()` manually. `DefaultCodecLookup` already implements `InitializingBean`; Spring's `afterPropertiesSet()` performs the single startup registration pass. This removes a full double-registration at bean creation.
- `GrailsWebUnitTest.mockCodec` either adds the artefact and `reInitialize()`s, or calls `configureCodecMethods()` once - not both.

### 2. Light registry only where still needed (`CodecMetaClassSupport`)

- Caffeine `weakKeys()` cache keyed on **factory identity**.
- Registration keys use target **`Class<?>` + method name** (not class name strings), so reloaded/plugin classloaders with the same simple name do not collide.
- Claim-and-register is serialized with a **per-factory lock** from a second Caffeine `weakKeys()` cache - never `synchronized(emc)` on globally shared metaclasses.
- If a target metaclass was replaced and the method is missing, registration is re-done even when the factory was seen before.
- Non-cached (development/reload) path is unchanged: still re-resolves and reattaches.

## What is deliberately preserved

- Distinct factory instances stay distinct even with the same codec name (last writer wins, as before).
- Aliases still register through the same path.
- No public API, config, dependency, or migration change - apps keep calling `value.encodeAsHTML()`, `value.encodeAsURL()`, `value.decodeSomeCodec()` exactly as before.

## Impact framing

**Problem shape (baseline):** repeated same-factory EMC writes from:

1. `CodecsConfiguration` calling `reInitialize()` and Spring then calling `afterPropertiesSet()` again
2. `mockCodec` calling `configureCodecMethods()` and then `reInitialize()` which calls it again
3. Multiple `DefaultCodecLookup.reInitialize()` passes over the same artefact instances

**What this PR reduces:** redundant same-factory writes via caller dedupe + Class-keyed Caffeine safety net.

**Measurement notes:** An opt-in microbenchmark remains at `:grails-encoder` (`grails.codec.benchmark.enabled=true`). Earlier real-app timing numbers on this PR were outlier-sensitive (mean vs median gap); treat them as directional only. The durable claim is fewer redundant EMC mutations on the startup/reInitialize path, not a single wall-clock number.

## Scope

Codec-registration slice of the broader #15374 / Indy-metaclass performance topic. GORM dynamic methods, taglib dispatch, static compilation, artefact indexing, and metaclass freeze mechanisms are out of scope here.

## Testing

- `:grails-encoder:test` - `CodecMetaClassSupportSpec` (same-factory idempotence, decoder/aliases, concurrent same-factory registration, re-add after metaclass replacement, distinct-factory isolation, factory churn, non-cached reload behavior)
- Public-behavior assertions only (no production counters / private-field reflection)
- Existing `:grails-test-suite-uber` `DefaultGrailsCodecClassTests` still covers double `configureCodecMethods` on the artefact path


grails-encoder/build.gradle
```diff
@@ -62,4 +62,13 @@ dependencies {
 apply {
     from rootProject.layout.projectDirectory.file('gradle/docs-config.gradle')
     from rootProject.layout.projectDirectory.file('gradle/test-config.gradle')
-}
\ No newline at end of file
+}
+
+tasks.register('codecMetaClassBenchmark', JavaExec) {
```
jdaugherty: Other benchmarks have used @Ignore and not been separate JavaExecs

grails-encoder/src/main/groovy/org/grails/encoder/CodecMetaClassSupport.groovy
```diff
@@ -38,6 +42,8 @@ class CodecMetaClassSupport {
     static final Object[] EMPTY_ARGS = []
     static final String ENCODE_AS_PREFIX = 'encodeAs'
     static final String DECODE_PREFIX = 'decode'
+    private static final Set<MetaMethodRegistrationKey> REGISTERED_META_METHODS = Collections.newSetFromMap(new ConcurrentHashMap<MetaMethodRegistrationKey, Boolean>())
```
jdaugherty: These static values are effectively unbounded now.  Why can't an LRU approach be used?

jdaugherty (commented):
I'm worried about the memory impacts this has on large applications.  It's rolling it's own caching mechanism when elsewhere in the code base we've used other solutions like caffeine, etc.

jdaugherty:
From using fabled to review this: 

## Review: PR #15800 — Reduce cached codec metaclass registration overhead

First, confirming one premise: repeated `emc."${name}" << closure` does silently replace
(Groovy's `registerIfClosure` only throws for native Java methods), so the "redundant
writes invalidate call sites" problem is real. The fix's *goal* is sound; the
implementation has several problems.

### Correctness / performance problems

**1. The distinct-factory path is now quadratic — and the PR's own control benchmark
proves it regressed.** `removeStaleMetaMethodRegistrationKeys()`
(`CodecMetaClassSupport.groovy:224-231`) runs a full scan of every registered key under
the global lock on *every single registration*:

```groovy
private static void removeStaleMetaMethodRegistrationKeys() {
    CodecFactoryKey codecFactoryKey = (CodecFactoryKey) STALE_CODEC_FACTORIES.poll()
    while (codecFactoryKey != null) {
        REGISTERED_META_METHODS.remove(codecFactoryKey)
        codecFactoryKey = (CodecFactoryKey) STALE_CODEC_FACTORIES.poll()
    }
    REGISTERED_META_METHODS.keySet().removeIf { CodecFactoryKey key -> key.stale }
}
```

With N distinct factories that's O(N²), which is exactly why the PR's control case went
from ~9.96s to ~14.4s (+45%). The `removeIf` sweep also makes the `ReferenceQueue`
completely redundant — two stale-cleanup mechanisms where the queue alone (O(1)
amortized) would do.

**2. `synchronized (emc)` locks globally shared metaclasses**
(`CodecMetaClassSupport.groovy:178`). In production these are the metaclasses of
`String`, `Object`, `StringBuilder`, etc. Groovy runtime internals and arbitrary user
code can synchronize on those same monitors; taking them and then nesting the global
`REGISTERED_META_METHODS` lock inside creates a contention and lock-ordering hazard on
some of the hottest objects in the JVM. All codec registration also serializes through
the one static map lock.

**3. The registration key uses the class *name*, not the class** (`registrationKey`,
line 235):

```groovy
private static MetaMethodRegistrationKey registrationKey(ExpandoMetaClass emc, String methodName) {
    new MetaMethodRegistrationKey(emc.getTheClass().getName(), methodName)
}
```

Under dev reload / plugin classloaders, a reloaded class with the same name collides
with the old entry, so its registration is skipped and only the `getMetaMethod`
fallback saves it — and that fallback is unreliable (next point).

**4. The fallback check is name-only and cross-factory** (line 219):

```groovy
registeredMetaMethodKeys(codecFactory).add(key) || emc.getMetaMethod(methodName, EMPTY_ARGS) == null
```

After a metaclass is replaced, factory A's re-registration is skipped if *any* factory
has already attached a same-named method to the new EMC — so "distinct factories with
the same codec name stay distinct" (the PR's own claim) breaks across metaclass
replacement, silently changing which encoder wins. Worse, with
`ExpandoMetaClass.enableGlobally()` a freshly created `String` EMC can inherit the
expando method from `Object`'s EMC, making `getMetaMethod` non-null and suppressing the
re-attach entirely.

**5. This is ~150 lines of hand-rolled weak-identity caching** — custom `WeakReference`
key with identity equals/hashCode, a `ReferenceQueue`, manual stale sweeps, hand-written
`equals`/`hashCode` on two key classes, double-checked locking.
`Caffeine.newBuilder().weakKeys()` gives exactly weak *identity* keys, thread-safe, with
none of this — and Caffeine is already used in `grails-web-url-mappings`,
`grails-datastore-core`, `grails-rest-transforms`, etc. All of that machinery exists to
dedupe what the PR itself says is ~110 writes per startup.

**6. Is this even the right layer?** `DefaultCodecLookup.reInitialize()` →
`GrailsCodecClass.configureCodecMethods()` is where the duplicates originate.
`DefaultGrailsCodecClass` already tracks an initialized flag for exactly this "called
more than once" situation (`DefaultGrailsCodecClass.java:324-330`). A per-codec-class
"already registered cached methods" guard there would be a few lines with no global
static registry at all. The PR never identifies *where* the same-factory duplicate calls
come from in a real app, which is the question that determines the simplest fix.

### Evidence problems

**7. The headline numbers aren't reproducible from the branch.** The
`realAppCodecBenchmark` (5 app starts, 31–41% startup improvement) is not in this diff,
and no baseline write count is given — "110 writes on the branch" is meaningless without
knowing the baseline count (if baseline was ~130, the dedup saved ~20 writes and can't
explain a 5-second startup change). A mean of 16.36s vs median of 10.03s across 5
samples means one or two massive outliers dominated the mean; that's noise, not a
measurement. The PR description also still references a
`:grails-encoder:codecMetaClassBenchmark` JavaExec task that was removed in the third
commit.

### Test / process problems

**8. Tests violate the repo's own rules (CLAUDE.md rule 9 — test via public APIs) and
the production class is polluted with test instrumentation.**
`META_METHOD_REGISTRATION_COUNT` is incremented on every registration forever in
production solely so tests can read it; `clearMetaMethodRegistrationState()` and
`getMetaMethodRegistrationKeyCount()` exist only for tests. The spec then goes further
and reflects into the private field (`CodecMetaClassSupportSpec.groovy:211-225`):

```groovy
private static Set registeredMetaMethodKeys() {
    def field = CodecMetaClassSupport.getDeclaredField('REGISTERED_META_METHODS')
    field.accessible = true
    ((Map) field.get(null)).keySet()
}
```

Tests that can only assert via private internals are a signal the design isn't
observable through its public behavior.

**9. Global-state test hygiene.** The spec clears process-wide static registration state
in setup/cleanup, and `DefaultGrailsCodecClassTests` now removes the metaclasses of
`String`, `GStringImpl`, `StringBuffer`, `StringBuilder`, and `Object` in tearDown —
global mutations in a suite the project explicitly warns runs in parallel. TestLens
already flagged a flaky test on this PR's CI run.

**10. Minor.** `grails-encoder/build.gradle` reads `System.properties` at configuration
time for every `Test` task (configuration-cache unfriendly); the benchmark spec carries
a `main()` method and `println` reporting inside a Spock spec; hand-rolled
`equals`/`hashCode` instead of `@EqualsAndHashCode`.

### Bottom line

The problem is real, but I'd push back on the shape of the fix:

- **(a)** Identify where the duplicate same-factory `configureCodecMethods` calls
  actually come from and dedupe at the caller
  (`DefaultGrailsCodecClass`/`DefaultCodecLookup`) if possible.
- **(b)** If a registry is genuinely needed, use Caffeine `weakKeys()` keyed on the
  factory with the target `Class` (not its name) in the value key, and drop the
  `synchronized(emc)` / global-lock / manual sweep machinery.
- **(c)** Remove the production-side counters and make tests observe public behavior.
- **(d)** Either include the real-app benchmark harness in the PR or restate the impact
  with the baseline write count and a defensible measurement.

jamesfredley:
The problem this targets is real - repeated `emc."${name}" << closure` does silently replace and invalidate call sites, and reducing those redundant startup writes is worth doing (it's the codec slice of the Indy/metaclass performance topic, Codebase 2.3 / Google Doc 1.1). But the review is right that the current *implementation* needs reshaping, and I'd rather fix the shape than defend it:

- **The stale-sweep is O(N²).** `removeStaleMetaMethodRegistrationKeys()` does a full `removeIf` scan of every registered key under the global lock on every registration, which is why the control benchmark regressed ~+45%. The `ReferenceQueue` alone is O(1) amortized and makes the sweep redundant.
- **`synchronized (emc)` locks globally-shared metaclasses** (`String`, `Object`, `StringBuilder`...), which Groovy internals and user code can also synchronize on - a real contention / lock-ordering hazard, made worse by nesting the global map lock inside it.
- **Name-based registration key** collides under dev-reload / plugin classloaders (a reloaded same-named class overwrites the old entry), and the name-only cross-factory fallback can silently change which encoder wins.
- **~150 lines of hand-rolled weak-identity caching** duplicates what `Caffeine.newBuilder().weakKeys()` gives for free - and Caffeine is already used in `grails-web-url-mappings` / `grails-datastore-core`.

So I agree with the direction you outlined:

1. Find where the duplicate *same-factory* `configureCodecMethods` calls actually originate (`DefaultCodecLookup.reInitialize()` -> `GrailsCodecClass.configureCodecMethods()`) and dedupe at the caller - `DefaultGrailsCodecClass` already tracks an initialized flag for exactly this, so a per-codec-class guard there may remove the need for a global registry entirely.
2. If a registry is still needed, use Caffeine `weakKeys()` keyed on the factory + target **`Class`** (not its name), and drop the `synchronized(emc)` / global-lock / manual-sweep machinery.
3. Move the test instrumentation (`META_METHOD_REGISTRATION_COUNT`, the reflect-into-private-field spec) out of the production class and assert via public behavior (repo test-via-public-API rule).
4. Either include the real-app benchmark harness in the PR or restate the impact with the baseline write count and a defensible measurement (a mean of 16.36s vs a 10.03s median is outlier-dominated).

I'll rework it this way. Thanks for the thorough review - this one's a genuinely better design after the feedback.

jdaugherty:
@jamesfredley  let me know once all of these changes are pushed and I'll review again.

jamesfredley:
## Reshape after review feedback

Pushed two commits on top of the previous approach:

1. **`Avoid double codec registration at startup and in mockCodec`**
   - `CodecsConfiguration`: drop manual `reInitialize()`; rely on `InitializingBean.afterPropertiesSet()`
   - `GrailsWebUnitTest.mockCodec`: either `reInitialize()` after `addArtefact`, or `configureCodecMethods()` once - not both

2. **`Key codec metaclass registration by Class and lock per factory`**
   - Registration keys: target `Class<?>` + method name (not class name strings)
   - Dropped `synchronized(emc)` on shared metaclasses
   - Per-factory claim+register lock via Caffeine `weakKeys()` (keeps same-factory concurrency atomic without locking `String`/`Object` EMCs)
   - Kept Caffeine factory-identity cache; no O(N²) stale sweep / hand-rolled weak refs
   - Tests assert via public encode/decode behavior only

### Intentionally not done

- No outer `metaMethodsConfigured` flag on `DefaultGrailsCodecClass` - that would skip `CodecMetaClassSupport`'s metaclass-replacement recovery path
- No real-app benchmark harness in this PR; impact restated as reduced redundant EMC writes + opt-in microbenchmark
- No production test instrumentation counters

### Verification

```
./gradlew :grails-encoder:test --tests org.grails.encoder.CodecMetaClassSupportSpec
```

BUILD SUCCESSFUL (7/7).

Ready for another review pass.

grails-encoder/src/main/groovy/org/grails/encoder/CodecMetaClassSupport.groovy
```diff
@@ -38,6 +43,16 @@ class CodecMetaClassSupport {
     static final Object[] EMPTY_ARGS = []
     static final String ENCODE_AS_PREFIX = 'encodeAs'
     static final String DECODE_PREFIX = 'decode'
+    private static final Cache<CodecFactory, Set<MetaMethodRegistrationKey>> REGISTERED_META_METHODS = Caffeine.newBuilder()
```
jdaugherty, line 47: This closes the unbounded-static concern from my earlier pass. `weakKeys()` means an entry can only outlive a factory that is already retained elsewhere, and each factory's value set is bounded by target classes x method names - roughly 30 small objects for the default five targets. No LRU needed at that size.

grails-encoder/src/main/groovy/org/grails/encoder/CodecMetaClassSupport.groovy
```diff
@@ -38,6 +43,16 @@ class CodecMetaClassSupport {
...
+            .weakKeys()
+            .build()
+    /**
+     * Per-factory locks keep claim+register atomic for the same CodecFactory without
+     * synchronizing on globally shared ExpandoMetaClass instances (String, Object, ...).
+     */
+    private static final Cache<CodecFactory, Object> FACTORY_REGISTRATION_LOCKS = Caffeine.newBuilder()
```
jdaugherty: This second cache is redundant. `registeredMetaMethodKeys(codecFactory)` already returns a stable per-factory instance, so it can serve as the monitor directly:

```groovy
synchronized (registeredMetaMethodKeys(codecFactory)) {
```

Same locking granularity, one fewer static cache, and `factoryRegistrationLock` plus its weak-key bookkeeping goes away.

grails-encoder/src/main/groovy/org/grails/encoder/CodecMetaClassSupport.groovy
```diff
@@ -152,8 +168,79 @@ class CodecMetaClassSupport {
     }
 
     protected void addMetaMethod(List<ExpandoMetaClass> targetMetaClasses, String methodName, Closure closure) {
+        addMetaMethod(targetMetaClasses, methodName, closure, false, null)
+    }
+
+    protected void addMetaMethod(List<ExpandoMetaClass> targetMetaClasses, String methodName, Closure closure, boolean cacheLookup, CodecFactory codecFactory) {
```
jdaugherty: `addMetaMethod(List, String, Closure)` is `protected` on a public class, so it is an extension point. Cached registration now routes through this five-arg overload and never calls the three-arg one, which means an external subclass overriding it silently stops intercepting.

Nothing in this repo overrides it, so this is not urgent. Keeping the three-arg method as the single hook and making the claim/skip decision in `configureCodecMethods` before calling it would avoid changing the protected surface at all.

grails-encoder/src/main/groovy/org/grails/encoder/CodecMetaClassSupport.groovy
```diff
@@ -152,8 +168,79 @@ class CodecMetaClassSupport {
...
+                return false
+            }
+
+            MetaMethodRegistrationKey otherKey = (MetaMethodRegistrationKey) other
+            targetClass == otherKey.targetClass &&
+                    methodName == otherKey.methodName
+        }
+
+        @Override
+        int hashCode() {
```
jdaugherty: `@EqualsAndHashCode` on the class replaces both of these. As written the `hashCode()` body reads like an unfinished `Objects.hash` - the intermediate `result` assignment does not earn its keep with the multiplier on the next line.

grails-encoder/build.gradle
```diff
@@ -63,4 +64,11 @@ dependencies {
 apply {
     from rootProject.layout.projectDirectory.file('gradle/docs-config.gradle')
     from rootProject.layout.projectDirectory.file('gradle/test-config.gradle')
-}
\ No newline at end of file
+}
+
+tasks.withType(Test).configureEach {
+    systemProperties System.properties.findAll { key, value -> key.toString().startsWith('grails.codec.benchmark.') }
```
jdaugherty: This does not follow the property-bridging pattern already established in `gradle/test-config.gradle`, which this module applies:

```groovy
project.properties.each { key, value ->
    if (key.toString().startsWith('grails.geb.')) {
        systemProperty key, value
    }
}
```

Two consequences of reading `System.properties` and using the bulk form instead: only `-D` works, not `-P` or `local.properties` like every other switch in this build; and iterating the full property set at configuration time becomes a configuration-cache input the moment `org.gradle.configuration-cache` stops being `false`. Please match the existing `grails.geb.` shape.

grails-encoder/src/test/groovy/org/grails/encoder/CodecMetaClassBenchmarkSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+import groovy.transform.CompileStatic
+import groovy.transform.TypeCheckingMode
+import org.codehaus.groovy.runtime.InvokerHelper
+
+import grails.util.GrailsMetaClassUtils
+import spock.lang.IgnoreIf
+import spock.lang.Specification
+
+@IgnoreIf({ !Boolean.getBoolean('grails.codec.benchmark.enabled') })
+class CodecMetaClassBenchmarkSpec extends Specification {
```
jdaugherty: The separate JavaExec is gone, which was the ask. The open question is placement: #16071 adds a `grails-benchmarks` module with real JMH harnesses plus per-PR regression reporting, also targeting 8.0.x. A hand-rolled `System.nanoTime()` loop here becomes a third benchmark pattern in the tree, and the description already concedes the wall-clock numbers were outlier-sensitive and directional only.

I would rather this spec and the `build.gradle` block come out of this PR, and the measurement land as a JMH benchmark on top of #16071 where the numbers get compared automatically instead of by hand. That also makes the `build.gradle` comment moot.

grails-codecs/src/main/groovy/org/grails/plugins/codecs/CodecsConfiguration.java
```diff
@@ -36,9 +36,7 @@ public class CodecsConfiguration {
 
     @Bean("codecLookup")
     @Primary
-    public CodecLookup codecLookup(GrailsApplication grailsApplication) throws Exception {
-        final DefaultCodecLookup defaultCodecLookup = new DefaultCodecLookup(grailsApplication);
-        defaultCodecLookup.reInitialize();
-        return defaultCodecLookup;
+    public CodecLookup codecLookup(GrailsApplication grailsApplication) {
```
jdaugherty, line 39: Correct - `BasicCodecLookup.afterPropertiesSet()` already calls `reInitialize()`, and this is the only `codecLookup` bean definition, so the second pass was pure waste.

Registration now rides entirely on the `InitializingBean` contract, and there is no test anywhere touching this class. Please add one asserting the `codecLookup` bean has codecs registered after a normal context refresh, so this cannot silently regress if the bean is ever constructed somewhere that skips the lifecycle.

grails-testing-support-web/src/main/groovy/grails/testing/web/GrailsWebUnitTest.groovy
```diff
@@ -163,10 +163,11 @@ trait GrailsWebUnitTest implements GrailsUnitTest {
         }
         loadedCodecs << codecClass
         DefaultGrailsCodecClass grailsCodecClass = new DefaultGrailsCodecClass(codecClass)
-        grailsCodecClass.configureCodecMethods()
         grailsApplication.addArtefact(CodecArtefactHandler.TYPE, grailsCodecClass)
         if (reinitialize) {
             applicationContext.getBean(DefaultCodecLookup).reInitialize()
+        } else {
+            grailsCodecClass.configureCodecMethods()
```
jdaugherty, line 170: This branch is now the only direct `configureCodecMethods()` caller left in the tree, and `mockCodec` has exactly one caller outside this trait (`CodecSpec`, which takes the `reinitialize = true` default). So the `false` path - the one `WebSetupSpecInterceptor` drives for every web spec - has no coverage at all.

Please add a spec asserting `mockCodec(SomeCodec, false)` still installs the `encodeAs*` methods.

jdaugherty (approved):
Approving.

I traced the lifecycle rather than taking the description at face value, and the mechanism holds up:

- `BasicCodecLookup implements InitializingBean` with `afterPropertiesSet() -> reInitialize()`, and this `@Bean` is the only `codecLookup` definition in the tree, so dropping the manual `reInitialize()` loses nothing. The legacy GSP harness registers its own `codecLookup` *definition* (`AbstractGrailsTagTests`), so it still gets the Spring lifecycle too.
- `cacheLookup` is driven off `Environment.isDevelopmentMode()`, so the reload path is genuinely untouched.
- The `getMetaMethod(...) == null` fallback is load-bearing, and the spec proves it: drop the fallback and `re-adds methods after metaclass replacement` fails. The tests pin the mechanism rather than a usage pattern, which is what I want to see here.
- The cached closures capture the resolved encoder eagerly, but `DefaultGrailsCodecClass.initializeCodec()` is one-shot guarded, so there is no stale-encoder path through the real factory implementation.
- Lock ordering is consistently factory-lock -> EMC and never touches the globally shared metaclasses, so there is no deadlock to worry about.

One behaviour change worth calling out in the description: when two codecs register the same method name or alias, a second registration pass over a *replaced* metaclass now lets whichever factory arrives first win, where previously each pass re-asserted last-writer-wins. It takes a name collision plus a metaclass reset to observe, so I am not treating it as a blocker, but it is the one semantic edge the fallback introduces.

Comments below. I would like the two test gaps closed before this lands - `CodecsConfiguration` and the `mockCodec` `false` branch are both touched with nothing covering them - and a decision on where the benchmark lives now that #16071 is in flight. Approving so neither of those needs another review round trip.

Issue #15454, opened 2026-02-25
Create a "Grails 8 Upgrade" skill file that we offer end users


jdaugherty:
### Issue description

Create a "Grails 8 Upgrade" skill file that we offer end users

jamesfredley:
Exists on https://github.com/apache/grails-core/tree/8.0.x/.agents/skills/grails-8-upgrade

PR #15805, opened 2026-07-01
Publish Grails skills through SkillsJars

jamesfredley:
## The Problem

[SkillsJars](https://skillsjars.com) discovers and publishes agent skills from a root-level `skills/**/SKILL.md` layout. Grails keeps its canonical skill sources under `.agents/skills`, which SkillsJars does not scan by default - and that directory also contains **framework-contributor-only** skills that should not be published to application developers.

## The Fix

Expose only the app-facing Grails skills through the SkillsJars discovery path, without duplicating any content, while keeping contributor skills internal.

**Published** via root-level `skills/` (real directories whose `SKILL.md` is a symlink to the canonical source):

- `skills/grails-developer/SKILL.md` -> `.agents/skills/grails-developer/SKILL.md`
- `skills/grails-8-upgrade/SKILL.md` -> `.agents/skills/grails-8-upgrade/SKILL.md`

**Kept internal** under `.agents/skills` only: `hibernate-developer`, `test-fixer`, `violation-fixer`, and other core-contributor workflow skills.

The change also:

- uses real skill directories with `SKILL.md` symlinks so SkillsJars discovers normal skill folders;
- adds a **`verifySkillsJarsSources`** Gradle task asserting the exact published skill set, symlink targets, git mode `120000`, and canonical source markers;
- excludes only the symlink alias paths from RAT (the licensed sources remain under `.agents/skills`);
- documents the canonical location, the public discovery aliases, and the publish workflow.

## How to publish after merge

Submit the SkillsJars form with GitHub Org `apache` and Repo `grails-core`. SkillsJars shallow-clones the public repo, scans `skills/**/SKILL.md`, and publishes Maven Central artifacts under `com.skillsjars`:

- `com.skillsjars:apache__grails-core__grails-developer:<date>-<commit>`
- `com.skillsjars:apache__grails-core__grails-8-upgrade:<date>-<commit>`

## Testing

- `./gradlew verifySkillsJarsSources` passes; `git diff --cached --check` clean.
- `./gradlew rat` no longer reports the `skills/**` or tracked `.claude/skills/**` symlink aliases.

Refs #15454

### DIFF
--- a/.agents/skills/README.md
+++ b/.agents/skills/README.md
@@ -1,0 +1,26 @@
+<!--
+SPDX-License-Identifier: Apache-2.0
+
+Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements; and to You under the Apache License, Version 2.0.
+-->
+
+# Grails Agent Skills
+
+The canonical Grails agent skills live in this directory.
+
+The repository root `skills` path is intentionally a symbolic link to this directory so [SkillsJars](https://www.skillsjars.com/) can discover the skills without maintaining duplicate files. SkillsJars deploys from public GitHub repositories by scanning `skills/**/SKILL.md` and then publishes one Maven Central artifact per skill under `com.skillsjars`.
+
+Run this check before publishing or updating a skill:
+
+```bash
+./gradlew verifySkillsJarsSources
+```
+
+After changes are merged to the public branch, publish them from SkillsJars by submitting:
+
+| Field | Value |
+|-------|-------|
+| GitHub Org | `apache` |
+| GitHub Repo | `grails-core` |
+
+Expected coordinates use the `apache__grails-core__<skill-name>` artifact pattern, for example `com.skillsjars:apache__grails-core__grails-developer:<date>-<commit>`.
--- a/.agents/skills/README.md
+++ b/.agents/skills/README.md
@@ -6,9 +6,16 @@ Licensed to the Apache Software Foundation (ASF) under one or more contributor l
 
 # Grails Agent Skills
 
-The canonical Grails agent skills live in this directory.
+The canonical Grails agent skills live in this directory. Skills that are only useful while working in the Grails framework repository stay here and are not published through SkillsJars.
 
-The repository root `skills` path is intentionally a symbolic link to this directory so [SkillsJars](https://www.skillsjars.com/) can discover the skills without maintaining duplicate files. SkillsJars deploys from public GitHub repositories by scanning `skills/**/SKILL.md` and then publishes one Maven Central artifact per skill under `com.skillsjars`.
+The repository root `skills/` directory contains normal skill directories for app-facing skills that are useful when building or upgrading end-user Grails applications. Each published directory contains only a `SKILL.md` symbolic link back to the canonical source in `.agents/skills`, so [SkillsJars](https://www.skillsjars.com/) can discover and package the skills without duplicating content. SkillsJars deploys from public GitHub repositories by scanning `skills/**/SKILL.md` and then publishes one Maven Central artifact per discovered skill under `com.skillsjars`.
+
+| Published skill | Source skill | Use |
+|-----------------|--------------|-----|
+| `grails-developer` | `.agents/skills/grails-developer` | Building current Grails web applications, REST APIs, GORM models, controllers, services, views, plugins, and tests |
+| `grails-8-upgrade` | `.agents/skills/grails-8-upgrade` | Upgrading Grails applications from Grails 7.x to Grails 8 |
+
+The repository-specific `hibernate-developer`, `test-fixer`, and `violation-fixer` skills are intentionally not linked from `skills/` because they are for Grails framework development. Contributors already receive them from this repository when working on Grails core.
 
 Run this check before publishing or updating a skill:
 
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,48 @@ tasks.register('publishAllToMavenLocal') {
     }
 }
 
+tasks.register('verifySkillsJarsSources') {
+    description = 'Verifies the SkillsJars publication path points at the agent-local skill sources.'
+    group = 'verification'
+
+    def agentSkillsDir = layout.projectDirectory.dir('.agents/skills')
+    def skillsJarsLink = layout.projectDirectory.file('skills')
+
+    inputs.dir(agentSkillsDir)
+    inputs.files(skillsJarsLink)
+
+    doLast {
+        def expectedTarget = '.agents/skills'
+        def skillsPath = skillsJarsLink.asFile.toPath()
+        def actualTarget = null
+
+        if (java.nio.file.Files.isSymbolicLink(skillsPath)) {
+            actualTarget = java.nio.file.Files.readSymbolicLink(skillsPath).toString().replace('\\', '/')
+        }
+        else if (skillsJarsLink.asFile.isFile()) {
+            actualTarget = skillsJarsLink.asFile.getText('UTF-8').trim().replace('\\', '/')
+        }
+
+        if (actualTarget != expectedTarget) {
+            throw new GradleException("The SkillsJars publication path must be a symlink to ${expectedTarget}, but found ${actualTarget ?: 'no link target'}.")
+        }
+
+        def gitMode = providers.exec {
+            commandLine 'git', 'ls-files', '-s', 'skills'
+        }.standardOutput.asText.get().trim().split(' ')[0]
+        if (gitMode != '120000') {
+            throw new GradleException("The SkillsJars publication path must be committed as a symlink with git mode 120000, but found mode ${gitMode ?: 'none'}.")
+        }
+
+        def skillMarkers = fileTree(agentSkillsDir).matching {
+            include '**/SKILL.md'
+        }.files
+        if (skillMarkers.empty) {
+            throw new GradleException("No SkillsJars skill marker files found under ${expectedTarget}.")
+        }
+    }
+}
+
 apply {
     from layout.projectDirectory.file('gradle/publish-root-config.gradle')
     from layout.projectDirectory.file('gradle/rat-root-config.gradle')
--- a/build.gradle
+++ b/build.gradle
@@ -120,43 +120,63 @@ tasks.register('publishAllToMavenLocal') {
 }
 
 tasks.register('verifySkillsJarsSources') {
-    description = 'Verifies the SkillsJars publication path points at the agent-local skill sources.'
+    description = 'Verifies the SkillsJars publication paths point at the app-facing Grails skill sources.'
     group = 'verification'
 
-    def agentSkillsDir = layout.projectDirectory.dir('.agents/skills')
-    def skillsJarsLink = layout.projectDirectory.file('skills')
+    def skillsJarsDir = layout.projectDirectory.dir('skills')
+    def publishedSkills = [
+            'grails-8-upgrade': '../../.agents/skills/grails-8-upgrade/SKILL.md',
+            'grails-developer': '../../.agents/skills/grails-developer/SKILL.md',
+    ]
 
-    inputs.dir(agentSkillsDir)
-    inputs.files(skillsJarsLink)
+    inputs.dir(layout.projectDirectory.dir('.agents/skills'))
+    inputs.dir(skillsJarsDir)
 
     doLast {
-        def expectedTarget = '.agents/skills'
-        def skillsPath = skillsJarsLink.asFile.toPath()
-        def actualTarget = null
-
-        if (java.nio.file.Files.isSymbolicLink(skillsPath)) {
-            actualTarget = java.nio.file.Files.readSymbolicLink(skillsPath).toString().replace('\\', '/')
-        }
-        else if (skillsJarsLink.asFile.isFile()) {
-            actualTarget = skillsJarsLink.asFile.getText('UTF-8').trim().replace('\\', '/')
+        if (!skillsJarsDir.asFile.directory) {
+            throw new GradleException('The SkillsJars publication path must be a skills directory containing app-facing skill symlinks.')
         }
 
-        if (actualTarget != expectedTarget) {
-            throw new GradleException("The SkillsJars publication path must be a symlink to ${expectedTarget}, but found ${actualTarget ?: 'no link target'}.")
+        def actualSkillNames = skillsJarsDir.asFile.listFiles().findAll { File file -> file.directory }.collect { File file -> file.name } as Set
+        def expectedSkillNames = publishedSkills.keySet() as Set
+        def missingSkillNames = expectedSkillNames - actualSkillNames
+        def extraSkillNames = actualSkillNames - expectedSkillNames
+        if (missingSkillNames || extraSkillNames) {
+            throw new GradleException("SkillsJars must publish exactly ${expectedSkillNames}, but missing ${missingSkillNames} and found extra ${extraSkillNames}.")
         }
 
-        def gitMode = providers.exec {
-            commandLine 'git', 'ls-files', '-s', 'skills'
-        }.standardOutput.asText.get().trim().split(' ')[0]
-        if (gitMode != '120000') {
-            throw new GradleException("The SkillsJars publication path must be committed as a symlink with git mode 120000, but found mode ${gitMode ?: 'none'}.")
-        }
+        publishedSkills.each { String skillName, String expectedSkillTarget ->
+            def skillDir = skillsJarsDir.dir(skillName).asFile
+            if (!skillDir.directory) {
+                throw new GradleException("The SkillsJars skill ${skillName} must be a directory containing SKILL.md.")
+            }
 
-        def skillMarkers = fileTree(agentSkillsDir).matching {
-            include '**/SKILL.md'
-        }.files
-        if (skillMarkers.empty) {
-            throw new GradleException("No SkillsJars skill marker files found under ${expectedTarget}.")
+            def skillLink = skillsJarsDir.file("${skillName}/SKILL.md").asFile
+            def skillPath = skillLink.toPath()
+            def actualTarget = null
+
+            if (java.nio.file.Files.isSymbolicLink(skillPath)) {
+                actualTarget = java.nio.file.Files.readSymbolicLink(skillPath).toString().replace('\\', '/')
+            }
+            else if (skillLink.isFile()) {
+                actualTarget = skillLink.getText('UTF-8').trim().replace('\\', '/')
+            }
+
+            if (actualTarget != expectedSkillTarget) {
+                throw new GradleException("The SkillsJars skill ${skillName} marker must be a symlink to ${expectedSkillTarget}, but found ${actualTarget ?: 'no link target'}.")
+            }
+
+            def gitMode = providers.exec {
+                commandLine 'git', 'ls-files', '-s', "skills/${skillName}/SKILL.md"
+            }.standardOutput.asText.get().trim().split(' ')[0]
+            if (gitMode != '120000') {
+                throw new GradleException("The SkillsJars skill ${skillName} marker must be committed as a symlink with git mode 120000, but found mode ${gitMode ?: 'none'}.")
+            }
+
+            def sourceSkillFile = layout.projectDirectory.file(expectedSkillTarget.substring(6))
+            if (!sourceSkillFile.asFile.file) {
+                throw new GradleException("The SkillsJars skill ${skillName} target ${expectedSkillTarget} does not exist.")
+            }
         }
     }
 }
--- a/gradle/rat-root-config.gradle
+++ b/gradle/rat-root-config.gradle
@@ -26,6 +26,7 @@ tasks.named('rat') {
             'CHECKSUMS', // build artifact for storing checksums for easy verification
             'PUBLISHED_ARTIFACTS', // build artifact for storing published artifacts coordinates & paths
             'licenses/**', // licenses directory excluded
+            'skills', // SkillsJars discovery symlink; licensed skill sources live under .agents/skills
             '.git-blame-ignore-revs', // git configuration isn't code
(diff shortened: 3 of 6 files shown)


jamesfredley:
Updated this PR to publish only the app-facing Grails skills:

- `skills/grails-developer/SKILL.md` -> `.agents/skills/grails-developer/SKILL.md`
- `skills/grails-8-upgrade/SKILL.md` -> `.agents/skills/grails-8-upgrade/SKILL.md`

The internal Grails core skills remain under `.agents/skills` but are no longer exposed to SkillsJars. I also changed the shape from symlinked skill directories to real skill directories with `SKILL.md` symlinks so SkillsJars can discover/package them more reliably.

jamesfredley:
Maintenance pass complete. Copilot generated no inline review comments, there are no review threads to resolve, Codecov/TestLens are green on the existing branch, and I reran ./gradlew.bat --no-daemon --no-parallel --max-workers=1 verifySkillsJarsSources locally. No code changes were needed.

jdaugherty:
I dug into this more and I don't believe we can use SkillsJars - it's acting a publisher directly to maven central.  This goes against the ASF voting rules.  Has there been any discussion on the mailing lists about this problem?

jamesfredley:
Follow-up on the ASF voting concern: this branch is still SkillsJars-based end to end. The changed files add verifySkillsJarsSources in build.gradle, expose skills/grails-developer and skills/grails-8-upgrade solely for SkillsJars discovery, and document com.skillsjars Maven Central coordinates in .agents/skills/README.md; the final search did not find a separate in-repo release path for these skill artifacts. Given that, we should not try to land this as-is. Proposed path: pause or close this PR, take the publish-to-Maven-Central and voting question to the dev list or Infra, and replace it with an ASF-controlled publication mechanism only if the project agrees on release and vote semantics. Until then, keeping the skills as source docs under .agents/skills avoids publishing artifacts outside the normal Apache release process.

jamesfredley:
I dug into this, and the factual premise is correct for the PR **as written** - but there's a compliant path that still achieves the goal, so I don't think we have to abandon skill distribution.

**Why the objection holds for the current approach:** the SkillsJars web service clones the repo, generates `com.skillsjars:apache__grails-core__...` coordinates, signs/bundles them, and uploads them to Maven Central *itself*. That's third-party publication of unvoted, commit-derived artifacts, outside the ASF Nexus / PMC release path. ASF release policy requires artifacts to be staged in ASF Nexus, approved by a PMC vote (>=3 binding +1, >72h), and promoted to Central through ASF infra ([release policy](https://www.apache.org/legal/release-policy.html#release-approval), [publishing guide](https://infra.apache.org/publishing-maven-artifacts.html)). So routing through the SkillsJars publisher does conflict with the voting rules.

**The compliant alternative** (and I think what we actually want): build the skills jar in the `grails-core` build and publish it as a normal `org.apache.grails:*` artifact **as part of the standard Apache Grails release candidate** - staged, signed, and covered by the same PMC vote as every other `grails-*` jar. ASF policy [expressly permits](https://www.apache.org/legal/release-policy.html#compiled-packages) shipping binary jars alongside the voted source release when they're the same version and built from that release's source. That gives end users a Maven-Central-published skills artifact (closing the #15454 distribution blocker, on top of the already-merged skill content in #15798) with none of the direct-publish problem.

So the fix is to swap the *distribution mechanism*: package + publish the skills through the ASF release pipeline (voted) rather than via the SkillsJars external publisher. I'll rework the PR along those lines. No prior dev@ thread found on this - I'll start one so the packaging/coordinates get consensus before we wire it into the release.

jdaugherty:
I actually havea  side branch that already publishes these as skill jars, i'm happy to push it too - but how do you consume it?

jdaugherty:
FYI: I pushed my rework branch (this was a start, i didn't finish it once I saw that skilljars couldn't pull other jars): https://github.com/apache/grails-core/tree/rework/publish-grails-skills

borinquenkid (commented):
## Review

**Overview:** Exposes two app-facing skills (`grails-developer`, `grails-8-upgrade`) to the third-party SkillsJars discovery service via a new root-level `skills/` directory, where each skill folder contains only a `SKILL.md` symlink back to the canonical `.agents/skills/...` source. Adds a `verifySkillsJarsSources` Gradle task to assert the published set and symlink integrity, and excludes the alias paths from RAT.

### Code quality / correctness
- Symlink targets (`../../.agents/skills/<name>/SKILL.md`) resolve correctly from `skills/<name>/SKILL.md`.
- The verification task's dual-mode target read (`Files.isSymbolicLink` vs. plain-file fallback) sensibly handles checkouts where symlinks degrade to text files.
- Minor fragility: `expectedSkillTarget.substring(6)` hardcodes the `'../../'` prefix length to locate the source file. Works today, but breaks silently (wrong path) if a target's depth ever changes. Resolving the path properly (e.g. `Paths.get(...).resolve(target).normalize()`) would be more robust than the magic-number substring.

### Verification task isn't wired to `check`/CI
`verifySkillsJarsSources` is registered but not added as a dependency of `check`, `build`, or any GitHub Actions workflow — it's purely a manual, docs-invoked gate. If `.agents/skills/grails-developer/SKILL.md` is ever moved or renamed, the symlink breaks silently and CI stays green; nobody finds out until someone remembers to run the task by hand (or SkillsJars' scan quietly fails). Wiring this into `check` (or at least the RAT/CodeAnalysis job) would close that gap cheaply.

### Test coverage
No automated test (e.g. Gradle TestKit) proves `verifySkillsJarsSources` actually fails when a symlink is broken/missing/wrong-target. Combined with the task not running in CI, the verification logic is currently unverified in both directions.

### Governance note (not a code defect)
This introduces a new distribution channel: skill content sourced from `apache/grails-core` gets scanned and republished under `com.skillsjars:apache__grails-core__*` on Maven Central by a non-ASF third party. Nothing in the build itself talks to skillsjars.com (it's a manual post-merge submission per the new README), so there's no build-time supply-chain exposure — but worth an explicit maintainer/PMC nod that this publishing path is sanctioned, beyond code review (likely already implicit via #15454).

### Suggested (non-blocking) actions
1. Wire `verifySkillsJarsSources` into `check` so drift is caught automatically instead of relying on someone remembering to run it.
2. Replace the `substring(6)` prefix assumption with proper path resolution.
3. Confirm governance sign-off for the third-party publishing workflow.

Nothing here blocks the change — it's small, self-contained, and CI is green — but 1–2 are cheap hardening fixes and 3 is worth a one-line confirmation in the thread.

jdaugherty:
I do not think this should be in RC1.  We should discuss this as I don't understand how publishing a skill jar is consumable by skill jars.  From what I can tell, they have to publish.  We need to discuss this one before going forward.

borinquenkid:
Following up on the ASF publishing concern — I think there's a compromise that sidesteps it entirely rather than trying to find an ASF-compliant way to *route through* SkillsJars.

**The governance problem is coupled to SkillsJars specifically, not to distributing skills at all.** Per @jdaugherty's finding on the rework branch, SkillsJars' own discovery/registry only indexes artifacts published through *their* scan-and-publish pipeline under `com.skillsjars:*` coordinates — publishing a jar ourselves under `org.apache.grails:*` through the normal ASF release train doesn't make it show up there. So there's no version of "publish through SkillsJars" that avoids the voting-rules conflict; the third-party publish step *is* the product.

**Proposal: for `grails-developer`, stop trying to publish it at all — embed it directly into every app `grails-forge` (and/or the legacy `create-app` profiles) generates.**

This repo already has the exact mechanism for this, and it ships zero new files through any third-party service:

- `grails-forge` builds every generated file via `Feature.apply(GeneratorContext)` → `generatorContext.addTemplate(...)` (`DefaultProjectGenerator.java`, `Feature.java`). Static verbatim files (gradle wrapper, GSP layouts, i18n props) are already added this way as classpath resources under `grails-forge-core/src/main/resources/`. A `SKILL.md` would work identically — no new distribution channel, no new build task, no RAT carve-out, nothing outside grails-core's own already-ASF-compliant release pipeline.
- The legacy `grails create-app` path (`grails-shell-cli` + `grails-profiles/*/skeleton`) could get the same file for parity, since those skeletons are just static content copied into generated apps.

**Why this is actually a better fit for `grails-developer` specifically:** its whole purpose is guiding ongoing development of a Grails app — the exact moment `create-app` runs. Shipping it in the generated project puts it in front of 100% of new Grails developers automatically, with no dependency, no Maven coordinate, no separate install step.

**Where this doesn't help — and I don't think we should pretend it does:** `grails-8-upgrade` targets people upgrading an *existing* pre-8 app. A freshly generated app is already on 8, so embedding it at generation time doesn't reach that audience at all. I'd suggest we don't block on solving that one here — either leave it as source docs under `.agents/skills` (discoverable via the README this PR already added) until we have a real answer, or treat it as a separate follow-up once we know what channel we actually want for "skill for an existing project" distribution (which is the same question a compliant SkillsJars-alternative would have to answer anyway).

**Concretely, I'd propose:**
1. Close/pause this PR as-is (per @jamesfredley's earlier comment) rather than continue reworking the SkillsJars integration.
2. Open a new, small PR that adds a `grails-developer` feature/template to `grails-forge` (and optionally the legacy profiles) so it's seeded into every generated app.
3. Leave `grails-8-upgrade` distribution as an open question, not a blocker on #15454 — it was already satisfied by the source doc landing in #15798.

Happy to put together the `grails-forge` PR if folks are on board with this direction.

jdaugherty:
As of Sept 2026, new support has landed in skill jars.  Here's some background, I intend to update this PR now that support has been added.

jdaugherty:
AI Summary on how to implement this now: 

# Self-publishing SkillsJars under your own groupId — findings, a Groovy Gradle plan, and a few asks

## Summary

I like SkillsJars a lot: packaging Agent Skills as Maven artifacts solves versioning and distribution neatly, and the catalog makes skills easy to find. But as far as I can tell there's no way to publish a SkillsJar yourself — the only publishing path is the form on skillsjars.com, which deploys under `com.skillsjars`.

After reading the docs, `SPEC.md`, and the Gradle plugin README, my understanding is:

- The **catalog and the `com.skillsjars` namespace are closed.** Only their pipeline can publish there, and the catalog only lists that groupId.
- The **JAR format and the build plugins are open.** You can package skills yourself, publish them under your own groupId (Maven Central, GitHub Packages, a private Nexus/Artifactory), and the current Gradle/sbt/Maven plugins will extract them like any catalog JAR. You just give up the catalog listing.

Below is what I found, a worked plan for a Groovy Gradle build, the gotchas I hit, and some asks for the maintainers. Corrections welcome.

## What's closed vs. open

### Closed: the catalog and the `com.skillsjars` namespace

From [`[SPEC.md](https://github.com/skillsjars/skillsjars/blob/main/SPEC.md)`](https://github.com/skillsjars/skillsjars/blob/main/SPEC.md):

- The catalog browses Agent Skills that live on Maven Central under `com.skillsjars`; the groupId is always `com.skillsjars`, and the artifactId is `<org>__<repo>__<skill>`.
- Publishing is a `POST /deploy` with `org` and `repo`. The service shallow-clones the public GitHub repo, builds one JAR per `SKILL.md`, versions it as `YYYY_MM_DD-<shorthash>` from the latest commit, GPG-signs the JAR and POM, and uploads the bundle to Maven Central.
- Skills without a resolvable license are skipped.

Since SkillsJars owns that namespace on Central, there's no way to publish into it yourself, and self-published artifacts won't appear in the catalog.

### Open: the format and the plugins

- **JAR layout** ([[docs](https://www.skillsjars.com/docs)](https://www.skillsjars.com/docs)): `META-INF/skills/<org>/<repo>/<skill>/SKILL.md`, with any files the skill references alongside it. The docs already describe packaging SkillsJars yourself with the Maven plugin's `package` goal, and the POM property convention `skillsjars.skill.<name>.allowed-tools`, which the plugin validates against the frontmatter.
- **Gradle plugin** ([[README](https://github.com/skillsjars/skillsjars-gradle-plugin)](https://github.com/skillsjars/skillsjars-gradle-plugin)): [[PR #1](https://github.com/skillsjars/skillsjars-gradle-plugin/pull/1)](https://github.com/skillsjars/skillsjars-gradle-plugin/pull/1) (merged 2026-09-06, released as 0.1.0 per the maintainer's comment) brought sbt-plugin parity. The plugin now:
  - resolves dependencies from a dedicated `skill` configuration (`skill 'group:artifact:version'`), **for any groupId**, keeping them off the compile/runtime classpath;
  - looks for skill content in `META-INF/skills/` and `META-INF/resources/skills/`;
  - adds `packageSkillsJars`, which packs local `skills/` directories into `META-INF/skills/...` and is wired into `processResources`/`jar` when the `java` plugin is applied;
  - exposes a `skillsjars { gitHubUrl; allowedTools; outputDir }` extension and validates `allowed-tools` declarations during packaging.
- **Spring AI**: the SkillsTool reads straight from `classpath:/META-INF/skills`, so JVM agents pick up self-published skills with no extraction step and no groupId dependence.

> ⚠️ [[skillsjars.com/docs](https://www.skillsjars.com/docs)](https://www.skillsjars.com/docs) and [[/setup](https://www.skillsjars.com/setup)](https://www.skillsjars.com/setup) still show Gradle plugin `0.0.2`, and the setup guide says that version scans resolvable configurations for the `com.skillsjars` group only. Consumers on 0.0.2 will not extract self-published JARs; they need ≥ 0.1.0.

## Plan: self-publishing from a Groovy Gradle project

### Layout

One subproject per skill if you want independent versions (mirrors the one-JAR-per-skill model of the catalog); a single project if the skills should travel together — `packageSkillsJars` packs everything under `skills/`.

```
agent-skills/
├── settings.gradle
├── build.gradle
├── skills/
│   └── geb-testing/
│       ├── SKILL.md
│       └── references/
└── .github/workflows/publish.yml
```

### Producer `build.gradle`

```groovy
plugins {
    id 'java-library'
    id 'maven-publish'
    id 'signing'
    id 'com.skillsjars.gradle-plugin' version '0.1.0'
}

group = 'com.yourorg.skills'
version = '1.0.0'   // semver — no longer tied to the date-hash scheme

skillsjars {
    // Sets the path inside the JAR: META-INF/skills/yourorg/agent-skills/<skill>/
    gitHubUrl = 'https://github.com/yourorg/agent-skills'
    // Only for skills whose SKILL.md declares allowed-tools; the plugin validates the match
    allowedTools = ['geb-testing': 'Bash Read Edit']
}

// Central expects -sources and -javadoc jars even for resource-only artifacts; empty ones are fine
java {
    withSourcesJar()
    withJavadocJar()
}

// Mirror what SkillsJars.com does: derive POM name/description from the frontmatter
def frontMatter(File skillMd) {
    def m = (skillMd.text =~ /(?s)\A---\R(.*?)\R---/)
    m.find() ? m.group(1).readLines().findAll { it.contains(':') }
                .collectEntries { def (k, v) = it.split(/:/, 2); [(k.trim()): v.trim()] } : [:]
}
def fm = frontMatter(file('skills/geb-testing/SKILL.md'))   // single-line YAML values only

publishing {
    publications {
        skill(MavenPublication) {
            from components.java
            artifactId = 'geb-testing'
            pom {
                name = fm.name
                description = fm.description
                url = 'https://github.com/yourorg/agent-skills'
                licenses {
                    license {
                        name = fm.license ?: 'MIT'
                        url = 'https://opensource.org/licenses/MIT'
                    }
                }
                developers { developer { id = 'yourorg'; name = 'Your Org' } }
                scm {
                    url = 'https://github.com/yourorg/agent-skills'
                    connection = 'scm:git:https://github.com/yourorg/agent-skills.git'
                }
                // Same property convention as catalog JARs, readable without opening the JAR
                properties = ['skillsjars.skill.geb-testing.allowed-tools': 'Bash Read Edit']
            }
        }
    }
    repositories {
        maven {
            name = 'GitHubPackages'
            url = uri('https://maven.pkg.github.com/yourorg/agent-skills')
            credentials {
                username = System.getenv('GITHUB_ACTOR')
                password = System.getenv('GITHUB_TOKEN')
            }
        }
    }
}

signing {
    def key = System.getenv('SIGNING_KEY')
    if (key) {
        useInMemoryPgpKeys(key, System.getenv('SIGNING_PASSWORD'))
        sign publishing.publications
    }
}
```

### Where to publish

- **GitHub Packages / private Nexus / Artifactory** — no signing or POM ceremony required, but consumers need credentials in their `repositories {}` block.
- **Maven Central under your own groupId** — go through the Sonatype Central Portal with a portal-aware plugin (e.g. `com.vanniktech.maven.publish` or `com.gradleup.nmcp`; check current versions). Requires a verified namespace, GPG signatures, full POM metadata (name, description, url, licenses, developers, scm), and sources/javadoc jars.

### Consumer `build.gradle`

```groovy
plugins { id 'com.skillsjars.gradle-plugin' version '0.1.0' }

repositories {
    mavenCentral()
    maven {
        url = uri('https://maven.pkg.github.com/yourorg/agent-skills')
        credentials { /* token */ }
    }
}

dependencies {
    skill 'com.yourorg.skills:geb-testing:1.0.0'                        // self-published
    skill 'com.skillsjars:anthropics__skills__pdf:2026_02_25-3d59511'   // from the catalog, same mechanism
}

skillsjars {
    outputDir = layout.projectDirectory.dir('.claude/skills')
}
```

Then `./gradlew extractSkillsJars`, add the output directory to `.gitignore`, and put the extract command in `AGENTS.md` / `CLAUDE.md` so agents run it on checkout.

### Local verification

```bash
./gradlew publishToMavenLocal
unzip -l ~/.m2/repository/com/yourorg/skills/geb-testing/1.0.0/geb-testing-1.0.0.jar | grep META-INF/skills
```

Then add `mavenLocal()` to a consumer project and run `extractSkillsJars` before publishing anywhere remote.

### CI

One workflow: on tag → `./gradlew publish`, with `GITHUB_TOKEN` (or Central Portal credentials), `SIGNING_KEY`, and `SIGNING_PASSWORD` as secrets.

## Gotchas

1. **Extraction directory naming.** Extraction writes `skillsjars__yourorg__agent-skills__geb-testing/`, not `geb-testing/`. [[skillsjars-maven-plugin#4](https://github.com/skillsjars/skillsjars-maven-plugin/issues/4)](https://github.com/skillsjars/skillsjars-maven-plugin/issues/4) notes that some assistants require the skill directory to match the frontmatter `name`. A small `Sync` task that renames after extraction works around it.
2. **Plugin version skew.** The public docs still reference 0.0.2, whose extractor only saw the `com.skillsjars` group. Anyone consuming self-published JARs must be on ≥ 0.1.0.
3. **Libraries carrying their own skill.** Putting `skills/` in a library project bakes the skill into that library's JAR. Classpath-reading agents (Spring AI) find it automatically, but file-based assistants still need the consumer to list the coordinate under `skill` for extraction.
4. **Discoverability is the real cost.** For public skills the two paths coexist: self-publish under your groupId for semver and private consumers, *and* run the same public repo through the skillsjars.com form so it also appears in the catalog as `com.skillsjars:yourorg__agent-skills__geb-testing:<date-hash>`.

## Asks / questions for the maintainers

1. Update [[/docs](https://www.skillsjars.com/docs)](https://www.skillsjars.com/docs) and [[/setup](https://www.skillsjars.com/setup)](https://www.skillsjars.com/setup) to 0.1.0 and document the `skill` configuration, `packageSkillsJars`, and the `skillsjars {}` extension for Gradle.
2. Document self-publishing (JAR layout + POM property conventions) as a supported path, even if the catalog stays `com.skillsjars`-only. Right now the only hint is one sentence about packaging with the Maven plugin.
3. Would the catalog ever index skills outside `com.skillsjars` — e.g. an opt-in registration of coordinates, or an indexer that looks for `META-INF/skills` in Central artifacts? That would keep self-published skills discoverable, which is the main thing lost today.
4. Does the Gradle plugin's `allowedTools` map get written into the generated POM as `skillsjars.skill.<name>.allowed-tools`, or is that left to `maven-publish` (as in the plan above)? The README isn't explicit.
5. A flag on the extract tasks to use the frontmatter `name` as the output directory (see maven-plugin#4) would help assistants that require directory/name parity.

## References

- Docs: https://www.skillsjars.com/docs
- Setup guide: https://www.skillsjars.com/setup
- SPEC.md: https://github.com/skillsjars/skillsjars/blob/main/SPEC.md
- Gradle plugin: https://github.com/skillsjars/skillsjars-gradle-plugin
- Gradle plugin PR #1 (sbt parity, 0.1.0): https://github.com/skillsjars/skillsjars-gradle-plugin/pull/1
- sbt plugin: https://github.com/skillsjars/skillsjars-sbt-plugin
- Maven plugin: https://github.com/skillsjars/skillsjars-maven-plugin
- Directory-naming issue: https://github.com/skillsjars/skillsjars-maven-plugin/issues/4

matrei:
> The catalog and the com.skillsjars namespace are closed. Only their pipeline can publish there, and the catalog only lists that groupId.

@jdaugherty I do not think this is correct, to list self-published skillsjars in the catalog you can make a PR to add your namespace here: https://github.com/skillsjars/skillsjars/blob/main/src/main/resources/application.conf

An example: https://www.skillsjars.com/?bt=maven&q=Spring+Testing+Skills

jdaugherty:
Looks like the skills gradle plugin updates are still not published to maven central or anywhere.  I'm going to just write out own plugin for now and we can replace once they actually release.

jdaugherty:
@matrei I didn't see your comment, but that's probably a better solution. THe problem is their gradle plugin isn't published yet with this support.

jdaugherty:
I opened https://github.com/skillsjars/skillsjars-gradle-plugin/issues/9 too

jdaugherty:
More research: it seems the maven plugin (0.7.0) is released and does contain the support.  So the lag is specific to Gradle.

jdaugherty:
For now, I've written a small custom plugin that can be later replaced by the skilljars gradle plugin (if they release).  I've restructured these skills per the grails-skills/* suggestion I made.  @matrei @jamesfredley  can you please confirm you're ok with this? If you are, we can merge to 8.0.0.  We can also defer to later once skilljars releases their updated plugin.

jdaugherty (approved):

PR #15812, opened 2026-07-02
Remove unused Forge GitHub create and OAuth integration

jamesfredley:
## Summary

Rewrites this PR from OAuth-state hardening to **full removal** of the unused Forge server-side GitHub create / OAuth / push integration.

The start.grails.org UI removed "Push to GitHub" in [grails-forge-ui#61](https://github.com/apache/grails-forge-ui/pull/61) (2025-07). The server endpoints and OAuth app wiring were left behind. Maintaining them (including state validation) is unnecessary attack surface.

## What was removed

- `GitHubCreateController` and related create/OAuth/redirect services, DTOs, clients, and tests
- `GitHubUtil` (JGit push helper) and the `org.eclipse.jgit` dependency
- `GrailsForgeConfiguration.GitHubConfiguration` and `application.yml` OAuth/API client config
- JGit native-image metadata
- Cloud Run deploy env vars for the OAuth app (`GITHUB_OAUTH_APP_*`, `GITHUB_USER_AGENT`), with `--remove-env-vars` so existing services drop them on next deploy
- Threat-model references to the deleted controller

## What was kept

- Generated-app GitHub Actions workflow features (`feature/github/workflows/**`)
- `GITHUB_REDIRECT_URL` / `grails.forge.redirect-url` for browser redirect to the Forge UI

## Additional fix

- `GrailsForgeConfiguration.setRedirectUrl` previously self-assigned and ignored config; it now applies the provided URL (covered by `GrailsForgeConfigurationSpec`)

## Test plan

- [x] `./gradlew :grails-forge-api:test :grails-forge-core:test :grails-forge-web-netty:compileJava` (from `grails-forge/`)
- [x] `GrailsForgeConfigurationSpec` for redirect override

## Follow-ups (ops / UI)

- Revoke/delete the unused GitHub OAuth App credentials after deploy
- Optional: scrub dead UI residue (`getGitHubLink` / callback query handling) in grails-forge-ui


jdaugherty:
Didn't we remove the ui part of create in github?  Why not just remove the function completely

jdaugherty (changes requested):
It's my understanding we removed the ability to deploy to a github repo.  We should remove the functionality rather than maintain it.

jamesfredley:
Worth clarifying the current state, because "we removed the ability to deploy to a github repo" is only half-true in this repo - and the half that's left is the risky one.

What was removed is the **UI**: the "Create in GitHub" action lives in the separate `grails-forge-ui` (React) repo, which is where that removal happened. But the **backend** GitHub create/deploy flow is still fully present and wired here in `grails-core`:

- `grails-forge-api`: `.../create/github/GitHubCreateController` (routes `/github/{type}/{name}` taking `code`/`state`), `GitHubRedirectService` (builds the OAuth authorize URL with `client_id`/`redirect_uri`/`scope`/`state`), `GitHubCreateService` (token exchange -> repo creation -> push), `GitHubCreateOperation`.
- `grails-forge-core`: `.../client/github/oauth/GitHubOAuthClient` + `GitHubOAuthOperations`, `GitHubUtil`.
- `grails-forge-web-netty/src/main/resources/application.yml`: live `github-oauth` / `github-api-v3` service URLs and the OAuth client-id/secret/redirect config.

And critically, the `state` is generated (`UUID.randomUUID()`) but **not validated server-side** - which is exactly the CSRF gap this PR closes.

So we currently ship live, reachable OAuth endpoints with an unhardened `state`. That leaves two coherent options - and the current in-between (UI gone, backend live + unhardened) is the worst of the three:

1. **Retire it fully** - remove the backend too: the `grails-forge-api` `.../create/github` package, the `grails-forge-core` GitHub OAuth client, and the `application.yml` GitHub config, matching the UI removal. That's a separate, larger removal PR.
2. **Keep and harden** - if those endpoints stay shipped, they're live CSRF surface and the state validation here is the correct fix.

I'm happy either way. If the decision is to retire, I'll convert this into the backend-removal PR instead. But if any of those `/github/...` endpoints remain in a released Forge, I'd merge the state validation first so we're not shipping an unhardened OAuth callback. Which direction do you want?

jdaugherty:
@jamesfredley right, we didn't remove in forge backend, but why are we maintaining instead of removing it?

jamesfredley:
This was removed from the UI a long time ago and just needs to be removed from the server side.  It is not used.

jamesfredley:
Cleaning this up once and for all on both ends.

**Backend (this PR):** remove the unused Forge GitHub create / OAuth / push stack that the API still exposed after the UI dropped the button.

**Frontend companion:** [apache/grails-forge-ui#159](https://github.com/apache/grails-forge-ui/pull/159) scrubs the leftover Launch UI residue (github href helpers, OAuth callback routing, clone NextSteps, `createRepo` tooltip).

Together these match the UI change from [grails-forge-ui#61](https://github.com/apache/grails-forge-ui/pull/61) so neither side keeps a half-dead create-on-GitHub path.

borinquenkid (approved):
Verified all review feedback is addressed:

- jdaugherty's changes-requested (2026-07-09: "remove the functionality rather than maintain it") is resolved — the PR was rewritten (commit bfa1f8e, 2026-07-29) from OAuth-state hardening to full removal of the unused Forge GitHub create/OAuth/push backend, matching that request.
- Copilot's 3 inline comments all targeted code in `GitHubCreateController.java` and `GitHubRedirectService.java`, both of which are now fully deleted in the diff — the flagged issues no longer exist.

Approving; jdaugherty to re-review given the rewrite.

jdaugherty (approved):

PR #15814, opened 2026-07-02
Require authentication for Forge analytics report endpoint

jamesfredley:
## Description

The Forge analytics `POST /analytics/report` endpoint accepted unauthenticated writes, allowing fabricated project-generation statistics to be stored. This change authenticates writes while preserving public read-only aggregate endpoints.

## Changes

- Require a Google-issued bearer JWT for `POST /analytics/report`.
- Validate the Google signature/JWKS, issuer, analytics-service audience, and the Forge web service account subject.
- Keep `GET /analytics/top/*` explicitly anonymous and `@ReadOnly`.
- Configure the Forge web analytics client to obtain a Google identity token with the analytics Cloud Run URL as its audience.
- Wire the audience and caller subject through all Forge Cloud Run deployment workflows.
- Add configurable HSTS to the Forge web Netty service.
- Document the new deployment environment variables.

### Test auth and transaction isolation

- Restored `@ReadOnly` on public aggregate methods so transaction interception remains available.
- Replaced the brittle expected-500 anonymous-read assertion with a mocked `FeatureRepository` and a real `200 []` assertion.
- Added a test-only `NoOpTransactionOperationsFactory` scoped with `@Requires(property = 'datasources.default.enabled', value = 'false')` so it applies only to datasource-disabled security tests and does not shadow the real JDBC transaction manager in the Testcontainers/PostgreSQL persistence path.
- Use a lifecycle-managed injected `@Client('/')` in the security spec.
- Updated the persistence-path test to send an authorized bearer token.
- Added two happy-path guard tests to `AnalyticsControllerSecuritySpec` (absolute `@Client('/')` and client-relative `@Client('/analytics')`) that issue a valid bearer token and assert `202 ACCEPTED`.

### Entity / schema alignment (runtime-verified CI root cause)

After the transaction-factory scoping landed, CI still failed `StoreGeneratedProjectStatsSpec > test save generation data` with HTTP 500. Manual Postgres reproduction (Testcontainers 1.20.6 cannot start against Docker 29.5.3 locally) isolated a pre-existing entity/schema mismatch, not the security change itself:

1. **Micronaut Data mapping:** `Application` has field `reloading` / constructor param `reloading`, but the getter was misnamed `getDevelopmentReloading()`, which produced `MappingException: Constructor argument [reloading] of type [DevelopmentReloading] ... must have an associated getter`. Fixed by renaming the getter to `getReloading()`.
2. **Schema drift after TestFramework → DevelopmentReloading:** V1 still had `NOT NULL build_tool` / `test_framework` and no `reloading` column. Persisting an `Application` failed with `column "reloading" of relation "application" does not exist`. Flyway `V2__application_reloading.sql`:
   - `ADD COLUMN reloading varchar(255) NOT NULL DEFAULT 'DEVTOOLS'`
   - `ALTER COLUMN build_tool SET DEFAULT 'GRADLE'` (column retained; entity no longer maps it, but `FeatureRepository.topBuildTools()` still queries it, and the DB default keeps new rows meaningful)
   - `ALTER COLUMN test_framework DROP NOT NULL` (unmapped and unqueried)
3. **Stale assertion:** gorm aggregate name is `HIBERNATE5` after the `GormImpl` rename, not `HIBERNATE`.

## Verification

- `AnalyticsControllerSecuritySpec`: 6/6 passing - unauthenticated POST (`401`), wrong audience (`401`), wrong caller subject (`401`), anonymous GET (`200 []`), and two valid-token POSTs (`202 ACCEPTED`).
- `HstsHeaderFilterSpec`: passing with the deployment audience environment variable unset.
- Full-flow save + `topGorm` / `topBuildTools` / `topReloading` / `topJdkVersion` against manual Postgres 10: green after getter + V2 + assertion fix.
- Oracle and Codex review gate: GREEN on the entity/schema fix diff.
- `StoreGeneratedProjectStatsSpec` (Testcontainers/PostgreSQL) is validated by the Forge CI matrices; it cannot run in this environment because Testcontainers 1.20.6 rejects the local Docker 29.5.3 API with HTTP 400 before the container starts.

## Security boundary

| Endpoint | Access |
| --- | --- |
| `POST /analytics/report` | Authenticated Forge service account only |
| `GET /analytics/top/*` | Public aggregate statistics |

## Contributor Checklist

- [x] Security behavior is covered through public HTTP endpoints.
- [x] Deployment configuration and documentation are updated.
- [x] Entity mapping and schema are aligned for the authenticated persistence path.
- [x] The PR is labeled `ai-generated-starting-point`.
- [x] All contributed code is provided under the Apache License 2.0.

Assisted-by: Sisyphus:grok-4.5 [oracle] [codex]


grails-forge/grails-forge-analytics-postgres/src/main/java/org/grails/forge/analytics/postgres/AnalyticsController.java
```diff
@@ -47,26 +49,31 @@ public AnalyticsController(
     }
 
     @Get("/top/features")
+    @Secured(SecurityRule.IS_ANONYMOUS)
```
jdaugherty, line 50: I'm confused - these indicate they're anonymous but your'e requiring authentication?

jdaugherty: @jamesfredley isn't the point of this to ensure we're authenticated?  Shouldn't all of these IS_ANONYMOUS be removed for the analytics endpoint?

jdaugherty:
The tests are failing to authenticate:


AnalyticsControllerSecuritySpec > top analytics data is not rejected as unauthenticated FAILED
    io.micronaut.http.client.exceptions.HttpClientResponseException at DefaultHttpClient.java:2177

StoreGeneratedProjectStatsSpec > test save generation data FAILED
    io.micronaut.http.client.exceptions.HttpClientResponseException at DefaultHttpClient.java:2177

grails-forge/grails-forge-analytics-postgres/src/test/groovy/org/grails/forge/analytics/postgres/AnalyticsControllerSecuritySpec.groovy
```diff
@@ -0,0 +1,162 @@
...
+
+    void 'top analytics data is not rejected as unauthenticated'() {
+        when:
+        client.toBlocking().exchange(HttpRequest.GET('/top/features'), String)
+
+        then:
+        HttpClientResponseException e = thrown(HttpClientResponseException)
+        e.status == HttpStatus.INTERNAL_SERVER_ERROR
+        0 * applicationRepository._
+    }
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 4adff56413. The test now asserts a successful anonymous response (200 and []) with FeatureRepository mocked, so authentication behavior cannot be hidden by an internal datasource failure. The complete security spec passes.

jamesfredley:
Final review fixes are pushed in `4adff56413`.

The public analytics GET test now returns `200 []` using a mocked `FeatureRepository`, all aggregate methods retain `@ReadOnly`, and the security spec uses a test-only transaction factory plus a lifecycle-managed HTTP client. `AnalyticsControllerSecuritySpec` passes 4/4 and `HstsHeaderFilterSpec` passes with no analytics audience environment variable set.

The exact 19-file PR diff completed Oracle and Codex review GREEN. The persistence-path test remains locally blocked before execution by the Testcontainers 1.20.6 / Docker 29.5.3 API incompatibility documented in the PR description.

borinquenkid:
@jamesfredley — following up on @jdaugherty's question above (2026-07-09, still unanswered): should the `GET /analytics/top/*` endpoints stay `IS_ANONYMOUS`, or should this lock down reads too? Worth a reply in-thread so this isn't left open before merge — even if the answer is "yes, intentional, they're public dashboard aggregates," it'd be good to have that confirmed on the record.

jamesfredley:
Forge analytics will be removed entirely as we move these to AWS.

PR #15779, opened 2026-06-27
(1/3) feat: add SessionResolver infrastructure and extend core datastore APIs

borinquenkid:
> **📖 Reading order: 1 of 3** — this GormRegistry O(M+N) scaling change is split across three PRs; please review them in order:
>
> 1. **(1/3) #15779** — SessionResolver infrastructure & core datastore API extensions
> 2. **(2/3) #15780** — GormRegistry implementation (core production code)
> 3. **(3/3) #15790** — core-class unit tests
>
> _You are reading **(1/3)**._

---

## Summary

Prerequisite for the GormRegistry O(M+N) scaling stack. Introduces low-level infrastructure in `grails-datastore-core` that the GormRegistry layer depends on:

- Add `SessionResolver` and `ThreadLocalSessionResolver` for thread-safe session lookup decoupled from a specific `Datastore` reference
- Extend `AbstractDatastore`, `Datastore`, and `DatastoreUtils` with hooks for registry lifecycle events (register/deregister on open/close)
- Add `MappingContext` and `AbstractMappingContext` extensions needed by `GormRegistry` for entity lookup
- Minor additions to `AbstractPersistentEntity`, `ClassUtils`, `DirtyCheckingSupport`, `ConnectionSourceSettingsBuilder`, `CustomizableRollbackTransactionAttribute`, `AstUtils`

This PR carries no GORM API changes and has no dependency on any GormRegistry class — it can be reviewed and merged independently.

---

## Relationship to the upstream fix

This stack builds on top of [fix/gorm-api-registration-scaling](https://github.com/apache/grails-core/compare/8.0.x...fix/gorm-api-registration-scaling), which solves the same O(M×N) problem through a different strategy. The two approaches are deliberately complementary:

### Upstream fix strategy — incremental optimization (4 files, ~815 lines)

[...]

PR #15817, opened 2026-07-03
feat(grails-data-neo4j): wire Neo4j adapter to GormRegistry's GormApiFactory mechanism (PR3)

borinquenkid:
> **📖 Neo4j GormRegistry migration — reading order: 3 of 5**:
>
> 1. #15816 — PR1: Groovy4/Jakarta/GORM8 baseline migration (no GormRegistry coupling)
> 2. #15951 — PR2: fold standalone build into root `settings.gradle`
> 3. **#15817** — PR3: wire `Neo4jGormApiFactory` into GormRegistry — *this PR*
> 4. #15832 — PR4: re-add example apps and docs into monorepo
> 5. #15833 — PR5: Checkstyle/CodeNarc cleanup
>
> _You are reading **3 of 5**._
>
> This Neo4j chain branches off the core GormRegistry O(M+N) scaling stack's head (#15779 → #15780 → #15790) — it depends on that stack but is its own separate sequence, not additional steps within it.
>
> Previously combined PR2 (wiring) and PR3 (settings.gradle fold) in this one PR; split per [review feedback](https://github.com/apache/grails-core/pull/15816#issuecomment-4931129546) so the fold lands earlier (#15951, now PR2) and this PR is wiring-only.

## Summary

Completes the Neo4j GormRegistry migration plan's PR3, on top of #15816 (PR1, baseline) and #15951 (PR2, `settings.gradle` fold): registers a `Neo4jGormApiFactory` so entities backed by `Neo4jDatastore` resolve a `Neo4jGormStaticApi` through `GormRegistry` instead of silently falling back to the generic `GormStaticApi`. Without this, static/cypher-string API calls (`cypherStatic`, `findRelationship(s)`, `findPath*`, `findShortestPath`, `find`/`findAll` with a cypher string) threw `ClassCastException` or `UnsupportedOperationException`, since those methods only exist on `Neo4jGormStaticApi`.

**This turned out to require far less than the "rewrite Neo4jEntity/Node/Relationship traits" originally scoped in the migration plan** - a nice, concrete dividend of the GormRegistry refactor: adding a new datastore's wiring is now a small, mechanical exercise rather than an integration project. Neo4j's instance and validation APIs were already generic (`GormInstanceApi`/`GormValidationApi`, no Neo4j-specific subclass), matching `DefaultGormApiFactory`'s base implementations already - only the static API needed a factory override, mirroring `MongoGormApiFactory`'s exact shape (which only overrides `createStaticApi()` for the same reason). The whole change is a ~25-line factory class plus a single line registering it.

- `Neo4jGormApiFactory#createStaticApi()` resolves the datastore via `DatastoreResolver#resolve()` rather than Neo4j's old bespoke `getDatastoreForQualifier()`/`datastoresByConnectionSource` logic - qualifier/multi-datasource routing is now handled generically by `GormRegistry`/`GormApiResolver` (the O(M+N) scaling work this plan depends on). Verified this doesn't regress Neo4j's own multi-tenancy/multi-datasource tests.
- `registerApiFactory()` is called from `Neo4jDatastore#initialize()`, before constructing the `GormEnhancer` whose constructor eagerly registers this datastore's entities - registering after would leave those entities bound to the generic factory forever, since `GormEnhancer` only registers each entity once.
- 17 of the 34 tests marked `@PendingFeature` in the baseline PR now pass and had their annotations removed (`ApiExtensionsSpec`, `CypherQueryStringSpec`, `OneToManyUpdateSpec`, `MultiTenancySpec`, `PathSpec`, `RelationshipSpec`, `NativeIdentityGeneratorSpec`'s `saveAll` test - whose fix was already in place from the baseline PR but unreachable until this factory was registered).

## Test plan

- [x] `./gradlew :grails-datastore-gorm-neo4j:test` - `BUILD SUCCESSFUL`, 198/215 passing, 0 failures, 17 skipped (4 genuinely pending and unrelated to this change, plus pre-existing `@Ignore`'d tests)
- [x] Verified no regressions in previously-passing tests, including multi-tenancy/multi-datasource specs, after switching to the generic `DatastoreResolver`-based routing

Generated with Claude

### DIFF
--- a/grails-data-neo4j/grails-datastore-gorm-neo4j/src/main/groovy/org/grails/datastore/gorm/neo4j/Neo4jDatastore.java
+++ b/grails-data-neo4j/grails-datastore-gorm-neo4j/src/main/groovy/org/grails/datastore/gorm/neo4j/Neo4jDatastore.java
@@ -434,6 +434,10 @@ public class Neo4jDatastore extends AbstractDatastore implements Closeable, Stat
     protected GormEnhancer initialize(Neo4jConnectionSourceSettings settings) {
         registerEventListeners(this.eventPublisher);
 
+        // Must run before the GormEnhancer constructor (below) registers this datastore's entities,
+        // otherwise GormRegistry falls back to the generic DefaultGormApiFactory for them.
+        org.grails.datastore.gorm.GormRegistry.getInstance().registerApiFactory(Neo4jDatastore.class, new Neo4jGormApiFactory());
+
         this.mappingContext.addMappingContextListener(new MappingContext.Listener() {
             @Override
             public void persistentEntityAdded(PersistentEntity entity) {
--- a/grails-data-neo4j/grails-datastore-gorm-neo4j/src/main/groovy/org/grails/datastore/gorm/neo4j/Neo4jGormApiFactory.groovy
+++ b/grails-data-neo4j/grails-datastore-gorm-neo4j/src/main/groovy/org/grails/datastore/gorm/neo4j/Neo4jGormApiFactory.groovy
@@ -1,0 +1,49 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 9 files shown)


borinquenkid:
Superseded by the consolidated migration in #15972 (same commits, single diff). Closing to reduce PR-list noise.

PR #15827, opened 2026-07-05
Extract shared BOM build conventions

jamesfredley:
## Summary

This PR reduces duplicated Gradle BOM build logic by moving the shared publication, POM customization, constraint extraction, and release snapshot validation conventions into `org.apache.grails.buildsrc.bom-conventions`.

The BOM build scripts now keep only BOM-specific inputs such as descriptions, dependency declarations, and variant-specific dependency sets. The shared plugin owns the common task wiring and POM mutation behavior used by the seven Grails BOM projects.

## Changes

- Add `BomConventionsPlugin` in `build-logic/plugins` and register it as `org.apache.grails.buildsrc.bom-conventions`.
- Apply the shared plugin in all seven BOM projects.
- Preserve BOM-specific dependency and platform declarations in each BOM script.
- Reduce duplicated Groovy module override maps in `dependencies.gradle` while leaving one literal coordinate list for Dependabot visibility.
- Remove unreachable BOM override branches for Micronaut/Hibernate variants.
- Address Copilot feedback by using the configured `gradleBuildProjects` group mapping when registering Gradle-build artifacts for POM property replacement.
- Add TestKit coverage for shared BOM plugin wiring and Gradle-build artifact POM property generation.

## Verification

- `JAVA_HOME=/tmp/opencode/jdk21 ./gradlew :build-logic-root:build-logic:compileGroovy :build-logic-root:build-logic:test`
- `JAVA_HOME=/tmp/opencode/jdk21 ./gradlew :build-logic-root:build-logic:test --tests org.apache.grails.buildsrc.BomConventionsPluginSpec`
- Generated all seven BOM POMs with `-PincludeMicronautProjects`.
- Ran `extractConstraints` for all seven BOMs with `-PincludeMicronautProjects`.
- Compared generated constraints against a clean `origin/8.0.x` baseline: no differences.
- Compared generated POMs against the baseline: only the intended base BOM improvement differs, where Gradle-build artifacts now use generated version properties instead of inline `8.0.0-SNAPSHOT` versions.

## Notes

- LSP diagnostics are unavailable locally for `.groovy` / `.gradle` because no Groovy/Gradle LSP is configured.
- The full aggregate gate currently has unrelated baseline failures on `origin/8.0.x` (GraphQL static compile errors and MongoDB timeout flakes), reproduced in a clean baseline worktree.


build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+                propertyNameCalculator.addForcedGroupPrefix('org.apache.grails.profiles', 'grails-profile')
+                propertyNameCalculator.addProjects(project.rootProject.subprojects)
+                for (String gradleArtifactId : project.ext.gradleBuildProjects) {
+                    propertyNameCalculator.addProject(
+                            'org.apache.grails.gradle',
+                            gradleArtifactId,
+                            project.version as String,
+                            gradleArtifactId
+                    )
+                }
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 206a4b6129 by iterating the configured gradleBuildProjects map and passing each artifact-specific group into PropertyNameCalculator.addProject. The regenerated base BOM POM now emits version properties for the Gradle-build artifacts, and the constraints output remains unchanged versus the 8.0.x baseline.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+            project.rootProject.subprojects.each { Project subproject ->
+                project.evaluationDependsOn(subproject.path)
+            }
+            task.projectArtifactIds.set(project.provider { projectArtifactIds(project) })
+            task.forcedGroupPrefixes.set(['org.apache.grails.profiles': 'grails-profile'])
+            task.projectCoordinateProperties.set(project.provider { projectCoordinateProperties(project) })
+            task.dependsOn('generateMetadataFileForMavenPublication', 'generatePomFileForMavenPublication')
+        }
+
+        project.tasks.named('check') { Task task ->
```
jdaugherty: This adds new lifecycle behavior that was not present in any of the extracted BOM scripts: `check` now runs `extractConstraints`, which also runs both publication generators and resolves the full BOM graph. `grails-doc:generateBomDocumentation` already invokes `extractConstraints` explicitly. It also makes a previously configuration-cache-compatible BOM check fail: `./gradlew :grails-base-bom:check --configuration-cache --rerun-tasks` reports unsupported serialization of `ConfigurationContainer`, `DependencyHandler`, and `Project`. Could we remove this `check` dependency and keep constraint extraction explicit?

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+        configureExtractConstraints(project)
+        configureSnapshotValidation(project)
+    }
+
+    private static void applyDependencyDefinitions(Project project) {
+        project.apply([from: project.rootProject.layout.projectDirectory.file('dependencies.gradle')])
+    }
+
+    private static void configureCoordinates(Project project) {
+        project.version = project.findProperty('projectVersion')
```
jdaugherty: The extracted scripts used `version = projectVersion`, which fails with `MissingPropertyException` when `projectVersion` is not defined. `findProperty` returns null here, leaving `project.version` as `unspecified`; `projectCoordinateProperties()` and the POM customization then register coordinates like `org.apache.grails:grails-gradle-plugins:unspecified`, and version property replacement silently stops matching with no error. Suggest failing fast, e.g. `project.property('projectVersion')` or an explicit check.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        if (project.extensions.extraProperties.get('isReleaseBuild') &&
+                project.extensions.extraProperties.get('isPublishedExternal')) {
+            project.tasks.matching { Task task ->
+                task.name in ['generateMetadataFileForMavenPublication', 'generatePomFileForMavenPublication']
```
jdaugherty: Behavior change from the extracted scripts: they used `afterEvaluate { tasks.named(...).configure { ... } }`, which throws `UnknownTaskException` if the publication tasks are missing. `tasks.matching { }.configureEach { }` silently no-ops if these tasks are never created (e.g. the publish plugin renames its publication), so on a real release the snapshot validation could silently not run. Since this guard only matters when `isReleaseBuild && isPublishedExternal`, consider asserting the tasks exist in that case rather than matching by name.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+                                groupId,
+                                artifactId,
+                                inlineVersion,
+                                isBom
+                        )
+                        if (extractedConstraint?.versionPropertyReference) {
+                            dep.version[0].value = extractedConstraint.versionPropertyReference
+                            pomProperties.put(extractedConstraint.versionPropertyName, inlineVersion)
+                        }
+                    } else if (!inlineVersion) {
```
jdaugherty: `else if (!inlineVersion)` is always true at this point — carried over from the original scripts, but since this code was rewritten anyway, a plain `else` would be clearer.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+import org.gradle.api.artifacts.Dependency
+import org.gradle.api.plugins.JavaPlatformExtension
+import org.gradle.api.tasks.TaskProvider
+
+import org.apache.grails.gradle.tasks.bom.ExtractDependenciesTask
+import org.apache.grails.gradle.tasks.bom.ExtractedDependencyConstraint
+import org.apache.grails.gradle.tasks.bom.PropertyNameCalculator
+
+@CompileStatic
+class BomConventionsPlugin implements Plugin<Project> {
```
jdaugherty: I would rename this to GrailsBomPlugin

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/BomConventionsPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+        applyDependencyDefinitions(project)
+        configureCoordinates(project)
+        configureExtraProperties(project)
+        configurePomCustomization(project)
+
+        project.pluginManager.apply('java-platform')
+        project.extensions.configure(JavaPlatformExtension) { JavaPlatformExtension extension ->
+            extension.allowDependencies()
+        }
+        project.pluginManager.apply('org.apache.grails.buildsrc.publish')
```
jdaugherty: We have not used aggregate plugins to date. These should be left on the individual projects (publish & sbom)

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsBomPlugin.groovy
```diff
@@ -0,0 +1,261 @@
...
+        project.pluginManager.apply('org.apache.grails.buildsrc.publish')
+        project.pluginManager.apply('org.apache.grails.buildsrc.sbom')
+
+        configureBomDependencies(project)
+        configureExtractConstraints(project)
+        configureSnapshotValidation(project)
+    }
+
+    private static void applyDependencyDefinitions(Project project) {
+        project.apply([from: project.rootProject.layout.projectDirectory.file('dependencies.gradle')])
```
jdaugherty, line 73: The difference between this and before is it was applied in the buildscript so it was available in all contexts. I think this is ok, but it's important to note the difference.

dependencies.gradle
```diff
@@ -20,6 +20,18 @@
 
 // This file provides the 'grails-bom' dependencies, while grails-gradle/dependencies.gradle provides the 'grails-gradle-bom' dependencies
 // These files are split to facilitate separation of build vs application dependencies. These are the application dependencies.
+static Map<String, String> groovyModuleDependencies(String groovyVersion) {
+    [
+            '', '-ant', '-astbuilder', '-cli-commons', '-cli-picocli', '-console', '-contracts', '-datetime',
```
jdaugherty: This is harder to read and should be reverted.  the standard of this file is to list the dependencies. these extractions make it harder to maintain

jdaugherty (changes requested):
I've made several comments inline, but my major issue is we should never use .matching {} - this gradle code is non-lazy.

jamesfredley:
The Gradle-technical points are valid - I'll address them:

- **`.matching { }.configureEach { }`** - agreed this is the main issue. It silently no-ops if the publication tasks are never created (e.g. the publish plugin renames its publication), so on a real release the snapshot validation could silently not run. Since that guard only matters when `isReleaseBuild && isPublishedExternal`, I'll assert the tasks exist in that case rather than matching by name.
- **`check` -> `extractConstraints` lifecycle** - I'll remove it and keep constraint extraction explicit (`grails-doc:generateBomDocumentation` already invokes it). As you noted it also breaks the previously configuration-cache-compatible BOM check (`ConfigurationContainer`/`DependencyHandler`/`Project` serialization).
- **`projectVersion` handling** - fail fast (`project.property('projectVersion')` / explicit check) instead of letting `findProperty` return null and silently producing `:unspecified` coordinates.
- **Rename to `GrailsBomPlugin`**, and **keep `publish` & `sbom` on the individual projects** rather than an aggregate plugin (matching existing convention).
- **`else if (!inlineVersion)`** - simplify to a plain `else`.
- Noted your point that the buildscript-vs-plugin application difference (line 73) is OK but worth calling out.

On **`dependencies.gradle`** - fair. The standard for that file is a flat, readable list of dependencies, and the extractions make it harder to maintain. I'll revert the `dependencies.gradle` extraction and keep the flat listing.

Net: this becomes a smaller, lazier convention-plugin extraction (the parts that genuinely dedupe the BOM build logic) without the `dependencies.gradle` churn, the aggregate-plugin behavior, or the non-lazy `.matching`/`check`-hook changes. I'll push the rework.

jdaugherty:
@jamesfredley let me know when these changes are done and I can make a second pass at reviewing

borinquenkid:
@jamesfredley just a friendly nudge

PR #15828, opened 2026-07-05
Improve hot path virtual-thread and codec performance

### LINKED ISSUE #15374: Grails 8 grails-core: Invoke Dynamic (Indy) Optimization Opportunities

jamesfredley:
## Summary

This PR makes two Grails 8 hot paths friendlier to the JDK 21 runtime without changing the public API surface:

- Replaces `GrailsHttpSession` wrapper monitor synchronization with a private `ReentrantLock`.
- Static-compiles `grails-codecs-core` codec extension helpers and removes dynamic self-dispatch inside the codec implementations.
- Keeps existing Groovy extension method binary descriptors intact by preserving `Object` return types and the existing mutable `HEXDIGITS` property shape.

### Virtual-thread friendliness

`GrailsHttpSession` previously serialized wrapper access with `synchronized (this)`. That object monitor can pin a virtual-thread carrier if the servlet container or request/session access blocks while the monitor is held. A private `ReentrantLock` preserves the wrapper-level serialization semantics while avoiding monitor pinning, and it also makes lazy session creation occur under the same lock as subsequent session access.

### Codec hot-path static compilation

The codec extension methods are small, frequently used helpers. Static compilation removes runtime Groovy dispatch inside MD5, SHA-1, SHA-256, Base64, Hex, and shared digest conversion paths. The implementation still keeps the extension methods compatible for existing compiled callers by preserving the original public method descriptors.

## Compatibility notes

- `decodeHex` keeps the previous Groovy truth behavior for falsy targets by using Groovy's runtime truth conversion helper under `@CompileStatic`.
- `String`, primitive byte arrays, wrapper byte arrays, number lists, and null/`NullObject` cases remain covered.
- No new dependencies or user-facing configuration changes are introduced.

## Review feedback addressed

- Restored the old `decodeHex` Groovy truth guard behavior for falsy inputs after Copilot review feedback.
- Added regression coverage for falsy values against the static extension implementation.

## Verification

- `./gradlew :grails-codecs-core:test --tests "org.grails.web.codecs.HexCodecTests" :grails-codecs-core:codeStyle --rerun-tasks`
- `./gradlew :grails-codecs-core:test :grails-codecs-core:codeStyle --rerun-tasks`
- `./gradlew :grails-codecs-core:test :grails-web-common:test --tests "grails.web.servlet.mvc.GrailsHttpSessionSpec" :grails-codecs-core:codeStyle :grails-web-common:codeStyle --rerun-tasks`
- `./gradlew clean aggregateViolations :grails-test-report:check --continue`
  - Violation reports were clean: Checkstyle, CodeNarc, PMD, SpotBugs.
  - The full gate still hits pre-existing unrelated failures on clean `origin/8.0.x`: `grails-validation:compileGroovy`, `grails-data-graphql-core:compileGroovy`, and MongoDB/Testcontainers environment failures. The two compile failures reproduce in a separate clean `origin/8.0.x` worktree.

## Review gate

- Current full branch review gate passed for diff hash `8102545e609fd576dfc616e73c59e4c7da847dbf` with both Oracle and Codex reviewers green.
- Copilot follow-up commit review gate passed for staged diff hash `fcb2c9b25edffd17c4910e79aa23047e3169e608` with both Oracle and Codex reviewers green.


grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/HexCodecExtensionMethods.groovy
```diff
@@ -22,50 +22,45 @@ import java.nio.charset.StandardCharsets
...
+            result.append(hexDigits.charAt((unsignedValue & 0xF0) >> 4))
+            result.append(hexDigits.charAt(unsignedValue & 0x0F))
         }
         return result.toString()
     }
 
-    static decodeHex(theTarget) {
-        if (!theTarget) return null
+    static Object decodeHex(Object theTarget) {
+        if (theTarget == null || theTarget instanceof NullObject || theTarget.toString().length() == 0) return null
```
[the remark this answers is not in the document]
jamesfredley: Fixed in 211a1ffae6. `decodeHex` now preserves the old Groovy truth behavior under `@CompileStatic` by using `DefaultTypeTransformation.castToBoolean`, and `HexCodecTests` covers falsy values including empty string, zero, false, empty list, and empty byte array.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/DigestUtils.groovy
```diff
@@ -18,30 +18,69 @@
...
+    private static byte[] toByteArrayFromArray(Object data, int length) {
+        byte[] result = new byte[length]
+        for (int i = 0; i < length; i++) {
+            result[i] = toByte(Array.get(data, i))
+        }
+        return result
+    }
+
+    private static byte toByte(Object value) {
+        return ((Number) value).byteValue()
```
jdaugherty: This narrows the accepted element types compared to the old dynamic code. Previously `src[i] = v` went through Groovy's `DefaultTypeTransformation` coercion, which also accepted `Character` elements (and `char[]` arrays via the array path); now any non-`Number` element throws `ClassCastException`. If preserving the old tolerance matters, `DefaultTypeTransformation.castToNumber(value).byteValue()` keeps the coercion while staying statically compiled. If the narrowing is intentional, it's probably worth a line in the PR compatibility notes.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/DigestUtils.groovy
```diff
@@ -18,30 +18,69 @@
...
-        def src
-        if (data instanceof Byte[] || data instanceof byte[]) {
-            src = data
+        MessageDigest md = MessageDigest.getInstance(algorithm)
+        byte[] src = toByteArray(data)
+        md.update(src) // This probably needs to use the thread's Locale encoding
+        return md.digest()
+    }
+
+    protected static byte[] toByteArray(Object data) {
```
jdaugherty, line 43: Nit: `protected` reads as "for subclasses", but the actual consumers (`Base64CodecExtensionMethods`, `HexCodecExtensionMethods`) rely on `protected`'s package-access side effect. `@PackageScope` (or a brief comment) would make the intent clearer.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/HexCodecExtensionMethods.groovy
```diff
@@ -21,51 +21,47 @@ package org.grails.plugins.codecs
...
         }
 
-        def result = new StringBuilder()
-        if (theTarget instanceof String) {
-            theTarget = theTarget.getBytes(StandardCharsets.UTF_8)
-        }
-        theTarget.each() {
-            result << HexCodecExtensionMethods.HEXDIGITS[(it & 0xF0) >> 4]
-            result << HexCodecExtensionMethods.HEXDIGITS[it & 0x0F]
+        byte[] bytes = theTarget instanceof String ? ((String) theTarget).getBytes(StandardCharsets.UTF_8) : DigestUtils.toByteArray(theTarget)
```
jdaugherty, line 39: Behavior change for iterable targets that aren't `List`s or arrays (e.g. a `Set<Integer>` or an `Iterator`): the old code element-wise hex-encoded anything via `theTarget.each { ... }`, but `DigestUtils.toByteArray` only special-cases `byte[]`/`Byte[]`/`List`/arrays and otherwise falls through to `toString().getBytes(UTF_8)` — so a `Set` of numbers now silently encodes its string representation instead of its elements. Consider handling `Collection`/`Iterable` in `toByteArray` (or at least calling this out in the compatibility notes, since the failure mode is silent wrong output rather than an exception).

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/HexCodecExtensionMethods.groovy
```diff
@@ -21,51 +21,47 @@ package org.grails.plugins.codecs
...
-            else {
-                output << (currentByte | HEXDIGITS.indexOf(val))
-                currentByte = 0
-            }
+        byte[] result = new byte[str.size().intdiv(2)]
+        String hexDigits = (String) HEXDIGITS
+        for (int i = 0; i < str.size(); i += 2) {
+            int high = hexDigits.indexOf((int) str.charAt(i))
+            int low = hexDigits.indexOf((int) str.charAt(i + 1))
+            result[i.intdiv(2)] = (byte) ((high << 4) | low)
```
jdaugherty: Nit for a perf-focused PR: `i.intdiv(2)` boxes on every iteration; `i >> 1` (and `str.length() >> 1` on line 58) avoids it. Same for `str.size()` vs plain `str.length()`. Also, on line 51 the explicit `null`/`NullObject` checks are redundant — `DefaultTypeTransformation.castToBoolean` already returns `false` for both.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/MD5CodecExtensionMethods.groovy
```diff
@@ -18,15 +18,19 @@
...
+
+@CompileStatic
 class MD5CodecExtensionMethods {
 
     // Returns the byte[] of the digest, taken from UTF-8 of the string representation
     // or the raw data coerced to bytes
-    static encodeAsMD5(theTarget) {
-        theTarget.encodeAsMD5Bytes()?.encodeAsHex()
+    static Object encodeAsMD5(Object theTarget) {
+        byte[] digest = (byte[]) MD5BytesCodecExtensionMethods.encodeAsMD5Bytes(theTarget)
```
jdaugherty, line 29: Just confirming this is intended: the old `theTarget.encodeAsMD5Bytes()?.encodeAsHex()` dispatched dynamically, so a runtime metaclass override of `encodeAsMD5Bytes` (e.g. an application-supplied codec replacing the default) would also change what `encodeAsMD5` produced. The direct static call hardwires the default implementation. Same applies to the SHA-1/SHA-256 variants. Probably fine — it's the whole point of removing dynamic dispatch — but worth a note in the compatibility section.

grails-codecs-core/src/test/groovy/org/grails/web/codecs/HexCodecTests.groovy
```diff
@@ -48,6 +49,13 @@ class HexCodecTests {
         assertIterableEquals(new Byte[] {65, 32, 66, 32, 67, 32, 68, 32, 69}.toList(), result.toList())
         //make sure decoding null returns null
         assertEquals(null.decodeHex(), null)
+
+        //make sure decoding Groovy-falsy values returns null
+        assertEquals(null, HexCodecExtensionMethods.decodeHex(''))
+        assertEquals(null, HexCodecExtensionMethods.decodeHex(0))
+        assertEquals(null, HexCodecExtensionMethods.decodeHex(false))
+        assertEquals(null, HexCodecExtensionMethods.decodeHex([]))
+        assertEquals(null, HexCodecExtensionMethods.decodeHex(new byte[0]))
```
jdaugherty: Per the repo rule to test via public APIs, these could use the extension-method form users actually call — `''.decodeHex()`, `0.decodeHex()`, `false.decodeHex()`, `[].decodeHex()`, `new byte[0].decodeHex()` — which also exercises the extension-module registration path rather than the static class directly (matching the existing `null.decodeHex()` assertion above).

grails-codecs-core/src/test/groovy/org/grails/web/codecs/HexCodecTests.groovy
```diff
@@ -48,6 +49,13 @@ class HexCodecTests {
...
     }
 
     void testRoundtrip() {
```
jdaugherty, line 61: Pre-existing, but since this file is being touched: `testRoundtrip` has no `@Test` annotation, so it never runs — and it's exactly the round-trip coverage the new static implementations want. Note that enabling it as-is will likely fail on the first assertion: `assertIterableEquals` compares `Integer` expected values against the `Byte` elements from `decodeHex(...).toList()`, and `Integer(65).equals(Byte(65))` is false — the expectations need a byte-typed list.

jdaugherty (commented):
I don't see anything major with this at first glance.  I asked Fabled to take a look and the results are attached to the review.

The `GrailsHttpSession` lock conversion looks correct — the remaining `synchronized (this)` blocks are all in commented-out dead code, lazy session creation now happens under the same lock as access (an improvement over the old racy pattern), and the new spec covers the lazy-create/delegate path. The codec static compilation is a nice win too; my comments are mostly about a few places where the rewrite narrows the accepted input types compared to the old dynamic coercion, plus some test-convention and hot-loop nits.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/Base64CodecExtensionMethods.groovy
```diff
@@ -24,32 +24,35 @@ import org.codehaus.groovy.runtime.NullObject
...
 
-    static encodeAsBase64(theTarget) {
+    static Object encodeAsBase64(Object theTarget) {
         if (theTarget == null || theTarget instanceof NullObject) {
             return null
         }
 
         if (theTarget instanceof Byte[] || theTarget instanceof byte[]) {
-            return new String(Base64.encodeBase64(theTarget))
+            return new String(Base64.encodeBase64(DigestUtils.toByteArray(theTarget)), StandardCharsets.UTF_8)
```
jdaugherty, line 44: UTF_8 should be an added method argument that is defaulted to this value.

grails-codecs-core/src/main/groovy/org/grails/plugins/codecs/Base64CodecExtensionMethods.groovy
```diff
@@ -24,32 +24,35 @@ import org.codehaus.groovy.runtime.NullObject
...
         if (theTarget == null || theTarget instanceof NullObject) {
             return null
         }
 
         if (theTarget instanceof Byte[] || theTarget instanceof byte[]) {
-            return Base64.decodeBase64(theTarget)
+            return Base64.decodeBase64(DigestUtils.toByteArray(theTarget))
         }
 
         return Base64.decodeBase64(theTarget.toString().getBytes(StandardCharsets.UTF_8))
```
jdaugherty, line 59: I think you should take the opportunity to make UTF_8 an added method argument that's defaulted

jamesfredley:
Thanks for the detailed pass (and the Fabled review). These read as review-then-commit nits rather than blockers, and they're all reasonable - I'll take them:

1. **Narrowed element types in `DigestUtils`** - restore the old `Character` / `char[]` tolerance with `DefaultTypeTransformation.castToNumber(value).byteValue()` (keeps the coercion while staying statically compiled). If we decide the narrowing is intentional instead, I'll add a line to the compatibility notes.
2. **`protected` vs `@PackageScope`** - switch to `@PackageScope` so the package-access intent (for `Base64CodecExtensionMethods` / `HexCodecExtensionMethods`) is explicit rather than relying on `protected`'s side effect.
3. **Iterable-target behavior change** - handle `Collection` / `Iterable` in `DigestUtils.toByteArray` so a `Set<Integer>` / `Iterator` hex-encodes its *elements* instead of silently falling through to `toString().getBytes(UTF_8)` (the silent-wrong-output failure mode you flagged).
4. **Hot-loop nits** - `i >> 1` and `str.length() >> 1` instead of `intdiv` (avoids boxing), plain `str.length()`, and drop the redundant `null`/`NullObject` checks (`castToBoolean` already returns false).
5. **MD5/SHA static dispatch** - confirmed intentional (that's the point of removing dynamic dispatch); I'll note in the compatibility section that a runtime metaclass override of `encodeAsMD5Bytes`/SHA variants no longer changes the result.
6. **`HexCodecTests`** - move to the extension-method form (`''.decodeHex()`, `0.decodeHex()`, ...) per the test-via-public-API rule.
7. **`testRoundtrip`** - enable it (add `@Test`) with byte-typed expectations so `assertIterableEquals` compares `Byte` to `Byte`.
8. **`UTF_8`** - make it an added method argument defaulted to `StandardCharsets.UTF_8` in the Base64 methods.

I'll push these fixes. On the `GrailsHttpSession` lock conversion - glad that read correctly. Since this is COMMENTED (not blocked), let me know if you'd prefer I split the pure compatibility-note items from the code fixes.

jdaugherty (approved):

PR #15834, opened 2026-07-06
Merge 7.x (with spring security / mail / redis ) into 8.x

jdaugherty:
The 7.x spring security / mail / redis merge was done against the 7.x grails-core branch.  This PR merges 7.2.x to 8.x while also pulling in the changes made by @matrei in the 8.x spring security branch to support Grails 8.  

Please note: groovy 5 has been merged into grails-core so as part of this merge, I had to fix the spring security code to be compatible with Groovy 5.  I opened https://issues.apache.org/jira/browse/GROOVY-12134 after finding a `@Delegate` bug involving final methods.


jdaugherty:
There appears to be a hibernate issue in redis in Grails 8 too:

**RedisIntegrationSpec > testMemoizeDomainList and testMemoizeDomainObject both fail with org.hibernate.LazyInitializationException: could not initialize proxy [com.example.Book#N] - no Session.**

The failing assertion (line 62, cacheMissList*.id) throws LazyInitializationException during evaluation — meaning memoizeDomainList/memoizeDomainObject return uninitialized Hibernate proxies that need a session, and accessing them outside one now fails

jdaugherty:
This will need more work, there appears to be a hibernate regression here.  

**From AI researching (I agree):**
The actual issue: on line 136, cacheMissBook.id — accessing .id on a GORM Hibernate proxy — goes through ByteBuddyInterceptor.intercept → getImplementation → initialize and needs a session. Normally the identifier getter (getId()) short-circuits on a Hibernate proxy and returns the id without initializing. Here it's fully initializing, so this is a Groovy 5 ↔ GORM/hibernate5 proxy-identifier-access interaction.

jdaugherty:
The `@Delegate` fix did not suffice.  It seems the wrapped delegate was still null.  I'm not sure how it was set originally - likely in groovy?  I added an explicit assignment to work around this for now. We probably should at least understand why that issue exists.

jdaugherty:
This is all I can work on for now, if anyone has time to look into these issues, it would be appreciated: 
1. why the `@Delegate` change required an assignment addition (is this an intentional groovy 5 change?)
2. why the hibernate domain object is unwrapping accessing .id only.

jdaugherty:
The subsequent failures may still related to the `@Delegate` change, so there's likely more to do there.

jdaugherty:
The params are fixed locally.  I'm hopeful only the hibernate issue remains.

jdaugherty:
I figured out the cause for all of these and will push fixes later today.

jdaugherty:
The last issue is the proxy issue found by the redis test.  I'm researching this more.

jdaugherty:
It was only by side effect that I caught the proxy issues.  I'm going to fix the Redis test by just switching the `@Rollback` to the grails one, which will have a session.  I'll address the proxy issue in a separate PR since it's a behavior change.

jdaugherty:
https://github.com/apache/grails-core/pull/15841 is the PR for the hibernate proxy change that this PR found.

jdaugherty:
@jamesfredley I removed the opt-in for atCheckWaiting and included it by default on this PR. this should resolve a lot of the flaky test behavior you've seen.

jdaugherty:
I've made several more changes to help the flakiness of our tests and turned on reporting for the ones that have see repeat failures so we can debug them more if they continue to occur.

jdaugherty:
For convenience, here's the link that diffs the material changes built on top of the 7.2.x changes https://github.com/apache/grails-core/compare/92e60a0be04c48feedf6da1c812d9fe5ad8393ef...19ac8b275713a063f417ffb1001148680b8022ae

sbglasius (approved):
Reviewed changes in https://github.com/apache/grails-core/pull/15834/changes#diff-1856901f8b015db14695a910ac0e8a8a6e34bc8de5da05a0df5ec2388ff7e134

Looks sane to me.

jdaugherty:
@jamesfredley since this is a merge up, I'm going to go ahead and merge this. We can iterate on the 8.x branch if you feel I've done something incorrectly here.  I need these changes merged to fix the latest round of the patches in the 8.x branch.

PR #15916, opened 2026-07-07
Make i18n and URL-mappings framework beans overridable (@ConditionalOnMissingBean)

codeconsole:
### Summary

Guards several Grails framework beans with `@ConditionalOnMissingBean` so an application- or plugin-defined bean takes precedence cleanly, instead of relying on bean-definition overriding (which produces a startup warning, or fails outright when `spring.main.allow-bean-definition-overriding=false`):

- **grails-i18n** (`I18nAutoConfiguration`): `localeResolver`, `localeChangeInterceptor`, `messageSource`
- **grails-url-mappings** (`UrlMappingsAutoConfiguration`): `grailsCorsFilter`, `urlMappingsErrorPageCustomizer`, `urlMappingsInfoHandlerAdapter`

All beans still register by default — this only changes what happens when the user defines their own.

### Depends on #13863

This PR is stacked on #13863 (removal of the auto-injected `@EnableWebMvc`) and includes its commits; merge that first, after which this diff reduces to the changes described above.

The dependency is real, not just mechanical: with `@EnableWebMvc` present, Spring's `WebMvcConfigurationSupport` contributes its own `localeResolver` bean, which would win the `@ConditionalOnMissingBean` race and silently replace Grails' `SessionLocaleResolver`. With #13863 in, `I18nAutoConfiguration` is ordered before `WebMvcAutoConfiguration`, so the Grails bean registers first and Boot's conditional bean defers to it. This is what #15751 attempted and could not achieve standalone.

### Context

Extracted from #15755 as an independently useful piece while the plugin-lifecycle discussion there is resolved (see the review discussion about unifying bean registration ahead of auto-configuration).

### Testing

New auto-configuration specs for both modules covering default registration, the `grails.cors.filter` toggle, and user-bean back-off for every guarded bean. Both modules' full test suites and checkstyle/CodeNarc pass.


grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nAutoConfiguration.java
```diff
@@ -56,18 +57,21 @@ public class I18nAutoConfiguration {
...
     @Bean
+    @ConditionalOnMissingBean(name = "localeChangeInterceptor")
     public LocaleChangeInterceptor localeChangeInterceptor() {
         ParamsAwareLocaleChangeInterceptor localeChangeInterceptor = new ParamsAwareLocaleChangeInterceptor();
         localeChangeInterceptor.setParamName("lang");
         return localeChangeInterceptor;
     }
 
     @Bean(AbstractApplicationContext.MESSAGE_SOURCE_BEAN_NAME)
+    @ConditionalOnMissingBean(name = AbstractApplicationContext.MESSAGE_SOURCE_BEAN_NAME)
```
jdaugherty: Consider `@ConditionalOnMissingBean(name = AbstractApplicationContext.MESSAGE_SOURCE_BEAN_NAME, search = SearchStrategy.CURRENT)` here. Boot's own `MessageSourceAutoConfiguration` scopes its equivalent guard to `SearchStrategy.CURRENT` specifically because a `messageSource` is commonly present in a *parent* context, and each child context should still get its own. With the default search strategy (ancestors included), any hierarchical deployment would cause the Grails `PluginAwareResourceBundleMessageSource` to silently back off in favor of the parent's plain message source, losing plugin message-bundle resolution. The same consideration applies (more weakly) to `localeResolver` and `localeChangeInterceptor` above, but `messageSource` is the one with a real hierarchy precedent.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -1346,3 +1346,61 @@ grails:
...
+With `WebMvcAutoConfiguration` active, Boot would otherwise register its own `RequestContextFilter` that rebinds a plain `ServletRequestAttributes`, replacing the `GrailsWebRequest` bound earlier in the filter chain.
+Downstream code that expects a `GrailsWebRequest` would then fail with a `ClassCastException`.
+
+Grails 8 registers its request-binding filter as a `RequestContextFilter` bean so that Boot's filter backs off and the `GrailsWebRequest` remains bound for the whole request.
+This is handled internally and requires no configuration changes.
+Applications that defined their own `GrailsWebRequestFilter` bean, or their own `grailsWebRequestFilter` filter-registration bean, continue to override the Grails-provided ones.
+
+==== 32. More Framework Beans Are Cleanly Overridable
+
+Several Grails framework beans that were previously registered unconditionally are now guarded with `@ConditionalOnMissingBean`, so an application- or plugin-defined bean of the same name (or type) takes precedence cleanly instead of relying on bean-definition overriding:
```
jdaugherty: The "or plugin-defined" claim is stronger than what `@ConditionalOnMissingBean` can deliver. Beans contributed by a plugin's `doWithSpring` (or an app's `resources.groovy`) are registered by `GrailsApplicationPostProcessor`, a plain (unordered) `BeanDefinitionRegistryPostProcessor` that runs *after* `ConfigurationClassPostProcessor` has already evaluated auto-configuration conditions. Those beans therefore still land via bean-definition overriding, exactly as before — the conditional only sees beans defined in `@Configuration` classes and other auto-configurations. Suggest qualifying this so users don't expect `doWithSpring` beans to trigger the back-off (e.g. "a bean defined in application configuration classes"), or explicitly noting that `doWithSpring`/`resources.groovy` beans continue to rely on overriding.

grails-url-mappings/src/main/groovy/org/grails/plugins/web/mapping/UrlMappingsAutoConfiguration.java
```diff
@@ -76,19 +76,22 @@ public LinkGenerator grailsLinkGenerator() {
     }
 
     @Bean
+    @ConditionalOnMissingBean(GrailsCorsFilter.class)
```
jdaugherty: Question on guard choice: this is type-based on `GrailsCorsFilter`, so a user who replaces CORS handling with a plain Spring `CorsFilter` (or any non-`GrailsCorsFilter` registration) still gets the Grails filter registered alongside theirs — they'd have to know to set `grails.cors.filter=false` as well. If the intent is only "a custom `GrailsCorsFilter` wins", this is fine as-is; if the intent is "any user CORS filter wins", `@ConditionalOnMissingBean(CorsFilter.class)` would be broader. Worth a deliberate choice either way (the other two guards below are unambiguous since the types are Grails-specific).

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsViewResolverAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+ * <p>Disable with {@code grails.web.removeDefaultViewResolverBean=false}. The earlier
+ * {@code spring.gsp.removeDefaultViewResolverBean} (when removal lived in grails-gsp) is still
+ * honoured for backward compatibility, but is deprecated.
+ */
+@AutoConfiguration(after = WebMvcAutoConfiguration.class)
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+@Import(GrailsViewResolverAutoConfiguration.RemoveDefaultViewResolverRegistrar.class)
+public class GrailsViewResolverAutoConfiguration {
+
+    static final String REMOVE_PROPERTY = "grails.web.removeDefaultViewResolverBean";
```
jdaugherty: `grails.web.removeDefaultViewResolverBean` is now a documented, user-facing configuration property — consider declaring the constant in `grails.config.Settings` alongside the other `grails.web.*` keys instead of a package-private field here, for discoverability and consistency with how the rest of the module references settings (`Settings.WEB_SERVLET_PATH`, etc.).

grails-i18n/build.gradle
```diff
@@ -69,4 +71,5 @@ dependencies {
 
 apply {
     from rootProject.layout.projectDirectory.file('gradle/docs-config.gradle')
+    from rootProject.layout.projectDirectory.file('gradle/test-config.gradle')
 }
```
jdaugherty: Nit: file still ends without a trailing newline (same in `grails-url-mappings/build.gradle`). Worth fixing while touching these files.

jdaugherty (approved):
I don't see anything immediately wrong with this. I merged the latest 8.0.x branch into this one to see if there are any notable test failures.  I also asked AI to review it and it some minor comments that we should probably discuss.  Overall, approving.  We should wait to merge it until we have a second review though.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nAutoConfiguration.java
codeconsole: Applied — and to all three beans, not just `messageSource`. `DispatcherServlet` resolves `localeResolver`/`localeChangeInterceptor` from its own context, so the parent-suppression problem applies with the same force there: a child context whose bean backed off would fall back to `DispatcherServlet`'s default `AcceptHeaderLocaleResolver`, not the parent's bean. Also verified Boot 4 no longer contributes a WebMvc `localeResolver` at all (it did in Boot 3), so `SearchStrategy.CURRENT` has no auto-config interplay to worry about. Added a parent-context test covering all three.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Good catch — qualified. The section now scopes the clean back-off to `@Bean` methods in application configuration classes, and a NOTE spells out that `doWithSpring`/`resources.groovy` beans continue to take effect through bean-definition overriding, exactly as before. (Closing that gap is what #15934 does — once the retimed `doWithSpring` drain registers plugin beans before the conditions are evaluated, this note can be revisited.)

grails-url-mappings/src/main/groovy/org/grails/plugins/web/mapping/UrlMappingsAutoConfiguration.java
codeconsole: Made the deliberate choice: broadened to `@ConditionalOnMissingBean(CorsFilter.class)`. The intent is "any user CORS filter wins" — two CORS filters answering the same requests is never desirable, and since `GrailsCorsFilter extends CorsFilter` a custom Grails filter still triggers the back-off. The reasoning is documented in a comment at the guard, and there are now tests for both the plain-`CorsFilter` and custom-`GrailsCorsFilter` cases.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsViewResolverAutoConfiguration.java
codeconsole: Done — `Settings.WEB_REMOVE_DEFAULT_VIEW_RESOLVER_BEAN`, declared next to the other `grails.web.*` keys. The constant lives on the base branch (#13863, commit 84ea8179b8) since that PR introduces the property; this branch merges it and references the constant.

grails-i18n/build.gradle
codeconsole: Fixed, both files.

jdaugherty (approved):

sbglasius (approved):

PR #15934, opened 2026-07-07
Register plugin beans before Spring Boot auto-configuration by retiming doWithSpring

codeconsole:
### Summary

Supersedes #15755, implementing the design agreed in its review discussion: instead of adding a new lifecycle method, the **existing** plugin entry point is retimed. Completes the follow-up scoped out of #15409 ("we could move plugins completely to the autoconfiguration workflow") and delivers the visible half of #14915.

- **`doWithSpring`/`doWithRuntimeConfiguration` now runs before Spring Boot auto-configuration.** An `ApplicationContextInitializer` adds a manually-registered `BeanDefinitionRegistryPostProcessor`, which Spring runs ahead of `ConfigurationClassPostProcessor`. Plugin beans are in the registry before auto-configuration conditions are evaluated, so Boot's `@ConditionalOnMissingBean` beans defer to plugin beans — no more overriding Boot beans after the fact.
- **A single plugin manager and `GrailsApplication`**, built once in the early phase from `PluginDiscovery` (the 7.x bootstrap infrastructure) and promoted as singletons that `GrailsApplicationPostProcessor` reuses — so plugins are loaded by one manager pass rather than by a throwaway pass followed by the real one (the superseded #15755 approach). (Grails still constructs each plugin's `GrailsClass` reference instance plus its real instance, as it always has.)
- **Artefact discovery moves into the same early phase** (resolving application classes via the same `classes()` path as before, honoring the compile-time `packageNames()` injection), so `doWithSpring` closures that iterate artefacts — controllers, services, interceptors — work unmodified.
- **`BeanRegistrar` is the new native registration API for plugins**: `beanRegistrar()` on `GrailsApplicationLifeCycle`/`Plugin`, applied pre-refresh through Spring's own `BeanRegistryAdapter` (the mechanism `GenericApplicationContext.register(BeanRegistrar...)` uses). It is a functional interface, so a plugin can return a closure coerced with `as BeanRegistrar` — no separate class needed. AOT-friendly by design.
- **The bean DSL entry point is deprecated** (`@Deprecated(since = '8.0')` on `doWithSpring`) pointing at `beanRegistrar()`. Deprecated DSL closures keep draining through the existing `RuntimeSpringConfiguration` path — just earlier — until removal. The DSL internals are deliberately *not* reimplemented on `BeanRegistry`: the DSL supports constructs with no clean equivalent (parent/abstract beans, late-resolved references, `BeanConfiguration` post-processing), the plumbing is invisible to users, and that translation work would be deleted at DSL removal anyway.

`resources.groovy`/`resources.xml` and the application class's own `doWithSpring` drain where they always did, so application beans still override plugin beans under the default `allow-bean-definition-overriding=true`. The one behavioral edge — apps explicitly setting it `false` now get `BeanDefinitionOverrideException` where the old single-drain merge was silent — is documented in the upgrade notes.

### Regressions found and fixed by verification

The retiming sweep caught two real breakages, both fixed generally in this PR:

1. **OSIV silently disabled** (differential-tested: passes on 8.0.x, failed on the branch before the fix): the Hibernate/Mongo datastore initializers detected "web application" via `containsBeanDefinition('dispatcherServlet')` — an auto-configuration bean, always absent at the early drain. `AbstractDatastoreInitializer.isWebApplicationRegistry()` now also does a registration-order-independent `WebApplicationContext` type check. Follow-up: `Neo4jDataStoreSpringInitializer` needs the identical one-line change in its own build (separate PR).
2. **Controllers in sibling packages lost their beans**: early artefact discovery initially scanned only the Application class's package, missing the injected `packageNames()` override. Fixed by resolving application classes through `classes()` for exact parity with the previous behavior.

### Verification

All green, zero failures across grails-core, grails-test-suite-uber, grails-test-suite-web, grails-gsp, grails-controllers, grails-web-boot, sitemesh3, mongodb-core (testcontainers), hibernate5/7, plus the full hibernate7 `app1:integrationTest` and `app1:integrationTest` functional suites.

Test coverage includes: ordering specs proving a plugin `doWithSpring` bean and a plugin `beanRegistrar()` bean each win a `@ConditionalOnMissingBean` race (with a no-early-phase control) and that the `beanRegistrar()` closure form works; app-over-plugin override under the default settings, and a `BeanDefinitionOverrideException` when overriding is disabled; a single-manager-pass check (no throwaway extra load); the initializing flag resetting when the early phase throws; an end-to-end app3 integration test exercising both `doWithSpring` and `beanRegistrar()` in a real booted application; Hibernate and MongoDB OSIV registration; and the actuator `/env` and `/health` endpoints serializing to JSON. Neo4j's suite and the MongoDB example apps run in their dedicated CI jobs (independent Gradle build / require a Mongo instance); noted above.

Docs: new upgrade-notes section covering the retiming, the override edge, and the DSL deprecation; a "what's new" entry; the plugin guide's runtime-configuration page documents `beanRegistrar()` first (concise closure form) with `doWithSpring` as the deprecated-but-working alternative.

Related: #15755 (superseded), #15409 / #15408 (7.x groundwork), #14915 (bean DSL deprecation ticket), #13863 / #15916 (independent extractions from #15755).


jdaugherty:
I merged the latest 8.x branch changes, apologies for the test failures initially.  They were due to merging up 7 -> 8 + the vulnerability fixes + the spring boot 4 vs 3 mismatches.  Will review this after the tests are run.  Any failures related to the merge, I'll try to fix and merge into this PR

grails-core/src/main/groovy/grails/boot/config/ApplicationClassScanner.groovy
```diff
@@ -0,0 +1,85 @@
...
+ * Discovers the classes that constitute a Grails application by scanning the classpath relative
+ * to an application class. This is the single implementation of the default scanning performed by
+ * {@link GrailsAutoConfiguration#classes()}, also used by
+ * {@link GrailsEarlyPluginRegistrationPostProcessor} to perform artefact discovery before Spring
+ * Boot auto-configuration is processed.
+ *
+ * @since 8.0
+ */
+@CompileStatic
+final class ApplicationClassScanner {
```
jdaugherty: This scanner implies it finds the Application class but we already have the FindmainClass for that so we should rename this to reflect its purpose.

grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
```diff
@@ -194,8 +240,11 @@ class GrailsApplicationPostProcessor implements BeanDefinitionRegistryPostProces
         def application = grailsApplication
         Holders.setGrailsApplication(application)
 
-        // first register plugin beans
-        pluginManager.doRuntimeConfiguration(springConfig)
+        if (!earlyPluginRegistrationRan) {
+            // first register plugin beans; when the early phase ran they were
+            // already drained into the registry ahead of auto-configuration
+            pluginManager.doRuntimeConfiguration(springConfig)
```
jdaugherty, line 248: When the early phase did **not** run (no promoted `PluginDiscovery` — unit-test slices, contexts not booted through `GrailsApp`/the bootstrap registry), this fallback drains only `doWithSpring` via `doRuntimeConfiguration`. Plugin `beanRegistrar()`s are never applied on this path — the only registrar applied here is the application's own (`lifeCycle.beanRegistrar()` below). A plugin that migrates to the new recommended API would silently lose its beans in any context that takes this fallback, which is a nasty asymmetry with the deprecated DSL that still works everywhere. Suggest also iterating `pluginManager.allPlugins` here and applying each plugin's `getBeanRegistrar()` (same filtering as `GrailsEarlyPluginRegistrationPostProcessor.applyBeanRegistrars`), so plugin registrars behave identically on both paths.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) throws BeansException {
+        // Check the LOCAL singleton only — a parent context's discovery must not cause us to re-run
+        // the early phase in a child context (containsBean/getBean would delegate to the parent).
+        Object discovery = applicationContext.getBeanFactory().getSingleton(PluginDiscovery.BEAN_NAME);
+        if (!(discovery instanceof PluginDiscovery pluginDiscovery)) {
+            // No plugin discovery promoted to this context (e.g. unit-test slice) — nothing to do.
+            return;
+        }
+
+        Environment.setInitializing(true);
```
jdaugherty, line 126: If anything in this phase throws (`loadPlugins()`, a plugin's `doWithSpring`, a registrar), or the refresh fails later for any reason, `Environment.setInitializing(true)` leaks — the reset listener is only added at the very end of this method and only fires on a *successful* refresh. Since the flag is a system property, a single failed context poisons every subsequent context in the same JVM (test forks especially). The same pattern pre-exists in `GrailsApplicationPostProcessor`, but this phase runs in more contexts and does much more work under the flag. Suggest wrapping the body in try/catch that calls `Environment.setInitializing(false)` before rethrowing (the success path still resets via the refresh listener).

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+     * in other packages. The instance created here is used solely to compute the scan; the
+     * lifecycle bean the application interacts with is still created by Spring later.
+     */
+    private Collection<Class> scanApplicationSource(Class<?> source) {
+        if (GrailsAutoConfiguration.class.isAssignableFrom(source)) {
+            try {
+                GrailsAutoConfiguration application = (GrailsAutoConfiguration) source.getDeclaredConstructor().newInstance();
+                application.setApplicationContext(applicationContext);
+                return application.classes();
+            } catch (Throwable e) {
```
jdaugherty: Two notes on this method:

1. It instantiates the user's `Application` class a second time (once here for the scan, once later as the Spring bean). Constructor side effects — static init aside — now execute twice per boot, and `setApplicationContext` is invoked on a throwaway instance. Probably acceptable, but worth calling out explicitly in the upgrade notes since application-class constructors doing work is not unheard of.
2. `catch (Throwable)` also swallows `OutOfMemoryError`/`StackOverflowError` into the fallback scan. `catch (Exception | LinkageError e)` would cover the realistic failure modes (missing no-arg ctor, class-init failures) without masking fatal errors.

grails-core/src/main/groovy/grails/boot/GrailsApp.groovy
```diff
@@ -142,6 +143,22 @@ class GrailsApp extends SpringApplication {
...
+    /**
+     * Stashes the application source classes as a well-known singleton so that
+     * {@code GrailsEarlyPluginRegistrationPostProcessor} can perform artefact discovery before
+     * Spring Boot auto-configuration is processed. Runs before the context initializers are
+     * applied, so the singleton is available by the time the early registration phase executes.
+     */
+    @Override
+    protected void postProcessApplicationContext(ConfigurableApplicationContext applicationContext) {
+        super.postProcessApplicationContext(applicationContext)
+        Class<?>[] sourceClasses = getAllSources().findAll { it instanceof Class } as Class<?>[]
```
jdaugherty: `getAllSources()` can contain `String` class names (`SpringApplication.setSources` / `spring.main.sources`), which are dropped here. If the application class is supplied as a String while any other source is a `Class`, the stash is non-empty, so `resolveApplicationSourceClasses` never falls back to the registry scan — and the application class silently misses early artefact discovery (its controllers/services get no plugin beans). Suggest resolving String sources to classes here as well, or making the fallback in the post-processor also engage when the stash contains no `GrailsApplicationClass`.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+
+        RuntimeSpringConfiguration springConfig = new DefaultRuntimeSpringConfiguration();
+        pluginManager.doRuntimeConfiguration(springConfig);
+        springConfig.registerBeansWithRegistry(registry);
+        applyBeanRegistrars(pluginManager, registry);
+
+        ConfigurableListableBeanFactory beanFactory = applicationContext.getBeanFactory();
+        beanFactory.registerSingleton(GrailsApplication.APPLICATION_ID, grailsApplication);
+        beanFactory.registerSingleton(GrailsPluginManager.BEAN_NAME, pluginManager);
+        beanFactory.registerSingleton(EARLY_REGISTRATION_COMPLETE_BEAN_NAME, Boolean.TRUE);
```
jdaugherty: Minor: the marker (`Boolean`) and the `Class<?>[]` stash registered by `GrailsApp` remain manual singletons for the life of the context, so they show up as autowire-by-type candidates (`getBeansOfType(Boolean)` etc.). The marker has to survive until `GrailsApplicationPostProcessor` checks it, but the source-classes stash could be `destroySingleton`-ed once resolved to avoid leaking it into the context.

grails-core/src/main/resources/META-INF/spring.factories
```diff
@@ -22,4 +22,5 @@ org.springframework.boot.env.PropertySourceLoader=\
   org.grails.config.yaml.YamlPropertySourceLoader
 org.springframework.boot.EnvironmentPostProcessor=grails.boot.config.GrailsEnvironmentPostProcessor
 org.springframework.boot.SpringApplicationRunListener=grails.config.external.ExternalConfigRunListener
-org.springframework.boot.bootstrap.BootstrapRegistryInitializer=org.apache.grails.core.GrailsBootstrapRegistryInitializer
\ No newline at end of file
+org.springframework.boot.bootstrap.BootstrapRegistryInitializer=org.apache.grails.core.GrailsBootstrapRegistryInitializer
```
jdaugherty, line 25: Nit: still no trailing newline at EOF (this diff touched the last line — easy moment to fix it).

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
```diff
@@ -105,10 +106,29 @@ abstract class Plugin implements GrailsApplicationLifeCycle, GrailsApplicationAw
...
+     *
+     * <p>The returned registrar is applied before Spring Boot auto-configuration is processed, so
+     * beans registered here take precedence over Boot's {@code @ConditionalOnMissingBean} defaults.</p>
+     *
+     * @return A {@link org.springframework.beans.factory.BeanRegistrar} that registers beans,
+     * or {@code null} if none (the default)
+     * @since 8.0
+     */
+    @Override
+    BeanRegistrar beanRegistrar() { null }
```
jdaugherty, line 131: Why not initialize the bean registry and pass it as a method argument?  It's effectively a functional interface already and I'd assume people would just have the closure be that implementation.

grails-core/src/main/groovy/grails/core/DefaultGrailsApplication.java
```diff
@@ -219,6 +219,18 @@ public GrailsApplicationClass getApplicationClass() {
...
 
+    /**
+     * Sets the application class. Used to adopt the application class when this instance was
+     * constructed before the {@link GrailsApplicationClass} was available, e.g. during early
+     * plugin registration ahead of Spring Boot auto-configuration.
+     *
+     * @param applicationClass The application class
+     * @since 8.0
+     */
+    public void setApplicationClass(GrailsApplicationClass applicationClass) {
```
jdaugherty, line 232: What prevents someone calling this after initialize?  I think this has to be guarded so it can only be set at certain points in the life cycle.

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
```diff
@@ -105,10 +106,29 @@ abstract class Plugin implements GrailsApplicationLifeCycle, GrailsApplicationAw
      * Sub classes should override to provide implementations
      *
      * @return A closure that defines beans to be executed by Spring
+     * @deprecated since 8.0 in favour of {@link #beanRegistrar()}. The bean builder DSL continues
```
jdaugherty: I would be stronger in the language here - Spring has said the bean dsl code will remain, but it won't be supported.  If there's an issue, they basically won't fix it.  So it's strongly urged to migrate.

grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
```diff
@@ -122,11 +148,31 @@ class GrailsApplicationPostProcessor implements BeanDefinitionRegistryPostProces
...
+     * application classes it could not resolve (e.g. a customized {@code classes()} implementation)
+     * still need to be registered.
+     */
+    private void registerRemainingApplicationClasses() {
+        for (cls in classes) {
+            if (!grailsApplication.isArtefact(cls)) {
+                grailsApplication.addArtefact(cls)
+            }
+        }
     }
```
jdaugherty, line 178: Minor perf: `isArtefact(cls)` does a linear name-comparison scan over `allArtefactClasses`, so this loop is O(classes x registeredArtefacts) on every boot that took the early path — for a large app that's a quadratic pass over the full class list. Building a `Set` of registered class names once before the loop keeps it linear:

```suggestion
    private void registerRemainingApplicationClasses() {
        Set<String> registeredClassNames = grailsApplication.allArtefacts*.name as Set<String>
        for (cls in classes) {
            if (!registeredClassNames.contains(cls.name)) {
                grailsApplication.addArtefact(cls)
            }
        }
    }
```

grails-doc/src/en/guide/plugins/hookingIntoRuntimeConfiguration.adoc
```diff
@@ -23,7 +23,52 @@ Grails provides a number of hooks to leverage the different parts of the system
 ==== Hooking into the Grails Spring configuration
 
 
-First, you can hook in Grails runtime configuration overriding the `doWithSpring` method from the link:{api}grails/plugins/Plugin.html[Plugin] class and returning a closure that defines additional beans. For example the following snippet is from one of the core Grails plugins that provides link:i18n.html[i18n] support:
+The recommended way to register beans from a plugin is to override the `beanRegistrar` method from the link:{api}grails/plugins/Plugin.html[Plugin] class and return a Spring Framework {springapi}org/springframework/beans/factory/BeanRegistrar.html[BeanRegistrar]:
```
jdaugherty: I think we need to mention this in the what's new as well.

grails-doc/src/en/guide/plugins/hookingIntoRuntimeConfiguration.adoc
```diff
@@ -23,7 +23,52 @@ Grails provides a number of hooks to leverage the different parts of the system
...
+import org.springframework.context.support.ReloadableResourceBundleMessageSource
+import grails.plugins.Plugin
+
+class I18nGrailsPlugin extends Plugin {
+
+    def version = "0.1"
+
+    @Override
+    BeanRegistrar beanRegistrar() {
+        new I18nBeanRegistrar()
```
jdaugherty: Since only 1 registrar can be returned, wha'ts the benefit here?  It seems like it's a lot more boiler plate.  if we built our own registrar and passed in, people could call just call registerBean, yes?

jdaugherty (commented):
I made a first pass at this review.  I think this is the right direction overall.

Can we also add a test example that explicitly exercises the new registrar functionality?

grails-core/src/main/groovy/grails/boot/config/ApplicationClassScanner.groovy
codeconsole: Done — renamed to `ApplicationArtefactScanner` (it scans the application's *artefact* classes; the single Application/main class is found by `FindMainClass`).

grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
codeconsole: Good catch — fixed. The `GrailsApplicationPostProcessor` fallback path (contexts not booted through `GrailsApp`, e.g. unit-test slices) now applies each enabled plugin's `beanRegistrar()` too, with the same scope/environment filtering the early phase uses, after the DSL flush so registrars win DSL name conflicts. A plugin's `beanRegistrar()` now behaves identically on both paths.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
codeconsole: Fixed. The early phase wraps its body in try/catch and resets `Environment.setInitializing(false)` before rethrowing, so a failed context no longer poisons later contexts in the same JVM (test forks). The success path still resets via the refresh listener. (The pre-existing `GrailsApplicationPostProcessor` has the same shape — worth the same treatment as a follow-up.)

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
codeconsole: Both addressed: the scan catch is now `Exception | LinkageError` (no longer swallows `OutOfMemoryError`/`StackOverflowError`), and the upgrade notes call out that the application class is instantiated a second time for the early scan (a throwaway instance; the real bean is still created by Spring), so application-class constructors should stay side-effect free.

grails-core/src/main/groovy/grails/boot/GrailsApp.groovy
codeconsole: Fixed both ends: `GrailsApp` now resolves `String` sources (`spring.main.sources`) to classes when stashing, and the early phase recovers a `GrailsApplicationClass` from the registry whenever the stash contains none — so a String-specified application class is still discovered even alongside `Class` sources.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
codeconsole: The `Class[]` source-classes stash is now `destroySingleton`-ed once consumed, so it no longer lingers as an autowire-by-type candidate. The completion marker stays because `GrailsApplicationPostProcessor` reads it later (in its constructor) to decide whether to reuse the promoted singletons.

grails-core/src/main/resources/META-INF/spring.factories
codeconsole: Fixed.

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
codeconsole: Kept `beanRegistrar()` returning `BeanRegistrar` deliberately: it's Spring's own registration abstraction (so plugin authors learn a concept that transfers to plain Spring), the returned registrar is a first-class object (unit-testable, reusable, composable with Spring's `@Import`/AOT machinery), and it mirrors `doWithSpring` returning a value. You're right that the separate-class form is boilerplate, though — and `BeanRegistrar` is a functional interface, so a coerced closure is all that's needed. I've reworked the docs to lead with that concise form:

```groovy
BeanRegistrar beanRegistrar() {
    { BeanRegistry registry, Environment env ->
        registry.registerBean('myService', MyServiceImpl)
    } as BeanRegistrar
}
```

and added a test proving it compiles and works — that's essentially the closure you were expecting. If you'd still rather the method receive the `BeanRegistry` directly (`void doWithBeanRegistry(registry, env)`), I'm happy to switch it — I just held off changing the public signature unilaterally.

grails-core/src/main/groovy/grails/core/DefaultGrailsApplication.java
codeconsole: Guarded as set-once: it throws `IllegalStateException` on reassignment to a different class (idempotent for the same value). It's deliberately *not* guarded on `initialised` — the reuse path legitimately adopts the class after the early phase has already built and initialised the application without it, so a "not after initialise" guard would break that. Set-once prevents the real risk (silent reassignment). Test added.

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
codeconsole: Strengthened on both `GrailsApplicationLifeCycle.doWithSpring` and `Plugin.doWithSpring`: the DSL remains available but is no longer actively supported and won't receive fixes for new issues; users are strongly urged to migrate.

grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
codeconsole: Applied your suggestion — builds a `Set` of registered artefact names once, then a `contains` check, so it's linear again.

grails-doc/src/en/guide/plugins/hookingIntoRuntimeConfiguration.adoc
codeconsole: Added a "Plugin Beans Register Before Spring Boot Auto-Configuration" section to the what's-new guide covering the retimed lifecycle and the `beanRegistrar()` API.

grails-doc/src/en/guide/plugins/hookingIntoRuntimeConfiguration.adoc
codeconsole: Since `BeanRegistrar` is a functional interface, only a coerced closure is needed — no separate class and no real boilerplate. I've reworked the docs to lead with that form (see the reply on `beanRegistrar()` in `Plugin` above):

```groovy
BeanRegistrar beanRegistrar() {
    { BeanRegistry registry, Environment env ->
        registry.registerBean('myService', MyServiceImpl)
    } as BeanRegistrar
}
```

Returning a single `BeanRegistrar` is not a limitation — a registrar can register any number of beans, and returning one mirrors `doWithSpring` returning one closure. Kept the return type because it's Spring's native abstraction and a reusable first-class object; happy to switch to passing the `BeanRegistry` in if you'd prefer.

codeconsole:
> Can we also add a test example that explicitly exercises the new registrar functionality?

Added — the `loadafter` plugin now also registers a bean through `beanRegistrar()` (using the concise closure-coercion form), app3's `Application` declares a matching `@ConditionalOnMissingBean` default, and `PluginBeansBeforeAutoConfigurationSpec` asserts the registrar's bean wins in a real boot. It passes alongside the existing `doWithSpring` case. There's also unit coverage in `EarlyPluginRegistrationOrderingSpec` (a `beanRegistrar()` bean winning a `@ConditionalOnMissingBean` race, and the coerced-closure form).

grails-core/src/main/groovy/grails/plugins/Plugin.groovy
jdaugherty: I'm happy with this compromise.  Hopefully others are too.

jdaugherty (approved):

borinquenkid:
**Note:** This comment was generated by an AI coding assistant (Claude), directed by a human reviewer to verify that this PR's tests actually cover the situations described in its own description and commit messages. Findings below are grounded in specific file/line citations from the diff and the `8.0.x` base branch, not inference.

Overall, the core architectural claims are well covered: `EarlyPluginRegistrationOrderingSpec` and app3's `PluginBeansBeforeAutoConfigurationSpec` genuinely prove plugin `doWithSpring`/`beanRegistrar()` beans beat `@ConditionalOnMissingBean` Boot defaults, at both unit and true end-to-end level (confirmed `app3` actually depends on the `loadafter` plugin). The two headline regressions — "OSIV silently disabled" and "sibling-package controllers losing beans" — are also genuinely pinned: `OpenSessionInViewSpec` in both `hibernate5/app1` and `hibernate7/app1`, and the pre-existing `UriMatchingInterceptorFunctionalSpec`, which really does exercise a controller (`demo.InterceptorDemoController`) living outside the `Application` class's own package (`functionaltests`).

A few gaps, in rough priority order:

1. **Mongo OSIV fix is untested.** `MongoDbDataStoreSpringInitializer` gets the identical `isWebApplicationRegistry()` swap that fixed the Hibernate OSIV regression (`MongoDbDataStoreSpringInitializer.groovy:129-130`), but no test anywhere asserts `mongoOpenSessionInViewInterceptor` registers in a web app — the same bug class just proven to silently break Hibernate.
2. **`/actuator/health` half of the actuator regression is untested.** Commit `dd5ac787` states both `/actuator/env` and `/actuator/health` failed with `HttpMessageNotWritableException`, but `ActuatorEnvClosureSpec` only requests `/actuator/env` (lines 40, 48).
3. **`BeanDefinitionOverrideException` edge case (new upgrade-notes section) has no test.** The closest existing spec, `GrailsAppContextOverridingSpec`, only checks that `allow-bean-definition-overriding=false` propagates to the bean factory flag — it never registers a colliding plugin+app bean pair or asserts the exception.
4. **"Built/instantiated exactly once" claim is checked by existence, not count.** `EarlyPluginRegistrationOrderingSpec` asserts the `grailsApplication`/`pluginManager` singletons exist (or don't, in the control case), but nothing spies on construction to prove single instantiation.
5. **Flag-reset-on-exception path is unexercised.** `GrailsEarlyPluginRegistrationPostProcessor`'s `catch (RuntimeException | Error e) { Environment.setInitializing(false); }` branch has no test that makes the early phase throw and then checks the flag resets — only the successful-refresh reset is asserted.

Flagging 1-2 as the highest priority since they mirror a regression class this PR already proved happened once.

codeconsole:
Thanks — this audit was accurate and genuinely useful; all five gaps are now addressed in e737052.

1. **Mongo OSIV** — Added `OpenSessionInViewSpec` to `grails-test-examples/mongodb/base`, mirroring the Hibernate one, asserting `mongoOpenSessionInViewInterceptor` registers in a web app. It exercises the same `isWebApplicationRegistry()` fix. I couldn't run it green locally — the entire mongodb/base integration suite can't reach Mongo in my environment (the existing `BookSpec` fails with the identical `MongoTimeoutException` against a local container), so it will run in the MongoDB CI job alongside the others.
2. **/actuator/health** — Added a `/actuator/health` assertion to `ActuatorEnvClosureSpec` (passes locally). It goes through the same MVC JSON-converter path the `/env` regression exposed.
3. **`BeanDefinitionOverrideException` edge** — Added a test that sets `allow-bean-definition-overriding=false`, registers a colliding plugin+application bean pair, and asserts the exception (passes).
4. **"Built exactly once"** — Good catch, and it surfaced a real inaccuracy: adding a construction counter showed a plugin is instantiated **twice**, not once — but that's pre-existing Grails behavior (`DefaultGrailsPlugin` builds a `GrailsPluginClass` reference instance *and* the real instance), unrelated to this PR. What the retiming actually guarantees is a *single plugin-manager pass* rather than a throwaway pass plus the real one (the rejected earlier approach, which would have multiplied that count). I reworded the test to assert exactly that (early-phase count equals a single bare `loadPlugins()`), and corrected the upgrade note that claimed "plugins are instantiated once."
5. **Flag-reset-on-exception** — Added a test with a plugin whose `doWithSpring` throws during the early phase, asserting `Environment.isInitializing()` is `false` afterward (passes) — directly exercising the `catch` branch.

grails-core full suite, the ordering spec (all 9 cases), and the issue-10279 actuator spec pass locally; the mongo spec is the only one gated on CI infrastructure.

borinquenkid:
**Note:** This comment was generated by an AI coding assistant (Claude), directed by a human reviewer, as a follow-up to the coverage audit above.

All 5 gaps from the audit are addressed as of e737052, and confirmed locally:

1. **Mongo OSIV** — `OpenSessionInViewSpec` (mongodb/base) added.
2. **`/actuator/health`** — assertion added to `ActuatorEnvClosureSpec`, passes locally.
3. **`BeanDefinitionOverrideException` edge** — new test in `EarlyPluginRegistrationOrderingSpec`, passes locally.
4. **"Built exactly once"** — reworded to assert a single manager pass (not a throwaway-then-real double pass); the original wording was corrected in the upgrade notes since a plugin is legitimately instantiated twice (reference `GrailsClass` + real instance, pre-existing behavior). Passes locally.
5. **Flag-reset-on-exception** — new test with a throwing `doWithSpring`, asserts `Environment.isInitializing()` resets. Passes locally.

On top of that, `d054328fe1` adds a second, unit-testable version of the Mongo OSIV coverage in `grails-data-mongodb/core` (`MongoOpenSessionInViewSpec`), alongside the existing `mongodb/base` functional spec. The functional spec depends on an external Mongo already listening on `localhost:27017` and can't be verified without that; this one uses the `AutoStartedMongoSpec` base already in this module (Testcontainers' `MongoDBContainer`), so it only needs Docker. It exercises `MongoDbDataStoreSpringInitializer` against a real `GenericWebApplicationContext` with no `dispatcherServlet` bean definition present — the exact registration-order-independent scenario `isWebApplicationRegistry()` was introduced for — plus a non-web control case. Both new tests pass locally, and the full `grails-data-mongodb-core` suite and `codeStyle` are clean.

borinquenkid (approved):
You did all the fixes required, we added the mongo coverage

codeconsole:
Correcting my Mongo OSIV test (it broke the MongoDB Functional Tests job): the `OpenSessionInViewSpec` I added to `grails-test-examples/mongodb/base` asserted a bean that app never registers — the Mongo OSIV class lives in `grails-datastore-web`, which mongodb/base doesn't depend on — so it failed against real Mongo. I've removed it.

@wbduque's `MongoOpenSessionInViewSpec` (grails-data-mongodb/core, Testcontainers via `AutoStartedMongoSpec`) is the right coverage here — it exercises `MongoDbDataStoreSpringInitializer.isWebApplicationRegistry()` against a real `GenericWebApplicationContext` with no `dispatcherServlet` yet, plus a non-web control. Thanks for that; merged it in.

I also kept a small complementary unit test (`AbstractDatastoreInitializerWebApplicationSpec`) covering the shared method's other branches — null registry and the legacy `dispatcherServlet` signal — with no Docker needed. Between the two, plus the Hibernate OSIV integration tests, the `isWebApplicationRegistry()` fix is covered on all branches.

jdaugherty:
I've addressedthe flaky test on the 8.x branch.  Once grails-core passes we can merge this.

PR #15936, opened 2026-07-07
Neo4j: detect web applications without relying on the dispatcherServlet bean

codeconsole:
### Summary

Follow-up to #15934, which retimes plugin bean registration to run before Spring Boot auto-configuration. `Neo4jDataStoreSpringInitializer` gates the open-session-in-view interceptor on `containsBeanDefinition('dispatcherServlet')` — a bean registered by Boot auto-configuration, so under the retimed lifecycle the check always returns false and the interceptor silently stops registering in web applications. This is the same defect #15934 found and fixed (with differential tests) in the hibernate5, hibernate7, and mongodb initializers.

The check still honours a `dispatcherServlet` definition when present, but otherwise tests whether the registry is the `WebApplicationContext` itself — a signal independent of auto-configuration ordering. Safe to merge before or after #15934: under the current lifecycle both signals agree.

### Why inlined rather than shared

`AbstractDatastoreInitializer.isWebApplicationRegistry()` was added in grails-datamapping-core by #15934, but grails-data-neo4j is an independent Gradle build resolving grails-datamapping-core from published artifacts that don't yet contain the shared method. The method is mirrored here with identical logic (documented in its groovydoc); once this build depends on a grails-datamapping-core release that ships it, the local copy is a redundant override and can be deleted.

### Verification

The identical logic is exercised in #15934 by new `OpenSessionInViewSpec` integration tests in the hibernate5 and hibernate7 `app1` example apps (differential-tested: interceptor present on 8.0.x, absent under the retimed lifecycle before the fix, present after). The grails-data-neo4j build could not be run locally — it pins `grails-gradle-plugins:6.1.2`, which is not resolvable from Maven Central and requires credentials for `repo.grails.org/restricted` — so compile/test verification for this module relies on CI.

Related: #15934 (parent change and shared fix), #15755 (superseded design discussion).


jdaugherty:
B/c this source isn't part of our normal build, i've merged it to not loose track.

PR #15939, opened 2026-07-08
Add method-based doWithSpring(BeanBuilder) plugin lifecycle hook

codeconsole:
## Summary

Adds a method-based `doWithSpring(BeanBuilder)` hook to the `grails.plugins.Plugin` base class, as an alternative to the closure-returning `doWithSpring()`. This is the plugin analog of the method-based TagLib syntax added in #15465, applied to plugin Spring configuration.

Instead of returning a closure whose delegate the container wires up at runtime, a plugin can register beans directly against the supplied builder:

```groovy
class I18nGrailsPlugin extends Plugin {

    def version = "0.1"

    @Override
    void doWithSpring(BeanBuilder beans) {
        beans.messageSource(ReloadableResourceBundleMessageSource) {
            basename = "WEB-INF/grails-app/i18n/messages"
        }
        beans.localeResolver(CookieLocaleResolver)
    }
}
```

## Scope

The hook is on the `grails.plugins.Plugin` base class and is dispatched by `DefaultGrailsPlugin` for plugin descriptors. It is intentionally **not** added to the `GrailsApplicationLifeCycle` interface: that interface is map-coerced in the testing support (`GrailsApplicationBuilder` builds `[doWithSpring: {…}] as GrailsApplicationLifeCycle`), and Groovy's map-to-interface proxies do not honour Java default methods, so an interface-level hook would break every unit test that builds an application context. This applies to plugin descriptors (`Plugin` subclasses), not to the application class or arbitrary `GrailsApplicationLifeCycle` implementors.

## Backwards compatibility

Fully preserved. The new hook is a no-op on `Plugin`, so existing plugins are unaffected. `DefaultGrailsPlugin` invokes both hooks for a `Plugin` descriptor, so a plugin may override `doWithSpring()`, `doWithSpring(BeanBuilder)`, or both. The legacy closure-returning form and the legacy `def doWithSpring = { }` closure-property form on non-`Plugin` descriptors remain fully supported.

### Defining both forms

A plugin must define **at most one** of `doWithSpring()` or `doWithSpring(BeanBuilder)`. Both forms are explicit overrides and are alternatives, so defining both throws a `PluginException` and fails plugin load rather than silently combining or preferring one — surfacing what is almost always an incomplete migration between the two forms. (This deliberately differs from the auto-detected TagLib case in #15465, where method-tags are convention-discovered; see the discussion thread.)

## Tests

`PluginDoWithSpringMethodSpec` covers: the `Plugin` no-op default, direct method-form registration against a `BeanBuilder`, the `DefaultGrailsPlugin` dispatch for the method form / both hooks / legacy closure form, and the warning emitted (and not emitted) depending on whether both forms are defined.

`./gradlew clean aggregateViolations` reports no Checkstyle / CodeNarc / PMD / SpotBugs violations.

## Docs

Documents the new hook and the both-forms behavior in the plugin runtime-configuration guide (`grails-doc`).


grails-core/src/main/groovy/org/grails/plugins/DefaultGrailsPlugin.java
```diff
@@ -391,13 +391,17 @@ public void doWithRuntimeConfiguration(RuntimeSpringConfiguration springConfig)
...
+            // Legacy closure-returning hook: doWithSpring() returns a bean-defining closure
+            Closure c = pluginObject.doWithSpring();
             if (c != null) {
-                BeanBuilder bb = new BeanBuilder(getParentCtx(), springConfig, grailsApplication.getClassLoader());
-                bb.setBinding(b);
                 c.setDelegate(bb);
                 bb.invokeMethod("beans", new Object[]{c});
             }
+            // Method-based hook: doWithSpring(BeanBuilder) registers beans directly against the builder
+            pluginObject.doWithSpring(bb);
```
jdaugherty, line 411: Don't we want to error if they define both?  shouldn't we document that the closure form is legacy and will be removed in a future version?  You can output a warning even.

borinquenkid:
*(Analysis assisted by Claude Code — I reviewed and am posting the findings below, but flagging the tooling for transparency.)*

Nice feature — the method-based `doWithSpring(BeanBuilder)` hook and its dispatch through `DefaultGrailsPlugin` are correctly implemented and well tested for `Plugin`-based plugin descriptors.

However, the PR description claims broader coverage than the diff actually delivers, and I think that gap is worth closing (or the description narrowing) before merge.

## The hook isn't wired into `GrailsApplicationLifeCycle` at all

The description says the new hook is "added as a `default` no-op on `GrailsApplicationLifeCycle`" and "overridden as a no-op on ... `GrailsApplicationLifeCycleAdapter`", and that **both** dispatch sites — `DefaultGrailsPlugin` and `GrailsApplicationPostProcessor` — invoke both hooks. Checking the diff against `21a937de06c` (this PR's head):

- `GrailsApplicationLifeCycle.groovy` is untouched — the interface still only declares `Closure doWithSpring()`:
  https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/core/GrailsApplicationLifeCycle.groovy#L29-L36
- `GrailsApplicationLifeCycleAdapter.groovy` is untouched — same, only the closure-returning override:
  https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/core/GrailsApplicationLifeCycleAdapter.groovy#L34
- `GrailsApplicationPostProcessor.groovy` still only calls the closure form:
  ```groovy
  if (lifeCycle) {
      def withSpring = lifeCycle.doWithSpring()
      if (withSpring) {
          def bb = new BeanBuilder(null, springConfig, application.classLoader)
          bb.beans(withSpring)
      }
  }
  ```
  https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy#L227-L231
- The new method exists only on `Plugin`, and (correctly, given the above) carries no `@Override` — it isn't implementing anything from the interface:
  https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/plugins/Plugin.groovy#L109-L121

## Why this matters in practice

`GrailsApplicationPostProcessor`'s `lifeCycle` field is typed `GrailsApplicationLifeCycle`, and a standard Grails app's `Application` class reaches it via `GrailsAutoConfiguration implements GrailsApplicationClass` → `GrailsApplicationClass implements GrailsApplicationLifeCycle` (not `Plugin`):

- https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/core/GrailsApplicationClass.groovy#L27
- https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/main/groovy/grails/boot/config/GrailsAutoConfiguration.groovy#L49

So as written, only actual `grails.plugins.Plugin` subclasses dispatched through `DefaultGrailsPlugin` (i.e. plugin descriptors) can use the new hook. An app's own `Application` class — or any other `GrailsApplicationLifeCycle` implementor that isn't a `Plugin` — has no `doWithSpring(BeanBuilder)` method to override, and nothing would call it if it did.

## Test coverage mirrors the gap

Every fixture class in `PluginDoWithSpringMethodSpec` extends `Plugin` (`MethodHookPlugin`, `NoOpPlugin`, `MethodHookGrailsPlugin`, `BothHooksGrailsPlugin`, `LegacyClosureGrailsPlugin`). Nothing exercises `GrailsApplicationLifeCycleAdapter` or `GrailsApplicationPostProcessor`, which is consistent with the interface/adapter/post-processor never actually being touched:
https://github.com/apache/grails-core/blob/21a937de06ccd318036e1d7cb1b0370c243225dd/grails-core/src/test/groovy/grails/plugins/PluginDoWithSpringMethodSpec.groovy

## Suggestion

Either:
1. Add `doWithSpring(BeanBuilder)` as a default no-op on `GrailsApplicationLifeCycle`, override it on `GrailsApplicationLifeCycleAdapter`, and dispatch it from `GrailsApplicationPostProcessor` alongside the closure form (with a test covering that path) — so the feature actually matches the description, or
2. Narrow the PR description/docs to state this hook applies to plugin descriptors specifically (via `Plugin`/`DefaultGrailsPlugin`), not to the general `GrailsApplicationLifeCycle` contract or the application class.

grails-core/src/main/groovy/org/grails/plugins/DefaultGrailsPlugin.java
codeconsole: Good question. I looked at how the sibling method-based TagLib PR (#15465) handles a tag defined in both forms: it doesn't error — `TagOutput.captureTagOutput` runs the closure and treats the method as a fallback, and registration only emits a `LOG.debug`. But I don't think we should copy that precedence here, because its rationale doesn't transfer.

TagLib method-tags are *auto-detected by convention* (`TagMethodInvoker.isTagMethodCandidate` promotes any public `(Map attrs)` / `(Closure body)` / `(Map attrs, Closure body)` method into a tag). So closure-first exists there to stop the new automatic scanning from silently overriding existing *explicit* closures — it's a "do no harm to existing taglibs" guard. Our two `doWithSpring` forms are both *explicit* overrides, and nothing pre-existing defines both, so there's no back-compat behavior to protect and no accidental-collision risk.

Given both forms are deliberate, "both defined" almost always means a migration between them is incomplete. Rather than error (which fails plugin load and breaks that transient state) or silently drop one form (app-breaking behavior hiding under a warning), the PR now **runs both and logs a WARN** telling the author to consolidate. That keeps a mid-migration plugin working while surfacing the issue.

I've pushed this. I also scoped the hook to the `Plugin` class rather than the `GrailsApplicationLifeCycle` interface (the interface is map-coerced in the testing support and Groovy's map proxies don't honour default methods, which would break unit tests), and corrected the PR description, which previously overclaimed the interface/adapter/post-processor coverage.

Happy to switch to a hard error instead if you'd prefer forcing consolidation over tolerating the migration window.

jdaugherty: we should error if both are defined, we want people to migrate and not define both

codeconsole: Switched to a hard error as you prefer. A plugin that defines both `doWithSpring()` and `doWithSpring(BeanBuilder)` now throws a `PluginException` and fails to load, rather than running both with a warning:

```java
Closure c = pluginObject.doWithSpring();
if (c != null && isDoWithSpringMethodOverridden(pluginObject)) {
    throw new PluginException("Plugin [" + this + "] defines both the closure-returning doWithSpring() " +
            "and the doWithSpring(BeanBuilder) method. Define only one Spring configuration hook.");
}
```

`PluginDoWithSpringMethodSpec` now asserts the exception (and its message identifies both conflicting hooks) instead of asserting both beans register. Docs updated to state that defining both fails plugin load. Pushed in 029aa5e.

codeconsole:
Thanks — you're right, the description had drifted ahead of the diff. I took your option 2 and narrowed both the description and the scope: the hook lives on `grails.plugins.Plugin` and is dispatched by `DefaultGrailsPlugin` for plugin descriptors only. I've updated the PR description to match.

For the record on why not option 1 (wire it into `GrailsApplicationLifeCycle`): an earlier revision of this PR did exactly that — a `default` method on the interface, an override on `GrailsApplicationLifeCycleAdapter`, and dispatch from `GrailsApplicationPostProcessor`. It broke CI across every module that builds an app context in a unit test. The testing support map-coerces the interface:

```groovy
// GrailsApplicationBuilder
super([doWithSpring: { -> doWithSpringClosure }] as GrailsApplicationLifeCycle, ...)
```

and Groovy's map-to-interface proxies don't honour Java `default` methods, so `lifeCycle.doWithSpring(bb)` threw `MissingMethodException` for the whole `GrailsUnitTest` path. So the interface-level version isn't a small follow-up — it needs the testing support reworked off map coercion first. Scoping to `Plugin` keeps this PR focused and safe; extending it to the application class can be a separate change.

Separately, per @jdaugherty's review, defining both hook forms is now a hard error (`PluginException`) rather than running both.

jdaugherty (approved):

borinquenkid (approved):
Documentation fixed

PR #15940, opened 2026-07-08
Add available-locale discovery and a language selector to i18n

codeconsole:
## Summary

Grails already ships the *switch-language* half of i18n — `I18nAutoConfiguration` wires a `SessionLocaleResolver` and a `?lang=`-driven `LocaleChangeInterceptor`, and `FormTagLib` has a `g:localeSelect`. What was missing was a way to discover **which locales an application is actually translated into** (`g:localeSelect` lists all ~150 `Locale.getAvailableLocales()`, which is not useful as a language switcher) and a ready-made selector in the generated app.

This PR adds that discovery to `grails-i18n` and surfaces a language dropdown in the default `create-app` layout. Because a generated web app already ships the `messages_*.properties` bundles (via the forge default plugins / the `base` profile skeleton), a freshly created app shows a fully-populated language switcher out of the box with no configuration.

## Changes

**`grails-i18n`**
- New `AvailableLocaleResolver` — scans the classpath for `messages_*.properties` and returns the translated locales (plus the configurable default locale, `grails.i18n.default.locale`, default `en`), sorted by display name and cached.
- `I18nAutoConfiguration` registers it as a bean (`@ConditionalOnMissingBean`).
- `I18nGrailsPlugin` publishes the list to the servlet context (`availableLocales`) on startup and re-scans on dev-mode bundle changes. Reading a servlet-context attribute keeps consumers decoupled from this module (no new compile dependency in `grails-gsp` / `grails-forge`).

**`grails-gsp`**
- `g:localeSelect` gains an `available="true"` attribute that lists only the translated locales instead of every JVM locale. Default behaviour is unchanged.

**Generated app (`grails-forge` template + classic `web` profile skeleton)**
- The default `main.gsp` navbar gets a Bootstrap language dropdown that iterates the published locales and switches via the existing `?lang=` interceptor. It renders only when more than one locale is available.

**Docs**
- `localeSelect` tag reference + the "Changing Locales" guide.

## Tests
- `AvailableLocaleResolverSpec` (resolver discovery, sorting, default locale, caching, immutability)
- `LocaleSelectAvailableSpec` (`available="true"` behaviour + a render test of the generated navbar snippet producing correct `?lang=` links)
- `GrailsGspSpec` gains a check that the generated layout contains the dropdown

All green.

## Update: plugin-contributed locales (default on)

`AvailableLocaleResolver` now discovers locales contributed by plugins too — their bundles are namespaced (e.g. `spring-security-core_fr.properties`) and were previously missed. When enabled it scans every root-level `*.properties` bundle on the classpath and validates each candidate suffix against `Locale.getISOLanguages()`, so non-i18n files (`application.properties`, etc.) are ignored.

Controlled by `grails.i18n.availableLocales.includePlugins`, **default `true`** — a selector lists every locale the app or its plugins provide out of the box. Set it to `false` to restrict discovery to the application's own `messages_*.properties`.

### DIFF
--- a/grails-doc/src/en/guide/i18n/changingLocales.adoc
+++ b/grails-doc/src/en/guide/i18n/changingLocales.adoc
@@ -28,7 +28,14 @@ Grails will automatically switch the user's locale and subsequent requests will 
 
 To offer a language selector, the i18n plugin discovers the locales your application is actually translated into by scanning the classpath for `messages_&#42;.properties` bundles (plus the default locale, configurable via `grails.i18n.default.locale`). This list is exposed as the `AvailableLocaleResolver` bean and published to the servlet context under the `availableLocales` attribute. The `<g:localeSelect available="true"/>` tag renders a `<select>` limited to those locales, and you can iterate the list directly in a view to build your own switcher:
 
-By default every `&#42;.properties` bundle on the classpath is considered, so locales contributed by plugins &mdash; whose bundles are namespaced (for example `spring-security-core_fr.properties`) &mdash; are listed alongside your application's own. Candidate suffixes are validated against `Locale.getISOLanguages()`, so non-i18n properties files (such as `application.properties`) are ignored. To restrict discovery to your application's own `messages_&#42;.properties` bundles, set `grails.i18n.availableLocales.includePlugins=false`:
+[source,xml]
+----
+<g:each in="${application.getAttribute('availableLocales')}" var="locale">
+    <a href="?lang=${locale.toLanguageTag()}">${locale.getDisplayName(locale)}</a>
+</g:each>
+----
+
+By default every `&#42;.properties` bundle on the classpath is considered, so locales contributed by plugins &mdash; whose bundles are namespaced (for example `spring-security-core_fr.properties`) &mdash; are listed alongside your application's own. Candidate suffixes are validated against the ISO language codes (and the ISO country codes when a country is present), so non-i18n properties files (such as `application.properties`) are ignored. To restrict discovery to your application's own `messages_&#42;.properties` bundles, set `grails.i18n.availableLocales.includePlugins=false`:
 
 [source,yaml]
 .grails-app/conf/application.yml
@@ -37,14 +44,6 @@ grails:
     i18n:
         availableLocales:
             includePlugins: false   # list only the application's own translations
-----
-
-
-[source,xml]
-----
-<g:each in="${application.getAttribute('availableLocales')}" var="locale">
-    <a href="?lang=${locale.toLanguageTag()}">${locale.getDisplayName(locale)}</a>
-</g:each>
 ----
 
 Applications created with `grails create-app` include a ready-made language dropdown in the default layout that uses exactly this mechanism.
--- a/grails-doc/src/en/guide/i18n/changingLocales.adoc
+++ b/grails-doc/src/en/guide/i18n/changingLocales.adoc
@@ -28,6 +28,18 @@ Grails will automatically switch the user's locale and subsequent requests will 
 
 To offer a language selector, the i18n plugin discovers the locales your application is actually translated into by scanning the classpath for `messages_&#42;.properties` bundles (plus the default locale, configurable via `grails.i18n.default.locale`). This list is exposed as the `AvailableLocaleResolver` bean and published to the servlet context under the `availableLocales` attribute. The `<g:localeSelect available="true"/>` tag renders a `<select>` limited to those locales, and you can iterate the list directly in a view to build your own switcher:
 
+By default every `&#42;.properties` bundle on the classpath is considered, so locales contributed by plugins &mdash; whose bundles are namespaced (for example `spring-security-core_fr.properties`) &mdash; are listed alongside your application's own. Candidate suffixes are validated against `Locale.getISOLanguages()`, so non-i18n properties files (such as `application.properties`) are ignored. To restrict discovery to your application's own `messages_&#42;.properties` bundles, set `grails.i18n.availableLocales.includePlugins=false`:
+
+[source,yaml]
+.grails-app/conf/application.yml
+----
+grails:
+    i18n:
+        availableLocales:
+            includePlugins: false   # list only the application's own translations
+----
+
+
 [source,xml]
 ----
 <g:each in="${application.getAttribute('availableLocales')}" var="locale">
--- a/grails-doc/src/en/guide/i18n/changingLocales.adoc
+++ b/grails-doc/src/en/guide/i18n/changingLocales.adoc
@@ -26,6 +26,17 @@ By default, the user locale is detected from the incoming `Accept-Language` head
 
 Grails will automatically switch the user's locale and subsequent requests will use the switched locale.
 
+To offer a language selector, the i18n plugin discovers the locales your application is actually translated into by scanning the classpath for `messages_&#42;.properties` bundles (plus the default locale, configurable via `grails.i18n.default.locale`). This list is exposed as the `AvailableLocaleResolver` bean and published to the servlet context under the `availableLocales` attribute. The `<g:localeSelect available="true"/>` tag renders a `<select>` limited to those locales, and you can iterate the list directly in a view to build your own switcher:
+
+[source,xml]
+----
+<g:each in="${application.getAttribute('availableLocales')}" var="locale">
+    <a href="?lang=${locale.toLanguageTag()}">${locale.getDisplayName(locale)}</a>
+</g:each>
+----
+
+Applications created with `grails create-app` include a ready-made language dropdown in the default layout that uses exactly this mechanism.
+
 By default, Grails uses {springapi}org/springframework/web/servlet/i18n/SessionLocaleResolver.html[SessionLocaleResolver] as the `localeResolver` bean.
 
 You can change the default locale easily: 
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
@@ -16,6 +16,24 @@
         <a class="navbar-brand d-flex align-items-center" href="${request.contextPath}/">
             <asset:image class="w-75" src="grails.svg" alt="Grails Logo"/>
         </a>
+        <g:set var="availableLocales" value="${application.getAttribute('availableLocales')}"/>
+        <g:if test="${availableLocales && availableLocales.size() > 1}">
+            <g:set var="currentLocale" value="${org.springframework.web.servlet.support.RequestContextUtils.getLocale(request)}"/>
+            <ul class="navbar-nav ms-auto">
+                <li class="nav-item dropdown">
+                    <a class="nav-link dropdown-toggle" href="#" id="localeDropdown" role="button" data-bs-toggle="dropdown" aria-expanded="false">
+                        <i class="bi bi-globe me-1"></i>${currentLocale.getDisplayName(currentLocale)}
+                    </a>
+                    <ul class="dropdown-menu dropdown-menu-end" aria-labelledby="localeDropdown">
+                        <g:each in="${availableLocales}" var="availableLocale">
+                            <li>
+                                <a class="dropdown-item${availableLocale == currentLocale ? ' active' : ''}" href="?lang=${availableLocale.toLanguageTag()}">${availableLocale.getDisplayName(availableLocale)}</a>
+                            </li>
+                        </g:each>
+                    </ul>
+                </li>
+            </ul>
+        </g:if>
     </div>
 </nav>
 
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
@@ -28,7 +28,7 @@
                     <ul class="dropdown-menu dropdown-menu-end" aria-labelledby="localeDropdown">
                         <g:each in="${availableLocales}" var="availableLocale">
                             <li>
-                                <a class="dropdown-item${availableLocale == currentLocale ? ' active' : ''}" href="?lang=${availableLocale.toLanguageTag()}">${availableLocale.getDisplayName(availableLocale)}</a>
+                                <a class="dropdown-item${availableLocale.language == currentLocale.language ? ' active' : ''}" href="?lang=${availableLocale.toLanguageTag()}">${availableLocale.getDisplayName(availableLocale)}</a>
                             </li>
                         </g:each>
                     </ul>
--- a/grails-i18n/src/main/groovy/org/grails/plugins/i18n/AvailableLocaleResolver.java
+++ b/grails-i18n/src/main/groovy/org/grails/plugins/i18n/AvailableLocaleResolver.java
@@ -50,8 +50,9 @@ import org.springframework.core.io.support.ResourcePatternResolver;
  * scanned. When {@code includePluginBundles} is enabled, every {@code *.properties}
  * bundle on the classpath is considered, so locales contributed by plugins &mdash; whose
  * bundles are namespaced (e.g. {@code spring-security-core_fr.properties}) &mdash; are
- * included too. Candidate suffixes are validated against {@link Locale#getISOLanguages()}
- * so non-i18n properties files (e.g. {@code application.properties}) are ignored. See
+ * included too. Candidate suffixes are validated against the ISO language codes (and the ISO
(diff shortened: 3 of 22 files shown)


borinquenkid (commented):
## Test coverage audit for this PR (grounded in Codecov's patch report + a line-by-line read of the head commit `8f9d7aa`)

Codecov: **75% patch coverage**, 17 lines uncovered. Per-file: `AvailableLocaleResolver.java` 80.85% (4 missing/5 partial), `I18nGrailsPlugin.groovy` 53.85% (4 missing/2 partial), `FormTagLib.groovy` 66.67% (0 missing/2 partial branches). Below is what those numbers correspond to in the code, confirmed by grepping the actual test files (no guessing from the summary alone).

### 1. `I18nGrailsPlugin.doWithApplicationContext()` / `publishAvailableLocales()` — no plugin-lifecycle test at all
https://github.com/apache/grails-core/blob/8f9d7aa2bc65769b371bc392fa185df5f3edb261/grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nGrailsPlugin.groovy#L53-L68

Nothing in this PR (or pre-existing) boots the plugin against a `WebApplicationContext` + `ServletContext` and asserts `servletContext.getAttribute('availableLocales')` (L66) is actually populated. The two early-return branches are also unexercised:
- L60-62: `ctx` is not a `WebApplicationContext`
- L64: `servletContext == null` or the `availableLocaleResolver` bean is absent

The `onChange()` addition at L123-126 (clear the resolver's cache and republish when a bundle changes) has no test either — I grepped both new specs (`AvailableLocaleResolverSpec`, `LocaleSelectAvailableSpec`) and neither calls `onChange` or references plugin reload behavior.

### 2. `I18nAutoConfiguration.availableLocaleResolver` bean — wiring itself is unasserted
https://github.com/apache/grails-core/blob/8f9d7aa2bc65769b371bc392fa185df5f3edb261/grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nAutoConfiguration.java#L151-L158

Every sibling bean in this class (`localeResolver`, `localeChangeInterceptor`, `messageSource`) has a `WebApplicationContextRunner`-based test in `I18nAutoConfigurationSpec.groovy` — property-driven variants and `@ConditionalOnMissingBean` back-off against a user-supplied bean. `grep -n "availableLocaleResolver\|AvailableLocaleResolver" I18nAutoConfigurationSpec.groovy` returns nothing: the new bean isn't checked for presence, isn't checked for property wiring (`grails.i18n.default.locale`, `grails.i18n.availableLocales.includePlugins`), and the `@ConditionalOnMissingBean(AvailableLocaleResolver.class)` back-off path is untested. `AvailableLocaleResolverSpec` only exercises `new AvailableLocaleResolver(...)` directly, not the Spring wiring.

### 3. `grails-profiles/web/skeleton/.../main.gsp` — zero coverage (the forge twin got a test, this didn't)
https://github.com/apache/grails-core/blob/8f9d7aa2bc65769b371bc392fa185df5f3edb261/grails-profiles/web/skeleton/grails-app/views/layouts/main.gsp#L21-L36

Identical dropdown markup was added to `grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp`, and that copy got a new `GrailsGspSpec` test ("test default layout includes the language selector dropdown"). This classic-profile copy has no equivalent — `grails-profiles` has no `src/test` source set in the repo at all, so nothing verifies this template even renders, let alone that the `g:if` gate on `size() > 1` behaves.

### 4. `AvailableLocaleResolver` — well-tested class, but specific branches are still open (matches Codecov's 4 missing/5 partial)
https://github.com/apache/grails-core/blob/8f9d7aa2bc65769b371bc392fa185df5f3edb261/grails-i18n/src/main/groovy/org/grails/plugins/i18n/AvailableLocaleResolver.java

- L128: `defaultLocale == null` or empty-language (`Locale.ROOT`) — the "don't add a default" branch is never hit by any spec.
- L143-144: the `IOException` catch in `computeAvailableLocales()` — unreachable without mocking `ResourcePatternResolver`.
- L157: `filename == null` / non-`.properties` filename branch.
- L162-163: a `*.properties` file with no underscore, reached only when `includePluginBundles=true` — currently only covered incidentally by whatever happens to be on the test classpath (e.g. `application.properties`), not by a deliberate fixture.
- L166-167: an empty suffix after the underscore (`messages_.properties`).
- L170: a non-ISO-language suffix on a *deliberate* fixture — same incidental-vs-intentional gap as above.

### 5. `FormTagLib.localeSelect` — one untested branch
https://github.com/apache/grails-core/blob/8f9d7aa2bc65769b371bc392fa185df5f3edb261/grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy#L979-L989

`LocaleSelectAvailableSpec` covers `available="true"` and the attribute being absent, but never an explicit `available="false"`. L981 (`Boolean.valueOf(availableAttr.toString())`) has an untested false-but-present path. Confirmed: `grep -n "available=\"false\""` in the spec returns nothing.

---
*Findings generated by automated analysis of the Codecov patch report and the PR head commit; reviewed by a human before posting.*

codeconsole:
Thanks for the thorough audit — all five findings were accurate. Addressed in 170ae02d0f:

**1. Plugin lifecycle** — Added `I18nGrailsPluginSpec` covering `doWithApplicationContext()` publishing the resolved locales to the servlet context, the non-`WebApplicationContext` and missing-bean early returns, and `onChange` clearing the resolver cache and republishing a freshly computed list. This also surfaced an NPE in `onChange` when `BuildSettings.RESOURCES_DIR` is unset (as in unit tests), now null-guarded.

**2. `availableLocaleResolver` bean wiring** — `I18nAutoConfigurationSpec` now asserts the bean registers by default, honours `grails.i18n.default.locale` (including the underscore form `pt_BR`) and `grails.i18n.availableLocales.includePlugins=false`, and backs off to a user-defined `AvailableLocaleResolver` bean.

**3. Web-profile skeleton `main.gsp`** — Acknowledged, but leaving as-is: `grails-profiles` modules are resource-only packaging projects with no test source sets anywhere in the repo, so there is no existing place to test profile templates. The markup is byte-identical to the forge template, which `GrailsGspSpec` covers.

**4. `AvailableLocaleResolver` branches** — Covered everything reachable through the public API: null and `Locale.ROOT` default locales, the `IOException` catch (via a `ClassLoader` whose `getResources` throws, through the public constructor), and deliberate malformed-bundle fixtures (`standalone.properties`, `messages_.properties`, `bogus-plugin_zz.properties`) replacing the incidental coverage. The remaining two branches (`filename == null`, non-`.properties` filename) are defensive dead code on the scan path — the classpath pattern guarantees `.properties` filenames — and testing them would require invoking the private method directly, which this repo's testing rules disallow.

**5. `available="false"`** — Added an explicit test: with locales published to the servlet context, `available="false"` still renders the full JVM locale list.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nGrailsPlugin.groovy
```diff
@@ -42,6 +43,30 @@ class I18nGrailsPlugin extends Plugin {
...
+        publishAvailableLocales()
+    }
+
+    private void publishAvailableLocales() {
+        def ctx = applicationContext
+        if (!(ctx instanceof WebApplicationContext)) {
+            return
+        }
+        def servletContext = ((WebApplicationContext) ctx).servletContext
+        if (servletContext != null && ctx.containsBean('availableLocaleResolver')) {
```
jdaugherty: The resolver is looked up by bean **name** here (and again in `onChange`), but the auto-configuration backs off by **type** (`@ConditionalOnMissingBean(AvailableLocaleResolver.class)`). If a user registers their own `AvailableLocaleResolver` under any other bean name, the auto-configured bean backs off, `containsBean('availableLocaleResolver')` is false, and nothing is ever published to the servlet context — the selector silently disappears even though a resolver exists.

Consider resolving by type instead, e.g. `ctx.getBeanProvider(AvailableLocaleResolver).ifAvailable { ... }`, so the customization path promised by `@ConditionalOnMissingBean` actually works end-to-end.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nAutoConfiguration.java
```diff
@@ -130,4 +130,30 @@ public MessageSource messageSource(GrailsApplication grailsApplication, GrailsPl
...
+     *
+     * @param defaultLocale the base {@code messages.properties} locale, always included
+     * ({@code grails.i18n.default.locale}, defaults to {@code en})
+     * @param includePlugins whether to also scan plugin-contributed message bundles
+     * ({@code grails.i18n.availableLocales.includePlugins}, defaults to {@code true})
+     */
+    @Bean
+    @ConditionalOnMissingBean(AvailableLocaleResolver.class)
+    public AvailableLocaleResolver availableLocaleResolver(GrailsApplication grailsApplication,
+            @Value("${grails.i18n.default.locale:en}") String defaultLocale,
```
jdaugherty: `grails.i18n.default.locale` now has two different defaults within the same class: the `defaultLocale` field above defaults to *empty* (falling back to `Locale.getDefault()` via `fixedLocale()`), while this parameter defaults to `en`. For an app whose base `messages.properties` is not English, the selector will always advertise English even though no English translation exists.

The parsing also diverges (`Locale.forLanguageTag(...replace('_','-'))` here vs `StringUtils.parseLocale` in `fixedLocale()`). Suggest injecting the same empty-default property and reusing the existing field/`StringUtils.parseLocale` so the one config key behaves consistently across the class.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/AvailableLocaleResolver.java
```diff
@@ -0,0 +1,173 @@
...
+        int underscore = base.indexOf('_');
+        if (underscore < 0) {
+            return null;
+        }
+        String code = base.substring(underscore + 1);
+        if (code.isEmpty()) {
+            return null;
+        }
+        Locale locale = Locale.forLanguageTag(code.replace('_', '-'));
+        return ISO_LANGUAGES.contains(locale.getLanguage()) ? locale : null;
```
jdaugherty: Only the language subtag is validated, so with `includePlugins=true` (the default) any root-level properties file whose suffix happens to parse to a known language leaks into the selector:

- `db_it.properties` → contributes Italian
- `foo_en_prod.properties` → `forLanguageTag("en-prod")` treats `prod` as a *script* subtag, producing an `en-Prod` locale that passes the ISO-language check and shows up as a second English-looking entry

Since `.properties` bundles follow the `basename_language(_COUNTRY(_variant))` convention rather than BCP 47, validating the whole suffix (ISO language, and ISO country when present) before accepting it would cut these false positives down considerably.

grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
```diff
@@ -18,6 +18,22 @@
...
+            <g:if test="${availableLocales && availableLocales.size() > 1}">
+                <g:set var="currentLocale" value="${org.springframework.web.servlet.support.RequestContextUtils.getLocale(request)}"/>
+                <li class="nav-item dropdown">
+                    <a class="nav-link dropdown-toggle" href="#" id="localeDropdown" role="button" data-bs-toggle="dropdown" aria-expanded="false">
+                        <i class="bi bi-globe me-1"></i>${currentLocale.getDisplayName(currentLocale)}
+                    </a>
+                    <ul class="dropdown-menu dropdown-menu-end" aria-labelledby="localeDropdown">
+                        <g:each in="${availableLocales}" var="availableLocale">
+                            <li>
+                                <a class="dropdown-item${availableLocale == currentLocale ? ' active' : ''}" href="?lang=${availableLocale.toLanguageTag()}">${availableLocale.getDisplayName(availableLocale)}</a>
```
jdaugherty: Two behavioural nits (both also apply to the identical copy in `grails-profiles/web/skeleton/grails-app/views/layouts/main.gsp`):

1. `availableLocale == currentLocale` compares full locales, but `currentLocale` usually carries a country on first visit (e.g. `en_US` from `Accept-Language`) while the discovered list holds language-only locales — so the `active` highlight never fires until the user explicitly picks an entry. A language-level fallback comparison would behave better out of the box.
2. `href="?lang=..."` replaces the entire query string, so switching language drops any existing request parameters (pagination, filters, etc.). Probably acceptable for the scaffold, but worth a conscious decision.

grails-doc/src/en/guide/i18n/changingLocales.adoc
```diff
@@ -26,6 +26,29 @@ By default, the user locale is detected from the incoming `Accept-Language` head
 
 Grails will automatically switch the user's locale and subsequent requests will use the switched locale.
 
+To offer a language selector, the i18n plugin discovers the locales your application is actually translated into by scanning the classpath for `messages_&#42;.properties` bundles (plus the default locale, configurable via `grails.i18n.default.locale`). This list is exposed as the `AvailableLocaleResolver` bean and published to the servlet context under the `availableLocales` attribute. The `<g:localeSelect available="true"/>` tag renders a `<select>` limited to those locales, and you can iterate the list directly in a view to build your own switcher:
```
jdaugherty, line 29: This paragraph ends with "build your own switcher:" but the `g:each` example it introduces is pushed below the plugin-discovery paragraph and the YAML block. Reordering to intro → example → plugin-discovery paragraph → YAML would read better.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nGrailsPlugin.groovy
codeconsole: Good catch — fixed in 9d79f67d09. Both `publishAvailableLocales()` and `onChange` now resolve the resolver by type via `ctx.getBeanProvider(AvailableLocaleResolver).getIfAvailable()`, so a user-defined resolver registered under any bean name is published. Added a regression test registering the resolver under a custom name.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/I18nAutoConfiguration.java
codeconsole: Agreed — fixed in 9d79f67d09. The `availableLocaleResolver` bean now reuses the existing `defaultLocale` field via `fixedLocale()` (`StringUtils.parseLocale`, falling back to `Locale.getDefault()`), so the one config key parses and defaults identically everywhere in the class. The `I18nAutoConfigurationSpec` default-wiring test now asserts the JVM-default fallback.

grails-i18n/src/main/groovy/org/grails/plugins/i18n/AvailableLocaleResolver.java
codeconsole: Fixed in 9d79f67d09. The suffix is now parsed by the resource-bundle convention `language(_COUNTRY(_variant))` instead of BCP 47: the language must be an ISO language and, when a second segment is present, it must be an ISO country — so `foo_en_prod.properties` is rejected outright rather than misparsed as `en-Prod`. Added a `report_en_prod.properties` fixture plus a country-parsing test (`messages_pt_BR` → country `BR`). The `db_it.properties` → Italian case is inherent to the name-based heuristic; `includePlugins=false` remains the escape hatch when an app's classpath has ambiguous root-level bundles.

grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
codeconsole: 1: Fixed in 9d79f67d09 — the `active` highlight now compares the language subtag (`availableLocale.language == currentLocale.language`), so it fires on first visit when the request locale carries a country (`en_US` vs `en`). Applied to both copies.

2: Keeping `?lang=` replacing the query string as a conscious scaffold decision — the generated layout links from simple pages, and preserving arbitrary parameters via `g:link params:` in a layout misbehaves on error/404 pages that have no resolvable controller. Apps with stateful query params can adapt the snippet.

grails-doc/src/en/guide/i18n/changingLocales.adoc
codeconsole: Fixed in 9d79f67d09 — reordered to intro → `g:each` example → plugin-discovery paragraph → YAML, and updated the validation sentence to mention the new ISO-country check.

jdaugherty (approved):

PR #15944, opened 2026-07-09
Forge - Grails 8 Updates around Grails Data

### LINKED ISSUE #15942: Unresolvable dependency jandex in grails-data-hibernate7
### Expected Behavior

I want to create e new Grails application using Hibernate 7 and i am able to create a dependency report.

### Actual Behaviour

An error occured:
`Variant Selection Exception: org.gradle.internal.component.resolution.failure.exception.ConflictingConstraintsException caused by Resolution Failure: org.gradle.internal.component.resolution.failure.type.ModuleRejectedFailure
Conflicting constraints detected: Cannot find a version of 'io.smallrye:jandex' that satisfies the version constraints:
   Dependency path: 'root' (classpath) --> 'org.apache.grails:grails-data-hibernate7:8.0.0-M2' (runtimeElements) --> 'org.apache.grails.data:grails-data-hibernate7-core:8.0.0-M2' (runtimeElements) --> 'io.smallrye:jandex:3.2.3'
   Constraint path: 'root' (classpath) --> 'org.apache.grails:grails-data-hibernate7:8.0.0-M2' (runtimeElements) --> 'org.hibernate.orm:hibernate-core:7.4.1.Final' (runtimeElements) --> 'org.hibernate.orm:hibernate-platform:7.4.1.Final' (runtimeElements) --> 'io.smallrye:jandex:3.3.2'
   Constraint path: 'root' (classpath) --> 'org.apache.grails:grails-data-hibernate7:8.0.0-M2' (runtimeElements) --> 'org.apache.grails:grails-hibernate7-bom:8.0.0-M2' (runtimeElements) --> 'io.smallrye:jandex:{strictly 3.2.3}'
`
Imho `grails-data-hibernate7-core` and `hibernate-platform` require different jandex versions strictly.

### Steps To Reproduce

1. Install Grails 8.0.0-M2
2. Create new App (`grailsw.bat create-app TestApp --profile=web`)
3. Modify `build.gradle` (grails-data-hibernate5 -> grails-data-hibernate7)
4. Run `gradlew.bat -i projectReport`

### Environment Information

- Windows 11
- Java amazon-corretto-21.0.11.10.1-windows-x64-jdk

### Example Application

_No response_

### Version

8.0.0-M2

jdaugherty:
## Description

  Fixes #15942.

  Grails 8.0.0-M2 users following the documented migration path (create an app, then swap
  `grails-data-hibernate5` → `grails-data-hibernate7` in `build.gradle`) hit an unresolvable
  dependency error:

  Conflicting constraints detected: Cannot find a version of 'io.smallrye:jandex' that satisfies the version constraints

  The root cause is that a manually-migrated app ends up with **two conflicting BOMs** on the same
  classpath: the default `grails-bom` (generated into `buildSrc`/`buildscript` regardless of the data
  implementation) and `grails-hibernate7-bom` (pulled transitively by `grails-data-hibernate7`).
  `grails-hibernate7-bom` strictly pins `io.smallrye:jandex:3.2.3` while `hibernate-core` endorses
  `hibernate-platform`, which requires jandex `3.3.2` — an irreconcilable strict-constraint conflict.
  Notably, `grails-data-hibernate7` resolves fine when the *matching* Hibernate 7 BOM variant is the
  only Grails BOM in the graph.

  Rather than requiring users to hand-edit BOMs, this PR adds first-class Hibernate 7 support to
  Grails Forge so generated apps consume a consistent set of BOMs from the start:

  - **New `gorm-hibernate7` feature** (`GrailsDataHibernate7`) and `GormImpl.HIBERNATE7`, selectable
    via `create-app --data hibernate7` or `--features gorm-hibernate7`.
  - **BOM selection matrix** applied consistently to the app dependencies, the `buildscript`
    classpath, and the generated `buildSrc`:
    | Data implementation | Without Micronaut | With Micronaut |
    |---|---|---|
    | Hibernate 5 | `grails-bom` | `grails-micronaut-bom` (enforced) |
    | Hibernate 7 | `grails-hibernate7-bom` | `grails-hibernate7-micronaut-bom` (enforced) |
  - **`database-migration`** now emits `grails-data-hibernate7-dbmigration` for Hibernate 7 apps.
  - **Guard rail**: selecting both `gorm-hibernate5` and `gorm-hibernate7` is rejected with a clear
    error (`GrailsDataHibernateValidator`).
  - **CLI switch renamed** to `-d, --data` with `-g, --gorm` retained as legacy aliases.
    `GormImpl.HIBERNATE` was migrated to `HIBERNATE5`; the selection values are now `hibernate5`,
    `hibernate7`, `mongodb`, and the legacy value `hibernate` is still accepted (handled in both the
    picocli converter and a new Micronaut `TypeConverter` for the Forge HTTP API).
  - **Naming cleanup in Forge**: feature classes renamed to match their artifacts
    (`GrailsDataHibernate5`, `GrailsDataHibernate7`, `GrailsDataMongoDB`) and user-facing "GORM"
    text updated to "Grails Data" (feature titles/descriptions, CLI help and error messages, OpenAPI
    schema descriptions). Feature names (`gorm-hibernate5`, etc.) are unchanged for compatibility.
  - **Documentation**: `creatingProject.adoc`, the `create-*` command reference pages, and the sample
    help output updated for the new switch, values, and legacy aliases (also fixes stale
    `hibernate, mongodb, neo4j` value lists).

  ### Testing

  - New `GrailsDataHibernate7Spec` (12 tests): dependencies, all four BOM combinations across
    `build.gradle` and `buildSrc`, dbmigration artifact switch, mutual-exclusion validation, and
    datasource config.
  - New `GormImplSpec` covering `GormImpl.parse`, including legacy and unknown values.
  - `CreateAppCommandSpec` extended for `--data`, `-d`, legacy `-g`/`--gorm`, legacy `hibernate`
    value, and the updated invalid-value error.
  - Full `grails-forge` `check` passes, except `CreateControllerCommandSpec."test app with controller"`,
    which requires a running Docker daemon (Testcontainers) and fails identically without this change
  when Docker is unavailable.

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. (#15942)
- [x] This PR targets the **correct branch** for the type of change. (`8.0.x` — new Forge feature and CLI switch rename)

### Code Quality

- [x] My code follows the project's **code style** guidelines.
- [x] This PR does **not** include mass reformatting, style-only changes, or large-scale refactoring
      unless it was **explicitly approved** in the linked issue. (Forge feature-class renames were
      requested as part of this change.)
- [x] If generative AI tooling was used in preparing this contribution, a quality model was used to
      ensure contributions are **consistent with the project's quality standards**.

### Licensing and Attribution

- [x] All contributed code is provided under the Apache License 2.0, and new source files include
      the appropriate **Apache license header**.
- [x] If generative AI tooling was used in preparing this contribution, I have followed the ASF
      policy on generative tooling and have properly attributed its use.
      (Prepared with assistance from Claude Code; reviewed and directed by the author.)

### Documentation

- [x] If this PR introduces breaking changes or changes that require user action during an upgrade,
      I have updated the **Upgrade Notes** for the corresponding version in the Grails Guide.
      (No user action required: `-g`, `--gorm`, and the `hibernate` value remain supported as legacy aliases.)

### DIFF
--- a/grails-forge/grails-forge-api/src/test/groovy/org/grails/forge/api/FeatureControllerSpec.groovy
+++ b/grails-forge/grails-forge-api/src/test/groovy/org/grails/forge/api/FeatureControllerSpec.groovy
@@ -21,9 +21,7 @@ package org.grails.forge.api
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import io.micronaut.http.HttpRequest
-import io.micronaut.http.HttpStatus
 import io.micronaut.http.client.HttpClient
-import io.micronaut.http.client.exceptions.HttpClientResponseException
 import io.micronaut.http.client.annotation.Client
 import io.micronaut.test.extensions.spock.annotation.MicronautTest
 import jakarta.inject.Inject
--- a/grails-forge/grails-forge-api/src/test/groovy/org/grails/forge/api/FeatureControllerSpec.groovy
+++ b/grails-forge/grails-forge-api/src/test/groovy/org/grails/forge/api/FeatureControllerSpec.groovy
@@ -21,7 +21,9 @@ package org.grails.forge.api
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpStatus
 import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.exceptions.HttpClientResponseException
 import io.micronaut.http.client.annotation.Client
 import io.micronaut.test.extensions.spock.annotation.MicronautTest
 import jakarta.inject.Inject
@@ -173,9 +175,8 @@ class FeatureControllerSpec extends Specification {
 
     void "test feature filter - invalid option as query parameter"() {
         when:
-        String response = httpClient.toBlocking().withCloseable { client ->
-            client.exchange(HttpRequest.GET('/application-types/' + ApplicationType.WEB.name + '/features?javaVersion=invalid'), String.class).body()
-        }
+        String response = httpClient.toBlocking()
+                .exchange(HttpRequest.GET('/application-types/' + ApplicationType.WEB.name + '/features?javaVersion=invalid'), String.class).body()
 
         then:
         response
@@ -185,5 +186,28 @@ class FeatureControllerSpec extends Specification {
         def map = mapper.readValue(response, Map)
 
         map.features.collect { it -> it.name }.find { it == 'gorm-mongodb'}
+    }
+
+    void "test feature filter - Grails Data implementation as query parameter"() {
+        when: 'a valid Grails Data implementation is accepted'
+        String response = httpClient.toBlocking()
+                .exchange(HttpRequest.GET('/application-types/' + ApplicationType.WEB.name + '/features?gorm=hibernate7'), String.class).body()
+
+        then:
+        response
+
+        when: 'the legacy hibernate value is accepted'
+        response = httpClient.toBlocking()
+                .exchange(HttpRequest.GET('/application-types/' + ApplicationType.WEB.name + '/features?gorm=hibernate'), String.class).body()
+
+        then:
+        response
+
+        when: 'an invalid value falls back to the default like other filter options'
+        response = httpClient.toBlocking()
+                .exchange(HttpRequest.GET('/application-types/' + ApplicationType.WEB.name + '/features?gorm=invalid'), String.class).body()
+
+        then:
+        response
     }
 }
--- a/grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/command/CreateCommand.java
+++ b/grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/command/CreateCommand.java
@@ -50,7 +50,7 @@ public abstract class CreateCommand extends BaseCommand implements Callable<Inte
     DevelopmentReloading reloading;
 
     @ReflectiveAccess
-    @CommandLine.Option(names = {"-g", "--gorm"}, paramLabel = "GORM Implementation", description = "Which GORM Implementation to configure. Possible values: ${COMPLETION-CANDIDATES}.", completionCandidates = GormImplCandidates.class, converter = GormImplConverter.class)
+    @CommandLine.Option(names = {"-d", "--data", "-g", "--gorm"}, paramLabel = "Grails Data Implementation", description = "Which Grails Data implementation to configure (-g, --gorm are legacy aliases). Possible values: ${COMPLETION-CANDIDATES}.", completionCandidates = GormImplCandidates.class, converter = GormImplConverter.class)
     GormImpl gormImpl;
 
     @ReflectiveAccess
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/DatabaseDriverFeature.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/DatabaseDriverFeature.java
@@ -29,15 +29,15 @@ import java.util.Map;
 public abstract class DatabaseDriverFeature implements OneOfFeature {
 
     private final TestContainers testContainers;
-    private final HibernateGorm hibernateGorm;
+    private final GrailsDataHibernate5 grailsDataHibernate5;
 
     public DatabaseDriverFeature() {
         this.testContainers = null;
-        this.hibernateGorm = null;
+        this.grailsDataHibernate5 = null;
     }
 
-    public DatabaseDriverFeature(HibernateGorm hibernateGorm, TestContainers testContainers) {
-        this.hibernateGorm = hibernateGorm;
+    public DatabaseDriverFeature(GrailsDataHibernate5 grailsDataHibernate5, TestContainers testContainers) {
+        this.grailsDataHibernate5 = grailsDataHibernate5;
         this.testContainers = testContainers;
     }
 
@@ -56,8 +56,8 @@ public abstract class DatabaseDriverFeature implements OneOfFeature {
         if (!featureContext.isPresent(TestContainers.class) && testContainers != null) {
             featureContext.addFeature(testContainers);
         }
-        if (!featureContext.isPresent(HibernateGorm.class) && hibernateGorm != null) {
-            featureContext.addFeature(hibernateGorm);
+        if (!featureContext.isPresent(GrailsDataHibernate5.class) && grailsDataHibernate5 != null) {
+            featureContext.addFeature(grailsDataHibernate5);
         }
     }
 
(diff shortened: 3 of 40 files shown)


grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/command/CreateCommand.java
```diff
@@ -50,7 +50,7 @@ public abstract class CreateCommand extends BaseCommand implements Callable<Inte
     DevelopmentReloading reloading;
 
     @ReflectiveAccess
-    @CommandLine.Option(names = {"-g", "--gorm"}, paramLabel = "GORM Implementation", description = "Which GORM Implementation to configure. Possible values: ${COMPLETION-CANDIDATES}.", completionCandidates = GormImplCandidates.class, converter = GormImplConverter.class)
+    @CommandLine.Option(names = {"-d", "--data", "-g", "--gorm"}, paramLabel = "Grails Data Implementation", description = "Which Grails Data implementation to configure (-g, --gorm are legacy aliases). Possible values: ${COMPLETION-CANDIDATES}.", completionCandidates = GormImplCandidates.class, converter = GormImplConverter.class)
```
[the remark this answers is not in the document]
jdaugherty: No, this is intentional

borinquenkid (approved):
LGTM

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/DatabaseDriverFeature.java
```diff
@@ -56,8 +56,8 @@ public void processSelectedFeatures(FeatureContext featureContext) {
         if (!featureContext.isPresent(TestContainers.class) && testContainers != null) {
             featureContext.addFeature(testContainers);
         }
-        if (!featureContext.isPresent(HibernateGorm.class) && hibernateGorm != null) {
-            featureContext.addFeature(hibernateGorm);
+        if (!featureContext.isPresent(GormFeature.class) && grailsDataHibernate5 != null) {
```
codeconsole: Changing the guard from `isPresent(HibernateGorm.class)` to `isPresent(GormFeature.class)` also changes behavior for the MongoDB case: since `GrailsDataMongoDB extends GormOneOfFeature extends GormFeature` (and `isPresent` uses `isAssignableFrom`), selecting e.g. `postgres` + `gorm-mongodb` previously auto-added Hibernate 5 and now doesn't.

That's needed for the Hibernate 7 case and arguably more correct, but the generated app now has a JDBC driver with no SQL data implementation and no `dataSource` config (that config is only applied from the Hibernate features' `apply`). Is that intended? If so, could you add a test locking the mongo + SQL-driver output so this doesn't regress silently?

grails-forge/grails-forge-api/src/test/groovy/org/grails/forge/api/FeatureControllerSpec.groovy
```diff
@@ -21,7 +21,9 @@ package org.grails.forge.api
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpStatus
```
codeconsole: `HttpStatus` (line 24) and `HttpClientResponseException` (line 26) are imported but never used — the new test only asserts success responses. Can be dropped.

codeconsole (commented):
Nice work — this fixes the root cause of #15942 (the conflicting `grails-bom` + `grails-hibernate7-bom` pair) rather than patching versions, and the BOM matrix is applied consistently across the app dependencies, the buildscript classpath, and `buildSrc`. Test coverage is thorough: all four BOM combinations in both `build.gradle` and `buildSrc`, the dbmigration artifact switch, mutual exclusion, and legacy value/alias parsing at every entry point (picocli, HTTP query binding, default option). Docs are complete, including the stale `hibernate, mongodb, neo4j` value lists and the copy-paste `ServletImplDTO` schema description.

Verified locally under JDK 25: `GrailsDataHibernate7Spec` (15 tests), `GormImplSpec` (9 tests), and `CreateAppCommandSpec` (8 tests, including the new `--data`/`-d`/legacy alias cases) all pass. I also ran the Forge CLI itself (`create-app --data hibernate7`) — the generated app has `grails-hibernate7-bom` + `grails-data-hibernate7` on both the application dependencies and the buildscript classpath, and `./gradlew dependencies --configuration runtimeClasspath` in that app resolves cleanly against the ASF snapshots repo: the #15942 jandex strict-constraint conflict does not occur (jandex resolves to the pinned 3.2.3).

A few comments inline — nothing blocking. The only one I'd like confirmed before merge is the `DatabaseDriverFeature` behavior change for the MongoDB + SQL-driver combination.

Two non-blocking follow-up thoughts that don't need to hold up this PR:

- `GrailsDataHibernate5` and `GrailsDataHibernate7` are identical except for the artifact id (same key constants, same `apply` body, same `processSelectedFeatures`). An abstract base with a `getDataArtifactId()` hook would keep future config changes in one place.
- The BOM selection matrix now lives in two places: `GrailsBase.apply` and the `@if/@else` chain in `buildSrcBuildGradle.rocker.raw`. A future BOM variant has to be added to both. If the computed BOM artifact could be passed to the template (it already receives `grailsVersion`), the matrix would have a single source of truth; at minimum the comment in `GrailsBase` could mention the template mirror.

One release-note consideration: renaming `GormImpl.HIBERNATE` to `HIBERNATE5` changes the serialized `value` in the select-options API. Query-parameter input is protected by the legacy-value handling, but any external client echoing the old enum name will break. Fine for a major, just worth remembering if the Forge HTTP API is considered public.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/options/GormImplTypeConverter.java
```diff
@@ -0,0 +1,48 @@
...
+public class GormImplTypeConverter implements TypeConverter<CharSequence, GormImpl> {
+
+    @Override
+    public Optional<GormImpl> convert(CharSequence object, Class<GormImpl> targetType, ConversionContext context) {
+        if (object == null) {
+            return Optional.empty();
+        }
+        GormImpl gormImpl = GormImpl.parse(object.toString());
+        if (gormImpl == null) {
+            throw new ConversionErrorException(Argument.of(GormImpl.class),
```
codeconsole: Minor/optional: Micronaut's `TypeConverter` contract prefers `context.reject(object, e); return Optional.empty();` over throwing. Throwing `ConversionErrorException` happens to produce the behavior the tests lock in (400 on `/create`, fallback-to-default on the bean-bound feature filter), but it relies on which binder catches the exception. If `reject` + empty still yields the 400 on `/create`, it would be the more robust form — fine as a follow-up.

codeconsole (approved):
Approving — verified locally end-to-end (specs, CLI `create-app --data hibernate7`, and dependency resolution of the generated app confirming the #15942 jandex conflict is gone). Details in my earlier review; the inline comments are minor and can be addressed here or as follow-ups.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/database/DatabaseDriverFeature.java
jdaugherty: the behavior change was not intended. The guard now targets DatabaseDriverConfigurationFeature (implemented only by the two Hibernate features), so mongo + SQL-driver output is byte-identical to pre-PR, and a test locks it.

PR #15947, opened 2026-07-09
Add opt-in deny-by-default data binding and always honor bindable:false

jamesfredley:
## Description

Grails data binding stays **compatible by default** and gains an **opt-in** deny-by-default mode that hardens `bindData` against mass assignment (OWASP / CWE-915). Unconfigured applications keep binding exactly as they did in prior Grails versions; applications that want mass-assignment protection enable secure mode with a single property.

Independently of the default, an explicit `bindable: false` constraint is now honored in **every** mode - including the nested `Map`-constructor fallback and the explicit bind-all path - so a property a developer marked non-bindable can never be mass-assigned.

## The contract

| Case | Behavior |
| --- | --- |
| Unconfigured app (default) | **Permissive** - binds properties as before (backward compatible) |
| `grails.databinding.legacyBindableDefault=false` | **Secure / deny-by-default** - only allowlisted properties bind |
| `bindable: false` on a property | **Never bound**, in either mode (explicit developer denial) |
| App using an allowlist in secure mode | `bindable: true` (incl. `importFrom` / shared constraints), explicit `include:` lists, and `@BindAllowed([...])` are all honored |
| Exclude-only `bindData` (compatibility mode) | Binds every eligible property except the excludes (no class-allowlist intersection) |
| Empty `include` list | Binds nothing, including for direct `SimpleDataBinder` callers (not restored by the compatibility flag) |

## What changed

| Area | Change |
| --- | --- |
| Default | Permissive (compatible) is the unconfigured behavior; `grails.databinding.legacyBindableDefault=false` opts in to deny-by-default |
| `bindable: false` | Always honored in both modes, including the nested `Map`-constructor fallback and the explicit bind-all path |
| Allowlist sources (secure mode) | Generated `bindable: true` allowlist (compile-time and runtime `importFrom` / shared constraints, unioned), explicit `include:` lists, and `@BindAllowed([...])` on action parameters |
| Nested enforcement (secure mode) | Allowlist applies through nested associations, collections/`List`s, object arrays, indexed properties, typed `Map<K,V>` values, JSON-shaped nested objects, and the listener fallback |
| Typed `Map` values | Converted to the declared value type; `beforeBinding` can veto before mutation; getter-only maps are updated in place; conversion failures register binding errors |
| Hot path | Negative include-list results and class-derived unbindable names are cached; instance-derived constraints are not class-cached |
| Config | `Settings.LEGACY_BINDABLE_DEFAULT` is the public key; unrecognised values log a warning and fail closed to secure mode |
| Public API | `bind(...)` overloads normalize a `null` include to the resolved allowlist and an empty include to bind-nothing; intentional bind-all uses a private marker |
| Docs / tests | Permissive default + opt-in secure mode documented; suite verifies the shipping default on the primary `bindData` surface |

## Enabling secure (deny-by-default) mode

```yaml
grails:
  databinding:
    legacyBindableDefault: false
```

## Verification

`:grails-web-databinding:test`, `:grails-databinding-core:test`, and focused `:grails-test-suite-web:test` (including permissive-default, explicit-secure, `bindable: false`, exclude-only, and bind-all regressions) pass.

## Related

| Item | Status |
| --- | --- |
| #15808 (`secureBindData` explicit-API approach) | Superseded by this PR; close once this lands |
| #15950 (`nullMissing` / stale-data clearing) | Stacked on this PR's binding base |

## Contributor Checklist

- [x] Compatible by default, with an opt-in deny-by-default mode; `bindable: false` always honored
- [x] Shipping default covered by `bindData` regressions (not only secure mode)
- [x] Nested/array/map/inheritance regressions plus permissive/secure/bind-all coverage
- [x] Targets `8.0.x`
- [x] Apache License 2.0
- [x] `ai-generated-starting-point` label applied


grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -119,13 +120,19 @@ public static BindingResult bindObjectToInstance(Object object, Object source) {
...
+        List includeList = legacyBindableDefaultEnabled ? Collections.emptyList() : Collections.singletonList(DefaultASTDatabindingHelper.NO_BINDABLE_PROPERTIES);
         try {
             final Class<? extends Object> objectClass = object.getClass();
-            if (CLASS_TO_BINDING_INCLUDE_LIST.containsKey(objectClass)) {
-                includeList = CLASS_TO_BINDING_INCLUDE_LIST.get(objectClass);
+            if (includeListCache.containsKey(objectClass)) {
+                includeList = includeListCache.get(objectClass);
             } else {
-                final Field whiteListField = objectClass.getDeclaredField(DefaultASTDatabindingHelper.DEFAULT_DATABINDING_WHITELIST);
+                final Field whiteListField = objectClass.getDeclaredField(whiteListFieldName);
```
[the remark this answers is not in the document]
jamesfredley: Addressed. `getBindingIncludeList` no longer uses `getDeclaredField` on the target class alone. It now resolves the injected whitelist fields by walking the class hierarchy (a `getField`/`getPublicDeclaredField` helper), so an inherited `public static` allowlist field is correctly found for CGLIB/ByteBuddy/Hibernate proxy subclasses and ordinary subclasses (the deny-by-default binding no longer collapses to bind-nothing for proxied instances).

For security it also does more than a plain inherited lookup: it trusts a `$defaultDatabindingWhiteList` value only when it is paired with a same-class `$legacyDatabindingWhiteList`, so a mixed-generation hierarchy (an old precompiled subclass declaring only the broad default field, extending a Grails 8 parent) cannot have its old permissive whitelist mistaken for a secure generated allowlist. Both the proxy-subclass and mixed-generation paths are covered by new regression tests.

jamesfredley:
## Reworked to deny-by-default (this supersedes the earlier opt-in approach)

Per maintainer direction this PR now makes Grails data binding **deny-by-default** rather than an opt-in flag. It went through an extensive dual security review; here is the final shape.

**The contract**

- **Unconfigured Grails 8 binds only allowlisted properties.** `grails.databinding.legacyBindableDefault=true` restores the previous permissive behavior for the whole application (the documented breaking-change opt-out).
- **Existing correct usage keeps working with no changes.** `bindable: true` constraints (including those resolved at runtime via `importFrom` / shared constraints), explicit `include:` lists, and `@BindAllowed` on action parameters are all honored unchanged.
- **Unsafe usage breaks with clear guidance.** When deny-by-default drops a request parameter, a one-time warning names the property and class and the exact remedies: declare it `bindable: true`, add it to the `include:` list, annotate with `@BindAllowed`, or set `legacyBindableDefault=true`.

**Security enforcement (the review closed each of these bypass paths, all with regression tests)**

- Allowlist applied through every recursive path: nested domain associations, collections/`List`s, object **arrays** (both indexed and non-indexed JSON `children: [[...]]`), indexed properties, typed `Map<K,V>` values, JSON-shaped nested objects, and the listener-rejection fallback - so a nested field (e.g. `admin`) cannot be bound past the child's own allowlist.
- Public `bind(...)` overloads normalized: a `null` include resolves the allowlist and an empty include binds nothing; intentional bind-all flows through a private marker (an empty list no longer silently means bind-all).
- Value types without a no-arg constructor **fail closed** in secure mode instead of mass-assigning through a `Map` constructor (legacy mode preserves the old behavior).
- Inherited allowlist fields resolve for proxy subclasses, and a generated allowlist is trusted only when both generated fields are co-declared on the same class (mixed-generation upgrade safety).
- Persisted domain array/collection elements resolve by id before binding.

**Verification:** `:grails-databinding-core:test`, `:grails-web-databinding:test`, `:grails-test-suite-web:test`, and `:grails-test-suite-persistence:test` all pass.

**Follow-up (non-security, documented):** the nested-collection branches (array / typed map) currently bind elements before the parent `DataBindingListener` veto and bypass the property `bindProperty` setter lifecycle - this is consistent with the pre-existing collection path and is a lifecycle-consistency cleanup for a later PR, not a mass-assignment concern.

Related: #15808 (the earlier `secureBindData` explicit-API approach) can be closed once this lands; #15950 tracks `nullMissing` / stale-data clearing separately.

matrei:
There seems to be some static state leak of `legacyBindableDefault`.
In `HalJsonRendererSpec` and `JsonRendererSpec` the default value `true` does not seem to be active during data binding.

grails-controllers/src/main/groovy/grails/artefact/Controller.groovy
```diff
@@ -363,6 +363,11 @@ trait Controller implements ResponseRenderer, ResponseRedirector, RequestForward
      */
     @Generated
     def initializeCommandObject(final Class type, final String commandObjectParameterName) throws Exception {
+        initializeCommandObject(type, commandObjectParameterName, null)
+    }
+
+    @Generated
+    def initializeCommandObject(final Class type, final String commandObjectParameterName, final List bindAllowedProperties) throws Exception {
```
matrei: Consider not using `final` for method params and local variables?
I know this aligns with previous code but it is really bad for readability for, in my opinion, almost always no gain. New or changed code should not use `final`.

grails-core/src/main/groovy/grails/config/Settings.groovy
```diff
@@ -392,6 +392,8 @@ interface Settings {
 
     String DATE_LENIENT_PARSING = 'grails.databinding.dateParsingLenient'
 
+    String LEGACY_BINDABLE_DEFAULT = 'grails.databinding.legacyBindableDefault'
```
matrei: `String DATABINDING_DENY_BY_DEFAULT = grails.databinding.denyByDefault`?

jamesfredley:
@matrei Good catch - you were right that the default was not active, and the cause was worse than the specs suggested. Fixed in 4c9062e and 08bf7818.

## What was actually happening

It was not leaked state between specs - it reproduced with those two spec classes selected alone and `-PmaxTestParallel=1`. `DataBindingUtils.isLegacyBindableDefaultEnabled()` simply never returned the documented default. A runtime probe on the failing run:

```
[PROBE] branch=FLAT result=false valueClass=org.grails.config.NavigableMap$NullSafeNavigator value=null
[PROBE] branch=FLAT result=true  valueClass=null value=null
```

When there is no `GrailsApplication` in `Holders`, the value came from `Holders.getFlatConfig()`. For an **absent** key a navigable map answers with a `NavigableMap.NullSafeNavigator` placeholder, not `null` - and it renders as the text `null`. So:

```java
value == null || Boolean.TRUE.equals(value) || "true".equalsIgnoreCase(String.valueOf(value))
```

evaluated to `false` for an absent key, and any application without an explicit setting silently ran in secure deny-by-default mode. That is why the renderer specs only passed once `bindable: true` was added to their model classes - those constraints were masking the defect, so they have been removed and the specs now cover the unconfigured-application path again, as intended.

## A second, more serious defect found while reviewing this

The `application != null` branch used `getConfig().getProperty(key, Boolean.class, true)`. In [`NavigableMapConfig.convertValueIfNecessary`](https://github.com/apache/grails-core/blob/8.0.x/grails-core/src/main/groovy/org/grails/config/NavigableMapConfig.java#L343) the converted value is returned only when it is Groovy-truthy:

```java
return DefaultGroovyMethods.asBoolean(value) ? value : defaultValue;
```

A configured `false` that arrives as a **string** converts to `Boolean.FALSE`, which is falsy, so the `true` default was handed back instead. An explicit `grails.databinding.legacyBindableDefault=false` supplied through a properties file, a system property or an environment variable was therefore silently ignored and mass assignment stayed enabled. A YAML boolean happened to work only because it took the `targetType.isInstance` short-circuit.

## The fix

Both branches now read the **raw** value and share one resolver:

| Configured value | Result |
| --- | --- |
| absent (`null` or the `NullSafeNavigator` placeholder) | permissive (documented default) |
| `Boolean` | its own value |
| string `true` (trimmed, case-insensitive) | permissive |
| anything else, including unparseable input | secure - fails closed |

Explicitly-supplied values keep exactly their previous meaning; only the absent-key case changes, plus string-sourced values are now honored in the application branch. Unrecognised input deliberately fails closed rather than throwing, since raising at bind time on a typo'd setting would be a worse regression than choosing the safe mode.

## Coverage

New `LegacyBindableDefaultConfigSpec` (19 iterations) pins all of it, including an assertion that the absent-key lookup really does return a `NullSafeNavigator`, application-supplied `false` / `'false'` / `'FALSE'` / `' false '` all selecting secure mode, and unrecognised values failing closed. It was verified to fail against the old expression and pass against the new one.

Green: `:grails-web-databinding:test`, `:grails-databinding-core:test`, `:grails-test-suite-web:test`, `:grails-test-suite-persistence:test`, `:grails-rest-transforms:test`, `:grails-controllers:test`, plus `:grails-web-databinding:check` and `:grails-rest-transforms:check`.

borinquenkid:

grails-test-suite-web/src/test/groovy/org/grails/web/servlet/BindDataMethodTests.groovy
```diff
@@ -28,6 +36,10 @@ import spock.lang.Specification
  */
 class BindDataMethodTests extends Specification implements ControllerUnitTest<BindingController> {
 
+    void setup() {
+        grailsApplication.config.setAt(DefaultASTDatabindingHelper.LEGACY_BINDABLE_DEFAULT, false)
```
jdaugherty: This flips the whole spec into the opt-in secure mode, including the eight pre-existing `bindData` regressions below it (`Test bindData with Map`, `With Excludes`, `With Includes`, `Overriding Included With Excluded`, `With Prefix Filter`, `With Disallowed And GrailsParameterMap`, `With Prefix Filter And Disallowed`, `Converts Single String In Map To List`).

Those were the regression coverage for the default binding path. After this change nothing in the file exercises the shipping default except the handful of specs that explicitly set the flag back to `true`/`null`.

Please leave the pre-existing features running on the unconfigured default and scope the secure-mode configuration to the new specs — either a separate spec class, or set the flag in the `given:` block of the tests that need it. The new secure-mode cases should also have unconfigured-default counterparts, since "an unconfigured application binds as before" is the compatibility guarantee this PR makes and it currently has no assertion covering the `bindData` surface.

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DataBindingTests.groovy
```diff
@@ -454,6 +454,11 @@ class DataBindingTests extends Specification implements ControllerUnitTest<TestC
 class Clown {
     String name
     String hairColour
+
+    static constraints = {
+        name bindable: true
```
jdaugherty: These `bindable: true` additions are not required by the new default.

I reverted the constraint-only fixture edits on this branch — `DataBindingTests`, `JSONBindingToNullSpec`, `BindCommandObjectsSpec`, `BindToObjectWithEmbeddableTests`, `GrailsParameterMapBindingSpec`, `BindingToNullableTests`, `DirtyCheckBaseBindDataSpec`, `ControllerExceptionHandlerSpec`, `JSONConverterTests`, `RespondMethodSpec`, `SomeValidateableClass`, the `org.grails.web.binding.*` / `commandobjects` / `json` / `mime` specs, and `DataBindingConfigurationSpec` — and re-ran them: 25 classes, 151 tests, all green.

Please drop them. As written, every one of these fixtures binds identically with the flag on or off, so the specs can no longer fail if the compatible path regresses. Where a fixture genuinely needs an allowlist to demonstrate secure mode, it belongs in a new spec rather than in a shared fixture that other tests rely on for default-mode behavior.

grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
```diff
@@ -162,14 +209,77 @@ class GrailsWebDataBinder extends SimpleDataBinder {
         boolean bind = listenerWrapper.beforeBinding(object, bindingResult)
 
         if (bind) {
-            super.doBind(object, source, filter, whiteList, blackList, listenerWrapper, bindingResult)
+            List previousIncludeList = bindingIncludeList.get()
+            Class<?> previousTargetType = bindingTargetType.get()
+            bindingIncludeList.set(whiteList)
+            bindingTargetType.set(object.getClass())
+            try {
+                super.doBind(object, source, filter, whiteList, addUnbindablePropertyNames(object, blackList), listenerWrapper, bindingResult)
```
jdaugherty, line 221: `addUnbindablePropertyNames` runs on every bind, in both modes, and nothing about it is cached.

It calls `getConstrainedProperties(object)`, which tries up to four strategies in order — the `constraintsMap` property, `getConstraintsMap()`, the `constraints` property, then `Class.forName("org.grails.web.plugins.support.ValidationSupport")` plus a reflective `getConstrainedPropertiesForClass` — each wrapped in `catch (Exception ignored)`. For a target with no constraints, which is the common case for plain command objects, that is three thrown-and-discarded exceptions plus a reflective class load per bind. `getBindableConstraintValue` then repeats the same three-fallback pattern for every constrained property.

`bindable: false` is a per-class, compile-time-known fact. Please resolve it once per class into a `ConcurrentHashMap`, the way `getBindingIncludeList` already caches its result, so request binding does not pay this on the hot path.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -119,30 +126,290 @@ public static BindingResult bindObjectToInstance(Object object, Object source) {
...
+                        combinedIncludeList.addAll(runtimeBindableNames);
+                    }
+                    includeList = new ArrayList(combinedIncludeList);
+                }
+                if (!legacyBindableDefaultEnabled) {
+                    includeList = asGeneratedBindingIncludeList(includeList);
                 }
-                if (!Environment.getCurrent().isReloadEnabled()) {
-                    CLASS_TO_BINDING_INCLUDE_LIST.put(objectClass, includeList);
+                if (includeList != null && !Environment.getCurrent().isReloadEnabled()) {
```
jdaugherty: Guarding the cache write on `includeList != null` means the "this class has no allowlist field" result is never cached.

In the default mode that is the outcome for every target the AST helper did not enhance, so each bind re-runs `getField` twice over the whole superclass chain, and `getPublicDeclaredField` throws and swallows a `NoSuchFieldException` for each class/field pair that misses. The previous implementation cached `Collections.emptyList()` and did the reflection once per class.

Please cache the negative result as well — either a sentinel value, or keep returning an empty list for "nothing found" and reserve `null` for a genuine resolution failure.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -212,9 +479,12 @@ public static <T> void bindToCollection(final Class<T> targetType, final Collect
      * @return A BindingResult if there were errors or null if it was successful
      */
     public static BindingResult bindObjectToInstance(Object object, Object source, List include, List exclude, String filter) {
-        if (include == null && exclude == null) {
+        if (include == null) {
```
jdaugherty, line 518: Dropping `&& exclude == null` changes the default mode too, not only secure mode.

`bindData(target, params, [exclude: ['email']])` — and the `bindData(target, bindingSource, List excludes)` overload that delegates to it — previously bound every property except the excludes, with no allowlist applied. It now intersects with the class allowlist as well, so anything absent from the generated list silently stops binding: dynamically typed properties, `id`/`version` on a domain class, and properties the AST helper did not record.

`bindable: false` is already enforced independently by `addUnbindablePropertyNames` in `GrailsWebDataBinder.doBind`, so this line is not what closes that gap. If the narrowing is intentional it needs an entry in the upgrade notes and a default-mode test; otherwise please restore the `exclude == null` condition. Note that `Test bindData With Excludes`, which would have caught this, now runs in secure mode.

grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy
```diff
@@ -269,7 +271,18 @@ class SimpleDataBinder implements DataBinder {
     }
 
     protected boolean isOkToBind(String propName, List whiteList, List blackList) {
```
jdaugherty, line 273: `SimpleDataBinder` is public API in `grails-databinding-core` and is used outside the Grails web binding stack.

Changing `!whiteList` to `whiteList == null` flips an empty include list from "no restriction" to "bind nothing" for every caller of `bind(obj, source, whiteList)`. Unlike the rest of this PR, that flip is not gated by `grails.databinding.legacyBindableDefault` — an application that stays on the compatible default still gets the new behavior, and setting the flag to `true` does not restore the old one.

The upgrade note says "Empty `include` lists bind no properties" but does not say it reaches direct `SimpleDataBinder` callers or that the compatibility flag does not cover it. Please state both, and add a `SimpleDataBinderSpec` case pinning the new semantics at this level — the current coverage only goes through `bindData`.

grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy
```diff
@@ -394,6 +415,16 @@ class SimpleDataBinder implements DataBinder {
         }
     }
 
+    protected Object instantiateAndBindOrUseMapConstructor(Class referencedType, Map values, DataBindingListener listener) {
```
jdaugherty, line 422: The `try` covers the recursive `bind(...)` call as well as the instantiation.

A `NoSuchMethodException` or `IllegalAccessException` raised anywhere inside nested binding is therefore caught here and retried through the `Map` constructor, binding the same source twice into two different instances and returning the second. `GrailsWebDataBinder.instantiateAndBindNestedOrUseMapConstructor` has the same shape; there the consequence in secure mode is that the element is dropped and the original cause is lost entirely.

Please narrow the guarded region to `referencedType.getDeclaredConstructor().newInstance()` and move the `bind` call outside the catch.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -119,30 +126,290 @@ public static BindingResult bindObjectToInstance(Object object, Object source) {
...
+     * any string-valued source such as a properties file, a system property or an environment variable.
+     * <p>
+     * A navigable config answers an absent key with a placeholder object rather than {@code null}, so
+     * only a genuinely absent key may fall back to the permissive default. Any other unrecognised value
+     * fails closed, because this switch governs mass-assignment protection.
+     *
+     * @param value the raw configured value, which may be {@code null} or an absent-key placeholder
+     * @return true when the legacy (permissive) binding default applies
+     */
+    static boolean resolveLegacyBindableDefault(final Object value) {
```
jdaugherty: Failing closed on a value that cannot be parsed is defensible, but it fails closed silently.

`LegacyBindableDefaultConfigSpec` pins `''`, `0`, `1`, `'off'` and `'no'` as switching the application into deny-by-default binding; by the same rule so do `'yes'` and `'on'`, which a reader would expect to mean *true*. The only feedback is a per-property WARN, emitted after requests have already started dropping fields.

`spring-configuration-metadata.json` declares the property as `java.lang.Boolean`, which is what IDE completion shows and what Spring's relaxed binding accepts, so `on`/`yes` are plausible things for someone to write. Please log a warning naming the property and the unrecognised value when the raw value does not resolve, so a typo is visible immediately rather than inferred from missing data.

grails-core/src/main/groovy/grails/config/Settings.groovy
jdaugherty: This constant is added but never referenced. `DataBindingUtils`, `GrailsWebDataBinder`, `BindDataMethodTests` and `LegacyBindableDefaultConfigSpec` all use `DefaultASTDatabindingHelper.LEGACY_BINDABLE_DEFAULT`, which lives in an internal `org.grails.web.databinding` class.

Please keep the single public definition here and have the internal code and the specs reference it, so the key has one definition and applications have a supported constant to use.

grails-web-databinding/src/test/groovy/grails/web/databinding/LegacyBindableDefaultConfigSpec.groovy
```diff
@@ -0,0 +1,168 @@
...
+        Holders.grailsApplication = originalApplication
+    }
+
+    void 'an absent legacy bindable default in a PropertySourcesConfig remains permissive'() {
+        given:
+        Holders.setConfig(new PropertySourcesConfig([:]))
+
+        when:
+        Object value = Holders.flatConfig.get(DefaultASTDatabindingHelper.LEGACY_BINDABLE_DEFAULT)
+        boolean legacyDefaultEnabled = DataBindingUtils.isLegacyBindableDefaultEnabled()
```
jdaugherty: `isLegacyBindableDefaultEnabled()` is package-private internal API, and this spec sits in `grails.web.databinding` only in order to reach it. Per the repository's testing rule, specs should drive behavior through the API an application actually calls.

Every case here can be written as "given this config, bind this source, assert which properties were set", which covers the resolution *and* that it is wired up correctly. As it stands, a change that resolved the flag correctly but applied it to the wrong code path would still pass this spec.

Separately: `setup()` nulls `Holders.grailsApplication` and `Holders.config` process-wide. `cleanup()` restores them, but together with the static caches in `DataBindingUtils` and `GrailsWebDataBinder.WARNED_BINDING_SHAPES` — neither of which is reset — this leaves mode-dependent state that later specs in the same fork inherit.

grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
```diff
@@ -397,6 +556,39 @@ class GrailsWebDataBinder extends SimpleDataBinder {
                         }
                     }
                 }
+            } else if (Map.isAssignableFrom(metaProperty.type) && val instanceof Map) {
+                def referencedType = getReferencedTypeForCollection(propName, obj)
+                if (referencedType != null) {
+                    needsBinding = false
+                    Map map = initializeMap(obj, propName)
+                    map.clear()
```
jdaugherty: This branch is new behavior for any `Map`-typed property with a resolvable value type: previously the raw source map was assigned through `bindProperty`/`setPropertyValue`, and now entries are converted or recursively bound. Two follow-ups:

- It bypasses `bindProperty`, so `DataBindingListener` callbacks and error registration do not fire for this property. The object-array branch a few lines above does call `bindProperty`. A conversion failure inside this loop is invisible to listeners and to the `BindingResult`.
- Typed-map value conversion is a user-visible change independent of the mass-assignment work, so it needs an upgrade-note line and a default-mode test. `Test bindData in explicit secure mode applies the nested target allowlist` only covers it with the flag on.

jdaugherty (commented):
Making this opt-in and keeping `bindable: false` unconditional is the right shape, and the nested/collection/map enforcement is thorough.

My main concern is that the suite no longer verifies the shipping default. `BindDataMethodTests` — the primary `bindData` regression spec — is flipped to `legacyBindableDefault=false` for every feature via a new `setup()`, and roughly 25 pre-existing fixture classes across `grails-test-suite-web`, `grails-databinding` and `grails-test-suite-persistence` gained `bindable: true` on every property.

I reverted the constraint-only fixture edits on this branch and re-ran the affected specs: 25 classes / 151 tests pass unchanged. So the new constraints are not required by the default. But adding them means those fixtures bind identically in both modes, so none of those specs can fail if the compatible path regresses — which is the claim this PR rests on.

Beyond the test coverage, there are a few behavior and cost changes that reach the default path as well. Details inline.

jamesfredley:
@jdaugherty Addressed in 73a0555772 (after merging latest `8.0.x`).

## Review responses

| Item | Resolution |
| --- | --- |
| Suite flipped to secure mode | Removed the global `setup()` that forced `legacyBindableDefault=false`. Pre-existing `bindData` regressions run on the unconfigured permissive default again. Secure-mode cases call `enableSecureBinding()` in their own `given:`. Added unconfigured counterparts for the main compatibility claims. |
| Unnecessary `bindable: true` on fixtures | Dropped the constraint-only fixture edits across `grails-test-suite-web`, `grails-databinding`, and `grails-test-suite-persistence`. Shared fixtures no longer bind identically in both modes. |
| Cache `bindable: false` / unbindable names | Class-level `ConcurrentHashMap` for class-evaluated unbindables. Object-derived constraints are **not** class-cached (instance maps can differ). |
| Cache negative include-list results | Cache miss of "no allowlist field" is stored as a sentinel and returned as `null` on hit. |
| `exclude == null` dropped | Restored. Exclude-only in compatibility mode uses the bind-all marker so it does not intersect the class allowlist; secure mode still applies the allowlist when only excludes are supplied. Covered by default-mode tests. |
| `SimpleDataBinder` empty whiteList | Documented that empty include binds nothing for direct callers and that the compatibility flag does not restore the old empty-list meaning. Added `SimpleDataBinderSpec` coverage. |
| try covers recursive `bind` | Narrowed to `getDeclaredConstructor().newInstance()` in both `SimpleDataBinder` and `GrailsWebDataBinder`. |
| Silent fail-closed on bad config | Unrecognised `legacyBindableDefault` values log a WARN naming the property and value, then fail closed. |
| `Settings.LEGACY_BINDABLE_DEFAULT` unused | Single public definition in `Settings`; internal helper delegates to it; specs use `Settings`. |
| `LegacyBindableDefaultConfigSpec` | Rewritten to drive public bind APIs; clears binding caches and warning state in cleanup. |
| Typed `Map` branch | Builds converted entries only after `beforeBinding` on the source value; mutates the target map in place (getter-only safe); always calls `afterBinding`; conversion failures go through `addBindingError`. Upgrade notes + default-mode typed-map test added. |

## Verification

`:grails-databinding-core:test`, `:grails-web-databinding:test`, and focused `:grails-test-suite-web:test` (`BindDataMethodTests`, `CommandObjectsSpec`, `DataBindingTests`, `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec`, `LegacyBindableDefaultConfigSpec`, `SimpleDataBinderSpec`) all green.

sbglasius (approved):
Used AI to review and it had the following comment:

1. **[SimpleDataBinder.groovy:273](https://github.com/apache/grails-core/blob/b89e6a02dda9f85ab3ecb923e2d997e702ce9989/grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy#L273)** — `isOkToBind(String,...)` only checks the `.` wildcard suffix, but the PR's new whitelist generator also emits `_` suffix markers, so underscore/indexed-convention properties get wrongly denied under deny-by-default mode.

2. **[DefaultASTDatabindingHelper.java:250](https://github.com/apache/grails-core/blob/b89e6a02dda9f85ab3ecb923e2d997e702ce9989/grails-web-databinding/src/main/groovy/org/grails/web/databinding/DefaultASTDatabindingHelper.java#L250)** — `getLegacyPropertyNamesToIncludeInWhiteList` duplicates ~80 lines of `getPropertyNamesToIncludeInWhiteList` instead of sharing the traversal.
 
3. **[DataBindingUtils.java:556](https://github.com/apache/grails-core/blob/b89e6a02dda9f85ab3ecb923e2d997e702ce9989/grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java#L556)** — `bindObjectToInstance` and `bindObjectToDomainInstance` duplicate the same include-list normalization block, and the former's copy is redundant since it delegates straight into the latter.

But otherwise it looks good to me.

matrei (approved):

PR #15948, opened 2026-07-09
Split CLI dependencies from the runtime classpath

### LINKED ISSUE #15377: Grails 8 - grails-data-hibernate5-dbmigration bootWar: Cannot invoke "String.split(String)" because "sources" is null -  In a classic servlet-container WAR deployment
`org.grails.cli.boot.SpringApplicationWebApplicationInitializer` class (causing the NPE during Tomcat WAR deployment) is being transitively included via the Grails Database Migration functionality. Specifically:

- This originates from `org.apache.grails:grails-data-hibernate5-dbmigration`
- In Grails 7+, the Database Migration plugin depends on `org.apache.grails:grails-shell` for certain scripts and commands (e.g., Liquibase-related operations like `dbm-*` scripts).
- `grails-shell` pulls in Spring Boot CLI components, including the problematic initializer that's only intended for standalone executable JARs—not traditional WAR deployments in an external container like Tomcat.

The presence of this transitive dependency explains why the class is on the classpath and gets invoked by Tomcat during context startup, leading to the null "sources" in `MANIFEST.MF`.

Workaround for Grails 7, build.gradle:

```
bootWar {
    classpath = classpath.filter { 
        !it.name.contains('spring-boot-cli') && 
        !it.name.contains('grails-shell')
    }
}
```

Potential updates for Grails 8:

Option 1: Gradle Feature Variant 
In the plugin's build.gradle:

```
java {
    registerFeature('cli') {
        usingSourceSet(sourceSets.main)
    }
}
dependencies {
    // Core - always included
    implementation("org.liquibase:liquibase-core:${liquibaseVersion}")
    implementation("org.liquibase.ext:liquibase-hibernate5:${liquibaseVersion}")
    
    // CLI feature - opt-in
    cliApi("org.apache.grails:grails-shell-cli:${grailsVersion}") {
        exclude group: 'org.slf4j', module: 'slf4j-simple'
        exclude group: 'org.codehaus.groovy'
    }
}
```

Consuming apps would then:

```
// Core functionality (in WAR)
implementation("org.apache.grails:grails-data-hibernate5-dbmigration:7.0.7")
// CLI commands (dev only)
developmentOnly("org.apache.grails:grails-data-hibernate5-dbmigration:7.0.7") {
    capabilities {
        requireCapability("org.apache.grails:grails-data-hibernate5-dbmigration-cli")
    }
}

```
---

Option 2: Use compileOnlyApi + Document for Apps (Simpler)
In the plugin's build.gradle:

```
// Available at compile time for the plugin, but consumers must explicitly add it
compileOnlyApi("org.apache.grails:grails-shell-cli:${grailsVersion}") {
    exclude group: 'org.slf4j', module: 'slf4j-simple'
    exclude group: 'org.codehaus.groovy'
}

```
Consuming apps add:

`developmentOnly("org.apache.grails:grails-shell-cli:7.0.7")`

---
Option 3: Separate Artifact
Split into two published artifacts:
- grails-data-hibernate5-dbmigration - core runtime (no CLI)
- grails-data-hibernate5-dbmigration-cli - CLI commands (depends on core + grails-shell-cli)
Apps use:

```
implementation("org.apache.grails:grails-data-hibernate5-dbmigration:7.0.7")
developmentOnly("org.apache.grails:grails-data-hibernate5-dbmigration-cli:7.0.7")
```

jdaugherty:
## Description

Grails commands (`ApplicationCommand` implementations) have always shipped inside runtime plugin
artifacts, dragging CLI-only dependencies onto application runtime classpaths. This causes real
failures: `grails-shell-cli` leaks a conflicting Groovy onto the app classpath (see the
long-standing exclusion + TODO in `grails-data-hibernate{5,7}/dbmigration/build.gradle`), and a
CLI-only Spring Boot `WebApplicationInitializer` breaks classic WAR deployment with an NPE
(#15377) unless users hand-filter the `bootWar` classpath.

(no classifiers or capability syntax for consumers), fully consumable by Maven as well as Gradle.

### Framework changes

- **`grails-core-cli`**: the command contract moves from `grails.dev.commands.*` to
  `org.apache.grails.core.cli.*`. Command registrations move from `META-INF/grails.factories` to a
  dedicated `META-INF/grails-cli.factories`, written by a new `CommandFactoriesTransformation`
  that ships in the cli artifact (so it is active exactly when a command can compile). The default
  `grails-core` jar contains no command code and no command contract.
- **Companion artifacts** for all command-bearing modules, each with a unique package and
  `Automatic-Module-Name`: dbmigration (h5/h7, 29 `dbm-*` commands each), the hibernate plugins
  (`schema-export`), scaffolding (9 `generate-*` commands), web-url-mappings
  (`url-mappings-report`), spring-security, and spring-security-oauth2. `grails-shell-cli` is gone
  from every runtime dependency graph.
- **BOM**: the base BOM enumerates and manages every companion coordinate.

### Build/tooling changes

- **New `org.apache.grails.gradle.grails-cli` plugin** (applied automatically by the Grails
  application/plugin plugins): registers the `grailsCli` configuration — compile-visible for
  `grails-app/commands` sources and on the command-runner and test classpaths, but never on the
  main `runtimeClasspath`, so commands and their CLI-only dependencies are excluded from
  `bootRun`/`bootJar`/`bootWar`. It auto-provisions `grails-core-cli` + `grails-console`, and
  **discovers plugin companions automatically**: a command-bearing plugin advertises its companion
  through a `Grails-Cli-Artifact` manifest attribute on its runtime jar, and the plugin walks the
  resolved dependency graph (including transitive plugins) and adds each advertised companion to
  `grailsCli`. Per-command Gradle tasks (`dbmUpdate`, …) register from the discovered companions —
  no `buildscript { classpath }` entry needed. Opt out with `grails { cliAutoProvision = false }`.
  The `console`/`shell` tasks draw from the same provisioned tier, so generated apps no longer
  declare `console "org.apache.grails:grails-console"`.
- **New `org.apache.grails.gradle.grails-plugin-cli` plugin** for plugin authors (dogfooded by the
  framework's own modules): one `apply` sets up the cli source set as a feature variant, the
  command contract on its compile classpath, the manifest advertisement, and the companion
  publication.
- **grails-publish**: builds on the additional-publications support added in
  apache/grails-gradle-publish#34 (requires `grails-publish` ≥ `1.0.0-SNAPSHOT`; the version is
  bumped in `dependencies.gradle`).
- **Forge/profiles**: generated apps declare only runtime plugins; all CLI wiring is automatic.

### Consumer impact

Most applications upgrade with **no build changes**: declaring a command-bearing plugin is enough,
and its commands (and Gradle tasks) follow automatically. Application commands in
`grails-app/commands` only need the import rename (`grails.dev.commands.*` →
- [x] This PR targets the **correct branch** for the type of change (`8.0.x` — breaking changes are part of the major release).

### Code Quality

- [x] I have **added or updated tests** that cover the changes introduced in this PR (module unit
      suites, dbmigration integration tests, the config-report end-to-end integration test
      exercising the full discovery → factories → runner chain, forge feature specs, and
      grails-gradle plugin tests).
- [x] I have verified that all existing tests pass by running `./gradlew build --rerun-tasks`.
- [x] My code follows the project's **code style** guidelines (`aggregateViolations` reports zero
      Checkstyle/CodeNarc/PMD/SpotBugs findings).
- [x] This PR does **not** include mass reformatting, style-only changes, or large-scale refactoring
      beyond the approved scope.

### Licensing and Attribution

- [x] All contributed code is provided under the Apache License 2.0, and new source files include the **Apache license header**.
- [x] If generative AI tooling was used in preparing this contribution, I have followed the ASF policy on generative tooling and have properly attributed its use.

### Documentation

- [x] User-facing changes are documented (Creating Custom Commands, Providing Basic Artefacts,
      Gradle plugins catalog, `create-command` reference).
- [x] The **What's New** section covers the companion `-cli` artifacts, automatic discovery, and the new plugins.
- [x] The **Upgrade Notes** (section 33) cover the package renames, the `grails-cli.factories`
      clean break, the `grailsCli` configuration, and the third-party plugin migration path.


jdaugherty:
These were resolving locally due to having maven local search on; I need to fix the dependency substitution to fix these resolution issues.

jdaugherty:
Turns out there's a real build issue separate from this change, micronaut-singleton isn't including our functional test app so it was pulling from upstream instead of testing with the local copy.

jdaugherty:
The published Gradle Module Metadata still contains capability-style dependencies (org.apache.grails:grails-core + requireCapability('org.apache.grails:grails-core-cli')) — violating the split's design rule that external consumers must see the plain -cli coordinate, never capabilities. When Gradle resolves that from a Maven repo, it selects grails-core's cliRuntimeElements variant, follows its available-at redirect to the standalone grails-core-cli module, and then sees both nodes providing the same capability — a self-conflict. I reproduced it with a one-dependency project against build/local-maven, and it fails with both timestamped and non-unique snapshots, so it's not a snapshot-timestamp issue.

There are two distinct leak paths:

1. Every -cli companion (grails-scaffolding-cli, grails-data-hibernate5-cli, grails-web-url-mappings-cli, dbmigration, spring-security plugins): configureDependencyMapping (publishResolvedCoordinates in GrailsCliArtifactGradlePlugin) rewrites the POM correctly to grails-core-cli, but the GMM still records grails-core + requestedCapabilities — and Gradle consumers prefer the GMM.
2. grails-console declares the capability dep in its main api (grails-console/build.gradle:43). Its main java component has no dependency mapping at all, so its GMM leaks the capability dep — and worse, its POM collapses it to plain grails-core, meaning Maven consumers silently lose the cli dependency entirely.


I'm still researching possible solutions for this one.

jdaugherty:
This is now fixed.  We have to support cases where the library is a regular library imported by another project that uses the requireCapbility - this is only when the projects are defined in the same build.  I've created a dedicated plugin for this scenario.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
```diff
@@ -0,0 +1,435 @@
...
+     * the (possibly not yet built) jar is not read — the coordinate comes from the project's
+     * {@code cliArtifactId} property exported by the cli-artifact convention.
+     */
+    @CompileDynamic
+    protected String findAdvertisedCliArtifact(Project project, ResolvedArtifactResult artifact) {
+        def componentIdentifier = artifact.id.componentIdentifier
+        if (componentIdentifier instanceof ProjectComponentIdentifier) {
+            Project target = project.rootProject.findProject(((ProjectComponentIdentifier) componentIdentifier).projectPath)
+            def cliArtifactId = target?.findProperty('cliArtifactId')
+            return cliArtifactId ? "${target.group}:${cliArtifactId}" as String : null
```
codeconsole: **Blocking:** companions advertised by a same-build project dependency are wired as unversioned *external module* dependencies, so auto-discovery is effectively broken for the plugin-development workflow.

Both branches of `findAdvertisedCliArtifact` return a bare `group:artifactId` string, and `autoProvisionCliDependencies` always adds it via `dependencies.create(companion)`. For a `ProjectComponentIdentifier` result this means Gradle tries to resolve the companion from a repository instead of binding to the sibling project's `cli` feature variant. Concretely: a multi-project build with `include 'app', 'my-plugin'` where `my-plugin` applies `grails-plugin-cli` and is unpublished — resolving `app`'s classpath fails (or silently picks up a stale previously-published artifact of the same coordinate).

The framework's own build corroborates the gap: root `build.gradle` disables auto-provisioning for every framework module, and `gradle/functional-test-config.gradle` adds a bespoke `dependencySubstitution` rule that performs exactly the project+capability wiring this method should produce itself.

Composite builds (`includeBuild`) hit a related edge: the included build's component is a `ProjectComponentIdentifier` that `project.rootProject.findProject(projectPath)` either fails to resolve (companion silently skipped) or resolves to a same-named project in the *wrong* build.

Suggested fix: when the component is a project of the current build, add `project.dependencies.project(path)` with `capabilities { requireCapability(...) }`; keep the coordinate path only for `ModuleComponentIdentifier` results. Alternatively, if in-build discovery is intentionally out of scope, document that and point at a substitution recipe.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
```diff
@@ -0,0 +1,435 @@
...
+     * {@code runCommand} task can always execute those commands regardless.
+     */
+    @CompileDynamic
+    protected Collection<String> loadCommandNamesFromCliClasspath(Project project) {
+        Set<String> names = new LinkedHashSet<String>()
+        Configuration cliClasspath = project.configurations.findByName(GRAILS_CLI_CLASSPATH_CONFIGURATION)
+        if (cliClasspath == null) {
+            return names
+        }
+        for (File file : cliClasspath.incoming.artifactView { it.lenient(true) }.files) {
```
codeconsole, line 343: This forces full resolution of `grailsCliClasspath` (and, transitively, the `grailsCliDetect` probe via `withDependencies`) inside `project.afterEvaluate` — i.e. at configuration time, on every invocation. `./gradlew help` or an unrelated test run pays for resolving `grails-core-cli`, `grails-console`, and every discovered companion, and an offline/registry failure silently drops per-command task registration rather than failing loudly. It is also likely hostile to the configuration cache.

I realize task names must be known at configuration time to `tasks.register(...)`, so some of this is a genuine design tension — but the resolution result should at least be cached/scoped so it is not repeated for builds that never touch a CLI task, and the lenient-swallow failure mode deserves a warning log at minimum.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliArtifactGradlePlugin.groovy
```diff
@@ -0,0 +1,216 @@
...
+        CliArtifactExtension extension = project.extensions.create('cliArtifact', CliArtifactExtension)
+        extension.artifactId.convention(project.provider { "${project.name}-cli" as String })
+        extension.defaultDependencies.convention(true)
+
+        SourceSetContainer sourceSets = project.extensions.getByType(SourceSetContainer)
+        SourceSet cliSourceSet = sourceSets.create(CLI_SOURCE_SET_NAME)
+
+        JavaPluginExtension java = project.extensions.getByType(JavaPluginExtension)
+        java.registerFeature(CLI_SOURCE_SET_NAME) { spec ->
+            // no explicit capability: the default (`${group}:${project.name}-cli:${version}`)
```
codeconsole, line 98: Latent contract mismatch: `registerFeature` here keeps the Gradle-computed default capability (`${group}:${project.name}-cli`), but `CliArtifactExtension.artifactId` is documented as customizable and renames the published jar and advertised coordinate. A consumer who sets `cliArtifact { artifactId = 'custom-name' }` and then requests `requireCapability("$group:custom-name")` in-build will fail to resolve, because the in-build capability never followed the extension.

Nothing in the framework customizes `artifactId` today, so this is unexercised — but either wire `spec.capability(...)` from the extension value or document the limitation on `CliArtifactExtension.artifactId`.

dependencies.gradle
```diff
@@ -30,7 +30,7 @@ ext {
             'directory-watcher.version'     : '0.19.1',
             'gradle-groovy.version'         : '4.0.32',
             'gradle-spock.version'          : '2.4-groovy-4.0',
-            'grails-publish-plugin.version' : '1.0.0-M1',
+            'grails-publish-plugin.version' : '1.0.0-SNAPSHOT',
```
codeconsole, line 33: **Blocking for release:** this `1.0.0-SNAPSHOT` flows into `gradleBomDependencies` and from there into the `constraints { api ... }` block of `grails-bom/base`, i.e. into the published `<dependencyManagement>` of `org.apache.grails:grails-bom` that every application consumes. A staged 8.0.x release would ship a BOM recommending a SNAPSHOT of `org.apache.grails.gradle:grails-publish`.

I understand this is pending apache/grails-gradle-publish#34 shipping — but the PR should be explicitly marked blocked on that release, with the version reverted to a GA coordinate before merge (or at latest before the first 8.0.x release candidate).

codeconsole (commented):
This is the right architecture for the CLI split, and most of the execution is excellent for a 260-file change: the renames are mechanically clean (h5/h7 command twins verified line-identical modulo package, no `grails.dev.commands` stragglers outside migration docs, license headers preserved), the BOM enumerates companion coordinates automatically via the lazy `cliArtifactId` sweep so there is nothing to keep in sync by hand, the AST transform ordering is verifiably correct with the `ConfigReportCommand` bootstrap problem cleanly solved, the docs match the implemented behavior including an honest upgrade note on the clean break, and the `grailsCli` configuration scoping is right (compile/test visible, absent from `runtimeClasspath`/`bootJar`/`bootWar`). The long-standing Groovy-conflict TODO in the dbmigration builds is genuinely resolved rather than papered over.

I left inline comments on the issues I think need resolution before merge — the two blocking ones are the project-dependency companion wiring in `GrailsCliGradlePlugin` and the `1.0.0-SNAPSHOT` of grails-publish flowing into the published BOM. Details and suggested fixes are on the relevant lines.

A few minor items not worth their own threads:

- `GrailsApplicationContextCommandRunner`'s pre-existing `Command not found for name: ...` message is now the symptom users will see when a companion was not provisioned; mentioning the `grailsCli` configuration / `-cli` artifact in that message would save support traffic.
- `FactoriesFileWriter.addToProps` carries over the old substring-based dedup check (`existing.contains(classNodeName)`), which miscounts when one FQCN is a prefix of another. Pre-existing, not a regression, but this refactor was the natural moment to fix it.
- Four spring-security command files lost their `@author Puneet Behl` javadoc tags during the rename (e.g. `S2QuickstartCommand`) — worth confirming that was intentional.

grails-bom/base/build.gradle
```diff
@@ -97,6 +97,23 @@ dependencies {
...
+// applying project exports its companion coordinate via the `cliArtifactId` extra property, which
+// only exists once that project has been evaluated — so the constraints are computed lazily, in
+// the mutation window Gradle provides right before the configuration is first observed.
+configurations.named('api').configure { apiConfiguration ->
+    apiConfiguration.withDependencies {
+        for (Project subproject : rootProject.subprojects) {
+            def cliArtifactId = subproject.findProperty('cliArtifactId')
+            if (cliArtifactId) {
+                apiConfiguration.dependencyConstraints.add(
+                        project.dependencies.constraints.create("${subproject.group}:${cliArtifactId}:${projectVersion}"))
```
codeconsole: Related to the `grails-publish` SNAPSHOT in `dependencies.gradle`: note that `validateNoSnapshotDependencies` (below in this file) only iterates `config.allDependencies` — constraints created here via `dependencies.constraints.create(...)` (and the `constraints { api ... }` block above) live in `allDependencyConstraints`, so the release guard has a blind spot for exactly this kind of leak and would let a SNAPSHOT constraint through a release build unflagged. Worth extending the guard to scan constraints as part of this PR, since the new lazy constraint mechanism widens that surface.

grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/commands/CliCompanionPublishingSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+ * Verifies that cli-tier dependencies declared with a capability
+ * ({@code project(':owner') { capabilities { requireCapability('g:owner-cli') } }}) are published
+ * as plain dependencies on the companion coordinate ({@code owner-cli}) in both the pom and the
+ * Gradle Module Metadata, and that the published artifacts resolve from a Maven repository.
+ *
+ * A published capability request cannot be resolved against the published component tree: the
+ * owner's cli variant and its {@code available-at} redirect target both provide the capability,
+ * so consumers fail with a capability self-conflict (the grails-forge generated-app failure).
+ */
+class CliCompanionPublishingSpec extends GradleSpecification {
```
codeconsole, line 35: The publishing side is well covered here (publish → parse `.module`/`.pom` → resolve from a fresh repo — nice), but the discovery/wiring half of the feature has no coverage at all: nothing tests `autoProvisionCliDependencies`/`findAdvertisedCliArtifact`, the guarantee that `grailsCli` never leaks into `runtimeClasspath`/`bootJar`/`bootWar`, per-command task registration, or the `cliAutoProvision` opt-out.

That gap is not hypothetical — the one scenario that would have caught the project-dependency wiring bug (an app resolving a companion advertised by a project dependency, with `cliAutoProvision` at its default) is never exercised anywhere; the framework build globally disables the feature for its own modules. A functional test with an in-build plugin + consuming app would cover both the bug and the regression.

grails-core/src/test/groovy/org/apache/grails/core/cli/compiler/CommandFactoriesTransformationSpec.groovy
```diff
@@ -0,0 +1,45 @@
...
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+/**
+ * Commands shipped in a companion cli artifact compile from the {@code cli} source set
+ * ({@code src/cli/groovy}), which is not a standard project-source location — the transformation
+ * must still register them in {@code META-INF/grails-cli.factories}.
+ */
+class CommandFactoriesTransformationSpec extends Specification {
```
codeconsole, line 29: This spec only exercises the `isCliSource(URL)` helper. Nothing drives the transformation end-to-end: compiling a direct `ApplicationCommand` implementor, an indirect one via the `GrailsApplicationCommand` trait, and an abstract base class (which must be excluded), then asserting what lands in `META-INF/grails-cli.factories`. `ApplicationContextCommandRegistry` also has no spec pinning the intentional clean-break behavior — that a stale `ApplicationCommand` entry left in a legacy jar's `grails.factories` is silently ignored rather than loaded or erroring. Since the clean break is the headline breaking change, it deserves a test that locks it in.

dependencies.gradle
jdaugherty: @codeconsole in the past, we've merged these type of changes and then prior to shipping the next grails release, we stage and update grails as part of the release.

jdaugherty:
fixes https://github.com/apache/grails-core/issues/15377

sbglasius (approved):

jdaugherty:
@jamesfredley  I'd like to make sure you're ok with the direction this PR takes.  Can you review this before Sunday so we can possibly stage the next milestone?

grails-core/src/test/groovy/org/apache/grails/core/cli/compiler/CommandFactoriesTransformationSpec.groovy
davydotcom: seems like an edge case that isnt really relevant here yea?

grails-bom/base/build.gradle
davydotcom: out of scope comment.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliArtifactGradlePlugin.groovy
davydotcom: minor, can be dealt wiith later.

davydotcom (approved):
The overall intent of this PR is quite needed. This has been a problem for a while. Code looks good and while the file changed count is high. Most of that is due to classpath package changes. There are some adjustments that may be merited in a future PR but I dont see anything majorly wrong here. We have a tendancy here to let AI help us review PRs like this and that is awesome but they also can tend to point out issues out of scope or non critical to a merge. It is up to us to take a human look at it first and assess what those agents say if its needed or not. I approve this PR with hte understanding we may have some iteration on future grails 8 point releases to improve edge cases.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
davydotcom: A caching/scope mechanism would indeed be handy as a future.

codeconsole:
@davydotcom everyone understands AI is not perfect, but if it is going to speed up the time in getting PRs reviewed, I am 100% for it.   We need to move things forward at a MUCH faster rate.   

Grails has notoriously been wayyy behind on the development cycle and in a constant state of catch up.  The fact that it took this simple PR a week to get evaluated showcases this problem. 

I invite anyone willing to spend money on tokens on a top tier model to spend some on any PR.  Those comments were generated with Fable and done a week before anyone else set eyes on it.  It was just meant to get things moving forward. 

You are literally responding today to something generated a week ago.

codeconsole (approved):

davydotcom:
Didn't say it wasn't

codeconsole:
> Didn't say it wasn't

awesome, glad you are onboard. Don't get me wrong, I *definitely* appreciate all your feedback and contributions.

jamesfredley:
I am for this and the PR looks fine.

But I would want to know the clear path for end apps to quickly upgrade to Grails 8, which they will need to do, when we GA 8.0.0, due to Spring Boot EOL.  Will they have to wait on common plugins to re-release and will that bottleneck upgrading to Grails 8.

jdaugherty:
> I am for this and the PR looks fine.
> 
> But I would want to know the clear path for end apps to quickly upgrade to Grails 8, which they will need to do, when we GA 8.0.0, due to Spring Boot EOL. Will they have to wait on common plugins to re-release and will that bottleneck upgrading to Grails 8.

I suspect the breaking part of this PR is only the command part.  So from an upgrade perspective, that is likely the only blocker.  With that said, we've already made changes to 8.x that will likely break plugins (due to spring boot4 changes).  Based on the above, I'm assuming if the core plugin code continues to function, then that's probably good enough to merge.

jdaugherty:
@scott @davydotcom @jamesfredley I know @jamesfredley  brought up purchasing AI credits for key contributors.  We should have a wider discussion about this in the weekly.  If AI helps us get the reviews done easier, I'm for it.  What my concern is is AI generating 100's of PRs that then no one but AI reviews.  That's what often gets categorized as "AI Slop" and it can lead to contributor time being wasted.  Another way to say this: if AI helps accelerate contributions, great.  If AI helps accelerate only commits and then not reviews, it's a major problem for us.

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    private static Properties addToProps(Properties props, String superTypeName, String classNodeNames) {
+        final List<String> classNodesNameList = classNodeNames.tokenize(',')
+        classNodesNameList.forEach(classNodeName -> {
+            String existing = props.getProperty(superTypeName)
+            if (!existing) {
+                props.put(superTypeName, classNodeName)
+            } else if (existing && !existing.contains(classNodeName)) {
```
jamesfredley: **Dedup uses substring matching, which drops distinct registrations.** `existing.contains(classNodeName)` is a substring test, not an exact-membership test. If `com.example.FooCommand` is already registered and `com.example.Foo` is added (or vice-versa - e.g. `MyCommand` vs `MyCommand2`), `contains` returns true and the second, distinct command is **silently not registered**. Any two commands whose FQCNs are prefixes of one another collide.

Fix: split the existing value on `,`, trim into a `Set`, and test exact membership before joining:
```groovy
private static void addToProps(Properties props, String superTypeName, String classNodeNames) {
    def names = new LinkedHashSet<String>()
    def existing = props.getProperty(superTypeName)
    if (existing) names.addAll(existing.tokenize(',')*.trim())
    classNodeNames.tokenize(',')*.trim().each { names.add(it) }
    props.put(superTypeName, names.join(','))
}
```

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
```diff
@@ -0,0 +1,136 @@
...
+            def classNodeName = classNode.name
+            // Use SortedProperties to ensure a consistent order of entries for reproducible builds
+            def props = CollectionFactory.createSortedProperties(false)
+            def superTypeName = superType.getName()
+
+            File factoriesFile = new File(compilationTargetDirectory, factoriesLocation)
+            if (!factoriesFile.parentFile.exists()) {
+                factoriesFile.parentFile.mkdirs()
+            }
+            loadFromFile(props, factoriesFile)
```
jamesfredley, line 70: **Incremental compilation never removes stale entries.** This always loads the previously generated file and merges into it, but nothing ever prunes. When a command class is deleted or renamed, its old entry survives in `grails-cli.factories`, so an incremental build produces a *different* (superset) result than a clean build. At runtime that ghost entry becomes a `ClassNotFoundException` / failed instantiation for a command that no longer exists.

Fix: regenerate the machine-written registrations from the current compilation's class set each run, merging in only the hand-authored `sourceFactoriesLocations` entries - don't treat the prior generated output as an input to merge.

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
```diff
@@ -0,0 +1,136 @@
...
+            if (sourceDirectory != null) {
+                for (String sourceFactoriesLocation : sourceFactoriesLocations) {
+                    File sourceFactoriesFile = new File(sourceDirectory, sourceFactoriesLocation)
+                    loadFromFile(props, sourceFactoriesFile)
+                }
+            }
+
+            addToProps(props, superTypeName, classNodeName)
+
+            factoriesFile.withWriter { Writer writer ->
```
jamesfredley: **Read-modify-write is neither synchronized nor atomic.** Each invocation reads the file, mutates, and rewrites it in place (L65-85). Under parallel compilation (multiple classes / source units, or forked compile processes) targeting the same output directory, two writers can interleave and lose registrations or leave a partially written factories file. Also note the `withWriter` here uses the platform-default charset while reads go through byte-oriented `Properties.load`.

Fix: serialize updates per canonical output file, and write to a temp file followed by an atomic rename (`Files.move(..., ATOMIC_MOVE)`); pin the charset to UTF-8 on both read and write.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
```diff
@@ -0,0 +1,435 @@
...
+     * resolved artifact, or {@code null}. For artifacts produced by a project of the same build
+     * the (possibly not yet built) jar is not read — the coordinate comes from the project's
+     * {@code cliArtifactId} property exported by the cli-artifact convention.
+     */
+    @CompileDynamic
+    protected String findAdvertisedCliArtifact(Project project, ResolvedArtifactResult artifact) {
+        def componentIdentifier = artifact.id.componentIdentifier
+        if (componentIdentifier instanceof ProjectComponentIdentifier) {
+            Project target = project.rootProject.findProject(((ProjectComponentIdentifier) componentIdentifier).projectPath)
+            def cliArtifactId = target?.findProperty('cliArtifactId')
```
jamesfredley: **Project-dependency companion discovery has an evaluation-order race.** For a same-build project dependency this reads the sibling project's `cliArtifactId` extra property - but `GrailsCliArtifactGradlePlugin` only sets that property inside *its own* `afterEvaluate`. A consumer project (e.g. the application, or an aggregating module) that is evaluated before the producing subproject will observe `cliArtifactId == null`, this branch returns `null` without falling back to reading the jar manifest, and the companion (plus its per-command tasks) is **silently omitted**. This bites composite / multi-project builds - including this repo's own modules - even though the published-jar path works.

Fix: defer the project-companion lookup until all projects are evaluated (`gradle.projectsEvaluated`), or expose the companion coordinate as a lazy `Provider` / published variant that can be queried without depending on `afterEvaluate` ordering.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
```diff
@@ -0,0 +1,435 @@
...
+            return
+        }
+
+        // command authoring and execution work out of the box: the command contract + the runner
+        dependencies.add(project.dependencies.create('org.apache.grails:grails-core-cli'))
+        dependencies.add(project.dependencies.create('org.apache.grails:grails-console'))
+
+        Configuration probe = project.configurations.getByName(GRAILS_CLI_DETECT_CONFIGURATION)
+        Set<String> companions = [] as Set
+        def lenientArtifacts = probe.incoming.artifactView { it.lenient(true) }.artifacts
```
jamesfredley: **Eager dependency resolution at configuration time / configuration-cache hostility.** The `grailsCli.withDependencies { autoProvisionCliDependencies(...) }` callback (L152-154) resolves the `grailsCliDetect` probe graph and opens every dependency jar's manifest here, and `configureApplicationCommands` (L235-247) additionally realizes `grailsCliClasspath` artifact files inside `afterEvaluate`. This means **every** build - even `./gradlew help` or a task that never touches the CLI - pays full dependency resolution + jar I/O, defeats task-configuration avoidance, and is fragile under the Gradle 9.6 configuration cache (project/artifact access at the wrong time).

Fix: model command discovery as lazy task inputs (a `Provider<Set<String>>` fed by an `ArtifactView`/`FileCollection`) resolved at execution time, and avoid resolving configurations while registering tasks. Recommend adding a `--configuration-cache` functional test to lock this down.

grails-bom/base/build.gradle
```diff
@@ -97,6 +97,23 @@ dependencies {
     }
 }
 
+// Companion cli artifacts (published by the cli-artifact convention plugin) are additional
+// publications of existing projects, so the subproject enumeration above cannot see them. Each
+// applying project exports its companion coordinate via the `cliArtifactId` extra property, which
+// only exists once that project has been evaluated — so the constraints are computed lazily, in
+// the mutation window Gradle provides right before the configuration is first observed.
+configurations.named('api').configure { apiConfiguration ->
```
jamesfredley: **Companion `-cli` constraints are missing for `enforcedPlatform` consumers of the derived BOMs.** This block adds the companion constraints dynamically only to the `api` configuration of `grails-bom/base`. The derived BOMs (`grails-bom/default`, `grails-bom/hibernate5`, ...) that consumers typically import via `enforcedPlatform(...)` re-declare only the *static* base constraints, so they do not carry these companion constraints in their own direct `constraints` block. Because `autoProvisionCliDependencies` adds `org.apache.grails:grails-core-cli` / `...-cli` with **no version**, an app that pins versions through `enforcedPlatform(grails-bom)` can end up with an unversioned `*-cli` dependency and a resolution failure - a consumer-facing break, contrary to AGENTS.md rule 14.

(Verified separately: the generated `grails-base-bom` POM does contain all 9 companions at `8.0.0-SNAPSHOT`; the gap is specifically the derived BOMs used through `enforcedPlatform`.)

Fix: share/reapply the companion-constraint enumeration in every derived BOM's direct `constraints` block, and add a resolution test using `enforcedPlatform(grails-bom)` that asserts each `*-cli` resolves to a version.

jamesfredley (commented):
### Third-party plugin authors: this is the heavy part of the migration

The PR description says most applications upgrade with **no build changes**, which is accurate for apps. But that promise rests entirely on the *plugin ecosystem* being rebuilt for Grails 8 first, and the migration burden that lands on third-party plugin authors is worth calling out explicitly in the upgrade notes.

**An unchanged Grails 7 command plugin dropped onto a Grails 8 app is not command-compatible** - and it fails silently. Three independent reasons, each sufficient on its own:

1. Its `META-INF/grails.factories` command entries are never read - `ApplicationContextCommandRegistry` now loads **only** `META-INF/grails-cli.factories`.
2. Its command implements the now-absent `grails.dev.commands.*` contract (moved to `org.apache.grails.core.cli.*`, no forwarding alias).
3. Its runtime jar lacks the `Grails-Cli-Artifact` manifest attribute that the app-side `GrailsCliGradlePlugin` scans for during companion auto-discovery.

The plugin's non-command functionality may keep working, but **all of its commands disappear with no error and no warning**. There is no transition path where the old packaging keeps working - it is a hard break, not a deprecation.

The supported migration is not a recompile; it is a repackaging with a second published artifact:

1. Apply `org.apache.grails.gradle.grails-plugin-cli`.
2. Move command sources to `src/cli/groovy`.
3. Rewrite the contract imports (`grails.dev.commands.*` -> `org.apache.grails.core.cli.*`).
4. Move CLI-only dependencies to `cliApi` / `cliImplementation`.
5. Rename hand-authored registrations from `META-INF/grails.factories` to `META-INF/grails-cli.factories` (generated registrations need no file).
6. Publish the companion `-cli` artifact alongside the plugin.

Request: since the silent-failure mode is the most user-hostile part, it would help to (a) state prominently in the upgrade notes that unchanged 7.x command plugins lose their commands **silently**, and (b) consider emitting a build- or runtime-time warning when `META-INF/grails.factories` still contains `ApplicationCommand` registrations, so the failure is diagnosable instead of invisible.

Line-level comments follow for the specific correctness bugs found in the new machinery.

jdaugherty:
I was testing grails-web-console with this code line and seeing what the breakage would be since it doesn't have commands. I can confirm that the plugin should work, but I did find a specific issue on the web console plugin.  We removed the get/getAt methods on Metadata as part of the early 8.0 deprecation work.  This means the following breaks your app now, unless you fix it: 

image

Older grails apps would have generated using this syntax.  To @jamesfredley original point, this change is only breaking for the command portion.  plugins continue to work with this implementation otherwise.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
jdaugherty: Fixed. findAdvertisedCliArtifact now binds a companion advertised by a project of the current build as a project dependency requesting its cli feature capability, instead of a bare group:artifactId module coordinate. The module-coordinate path is kept for external modules and included-build components (distinguished by BuildIdentifier.buildPath). Verified end-to-end: the companion resolves to the sibling project's cli variant with no repository lookup.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
jdaugherty: Agreed the caching/scoping is a good future improvement. For the silent-drop concern I've added a warning log over the discovery probe's lenient resolution failures, so a companion (and its per-command tasks) dropped by an unresolvable dependency is surfaced rather than lost quietly.

grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/commands/CliCompanionPublishingSpec.groovy
jdaugherty: Added CliAutoDiscoverySpec and a cli-companion-autodiscovery test project: an in-build command-bearing plugin plus a consuming app with cliAutoProvision at its default. It asserts the companion is wired as a project dependency on its cli capability (the bug this pins), that grailsCli never leaks onto runtimeClasspath, and that cliAutoProvision = false opts out entirely.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
jdaugherty: I'm not sure how this would work since dependencies can't be added to in an afterEvaluate.  For now, this is going to be a feature and I'll move forward with the warning only.

dependencies.gradle
jdaugherty: I'm going to resolve this - as part of the release vote we'll set it to a non-snapshot, staged version.

grails-bom/base/build.gradle
jdaugherty: Resolving this since it won't apply as part of the release vote - we can't release snapshots in our builds anyhow.

grails-core/src/test/groovy/org/apache/grails/core/cli/compiler/CommandFactoriesTransformationSpec.groovy
jdaugherty: Added FactoriesFileWriterSpec, which drives the registration mechanism end-to-end: a concrete subtype is registered in the factories file, an abstract base class is excluded, a non-subtype is ignored, and two distinct commands whose fully-qualified names share a prefix are both kept. The stale legacy grails.factories clean-break scenario is left as-is per the thread.

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
jdaugherty: Fixed. addToProps now dedups on exact membership (a LinkedHashSet of trimmed names) rather than a substring contains() test, so distinct commands whose fully-qualified names are prefixes of one another (Foo vs FooCommand) are both registered. FactoriesFileWriterSpec includes a regression case for the prefix collision.

grails-bom/base/build.gradle
jdaugherty: Fixed. The companion-constraint enumeration is extracted into gradle/cli-companion-bom-constraints.gradle and applied to the base and every derived BOM, so the -cli constraints are declared directly on each platform rather than only inherited via platform(), and enforcedPlatform() consumers receive forced companion versions. Verified from the generated POMs that grails-bom, both hibernate BOMs, and all three micronaut BOMs now carry all nine companions.

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
jdaugherty: Gradle already handles "exclusive" locks because the project doesn't parallel compile these files - they're part of a single project and the build system is what handles these locks. We would not want to add exclusivitiy here - it's the role of the build system instead.

grails-core/src/main/groovy/org/grails/compiler/injection/FactoriesFileWriter.groovy
jdaugherty: This code was modeled off the existing factories writer.  Yes, this is an issue but it's been a long standing issue that would require tracking the current class set.  Incremental compile is considered an incubating feature of Gradle and not a default feature.  This is something we should track separately from this change. I've opened https://github.com/apache/grails-core/issues/16007 for this.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
jdaugherty: The concern is real today — Grails doesn't yet support the configuration cache (our own build runs with `org.gradle.configuration-cache=false`), so this cost is currently paid on every invocation. But it's worth separating what Gradle makes fixable here from what it doesn't.

**The hard constraint:** the per-command task names (`dbmUpdate`, ...) come from `META-INF/grails-cli.factories` inside resolved dependency jars, and Gradle requires task names to exist when the task graph is assembled. There is no API for lazily-materialized task *names* — a `Provider<Set<String>>` fed by an `ArtifactView` works for a task's inputs, but not for the existence of the tasks themselves. So "dynamically named tasks, visible in `gradle tasks`/IDE sync, with zero configuration-time resolution" is not a combination Gradle offers; any plugin registering tasks from dependency content has this shape.

**What is achievable within the current design:**

1. *Task rules* (`tasks.addRule`) would defer discovery until an unmatched task name is actually requested — `./gradlew help` and unrelated builds would pay nothing. The tradeoff is real, though: the commands would no longer be listed individually in `gradle tasks` or IDE task views, which is a big part of why per-command tasks exist. Could be offered behind a flag for build-speed-sensitive projects.
2. *Configuration cache is the caching mechanism* — configuration-time resolution is supported under CC; the resolved result is serialized into the cache entry, so discovery would be paid once and replayed on every subsequent build. The genuine CC blockers in this code are incidental rather than structural (cross-project `findProject` access during discovery, `Project` captures in task closures) and are tractable cleanup. So the right long-term answer to this comment largely coincides with Grails gaining CC support.
3. *Cheap scoping wins* meanwhile: cache the manifest/factories scan per resolved artifact set so jars aren't re-opened within an invocation, and the generic `runCommand` task already executes any command without discovery.

**The deeper issue** is that all of this complexity exists because we decided to surface commands as Gradle *tasks*. A better solution may be to keep tracking the dependencies in `grailsCli` but resolve and use them in the Grails CLI itself — fully separating the build system from the CLI system. That's a redesign (and a partial revert to the older design), and we haven't agreed on how to restructure profile or forge, so it's out of scope for this PR. The significance of this problem is why I've suggested we use the hackathon to address the profile & forge design — moving forward with a single CLI design is critical long term, and this thread is a good illustration of why.

Proposed follow-up for the tactical items: keep eager registration as the default (discoverability is the feature), do the CC-compatibility cleanup with a `--configuration-cache` functional test to lock it in, and consider the task-rule mode as an opt-in. Agreed this shouldn't block the PR — tracked in #16008.

jdaugherty: I'm going to resolve this since this issue already exists with the existing implementation - this isn't new to this change. The ticket was created to address that issue.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliArtifactGradlePlugin.groovy
jdaugherty: Fixed. When `cliArtifact { artifactId }` is customized, the cli variants now additionally carry the advertised capability (`$group:$artifactId`) alongside the Gradle default, so an in-build requireCapability against the advertised coordinate resolves. The published-metadata rewrite also recognizes customized companion coordinates now (previously only the `<module>-cli` naming convention), so the capability request is still replaced with the plain companion coordinate in the GMM. Covered end-to-end in CliAutoDiscoverySpec (in-build discovery + resolution) and CliCompanionPublishingSpec (publish + external resolve) with a customized acme-tools companion.

grails-core/src/test/groovy/org/apache/grails/core/cli/compiler/CommandFactoriesTransformationSpec.groovy
jdaugherty: Also added the indirect-implementor case: a command implementing the contract through a trait (the GrailsApplicationCommand pattern) is registered correctly.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/commands/GrailsCliGradlePlugin.groovy
jdaugherty: Largely addressed across the last three commits: discovery now reads the `cliArtifact` extension directly (it exists with its convention from plugin-apply time, so the afterEvaluate ordering dependency is gone), matches the producing project by build tree path, and falls back to the jar-manifest read otherwise. The residual window is a producer that is both unevaluated and unbuilt when `grailsCli` is first observed — that's the fully-lazy discovery redesign, tracked with the related configuration-time work in #16008.

jdaugherty:
@jamesfredley following up on the upgrade-path question: the docs now spell out the bounded impact (upgrade notes §35 "What this change means for upgrading an application" + What's New). Verified empirically as well — a Grails 7 plugin (grails-web-console 7.1.0) runs on a Grails 8 app built from this branch with its runtime functionality intact; the split itself required no plugin changes. Plugin re-releases are only needed for commands (companion `-cli` artifact) or where a plugin relied on APIs deprecated in 7 and removed in 8.

jdaugherty:
@jamesfredley Given that:
1. your time is limited right now to respond & this needs to make it into tomorrow's milestone stage
2. your comment wasn't a blocking change (requested changes)
3. all of the comments have been addressed except those that exist with the current code line (gradle task names / incremental compile)
4. this change doesn't require rebuilding every grails plugin (only if you want the commands, which would indicate active development)
5. there are multiple other approvals
6. we can always iterate on this design before a RC or the final release 

I'm going to merge this after the tests finish.  If you feel this needs reverted, let's revisit.

PR #15949, opened 2026-07-09
Restore lenient config handling

### LINKED ISSUE #15818: relaxed/lenient config property resolution not working
### Expected Behavior

this config should "just work":
```yml
management:
  endpoint:
    heapdump:
      access: unrestricted
```
(ref. https://docs.spring.io/spring-boot/3.5/reference/actuator/endpoints.html#actuator.endpoints.controlling-access)

but it fails with the error below.

that property is not subject to config-property binding (where relaxed binding works) but is resolved via `environment.getProperty(String, Enum)`:
* https://github.com/spring-projects/spring-boot/blob/v3.5.15/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/PropertiesEndpointAccessResolver.java#L89

looks like `GrailsApp` is overriding `SpringApplication.configureEnvironment` but does not add `ApplicationConversionService`. ref.:
* https://github.com/apache/grails-core/blob/v7.1.2/grails-core/src/main/groovy/grails/boot/GrailsApp.groovy#L131-L143
* https://github.com/spring-projects/spring-boot/blob/v3.5.15/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L481-L498

thanks & regards.

### Actual Behaviour

```
2026-07-03T07:58:28.364+02:00 ERROR 82081 --- [  restartedMain] o.s.b.d.LoggingFailureAnalysisReporter   : 

***************************
APPLICATION FAILED TO START
***************************

Description:

Invalid value 'unrestricted' for configuration property 'management.endpoint.heapdump.access' (originating from '"management.endpoint.heapdump.access" from property source "Config resource 'class path resource [application.yml]' via location 'optional:classpath:/'"'). Validation failed for the following reason:

Failed to convert to type org.springframework.boot.actuate.endpoint.Access
```

### Steps To Reproduce

* run sample app
* see error

### Environment Information

- ubuntu-24.04
- jdk 17

### Example Application

https://github.com/zyro23/grails-core-15818

### Version

7.1.2

jdaugherty:
## Description

  Fixes #15818

  Grails applications reject lenient configuration values (e.g. lowercase enum names) that work in plain Spring Boot applications. For example, this standard actuator configuration causes the application to fail on startup:

  ```yml
  management:
    endpoint:
      heapdump:
        access: unrestricted
   ``` 
Invalid value 'unrestricted' for configuration property 'management.endpoint.heapdump.access' ...
  Failed to convert to type org.springframework.boot.actuate.endpoint.Access

  Root cause: GrailsApp has overridden SpringApplication.configureEnvironment() since Grails 3 without delegating to super. Spring Boot's implementation installs the ApplicationConversionService on the environment, which provides the lenient converters (lowercase/hyphenated enum names, durations, etc.). Because Grails skipped that step, the environment fell back to a plain DefaultConversionService, whose string-to-enum conversion is case-sensitive. This only affects properties resolved directly through environment.getProperty(name, type) — such as the actuator's PropertiesEndpointAccessResolver — since @ConfigurationProperties binding supplies its own conversion service.affects properties resolved directly through environment.getProperty(name, type) — such as the actuator's PropertiesEndpointAccessResolver — since @ConfigurationProperties binding supplies its own conversion service.

Change: GrailsApp.configureEnvironment() now delegates to super.configureEnvironment(environment, args) in place of the direct configurePropertySources() call. This installs the conversion service while preserving identical behavior otherwise (configureProfiles() is an empty method in Spring Boot 3.x, and the Grails-specific profile handling is unchanged).

Tests:
- GrailsAppEnvironmentConversionSpec (grails-core) boots a minimal GrailsApp and verifies lowercase (unrestricted) and hyphenated (read-only) enum values resolve via environment.getProperty(). This spec fails without the fix.
- The grails-test-examples/external-configuration functional app now declares the exact configuration from the issue (management.endpoint.heapdump.access: unrestricted, threaddump.access: read-only), which previously prevented startup, and the new RelaxedPropertyResolutionSpec verifies both values resolve to the Access enum through the Spring environment.

Contributor Checklist

Issue and Scope

- [x] This PR is linked to an existing issue that has been acknowledged or approved by the project team.
- [x] This PR targets the correct branch for the type of change (bug fix targeting 7.0.x).

Code Quality

- [x] I have added or updated tests that cover the changes introduced in this PR.
- [ ] I have verified that all existing tests pass by running ./gradlew build --rerun-tasks.
- [x] My code follows the project's code style guidelines.
- [x] This PR does not include mass reformatting, style-only changes, or large-scale refactoring.
- [x] If generative AI tooling was used in preparing this contribution, a quality model was used to ensure contributions are consistent with the project's quality standards.

Licensing and Attribution

- [x] All contributed code is provided under the Apache License 2.0, and new source files include the appropriate Apache license header.
- [x] If generative AI tooling was used in preparing this contribution, I have followed the ASF's policy on generative tooling and have properly attributed its use.

Documentation

- [x] If this PR introduces user-facing changes, I have included or updated the relevant documentation. (No doc change needed — this restores documented Spring Boot behavior; no new or changed public API.)
- [x] If this PR adds a new feature, I have updated the What's New section. (N/A — bug fix.)
- [x] If this PR introduces breaking changes... (N/A — no user action required.)
- [x] The PR description clearly explains what was changed and why.


grails-test-examples/external-configuration/grails-app/conf/application.yml
```diff
@@ -99,6 +99,15 @@ environments:
...
         defaultTransactionIsolation: 2 # TRANSACTION_READ_COMMITTED
 ---
+# Lenient enum values resolved via environment.getProperty(name, Access) rather than
+# relaxed configuration-property binding - regression test for issue #15818
+management:
+  endpoint:
+    heapdump:
+      access: unrestricted
+    threaddump:
+      access: read-only
```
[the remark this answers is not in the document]
jdaugherty: Let's leave this.  They're only designed to run locally.

grails-core/src/main/groovy/grails/boot/GrailsApp.groovy
```diff
@@ -133,7 +133,9 @@ class GrailsApp extends SpringApplication {
 
     @Override
     protected void configureEnvironment(ConfigurableEnvironment environment, String[] args) {
-        configurePropertySources(environment, args)
+        // Delegating to super installs the ApplicationConversionService so relaxed property
+        // resolution (e.g. lowercase enum values) works via environment.getProperty()
+        super.configureEnvironment(environment, args)
```
codeconsole, line 170: Two side effects of this delegation worth noting (neither blocking):

1. `GrailsApplicationPostProcessor.loadApplicationConfig()` mutates `environment.getConversionService()` via `addConverter(...)` and then propagates it into `PropertySourcesConfig`. That still works only because Boot 3.5 installs a **new mutable** `ApplicationConversionService` here rather than the unmodifiable `ApplicationConversionService.getSharedInstance()`. If a future Boot version switches to the shared instance, those `addConverter` calls will throw at startup — worth remembering during Boot upgrades.

2. Because that same conversion service flows into `PropertySourcesConfig`, typed access through `grailsApplication.config.getProperty(name, type)` also becomes lenient (durations, lowercase/hyphenated enums, delimited collections, formatters), not just `environment.getProperty()`. It's strictly in the direction of Spring Boot alignment, so I consider it a plus — but the behavior change is a bit broader than the PR description states. Maybe add a sentence to the description so it's discoverable later.

grails-core/src/test/groovy/grails/boot/GrailsAppEnvironmentConversionSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        System.setProperty('test.access.lowercase', 'unrestricted')
+        System.setProperty('test.access.hyphenated', 'read-only')
+        GrailsApp app = new GrailsApp(EnvironmentConversionTestConfiguration)
+        app.webApplicationType = WebApplicationType.NONE
+
+        when:
+        ConfigurableApplicationContext context = app.run()
+
+        then:
+        context.environment.conversionService instanceof ApplicationConversionService
```
codeconsole, line 50: Nit: this asserts a Boot implementation detail on top of the behavioral assertions below it, but since the installed conversion service *is* the contract being restored (and it would still pass if Boot ever reverted to `getSharedInstance()`, which is also an `ApplicationConversionService`), it works well as a diagnostic — when this line fails you know the service was never installed, vs. a converter behavior change. Fine as is.

grails-test-examples/external-configuration/src/integration-test/groovy/test/app/RelaxedPropertyResolutionSpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+ */
+@Integration
+class RelaxedPropertyResolutionSpec extends Specification {
+
+    @Autowired
+    ConfigurableEnvironment springEnvironment
+
+    void 'the environment uses the ApplicationConversionService'() {
+        expect: 'the conversion service installed by Spring Boot is present'
+        springEnvironment.conversionService instanceof ApplicationConversionService
```
codeconsole, line 44: Observation: the two enum-resolution features below already prove everything this feature asserts, so this one is technically redundant — but it's useful as a first-failure diagnostic that distinguishes "conversion service missing" from "converter behavior changed", and it documents intent. No change needed.

codeconsole (approved):
LGTM. Verified the core claim against the Spring Boot 3.5.16 bytecode the build resolves: `SpringApplication.configureEnvironment()` does exactly three things — installs a new `ApplicationConversionService` (guarded by `addConversionService`, default `true`), calls `configurePropertySources()` (the same call the old code made directly, so no double invocation and no ordering change for the profile logic that follows), and calls `configureProfiles()`, which has an empty body in Boot 3.5. Delegating to super is a clean restoration of the Boot contract.

Also confirmed no other entry point needs the same fix: `GrailsSwingConsole`, `GrailsShell`, and `DevelopmentGrailsApplication` all extend `GrailsApp` without overriding `configureEnvironment`, so they inherit it, and no other override exists in the repo.

Tests are well-placed: the unit spec follows the existing `GrailsAppPidFileSpec` pattern and fails without the fix, and since the pre-fix failure mode was startup failure, booting the `external-configuration` app with the issue's exact yml is itself the regression test. Nice that `setAddConversionService(false)` remains as an escape hatch if the new leniency ever causes trouble for an app.

A couple of non-blocking observations inline.

grails-core/src/main/groovy/grails/boot/GrailsApp.groovy
jdaugherty: I've updated the comment & PR title

PR #15950, opened 2026-07-09
Clear omitted bindData fields when clearMissing is enabled

jamesfredley:
## Description

Adds opt-in stale-data clearing to `bindData`, stacked on the binding base from #15947.

When `nullMissing: true` is supplied together with an explicit `include` list, included properties omitted from the source are cleared (reference types to `null`, primitives to their type default). Existing behavior is unchanged when the option is absent, false, or no explicit include was supplied.

## Final contract

| Area | Behavior |
| --- | --- |
| Opt-in | `bindData(target, source, [include: [...], nullMissing: true])` |
| Default | Omitted fields retain their existing value |
| Authorization | Explicit includes and (in secure mode) generated binding allowlists authorize clearing |
| Denials | Excludes, `bindable: false`, and framework-managed properties remain protected |
| Primitives | Omitted included primitives reset to the type default (e.g. `0` / `false`) |
| Failures | Clear failures are reported on the returned `BindingResult` |
| Paths | Nested objects, lists, arrays, maps, indexed paths, and filters are supported |
| Wildcards | Wildcard matching is isolated to `nullMissing`; it cannot broaden normal request binding |
| Compatibility | Existing public overloads continue to delegate with `nullMissing = false` |
| Modes | Works under the shipping permissive default and under opt-in secure mode |

## Implementation notes

- Clearing logic lives in package-private `NullMissingPropertyClearer` (not expanded public API).
- Framework property names are centralized in `FrameworkPropertyNames` (intrinsic runtime vs Grails-managed).
- Stacked on #15947: permissive default + opt-in deny-by-default secure mode; `bindable: false` always honored.

## Example

```groovy
bindData book, params, [include: ['title', 'description'], nullMissing: true]
```

## Verification

- `:grails-web-databinding:test`
- `:grails-databinding-core:test`
- Focused `BindDataMethodTests` (default-mode and secure-mode nullMissing paths, primitives, BindingResult errors, nested/indexed/maps, excludes, `bindable: false`)
- `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec` (explicit bindable special properties)

## Related

| PR | Relationship |
| --- | --- |
| #15947 | Binding base (permissive default + opt-in secure mode) merged into this PR |
| #15808 | Earlier combined approach, closed as superseded by #15947 and #15950 |

## Contributor Checklist

- [x] Tests cover the affected public behavior in the shipping default mode
- [x] User-facing behavior is documented in the `bindData` reference and Grails 8 upgrade guide
- [x] Default behavior and existing overloads remain compatible
- [x] The PR is labeled `ai-generated-starting-point`
- [x] All contributed code is provided under the Apache License 2.0


grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -300,6 +320,586 @@ public static BindingResult bindObjectToDomainInstance(PersistentEntity entity,
...
+        catch (Exception e) {
+            return null;
+        }
+    }
+
+    private static Object getIndexedValue(Object indexedProperty, String index) {
+        if (indexedProperty instanceof List) {
+            List list = (List) indexedProperty;
+            Integer parsedIndex = parseIndex(index);
+            return parsedIndex != null && parsedIndex >= 0 && parsedIndex < list.size() ? list.get(parsedIndex) : null;
```
jdaugherty: This PR appears to be doing more than what it says it does.  It appears to have overlap with the other changes.

jamesfredley: Resolved in 018220fb3e by explicitly incorporating #15947 as this PR's deny-by-default security base. The remaining stacked delta is focused on nullMissing: six files covering the opt-in API, path traversal, public regressions, and docs. Security logic is reused rather than duplicated, and the final PR description now makes the relationship and scope explicit.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -300,6 +320,586 @@ public static BindingResult bindObjectToDomainInstance(PersistentEntity entity,
...
+        for (Object includedProperty : bindingIncludeList) {
+            if (includedProperty instanceof CharSequence) {
+                String includedPropertyName = includedProperty.toString();
+                if (propertyName.equals(includedPropertyName) || includedPropertyName.startsWith(propertyName + ".")) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
jamesfredley: Addressed in 018220fb3e. The normal binding predicate retains #15947 exact/nested-prefix allowlist semantics, while wildcard path matching now lives only in the nullMissing helper. Regression coverage proves generated foo_* / foo.* entries cannot authorize the sibling foo_admin, while nested and indexed nullMissing paths still work. The four affected module suites pass.

jamesfredley:
Final update pushed in `018220fb3e`.

The branch now incorporates #15947 as its deny-by-default security base and keeps `nullMissing` as a focused opt-in behavior. Review fixes cover wildcard sibling authorization, indexed excludes and includes, explicit include semantics, null generated allowlists, public boolean overload safety, recursive `bindable: false`, and framework-managed properties.

Verification passed for `:grails-databinding-core:test`, `:grails-databinding:test`, `:grails-web-databinding:test`, and `:grails-test-suite-web:test`. The final delta completed three rounds of Oracle and Codex review; all findings within #15950's six-file stacked scope are resolved.

borinquenkid:
Thanks for splitting this out — since #15950 is branched directly off #15947, its diff currently duplicates all of #15947's changes on top of the `nullMissing` additions, which means reviewers end up reviewing the mass-assignment hardening twice.

Would you be open to one of:
1. Retargeting this PR's base branch to `fix/binddata-mass-assignment` (the #15947 branch), so this shows only the incremental `nullMissing` commits — GitHub will auto-retarget it to `8.0.x` once #15947 merges, or
2. Folding this PR's commits into #15947 so it's reviewed as a single PR.

Whichever is easier on your end — just flagging so we don't duplicate review effort.

jamesfredley:
PR updated to target fix/binddata-mass-assignment

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -576,6 +605,619 @@ else if (include.isEmpty()) {
         return bindingResult;
     }
 
+    private static void assignNullToMissingIncludedProperties(Object object, DataBindingSource bindingSource, List include, List exclude, String filter) {
```
jdaugherty: This is where the cost of the design shows up. `assignNullToMissingIncludedProperties` re-walks the binding source after `grailsWebDataBinder.bind(...)` has returned, and to do that it re-implements path splitting (`splitPropertyPath`, `propertyPathSeparator`), bracket and index parsing, checkbox marker names, collection and map expansion, and prefix-filter handling — a second implementation of the grammar the binder just traversed with `IndexedPropertyReferenceDescriptor` and `processIndexedProperty`.

Two implementations of the same path grammar will drift, and this is the copy that decides which properties get written to. `shouldExpandMapEntries` / `isStructuredMapValueType` inferring "is this a nested object map?" from the package name of the value type is a symptom of working without the type information the binder had already resolved.

Could the clearing be driven from inside the binder instead — it already holds the resolved allowlist and knows which properties it set, so the remainder is the set to null? If that is not workable, please extract this into its own collaborator with its own unit tests rather than growing `DataBindingUtils`, which is already carrying the include-list resolution and the mode switch.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -576,6 +605,619 @@ else if (include.isEmpty()) {
...
+            Integer parsedIndex = parseIndex(index);
+            return parsedIndex != null && parsedIndex >= 0 && parsedIndex < list.size() ? list.get(parsedIndex) : null;
+        }
+        if (indexedProperty instanceof Map) {
+            return ((Map) indexedProperty).get(index);
+        }
+        return null;
+    }
+
+    private static void setPropertyValueToNull(Object object, String propertyName) {
```
jdaugherty: Every failure here is swallowed. A `nullMissing` clear that cannot be applied — a primitive-typed property, a property with no setter, a setter that throws — silently leaves the stale value in place, which is exactly the outcome the option exists to prevent, and nothing is recorded in the `BindingResult` that `bindData` returns.

The catch comment says "ignore invalid indexed nullMissing paths", but the `try` also covers the non-indexed `mc.setProperty(...)` branch, so it is broader than the comment claims.

Please narrow the catch to the path/index parsing, decide explicitly what a primitive-typed include should do (reject it at the API, or set the type default), and surface a failure to clear through the errors object the way a binding failure is surfaced. A test with a primitive property in the include list would pin whichever behavior you choose.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -77,6 +82,9 @@ public class DataBindingUtils {
     private static final String BLANK = "";
     private static final Map<Class, List> CLASS_TO_BINDING_INCLUDE_LIST = new ConcurrentHashMap<>();
     private static final Map<Class, List> CLASS_TO_LEGACY_BINDING_INCLUDE_LIST = new ConcurrentHashMap<>();
+    private static final Set<String> FRAMEWORK_MANAGED_PROPERTIES = Set.of(
```
jdaugherty: This is now the third list of "properties the binder must never touch" in the stack, and the three disagree:

- `SimpleDataBinder.isFrameworkProperty`: `class, classLoader, protectionDomain, metaClass, metaPropertyValues, properties`
- `GrailsWebDataBinder.FRAMEWORK_MANAGED_PROPERTIES`: `class, errors, id, version, dateCreated, lastUpdated`
- this one: the union of both

A property added to one and not the others becomes clearable but not bindable, or the reverse. Please consolidate to a single constant that all three consume.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -479,6 +487,11 @@ public static <T> void bindToCollection(final Class<T> targetType, final Collect
      * @return A BindingResult if there were errors or null if it was successful
      */
     public static BindingResult bindObjectToInstance(Object object, Object source, List include, List exclude, String filter) {
+        return bindObjectToInstance(object, source, include, exclude, filter, false);
+    }
+
+    public static BindingResult bindObjectToInstance(Object object, Object source, List include, List exclude, String filter, boolean nullMissing) {
```
jdaugherty: This overload and the matching `bindObjectToDomainInstance(..., boolean nullMissing)` are new public API on a class where the neighbouring overloads are documented.

Please add javadoc covering the parameter, the fact that it is ignored unless the caller supplied a non-null `include`, and that the clearing happens after binding completes — outside the listener callbacks and outside the `BindingResult` the method returns.

grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy
```diff
@@ -271,7 +271,23 @@ class SimpleDataBinder implements DataBinder {
     }
 
     protected boolean isOkToBind(String propName, List whiteList, List blackList) {
-        'class' != propName && 'classLoader' != propName && 'protectionDomain' != propName && 'metaClass' != propName && 'metaPropertyValues' != propName && 'properties' != propName && !blackList?.contains(propName) && (whiteList == null || isBindAllBindingIncludeList(whiteList) || whiteList.contains(propName) || whiteList.find { it -> it?.toString()?.startsWith(propName + '.') })
+        !isFrameworkProperty(propName) && !blackList?.contains(propName) &&
+                (whiteList == null || isBindAllBindingIncludeList(whiteList) || whiteList.contains(propName) ||
+                        whiteList.any { item -> item?.toString()?.startsWith(propName + '.') })
+    }
+
+    static boolean isPropertyExcluded(String propertyName, List excludeList) {
```
jdaugherty: This adds a public static method to `grails.databinding.SimpleDataBinder` — public API in `grails-databinding-core` — solely so `DataBindingUtils` in `grails-web-databinding` can call it. It also implements exclusion semantics (`.*` and `_*` prefix matching, and nested-path prefixes) that the binder's own `isOkToBind` does not apply, so `SimpleDataBinder` now exposes two different notions of "excluded".

It has no javadoc and no coverage in `SimpleDataBinderSpec`. If it is only needed by the web binding layer, please make it package-private there or move it to a shared internal utility; if it is genuinely meant to be public API, document it and test it at the `SimpleDataBinder` level.

grails-test-suite-web/src/test/groovy/org/grails/web/servlet/BindDataMethodTests.groovy
```diff
@@ -133,6 +133,158 @@ class BindDataMethodTests extends Specification implements ControllerUnitTest<Bi
         target.email == null
     }
 
+    void 'Test bindData With Null Missing Clears Omitted Included Field'() {
```
jdaugherty: This spec and every other `nullMissing` case in the file inherit `legacyBindableDefault=false` from the `setup()` at the top, so the whole feature is only exercised in the opt-in secure mode.

The reference docs and the 8.0.x upgrade note both present `nullMissing` as a plain `bindData` option with no mention of the binding mode, and the authorization path genuinely differs between the two: `isNullMissingPropertyBindable` consults `getBindingIncludeList(object)`, which resolves the legacy allowlist under the default and the generated one under secure mode.

Please add default-mode coverage for the main paths — clearing an omitted included field, leaving an excluded one, honouring `bindable: false`, and the nested/indexed cases — so the documented behavior is verified in the mode most applications will be running.

jdaugherty (commented):
The `nullMissing` contract is well scoped — opt-in, requires an explicit `include`, defaults unchanged — and the reference docs read clearly.

Two structural concerns before this lands.

First, the clearing pass re-derives "which included properties were absent from the source" by re-parsing the binding source after binding has already finished. That is why `DataBindingUtils` grows roughly 640 lines of bespoke path handling, duplicating traversal the binder just performed and already has type information for.

Second, every `nullMissing` spec inherits `legacyBindableDefault=false` from the `BindDataMethodTests` `setup()`, so the feature has no coverage under the shipping default — even though the docs present it as a plain `bindData` option with no mention of the binding mode.

Smaller items inline.

jamesfredley:
@jdaugherty Addressed after merging the updated #15947 base (`50714be10c`).

## Review responses

| Item | Resolution |
| --- | --- |
| Post-bind path re-walk (~640 lines in `DataBindingUtils`) | Extracted to package-private `NullMissingPropertyClearer` with coverage through public `bindData` / `DataBindingUtils` APIs. Clearing remains intentionally post-bind (documented). |
| nullMissing only tested under secure mode | After the #15947 merge, `BindDataMethodTests` runs on the unconfigured permissive default. Added/kept default-mode coverage for clear omitted include, leave excluded, `bindable: false`, nested/indexed/map paths. Secure-mode cases stay explicit via `enableSecureBinding()`. |
| Swallowed clear failures | Primitive includes reset to the type default. Clear failures produce `FieldError`s on the returned `BindingResult` (full nested path preserved); existing conversion errors are kept. |
| Three disagreeing framework-property lists | Single `FrameworkPropertyNames` with intrinsic runtime vs Grails-managed sets. `SimpleDataBinder` only hard-denies intrinsic runtime properties so `bindable: true` on `id`/`dateCreated`/`lastUpdated` still works; nullMissing uses the full managed set. |
| Missing javadoc on public `nullMissing` overloads | Documented: ignored without a non-null `include`; clearing runs after bind; listener/`BindingResult` behavior described. |
| Public `SimpleDataBinder.isPropertyExcluded` | Removed from public API; exclusion matching lives in the web-layer clearer. |

## Verification

`:grails-databinding-core:test`, `:grails-web-databinding:test`, and focused `:grails-test-suite-web:test` (`BindDataMethodTests`, `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec`) all green.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
jamesfredley: Addressed in 69ed46edca. The second-pass path grammar and clearing logic are extracted from `DataBindingUtils` into the focused `NullMissingPropertyClearer` collaborator. Public `bindData` regressions cover nested and indexed collection paths, map paths, prefixes, exclusions, blank values, framework properties, bindability, primitives, and failure paths, without testing package-private implementation details directly.

matrei:
@jamesfredley Crap, those commits were supposed to be pushed to a new branch as a suggestion, but I messed it up.
I'll leave them for now and revert them if you don't agree with them.

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec.groovy
```diff
@@ -26,6 +26,7 @@ import groovy.transform.CompileStatic
 import org.grails.config.PropertySourcesConfig
 import org.grails.validation.ConstraintEvalUtils
 import spock.lang.Issue
+import spock.lang.Shared
```
sbglasius: Stray import?

sbglasius (approved):

jamesfredley:
@matrei not a problem.  Ready to merge?

matrei:
> @matrei not a problem.  Ready to merge?

@jamesfredley In that case, yes!

PR #15952, opened 2026-07-10
Make AutoTimestampEventListener temporary disabling thread-safe

### LINKED ISSUE #14510: GRAILS 6+ BUG: autoTimestampEventListener.without* Results in Unexpected Behavior Across All Threads
https://github.com/codeconsole/grails-autotimestamp-bug

https://gorm.grails.org/latest/hibernate/manual/index.html#eventsAutoTimestamping

Since Apr 18, 2017 https://github.com/grails/grails-data-mapping/commit/353a25427c55cb1dbe52a29b40c2308ef394fda0

codeconsole:
Fixes #14510

### The bug

`withoutTimestamps`, `withoutDateCreated` and `withoutLastUpdated` disabled timestamping by mutating the listener-wide entity maps (capturing each entry and setting it to `Optional.empty()`, then restoring). That state is JVM-global, so overlapping windows on concurrent threads corrupt each other:

1. Thread A enters: captures ON, sets OFF
2. Thread B enters: captures OFF, sets OFF
3. Thread A exits: restores ON
4. Thread B's save now runs **with** timestamping — its historical `dateCreated`/`lastUpdated` values are silently clobbered

Reproduction: https://github.com/codeconsole/grails-autotimestamp-bug

A second facet: there was no `try`/`finally`, so an exception inside the closure left timestamping **permanently disabled** for the whole JVM.

### The fix

The shared metadata maps are no longer mutated. Suppression is tracked per thread: the property-name getters consult a `ThreadLocal` holding an all-entities nesting depth plus the set of entity names disabled by per-class scopes. Each scope only re-enables the names it added, so nested and overlapping scopes restore correctly; `try`/`finally` guarantees restoration when the closure throws; the `ThreadLocal` is removed when empty so nothing is retained on pooled threads.

Public API, protected fields and single-thread semantics are unchanged (binary and source compatible).

### Behavioral note

Disabling is now scoped to the calling thread. Threads spawned *inside* the closure are no longer affected — under the old global behavior they were (by accident). Code relying on that must call `withoutTimestamps` on each thread. The auto-timestamping guide now documents the thread scoping. Worth a mention in the release notes.

### Tests

The spec was rewritten to exercise the listener through the public `beforeInsert`/`beforeUpdate` API and extended with tests that reproduce the bug — written first and verified to fail on the old implementation:

* cross-thread isolation (the main #14510 repro)
* overlapping windows on different threads restoring independently (the interleaving above)
* nested and overlapping same-class scopes restoring the enclosing scope
* restoration when the closure throws, including a partial failure while registering a class list
* an 8-thread contention stress test

All 218 tests in `grails-datamapping-core` pass, as do the timestamp specs in `grails-datamapping-core-test`, `grails-data-hibernate5-core` and `grails-test-suite-uber`; Checkstyle and CodeNarc are clean.

### DIFF
--- a/grails-data-hibernate5/docs/src/docs/asciidoc/advancedGORMFeatures/eventsAutoTimestamping.adoc
+++ b/grails-data-hibernate5/docs/src/docs/asciidoc/advancedGORMFeatures/eventsAutoTimestamping.adoc
@@ -408,3 +408,5 @@ autoTimestampEventListener.withoutTimestamps {
 ----
 
 WARNING: Because the timestamp handling is only disabled for the duration of the closure, you must flush the session during the closure execution!
+
+NOTE: The timestamp handling is only disabled for the thread executing the closure. Other threads persisting entities at the same time are unaffected and will continue to have their timestamps applied.
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/events/AutoTimestampEventListener.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/events/AutoTimestampEventListener.java
@@ -22,7 +22,6 @@ import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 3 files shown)


jdaugherty:
This has been like this for years and at this point it does revert a behavior so I think this shouldn't go in 7.0

PR #15956, opened 2026-07-10
Seed URL mappings index with runtime fallback

jamesfredley:
### What was found

| Problem | Impact |
| --- | --- |
| URL mappings are resolved at runtime with linear/holder matching | Startup and request routing cost stays higher than needed |
| No build-time index SPI for mappings | Future precomputation work has no safe landing path |
| Google Doc 4.2 listed this as 8.1+ architecture | Needs a starter that preserves runtime fallback |

### What changed

| Area | Change |
| --- | --- |
| Index properties | Add `UrlMappingsIndexProperties` for optional `META-INF/grails/url-mappings-index.properties` |
| Runtime holder | `DefaultUrlMappingsHolder` only *detects* the descriptor (debug log); runtime mapping evaluation stays authoritative whether or not a descriptor is present |
| Fallback | Missing/malformed/unreadable descriptor is a non-fatal soft miss; existing runtime mapping behavior is unchanged |
| Docs / tests | Upgrade note + missing/malformed/unreadable/precedence/valid coverage |

### Review feedback addressed

| Source | Item | Resolution |
| --- | --- | --- |
| Copilot | `load()` skipped the TCCL and did not wrap `IllegalArgumentException` like `IOException` | Try the thread context classloader first, then the provided one; catch `IOException`/`RuntimeException` **per loader** so a throwing TCCL still falls through to the fallback loader and an optional/malformed descriptor never aborts startup (`Error` still propagates) |
| Review pass | New `.adoc` missing ASF header; `@since`/docs said 8.1; wording implied conditional fallback | Added ASF header, corrected to 8.0, and clarified that runtime evaluation stays authoritative regardless of descriptor presence |

### Out of scope / follow-up

| Topic | Status |
| --- | --- |
| Full trie/routing precomputation | 8.1+ |
| Reverse-routing table generation | Follow-up |
| AST transform writing the index at compile time | Follow-up |

### Related MD topics

| Source | Topic |
| --- | --- |
| Google Doc 4.2 | URL mapping pre-computation |
| Status map | Not covered / defer -> starter now |

### Issue and Scope
- [x] Starter PR for future architecture; no acknowledged issue required for discussion.
- [x] Runtime behavior preserved without an index.
- [x] Single focused seed change.
- [x] Targets `8.0.x` as an experimental starter for 8.1 work.

### Code Quality
- [x] Tests cover the fallback path.
- [x] Focused module tests run locally.
- [x] No mass reformatting.
- [x] Generative AI starting point; labeled.

### Licensing and Attribution
- [x] Apache License 2.0.

### Documentation
- [x] Upgrade note added.
- [x] PR description explains scope.


grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/UrlMappingsIndexProperties.java
```diff
@@ -0,0 +1,80 @@
...
+                return EMPTY;
+            }
+            Properties properties = new Properties();
+            properties.load(inputStream);
+            return new UrlMappingsIndexProperties(true, properties);
+        }
+        catch (IOException e) {
+            throw new IllegalStateException("Unable to load " + LOCATION, e);
+        }
+    }
```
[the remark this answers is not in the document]
jamesfredley: Addressed. `load(ClassLoader)` now tries the thread context classloader **first**, then falls back to the provided classloader, guarding for `null` (and a `SecurityException` from acquiring the TCCL). Failures are isolated **per loader**: an `IOException` *or* any non-fatal `RuntimeException` (including the `IllegalArgumentException` from malformed properties, and a restricted/throwing TCCL) is logged at debug and skips that loader, so a broken TCCL still falls through to the provided classloader; `EMPTY` is only returned after every loader has been tried, and `Error` is never swallowed. Added tests for the throwing-classloader (soft miss), malformed content, TCCL precedence, and valid-descriptor paths.

jamesfredley:
## Review feedback addressed

Merged the latest `8.0.x` and pushed follow-up commits.

**Copilot review comment (classloader / exception handling on `UrlMappingsIndexProperties.load`):**

- The loader now tries the **thread context classloader first**, then the provided classloader, so an application-packaged `META-INF/grails/url-mappings-index.properties` is discovered.
- Failures are isolated **per loader** and treated as a soft miss: an `IOException` or any non-fatal `RuntimeException` (including the malformed-properties `IllegalArgumentException`, and a restricted/throwing TCCL) is logged at debug and skips that loader, so a broken TCCL still falls through to the provided classloader. `EMPTY` is only returned after every loader has been tried, and `Error` is never swallowed - so an optional/missing/malformed descriptor can never abort application startup.

**Review-pass fixes:** added the ASF license header to the new `urlMappingsPrecompute.adoc`, corrected the version references from 8.1 to **8.0** (matching the target branch and `@since`), and clarified the wording so it is explicit that runtime URL-mapping evaluation stays authoritative whether or not a descriptor is present (the descriptor is only *reserved/detected* in this seed). Added tests for missing, malformed, unreadable/throwing, TCCL-precedence, and valid-descriptor paths.

Local verification: `:grails-web-url-mappings:test --tests UrlMappingsIndexPropertiesSpec` (5 tests) passes.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/UrlMappingsIndexProperties.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Descriptor for a future build-time URL mappings index.
+ *
+ * @since 8.0.x
+ */
+public final class UrlMappingsIndexProperties {
+
+    public static final String LOCATION = "META-INF/grails/url-mappings-index.properties";
```
jdaugherty: `getResourceAsStream` returns the first match on the classpath, so a single well-known location only works while exactly one jar provides it. URL mappings come from the application *and* from every plugin (`UrlMappingsHolderFactoryBean` merges each plugin's `UrlMappings` with its `pluginIndex`), and a build-time generator would naturally emit one index per compiled jar. At that point classpath order silently decides which index wins.

`grails-plugin.xml` handles this by enumerating every copy with `getResources` (`PluginUtils`). #16000 handles it by resolving relative to the application's code-source root (`IOUtils.findRootResource`). Either one, chosen with the generator in mind, would fit better than a single classpath-wide lookup.

The format is worth deciding at the same time. `Properties.load(InputStream)` reads ISO-8859-1 into a flat string map, which is an awkward fit for a trie or reverse-routing table, and non-Latin-1 URL segments would need `\u` escaping.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/UrlMappingsIndexProperties.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public static UrlMappingsIndexProperties load(ClassLoader classLoader) {
+        ClassLoader threadContextClassLoader;
+        try {
+            threadContextClassLoader = Thread.currentThread().getContextClassLoader();
+        }
+        catch (RuntimeException e) {
+            threadContextClassLoader = null;
+        }
+        for (ClassLoader loader : new ClassLoader[] {threadContextClassLoader, classLoader}) {
```
jdaugherty: With the TCCL tried first, `load(classLoader)` ignores its argument whenever the TCCL can see a descriptor, so a caller that passes an explicit loader gets the TCCL's copy instead. If `DefaultUrlMappingsHolder` needs TCCL-first, resolving it at the call site keeps the parameter meaningful. For example, `ClassUtils.getDefaultClassLoader()` tries the TCCL and then the class's loader.

Falling through after a failure has a similar problem. If the TCCL's copy is malformed, the second loader usually finds the same resource again through parent delegation, or picks up a different jar's copy. Neither seems like what we'd want once something actually reads the index.

The `RuntimeException` catch around `getContextClassLoader()` above can't be reached. That call only throws under a SecurityManager, which is off by default on our JDK 21 baseline and permanently disabled from JDK 24. I'd drop it.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/UrlMappingsIndexProperties.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+            try (InputStream inputStream = loader.getResourceAsStream(LOCATION)) {
+                if (inputStream == null) {
+                    continue;
+                }
+                Properties properties = new Properties();
+                properties.load(inputStream);
+                return new UrlMappingsIndexProperties(true, properties);
+            }
+            catch (IOException | RuntimeException e) {
```
jdaugherty: Two things on this catch:

- `Properties.load` signals bad input with `IOException` or `IllegalArgumentException`. Catching every `RuntimeException` also turns real bugs (an NPE, say) into a debug-level soft miss.
- The per-loader isolation this enables isn't tested. In the malformed and unreadable specs the failing loader is always the *last* one tried, so changing this block to `return EMPTY;` leaves all seven specs passing. A spec with a throwing TCCL and a valid fallback loader would pin it down. I tried that locally with a TCCL whose `getResourceAsStream` throws only for `LOCATION` and delegates everything else, and it finds the fallback copy on this branch. (Throwing for every resource breaks slf4j-simple, which loads `simplelogger.properties` through the TCCL.)

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingsHolder.java
```diff
@@ -113,6 +114,7 @@ public DefaultUrlMappingsHolder(List<UrlMapping> mappings, List excludePatterns)
     public DefaultUrlMappingsHolder(List<UrlMapping> mappings, List excludePatterns, boolean doNotCallInit) {
         urlMappings = mappings;
         this.excludePatterns = excludePatterns;
+        this.precomputedIndexProperties = UrlMappingsIndexProperties.load(DefaultUrlMappingsHolder.class.getClassLoader());
```
jdaugherty: This does two classpath lookups on every `new DefaultUrlMappingsHolder(...)`: the factory bean, `UrlMappingsFactory`, `GspAutoConfiguration`, and unit tests all construct one. It runs even when `doNotCallInit` is true, and today the result only feeds a debug log. Since the goal is lower startup cost, I'd leave the holder alone until something consumes the index. Then load it once (e.g. in `UrlMappingsHolderFactoryBean`) rather than per instance.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingsHolder.java
```diff
@@ -233,6 +238,10 @@ public List getExcludePatterns() {
         return excludePatterns;
     }
 
+    public UrlMappingsIndexProperties getPrecomputedIndexProperties() {
```
jdaugherty: This getter and the public `UrlMappingsIndexProperties` class become API we have to keep once released, and nothing outside the spec calls them. #16000 keeps its reader package-private. That works here too, since the spec is already in `org.grails.web.mapping`.

grails-web-url-mappings/src/test/groovy/org/grails/web/mapping/UrlMappingsIndexPropertiesSpec.groovy
```diff
@@ -0,0 +1,130 @@
...
+
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+
+import grails.web.mapping.UrlMapping
+import spock.lang.Specification
+
+class UrlMappingsIndexPropertiesSpec extends Specification {
+
+    void 'missing build-time URL mapping index keeps runtime fallback active'() {
```
jdaugherty: This passes no matter what the index code does. There's no descriptor on the test classpath and the mapping list is empty, so `matchAll('/books').length == 0` holds either way. The behavior the PR promises, that runtime matching stays authoritative *when a descriptor is present*, isn't tested.

That case can be reached through the TCCL, the same way the precedence spec does it. Build a holder with a descriptor on the TCCL and a real `"/books"(controller: 'book', action: 'list')` mapping (e.g. via `AbstractUrlMappingsSpec`). Then assert `precomputedIndexProperties.present` and that `/books` still matches `book`/`list`. I tried this locally and it passes on this branch, so it only needs adding.

grails-web-url-mappings/src/test/resources/simplelogger.properties
```diff
@@ -17,4 +17,9 @@
...
 #
 
-org.slf4j.simpleLogger.defaultLogLevel=info
\ No newline at end of file
+org.slf4j.simpleLogger.defaultLogLevel=info
+
+# Enabled at DEBUG so UrlMappingsIndexPropertiesSpec/DefaultUrlMappingsHolderSpec can
+# exercise the (non-behavioral) debug-log lines guarded by LOG.isDebugEnabled().
+org.slf4j.simpleLogger.log.org.grails.web.mapping.UrlMappingsIndexProperties=debug
+org.slf4j.simpleLogger.log.org.grails.web.mapping.DefaultUrlMappingsHolder=debug
```
jdaugherty: This DEBUG override applies to every spec in the module, not just the new ones, and `DefaultUrlMappingsHolder` logs once per mapping on every match. Running `:grails-web-url-mappings:test` on this branch produced 5,571 DEBUG lines from it: about 1 MB of the module's ~1.03 MB of captured test output. If the debug line needs verifying, I'd capture it inside the spec rather than raise the level module-wide. Also, the `DefaultUrlMappingsHolderSpec` named in the comment above doesn't exist.

grails-doc/src/en/guide/toc.yml
```diff
@@ -37,6 +37,7 @@ gettingStarted:
 upgrading:
   title: Upgrading from the previous versions
   upgrading80x: Upgrading from Grails 7 to Grails 8
+  urlMappingsPrecompute: URL mapping precomputation seed
```
jdaugherty: This sits alongside the version upgrade guides, but an upgrading user has nothing to act on: no descriptor is generated or read yet. "Seed" is also internal terminology. Advertising a reserved location now invites apps and plugins to put files there before the format is defined. I'd drop `urlMappingsPrecompute.adoc` until the generator exists. At that point it belongs in the feature docs with a What's New entry, not in the upgrade guide.

jdaugherty (changes requested):
I've moved this to `8.1.x`  because this adds public API (`UrlMappingsIndexProperties`, `DefaultUrlMappingsHolder.getPrecomputedIndexProperties()`, and a documented reserved location) that nothing reads yet. Once 8.0.0 ships, all three become compatibility commitments before the index format or its generator exists.

Before this merges, I'd like it to arrive with (or directly ahead of) the code that generates and consumes the index, so the location, format, and lookup rules are designed around real use. #16000 is a useful comparison. It tackles the same kind of problem with:
- a package-private reader
- a lookup scoped to the application's code-source root
- a UTF-8 line format
- a system-property kill switch

If both land as they are, `META-INF/grails/` ends up with two inconsistent index conventions, so it'd be worth settling on one approach for both.

The description points to "Google Doc 4.2" for the design. Could the relevant part go into an issue or a dev@ thread so everyone can follow it?

Details are inline. Two small nits that don't need their own threads:
- `@since 8.0.x`: elsewhere we use `@since 8.0` (no `.x`); with the move this becomes `@since 8.1`.
- `propertyNames()` doesn't need the `present ?` check, since `EMPTY` already wraps an empty `Properties`.

PR #15958, opened 2026-07-10
Seed groovy-json facade over org.grails.web.json

jamesfredley:
### What was found

| Problem | Impact |
| --- | --- |
| `org.grails.web.json` is a vendored JSON.org-era fork | Maintenance and security risk |
| Converters still depend on public facade types | Cannot rip out internals in one step |
| Codebase 4.1 marked this strategic 8.x/9.0 work | Needs a deprecation facade seed |

### What changed

| Area | Change |
| --- | --- |
| Facade | Add `GroovyJsonFacade` bridging toward `groovy-json` |
| Public API | Keep existing `org.grails.web.json` types |
| Docs | Upgrade note with 9.0 removal plan |
| Tests | Facade parse/serialize coverage |

### Out of scope / follow-up

| Topic | Status |
| --- | --- |
| Full internal rewrite of all converters | Follow-up |
| Removing public JSONObject/JSONArray | 9.0 |
| Benchmarks vs current fork | Follow-up |

### Related MD topics

| Source | Topic |
| --- | --- |
| Codebase 4.1 | Vendored JSON.org-era fork |

### Issue and Scope
- [x] Starter deprecation path.
- [x] Public API preserved.
- [x] Single focused seed.
- [x] Targets `8.0.x`.

### Code Quality
- [x] Tests added.
- [x] Focused web-common tests run.
- [x] No mass reformatting.

### Licensing and Attribution
- [x] Apache License 2.0.

### Documentation
- [x] Upgrade note added.
- [x] Scope explained.


jdaugherty:
Instead of using our vendored version we should just adopt the upstream version.  I disagree with this change.

jamesfredley:
Just to align on the goal - I think we actually want the same end state ("adopt upstream"), and the facade is the non-breaking route to it rather than an alternative.

Per the pre-release review (Codebase 4.1): `grails-web-common` carries a custom JSON stack - `JSONObject.java` (~1,191 lines), `JSONArray`, and a JavaCC-generated parser (`JsonParserTokenManager.java`, ~1,038 lines) - that predates `groovy-json`/Jackson and is a standing maintenance and security-surface cost (custom JSON parsers are where JSON CVEs live).

The recommended starting point there is exactly to move onto upstream `groovy-json` **internally** while keeping the public `org.grails.web.json.*` type surface as a **deprecated facade**:

- `groovy-json` is already a BOM dependency, so this adds **zero new deps**.
- The reason for a facade instead of a hard swap is backward compatibility: applications, plugins, and `grails-converters` (`JSON.java`) reference `org.grails.web.json.JSONObject` / `JSONArray` directly. Deleting that public API outright is a breaking change and isn't safe in 8.x.
- So the facade *is* adopting upstream - it reimplements the internals on `groovy-json` and validates against the existing converter test suites in `grails-test-suite-web` - while deferring removal of the vendored public API to 9.0.

If your objection is specifically to keeping the vendored **public API** around at all, that's the 9.0 removal step this facade sets up. Would it help if I framed the PR explicitly as "reimplement internals on groovy-json; deprecate the public facade for 9.0 removal" so the upstream-adoption intent is unambiguous?

jdaugherty:
groovy-json has behavioral differences that has blocked me on trying this before and changing the json parsing stack is a major change.  I'd rather we update to the successor of what was vendored, which in this case is [openjson](https://github.com/openjson/openjson).  It's a clean room implementation that is maintained, and it was driven by the ASF having problems with the original library.  This maintains strict compatibility and is less likely to introduce regressions to behavioral differences.

PR #15959, opened 2026-07-10
Add Automatic-Module-Name jar manifests

jamesfredley:
### What was found

| Problem | Impact |
| --- | --- |
| Published jars lack `Automatic-Module-Name` | Poor JPMS readiness |
| Google Doc 6.2 listed this as low-risk 8.x step | Easy starter before full module-info |
| Manual per-module names would drift | Needs build-logic convention |

### What changed

| Area | Change |
| --- | --- |
| Build-logic | Generate `Automatic-Module-Name` in jar manifests |
| Convention | Applied through CompilePlugin/build conventions |
| Docs | Upgrade note describing module names |
| Tests | Build-logic coverage for manifest attribute |

### Out of scope / follow-up

| Topic | Status |
| --- | --- |
| Full `module-info.java` | Later |
| Split-package remediation | Later |
| Consumer module-path validation suite | Follow-up |

### Related MD topics

| Source | Topic |
| --- | --- |
| Google Doc 6.2 | JPMS module-system readiness |

### Issue and Scope
- [x] Low-risk JPMS readiness starter.
- [x] No module-info rewrite.
- [x] Single focused change.
- [x] Targets `8.0.x`.

### Code Quality
- [x] Tests added.
- [x] Focused build-logic tests run.
- [x] No mass reformatting.

### Licensing and Attribution
- [x] Apache License 2.0.

### Documentation
- [x] Upgrade note added.
- [x] Scope explained.

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
@@ -83,9 +83,25 @@ class CompilePlugin implements Plugin<Project> {
                     'Implementation-Version': lookupPropertyByType(project, 'grailsVersion', String),
                     'Implementation-Vendor': 'grails.apache.org'
             )
+            if (!jar.archiveClassifier.present) {
+                jar.manifest.attributes('Automatic-Module-Name': automaticModuleName(project))
+            }
             // Explicitly fail since duplicates indicate a double configuration that needs fixed
             jar.duplicatesStrategy = DuplicatesStrategy.FAIL
         }
+    }
+
+    static String automaticModuleName(Project project) {
+        String configuredModuleName = project.findProperty('automaticModuleName') as String
+        if (configuredModuleName) {
+            return configuredModuleName
+        }
+        String groupPrefix = project.group == null ? 'org.apache.grails' : project.group.toString()
+        return "${groupPrefix}.${project.name}"
+                .replace('-', '.')
+                .replaceAll('[^A-Za-z0-9_.]', '.')
+                .replaceAll('\\.+', '.')
+                .replaceAll('^\\.|\\.$', '')
     }
 
     private static void configureCompiler(Project project) {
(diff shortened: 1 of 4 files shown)


jdaugherty:
We cannot do this PR because we have overlapping module names.  Until we repackage all of our jars, we cannot proceed here and thus is why we've excluded this information to date (we don't want to give the appearance we're compatible with JPMS until we actually are).

jamesfredley:
You're right that we have split packages - I verified it against the source tree. The same package is declared across multiple published jars in at least:

- `org.grails.plugins.web.taglib` -> **4 modules** (`grails-gsp`, `grails-layout`, `grails-sitemesh3`, `grails-web-gsp-taglib`)
- `org.grails.events` -> `grails-events-core`, `grails-events-transforms`
- `org.grails.events.spring` -> `grails-events`, `grails-events-spring`
- `org.grails.async.transform.internal` -> `grails-async`, `grails-async-core`

Those genuinely break module-path consumption - a consumer placing these jars on the module path hits "package X in both module A and module B" - so the concern about advertising a JPMS readiness we don't have is valid.

One scope clarification for the record: `Automatic-Module-Name` only stabilizes the module *name* (versus the volatile filename-derived automatic name); it doesn't itself create or enforce the split-package rule, and those packages already break module-path use today regardless of this PR. The pre-release review (Google Doc 6.2) framed automatic module names as a low-risk manifest step and full `module-info` (which would *expose* split packages) as later work. But your point that a stable module name signals module-path intent stands.

Given that, I think the right call is your (a): **defer this until the split-package modules are repackaged**, rather than shipping mixed signals. The alternative - adding `Automatic-Module-Name` only to the clean majority and explicitly excluding the split-package modules (the gsp/layout/sitemesh3/taglib set, events*, async*) - is possible but muddier. I'll hold/close this PR and instead open a tracked task for the split-package repackaging, which is the real prerequisite. Sound right?

jdaugherty:
I agree, this is the *right* solution wrong time, but we need to focus on repackaging.  I'd like to make this an effort in Grails 9 if possible or let it gradually happen as we refactor.  I don't think we should ship anything that says we're JPMS compatible until we actually are.

PR #15962, opened 2026-07-10
Fix embedded where queries

### LINKED ISSUE #15955: where {} queries fail when accessing properties of embedded components
where {} queries fail when accessing a property of an embedded component, even though the property is referenced using the expected dotted notation.

Hibernate throws the following exception:

```
org.hibernate.QueryException: Criteria objects cannot be created directly on components.
Create a criteria on owning entity and use a dotted property to access component property: extRef1
```

The query already uses a dotted property (`extRef1.value`), so it appears that GORM is translating the where {} closure into a Hibernate Criteria incorrectly.

The same happens with:
```
extRef1.provider == "SAP"
```
so the issue does not appear to be related to the =~ operator, but to accessing any property of an embedded component.

## Expected behavior

The query should be translated into a predicate on the embedded property (e.g. `extRef1.value`) and execute successfully.

## Actual Behaviour

Hibernate throws:
```
org.hibernate.QueryException: Criteria objects cannot be created directly on components.
Create a criteria on owning entity and use a dotted property to access component property: extRef1
```

*NOTE: Removing the extRef1.value predicate makes the query execute correctly.*

### Domain model
```
class WorkItem extends SystemManaged {
    String description
}
```

Classes in `/main/src`
```
abstract class SystemManaged implements GormEntity {
    ExternalRef extRef1

    static embedded = ['extRef1']
}

class ExternalRef {
    String provider
    String value
}
```

### Query
```
def query = WorkItem.where {}

String search = "ABC"

query = query.where {
    extRef1.value =~ "%${search}%"
}
```

## Environment Information

Grails: 7.1.0

## Example Application

_No response_

## Version

7.1.0

jdaugherty:
## Description

Fixes #15955

`where {}` queries failed with a Hibernate `QueryException` ("Criteria objects cannot be created directly on components") when a predicate referenced a property of an embedded component (e.g. `extRef1.value =~ "%ABC%"`) and was combined with any other condition (`&&`, `||`, or additional criteria). A lone embedded predicate worked because `AbstractHibernateQuery.add()` special-cases `Embedded` associations, but once the predicate was nested inside a junction, translation went through `AbstractHibernateCriterionAdapter`, whose `DetachedAssociationCriteria` and `AssociationQuery` adapters unconditionally created a Hibernate sub-criteria/alias on the component — which Hibernate rejects.

Changes:

- **grails-datamapping-core** — `DetachedCriteriaTransformer` no longer discards an embedded block's component type when resolving the association type from generics (a collection association carries its element type in its generics; an embedded component does not), which previously caused the block's criteria to be silently dropped. It also recomputes variable scopes after applying the where transform so the generated nested closures capture exactly the variables they reference instead of inheriting the outer closure's scope nondeterministically. `AbstractDetachedCriteria` now uses explicit null checks instead of Groovy truth when looking up an existing association criteria — Groovy truth on a `DetachedCriteria` invokes `asBoolean()`, which executed the criteria as a spurious query on every repeated association-block reference (and failed outright for embedded components, whose class is not a queryable root entity).
- **grails-data-hibernate5** — the `DetachedAssociationCriteria` and `AssociationQuery` criterion adapters now short-circuit for `Embedded` associations and apply the nested criteria using dotted property paths (`extRef1.value`), mirroring the existing top-level handling in `AbstractHibernateQuery.add()`.
- **grails-data-hibernate7** — `CriteriaMethodInvoker` now recognizes JPA `EMBEDDED` attributes (which the JPA metamodel does not consider associations, but the GORM model does) and builds a `DetachedAssociationCriteria` for the embedded block without creating a join — a component's columns live in the owning entity's table, so an explicit join-type argument on an embedded block is intentionally ignored.
- **grails-data-mongodb** — `MongoQuery` previously prefixed *every* key of an embedded sub-query with the embedded property's path, so an embedded block containing a junction produced keys like `{"extRef1.$or": [...]}` that silently matched nothing. The rewrite is now recursive: logical operators (`$and`, `$or`, …) stay at the correct level and only property keys are qualified. A `$`-operator whose value cannot be rewritten (e.g. `$where` from a property-to-property comparison, or `$text`) inside an embedded block now throws `UnsupportedOperationException` instead of silently matching nothing.
- **grails-data-simple** — the in-memory implementation had the same class of bug: `SimpleMapQuery.queryAssociation()` treated `Embedded` (a `ToOne` subtype) as a real association and called `session.retrieve()` with the inlined embedded map as an id, throwing `NonPersistentTypeException`. It now evaluates the criteria directly against the stored embedded value.
- **grails-datamapping-tck** — new `WhereQueryEmbeddedSpec` plus `WorkItem`/`SystemManaged`/`ExternalRef` test domains reproducing the reported shape (embedded component declared in an abstract non-entity superclass). Covers `=~` and `==` on embedded properties, chained `query.where {}` composition, aliased blocks, and the conjunction/disjunction cases that triggered the bug. Where-query specs that were previously duplicated per implementation (`WhereQueryIssueVerificationSpec`, `WhereQueryLeftJoinSpec`, `DeleteAllSpec`) were moved into the TCK, and the TCK test properties were standardized to one location in the Gradle test configs. The TCK runs against all four implementations: hibernate5, hibernate7, simple (in-memory), and mongodb.
- **grails-test-examples (geb)** — includes an unrelated flaky-test fix in `PerTestRecordingSpec`: recording-directory discovery is now scoped to directories timestamped at or after the current JVM's start time instead of racing on the "most recent" directory, which broke under parallel test execution.

All test suites for the affected modules pass, along with the project style checks. The full `./gradlew build --rerun-tasks` suite was not run; verification covered the affected modules.

## Contributor Checklist

Please review the following checklist before submitting your pull request. Pull requests that do not meet these requirements may be closed without review.

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. If no approved issue exists, please give background on why this change is necessary.  Tickets are preferred for release change log history.
- [x] This PR contains a **single, focused change**. Unrelated changes should be submitted as separate pull requests. (One exception, disclosed above: the `PerTestRecordingSpec` flaky-test fix.)

### Code Quality

### Licensing and Attribution

### Documentation

- [x] If this PR introduces breaking changes or changes that require user action during an upgrade, I have updated the **Upgrade Notes** for the corresponding version in the Grails Guide.


borinquenkid (changes requested):
Needs coverage added as per codecov

jdaugherty:
I've gone ahead and moved the duplicated tests in h5/h7 for where queries to the tck as well as part of this

borinquenkid (changes requested):
Got it. Looking at the expanded file list in the Codecov report, the remaining coverage gaps are spread across the Simple, Core, and MongoDB modules.

Here is the regenerated list of missing test scenarios needed to cover the newly visible files:

### 1. `SimpleMapQuery.groovy` (grails-data-simple)

The newly added `Embedded` handling block (lines 295-300) inside `queryAssociation` is completely uncovered (0% patch coverage).

* **Missing Scenario:** You need a test that executes a criteria or Where query against an `Embedded` property, specifically utilizing the `SimpleMapEntityPersister` (the in-memory/simple datastore).
* *Note:* To fully cover this, ensure the test hits both branches of the `if (embedded != null)` check on line 297 (i.e., one record where the embedded component exists, and one where it is null).

### 2. `DetachedCriteriaTransformer.java` (grails-datamapping-core)

The AST transformation logic added to retain the type of embedded components (lines 812-818) is completely missed.

* **Missing Scenario:** A unit test verifying the AST transformation of a Where query block that targets an embedded component. The test needs to trigger compilation of a detached criteria where `AstUtils.isDomainClass(type)` is false, forcing it into the `getAssociationTypeFromGenerics` block to retain the `associationType`.

### 3. `MongoQuery.java` (grails-data-mongodb)

Most of the MongoDB embedded query rewriting is covered, but there are a few missed branches (partial coverage) regarding logical operators within embedded queries.

* **Missing Scenario A (Line 202):** The condition `if (associatedEntity instanceof EmbeddedPersistentEntity || association instanceof Embedded)` has partial coverage. You likely have a test covering one side of this OR condition but not the other.
* **Missing Scenario B (Line 761):** The condition `if (key.charAt(0) == '$' && value instanceof List)` inside `prefixEmbeddedQuery` has partial coverage. You need a MongoDB query on an embedded property that utilizes a logical operator like `$and` or `$or` (which generates a list of conditions) to hit the `true` branch of this check.
* **Missing Scenario C (Line 764):** The condition `if (element instanceof Document)` within the list iteration also has partial coverage. The test for Scenario B will likely cover the `true` branch, but you may need a scenario where an element inside a Mongo `$in` or similar array operator evaluates to `false` (e.g., checking a list of primitives instead of nested Documents).

borinquenkid (approved):
LGTM

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/tests/WhereQueryEmbeddedSpec.groovy
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+
+        then:
+        results.size() == 1
+        results[0].description == 'first'
+    }
+
+    @Issue('https://github.com/apache/grails-core/issues/15955')
+    void 'where query with a disjunction inside an embedded association block'() {
```
codeconsole: This feature method has the same name as the one at line 223 (`'where query with a disjunction inside an embedded association block'`). It only compiles because Spock renames feature methods before Groovy's duplicate-signature check, but both tests report under an identical display name, so a failure in either is ambiguous in test reports. This one combines the embedded-block junction with an outer predicate — suggest renaming to something like `'where query with a junction inside an embedded association block combined with another predicate'`.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
```diff
@@ -60,62 +65,75 @@ class PerTestRecordingSpec extends ContainerGebSpec {
...
-        if (timestampedDirs) {
-            // Get the most recent directory
-            recordingDir = timestampedDirs.sort { it.name }.last()
-        }
-
-        then: 'the recording directory should be found'
-        recordingDir != null
-
-        when: 'getting all video recording files (mp4 or flv) from the recording directory'
+        when: 'collecting the recordings this spec produced during this test run'
```
codeconsole, line 72: This flaky-test fix is sound (scoping recording-dir discovery to the current JVM's start time instead of racing on "most recent directory"), but it's unrelated to embedded where queries. Ideally it would be its own PR; at minimum please mention it in the PR description so it doesn't get lost in the merge record.

grails-data-hibernate7/core/src/main/groovy/grails/orm/CriteriaMethodInvoker.java
```diff
@@ -193,21 +193,29 @@ protected Object tryAssociationOrJunction(String name, CriteriaMethods method, O
...
                 builder.setTargetClass(associationClass);
-                JoinType joinType;
-                if (hasMoreThanOneArg) {
-                    joinType = builder.convertFromInt((Integer) args[0]);
-                } else if (associationClass.equals(oldTargetClass)) {
-                    joinType = JoinType.LEFT; // default to left join if joining on the same table
-                } else {
-                    joinType = builder.convertFromInt(0);
-                }
+                if (!embedded) {
```
codeconsole, line 208: Nit: an explicit join-type argument on an embedded block (e.g. `extRef1(JoinType.LEFT) { ... }`) is now silently ignored. That's the right behavior — a component can't be joined — but a short comment noting the argument is intentionally dropped (or a debug log) would help the next reader.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/query/MongoQuery.java
```diff
@@ -751,6 +749,33 @@ public static void populateMongoQuery(final EmbeddedQueryEncoder queryEncoder, D
         }
     }
 
+    /**
+     * Rewrites a query built against an embedded entity so it applies to the owning document,
+     * qualifying each property name with the embedded property's path. Logical operators such as
+     * {@code $and} and {@code $or} must stay at the current level (a key like {@code extRef1.$and}
+     * matches nothing), so their nested documents are rewritten recursively instead.
+     */
+    private static void prefixEmbeddedQuery(String prefix, Document source, Document target) {
```
codeconsole, line 758: Two notes, neither blocking:

1. `$`-keys with non-List values — `$where` (from property-to-property comparisons) and `$text` — still fall into the else branch and get mis-prefixed to e.g. `extRef1.$where`. That matches the old (broken) behavior so it's not a regression, but since this method now understands operator keys, consider either documenting the limitation here or throwing `UnsupportedOperationException` for a `$`-key that isn't a rewritable list, so it fails loudly instead of silently matching nothing.

2. Nit: raw `List`/`ArrayList` — `List<Object>` would avoid the unchecked warning, though the surrounding file is raw-typed too.

codeconsole (approved):
Approving with a few minor changes requested (inline). The fix is correct and consistent across all four implementations, and I verified the two extra concerns beyond the diff itself:

**MongoDB impact**: positive, not neutral. The `MongoQuery` change fixes a real silent wrong-results bug — the old code prefixed every key of an embedded sub-query, so an embedded block containing a disjunction produced `{"extRef1.$or": [...]}`, a key that matches nothing. The new recursive rewrite keeps logical operators at the correct level, and the new TCK tests exercise exactly this shape on the Mongo suites (all green on the head commit).

**Performance impact**: neutral to slightly positive. Non-embedded association paths gain only an `instanceof`/attribute-type branch. The `AbstractDetachedCriteria` null-check fix is an actual win — the old Groovy-truth check invoked `DetachedCriteria.asBoolean()`, executing the criteria as a spurious database query on every repeated association-block reference. The Mongo handler also now computes `getPropertyName()` once instead of per key.

Two things to clean up in the **PR description** before merge:
- It states "Hibernate 7 and MongoDB required no code changes" — no longer true; both `CriteriaMethodInvoker` and `MongoQuery` have functional changes in the current head.
- The closing sentence ("Say the word and I'll run the full build in the background...") is a leftover conversational artifact and should be removed — the description becomes the merge record.

PR #15963, opened 2026-07-10
Add workaround for GROOVY-12146

jdaugherty:
See https://lists.apache.org/thread/98vpor0k2crzlqn4qk91ms7kp8pfm497 for history.

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
@@ -125,6 +125,13 @@ class CompilePlugin implements Plugin<Project> {
                 if (System.getenv('SUPPRESS_DEPRECATION_WARNINGS') == 'true') {
                     it.options.compilerArgs += ['-Xlint:-removal']
                 }
+                // Canonicalize annotation member order for reproducible builds. Annotations copied from
+                // precompiled classes (e.g. @DelegatesTo on trait methods woven into controllers and GORM
+                // entities) have their members ordered by Class.getDeclaredMethods(), which varies between
+                // JVM runs. The GrailsGradlePlugin merges this script with its own configuration script
+                // when both are present.
+                it.groovyOptions.configurationScript =
+                        GradleUtils.findRootGrailsCoreDir(project).file('gradle/groovy-compile-configscript.groovy').asFile
             }
         }
     }
--- a/etc/bin/normalize-annotations.groovy
+++ b/etc/bin/normalize-annotations.groovy
@@ -1,0 +1,147 @@
+#!/usr/bin/env groovy
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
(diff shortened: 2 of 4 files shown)


borinquenkid (approved):
I imagine the proof in the workaround working

codeconsole (approved):
Reviewed the issue and the fix, including verifying both scripts locally. Approving.

**Issue (GROOVY-12146):** Groovy 5 populates `AnnotationNode` members for annotations copied from precompiled classes (e.g. `@DelegatesTo` on trait methods woven into controllers/GORM entities) via `Class.getDeclaredMethods()`, whose order is unspecified (HotSpot orders by internal Symbol address). The Groovy 4 fix (GROOVY-11715, TreeMap canonicalization) was reverted on master, so Grails 8 artifacts fail reproducible-build verification. The two-layer workaround here is the right shape: fix emission order at compile time, and keep a normalization backstop in the decompile-diff so verification is resilient to any residual nondeterminism.

**What I verified:**

- `gradle/groovy-compile-configscript.groovy` — reproduced the scenario locally with Groovy 5.0.0: compiled a trait with a multi-member `@DelegatesTo` parameter annotation, then compiled an implementing class with and without `--configscript`. Without: members emitted in reflection order (`genericTypeIndex, type, value, target, strategy`). With: alphabetical. Works as advertised.
- `AnnotationNode.getMembers()` returns the live internal `LinkedHashMap` (or `Collections.emptyMap()` when unset, which the `size() > 1` guard never mutates), so the `clear()`/`putAll(new TreeMap<>(...))` in-place sort is safe and order-preserving through bytecode generation. `INSTRUCTION_SELECTION` runs after trait weaving and before class generation — the right phase.
- The claim in the `CompilePlugin` comment holds: `GrailsGradlePlugin.configureGroovyCompiler` reads any pre-set `configurationScript` in its `doFirst` and concatenates it with its own generated script, so projects applying both plugins keep both behaviors.
- `GradleUtils.findRootGrailsCoreDir` resolves via `.asf.yaml`, so the script path also resolves correctly from the `grails-gradle`/`grails-forge` included builds, whose subprojects apply the same compile convention plugin.
- `etc/bin/normalize-annotations.groovy` — exercised it on synthetic decompiled sources: sorts flat members, nested annotations, and annotation arrays; leaves positional values (`@Single("x")`, `@SuppressWarnings({...})`) untouched; handles commas/parens inside string literals; is idempotent on a second pass.

**Non-blocking observations:**

1. The normalize script's regex scan also rewrites annotation-*like* text inside string literals of decompiled sources (e.g. `"@Fake(b=2, a=1)"` becomes `"@Fake(a=1, b=2)"`). Since the transformation is deterministic and applied to both sides of the diff, it can't cause a false failure; the theoretical case it could mask (two builds whose string constants differ only by member order in annotation-shaped text) is vanishingly unlikely, and the checksum stage still flags the jar. Fine as-is for a comparison-only tool.
2. The config script's `visitClass` doesn't visit record components or package-info annotations, and classes materialized during `CLASS_GENERATION` (e.g. closure classes) are outside its reach. All are unlikely carriers of copied multi-member annotations, and the verify-script normalization backstops them anyway.

Might be worth linking GROOVY-12146 in a `// TODO remove when fixed upstream` so this can be unwound if Groovy restores the GROOVY-11715 behavior — but the file headers already reference the issue context well enough.

PR #15964, opened 2026-07-10
Apply SiteMesh 3 decoration at the bean-definition level

codeconsole:
### Summary

Since #13863 stopped auto-adding `@EnableWebMvc`, Spring Boot's `WebMvcAutoConfiguration` is active for Grails servlet apps and registers a `ContentNegotiatingViewResolver` as the highest-precedence `viewResolver`. While initializing, CNVR collects every `ViewResolver` bean via `getBeansOfType`, which force-creates the lazy `jspViewResolver`. If that happens before the SiteMesh 3 `BeanPostProcessor` wrap (#15585) is in effect, CNVR captures the **raw, non-decorating resolver** and keeps rendering HTML through it — GSP layouts silently stop applying. Whether the race is hit depends on bean-creation order, so it varies with the plugin set: some apps on 8.0.0-M3 decorate fine, others lose all layouts.

### Diagnosis (reproduced on 8.0.0-M3)

In an affected app:

- `jspViewResolver` bean = `GrailsSiteMeshViewResolver@55871443` wrapping `ScaffoldingViewResolver@344207224` — the wrap itself worked.
- CNVR's delegate list contained `ScaffoldingViewResolver@344207224` — **the same instance as the wrapper's inner**, captured pre-wrap.
- Requests with `Accept: text/html` rendered undecorated (CNVR short-circuits); requests with an Accept header CNVR cannot satisfy fell through to the wrapped resolver and rendered **fully decorated**.

### Fix

- **`Sitemesh3ViewResolverDefinitionPostProcessor`** (registered by `Sitemesh3AutoConfiguration`) rewrites the `jspViewResolver` bean **definition** into a `GrailsSiteMeshViewResolver` with the original definition embedded as the inner bean — mirroring the SiteMesh 2 module's `GrailsLayoutViewResolverPostProcessor`. The decorating resolver is what gets instantiated no matter how early a consumer forces the bean into existence, so the race is structurally impossible. It skips contexts where decoration is impossible (no `contentProcessor`/`decoratorSelector`, e.g. grails-testing-support unit-test contexts) and definitions that already decorate (the SiteMesh 2 layout resolver, or an existing `SiteMeshViewResolver`).
- The existing bean post-processor is retained as a safety net for instance-registered resolvers and to keep upstream's wrap-all post-processor from registering; upstream never re-wraps a `SiteMeshViewResolver`, so the two mechanisms cannot double-decorate.

### Modernization (builds on #15934)

`Sitemesh3GrailsPlugin` no longer declares `doWithSpring()` at all:

- The SiteMesh property defaults moved to **`Sitemesh3EnvironmentPostProcessor`** (`META-INF/spring.factories`), contributed with lowest precedence before refresh — which also removes the post-hoc `PropertySourcesConfig` reassignment the plugin used to perform.
- `grailsRenderViewMutator` is registered through **`beanRegistrar()`**, making this the first in-tree plugin on the modern registration API.

### Testing

- New `Sitemesh3ViewResolverDefinitionPostProcessorSpec` (9 tests) including a regression test that simulates CNVR's early `getBeansOfType(ViewResolver)` sweep against a registry with **no bean post-processors at all** and asserts the captured instance decorates.
- New `Sitemesh3EnvironmentPostProcessorSpec` (7 tests) covering defaults, application-set keys winning, and the SiteMesh 2 default-layout fallback.
- `:grails-sitemesh3:test` and `:grails-layout:test` pass.
- Verified end-to-end against a previously affected application on 8.0.0-M3: with only the patched `grails-sitemesh3` jar swapped in, both a `(view:)` URL mapping and a controller-rendered page produce fully decorated output, byte-identical through the CNVR path and the fall-through path.

### DIFF
--- a/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
+++ b/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
@@ -28,7 +28,10 @@ import grails.plugins.Plugin
 import org.grails.plugins.web.taglib.RenderSitemeshTagLib
 
 /**
- * Configures Grails to use SiteMesh 3 instead of SiteMesh 2.
+ * Provides GSP layout decoration through SiteMesh 3's filter-less,
+ * view-resolver-based integration. Installing it alongside the legacy
+ * SiteMesh 2 module (grails-layout) is supported; in that case the
+ * SiteMesh 2 integration keeps decorating and this one stands down.
  *
  * <p>The heavy lifting lives outside this class: default configuration
  * properties are contributed by {@link Sitemesh3EnvironmentPostProcessor}
@@ -49,7 +52,7 @@ class Sitemesh3GrailsPlugin extends Plugin {
     def title = 'SiteMesh 3'
     def author = 'Scott Murphy'
     def authorEmail = ''
-    def description = 'Configures Grails to use SiteMesh 3 instead of SiteMesh 2'
+    def description = 'Provides GSP layout decoration using SiteMesh 3'
     def profiles = ['web']
 
     def license = 'APACHE'
@@ -66,9 +69,6 @@ class Sitemesh3GrailsPlugin extends Plugin {
     @Override
     BeanRegistrar beanRegistrar() {
         { BeanRegistry registry, Environment environment ->
-            // Unwraps the SiteMesh view for "render template:" partials so
-            // they are never decorated with a layout (the SiteMesh 2 plugin
-            // does the same with its GrailsLayoutRenderViewMutator).
             registry.registerBean('grailsRenderViewMutator', Sitemesh3RenderViewMutator)
         } as BeanRegistrar
     }
--- a/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
+++ b/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
@@ -29,9 +29,9 @@ import org.grails.plugins.web.taglib.RenderSitemeshTagLib
 
 /**
  * Provides GSP layout decoration through SiteMesh 3's filter-less,
- * view-resolver-based integration. Installing it alongside the legacy
- * SiteMesh 2 module (grails-layout) is supported; in that case the
- * SiteMesh 2 integration keeps decorating and this one stands down.
+ * view-resolver-based integration. It is a drop-in replacement for the
+ * grails-layout module — the two must never share a classpath, which
+ * {@link Sitemesh3EnvironmentPostProcessor} enforces at startup.
  *
  * <p>The heavy lifting lives outside this class: default configuration
  * properties are contributed by {@link Sitemesh3EnvironmentPostProcessor}
--- a/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
+++ b/grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
@@ -18,16 +18,23 @@
  */
 package org.grails.plugins.sitemesh3
 
-import org.springframework.core.env.ConfigurableEnvironment
-import org.springframework.core.env.MapPropertySource
-import org.springframework.core.env.PropertySource
+import org.springframework.beans.factory.BeanRegistrar
+import org.springframework.beans.factory.BeanRegistry
+import org.springframework.core.env.Environment
 
-import grails.core.DefaultGrailsApplication
 import grails.plugins.Plugin
-import org.grails.config.PropertySourcesConfig
 import org.grails.plugins.web.taglib.RenderSitemeshTagLib
-import org.grails.web.util.WebUtils
 
+/**
+ * Configures Grails to use SiteMesh 3 instead of SiteMesh 2.
+ *
+ * <p>The heavy lifting lives outside this class: default configuration
+ * properties are contributed by {@link Sitemesh3EnvironmentPostProcessor}
+ * (registered in {@code META-INF/spring.factories}) and the view-resolver
+ * decoration is applied by {@link Sitemesh3AutoConfiguration}. The only bean
+ * this plugin itself contributes is registered through {@link #beanRegistrar()},
+ * the modern replacement for the deprecated {@code doWithSpring()} bean DSL.</p>
+ */
 class Sitemesh3GrailsPlugin extends Plugin {
 
     def grailsVersion = '7.0.0-SNAPSHOT > *'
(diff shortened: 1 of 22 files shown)


codeconsole:
Design refinement after field testing: the definition post-processor now extends upstream's `SiteMeshViewResolverPostProcessor` (the class behind `sitemesh.viewResolver.wrapMode=bean-definition`) and registers under upstream's bean name, so every upstream wrap mode is now preempted through upstream's own `@ConditionalOnMissingBean` contract — `bean-definition` by the definition post-processor, `all`/`bean-instance` by the bean post-processor (now the explicit fallback tier). This also resolves a spurious startup warning ("SiteMesh did not wrap the target ViewResolver bean 'jspViewResolver'") that the upstream base class logs when the instance-level tier wraps nothing, which is the expected state when the definition-level wrap is active.

One deliberate divergence from upstream's `bean-definition` implementation, documented in the class javadoc: upstream re-registers the unwrapped resolver as a separate *named* bean that remains visible to `getBeansOfType(ViewResolver)` sweeps — the same exposure this PR closes — so the Grails implementation embeds the original definition as an anonymous inner-bean definition instead. That may be worth an upstream sitemesh issue in its own right.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3EnvironmentPostProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+ * start, both the SiteMesh starter's {@code @Value} placeholders and the Grails
+ * configuration (which is built from the environment) observe them without any
+ * post-hoc reassignment. This replaces the property-source manipulation the
+ * plugin previously performed in {@code doWithSpring()}.</p>
+ *
+ * <p>Each default is contributed only when the application has not set the key
+ * itself, and the source is appended with lowest precedence, so application
+ * configuration always wins.</p>
+ */
+public class Sitemesh3EnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
```
jdaugherty: Why not use groovy?

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
```diff
@@ -49,39 +66,10 @@ class Sitemesh3GrailsPlugin extends Plugin {
...
-            (grailsApplication as DefaultGrailsApplication).config = new PropertySourcesConfig(propertySources)
-
-            // Unwraps the SiteMesh view for "render template:" partials so
-            // they are never decorated with a layout (the SiteMesh 2 plugin
-            // does the same with its GrailsLayoutRenderViewMutator).
-            grailsRenderViewMutator(Sitemesh3RenderViewMutator)
+    @Override
+    BeanRegistrar beanRegistrar() {
+        return { BeanRegistry registry, Environment environment ->
+            registry.registerBean('grailsRenderViewMutator', Sitemesh3RenderViewMutator)
```
jdaugherty: The class javadoc states that installing this plugin alongside grails-layout "is supported; in that case the SiteMesh 2 integration keeps decorating and this one stands down" — but SiteMesh 2 and SiteMesh 3 are never meant to be installed together (grails-layout only exists when sitemesh3 is missing). The javadoc should say the two are mutually exclusive rather than promise coexistence semantics.

Worth noting because the stand-down promise is no longer fully true anyway: registrar beans are applied after the `doWithSpring()` drain and win name conflicts (see `GrailsEarlyPluginRegistrationPostProcessor.applyBeanRegistrars`), so if both modules ever do end up on the classpath, this `grailsRenderViewMutator` now displaces SM2's `GrailsLayoutRenderViewMutator` (which previously won via plugin order) and SM2 layout views would no longer be unwrapped for `render template:` partials. If mutual exclusion is the contract, consider having the plugin fail fast (or at least warn) when grails-layout is detected, instead of documenting cooperative behavior.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3ViewResolverDefinitionPostProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+            MethodMetadata factoryMethod = annotated.getFactoryMethodMetadata();
+            if (factoryMethod != null) {
+                className = factoryMethod.getReturnTypeName();
+            }
+        }
+        if (className == null) {
+            return false;
+        }
+        try {
+            Class<?> beanClass = ClassUtils.forName(className, getClass().getClassLoader());
```
jdaugherty: `getClass().getClassLoader()` is the classloader that loaded the grails-sitemesh3 jar, so it can miss application-defined classes when app classes live in a child/restart classloader (e.g. devtools dev-mode reloading). A user-defined `SiteMeshViewResolver` subclass registered as `jspViewResolver` would then fail `forName`, be treated as non-decorating, and get double-wrapped.

The check should resolve with the same loader the container will use when it instantiates the definition — the bean factory's bean classloader (`AbstractBeanDefinition.resolveBeanClass` uses it), falling back to the thread context classloader:

```java
ClassLoader loader = registry instanceof ConfigurableBeanFactory cbf
        ? cbf.getBeanClassLoader()
        : ClassUtils.getDefaultClassLoader(); // TCCL with fallbacks
Class<?> beanClass = ClassUtils.forName(className, loader);
```

(the registry passed to `postProcessBeanDefinitionRegistry` is the `DefaultListableBeanFactory` itself in practice, and `getBeanClassLoader()` defaults to `ClassUtils.getDefaultClassLoader()` anyway, so both branches agree unless a custom bean classloader was set). This requires threading the registry into `isAlreadyDecorating`.

jdaugherty (commented):
Have you tested this with sitemesh2?  We still technically support it via grails-layout and the bean wiring now occurs at a different level.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3EnvironmentPostProcessor.java
codeconsole: Deliberate: this module keeps its Spring infrastructure in Java (`CaptureAwareContentProcessor`, `Sitemesh3LayoutFinder`, `GrailsSiteMeshViewResolver` are all Java), these classes have no Groovy idiom to benefit from, and Java puts them under checkstyle. As a bonus it avoids the `grails.util.Environment` / Spring `Environment` name clash that forces fully-qualified references in Groovy.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
codeconsole: Agreed — the coexistence framing was wrong, and as you noted the stand-down promise was already broken by registrar ordering. Fixed in ba2c2ec2: all SiteMesh 2 accommodation logic is removed (including the `ApplicationListener` exclusions in both post-processors, which only existed to detect the SM2 resolver), the javadoc now states the modules are mutually exclusive, and `Sitemesh3EnvironmentPostProcessor` fails startup with a clear message when grails-layout is detected on the classpath. The `grails.views.layout.default` fallback stays — honoring existing config is what makes the replacement drop-in.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3ViewResolverDefinitionPostProcessor.java
codeconsole: Good catch — fixed in ba2c2ec2 exactly as suggested: the check resolves against the bean factory's bean class loader when the registry is a `ConfigurableBeanFactory`, falling back to `ClassUtils.getDefaultClassLoader()`, so app-defined `SiteMeshViewResolver` subclasses in a child/restart class loader are recognized instead of double-wrapped.

codeconsole:
The functional-test failures were caused by the mutual-exclusion fail-fast from ba2c2ec2: it turns out a combined classpath is not a freak misconfiguration but the standard state for SiteMesh 2 applications — grails-sitemesh3 arrives transitively via `grails-dependencies-starter-web`, so declaring `grails-layout` puts both on the classpath (as half a dozen apps in grails-test-examples do, e.g. `spring-dependency-management`).

32bb99760f keeps the mutual-exclusion *contract* but tolerates co-presence the way those applications already relied on: SiteMesh 2 keeps decorating and this module stands down — one `isSiteMesh2Present()` classpath check gating the definition post-processor, the bean post-processor, and the `grailsRenderViewMutator` registrar registration (which would otherwise displace SM2's mutator via registrar name-conflict precedence, the regression flagged earlier). `Sitemesh3EnvironmentPostProcessor` warns loudly with exclusion instructions, and the tolerance is documented as subject to removal.

Escalating to a hard fail-fast would need the in-tree apps fixed (exclude grails-sitemesh3 from the starter) and is really a separate policy decision — happy to prepare that as a follow-up if that's the direction you want.

Verified locally: `:grails-sitemesh3` suite + checkstyle/codenarc, whole-repo compile, and `--rerun-tasks` integration tests for `spring-dependency-management` (the failing app), `mail` (also combined-classpath), `app1`, and `mongodb-base`.

grails-gsp/grails-sitemesh3/src/main/java/org/grails/plugins/sitemesh3/Sitemesh3EnvironmentPostProcessor.java
jdaugherty: The groovy idiom is the code is easier to read because of property references etc

PR #15965, opened 2026-07-10
Remove jodd-wot from fields tests

jamesfredley:
## Summary

Removes the `jodd-wot` test dependency from `grails-fields` by replacing its only usage with direct Spock string assertions against the rendered field output.

## Changes

| Area | Change |
| --- | --- |
| `grails-fields` tests | Removed `jodd.lagarto.dom.jerry.Jerry` imports and helper usage in `DefaultFieldTemplateSpec`. |
| `grails-fields` dependencies | Removed `org.jodd:jodd-wot` from test dependencies. |
| Root properties | Removed the now-unused `joddWotVersion` property. |

## Verification

| Check | Result |
| --- | --- |
| `rg -n "jodd|jodd-wot"` | No matches. |
| `./gradlew.bat :grails-fields:test --tests "grails.plugin.formfields.DefaultFieldTemplateSpec"` | Passed. |
| `./gradlew.bat :grails-fields:test` | Passed. |

## Checklist

- [x] Searched for `jodd-wot` / `jodd` references.
- [x] Replaced test usage with plain Spock assertions.
- [x] Removed unused dependency/version wiring.
- [x] Verified affected module tests pass.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -39,7 +39,6 @@ gradleToolingApiVersion=9.6.0
 greenmailVersion=2.1.2
 javassistVersion=3.30.2-GA
 jnrPosixVersion=3.1.20
-joddWotVersion=3.3.8
 joptSimpleVersion=5.0.4
 jspApiVersion=4.0.0
 openTest4jVersion=1.3.0
--- a/gradle.properties
+++ b/gradle.properties
@@ -40,6 +40,7 @@ greenmailVersion=2.1.2
 javassistVersion=3.30.2-GA
 jnrPosixVersion=3.1.20
 joptSimpleVersion=5.0.4
+jsoupVersion=1.22.1
 jspApiVersion=4.0.0
 openTest4jVersion=1.3.0
 picocliVersion=4.7.6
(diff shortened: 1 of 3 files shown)


jdaugherty:
I'm not a fan of the string comparisons - they're fragile and brittle.  Why are we not using a standard html parser?

jamesfredley:
Fair point on the brittleness. Worth separating the goal from the assertion style:

The primary objective (pre-release review, Codebase 3 / hygiene quick-win) is to **remove the `jodd-wot` dependency** - `jodd-wot 3.3.8` dates to 2012 and is effectively unmaintained, and it's a single `testImplementation` usage in `grails-fields/build.gradle`. Getting it off the dependency list is the win.

On *how* to replace the assertion, the review lists two options: "modern assertions" **or** "Jsoup or equivalent". The trade-off:

- **String assertions** (current PR): removes `jodd-wot` with **zero new dependencies**, but is more brittle as you note.
- **Jsoup**: robust HTML parsing, but re-adds a test dependency to replace the one we're removing.

I'm happy to switch these to Jsoup (`org.jsoup:jsoup`, `testImplementation`) if you'd rather have a real parser and consider the added test dep acceptable - that still achieves the `jodd-wot` removal, just trading one modern dep for one archived one. Which would you prefer: keep it dependency-free with tightened assertions, or move to Jsoup?

jdaugherty:
I'm happy with Jsoup, it's a parser that's well understood.  I think we need to parse these values though, especially with some of the refactoring we've talked about.  Otherwise, the tests are too brittle.

PR #15966, opened 2026-07-10
Register the scaffolding view resolver without the bean DSL

codeconsole:
### Summary

Migrates `ScaffoldingGrailsPlugin` off the deprecated `doWithSpring()` bean DSL (deprecated by #15934) and makes the descriptor fully `@CompileStatic` — the second in-tree plugin on the modern registration API, following the SiteMesh 3 module in #15964. The two PRs are independent; both preserve the same view-resolver pipeline.

### Why not a straight `beanRegistrar()` port

The old DSL definition relied on `bean.parent = 'abstractViewResolver'`, and parent/abstract bean templates are one of the DSL constructs `BeanRegistry` deliberately has no equivalent for (per #15934). Re-declaring the GSP resolver defaults inside scaffolding would duplicate them and drift.

Instead, `beanRegistrar()` registers a small `ScaffoldingViewResolverDefinitionPostProcessor` that contributes the `jspViewResolver` **definition** — same class, same `abstractViewResolver` parent, same `lazyInit` — at order **-2**, which slots into the established post-processor pipeline exactly where the DSL registration used to sit:

| order | post-processor | behaviour |
|---|---|---|
| -2 | `ScaffoldingViewResolverDefinitionPostProcessor` (this PR) | contributes the scaffolding resolver definition unless one exists |
| -1 | `GrailsLayoutViewResolverPostProcessor` (grails-layout / SiteMesh 2) | embeds the existing definition as its inner view resolver |
| 0 | `GroovyPagesPostProcessor` (grails-gsp) | contributes the plain GSP resolver only when no definition exists |
| +10 | `Sitemesh3ViewResolverDefinitionPostProcessor` (#15964) | wraps the definition for SiteMesh 3 decoration |

Behavioural parity notes:

- An existing `jspViewResolver` definition (application `resources.groovy`, another plugin) still wins, matching the old override semantics.
- `grails.scaffolding.enableNamespaceViewDefaults` is now read from the Spring `Environment` rather than the Grails config object — equivalent since the config is built from the environment.
- The reload-enabled computation (`Environment.current` / development-mode metadata) is unchanged, just relocated.

### Testing

- New `ScaffoldingViewResolverDefinitionPostProcessorSpec` covering registration (parent template, lazy-init, property values), the environment-driven `enableNamespaceViewDefaults`, existing-definition precedence, and the pipeline ordering contract.
- `:grails-scaffolding:test` passes.
- Verified end-to-end on a real application with this module and the #15964 module swapped in together (the eventual merged state): scaffolded controller pages and `(view:)` URL mappings both render with layouts applied through the SiteMesh 3 view-resolver chain, and the resolver arrives wrapped as `GrailsSiteMeshViewResolver[inner=ScaffoldingViewResolver]`.

### DIFF
--- a/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessor.java
+++ b/grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessor.java
@@ -16,20 +16,18 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-package grails.plugin.scaffolding
+package grails.plugin.scaffolding;
 
-import groovy.transform.CompileStatic
+import org.springframework.beans.BeansException;
+import org.springframework.beans.MutablePropertyValues;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
+import org.springframework.beans.factory.support.GenericBeanDefinition;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.core.Ordered;
+import org.springframework.core.env.Environment;
 
-import org.springframework.beans.BeansException
-import org.springframework.beans.MutablePropertyValues
-import org.springframework.beans.factory.support.BeanDefinitionRegistry
-import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor
-import org.springframework.beans.factory.support.GenericBeanDefinition
-import org.springframework.context.EnvironmentAware
-import org.springframework.core.Ordered
-import org.springframework.core.env.Environment
-
-import grails.util.Metadata
+import grails.util.Metadata;
 
 /**
  * Registers the {@code jspViewResolver} bean definition as a
@@ -40,8 +38,7 @@ import grails.util.Metadata
  * than building the resolver directly keeps the view-resolver configuration in
  * one place and preserves the established post-processor pipeline.
  */
-@CompileStatic
-class ScaffoldingViewResolverDefinitionPostProcessor implements BeanDefinitionRegistryPostProcessor, EnvironmentAware, Ordered {
+public class ScaffoldingViewResolverDefinitionPostProcessor implements BeanDefinitionRegistryPostProcessor, EnvironmentAware, Ordered {
 
     /**
      * Runs before the SiteMesh 2 module's {@code GrailsLayoutViewResolverPostProcessor}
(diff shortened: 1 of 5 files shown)


grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+    public static final int ORDER = GroovyPagesPostProcessor.ORDER - 2;
+
+    private static final String JSP_VIEW_RESOLVER_BEAN_NAME = "jspViewResolver";
+
+    private Environment environment;
+
+    @Override
+    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) throws BeansException {
+        if (registry.containsBeanDefinition(JSP_VIEW_RESOLVER_BEAN_NAME)) {
+            // an application- or plugin-supplied view resolver wins
```
jdaugherty: The plugin-vs-plugin direction of this precedence actually flips relative to the old DSL registration, so "matching the old override semantics" isn't quite exact. Previously the definition was registered during the `doWithSpring()` drain in plugin load order: scaffolding (`loadAfter = ['groovyPages']`) *overwrote* a `jspViewResolver` registered by any plugin loading before it, and was itself overwritten by plugins loading after it and by the application's `resources.groovy`. Now any definition present after the drain — including one from a plugin that loads *before* scaffolding — suppresses this registration.

Application precedence is preserved, and stand-down-if-present is arguably the better contract (it's what `GroovyPagesPostProcessor` does), so this is probably fine — just confirming the flip is intentional, and the inline comment could say "any existing definition wins" rather than claiming parity with the old override semantics.

grails-scaffolding/src/test/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessorSpec.groovy
```diff
@@ -0,0 +1,85 @@
...
+        when:
+        postProcessor.postProcessBeanDefinitionRegistry(registry)
+
+        then:
+        registry.getBeanDefinition('jspViewResolver').beanClassName == InternalResourceViewResolver.name
+    }
+
+    void "runs before the SiteMesh 2 layout post-processor and the GSP default post-processor"() {
+        expect: "GrailsLayoutViewResolverPostProcessor runs at -1 and GroovyPagesPostProcessor at 0"
+        postProcessor.order < -1
```
jdaugherty: This assertion hardcodes the collaborators' orders (-1 and 0) in a comment rather than checking them, so if `GrailsLayoutViewResolverPostProcessor` or `GroovyPagesPostProcessor` were ever renumbered the test would keep passing while the pipeline contract silently broke. Asserting against the real constants pins the contract:

```groovy
postProcessor.order < new GrailsLayoutViewResolverPostProcessor().order
postProcessor.order < GroovyPagesPostProcessor.ORDER
```

(the second works with the existing `implementation project(':grails-gsp')`; the first needs a `testImplementation project(':grails-layout')` — if that dependency isn't wanted, at least the `GroovyPagesPostProcessor.ORDER` half is free).

jdaugherty (approved):

grails-scaffolding/src/main/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessor.java
codeconsole: Confirmed intentional — stand-down-if-present is the contract, consistent with `GroovyPagesPostProcessor`. You're right that my "matching the old override semantics" claim was only true for application precedence, not plugin-vs-plugin: the old `doWithSpring()` registration overwrote definitions from earlier-loading plugins, and this doesn't. Fixed the inline comment in be023f8e42 to state "any existing definition wins" and to note explicitly that this is stricter than the DSL registration was.

grails-scaffolding/src/test/groovy/grails/plugin/scaffolding/ScaffoldingViewResolverDefinitionPostProcessorSpec.groovy
codeconsole: Done in be023f8e42 — the test now asserts against both real constants (`new GrailsLayoutViewResolverPostProcessor().order` and `GroovyPagesPostProcessor.ORDER`), with `grails-layout` added as a `testImplementation` dependency. That dep felt justified rather than merely tolerable: the -2 slot exists precisely because grails-layout's post-processor embeds this definition at -1, so the contract test naming that collaborator directly is the point.

PR #15967, opened 2026-07-10
Add default configurable Grails security response headers

jamesfredley:
### What was found

| Problem | Impact |
| --- | --- |
| Grails apps do not set common security headers by default | Missing X-Content-Type-Options, X-Frame-Options, Referrer-Policy, etc. |
| Google Doc 2.1 listed framework default security headers as not covered | No existing open PR for this |
| Apps that use Spring Security still benefit from baseline headers when plugin is absent | Framework should provide safe defaults with opt-out |

### What changed

| Area | Change |
| --- | --- |
| Filter | `GrailsSecurityHeadersFilter` (`OncePerRequestFilter`) |
| Auto-config | `GrailsSecurityHeadersAutoConfiguration` with `@ConditionalOnMissingBean` |
| Defaults | `X-Content-Type-Options: nosniff`, `X-Frame-Options: SAMEORIGIN`, `Referrer-Policy: strict-origin-when-cross-origin`, `X-XSS-Protection: 0` |
| HSTS | Only applied on secure requests when configured |
| CSP | Disabled by default; set via config when desired |
| Config | `grails.security.headers.*` enable/disable and per-header values |
| Docs / tests | Security guide, upgrade notes, auto-config specs |

### Review feedback addressed

| Source | Item | Resolution |
| --- | --- | --- |
| Copilot | Apply headers in a `finally` block after the chain | Declined: for security headers that skips redirect/error/streaming responses that commit during the chain. Kept eager before-chain application with `containsHeader` gap-fill (downstream `setHeader` still wins) + added a redirect-committed regression test |
| Review pass | ERROR redispatches lacked headers | Overrode `shouldNotFilterErrorDispatch()` to `false` + `DispatcherType.ERROR` test |

### Out of scope / follow-up

| Topic | Status |
| --- | --- |
| Full CSP policy templates | Follow-up |
| Permissions-Policy defaults | Follow-up |
| Spring Security ordering (eager defaults vs SS skip-if-present writers; needs on-commit wrapper) | Follow-up (SS is an optional plugin) |

### Related MD topics

| Source | Topic |
| --- | --- |
| Google Doc 2.1 | Default framework security headers |
| Status map | Not covered |

### Issue and Scope
- [x] Background explains the security-headers gap.
- [x] Scoped to default response headers only.
- [x] Single focused feature seed.
- [x] Targets `8.0.x`.

### Code Quality
- [x] Tests cover enable/disable and defaults.
- [x] Focused controllers tests intended for CI.
- [x] No mass reformatting.

### Licensing and Attribution
- [x] Apache License 2.0.

### Documentation
- [x] User-facing docs updated.
- [x] PR description explains what changed and why.


grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
```diff
@@ -0,0 +1,60 @@
...
+            throws ServletException, IOException {
+        applyHeader(response, "X-Content-Type-Options", properties.getContentTypeOptions());
+        applyHeader(response, "X-Frame-Options", properties.getFrameOptions());
+        applyHeader(response, "Referrer-Policy", properties.getReferrerPolicy());
+        applyHeader(response, "X-XSS-Protection", properties.getXssProtection());
+        if (request.isSecure()) {
+            applyHeader(response, "Strict-Transport-Security", properties.getHsts());
+        }
+        applyHeader(response, "Content-Security-Policy", properties.getContentSecurityPolicy());
+        filterChain.doFilter(request, response);
```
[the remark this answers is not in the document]
jamesfredley: Thanks - I evaluated moving header application into a `finally` block after the chain, but for **security** headers that regresses a critical case: on responses that commit *during* the chain (`sendRedirect`, `sendError`, `flushBuffer`, streaming), `response.isCommitted()` is already true when the chain returns, so a post-chain block would skip every header - redirects and error pages would ship with no hardening headers.

Instead the filter keeps applying the defaults **eagerly before** the chain, but `applyHeader` only sets a header when it is not already present (`containsHeader`), so a value a downstream component set earlier still wins and an explicit downstream `setHeader` still replaces it. Added a regression test that a redirect-committed response still carries the headers, plus a `DispatcherType.ERROR` test after overriding `shouldNotFilterErrorDispatch()` so ERROR redispatches are covered too. See the commit comment for the follow-up note on Spring Security ordering.

jamesfredley:
## Review feedback addressed

Merged the latest `8.0.x` and pushed follow-up commits.

**Copilot review comment - apply headers in a `finally` block after the chain:** evaluated and intentionally **not** adopted, because for security headers it regresses committed responses. If a downstream `sendRedirect` / `sendError` / `flushBuffer` / streaming write commits the response during the chain, `isCommitted()` is already true on return and a post-chain block would skip **all** hardening headers on exactly the redirect/error/streaming responses that need them. Instead the filter keeps applying the defaults eagerly before the chain with a `containsHeader` gap-fill (so a value set earlier wins, and an explicit downstream `setHeader` still replaces it), and I added:

- a regression test that a **redirect-committed** response still carries the default headers, and
- an override of `shouldNotFilterErrorDispatch()` to `false` plus a `DispatcherType.ERROR` test, so container **ERROR** redispatches (which reset the response) also get the headers.

**Follow-up (documented, not changed here):** when the optional Spring Security plugin is installed, its header writers skip headers already present, so eagerly-set Grails defaults can win over a Spring-Security-configured policy. The robust fix is an on-commit response wrapper that fills only-missing headers after downstream writers run; that is a larger change and Spring Security is an optional plugin, so I left it as a follow-up rather than expand this starter filter's scope.

Local verification: `:grails-controllers:test --tests GrailsSecurityHeadersAutoConfigurationSpec` (9 features) passes.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private static void applyHeader(HttpServletResponse response, String name,
+            GrailsSecurityHeadersProperties.Header header) {
+        if (header != null && header.isEnabled() && StringUtils.hasText(header.getValue()) &&
+                !response.containsHeader(name)) {
```
jdaugherty, line 61: The `containsHeader` guard can't see headers injected by a reverse proxy, because those are added *after* the response leaves the app. In the very common setup where nginx already sends these headers via `add_header` (which appends and never replaces upstream headers), upgrading to 8.0 means the client suddenly receives the header twice — e.g. proxy `X-Frame-Options: DENY` plus app `SAMEORIGIN`. Browsers handle conflicting duplicates inconsistently: duplicate conflicting `X-Frame-Options` causes Chrome to block framing outright, duplicate `Referrer-Policy` resolves to the last valid value, and multiple `Content-Security-Policy` headers are enforced as the intersection of all policies (strictly tighter, can break pages). haproxy `http-response set-header` replaces, so it's unaffected, but nginx `add_header` deployments will double-send by default.

Since this is on by default, existing proxy-managed deployments change behavior silently on upgrade. At minimum the security guide and the 8.0 upgrade notes should call this out explicitly with the mitigation (set `grails.security.headers.enabled: false` — or per-header `enabled: false` — when the edge proxy owns these headers, or strip the app's copies at the proxy). Worth also considering whether on-by-default is the right trade-off for an upgrade release given this.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        applyHeader(response, "X-Content-Type-Options", properties.getContentTypeOptions());
+        applyHeader(response, "X-Frame-Options", properties.getFrameOptions());
+        applyHeader(response, "Referrer-Policy", properties.getReferrerPolicy());
+        applyHeader(response, "X-XSS-Protection", properties.getXssProtection());
+        if (request.isSecure()) {
```
jdaugherty, line 51: Behind a TLS-terminating proxy (the typical nginx/haproxy topology), the backend connection is plain HTTP, so `request.isSecure()` is `false` and HSTS is silently never sent — even when the user has explicitly set `grails.security.headers.hsts.enabled: true`. It only works if the app is also configured with `server.forward-headers-strategy: framework|native` so `X-Forwarded-Proto` is honored (the `ForwardedHeaderFilter` / `RemoteIpValve` run well before this filter, so ordering is fine once that's set).

The docs currently just say HSTS is sent "only for secure requests" — that should spell out the proxy case: either configure `server.forward-headers-strategy`, or (usually better) set HSTS at the proxy where TLS actually terminates. Otherwise enabling HSTS here is a silent no-op for most real deployments.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -547,6 +547,10 @@ Spring Security is moving toward the fluent `HttpSecurity` configuration API for
 If your application references `DEFAULT_FILTER_ORDER` for custom filter positioning, replace it with the concrete value `-100` (computed as `OrderedFilter.REQUEST_WRAPPER_FILTER_MAX_ORDER - 100`).
 Longer term, consider migrating to the fluent `HttpSecurity` API for filter chain configuration.
 
+Grails 8 also registers a servlet filter that sends default browser hardening headers for servlet web applications.
+Applications that already set these headers through Spring Security or a custom filter keep their existing response values.
```
jdaugherty: I believe this claim is inverted for Spring Security. This filter registers at `GrailsFilters.LAST` (order -110), which runs *before* the Spring Security filter chain (default order -100), and it sets its headers eagerly on the way in. Spring Security's header writers (`XFrameOptionsHeaderWriter`, `HstsHeaderWriter`, `StaticHeadersWriter` — verified in spring-security-web 7.0.4 sources) all skip writing when the header is already present on the response.

So an app that configures e.g. `frameOptions { deny() }` in its security config will still send the Grails default `SAMEORIGIN` after upgrading — the Grails filter wins, not Spring Security. Same for a customized HSTS policy. That's a behavioral regression for existing Spring Security users, and the opposite of what this sentence promises. Options: back off when Spring Security is on the classpath (it already provides secure defaults for these headers), or write the headers at commit time only when still absent (the `HeaderWriterFilter` pattern) so anything set later in the chain wins. If the current precedence is intentional, this doc needs to say the opposite of what it says now.

jdaugherty (commented):
Took a look at this with reverse-proxy deployments (nginx/haproxy) in mind, since that's how most production Grails apps run. A few concerns inline — the common thread is that the `containsHeader` guard only protects against headers set *inside* the servlet container, but in real deployments these headers are frequently owned by the edge proxy or by Spring Security, and in both cases the current behavior is surprising.

jdaugherty (changes requested):

jdaugherty:
The reverse proxy scenarios are real regressions that need addressed.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersAutoConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+ * one would and only writes a header when it is still absent, so an eagerly-applied
+ * Grails default would silently win over an application's explicit Spring Security
+ * header configuration. Spring Security already ships secure header defaults of its
+ * own, so this auto-configuration only fills the gap for applications that don't have
+ * it.</p>
+ */
+@AutoConfiguration
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+@ConditionalOnBooleanProperty(name = "grails.security.headers.enabled", matchIfMissing = true)
+@ConditionalOnMissingClass("org.springframework.security.web.header.HeaderWriterFilter")
```
jdaugherty: This back-off makes things worse for Spring Security users rather than fixing the precedence problem.

The Grails Spring Security Core plugin (`grails-spring-security/plugin` in this repo) never registers a `HeaderWriterFilter`. `SecurityFilterPosition` reserves a `HEADERS_FILTER` slot, but nothing in the plugin fills it, so plugin users get no security headers from Spring Security today. The plugin declares `spring-security-web` as an `api` dependency, which means `HeaderWriterFilter` is always on their classpath and this condition now disables the Grails filter for every one of them. The net result is zero security headers for every Grails Spring Security application, which is the opposite of what the PR description and the upgrade note promise.

Two further problems with a class-presence condition:

- Even in apps that do use the `HttpSecurity` DSL, `HeaderWriterFilter` only runs on requests matched by a `SecurityFilterChain`. Requests outside a `securityMatcher`, or chains with `headers.disable()`, get nothing, and this filter is no longer there to fill the gap.
- The condition sits on the whole auto-configuration, so an explicitly configured value such as `grails.security.headers.content-security-policy.value` silently does nothing whenever `spring-security-web` is present. Explicit config that no-ops without any signal is a defect on its own.

The precedence problem is real, but the fix belongs in the filter (write on commit so later writers win; see the comment there), not in a classpath condition. Please remove this.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    protected boolean shouldNotFilterErrorDispatch() {
+        return false;
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        applyHeader(response, "X-Content-Type-Options", properties.getContentTypeOptions());
```
jdaugherty: Applying eagerly before the chain is the root cause of the Spring Security precedence problem, and the objection to the `finally`-block suggestion (redirects / `sendError` / streaming commit during the chain) has a standard answer: write the headers at commit time.

Wrap the response in a small `HttpServletResponseWrapper` that intercepts the commit points (`sendError`, `sendRedirect`, `flushBuffer`, `getWriter().flush()/close()`, `getOutputStream().flush()/close()`, and content-length reached) and fills any still-missing headers immediately before delegating. If the chain returns without committing, fill them then. This is exactly what Spring Security's `HeaderWriterFilter` does with `OnCommittedResponseWrapper`, and when both filters are active the nesting works out naturally: the later filter's wrapper fires first and writes its configured values, then this wrapper fills only what is still absent. Anything set by a controller, an interceptor, or Spring Security wins; Grails only supplies defaults for headers nobody else set.

With that in place the `@ConditionalOnMissingClass` back-off in the auto-configuration is unnecessary and should go. The existing redirect-committed and `DispatcherType.ERROR` tests should keep passing, and there should be a new test that a downstream filter setting `X-Frame-Options: DENY` after this filter runs ends up with `DENY`, not `SAMEORIGIN`.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
jdaugherty: Still a silent no-op behind a TLS-terminating proxy unless `server.forward-headers-strategy` happens to be configured. Documenting the workaround is not a fix for an explicitly enabled setting that does nothing.

When `request.isSecure()` is false, fall back to the forwarded scheme: `X-Forwarded-Proto: https` or RFC 7239 `Forwarded: proto=https`. Trust is not a concern for this particular header: RFC 6797 section 8.1 requires user agents to ignore `Strict-Transport-Security` received over a non-secure transport, so a client spoofing a forwarded header on a plain-HTTP direct connection gains nothing. There is already precedent for reading `X-Forwarded-Proto` directly in `GrailsWebRequest.getBaseUrl`.

When `ForwardedHeaderFilter` or `RemoteIpValve` is configured, `isSecure()` is already correct and the fallback is never consulted, so the two approaches compose.

grails-doc/src/en/guide/security.adoc
```diff
@@ -31,3 +31,65 @@ Grails has a few built in safety mechanisms by default.
...
+            content-security-policy:
+                enabled: true
+                value: default-src 'self'
+----
+
+Set `grails.security.headers.<header>.enabled` to `false` to omit a single header.
+For example, `grails.security.headers.xss-protection.enabled: false` omits `X-XSS-Protection`.
+If an application or another filter has already set one of these headers on the response, Grails leaves that value unchanged.
+
+This "already set" check can only see headers set inside the servlet container - it has no visibility into headers a reverse proxy adds to the response after it leaves the application.
```
jdaugherty: The duplication problem needs a behavioral fix, not a paragraph telling users to turn the feature off after they discover it. This is on by default, so a proxy-managed deployment changes what the browser receives the moment it upgrades, and nothing in the app logs or tests will tell them.

The application cannot see headers the proxy appends on the way out, but it can detect that the request came through a proxy. Signals available per request: `Forwarded`, `X-Forwarded-For`, `X-Forwarded-Proto`, `X-Forwarded-Host`, `Via`, `X-Real-IP`. Signals available from configuration: `server.forward-headers-strategy` set to `framework` or `native`, or an active Spring Boot `CloudPlatform` (Boot itself uses that signal to turn forwarded-header handling on).

Suggested behavior, as a tri-state setting defaulting to `auto`: when a proxy is detected, apply only the headers the application configured explicitly and skip the implicit defaults, logging once so operators can see why. Direct-exposed apps, the ones that most need these defaults, keep them. Proxied apps send exactly what they sent on 7.x unless they opt in with `always`, so there is no upgrade regression in either direction. `never` skips detection entirely for apps that know the proxy does not touch these headers.

One caveat worth writing into the design: a client can add a forwarded header to its own direct request and suppress the defaults for that one response. That is not a practical bypass for clickjacking or MIME sniffing, since the attacker does not control the victim browser's request headers, but it should be stated.

jdaugherty (dismissed):
Re-reviewed after the latest push. The three reverse-proxy / Spring Security findings from the previous round are still open: two of them were addressed with documentation only, and the one code change (backing off on `HeaderWriterFilter` class presence) removes headers from the largest group of users this feature is meant to protect. Details inline. The short version of what I think the fix needs to be:

1. Write the headers at commit time via a response wrapper (the same pattern Spring Security's `HeaderWriterFilter` uses) instead of eagerly before the chain, and drop the `@ConditionalOnMissingClass` back-off. That lets any later writer win and still covers redirects, `sendError`, and streaming responses.
2. Make HSTS forwarded-proto aware when `request.isSecure()` is false.
3. Detect that a request arrived through a reverse proxy and, by default, apply only explicitly configured headers in that case, so proxy-managed deployments do not double-send on upgrade.

Documentation alone is not a fix for on-by-default behavior that changes what existing deployments send to browsers.

jdaugherty:
I'm going to make  pass at fixing this PR

jdaugherty:
@jamesfredley I've adjusted this PR so it supports reverse proxies the way Spring does.  Can you please take a look so I know if we can merge?

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/SecurityHeadersResponseWrapper.java
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+    }
+
+    private void trackWritten(long count) {
+        if (this.fired) {
+            return;
+        }
+        this.contentWritten += count;
+        if (this.contentLength >= 0 && this.contentWritten >= this.contentLength) {
```
codeconsole: This misses the most common commit path: the container's response buffer filling up. With no `Content-Length` and no explicit flush, Tomcat commits as soon as the body outgrows `getBufferSize()` (8 KB by default). The filter's `finally` then calls `setHeader` on a committed response, and Tomcat silently ignores it.

I put the filter in front of a servlet on embedded Tomcat 11.0.26:

```
path            body                   X-Content-Type-Options  X-Frame-Options
/small-stream   100 B, OutputStream    nosniff                 SAMEORIGIN
/large-stream   50 KB, OutputStream    <missing>               <missing>
/large-writer   50 KB, getWriter()     <missing>               <missing>
```

So large JSON responses, file downloads, and any big page not buffered by SiteMesh go out with no hardening headers. That contradicts the "streamed body" sentence in `security.adoc`. `OnCommittedResponseWrapper` checks the buffer size next to the content length for exactly this reason. Adding the same check fixed both large cases in the same run:

```suggestion
        int bufferSize = getBufferSize();
        if ((this.contentLength >= 0 && this.contentWritten >= this.contentLength)
                || (bufferSize > 0 && this.contentWritten >= bufferSize)) {
```

The spec can't catch this because `MockHttpServletResponse` still accepts `setHeader` after it commits. A variant of the streaming test would: set `response.bufferSize = 16`, write 64 bytes with no flush, and assert the headers are already present inside the chain.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/SecurityHeadersResponseWrapper.java
```diff
@@ -0,0 +1,249 @@
...
+            super.write(buf, off, len);
+        }
+
+        @Override
+        public void write(String s, int off, int len) {
+            trackWritten(len);
+            super.write(s, off, len);
+        }
+
+        @Override
```
codeconsole: `PrintWriter.println()` writes the line separator straight to the wrapped `out` through its private `newLine()`, bypassing these overrides, so line separators never count toward `Content-Length`. In the same Tomcat run, `setContentLength(3)` + `writer.println('hi')` counted 2 of 3. Tomcat treats the response as committed once the declared length has been written, so the headers written in `finally` were dropped. Counting the separator fixes it (verified):

```suggestion
        @Override
        public void println() {
            trackWritten(System.lineSeparator().length());
            super.println();
        }

        @Override
```

A `println` variant of the content-length test would cover it.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersAutoConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+    @Bean
+    @ConditionalOnMissingBean(name = "grailsSecurityHeadersFilter")
+    public FilterRegistrationBean<GrailsSecurityHeadersFilter> grailsSecurityHeadersFilter(
+            GrailsSecurityHeadersFilter securityHeadersFilter) {
+        FilterRegistrationBean<GrailsSecurityHeadersFilter> registrationBean = new FilterRegistrationBean<>();
+        registrationBean.setFilter(securityHeadersFilter);
+        registrationBean.setDispatcherTypes(EnumSet.of(DispatcherType.REQUEST, DispatcherType.FORWARD,
+                DispatcherType.INCLUDE, DispatcherType.ERROR));
+        registrationBean.addUrlPatterns("/*");
+        registrationBean.setOrder(GrailsFilters.LAST.getOrder());
```
codeconsole: `LAST` (-110) nests this filter inside every filter registered earlier, so an earlier filter that writes the response itself without calling `chain.doFilter` never reaches it. Asset-pipeline does exactly that: `AssetPipelineGrailsPlugin` registers `AssetPipelineFilter` at `GrailsFilters.ASSET_PIPELINE_FILTER` (-190), and on a hit it streams the file and calls `flushBuffer()` without continuing the chain. So `/assets/**` JS and CSS, where `nosniff` matters most, never get these headers.

With commit-time writing, the outermost slot is also the one that makes precedence work. Nested wrappers fire innermost-first, both on commit and when the chain returns, so the outermost filter fills last and everything inside it wins. At -110 that is inverted for anything registered before it. For example, with `spring.security.filter.order: -170` (security ahead of SiteMesh) and `frameOptions { deny() }`, this wrapper fires first and writes `SAMEORIGIN`, then `XFrameOptionsHeaderWriter` skips because the header is already present. That breaks the "regardless of filter ordering" promise in `upgrading80x.adoc`. Registering at `GrailsFilters.FIRST` (-200) or earlier would fix both cases.

codeconsole (commented):
Checked the commit-time wrapper against a real container (the filter in front of a servlet on embedded Tomcat 11.0.26). Two commit paths still drop the headers, and at its current order the filter never sees asset-pipeline responses. Details inline; the two wrapper suggestions were verified in that run.

Two things outside the diff:
- `grails-doc/src/en/guide/introduction/whatsNew.adoc` needs an entry. This changes the headers every servlet app sends by default, so it should appear there as well as in the upgrade notes.
- The new public classes (`GrailsSecurityHeadersFilter`, `GrailsSecurityHeadersProperties`, `GrailsSecurityHeadersAutoConfiguration`) need `@since 8.0` in their class Javadoc, as `GrailsHiddenHttpMethodFilterAutoConfiguration` has.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersProperties.java
```diff
@@ -0,0 +1,173 @@
...
+package org.grails.plugins.web.controllers;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "grails.security.headers")
+public class GrailsSecurityHeadersProperties {
+
+    private boolean enabled = true;
+
+    private Defaults defaults = Defaults.AUTO;
```
codeconsole: With `AUTO` as the default, most production deployments get none of the default headers. Every request through a load balancer or ingress carries `X-Forwarded-For`/`X-Real-IP`, and every Kubernetes pod has `KUBERNETES_SERVICE_HOST`/`KUBERNETES_SERVICE_PORT` set, which makes Boot report `CloudPlatform.KUBERNETES`. I ran the auto-configured filter (`securityHeadersFilter(properties, new StandardEnvironment())`) on embedded Tomcat 11.0.26:

```
environment                request                        nosniff / X-Frame-Options / Referrer-Policy
plain host                 direct                         sent
plain host                 X-Forwarded-For: 203.0.113.7   <missing>
KUBERNETES_SERVICE_* set   direct                         <missing>
KUBERNETES_SERVICE_* set   X-Forwarded-For: 203.0.113.7   <missing>
```

The headers only appear on requests that hit the app directly, which in practice means local dev and tests. The production traffic they're meant to protect doesn't get them, and the only trace is one INFO line. `AUTO` assumes the proxy supplies these headers, but ingress-nginx, AWS ALB, the GKE ingress and Heroku's router add none of them unless someone configures them to.

The risk `AUTO` guards against, a proxy that already sends one of these headers, is small for these four defaults:
- `X-Content-Type-Options`: browsers only look at the first value, and `nosniff` twice is still `nosniff`.
- `X-Frame-Options`: identical values collapse to one. Conflicting values (proxy `DENY`, app `SAMEORIGIN`) make the browser block framing, which is the stricter outcome the proxy asked for.
- `Referrer-Policy`: the last valid token wins, and nginx's `add_header` goes after the upstream headers, so the proxy's policy still applies.
- `X-XSS-Protection`: no current browser implements the XSS auditor, so this header does nothing either way.

CSP is the header where duplicates really hurt (multiple policies are intersected), and it isn't a default. Explicitly configured headers are sent whether or not a proxy is detected, so `AUTO` doesn't prevent that case anyway. I'd default to `ALWAYS` and keep `AUTO` as the opt-in for deployments whose proxy owns these headers, which the upgrade note's nginx `add_header` guidance already describes.

```suggestion
    private Defaults defaults = Defaults.ALWAYS;
```

The metadata JSON default, the `security.adoc` table, the upgrade note and the defaults specs would need to follow.

matrei:
Reviewed at head `5c8fc21206` (base `8.0.x`, merge-base `7ea4cb9d8e`; merges cleanly into current `8.0.x` at `a441334507`).

What I ran locally:
- `:grails-controllers:cleanTest :grails-controllers:test --no-build-cache`: 182 tests, 0 failures (84 in `GrailsSecurityHeadersAutoConfigurationSpec`, 9 in `GrailsSecurityHeadersFilterTomcatSpec`).
- `:grails-controllers:codeStyle`: clean.
- Throwaway specs (not committed) that put the filter in front of servlets on the embedded Tomcat from the test classpath, to try commit paths and request shapes the PR's specs don't cover. Results are below.

The commit-time wrapper, the move to `GrailsFilters.FIRST` and the `always` default fix the reverse-proxy, Spring Security precedence and asset-pipeline problems raised in the earlier reviews by @jdaugherty and @codeconsole. The latest commit adds re-arming on `reset()`, the typed-registration back-off and the blank-value warning, which all look right. Three things are left before merge: the switch to `ForwardedHeaderUtils` turns some requests into 500s when HSTS is enabled, one commit path still drops the headers, and the PR description is stale. The rest are small.

### 1. With HSTS enabled, URLs Tomcat accepts but `java.net.URI` rejects now fail with 500

`GrailsSecurityHeadersFilter.java:185-191`: `isSecure` now builds a `ServletServerHttpRequest` and calls `getURI()` so it can hand the request to `ForwardedHeaderUtils`. `getURI()` throws `IllegalStateException` ("Could not resolve HttpServletRequest as URI") when the request URL isn't a valid `java.net.URI`, and only `IllegalArgumentException` is caught. Tomcat lets such paths through when `relaxedPathChars` is configured (`server.tomcat.relaxed-path-chars`), which apps set for legacy links containing `|`, `{}` or `[]`. I ran it on Tomcat with `relaxedPathChars='|{}[]^`'` and `hsts.enabled: true`:

```
request                                      status   nosniff   HSTS
GET /plain                                   200      sent      -
GET /p|x                                     500      -         -
GET /p{x}                                    500      -         -
GET /plain?a=|b                              200      sent      -     (getURI re-encodes a bad query)
GET /p|x         X-Forwarded-Proto: https    500      -         -
GET /plain       X-Forwarded-Proto: https    200      sent      sent
```

With HSTS disabled (the default), `isSecure` isn't called and these requests return 200, so the failure only shows up after someone turns HSTS on. The previous hand-written parser didn't touch the URI. Only the scheme is needed here, so it can be resolved without the request URL at all, for example:

```java
HttpHeaders headers = new ServletServerHttpRequest(request).getHeaders();
String scheme = ForwardedHeaderUtils.adaptFromForwardedHeaders(URI.create("http://localhost"), headers)
        .build().getScheme();
```

Catching `IllegalStateException` as well would also stop the 500s, but it would silently skip HSTS on those requests. A spec row with a relaxed path char and HSTS enabled would cover it.

### 2. Multi-byte text written through the writer can commit before the wrapper fires

`SecurityHeadersResponseWrapper.java:229`: the writer counts characters, but the buffer it is compared against (`getBufferSize()`) is in bytes. On Tomcat the default 8 KB char buffer usually hides this. Once the response buffer is larger than that, UTF-8 text with 3-byte characters fills the byte buffer, and commits, at about a third of the character count the wrapper is waiting for:

```
servlet                                                       nosniff / X-Frame-Options
setBufferSize(32768), 400 × write('日' * 50)   (60 KB)        <missing> / <missing>   (every run, at both heads)
setBufferSize(32768), 400 × write('ж' * 50)    (40 KB)        sent
setBufferSize(32768), 1000 × write('a' * 50)   (50 KB)        sent
default buffer,       400 × write('日' * 50)                  sent (4 of 4 runs)
```

It also leaked from one request to the next. In my first run, a plain default-buffer `/cjk-chunks` request lost its headers because it came right after a request on the same keep-alive connection that had called `setBufferSize(32768)`. Tomcat had kept the enlarged buffer on the recycled response. It happened again at `5c8fc21206`. So an app doesn't have to call `setBufferSize` on the affected request itself.

The class Javadoc (`SecurityHeadersResponseWrapper.java:43-47`) says a multi-byte encoding reaches the buffer "a little before the count does" and that "the callback still runs no later than the chain returning". It does run then, but the response has already committed, so the `setHeader` calls are dropped without any error. The comment describes this as safe when it isn't.

Spring Security's `OnCommittedResponseWrapper` makes the same approximation, so this isn't a regression compared with Spring Security. Still, CJK sites are exactly the users who would hit it. Options:
- Count writer output conservatively as `chars * charsetEncoder.maxBytesPerChar()` for the response's character encoding. This fires early rather than late. The trade-off is that an inner skip-if-present writer that counts characters (Spring Security's `HeaderWriterFilter`) may fire after this one on large multi-byte bodies and lose precedence. Given the choice, sending the default is better than sending nothing.
- At minimum, correct the Javadoc and document the limitation in `security.adoc` next to the "body large enough to fill the container's response buffer" sentence.

The latest commit caches the buffer size and refreshes it on `setBufferSize`/`reset`/`resetBuffer`, which doesn't change this: the count is still in characters.

Either way, a Tomcat spec case with `setBufferSize(32768)` and chunked 3-byte characters would pin it down.

### 3. The PR description no longer matches the code

The squash commit will most likely be built from the description, and it currently says:
- "Kept eager before-chain application with `containsHeader` gap-fill" / Copilot `finally` suggestion "Declined". The filter now writes at commit time.
- "Spring Security ordering (… needs on-commit wrapper) — Follow-up". That is done in this PR.
- "`GrailsSecurityHeadersAutoConfiguration` with `@ConditionalOnMissingBean`". There is now also a custom condition that backs off on a typed `FilterRegistrationBean<GrailsSecurityHeadersFilter>`.
- Nothing about `grails.security.headers.defaults` (`always`/`auto`/`never`), forwarded-scheme HSTS, `GrailsFilters.FIRST`, or the Tomcat spec.
- References to "Google Doc 2.1" and a "Status map", which readers of the public PR can't see.

Please rewrite it to describe the final behaviour before merging.

### 4. The upgrade note is under the wrong heading

`upgrading80x.adoc:500`: the note is a paragraph inside "13. Spring Security Filter Ordering". This change alters the response headers of every servlet application, not just Spring Security users, so someone skipping the Spring Security section will miss it. It should be its own numbered section (for example "14. Default Security Response Headers", renumbering the ones after it).

### 5. The "explicitly configured is always sent" wording is too broad

`security.adoc:97`: "Any header you configure explicitly (setting either its `enabled` or its `value` key) is always sent". That isn't true for:
- `content-security-policy.value` or `hsts.value` set on its own. Both keep `enabled: false`, so nothing is sent.
- `<header>.enabled: false`, which also counts as explicit.

Suggest wording along the lines of "Explicitly configured headers are not affected by `defaults`: a header you have enabled and given a value is sent whether or not a proxy is detected." It would also help to say directly that CSP and HSTS need `enabled: true` as well as a value. (The latest commit's blank-value warning covers the `enabled: true` without a value case, and the new sentence in `security.adoc` explains it. The "always sent" sentence itself is unchanged.)

### 6. Minor

- The new ordering paragraph says Spring Boot's "character-encoding" filter runs outside the Grails filter (`security.adoc:65`, `upgrading80x.adoc:503`, `GrailsSecurityHeadersAutoConfiguration.java:61`). In a Grails app it doesn't: `ControllersAutoConfiguration` registers its own `CharacterEncodingFilter` at `GrailsFilters.CHARACTER_ENCODING_FILTER` (-180), which is inside `FIRST`, and Boot's `HttpEncodingAutoConfiguration` backs off. Boot's `ErrorPageFilter` is only registered for WAR deployments, which is worth saying too. Only the forwarded-header filter is outside in a default Grails app.
- `GrailsSecurityHeadersFilter.java:71`: `REVERSE_PROXY_REQUEST_HEADERS` is `public static final`, which adds to the public API a list that only the filter and the spec's `where:` block use. Package-private would do, and it avoids having to keep the list stable.
- `GrailsSecurityHeadersFilterTomcatSpec.groovy:72,75`: the two temp directories are never deleted. Spock's `@TempDir` on a `@Shared` field would clean them up.
- Optional: `GrailsSecurityHeadersFilterTomcatSpec` could implement `HttpClientSupport` from `grails-testing-support-http-client` instead of its own `get()` helper and `firstValue(...).present` checks. `assertHeaders(..., status)` would then check all four default values and the status in one call; today only two of the four values are checked. This needs `testImplementation project(':grails-testing-support-http-client')`, which doesn't create a cycle (`grails-gsp` already uses the testing-support modules in its tests). There's no Spring context to inject `local.server.port`, so the spec overrides `getHttpBaseUrl()`. The shared client follows redirects, so the `/redirect` row needs a client that doesn't, passed through `http(path, client)`. I tried this locally on the previous head (`37b983c60c`): all 8 features pass and `codeStyle` is clean. The `/reset` row added since then carries over with the same `referrerPolicy` default. The unused `DEFAULT_HEADER_NAMES` and the `get()` helper go away, and the `java.net.http.HttpRequest`/`HttpResponse` imports are no longer needed:

  ```groovy
  class GrailsSecurityHeadersFilterTomcatSpec extends Specification implements HttpClientSupport {

      @Shared
      HttpClient noRedirects = HttpClient.newBuilder().followRedirects(HttpClient.Redirect.NEVER).build()

      @Override
      String getHttpBaseUrl() {
          "http://localhost:${port}"
      }

      @Unroll
      void 'headers are on the wire for #path (#description)'() {
          when:
          def response = http(path, noRedirects)

          then:
          response.assertHeaders([
                  'X-Content-Type-Options': 'nosniff',
                  'X-Frame-Options'       : 'SAMEORIGIN',
                  'Referrer-Policy'       : referrerPolicy,
                  'X-XSS-Protection'      : '0'
          ], status)
          bodyLength == null || response.body().length() == bodyLength

          where:
          path            | description                                                | status | bodyLength                | referrerPolicy
          '/small-stream' | 'body below the response buffer'                           | 200    | 100                       | 'strict-origin-when-cross-origin'
          '/large-stream' | 'output stream body outgrowing the buffer, no flush'       | 200    | 50 * 1024                 | 'strict-origin-when-cross-origin'
          '/large-writer' | 'writer body outgrowing the buffer, no flush'              | 200    | 50 * 1024                 | 'strict-origin-when-cross-origin'
          '/println'      | 'Content-Length reached by println'                        | 200    | 2 + LINE_SEPARATOR_LENGTH | 'no-referrer'
          '/redirect'     | 'redirect committed inside the chain'                      | 302    | null                      | 'strict-origin-when-cross-origin'
          '/error'        | 'sendError committed inside the chain'                     | 404    | null                      | 'strict-origin-when-cross-origin'
          '/reset'        | 'reset after the buffer filled but before it flushed'      | 500    | 5                         | 'strict-origin-when-cross-origin'
          '/assets/a.js'  | 'served by an inner filter that never continues the chain' | 200    | 14                        | 'strict-origin-when-cross-origin'
      }

      void 'a header set by the servlet wins over the Grails default'() {
          expect:
          http('/println').hasHeaderValue('Referrer-Policy', 'no-referrer')
      }
  ```

  The `referrerPolicy` column is needed because the exact-value check catches something the current `present` check misses: the `/println` servlet sets `Referrer-Policy: no-referrer` itself, so that row can't expect the default. `body()` returns a String, so its length only equals the byte count for ASCII bodies. A multi-byte case for item 2 would need to compare byte length.

### Verified as correct

- Commit-time writing: the wrapper intercepts `sendError`/`sendRedirect` (all four overloads), `flushBuffer`, writer and stream `flush`/`close`, declared `Content-Length` (including via `setHeader`/`addHeader`/`setIntHeader`/`addIntHeader` and when declared after the body), buffer-full, and the `println` line separator. The Tomcat spec shows each of these on the wire.
- The Servlet 6.1 `ServletOutputStream.write(ByteBuffer)` default goes through the wrapper's `write(byte[])`, so it is counted. Checked on Tomcat for a 100 B and a 50 KB `ByteBuffer`: headers sent in both cases.
- `println` is not double-counted. `PrintWriter.print`/`append`/`format` all funnel through the overridden `write` methods, and only `newLine()` bypasses them.
- Ordering: at `FIRST` (-200) the filter is outside the asset-pipeline filter (-190), the character encoding filter, SiteMesh, and Spring Security at its default -100. Innermost-first firing means every one of them wins over the Grails defaults, and the spec covers Spring Security's `HeaderWriterFilter` on both sides.
- The `@ConditionalOnMissingClass` back-off is gone. With `spring-security-web` on the classpath the filter stays active, which fixes the problem that the Grails Spring Security plugin never registers a `HeaderWriterFilter`.
- HSTS behind a TLS-terminating proxy: apart from item 1, reading the scheme through `ForwardedHeaderUtils` matches `ForwardedHeaderFilter` (first `Forwarded` element, `X-Forwarded-Proto`, `X-Forwarded-Ssl`). A forwarded host with a space doesn't break it, and the scheme is only resolved when HSTS is enabled. `ForwardedHeaderFilter` and `RemoteIpValve` run earlier, so they compose with it.
- The new Spring Security ordering docs are right about `X-Frame-Options`: in spring-security-web 7.1.1, `XFrameOptionsHeaderWriter` in `DENY`/`SAMEORIGIN` mode calls `setHeader` unconditionally, while the `Referrer-Policy`, HSTS and static writers (including `X-Content-Type-Options`) skip a header that is already present.
- `reset()` re-arms the callback and `resetBuffer()` restarts the count, and the Tomcat `/reset` row shows the replacement response gets the headers. `super.reset()` throws on a committed response before any state is cleared.
- The custom `OnMissingSecurityHeadersFilterRegistration` condition matches only on the declared generic type without eager init, and the specs cover another name, a registration that builds the filter itself, and registrations for other filters.
- `defaults: always` is the default, and `auto`/`never` behave as documented, including the config-level signals (`server.forward-headers-strategy`, active `CloudPlatform`) and the log line emitted once.
- `explicit` tracking works through Boot binding: nested `Header` instances are bound in place through the getters, so only keys that are actually set flip the flag.
- `ERROR` dispatches are filtered (`shouldNotFilterErrorDispatch() == false`). Spring Boot always registers an error page, so an exception that resets the response still gets headers on the error dispatch.
- The `whatsNew.adoc` entry and `@since 8.0` on the three public classes that @codeconsole asked for are in. There are no `@author` tags, Apache headers are present, and the Spring Security and Tomcat test dependencies are test-only and BOM-managed.
- CI was still running at the time of review (24 passed, 50 pending). The one failure, "Update Release Draft", is the release-drafter job and doesn't run the PR's code.

jdaugherty:
@matrei thanks, both behaviour findings reproduced on embedded Tomcat and are fixed in 8da7b05c58.

1. `isSecure` no longer touches the request URL. It applies the forwarded headers to the request's scheme alone, so `/p|x` with `relaxedPathChars` and HSTS enabled returns 200 with HSTS (when forwarded as https) instead of 500. Added a spec over `|`, `{}`, `[]`, `^` and `` ` `` paths, plus one pinning that a container scheme of `https` still counts when `isSecure()` is false.
2. Writer output is now counted in encoded bytes: exactly for UTF-8 and single-byte encodings, and at `maxBytesPerChar` for any other. The Tomcat spec has a `setBufferSize(32768)` + 400 × `'日' * 50` row, and unit specs cover 2-, 3- and 4-byte UTF-8 through `write(String)`, `write(char[])` and `write(int)`, `Shift_JIS`, and a multi-byte body reaching `Content-Length`. The enlarged buffer also carried over to a request on a *new* connection in my run, not only the same keep-alive one, since Tomcat recycles the processor along with its buffer. Byte counting covers that too, because the wrapper reads the real buffer size. The Javadoc now describes the trade-off you pointed out: for a large multi-byte body, Grails can fire before Spring Security's character-counting wrapper. The Spring Security ordering section of `security.adoc` says so as well.
3. The PR description still needs rewriting; that is not part of this push.
4. The upgrade note is now its own section, 73 (appended rather than renumbered, because the guide cross-references sections by number, e.g. "section 53").
5. Reworded: `defaults` doesn't apply to explicitly configured headers; an explicit header is sent if it is enabled and has a value; `enabled: false` is never sent; CSP and HSTS need `enabled: true` as well as a value.
6. Character-encoding claim corrected in all three places (Grails' own filter at -180 runs inside; only the forwarded-header filter, and `ErrorPageFilter` in a WAR, are outside). `REVERSE_PROXY_REQUEST_HEADERS` is package-private and the spec lists the headers literally. The Tomcat spec uses `@TempDir` and now checks the exact value of all four headers, with a `referrerPolicy` column for the `/println` row. I kept its own client rather than `HttpClientSupport` for now, because the new multi-byte row compares the raw byte length.

matrei:
Re-reviewed at head `8da7b05c58` (base `8.0.x`, merge-base `7ea4cb9d8e`; merges cleanly into current `8.0.x` at `a441334507`). This round only looks at what changed since `5c8fc21206`.

What I ran locally:
- `:grails-controllers:cleanTest :grails-controllers:test --no-build-cache`: 199 tests, 0 failures (100 in `GrailsSecurityHeadersAutoConfigurationSpec`, 10 in `GrailsSecurityHeadersFilterTomcatSpec`).
- `:grails-controllers:codeStyle`: clean.
- My throwaway Tomcat specs from the last round, re-run at the new head with more cases added (results below).

Everything from the last round is addressed, and I have nothing blocking. Both behaviour problems are fixed on the wire. The PR description now matches the code. One optional nit is below.

### Re-run of the previous findings

Relaxed path chars (`relaxedPathChars='|{}[]^`'`) with `hsts.enabled: true`:

```
request                                                  status   nosniff   HSTS
GET /p|x                                                 200      sent      -       (was 500)
GET /p{x}                                                200      sent      -       (was 500)
GET /p|x      X-Forwarded-Proto: https                   200      sent      sent    (was 500)
GET /p|x      X-Forwarded-Proto: http                    200      sent      -
GET /plain    X-Forwarded-Host: [::1, Proto: https       200      sent      -       (malformed host: IAE caught, HSTS skipped)
GET /plain    X-Forwarded-Port: abc, Proto: https        200      sent      -
GET /plain    X-Forwarded-Proto: https:x / "ht tps"      200      sent      -
GET /plain    Content-Type: text/plain;charset=bogus     200      sent      sent
```

Writer output with `setBufferSize(32768)`, every row showing `nosniff` / `X-Frame-Options` on the wire (all were checked at least once; the two CJK rows twice each, interleaved so the enlarged buffer is recycled into a default-buffer request):

```
400 × write('日' * 50)               UTF-8        60 000 B    sent   (was <missing>)
400 × write('日' * 50), default buf  UTF-8        60 000 B    sent   (was <missing> after a 32 KB request)
400 × write('ж' * 50)                UTF-8        40 000 B    sent
400 × write('😀' * 25)               UTF-8        40 000 B    sent   (surrogate pairs)
20000 × write((int) '日')            UTF-8        60 000 B    sent
400 × printf('%s', '日' * 50)        UTF-8        60 000 B    sent
400 × write('日' * 50)               Shift_JIS    40 000 B    sent
400 × write('日' * 50)               UTF-16       40 002 B    sent
1000 × write('é' * 50)               ISO-8859-1   50 000 B    sent
```

### Nit (optional)

`GrailsSecurityHeadersFilter.java:196-197`: the `catch` comment says "A forwarded port that is not a number", but it also catches `ForwardedHeaderUtils` rejecting a malformed forwarded host ("Invalid IPv4 address", the `[::1` row above) and `URI.create` rejecting the scheme. "Malformed forwarded headers: no trustworthy scheme to go on" would cover all three.

### Verified as correct

- `isSecure` builds the base URI from `request.getScheme()` alone, so the request path and query never reach `java.net.URI`. `URI.create` is inside the `try`, and a container scheme of `https` on a connection that isn't marked secure still counts. This is covered by the new spec.
- Writer byte counting: `utf8Length` counts 1, 2 or 3 bytes, and 2 for each half of a surrogate pair, so 4 for the pair. Any other encoding counts `ceil(maxBytesPerChar)` per character, which can only over-count, so the callback fires early rather than late. The charset is read after `super.getWriter()`, once the servlet API has locked the encoding. An unknown or decode-only charset falls back to UTF-8, and the container fails that writer anyway. Once the callback has fired, the counting loop is skipped. `CharBuffer.wrap(buf)` with `off` as an absolute index is right. `print`, `append`, `printf`/`format` and `println` all go through the counted methods, and the `printf` and `write(int)` rows confirm it on Tomcat.
- The new Tomcat `/multi-byte-writer` row fails at `5c8fc21206` (headers missing in every run last round) and passes now, so it pins the fix.
- The Spring Security ordering paragraph in `security.adoc` describes the trade-off accurately. At Tomcat's default buffer, UTF-8 text from about 8 KB of encoded bytes now gets the Grails defaults before Spring Security's character-counting wrapper fires. With Spring Security's default headers this changes nothing: `X-Frame-Options` is written unconditionally, the `nosniff` and `X-XSS-Protection: 0` values are the same, and Grails HSTS is off by default. Only a non-default Spring Security value for a skip-if-present header (for example `Referrer-Policy`) is affected, and the paragraph says how to hand that header over.
- The docs changes: the character-encoding claim is corrected in all three places. The "explicitly configured" wording now matches the `explicit`/`enabled`/value logic, including `enabled: false` and the CSP/HSTS `enabled: true` requirement. Appending the upgrade note as section 73 instead of renumbering is reasonable, given the numbered cross-references such as "section 53".
- `REVERSE_PROXY_REQUEST_HEADERS` is package-private. The Tomcat spec cleans up with `@Shared @TempDir` and now checks the exact values of all four headers. Keeping its own client instead of `HttpClientSupport` makes sense, because the new multi-byte row compares the raw byte length.
- CI was still running at the time of review (5 passed, 66 pending, 0 failed).

matrei (approved):

jdaugherty (approved):

jdaugherty:
Once @codeconsole approves, we can merge this.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
```diff
@@ -0,0 +1,226 @@
...
+        try {
+            // Only the scheme is needed, so the forwarded headers are applied to the request's
+            // scheme alone rather than to its URL: java.net.URI rejects request paths the
+            // container may accept, such as those allowed through Tomcat's relaxedPathChars.
+            URI base = URI.create(request.getScheme() + "://localhost");
+            String scheme = ForwardedHeaderUtils.adaptFromForwardedHeaders(base, headers).build().getScheme();
+            return HTTPS.equalsIgnoreCase(scheme);
+        }
+        catch (IllegalArgumentException malformedForwardedHeaders) {
+            // A forwarded port that is not a number: no trustworthy scheme to go on.
```
codeconsole: Nit (optional): this also catches `ForwardedHeaderUtils` rejecting a malformed forwarded host and `URI.create` rejecting the scheme, not only a non-numeric port.

```suggestion
            // Malformed forwarded headers: no trustworthy scheme to go on.
```

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/SecurityHeadersResponseWrapper.java
```diff
@@ -0,0 +1,402 @@
...
+    @Override
+    public ServletOutputStream getOutputStream() throws IOException {
+        return new CommitAwareOutputStream(super.getOutputStream());
+    }
+
+    @Override
+    public PrintWriter getWriter() throws IOException {
+        PrintWriter writer = super.getWriter();
+        return new CommitAwareWriter(writer, getCharacterEncoding());
+    }
```
codeconsole: Nit (optional): each call builds a new `CommitAwareWriter` and resolves the charset again (plus `newEncoder()` for anything but UTF-8), whereas the container hands back the same writer every time. Caching it keeps `response.writer.is(response.writer)` true like the unwrapped response. `reset()` has to clear it, since the encoding can change after a reset:

```java
private PrintWriter writer;

@Override
public PrintWriter getWriter() throws IOException {
    if (this.writer == null) {
        this.writer = new CommitAwareWriter(super.getWriter(), getCharacterEncoding());
    }
    return this.writer;
}
```

plus `this.writer = null;` in `reset()`. I tried it: the same instance comes back on repeated calls, a new one after `reset()`, and the Tomcat commit-path cases above still send the headers. `getOutputStream()` could get the same treatment.

codeconsole (approved):
Re-ran my Tomcat probes at `8da7b05c58`. Everything I raised is fixed on the wire:

```
case                                                   ee6be46      8da7b05
50 KB body, OutputStream, no Content-Length            <missing>    sent
50 KB body, getWriter(), no Content-Length             <missing>    sent
setContentLength(3) + println('hi')                    <missing>    sent
direct request, KUBERNETES_SERVICE_HOST/PORT set       <missing>    sent
request with X-Forwarded-For                           <missing>    sent
```

I also read through the byte counting, the `reset()` re-arming, the typed-registration condition and the `ForwardedHeaderUtils` scheme resolution, and found nothing blocking. Two optional nits inline.

The one failing check, the Groovy snapshot canary, failed to download Gradle 8.14.4 (HTTP 500 from GitHub) and doesn't run this PR's code.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/GrailsSecurityHeadersFilter.java
jdaugherty: Done in 63aaf3d594: the comment now reads "Malformed forwarded headers: no trustworthy scheme to go on." The spec asserting HSTS is withheld now also runs with a malformed `X-Forwarded-Host: a:b:c` and `Forwarded: host=a:b:c;proto=https`, alongside the non-numeric port. Both carry `proto=https`, so they only pass if this catch is reached.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/SecurityHeadersResponseWrapper.java
jdaugherty: Done in 63aaf3d594: `getWriter()` and `getOutputStream()` return the same instance on every call, and `reset()` drops the cached writer.

One thing that came up while testing: on Tomcat 11 the encoding change after a reset can't be observed. Once `getWriter()` has been called, `Response.reset()` resets the `OutputBuffer` but keeps its converter, so a charset set after the reset never reaches the wire. The Tomcat spec therefore checks the identity contract (same writer and stream on repeated calls, a new writer after `reset()`), and the mock spec covers a reset from UTF-8 to UTF-16BE whose body fills the buffer only when counted in the new encoding. Removing either the caching or the `reset()` clear fails a test in both specs.

PR #15969, opened 2026-07-10
Log plugin loaded successfully messages at DEBUG instead of INFO

codeconsole:
The per-plugin "Grails plug-in [...] loaded successfully" messages from `DefaultGrailsPluginManager` and `DefaultPluginDiscovery` add 20+ INFO lines to every application startup:

```
2026-07-10T18:21:53.386-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [dataBinding] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.387-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [sitemesh3] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.387-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [restResponder] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.388-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [i18n] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.389-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [core] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.389-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [groovyPages] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.389-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [urlMappings] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.390-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [controllers] with version [8.0.0-M2] loaded successfully
2026-07-10T18:21:53.390-07:00  INFO 17975 --- [  restartedMain] g.plugins.DefaultGrailsPluginManager     : Grails plug-in [codecs] with version [8.0.0-M2] loaded successfully
[...]

PR #15970, opened 2026-07-11
Fix domainClass plugin being registered and loaded twice

codeconsole:
Every Grails 8 application startup registers and loads the `domainClass` plugin twice (visible in M2, M3, and current SNAPSHOTs):

```
INFO ... g.plugins.DefaultGrailsPluginManager : Grails plug-in [domainClass] with version [8.0.0-M3] loaded successfully
...
INFO ... g.plugins.DefaultGrailsPluginManager : Grails plug-in [domainClass] with version [8.0.0-M3] loaded successfully
```

### Root cause

`org.grails.plugins.domain.DomainClassGrailsPlugin` is declared by **two** `META-INF/grails-plugin.xml` descriptors on every application classpath:

1. `grails-core.jar` — the hand-written descriptor declares `CoreGrailsPlugin` **and** `DomainClassGrailsPlugin`. This is a leftover from Grails 2.x, when the class still lived in grails-core; the class later moved to the `grails-domain-class` module but the entry was never removed.
2. `grails-domain-class.jar` — carries its own build-generated descriptor declaring the same class.

`ClasspathPluginFinder` creates one `PluginInfo` per *(descriptor, class)* pair and `DefaultPluginDiscovery.registerPlugin()` had no already-registered check, so both entries flow into the load order: the plugin is sorted, instantiated, and logged twice, with the second `plugins.put` silently overwriting the first instance.

### Fix

- Remove the stale `DomainClassGrailsPlugin` entry from grails-core's hand-written descriptor. This is safe: the generated descriptor is packaged inside the `grails-domain-class` jar, so the class and its descriptor always travel together — the stale entry could only ever produce a duplicate (jar present) or a class-not-found skip (jar absent).
- Defensively, `registerPlugin()` now skips a plugin whose name is already registered and logs a WARN including both versions, so a plugin accidentally declared by multiple descriptors can never double-load:

```
WARN ... Grails plug-in [domainClass] with version [8.0.0-M3] is already registered; skipping duplicate declaration with version [8.0.0-M3]
```

### Tests

Two new `PluginDiscoverySpec` features, exercised through public APIs only:

- the same plugin class declared in two classpath `grails-plugin.xml` descriptors (reproducing the actual bug via the thread-context class loader) registers exactly once
- the same plugin class supplied twice registers exactly once and the skipped duplicate is reported at WARN

Both tests were verified to fail against the previous code.

Note: the duplicate `domainClass` line is also visible in the startup output shown in #15969; the two PRs are independent and merge cleanly in either order (verified with `git merge-tree`).

### DIFF
--- a/grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
+++ b/grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
@@ -483,12 +483,26 @@ public class DefaultPluginDiscovery implements PluginDiscovery {
     /**
      * Registers a plugin that is eligible to be loaded.
      *
-     * <p>This method skips disabled plugins, records eviction and observer relationships, and appends the
-     * plugin to the discovered load order.</p>
+     * <p>This method skips disabled plugins and plugins already registered under the same name (a plugin
+     * class may be declared by more than one {@code META-INF/grails-plugin.xml} descriptor on the
+     * classpath), records eviction and observer relationships, and appends the plugin to the discovered
+     * load order.</p>
      *
      * @param plugin the plugin to register
      */
     private void registerPlugin(PluginInfo plugin) {
+        var registeredPlugin = plugins.get(plugin.getName());
+        if (registeredPlugin != null) {
+            LOG.warn(
+                    "Grails plug-in [{}] with version [{}] is already registered; " +
+                            "skipping duplicate declaration with version [{}]",
+                    plugin.getName(),
+                    registeredPlugin.getPluginVersion(),
+                    plugin.getPluginVersion()
+            );
+            return;
+        }
+
         if (!plugin.getMetadata().canRegisterPlugin()) {
             if (LOG.isInfoEnabled()) {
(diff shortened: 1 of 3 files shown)


grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
```diff
@@ -483,12 +483,26 @@ private void attemptRegisterPlugins(List<PluginInfo> filteredPlugins) {
...
+     * class may be declared by more than one {@code META-INF/grails-plugin.xml} descriptor on the
+     * classpath), records eviction and observer relationships, and appends the plugin to the discovered
+     * load order.</p>
      *
      * @param plugin the plugin to register
      */
     private void registerPlugin(PluginInfo plugin) {
+        var registeredPlugin = plugins.get(plugin.getName());
+        if (registeredPlugin != null) {
+            LOG.warn(
```
jdaugherty, line 496: This should throw instead of warn.

jdaugherty: This shouldn't throw - it was silently repressed in past versions.  Leaving this as -is

PR #15971, opened 2026-07-11
Compile-time check for GORM query strings flattened from GString (alternative to #15968)

borinquenkid:
## Background

While reviewing #15968 ("Warn once on GString-interpolated GORM HQL queries"), GitHub's Copilot reviewer left a comment on `HibernateGormStaticApi.groovy` noting that the new runtime warning had no test verifying it actually fires through the real Hibernate static API call path — only through the isolated `GormQuerySafetyWarnings` helper spec. That comment was the starting point for a deeper investigation (with Claude) into whether the warning mechanism actually closes the gap it claims to.

That investigation found:

1. **The warning does fire correctly when a `GString` is passed directly** to `find`/`findAll`/`executeQuery` — confirmed empirically by adding a log-capturing assertion to the existing `HibernateGormStaticApiSpec` "injection-safe" test and running it against #15968's branch. GORM binds the interpolated value as a real query parameter in this case; it is not exploitable.
2. **#15968 only wires the warning into one of three GORM implementations with the identical code path.** `grails-data-hibernate5`'s `AbstractHibernateGormStaticApi` and `grails-data-neo4j`'s `Neo4jGormStaticApi` have the same `instanceof GString` / `buildNamedParameterQueryFromGString` pattern (Neo4j's error message still literally says "HQL injection" even though it's building Cypher) — neither references the new warning helper at all.
3. **The actual dangerous case is structurally invisible to any runtime check.** GORM's own safe-binding mechanism only engages when the query argument is *still* a `GString` at the point of the call. The extremely common Groovy pattern
   ```groovy
   String query = "from Book where name = ${userInput}"   // GString -> String, right here
   Book.executeQuery(query)
   ```
   causes Groovy to coerce the `GString` to a plain `String` at the assignment, because `query` is declared `String`. By the time this reaches GORM's query builder, `instanceof GString` is `false`, and the interpolated value is used as raw, unescaped query text — a genuine injection vector. A `java.lang.String` carries no trace of ever having been a `GString`, so no amount of runtime checking at the query boundary can distinguish this from a safely hand-written HQL string. The information needed to tell them apart only exists at the source-code level, before compilation erases it.

That third point is the actual finding this PR addresses: a warning that fires on the already-safe case and stays silent on the actually-dangerous case is backwards from what a risk-reduction mechanism should do, regardless of whether it warns or (as considered and rejected) throws at runtime.

## What this PR does

Adds a **global, compile-time Groovy AST transformation** (`GormQuerySafetyTransformer` / `GlobalGormQuerySafetyASTTransformation`, in `grails-datamapping-core`) that detects the flattened-`String` pattern above and fails the build with a clear, actionable error — before the type information that would let a runtime check catch it is lost.

Because the check is syntactic (method name + argument shape), not tied to a specific datastore's runtime class, **one transform covers Hibernate5, Hibernate7, and Neo4j simultaneously**, with zero changes to any of those three modules and zero configuration required from application developers — it runs automatically wherever `grails-datamapping-core` is on the compile classpath, the same way `@GrailsCompileStatic` and domain-class enhancement already apply themselves invisibly.

### Detection algorithm

- Tracks local variables (declarations and reassignments) that go from a live, interpolated `GString` to a plain `String` — via explicit `String` typing, `.toString()`, `(String)` cast, or `as String`.
- Flags a call to `find` / `findAll` / `executeQuery` / `executeUpdate` / `findAllWithSql` / `cypherStatic` / `findPath` / `findPathTo` whose query argument is one of those flattened variables.
- Gates on the receiver looking like a GORM domain class (reusing the existing `AstUtils.isDomainClass`, the same utility `DetachedCriteriaTransformer`'s `where{}` rewriting already relies on) to avoid false positives on unrelated methods.
- A direct `GString` literal argument (`Book.executeQuery("... ${x} ...")`) is never flagged — that path is already safe.

### Corner cases handled (with test coverage)

- **`.toString()` / `(String)` cast / `as String`** coercions of an interpolated `GString` — all flagged, not just plain `String`-typed declarations.
- **Reassignment**, not just declaration (`String q; q = "...${x}..."`) — tracked via `visitBinaryExpression`, with a safe reassignment clearing prior unsafe tracking (last-write-wins; see limitations below).
- **`Collection.find` / `Collection.findAll`** (GDK methods present on nearly every `Iterable`) never collide with the GORM methods of the same name — they take a `Closure`, and the flagged argument must be a tracked `String` variable, so the two shapes can't overlap.
- **Non-domain classes that happen to define their own `find`/`findAll(String)`** methods do not false-positive, because of the `isDomainClass` receiver gate.
- **Closures capture the flattened state correctly** — a variable flattened in an enclosing method and referenced inside a nested closure (`.each { ... executeQuery(q) }`) still triggers, since closures are walked as part of their enclosing method's traversal.
- **Neo4j's `findPathTo(Class type, CharSequence query, Map params)`** has the query as its *second* argument, not first — handled as a special case in the argument-index map, with a dedicated test.
- **Suppression**: a reviewed, genuinely safe call site can opt out per-call-site with `@SuppressWarnings("GormUnsafeQueryString")` on the enclosing method — a deliberate, auditable exception rather than a global kill switch.
- **Multiple-assignment declarations** (`def (a, b) = [...]`) don't crash the transform — `DeclarationExpression.getVariableExpression()` returns `null` for these and is guarded.

### Known, deliberate limitations (documented in the class Javadoc and the upgrade guide)

This is intentionally scoped, not a general SQL-injection solution:

- **Intraprocedural only** — a flattened `String` built inside a helper method and returned is invisible to this check.
- **Last-write-wins reassignment tracking**, not full branch-sensitive dataflow.
- **Locals only, not fields**, in this first version.
- **Does not catch plain string concatenation with no `GString` involved at all** (`"select * ... " + userInput`) — arguably the more classic injection shape, and untouched by this or #15968's runtime check, since there's no `GStringExpression` node to detect.
- **Does not cover raw JDBC via `groovy.sql.Sql`**, which is outside GORM's static API entirely.
- **Does not apply to MongoDB** — its GORM API has no `CharSequence`-based raw query method at all (it inherits `GormStaticApi.executeQuery()`, which just throws `unsupported('executeQuery')`), so it isn't vulnerable to this specific pattern in the first place.

## Verification

- New `GormQuerySafetyTransformerSpec` — 12 tests, covering every case above (error cases assert `MultipleCompilationErrorsException` with the expected message; safe/suppressed/non-domain cases assert clean compilation).
- Full `grails-datamapping-core` test suite: pass.
- Full `grails-data-hibernate7-core` suite (2988 tests) and `grails-data-hibernate5-core` suite: zero failures, confirming the global transform introduces no false positives against real GORM/query source in either module.
- `codeStyle` (Checkstyle + CodeNarc) clean on touched modules.
- Doc build (`publishGuide`) verified to render correctly, including the new cross-reference between the upgrade guide and the security guide.

## Why this is opinionated, on purpose

Grails' own stated philosophy is "sensible defaults" — convention over configuration, with escape hatches for the cases that genuinely need them. Not permitting query-string injection by default is exactly that kind of sensible default: an application shouldn't have to opt in to safety, and a framework that discovers it's silently coercing a safely-parameterizable query into raw unescaped text has an obligation to say so loudly, not log it once and move on. That's why this fails the build rather than warning: for a compile-time check, the "pre-prod environment" every responsible team already runs *is* the build itself — failing to compile has no live-traffic blast radius, unlike a runtime throw would. Teams that hit a false positive on a reviewed-safe call get an explicit, auditable, per-call-site escape hatch (`@SuppressWarnings("GormUnsafeQueryString")`) rather than a single switch that quietly reopens the whole risk surface.

This is offered as an alternative to the runtime-only, single-module approach in #15968 — happy to discuss reconciling the two (e.g. keeping #15968's docs/warning as a softer signal for the cases this transform's intraprocedural limits can't reach, while this closes the case that matters most).

### DIFF
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
@@ -1,0 +1,57 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.datastore.gorm.query.transform;
+
+import java.util.List;
+
+import org.codehaus.groovy.ast.ASTNode;
+import org.codehaus.groovy.ast.ClassNode;
+import org.codehaus.groovy.ast.ModuleNode;
+import org.codehaus.groovy.control.CompilePhase;
+import org.codehaus.groovy.control.SourceUnit;
+import org.codehaus.groovy.transform.ASTTransformation;
+import org.codehaus.groovy.transform.GroovyASTTransformation;
+import org.codehaus.groovy.transform.TransformWithPriority;
+
+import org.apache.grails.common.compiler.GroovyTransformOrder;
+
+/**
+ * Global version of {@link GormQuerySafetyTransformer} - runs automatically against every class
+ * in every Grails application that has {@code grails-datamapping-core} on its compile classpath,
+ * with no developer configuration required.
+ *
+ * @since 8.1
+ */
+@GroovyASTTransformation(phase = CompilePhase.CANONICALIZATION)
+public class GlobalGormQuerySafetyASTTransformation implements ASTTransformation, TransformWithPriority {
+
+    public void visit(ASTNode[] nodes, SourceUnit source) {
+        GormQuerySafetyTransformer transformer = new GormQuerySafetyTransformer(source);
+        ModuleNode ast = source.getAST();
+        List<ClassNode> classes = ast.getClasses();
+        for (ClassNode aClass : classes) {
+            transformer.visitClass(aClass);
+        }
+    }
+
+    @Override
+    public int priority() {
+        return GroovyTransformOrder.QUERY_SAFETY_ORDER;
+    }
+}
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
@@ -36,12 +36,31 @@ import org.apache.grails.common.compiler.GroovyTransformOrder;
  * in every Grails application that has {@code grails-datamapping-core} on its compile classpath,
  * with no developer configuration required.
  *
+ * <p>Since this check is intraprocedural and does not cover every possible way query text can be
+ * built (see {@link GormQuerySafetyTransformer}'s documented limitations), a build that hits a
+ * false positive it can't otherwise work around may need to disable it entirely rather than per
+ * call site. Setting the {@value #PROTECT_SQL_INJECTION_ATTACKS_PROPERTY} system property to
+ * {@code false} skips this transformation for the whole compilation unit - this is a global kill
+ * switch of last resort, not a substitute for {@link GormQuerySafetyTransformer#SUPPRESS_WARNINGS_VALUE}.
+ *
  * @since 8.1
  */
 @GroovyASTTransformation(phase = CompilePhase.CANONICALIZATION)
 public class GlobalGormQuerySafetyASTTransformation implements ASTTransformation, TransformWithPriority {
 
+    /**
+     * System property that enables this check; defaults to {@code true}. There is no fully
+     * accurate way to guarantee protection against SQL/HQL/Cypher injection at compile time (this
+     * check has documented gaps), so the name describes intent, not a guarantee. Set to
+     * {@code false} only as a last resort, e.g. a false positive blocking a build with no other
+     * workaround.
+     */
+    public static final String PROTECT_SQL_INJECTION_ATTACKS_PROPERTY = "protectSqlInjectionAttacks";
+
     public void visit(ASTNode[] nodes, SourceUnit source) {
+        if (!Boolean.parseBoolean(System.getProperty(PROTECT_SQL_INJECTION_ATTACKS_PROPERTY, "true"))) {
+            return;
+        }
         GormQuerySafetyTransformer transformer = new GormQuerySafetyTransformer(source);
         ModuleNode ast = source.getAST();
         List<ClassNode> classes = ast.getClasses();
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
@@ -43,7 +43,7 @@ import org.apache.grails.common.compiler.GroovyTransformOrder;
  * {@code false} skips this transformation for the whole compilation unit - this is a global kill
  * switch of last resort, not a substitute for {@link GormQuerySafetyTransformer#SUPPRESS_WARNINGS_VALUE}.
  *
- * @since 8.1
+ * @since 8.0
  */
 @GroovyASTTransformation(phase = CompilePhase.CANONICALIZATION)
 public class GlobalGormQuerySafetyASTTransformation implements ASTTransformation, TransformWithPriority {
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
@@ -32,6 +32,7 @@ import org.codehaus.groovy.ast.AnnotationNode;
 import org.codehaus.groovy.ast.ClassCodeVisitorSupport;
 import org.codehaus.groovy.ast.ClassHelper;
 import org.codehaus.groovy.ast.ClassNode;
+import org.codehaus.groovy.ast.FieldNode;
 import org.codehaus.groovy.ast.MethodNode;
 import org.codehaus.groovy.ast.expr.ArgumentListExpression;
 import org.codehaus.groovy.ast.expr.BinaryExpression;
@@ -43,9 +44,13 @@ import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.GStringExpression;
 import org.codehaus.groovy.ast.expr.ListExpression;
 import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.codehaus.groovy.ast.expr.PropertyExpression;
 import org.codehaus.groovy.ast.expr.StaticMethodCallExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
+import org.codehaus.groovy.ast.stmt.IfStatement;
 import org.codehaus.groovy.control.SourceUnit;
+import org.codehaus.groovy.control.messages.WarningMessage;
+import org.codehaus.groovy.syntax.Token;
 import org.codehaus.groovy.syntax.Types;
 
 import org.grails.datastore.mapping.reflect.AstUtils;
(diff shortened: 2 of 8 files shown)


borinquenkid:
@bito-code-review Good catch, and correct — the detector only matched the immediate coercion expression, so this slipped through:

```groovy
def g = "from Book where name = ${x}"   // g: still a live GString
String q = g                             // flattened HERE, uncaught
Book.executeQuery(q)
```

Fixed in 46c4ba2486: replaced the single `flattenedStringVars` set with a two-state model (`LIVE_GSTRING` / `FLATTENED`) that resolves through `VariableExpression` references, so the flattening point is found correctly regardless of how many variable-to-variable hops separate the original `GString` literal from the query call.

Added regression tests for the exact case above, a two-hop alias chain (`def g = ...; def h = g; String q = h`), `.toString()` called on an alias, and re-aliasing an already-flattened variable — all in `GormQuerySafetyTransformerSpec`. Full `grails-datamapping-core`, `grails-data-hibernate7-core`, and `grails-data-hibernate5-core` suites still pass with zero regressions.

Still out of scope by design (see the class Javadoc / PR description): a flattened `String` built inside a *different* method and returned is invisible to this check, since the analysis is intraprocedural.

davydotcom (approved):
This supersedes #15968 and is a fairly elegant compile time safety check. It is important to note there are loopholes it cannot account for and it may be worth considering if this is even the best way to do sql injection protections long term since its hard to determine casting to a String. I approve this and recommend it for merge

borinquenkid:
Pushed a follow-up commit (`cceff551`) extending this check, based on a review discussion about what the check does and doesn't cover, and how confidently it should act on each case: for patterns the detection can't verify with full confidence, warn instead of failing the build - the goal isn't to prevent every possible unsafe pattern, just to close the gap decisively where detection is precise, and surface the rest for review rather than staying silent on it.

Three changes:

1. **Bug fix: branch-sensitive reassignment tracking.** The existing local-variable tracking was last-write-wins across an `if`/`else` - a variable flattened in only one branch could be silently cleared if the other branch reassigned it safely, depending on which branch the visitor happened to traverse last (not which one actually runs at runtime). `visitIfElse` now walks each branch from the same starting state and merges pessimistically: unsafe in either branch stays unsafe after the statement. Still build-breaking, since this is the same precise local-variable case as before.

2. **New warning: field-based flattening.** A `String`-typed field assigned an interpolated `GString` (via its initializer or `this.field = ...`) that's later read through `this.field` in a query call is now flagged. This is a compile-time warning, not an error - a field can be reassigned from a constructor, another method, or a subclass that this check never visits, so the detection is inherently less certain than the local-variable case.

3. **New warning: string concatenation.** Query text built with `+` from a non-constant value and no `GString` involved at all (e.g. `"select ... " + userInput`) now warns. Also a warning rather than an error, since concatenation is common enough for benign, non-query purposes that a hard failure would be too blunt an instrument. One subtlety this turned up: concatenating a *live* `GString` with more text (`"...${x}..." + " order by title"`) actually returns a plain `String` at runtime (`GString.plus` returns `String`), so that case is correctly promoted to the existing build-breaking error rather than this warning - it's the same irreversible coercion `.toString()` causes, not a lower-confidence pattern.

Both new warnings share the existing `@SuppressWarnings("GormUnsafeQueryString")` suppression.

Verification:
- 10 new cases added to `GormQuerySafetyTransformerSpec` (26 total, all passing) covering both warnings, the branch-sensitivity fix, and the GString-concatenation promotion.
- Full `grails-datamapping-core` suite green, `codeStyle` clean.
- Compiled `grails-data-hibernate5-core` and `grails-data-hibernate7-core`'s test sources under the stricter checks - clean, confirming no new false positives against real GORM implementation code.
- Doc guide (`securingAgainstAttacks.adoc`, `upgrading80x.adoc`) updated and rendered via `publishGuide` to cover both new warnings.

grails-datamapping-core/src/main/resources/META-INF/services/org.codehaus.groovy.transform.ASTTransformation
```diff
@@ -1,2 +1,3 @@
 org.grails.datastore.gorm.query.transform.GlobalDetachedCriteriaASTTransformation
 org.grails.compiler.gorm.GlobalJpaEntityTransform
+org.grails.datastore.gorm.query.transform.GlobalGormQuerySafetyASTTransformation
```
jdaugherty, line 3: Registering this globally makes it a **hard compile error, on by default, for every project with `grails-datamapping-core` on the classpath**, with the only escape hatch being per-call-site `@SuppressWarnings`. Two things worth weighing before that ships enabled-by-default on the 8.0.x line:

1. **False positives on trusted interpolation.** The check flags *any* interpolated GString flattened to `String`, but interpolating a non-user value is common and safe — `"from Book where status = ${Status.ACTIVE.name()}"`, an enum, a numeric id, a config constant. Each of those now breaks the build and forces a suppression annotation despite no injection risk. The check cannot distinguish trusted from untrusted data, so it fails safe code, not just unsafe code.
2. **No global opt-out.** On upgrade a team with even one such pattern cannot compile until every site is annotated. Given the approval note about the loopholes this can't close, consider whether the *default* should be a warning (matching the field/concatenation cases) with an opt-in property to promote it to an error, rather than error-by-default with only per-site suppression.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
```diff
@@ -0,0 +1,678 @@
...
+            }
+            if (concatenatedStringVars.containsKey(name)) {
+                return Finding.UNSAFE_CONCATENATION;
+            }
+            return Finding.NONE;
+        }
+        if (isThisFieldReference(argument) && flattenedFields.containsKey(fieldNameOf(argument))) {
+            return Finding.FLATTENED_FIELD;
+        }
+        Origin concatOrigin = classifyConcatenation(argument);
```
jdaugherty, line 587: Inline flattening *at the call site* isn't detected. `findUnsafeArgument` classifies the argument only when it's a `VariableExpression`, a `this.field` reference, or a concatenation — a `.toString()` call or a `(String)`/`as String` cast applied directly in the argument slot falls through to `classifyConcatenation`, which returns `NONE`:

```groovy
Book.executeQuery("from Book where t = ${x}".toString())   // not flagged
Book.executeQuery((String) "from Book where t = ${x}")     // not flagged
```

This is the same irreversible coercion the tracked-variable tests already cover (`def q = "...".toString(); find(q)`), so the asymmetry both leaves a real vector open and is surprising to a user. The logic already exists — `classify(argument, ClassHelper.STRING_TYPE)` (or `isUnsafeSource`) returns `FLATTENED` for both forms; running the raw argument through it here would close the gap. Please add explicit tests for each inline form as well.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
```diff
@@ -0,0 +1,678 @@
...
+ *     across loops, {@code switch}, or {@code try}/{@code catch} - and is last-write-wins for
+ *     fields, which are not branch-sensitive at all.</li>
+ *     <li>Field tracking only recognizes a directly-interpolated {@code GString} initializer or
+ *     {@code this.field = ...} assignment - it does not follow aliasing chains or
+ *     {@code .toString()}/cast coercions the way local tracking does.</li>
+ *     <li>Does not detect raw JDBC via {@code groovy.sql.Sql}, or any datastore whose query
+ *     methods use names outside {@link #CANDIDATE_METHODS}.</li>
+ * </ul>
+ *
+ * @since 8.1
```
jdaugherty: `@since 8.1` (here and in `GlobalGormQuerySafetyASTTransformation`) doesn't match the target. This PR is against `8.0.x` (`projectVersion=8.0.0-SNAPSHOT`) and the upgrade note is in `upgrading80x.adoc`, described as something "Grails 8 adds." Either the tag should be `8.0.0`, or — if this is genuinely intended for 8.1 — a compile-breaking change probably shouldn't land on the 8.0.x branch. Please reconcile the version.

jdaugherty (commented):
My largest concern is not having an opt out that can be configured - in the event of an AST related bug.  Otherwise, I love this change.

grails-datamapping-core/src/main/resources/META-INF/services/org.codehaus.groovy.transform.ASTTransformation
jdaugherty: Just checking for a system property as an opt out mechanism would suffice for my concern.

matrei (commented):
Nice work @borinquenkid!

I agree with @jdaugherty, there should be a global opt-out to not run the check at all and strings created at the call site should also be flagged for consistency.

jdaugherty:
re-reading my comments, i want to be clear, i'm ok for this on by default, i just want an opt out - mostly in case something goes wrong

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformationSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+
+    static List byTitle(String title) {
+        String q = "from Book where title = ${title}"
+        executeQuery(q)
+    }
+}
+'''
+
+    void cleanup() {
+        System.clearProperty(GlobalGormQuerySafetyASTTransformation.PROTECT_SQL_INJECTION_ATTACKS_PROPERTY)
```
jdaugherty: Can you please adopt the RestoreSystemProperties annotation like we have done in the other tests?

paulk-asert:
Groovy 6 throws a SqlException rather than just provide a warning in more cases with a system property for opt-out:

https://issues.apache.org/jira/browse/GROOVY-12118
https://github.com/apache/groovy/pull/2647

Not directly relevant but confirms the direction being taken here will become familiar to Groovy folks.

But having said that, the case AI is showing looks weird to me:

```
String query = "from Book where name = ${userInput}"
```

As a String, would this still be well-formed? For SQL, this would go to: "from Book where name = someInput" but this wouldn't be well-formed unless it was "from Book where name = 'someInput'". I am not sure how HQL handles that. The case that Groovy looks for is:

```
String query = "from Book where name = '${userInput}'"
```

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
jdaugherty, line 587: This is still open at the current head — I verified empirically that all three inline forms compile cleanly when they should fail:

```groovy
Book.executeQuery("from Book where title = ${title}".toString())   // not flagged
Book.executeQuery((String) "from Book where title = ${title}")      // not flagged
Book.executeQuery("from Book where title = ${title}" as String)     // not flagged
```

A non-variable argument still only reaches `classifyConcatenation`, so a `.toString()` or cast applied directly in the argument slot falls through to `NONE`. Handling the two shapes here closes it (`as String` is a coercing `CastExpression`, so it's covered by the first branch):

```suggestion
        if (argument instanceof CastExpression) {
            CastExpression cast = (CastExpression) argument;
            if (ClassHelper.STRING_TYPE.equals(cast.getType()) && isUnsafeSource(cast.getExpression())) {
                return Finding.FLATTENED_GSTRING;
            }
        }
        if (argument instanceof MethodCallExpression) {
            MethodCallExpression flatteningCall = (MethodCallExpression) argument;
            if ("toString".equals(flatteningCall.getMethodAsString()) && isUnsafeSource(flatteningCall.getObjectExpression())) {
                return Finding.FLATTENED_GSTRING;
            }
        }
        Origin concatOrigin = classifyConcatenation(argument);
```

And a matching test for `GormQuerySafetyTransformerSpec`:

```groovy
    @Unroll
    void "test inline #description at the call site fails to compile"() {
        when:
        new GroovyClassLoader().parseClass("""
import grails.gorm.annotation.Entity

@Entity
class Book {
    String title

    static List byTitle(String title) {
        executeQuery(${argument})
    }
}
""")

        then:
        def e = thrown(MultipleCompilationErrorsException)
        e.message.contains('GormUnsafeQueryString')

        where:
        description     | argument
        '.toString()'   | '"from Book where title = ${title}".toString()'
        '(String) cast' | '(String) "from Book where title = ${title}"'
        'as String'     | '("from Book where title = ${title}" as String)'
    }
```

I ran exactly that test against the current head and all three cases fail (no error is raised), so it will go green once the change above is in.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GormQuerySafetyTransformer.java
jdaugherty: Still `8.1` at the current head while the branch is `8.0.0-SNAPSHOT` and the doc note is in `upgrading80x.adoc`:

```suggestion
 * @since 8.0
```

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/transform/GlobalGormQuerySafetyASTTransformation.java
```diff
@@ -0,0 +1,76 @@
...
+ * with no developer configuration required.
+ *
+ * <p>Since this check is intraprocedural and does not cover every possible way query text can be
+ * built (see {@link GormQuerySafetyTransformer}'s documented limitations), a build that hits a
+ * false positive it can't otherwise work around may need to disable it entirely rather than per
+ * call site. Setting the {@value #PROTECT_SQL_INJECTION_ATTACKS_PROPERTY} system property to
+ * {@code false} skips this transformation for the whole compilation unit - this is a global kill
+ * switch of last resort, not a substitute for {@link GormQuerySafetyTransformer#SUPPRESS_WARNINGS_VALUE}.
+ *
+ * @since 8.1
```
jdaugherty: Same here:

```suggestion
 * @since 8.0
```

jdaugherty (approved):
All four points from my earlier rounds are addressed at `cb3a575`, verified locally against the PR head:

1. **Global opt-out** — `protectSqlInjectionAttacks` system property (defaults to `true`) short-circuits the global transform, with three tests covering unset/`true`/`false` and the behavior documented in `upgrading80x.adoc`.
2. **`@RestoreSystemProperties`** — adopted in `GlobalGormQuerySafetyASTTransformationSpec`.
3. **Inline flattening at the call site** — `findUnsafeArgument` now runs a non-variable argument through the `CastExpression` / `.toString()` checks, so all three forms I verified as slipping through previously (`"...${x}".toString()`, `(String) "...${x}"`, `"...${x}" as String`) are flagged, each with its own test.
4. **`@since`** — reconciled to `8.0` in both `GormQuerySafetyTransformer` and `GlobalGormQuerySafetyASTTransformation`.

Verified: `GormQuerySafetyTransformerSpec` (30) and `GlobalGormQuerySafetyASTTransformationSpec` (3) all pass, and the full `grails-datamapping-core` suite is green (289 tests, 0 failures).

The documented gaps remain (interprocedural flattening, fields, plain concatenation, `groovy.sql.Sql`), but they are called out in both the class Javadoc and the guide, and the kill switch covers the AST-bug scenario that was my main concern. LGTM.

PR #15973, opened 2026-07-11
Fix generated logback-spring.xml hiding WARN and INFO messages

codeconsole:
## Summary

The starter `grails-app/conf/logback-spring.xml` generated for new applications sets the root logger to `ERROR`, which suppresses the WARN and INFO output — including framework startup warnings — that Spring Boot shows by default.

This PR raises the root level to `INFO` and composes Spring Boot's own `defaults.xml` and `console-appender.xml` instead of duplicating the console appender, so the generated configuration matches Spring Boot's defaults. Jansi remains unused; Spring Boot's `%clr` converter supplies console colors (see #15663).

The change is applied to both generation paths so `create-app` and start.grails.org produce the same file:

- **Profile CLI**: `grails-profiles/base/skeleton/grails-app/conf/logback-spring.xml` (also fixes the commented development logger example to use the `@grails.codegen.defaultPackage@` placeholder instead of a hardcoded package name)
- **Grails Forge**: `grails-forge-core` `logback.rocker.raw` template, with `LogbackSpec` updated to assert the new content

This is the logback-spring.xml portion of #15757, extracted into its own PR. The opt-in zero-config Forge feature and documentation changes remain in #15757.

## Testing

- Full `grails-forge-core` test suite passes, including the updated `LogbackSpec` assertions (Spring Boot includes present, `INFO` root level, development `springProfile`, no Jansi) across all application types.

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/logging/template/logback.rocker.raw
@@ -22,43 +22,25 @@ under the License.
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
     <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
+    <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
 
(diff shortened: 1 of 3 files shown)


jdaugherty:
@jamesfredley I am merging this to ensure log output for the M3.  If you still disagree with any if this, lets fix in a subsequent PR?

PR #15974, opened 2026-07-11
Internationalize the default welcome page

codeconsole:
The application generated by grails-forge ships a language selector in its navbar, but `index.gsp` was hardcoded English — switching languages changed nothing on the page.

### Changes

- **`gsp/index.gsp` template**: every user-facing string now renders through `<g:message>` (headings, card titles, table headers, `aria-label`s). Product names (Grails, Spring Boot, Groovy, JVM) stay literal.
- **All 18 bundled locales** (`i18n/messages*.properties` + English defaults): 31 new `welcome.*` keys per bundle, translated for cs, da, de, es, fr, it, ja, nb, nl, pl, pt_BR, pt_PT, ru, sk, sv, th and zh_CN. The controller count renders through a `MessageFormat` choice, or a genitive phrasing where a language's plural rules can't be expressed as a choice (cs, pl, ru, sk).

### Tests

- `GrailsGspSpec."test default index page is internationalized"` — the generated page uses message codes and no hardcoded user-facing English remains.
- `GrailsDefaultPluginsSpec."test every i18n bundle translates the welcome page"` — every locale bundle carries the welcome keys, with byte-intact spot checks of the Russian/Japanese/Thai/Chinese titles surviving the template pipeline.
- Full `:grails-forge-core:test` run green.

All 18 locales were also verified end-to-end in a generated running application (response `charset=UTF-8`, strict UTF-8 byte validation, translated content matching the bundles).

### Note

Correct rendering of non-Latin-1 locales also depends on the SiteMesh `3.3.0-M3` upgrade (companion PR to follow once M3 syncs to Maven Central), which fixes the response charset in the view-resolver integration. Without it, decorated pages fall back to the container's locale-derived charset.

### Follow-up: bundle backfill and MessageFormat repairs (second/third commits)

Validating the bundles surfaced pre-existing i18n debt, fixed here so every locale is complete and renders correctly:

- **Missing keys backfilled**: `fr` and `zh_CN` lacked 32 keys each (all `default.*` scaffolding labels + all `typeMismatch` messages), `pt_PT` and `ru` lacked the 22 `default.*` labels, and every non-English bundle lacked the bare `typeMismatch` key — those locales silently fell back to English. Backfilled following each file's established vocabulary and conventions.
- **17 broken `MessageFormat` patterns repaired**: unescaped apostrophes in arg-bearing messages corrupted rendered output — six French validation messages (rendered `nest` for `n'est`), two Italian range/size messages (swallowed their `{3}`/`{4}` bounds), nine Thai `typeMismatch` messages (showed literal `{0}`).
- Two typos fixed (`pl`, `ru`), and two labels localized after translation review (`th` servlet container, `zh_CN` profile).
- New `GrailsDefaultPluginsSpec` feature pins bundle integrity: every locale declares every key, and every parameterized message formats without leftover placeholders.

### Follow-up: the cups return, filterable cards, and a controllers nav dropdown (fourth/fifth commits)

- **Welcome hero**: the classic three-cups mark is back as the page's signature — a contained charcoal panel holding the welcome copy, the full mark inlined as a theme-adaptive SVG with a warm halo in the brand accent, one load animation (disabled under `prefers-reduced-motion`), cups receding to a faint backdrop on mobile. With the copy in the hero, Runtime versions joins the other status cards as a uniform four-card row.
- **Filterable controllers/plugins**: both cards gain a funnel dropdown with a live name filter (auto-focus, engaged-state toggle, collapsing namespace groups, localized empty states — `welcome.filter.name`/`welcome.filter.none` in all 18 bundles). The Available Controllers card now mirrors the Installed plugins card (default padding, h6 title, count badge) instead of its oversized `p-md-5`/h2 layout.
- **Theme fix**: count badges swap hardcoded `text-bg-light` (a white glare in dark mode) for adaptive neutral pills. The unused `welcome.controllers.detected` key is removed from all bundles.
- **Layout navbar**: a controllers dropdown on every page, linking each controller's default action, labeled with the localized `welcome.artefact.controllers` key.
- `GrailsGspSpec` pins the hero, filters, badges, and nav dropdown; full `:grails-forge-core:test` green.

### Follow-up: layout chrome internationalized (sixth commit)

The layout wrapping every generated page still hardcoded English — the three footer cards, the light/dark/auto theme selector, and the loading spinner — so they stayed untranslated on all pages regardless of the selected language. All now render through `layout.*` message codes translated in the 18 bundles (including alt text and aria-labels), with a spec asserting no user-facing English remains hardcoded in the layout.

### Follow-up: actuators card and a pinned default language (seventh/eighth commits)

- **Actuators card on the welcome page** (the application-status data already lives in the page's cards): lists the exposed actuator web endpoints with links under the configured management base path, below Installed plugins. Renders only when Spring Boot Actuator is present and exposes endpoints — the supplier type is resolved via `ClassUtils.isPresent`, so removing the actuator dependency can't break the page. `welcome.actuators` added to all 18 bundles.
- **The language menu pins the default language above a divider** — a user who switches to a language they cannot read always has a recognizable way back.
- The navbar stays minimal: controllers, language and theme only — pinned by `GrailsGspSpec` (the intermediate commit adding status/artefacts/plugins navbar dropdowns is reverted by the follow-up).

### Follow-up: profile skeleton parity (tenth commit)

The profile-based `create-app` generates from `grails-profiles` skeletons, which carry their own copies of the welcome templates and had received none of this work — only forge-generated applications got the new page. All 22 shared files (index.gsp, layout, welcome.css/js, 18 bundles) are now synced byte-identical into the `web`/`base` skeletons, and a new `GrailsGspSpec` feature **pins the mirror**: any future change to one side without the other fails the build.


grails-forge/grails-forge-core/src/main/resources/i18n/messages_sk.properties
```diff
@@ -53,4 +53,51 @@ typeMismatch.java.lang.Long=Položka {0} musí byť celé číslo
 typeMismatch.java.lang.Short=Položka {0} musí byť celé číslo
 typeMismatch.java.math.BigDecimal=Položka {0} musí byť desatinné číslo
 typeMismatch.java.math.BigInteger=Položka {0} musí byť celé číslo
+typeMismatch=Položka {0} má nesprávny typ
```
jdaugherty: This adds a second `typeMismatch` key — line 57 already defines `typeMismatch=Položka {0} má nezhodný typ`, so the new line is silently shadowed (Properties last-one-wins). Only the Slovak bundle has this duplicate; the profiles mirror (`grails-profiles/base/skeleton/grails-app/i18n/messages_sk.properties`) carries the same pair. One of the two should be dropped from both copies. The key-completeness spec can't catch duplicates since `Properties` collapses them — might be worth asserting no duplicate keys per bundle while you're in there.

grails-forge/grails-forge-core/src/main/resources/assets/javascripts/welcome.js
```diff
@@ -106,11 +106,51 @@
             state.key = defaultKey;
             state.dir = 1;
             setSortIndicator(table, defaultTh, 'ascending', 'asc');
```
jdaugherty, line 108: The comment above still says "(and sort to match)" but the `sortTableBy` call was removed. The removal itself looks right — the GSP already renders `pluginsWithOrder` sorted by name ascending, so the initial sort was redundant — but the comment should be updated to say the DOM order already matches the indicator, otherwise the next reader will wonder whether the call was dropped by accident.

grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
```diff
@@ -26,33 +51,42 @@
                         <i class="bi bi-globe me-1"></i>${currentLocale.getDisplayName(currentLocale)}
                     </a>
                     <ul class="dropdown-menu dropdown-menu-end" aria-labelledby="localeDropdown">
+                        <%-- The default language stays pinned on top: a user who switched to a
+                             language they cannot read must always find a recognizable way back. --%>
+                        <g:set var="defaultLocale" value="${java.util.Locale.ENGLISH}"/>
```
jdaugherty: The pinned "way back" language is hardcoded to `Locale.ENGLISH`. That matches the template's defaults today, but if an app configures a different default locale the pinned entry no longer reflects it. Consider resolving the application's configured default locale (falling back to English) so the pin stays truthful.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
```diff
@@ -332,8 +379,40 @@
...
+                                <h6 class="card-title mb-0 fw-semibold"><g:message code="welcome.actuators"/></h6>
+                                <span class="badge bg-body-tertiary text-body border">
+                                    ${actuatorEndpoints.size()}
+                                </span>
+                            </div>
+                            <ul class="list-group list-group-flush small">
+                                <g:each var="endpoint" in="${actuatorEndpoints}">
+                                    <li class="list-group-item d-flex justify-content-between align-items-center px-0">
+                                        <span class="fw-medium">${endpoint.endpointId}</span>
+                                        <a href="${request.contextPath}${actuatorBasePath}/${endpoint.rootPath}" target="_blank" rel="noopener"
```
jdaugherty: These links assume the actuator endpoints are served on the application port. With `management.server.port` set, the card still renders but every link 404s on the app port. Probably fine for a dev welcome page, but worth either noting or skipping the card when a separate management port is configured.

grails-forge/grails-forge-core/src/main/resources/i18n/messages.properties
```diff
@@ -54,3 +54,49 @@ typeMismatch.java.lang.Short=Property {0} must be a valid number
...
+layout.docs.title=Documentation
+layout.docs.text=Ready to dig in? You can find in-depth documentation for all the features of Grails in the User Guide.
+layout.community.title=Join the Community
+layout.community.text=Get feedback and share your experience with other Grails developers on the Grails mailing lists, GitHub discussions, or the community Slack channel.
+layout.theme.toggle=Toggle theme
+layout.theme.light=Light
+layout.theme.dark=Dark
+layout.theme.auto=Auto
+layout.loading=Loading...
+welcome.actuators=Actuators
```
jdaugherty, line 103: `welcome.actuators` stays "Actuators" in every bundle (zh_CN uses "Actuator"). If that's deliberate — treating it as the Spring Boot product term — a short comment in the bundles would prevent future "missing translation" reports; otherwise it's the one card title the language selector doesn't translate.

grails-forge/grails-forge-core/src/main/resources/i18n/messages_sk.properties
codeconsole: Dropped the shadowed line in e03e68c5ff — kept `nezhodný typ` since as the last entry it was the value actually in effect. Removed from both the forge bundle and the profiles mirror, and the integrity spec now asserts no bundle declares the same key twice (parsing raw lines, since `Properties` collapses duplicates).

grails-forge/grails-forge-core/src/main/resources/assets/javascripts/welcome.js
codeconsole: Updated in e03e68c5ff — the comment now says the server already renders rows in that order, so the indicator-only initialization reads as intentional.

grails-forge/grails-forge-core/src/main/resources/gsp/main.gsp
codeconsole: Done in e03e68c5ff — the pin now resolves `spring.web.locale` via `StringUtils.parseLocale` and falls back to `Locale.ENGLISH` when unset, so a configured default locale is what stays pinned.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
codeconsole: Good catch — in e03e68c5ff the links now build against `management.server.port` (plus `management.server.base-path`) when a separate management port is configured, so the card stays useful instead of 404ing. Same-port setups keep using the app's context path.

grails-forge/grails-forge-core/src/main/resources/i18n/messages.properties
codeconsole: It's deliberate — 'Actuators' is the Spring Boot product name (zh_CN's "Actuator" included). Added a comment above the key in every bundle in e03e68c5ff so it doesn't get reported as a missing translation.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
```diff
@@ -13,32 +14,38 @@
...
     <div class="container-lg py-2 py-md-3">
-        <div class="row align-items-top g-4">
 
-            <%-- WELCOME MESSAGE --%>
-            <div class="col-12 col-md-7">
-                <h1 class="display-6 fw-semibold mb-2">Welcome to Grails</h1>
-                <p class="lead text-body-secondary">
-                    Congratulations, you have successfully started a Grails application.
+        <%-- WELCOME HERO --%>
+        <section class="welcome-hero position-relative overflow-hidden rounded-3 p-4 p-md-5 mb-4">
```
jdaugherty, line 26: On the design front, we'll probably want more feedback.  For now, since we need this to get the 8.0.0-M3 out, I'm ok merging.  We will probably have to readdress if anyone disagrees.

jdaugherty:
As discussed, this is required for the 8.0.0-M3 stage.  @jamesfredley @matrei  please review the design changes later and let's re-address any concerns in a follow-up PR.

PR #15976, opened 2026-07-12
Include the charset when DefaultHtmlRenderer sets the content type

codeconsole:
`respond()` rendering HTML sets the response content type to the bare mime type name (`text/html`) — `DefaultHtmlRenderer` is the only renderer that skips `GrailsWebUtil.getContentType`; its JSON and XML siblings both append the charset.

### Effect

A bare `text/html` leaves the response charset to the servlet container (ISO-8859-1 on Tomcat). GSP only applies its configured `text/html;charset=UTF-8` when the response has no content type yet, so it backs off — the response writer then encodes Latin-1 and **every non-Latin-1 character on the page is written as a literal `?`** (Czech, Cyrillic, Thai, CJK; Latin-1-mappable accents survive, which makes the corruption easy to miss on Western-European pages).

### Why it went unnoticed

- It only triggers when the `Accept` header negotiates a concrete HTML mime type. Browsers always send one; curl and most HTTP test clients send `Accept: */*`, which negotiates `MimeType.ALL`, skips the content-type stamp, and gets a correct UTF-8 page.
- The SiteMesh 2 filter historically masked it by unconditionally forcing UTF-8 onto decorated responses. The SiteMesh 3 view-resolver integration (#15975) deliberately respects the content type the application sets — exposing this latent bug on every scaffolded (`respond`-rendered) page.

### Fix

Mirror `DefaultJsonRenderer`: an `encoding` property defaulting to `GrailsWebUtil.DEFAULT_ENCODING` (UTF-8), applied via `GrailsWebUtil.getContentType(mimeType.name, encoding)`.

### Verification

- New spec pins `text/html;charset=UTF-8` on the negotiated render path (fails against the old code); `:grails-rest-transforms:test` and `:grails-controllers:test` green.
- Live: a generated application's scaffolded page requested with a real browser `Accept` header serves all 18 locales `charset=UTF-8`, strict byte-validated (`slovenčina`, `čeština`, `русский`, `中文` intact); in Chrome, `document.characterSet` is now `UTF-8` and the locale dropdown renders every native name correctly.

Related: #15974, #15975.

### DIFF
--- a/grails-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/grails-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -29,6 +29,12 @@
       "type": "java.util.List<grails.boot.GrailsBanner.DefaultVersionOption>"
     },
     {
+      "name": "grails.converters.encoding",
+      "description": "The character encoding used for converter output and for the content type of responses rendered via respond() and render(), e.g. text/html;charset=UTF-8.",
+      "type": "java.lang.String",
+      "defaultValue": "UTF-8"
+    },
+    {
       "name": "grails.banner.versions.order",
       "description": "A sorted list of versions to put first in the banner versions.",
       "type": "java.util.List<grails.boot.GrailsBanner.VersionOption>"
--- a/grails-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/grails-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -29,12 +29,6 @@
       "type": "java.util.List<grails.boot.GrailsBanner.DefaultVersionOption>"
     },
     {
-      "name": "grails.converters.encoding",
-      "description": "The character encoding used for converter output and for the content type of responses rendered via respond() and render(), e.g. text/html;charset=UTF-8.",
-      "type": "java.lang.String",
-      "defaultValue": "UTF-8"
-    },
-    {
       "name": "grails.banner.versions.order",
       "description": "A sorted list of versions to put first in the banner versions.",
       "type": "java.util.List<grails.boot.GrailsBanner.VersionOption>"
--- a/grails-rest-transforms/src/main/groovy/org/grails/plugins/web/rest/render/html/DefaultHtmlRenderer.groovy
+++ b/grails-rest-transforms/src/main/groovy/org/grails/plugins/web/rest/render/html/DefaultHtmlRenderer.groovy
@@ -28,6 +28,7 @@ import grails.core.support.proxy.ProxyHandler
 import grails.rest.render.RenderContext
 import grails.rest.render.Renderer
 import grails.util.GrailsNameUtils
+import grails.util.GrailsWebUtil
 import grails.web.mime.MimeType
 
 /**
@@ -47,6 +48,8 @@ class DefaultHtmlRenderer<T> implements Renderer<T> {
 
     String suffix = ''
 
+    String encoding = GrailsWebUtil.DEFAULT_ENCODING
+
     DefaultHtmlRenderer(Class<T> targetType) {
         this.targetType = targetType
     }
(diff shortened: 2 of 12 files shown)


grails-rest-transforms/src/main/groovy/org/grails/plugins/web/rest/render/html/DefaultHtmlRenderer.groovy
```diff
@@ -47,6 +48,8 @@ class DefaultHtmlRenderer<T> implements Renderer<T> {
 
     String suffix = ''
 
+    String encoding = GrailsWebUtil.DEFAULT_ENCODING
```
jdaugherty, line 53: Shouldn't this be configurable and then this assigned to default if not set?

codeconsole: Good call — done in 814b384df6. The encoding is now wired from configuration: `DefaultRendererRegistry` gained an `encoding` property that it propagates to every default renderer it creates (HTML, JSON, XML, and the `Errors` container renderers, plus the fallback HTML renderers the JSON/XML renderers construct when no view matches), and `RestResponderGrailsPlugin` populates it from the existing `grails.converters.encoding` setting, falling back to UTF-8 when unset. I reused that key rather than introducing a new one since these renderers are part of the same `respond()` pipeline the converters serve — one knob for response encoding. The setting is also documented in `additional-spring-configuration-metadata.json` so IDEs offer completion for it. Renderers registered as beans keep the independently settable `encoding` property. Covered by new specs at renderer level (custom encoding reaches the content type) and registry level (propagation to all default renderers).

jdaugherty: Shouldn't we default this to the configured value if set, otherwise use this default thought?

codeconsole:
Good call — moved the resolution into the classes themselves in bb78ee0197. `DefaultRendererRegistry` and every renderer with an `encoding` property (`DefaultHtmlRenderer`/`Json`/`Xml`, `AbstractLinkingRenderer`, `AbstractVndErrorRenderer`) now carry `@Value('${grails.converters.encoding:UTF-8}')`, and the explicit wiring in `RestResponderGrailsPlugin` is gone. That way any renderer registered as a bean — including user-registered HAL/Atom/Vnd.Error renderers — honors the config, not just the registry bean the plugin defines. The field default stays UTF-8 for instances constructed outside Spring, and the registry still pushes its encoding onto the non-bean defaults it creates. Added a spec that boots a context with `grails.converters.encoding=ISO-8859-1` and verifies both the registry and its renderers resolve it from the environment.

grails-rest-transforms/src/main/groovy/org/grails/plugins/web/rest/render/html/DefaultHtmlRenderer.groovy
codeconsole: Done in bb78ee0197 — the property is now annotated `@Value('${grails.converters.encoding:UTF-8}')` so it resolves the configured value when set and falls back to UTF-8 otherwise, on the registry and all the renderer classes. The plugin-side wiring is removed.

grails-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -28,6 +28,12 @@
       "description": "A list of versions to exclude from the banner.",
       "type": "java.util.List<grails.boot.GrailsBanner.DefaultVersionOption>"
     },
+    {
+      "name": "grails.converters.encoding",
```
jdaugherty: `grails.converters.encoding` is already declared in `grails-web-core/src/main/resources/META-INF/spring-configuration-metadata.json` (line ~211) with the narrower description "The character encoding for converter output (JSON or XML)." — this adds a second declaration of the same property with a different description, so IDEs will surface duplicate/conflicting entries. Suggest updating the existing grails-web-core entry's description in place and dropping this addition. The user-facing doc table (`grails-doc/src/en/ref/Configuration/Application Properties.adoc`, ~line 289) carries the same stale JSON/XML-only wording and should be broadened too, since this PR extends the key's effect to respond()-rendered content types. Also, if the entry stays here, it's currently inserted between the two `grails.banner.versions.*` entries — worth moving out of that group.

codeconsole: Fixed in 0a45f47908 — dropped the second declaration from grails-core and broadened the existing grails-web-core entry in place to cover the respond() renderer content types (which also resolves the banner-group placement nit). One correction on the doc pointer: `ref/Configuration/Application Properties.adoc` doesn't exist on 8.0.x and `grails.converters.encoding` wasn't documented anywhere in grails-doc, so instead of editing stale wording I added the setting to the runtime settings list in `guide/conf/config/builtInOptions.adoc` with the broadened description.

jdaugherty (approved):

PR #15984, opened 2026-07-14
Simplify forge deployments to speed up release times

jdaugherty:
@jamesfredley we spoke about simplifying the forge workflows in one of the previous weekly meetings, this PR does that so releases can be quicker.

### DIFF
--- a/.github/workflows/forge-deploy-next.yml
+++ b/.github/workflows/forge-deploy-next.yml
@@ -16,34 +16,13 @@
 name: "Forge - Next GCP Deploy"
 on:
   workflow_dispatch:
+# The application code is already built and tested by the main CI (gradle.yml)
+# and by release verification, so these deploy workflows only build the native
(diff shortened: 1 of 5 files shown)


.github/workflows/forge-deploy-next.yml
matrei: Do we need TestLens here?

matrei: TestLens?

.github/workflows/forge-deploy-prev-snapshot.yml
matrei: TestLens?

matrei: TestLens?

matrei (approved):
If we are not running tests, do we need TestLens?

.github/workflows/forge-deploy-next.yml
jdaugherty: Nope, i'll get that removed.

PR #15987, opened 2026-07-16
8.X UI OVERHAUL: Enhance the create-app welcome page, update g:localeSelect, and spring-security core/UI, and bump jQuery to 4.0.0

codeconsole:
## Summary

Create-app welcome page and `g:localeSelect` enhancements, a jQuery 4.0.0 bump, a Bootstrap rewrite of the spring-security login pages and spring-security-ui plugin, and i18n improvements — targeting `8.0.x`.

### 1. Bump the managed jQuery webjar to 4.0.0
`jquery.version` (`org.webjars.npm:jquery`) goes from `3.7.1` to `4.0.0` in both BOM maps in `dependencies.gradle`. The Forge and web-profile templates reference the webjar without an explicit version, so generated apps inherit `4.0.0` from the BOM automatically. Documented in `upgrading80x.adoc` (§35), including how to pin the previous version.

### 2. Welcome-page enhancements (Forge + profile templates)
- A **switchable artefact list**: the "Available Controllers" card title is now a dropdown that switches the card between Controllers, Domains, Services and Tag Libraries — per-type count badges, name filters and zero-states, with all four panels server-rendered and toggled client-side (no localized text in JS). The rows of the "Artefact counts" card now link to the card pre-switched to their type.
- A **runtime internals row**: a second switchable card lists the application context's `ApplicationListener`s (with the adapted `@EventListener` method as hover detail), the four data-binding bean types grouped under one Data Binding panel, and the `MimeTypeProvider`s; alongside it a sortable/filterable Mime Types table (Extension / Media type) and a conditional Datastores card showing each datastore's `mappingContext` event listeners (rendered only when GORM is on the classpath, resolved without a hard class reference). Three filter panels complete the card: **Servlet Filters** derives the request's effective filter pipeline from the rendering call stack (the true execution order, with the security chain's inner filters in position — something no registry can report), **Filter Registrations** lists the `FilterRegistrationBean`s by order value, and a **Security Filter Chain** panel (rendered only when Spring Security is present) groups each `FilterChainProxy` chain with its request matcher and inner filters. The card switcher is generalised to per-card scopes (`data-switch-*`) so multiple switchable cards stay independent.
- The controller listings respect **`allowedMethods`**: a controller whose default action cannot be requested with GET (e.g. the POST-only logout) is invoked through a form that submits with the allowed method — in both the navbar dropdown and the welcome card — with a method badge signalling the verb, so clicking Logout logs out instead of producing a 405. `LogoutController` declares its restriction through a computed `allowedMethods` getter that honors `logout.postOnly`.
- A **"jump to id"** control on controllers that expose a `show` action, navigating to `/controller/show/{id}` and progressively enhancing a plain `GET` fallback.
- A **navbar controllers filter** that appears once the controller list is long enough, with a collapsible navbar toggler.
- Fixed the layout's spinner (its `ajaxStart`/`ajaxStop` handlers were bound to `#spinner` rather than `document`, dead since jQuery 1.9) and gated the filter autofocus to hover/pointer devices so touch keyboards no longer cover the list.
- New i18n keys (`welcome.show.*`, `welcome.domains/services/taglibs.title`, `welcome.artefacts.*`, `welcome.listeners/binding/mime/datastores/filters/registrations/securitychain.*`, `layout.nav.toggle`) translated across all bundled locales.

### 3. Language selector: shared ordering and a more flexible `g:localeSelect`
- `AvailableLocaleResolver` now orders the available locales by their autonym using a fixed ROOT `Collator` (case-insensitive and identical in every UI language), so every consumer inherits a correct, stable order.
- `g:localeSelect` gains `labelType` / `sort` / `tags` for its `<select>`, a framework-agnostic `type="links"` render, and a **body form** (`var`) that exposes a per-locale model (`locale`, `tag`, `autonym`, `name`, `active`, `default`, `index`, …) so a Bootstrap dropdown or any custom switcher can reuse the tag's resolution, ordering and active/default detection without duplicating it. The default no-body `<select>` output is unchanged.
- The create-app welcome page now uses the body form in place of hand-rolled logic. The tag reference and the changing-locales guide are updated, with rendering tests added.

### 4. Spring-security plugin: Bootstrap login/denied pages and locale parity
- The plugin's login page (`login/auth.gsp`) drops its embedded custom CSS for plain Bootstrap 5 markup rendered through the host layout — a centered card with `form-control` fields, an input-group password show/hide toggle, a `form-check` remember-me and a full-width primary button. The denied page becomes a Bootstrap `alert-danger`.
- The login failure message is now rendered with `<g:flashMessages/>` — whose `_flashRendered` guard keeps a host layout's own `<g:flashMessages/>` from displaying the same message twice — and is reported via `flash.error` instead of `flash.message`, so it is styled as an error rather than a success.
- The plugin's `spring-security-core` message bundle is brought to locale parity with the create-app bundles: the eight missing locales are added (cs, da, nl, pl, pt_PT, sk, th, zh_CN), the `springSecurity.errors.login.max.sessions.exceeded` message is supplied to the nine locales that lacked it, a French typo (`verrouillé`) and German formal address (`Ihres`) are corrected, and a translator-attribution comment is removed from the pt_BR bundle.
- A **Traditional Chinese (`zh_TW`)** bundle — which the security plugin already shipped — is added to the create-app welcome bundles (Forge + base profile skeleton) at full key parity with the other locales, using Taiwan vocabulary.

### 5. spring-security-ui plugin: rewritten on jQuery 4 and Bootstrap 5
- The plugin's 2015-era client stack — jQuery 2.1.4, jQuery UI 1.10, jquery-form, jGrowl, DataTables, bgiframe and the bundled jdMenu/positionBy plugins — is replaced with the BOM-managed jQuery 4.0.0 and Bootstrap 5.3 webjars, referenced through version-wildcard asset manifests (`webjars/jquery/%/dist/jquery.js`) so they track the BOM without hardcoded versions. Every bundled stylesheet and image is deleted; there is no custom CSS anywhere.
- Replacements map one-to-one onto Bootstrap: the jdMenu menubar becomes a navbar with dropdowns, the jQuery UI ajax-login and delete-confirmation dialogs become modals (the login submitted with `jQuery.ajax` + `serialize()` in place of jquery-form), jQuery UI tabs become nav-tabs, jGrowl notifications delegate to `g:flashMessages`, and the jQuery UI autocomplete search becomes a native `datalist` fed by `fetch` from the existing `ajaxSearch` endpoints. The taglib emits Bootstrap form groups, cards and striped tables, so all 44 views restyle largely through it. The unused `Constants` class (which pinned the old jQuery versions) and the now-stale MIT license references are removed. The migration is documented in `upgrading80x.adoc` (§36).
- The plugin's screens now render **through the host application's own layout**: the `springSecurityUI` and `register` layouts are thin sub-layouts that `g:applyLayout` the app layout (`grails.plugin.springsecurity.ui.gsp.parentLayout`, default `main`), so plugin pages inherit the app's theme — including Bootstrap dark mode — locale selector, branding and assets. The security menu and login state are contributed to the host navbar through `nav`/`navActions` content blocks, which the create-app layout renders via new generic `<g:pageProperty name="page.nav"/>` and `<g:pageProperty name="page.navActions"/>` extension points that any page or plugin can use. The plugin loads no jQuery, Bootstrap or CSS of its own; its scripts register on `DOMContentLoaded` so they run after the host's end-of-body scripts, and `s2ui:title` delegates to the `grailsLayout` capture tags so page titles survive nested decoration.
- Fixes latent tag-dispatch issues that left several of these pages broken on Grails 7+ regardless of styling: tag-as-method calls with string bodies or missing closure arguments (`hasErrors`, `set`, `fieldError`) recursed `methodMissing` into a `StackOverflowError`, and unqualified tag calls interpolated into GStrings wrote directly to the output writer instead of returning markup, rendering field rows out of order with stray nulls. Tags are now invoked through the `g.` namespace or replaced with direct implementations, and the ajax-login modal form no longer duplicates the standalone login page's form element id.

## Notes
- jQuery 4.0.0 is a major release; the generated welcome page uses jQuery only for the small spinner/fade behaviour (the new controls are vanilla JS).
- The Forge and profile welcome-page templates are kept identical to each other.
- The oauth2 plugin's create-or-link account page and the `ChallengeQuestions*.gsp.template` files used by `s2ui-create-challenge-questions` are converted to the same Bootstrap markup; the registration mail partials keep their inline styles, which HTML email requires.


dependencies.gradle
```diff
@@ -35,7 +35,7 @@ ext {
             'javaparser-core.version'       : '3.28.2',
             'jline.version'                 : '3.30.6',
             'jna.version'                   : '5.19.1',
-            'jquery.version'                : '3.7.1',
+            'jquery.version'                : '4.0.0',
```
jdaugherty, line 38: jQuery 4.0.0 is a breaking major (drops IE and other legacy browsers, removes long-deprecated utilities like `$.trim`, `$.type`, `$.proxy`, `$.isFunction`, etc.), and because it's managed in both BOM maps, it isn't just the freshly-generated welcome page that moves — any existing app that references the jquery webjar without an explicit version will silently jump 3.7.1 → 4.0.0 when it upgrades to the Grails 8 BOM, and *their* JavaScript may well use the removed APIs.

The templates themselves are fine (I checked the 4.0.0 webjar: `dist/jquery.js` is at the same path so the `%` manifest reference still resolves, and the ajax event alias methods used in `application.js` are still present in the full build). But this needs a line in `upgrading80x.adoc`: note the major bump, link the jQuery 4 upgrade guide, and mention that apps relying on removed APIs can pin `org.webjars.npm:jquery:3.7.1` in their own build until they migrate. Per the repo guidelines, user-facing changes like this shouldn't merge without doc coverage.

grails-profiles/web/skeleton/grails-app/assets/javascripts/application.js
```diff
@@ -17,4 +17,40 @@ if (typeof jQuery !== 'undefined') {
             $(this).fadeOut();
         });
     })(jQuery);
-}
\ No newline at end of file
+}
```
jdaugherty, line 22: Not introduced here, but since this PR both bumps jQuery and touches this file: the `$('#spinner').ajaxStart(...)` block just above has been dead code since jQuery 1.9 — global ajax events fire only on `document`, so a handler bound to `#spinner` never runs and the spinner in the layout never shows. If we're keeping it, the working form is:

```js
$(document)
    .on('ajaxStart', () => $('#spinner').fadeIn())
    .on('ajaxStop', () => $('#spinner').fadeOut());
```

(Behavior unchanged in 4.0.0 — I checked the shipped `dist/jquery.js`.) Applies to the forge copy of this file too; worth fixing in both while they're being kept in lockstep, or deleting the block outright if the spinner isn't worth keeping.

grails-profiles/web/skeleton/grails-app/assets/javascripts/application.js
```diff
@@ -17,4 +17,40 @@ if (typeof jQuery !== 'undefined') {
...
+
+    document.addEventListener('DOMContentLoaded', () => {
+        document.querySelectorAll('.nav-filter-input').forEach((input) => {
+            input.addEventListener('input', () => applyNavFilter(input));
+            input.addEventListener('search', () => applyNavFilter(input));
+
+            const dropdown = input.closest('.dropdown');
+            if (!dropdown) return;
+            // Focus the field as the menu opens; clear it once the menu closes.
+            dropdown.addEventListener('shown.bs.dropdown', () => input.focus());
```
jdaugherty: Focusing the filter as the menu opens is right on desktop, but on touch devices it pops the on-screen keyboard the moment the Controllers menu opens, covering half the list the user was about to scan. Suggest gating the autofocus to pointer/hover environments:

```js
dropdown.addEventListener('shown.bs.dropdown', () => {
    if (window.matchMedia('(hover: hover)').matches) input.focus();
});
```

Touch users can still tap the field (Bootstrap 5.3's dropdown exempts inputs inside the menu from auto-close, so that works fine).

jdaugherty (commented):
Reviewed both halves of this. The welcome-page work itself looks solid — I verified the forge templates and profile skeletons are byte-identical as claimed, the new i18n keys are present in all bundled locales (and genuinely translated, not English copies), `GrailsControllerClass.getActions()` backs the `c.actions?.contains('show')` check, and the no-JS fallback (`GET /controller/show?id=…`) works with the default URL mapping.

On the jQuery 4.0.0 bump I checked the actual artifact on Maven Central: the webjar exists, `dist/jquery.js` is still at the same path so the version-agnostic `//= require webjars/jquery/%/dist/jquery.js` manifest reference keeps resolving, and the full build still ships the `.ajaxStart()`/`.ajaxStop()` alias methods that `application.js` calls — so the generated templates themselves survive the major bump. Also confirmed Bootstrap 5.3.8's dropdown explicitly exempts `input`/`textarea`/`form` targets inside the menu from auto-close, so the sticky filter field works with the default `autoClose` — no `data-bs-auto-close` needed.

Remaining items inline — main one is an upgrade-notes ask for the BOM bump.

dependencies.gradle
codeconsole: Added as section 35 in `upgrading80x.adoc` — the major bump and the removed utilities, a link to the jQuery 4.0 upgrade guide, and the pin-back for apps that aren't ready to migrate.

One adjustment to the suggested workaround: a plain `implementation 'org.webjars.npm:jquery:3.7.1'` doesn't actually downgrade, because the auto-applied Grails BOM is a non-enforced `platform()` and its 4.0.0 wins under Gradle's highest-version-wins resolution. The note recommends the BOM property override `jquery.version=3.7.1` instead, which is applied as a strict constraint that correctly downgrades (consistent with section 14). ea78cf8

grails-profiles/web/skeleton/grails-app/assets/javascripts/application.js
codeconsole: Good catch. Fixed rather than deleted, since `#spinner` is present in the layout — rebound the global `ajaxStart`/`ajaxStop` handlers to `document` (your suggested form) so the spinner actually shows. Applied to both the Forge and profile copies. 0e28a4d

grails-profiles/web/skeleton/grails-app/assets/javascripts/application.js
codeconsole: Gated the on-open autofocus to `window.matchMedia('(hover: hover)').matches` in both copies, so touch devices no longer pop the on-screen keyboard over the list. 0e28a4d

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -1601,3 +1601,22 @@ Custom field renderers should go through the documented `f:field`, `f:widget`, a
...
+
+The jQuery webjar managed by the `grails-bom` (`org.webjars.npm:jquery`) moves from 3.7.1 to 4.0.0.
+jQuery 4 is a major release: it drops Internet Explorer and other legacy browsers and removes long-deprecated utilities such as `jQuery.trim`, `jQuery.type`, `jQuery.isArray`, `jQuery.isFunction`, `jQuery.proxy`, and `jQuery.isWindow`.
+See the https://jquery.com/upgrade-guide/4.0/[jQuery 4.0 Upgrade Guide] for the complete list of changes.
+
+Because the version is managed by the BOM, any application that references the jQuery webjar without an explicit version picks up 4.0.0 automatically once it adopts the Grails 8 BOM.
+This includes the generated create-app welcome page, whose assets have been verified against 4.0.0.
+
+If your application's own JavaScript relies on the removed APIs, migrate those call sites.
+If you are not ready to migrate, pin the previous version by overriding the BOM-managed property (see section 14).
```
jdaugherty: I would keep this simple.  Replacing lines 1615 & 1616 with: 

If you are not ready to migrate, pin the previous version in your gradle build.

jdaugherty (approved):
It would be nice if we could update the spring-security-ui & spring-security-core projects so they rely on jquery too.  I understand this is a stretch though.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Done — trimmed to your wording. Kept the `gradle.properties` override snippet since that's the form that actually downgrades under the non-enforced BOM platform. 3af80d7

jdaugherty:
@codeconsole Looks like the forge tests are legitimately broken.

jdaugherty:
@codeconsole Let me know when it's good to re-review this and I'll take a second pass.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
```diff
@@ -963,39 +965,145 @@ class FormTagLib implements ApplicationContextAware, InitializingBean, TagLibrar
...
+            boolean exactActiveMatch = locales.any { it.language == current.language && it.country == current.country }
+            locales.eachWithIndex { locale, i ->
+                out << body([(varName): [
+                        locale: locale,
+                        tag: locale.toLanguageTag(),
+                        code: localeKey(locale),
+                        autonym: locale.getDisplayName(locale),
+                        name: locale.getDisplayName(current),
+                        label: label(locale),
+                        active: exactActiveMatch ?
```
jdaugherty: When the list offers no exact country match, this fallback marks *every* locale sharing the language active, not just the language-only entry the comment describes. The create-app bundles ship two country variants of the same language with no bare-language bundle (`pt_BR`/`pt_PT`, `zh_CN`/`zh_TW`), so a request locale of bare `pt`, bare `zh`, or an unlisted variant like `zh-HK` flags both variants active — and the welcome dropdown then highlights two rows at once. `LocaleSelectRenderingSpec` exercises the fallback only against single-variant lists (`[en, it]`), so this case is uncovered. Consider limiting the fallback to a single winner: the language-only entry when the list has one, otherwise the first language match.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
```diff
@@ -963,39 +965,145 @@ class FormTagLib implements ApplicationContextAware, InitializingBean, TagLibrar
...
+        }
+
+        String varName = attrs.remove('var')
+        if (varName) {
+            boolean pinDefault = Boolean.valueOf(attrs.remove('pinDefault')?.toString())
+            Locale defaultLocale = configuredDefaultLocale()
+            if (pinDefault) {
+                Locale pinned = locales.find { it.language == defaultLocale.language }
+                if (pinned) {
+                    locales = [pinned] + locales.findAll { it.language != defaultLocale.language }
```
jdaugherty: `findAll { it.language != defaultLocale.language }` removes every locale that shares the default's *language*, not just the pinned entry — and the pin itself matches by language only. With the bundle set the create-app ships, configuring `grails.i18n.default.locale=pt_BR` pins whichever `pt` variant the list yields first and silently drops the other (`pt_PT`) from the selector entirely; the `zh_CN`/`zh_TW` pair behaves the same. Pinning the exact locale (with a language-only fallback) and keeping the remainder — `locales = [pinned] + (locales - pinned)` — preserves every entry while still emitting the default first.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
```diff
@@ -281,26 +371,784 @@
...
+                                   if (prev != null && cls.isAssignableFrom(prev)) { return acc }
+                                   acc << cls
+                               }
+                               .unique()}"/>
+                <g:set var="filterRegistrations"
+                       value="${applicationContext.getBeansOfType(org.springframework.boot.web.servlet.FilterRegistrationBean)
+                               .entrySet().toList().sort { it.value.order }}"/>
+                <g:set var="filterChainProxyType"
+                       value="${ClassUtils.isPresent('org.springframework.security.web.FilterChainProxy', null) ? ClassUtils.forName('org.springframework.security.web.FilterChainProxy', null) : null}"/>
+                <g:set var="securityFilterChains"
```
jdaugherty, line 599: The runtime-internals surface renders in every environment — `Environment` is referenced on this page only for display. That includes each security filter chain's request matcher (here), the effective servlet filter pipeline, listener and internal bean names, and the server hostname/OS further up. The previous welcome page also rendered in production, but it exposed far less than this. Since this is the generated app's default `/` view and frequently survives into production unchanged, consider gating the internals cards (or the whole diagnostic block) on `Environment.current == Environment.DEVELOPMENT`. The same applies to the byte-identical copy at `grails-profiles/web/skeleton/grails-app/views/index.gsp`.

grails-spring-security/ui/plugin/grails-app/taglib/grails/plugin/springsecurity/ui/SecurityUiTagLib.groovy
```diff
@@ -705,59 +700,59 @@ class SecurityUiTagLib {
         }
 
         def out = getOut()
-        out << """<a id="$elementId" """
+        out << """<button type="submit" id="$elementId" class="btn btn-primary" """
```
jdaugherty: `$elementId`, `$text` and the remaining attributes appended by `writeRemainingAttributes` (line 885) are interpolated into the markup without HTML/attribute encoding. Every current caller passes fixed ids and message-bundle text, so nothing is exploitable today, but the rewrite funnels all `s2ui:submitButton`/`s2ui:linkButton` output through this path — a future caller passing user-derived `text` or attribute values would inject markup. While these tags are being reworked anyway, encoding the interpolated values (`encodeAsHTML()`) would close that door.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
codeconsole: Good catch — fixed in 47e8ebe933. `active` (and `default`) now elect a single winner per language via the exact match → bare-language entry → first language variant chain you suggested, so a bare `pt` or unlisted `zh-HK` request highlights exactly one row. Added a regression test using the shipped-bundle shape (`pt_BR`/`pt_PT` with no bare `pt`).

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
codeconsole: Fixed in 47e8ebe933 exactly as suggested: the pin resolves the exact configured variant (with a language-only fallback) and reorders with `[pinned] + remainder`, so `grails.i18n.default.locale=pt_BR` pins `pt_BR` while `pt_PT` stays in the selector. The `default` flag elects the same single entry. Covered by the new `LocaleSelectPinnedVariantSpec`.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
codeconsole: Agreed — as of 47e8ebe933 the whole diagnostic block renders only when `Environment.current == Environment.DEVELOPMENT`: the runtime-internals card (listeners, binding, mime providers, the filter pipeline, registrations, security chains and the URL-mappings resolver) plus the Mime Types and Datastores cards, with their `g:set` computations inside the gate so none of the introspection even executes in production. Applied to both template copies; the artefact/plugins/actuator cards that the previous welcome page also showed remain ungated.

grails-spring-security/ui/plugin/grails-app/taglib/grails/plugin/springsecurity/ui/SecurityUiTagLib.groovy
codeconsole: Fixed in 47e8ebe933 — `submitButton`/`linkButton` now encode the element id and text, and `writeRemainingAttributes` encodes every passed-through attribute key and value (the stale `TODO encodeAsHTML` is gone). Added `SecurityUiTagLibSpec` feeding markup-sensitive values through both tags to pin the encoding.

jdaugherty (approved):

jdaugherty:
I verified the fixes & checked out the new functionality.  seems good enough for our milestone.  We can revisit after the milestone if anyone objects.

PR #15990, opened 2026-07-17
Replace local StringBuffer usage with StringBuilder

jamesfredley:
## Summary

- replace invocation-local StringBuffer instances with StringBuilder
- preserve output and public APIs while removing unnecessary synchronization
- keep the starter intentionally limited to five safe local conversions

## Verification

- :grails-web-url-mappings:test
- :grails-web-common:test
- :grails-datamapping-core:test
- :grails-gsp-core:test
- :grails-gradle-model:test

This is an AI-generated starting point for the broader legacy JDK type cleanup.


davydotcom (approved):
Most of these calls are not used in a threaded manner so StringBuilder is indeed the smarter and faster solution here. This could have a benefit to overall GSP render performance (though small, it is a worth it low risk improvement)

jdaugherty (approved):

PR #15991, opened 2026-07-17
Enable CodeNarc EmptyCatchBlock checks

jamesfredley:
## Summary

- enable CodeNarc EmptyCatchBlock at priority 3
- mark each reviewed intentional empty catch explicitly with the supported ignored parameter convention
- add concise context or debug logging where silent fallback was not self-explanatory
- retain normal priority thresholds without global violation budgets or file exclusions

## Verification

- full root test suite: 2,990 tests, 0 failures, 54 skipped
- aggregateViolations: zero EmptyCatchBlock findings
- git diff --check

This is an AI-generated starting point for stricter empty-catch enforcement.


jdaugherty:
I'm interested to know what @matrei thinks about this. I'm good if he is.

matrei (approved):

matrei:
> I'm interested to know what @matrei thinks about this. I'm good if he is.

Yes, I like this. Maybe we can also remove the redundant comments that do not add any info, but just says `// ignore`?

jdaugherty (approved):
I'm with matrei on removing the redundant comments.  Let's do that and then this can be merged?

PR #15993, opened 2026-07-17
Use SLF4J for runtime diagnostics

jamesfredley:
## Summary

- replace direct stack traces and console output with structured SLF4J diagnostics across selected runtime paths
- preserve stderr visibility when logging is disabled or no provider is installed
- preserve historical logger categories and command return contracts
- add focused provider-enabled, API-only, lifecycle, malformed-input, and schema-export regression coverage

## Verification

- :grails-bootstrap:test
- :grails-core:test
- :grails-shell-cli:test
- :grails-data-hibernate5:test
- :grails-data-hibernate7:test
- grails-forge :grails-forge-core:test
- git diff --check

This is an AI-generated starting point for the broader runtime logging cleanup.


jdaugherty:
It's good to see this can work - I remember reading comments in the code base that historically this couldn't work due to some edge case.  I'll see if I can dig up the history to make sure this is indeed safe now and then review.

borinquenkid:
@jdaugherty — some findings that might help with your investigation into whether this can safely work.

I searched git history broadly for the historical comment/reason you're remembering (grepped commit messages and blamed the touched files - `GrailsConsole.java`, `SpringApplicationRunner.java`, `GrailsCli.groovy`) and couldn't turn up a specific citation. If you find it, I'd like to see it too.

What I did instead: identified the most plausible place the "edge case" could bite, and tested it directly. `SpringApplicationRunner.RunThread` deliberately runs with the *compiled application's own classloader* as its context classloader (its own Javadoc: "Thread used to launch the Spring Application with the correct context classloader"). SLF4J's provider discovery is classloader-sensitive, so a logger call whose *first-ever* SLF4J touch in the JVM happens from inside that thread is exactly the kind of scenario that could pick an ambiguous/wrong provider.

That doesn't happen here, structurally: `LOG` on `SpringApplicationRunner` is a `static final` field, so SLF4J's one-time provider binding is triggered when the class is first initialized - which happens when `new SpringApplicationRunner(...)` is constructed on the main CLI thread, before any `RunThread` with a foreign classloader is ever spawned. By the time `RunThread.run()` calls `logOrPrintStackTrace`, the logger is already resolved.

I verified this empirically rather than just reasoning about it: ran `SpringApplicationRunnerSpec` (6/6 pass), which forks a real JVM and exercises exactly this path - launch/reload/shutdown failures thrown from inside `RunThread` and logged via SLF4J. Also ran the full stated verification suite (`grails-bootstrap`, `grails-core`, `grails-shell-cli`, `grails-data-hibernate5`, `grails-data-hibernate7`, `grails-forge-core`) clean on a fresh `--rerun-tasks` run.

Separately, I confirmed and fixed the multi-SLF4J-provider classpath issue Copilot flagged (`grails-shell-cli`'s test classpath had both `slf4j-simple` and `logback-classic` on it simultaneously with no exclusivity) and rebased onto current `8.0.x`, which also surfaced a real semantic merge conflict: `ExecutionContext`/`SchemaExportCommand` moved packages upstream as part of unrelated work, and the two new `SchemaExportCommandSpec` files (added by this PR, so git had no history to rename-track through) needed their imports updated - now fixed.

This is evidence for your review, not a claim that the question is settled - happy to dig further if you want a different angle tested.

borinquenkid:
@jdaugherty following up on the classloader-binding question from July — wanted to loop back with the CI story since it's been a source of confusion on this PR.

**Status update:** I re-ran CI and Coverage against the current HEAD (`b8fd78d243`) via `workflow_dispatch`, and both came back fully green:
- CI: https://github.com/apache/grails-core/actions/runs/33202800706 (46/46 jobs, all platforms/JDKs)
- Coverage: https://github.com/apache/grails-core/actions/runs/33202803098

This matters because **no CI had actually run against the last two commits in three weeks** — the two `Merge branch '8.0.x'` commits (`9c1a834c15`, `b8fd78d243`) never triggered a workflow run, so the PR looked stale/red from the last real signal (commit `c75b73ec1e`, 2026-08-02, where all three `ProfilingGrailsPluginManagerSpec` tests failed on every OS/JDK).

That failure was a test-fragility issue, not a code defect: the original spec asserted on captured `System.err` text, which is sensitive to which SLF4J provider gets resolved in a given environment. The merge from `8.0.x` pulled in the `LogCapture` test fixture and rewrote the spec to assert on actual Logback events instead — I confirmed locally that all three tests pass reliably with that fix.

I also re-checked patch coverage against jacoco data for `ProfilingGrailsPluginManager.java` and `SpringApplicationRunner.java` (the two files Codecov flagged early on) — every line the diff touches is now covered; that earlier 13.9%-patch-coverage comment was against a since-superseded commit.

Ran the full verification suite from the PR description locally as well (`grails-bootstrap`, `grails-core`, `grails-shell-cli`, `grails-data-hibernate5`, `grails-data-hibernate7`, `grails-forge-core`) — all green.

Given CI is now actually green against HEAD and the open question from July has empirical evidence behind it, I'm approving.

borinquenkid (approved):
CI and Coverage re-verified green against current HEAD (see comment above) — approving.

PR #15994, opened 2026-07-17
Convert remaining plugin descriptors from doWithSpring() to beanRegistrar()

codeconsole:
### Summary

Continues the migration of in-tree framework plugins off the deprecated `doWithSpring()` bean DSL onto the Spring `BeanRegistrar` API (deprecated by #15934), following the SiteMesh 3 (#15964) and scaffolding (#15966) conversions. Descriptors are now statically compiled where the remaining lifecycle hooks allow it.

**17 descriptors converted:**

| Conversion style | Plugins |
|---|---|
| Straight registrar (config reads move to the Spring `Environment`; constructor/property refs become supplier-context lookups) | async, cache, converters, datasource, dbmigration (hibernate5 + 7), domain-class, events, gorm-graphql, fields, layout, markup views, json views, rest-transforms, url-mappings |
| Definitions post-processor (features `BeanRegistry` can't express) | controllers, interceptors, services, redis |

### Why some registrations need a post-processor

Per-artefact controller/interceptor/service beans autowire **by name** and use **arbitrary scopes**; the redis pool beans need a **`destroy` method** and **dynamic property values**. None of these can be expressed through the `BeanRegistry.Spec` API, so their definitions move to dedicated `BeanDefinitionRegistryPostProcessor`s registered through `beanRegistrar()`. They run at `PriorityOrdered` highest precedence so their definitions are visible to Spring Boot auto-configuration conditions — the same visibility the `doWithSpring()` drain gave them — and an existing definition for a bean name always wins, so applications (or other plugins) can still override the beans.

### Behavioural parity notes

- The interceptor handler adapter stays an **inner bean** of `grailsInterceptorMappedInterceptor`, as in the original DSL: inner beans are still fully autowired — the adapter's `@Autowired Interceptor[]` is injected — but are not returned by `getBeansOfType`, so `WebUtils.lookupHandlerInterceptors` does not collect the adapter a second time and interceptors do not run twice. Its definition (and the per-interceptor beans) is contributed by `InterceptorBeanDefinitionsPostProcessor`.
- The converters error-marshaller registerers reference the existing named marshaller beans instead of duplicating them as inner beans, preserving the `ApplicationContextAware` callback.
- `RedisConfigurationUtil` gains a `BeanDefinitionRegistry`-based `configureService` variant; the bean-builder closure form remains for existing consumers.
- `GrailsUnitTest.defineBeans(plugin)` now also applies a plugin's `beanRegistrar()`, so test setup that passes converted plugin instances (converters, JSON views) keeps working.
- The `chainedTransactionManagerPostProcessor` registration drops the unrefreshed-context `transactionManager` guard; the post-processor itself only acts when a `transactionManager` definition and multiple chainable managers exist.

The `plugin` and `web-plugin` profile skeletons and the plugin-developer guide now show `beanRegistrar()` instead of the deprecated hook.

### Overridable framework defaults moved to auto-configuration

Two conversions surfaced framework defaults that applications and plugins override *by bean name*. Because the plugin registration phase drains framework plugins before application plugins, keeping them as registrar-time registrations meant any such override triggered a bean-definition override (with Spring's startup notice). These defaults now register through auto-configuration guarded with `@ConditionalOnMissingBean` — the same treatment as the i18n and URL-mappings beans under "More Framework Beans Are Cleanly Overridable" — so a bean contributed by the application or another plugin simply wins and the default is never registered:

- **controllers**: `exceptionHandler` (`GrailsExceptionResolver` with the `Exception` → `/error` mapping) moved from the descriptor to the existing `ControllersAutoConfiguration`.
- **cache**: `grailsCacheManager` (config-driven selection via `grails.cache.cacheManager` preserved) and `customCacheKeyGenerator` moved from the descriptor to a new `GrailsCacheAutoConfiguration`, gated on `grails.cache.enabled` exactly like the registrar was. Cache-provider plugins replace `grailsCacheManager` by name, so this previously produced an override notice at every boot with a provider installed.

The upgrade guide's "More Framework Beans Are Cleanly Overridable" list now includes both.

### Still on `doWithSpring()` (deferred to follow-up PRs)

- `CoreGrailsPlugin` and `GroovyPagesGrailsPlugin` — abstract parent bean templates (`abstractGrailsResourceLocator`, `abstractViewResolver`) and an XML-namespace component-scan; the resolver-pipeline linchpin.
- The hibernate5/7, mongodb and neo4j plugins — their datastore `SpringInitializer`s return bean-DSL closures, so conversion means reworking the initializers.
- The spring-security suite — config-driven DSL throughout.

### Testing

- New `*GrailsPluginSpec` / `*PostProcessorSpec` for every converted descriptor, exercising the registrar through `BeanRegistryAdapter`.
- The auto-configured defaults are covered by new back-off tests in `ControllersAutoConfigurationSpec` and a new `GrailsCacheAutoConfigurationSpec`; the full-app `grails-test-examples-cache` unit + integration suites and `grails-spring-security-rest-grailscache` (a `grailsCacheManager` consumer) pass.
- Every touched module's test suite passes, plus `grails-test-suite-web`, `-uber` and `-persistence` (the full-lifecycle consumers of these plugins).
- `./gradlew codeStyle` and `./gradlew clean aggregateViolations` are clean across all four analysers (Checkstyle, CodeNarc, PMD, SpotBugs).


grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsUnitTest.groovy
```diff
@@ -118,6 +131,15 @@ trait GrailsUnitTest {
             }
         } catch (NoSuchMethodException e) {}
 
+        try {
+            Method beanRegistrarMethod = clazz.getMethod('beanRegistrar')
+            BeanRegistrar registrar = (BeanRegistrar) beanRegistrarMethod.invoke(plugin)
+            if (registrar != null) {
+                defineBeans(registrar)
+                return
```
jdaugherty: At boot, `GrailsApplicationPostProcessor` applies a lifecycle's `doWithSpring()` DSL first and then its `beanRegistrar()`, so a plugin that defines both hooks (e.g. one that is mid-migration) gets both sets of beans. Here the first hook found wins and the other is skipped — with the current ordering, a plugin defining both would only get its `doWithSpring()` beans in the test context and its registrar beans would silently be missing. Consider applying the DSL (when present) and then the registrar (when present) without the early returns, mirroring the boot order.

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+    @Bean
+    @ConditionalOnMissingBean(name = 'customCacheKeyGenerator')
+    CustomCacheKeyGenerator customCacheKeyGenerator() {
+        new CustomCacheKeyGenerator()
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(name = 'grailsCacheManager')
+    GrailsCacheManager grailsCacheManager(CachePluginConfiguration grailsCacheConfiguration) {
```
jdaugherty, line 62: `grailsCacheConfiguration` is only contributed by the cache plugin descriptor's registrar, but this auto-configuration is gated on `grails.cache.enabled` alone. In a context where the jar is on the classpath but the plugin is not active (plugin excluded, or a plain Boot context that never runs the plugin registration phase), this bean method still runs and the `CachePluginConfiguration` parameter fails to resolve, breaking refresh — before this change everything lived in the descriptor, so deactivating the plugin removed the whole set together. Gating this method (or the class) with `@ConditionalOnBean(CachePluginConfiguration)` would keep the two in lockstep; the registrar-phase definitions are visible to auto-configuration conditions per the ordering this PR already relies on.

grails-redis/src/main/groovy/grails/plugins/redis/util/RedisConfigurationUtil.groovy
```diff
@@ -99,6 +92,84 @@ class RedisConfigurationUtil {
         }
     }
 
+    /**
+     * Registers the pool-config, pool and service bean definitions for a redis connection
+     * directly against a {@link BeanDefinitionRegistry}, mirroring the beans the
+     * {@link #configureService} closure wires through the bean builder DSL. Used by the redis
+     * plugin's {@code beanRegistrar()}-registered post-processor.
+     */
+    static void configureService(BeanDefinitionRegistry registry, def redisConfigMap, String key, Class serviceClass) {
```
jdaugherty, line 103: Unlike the controllers/services/interceptors post-processors (and the guards in the url-mappings and datasource ones), this registry variant registers `redisPoolConfig*`, `redisPool*` and `redisService*` without a `containsBeanDefinition` check, so a definition contributed earlier under one of these names is overwritten rather than winning. That breaks the "an existing definition for a bean name always wins" contract the other converted post-processors follow — and since `RedisBeanDefinitionsPostProcessor` runs at highest precedence, an application override registered before it would be clobbered silently. Consider skipping names that already exist in the registry, either here or in the post-processor.

jdaugherty (approved):
Mostly minor issues that should be able to be fixed easily.

grails-testing-support-core/src/main/groovy/org/grails/testing/GrailsUnitTest.groovy
codeconsole: Fixed in fc6ac88995. `defineBeans(Object plugin)` now applies the `doWithSpring()` DSL first and the `beanRegistrar()` second, no early returns — matching the boot order, with registrar beans winning name conflicts. Covered by the new `DefineBeansPluginHooksSpec`.

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
codeconsole: Fixed in fc6ac88995. Added class-level `@ConditionalOnBean(CachePluginConfiguration)` so the auto-configuration backs off entirely when the plugin is not active; covers `customCacheKeyGenerator` too. Spec test added.

grails-redis/src/main/groovy/grails/plugins/redis/util/RedisConfigurationUtil.groovy
codeconsole: Fixed in fc6ac88995. The registry variant now skips `redisPoolConfig*`, `redisPool*` and `redisService*` names that already have definitions, so existing definitions win as in the other post-processors. Spec test added.

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * <p>Gated on the {@code CachePluginConfiguration} definition contributed by the cache plugin
+ * descriptor's registrar (which runs before auto-configuration conditions are evaluated), so the
+ * auto-configuration backs off entirely when the plugin is not active — e.g. the jar is on the
+ * classpath but the plugin is excluded — keeping it in lockstep with the descriptor.</p>
+ *
+ * @since 8.0
+ */
+@AutoConfiguration
+@ConditionalOnProperty(name = 'grails.cache.enabled', matchIfMissing = true)
```
matrei:
```suggestion
@ConditionalOnBooleanProperty(name = 'grails.cache.enabled', matchIfMissing = true)
```

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
```diff
@@ -0,0 +1,73 @@
...
+    @ConditionalOnMissingBean(name = 'grailsCacheManager')
+    GrailsCacheManager grailsCacheManager(CachePluginConfiguration grailsCacheConfiguration) {
+        if (cacheManagerType == 'GrailsConcurrentLinkedMapCacheManager') {
+            GrailsConcurrentLinkedMapCacheManager cacheManager = new GrailsConcurrentLinkedMapCacheManager()
+            cacheManager.configuration = grailsCacheConfiguration
+            return cacheManager
+        }
+        GrailsConcurrentMapCacheManager cacheManager = new GrailsConcurrentMapCacheManager()
+        cacheManager.configuration = grailsCacheConfiguration
+        return cacheManager
```
matrei:
```suggestion
            return new GrailsConcurrentLinkedMapCacheManager(configuration: grailsCacheConfiguration)
        }
        new GrailsConcurrentMapCacheManager(configuration: grailsCacheConfiguration)
```

jdaugherty (approved):
We should merge this once the tests pass.

grails-async/plugin/src/main/groovy/org/grails/plugins/web/async/ControllersAsyncGrailsPlugin.groovy
```diff
@@ -29,14 +35,17 @@ import org.grails.plugins.web.async.spring.PromiseFactoryBean
  * @author Graeme Rocher
  * @since 2.0
  */
+@CompileStatic
 class ControllersAsyncGrailsPlugin extends Plugin {
 
     def grailsVersion = '7.0.0-SNAPSHOT > *'
```
matrei: Should we update all these to `8.0.0-SNAPSHOT > *`?

grails-cache/src/test/groovy/grails/plugin/cache/CacheGrailsPluginSpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.core.env.MapPropertySource
+import org.springframework.core.env.StandardEnvironment
+
+import spock.lang.Specification
+
+class CacheGrailsPluginSpec extends Specification {
+
+    void "beanRegistrar registers the cache infrastructure beans"() {
+        given:
+        DefaultListableBeanFactory beanFactory = new DefaultListableBeanFactory()
```
matrei: For this line (and all other occurrences where the type can be inferred) the readability is really improved by using `def` (or `var` for Java).

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllerBeanDefinitionsPostProcessor.groovy
```diff
@@ -0,0 +1,94 @@
...
+            definition.autowireMode = AbstractBeanDefinition.AUTOWIRE_BY_NAME
+            if (beanScope == 'prototype') {
+                definition.dependencyCheck = AbstractBeanDefinition.DEPENDENCY_CHECK_NONE
+            }
+            if (useJsessionId) {
+                definition.propertyValues.addPropertyValue('useJessionId', useJsessionId)
+            }
+            registry.registerBeanDefinition(controllerClass.fullName, definition)
+        }
+    }
```
matrei, line 90:
```suggestion
    @Override
    void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) throws BeansException {
        for (def controllerClass : resolveControllerClasses(grailsApplication)) {
            log.debug('Configuring controller {}', controllerClass.fullName)
            if (!controllerClass.available || registry.containsBeanDefinition(controllerClass.fullName)) {
                continue
            }
            def beanDefinition = new GenericBeanDefinition(
                    beanClass: controllerClass.clazz,
                    lazyInit: resolveControllerLazyInit(controllerClass),
                    scope: controllerClass.scope,
                    autowireMode: AbstractBeanDefinition.AUTOWIRE_BY_NAME
            )
            if (controllerClass.scope == 'prototype') {
                beanDefinition.dependencyCheck = AbstractBeanDefinition.DEPENDENCY_CHECK_NONE
            }
            if (useJsessionId) {
                beanDefinition.propertyValues.addPropertyValue('useJessionId', useJsessionId)
            }
            registry.registerBeanDefinition(controllerClass.fullName, beanDefinition)
        }
    }

    private static List<GrailsControllerClass> resolveControllerClasses(GrailsApplication grailsApplication) {
        grailsApplication.getArtefacts(ControllerArtefactHandler.TYPE).collect {
            (GrailsControllerClass) it
        }
    }
    
    private static boolean resolveControllerLazyInit(GrailsControllerClass controllerClass) {
        controllerClass.hasProperty('lazyInit') ? controllerClass.getPropertyValue('lazyInit') : true
    }
```

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersGrailsPlugin.groovy
```diff
@@ -51,51 +56,35 @@ class ControllersGrailsPlugin extends Plugin {
...
-                        }
-                    }
+            // Controller beans autowire by name and use per-controller scopes, which the
+            // BeanRegistry API cannot express — their definitions are contributed by a
+            // dedicated post-processor instead
+            registry.registerBean('controllerBeanDefinitionsPostProcessor', ControllerBeanDefinitionsPostProcessor) { BeanRegistry.Spec<ControllerBeanDefinitionsPostProcessor> spec ->
+                spec.infrastructure().supplier { BeanRegistry.SupplierContext context ->
+                    new ControllerBeanDefinitionsPostProcessor(grailsApplication, useJsessionId)
                 }
             }
```
matrei:
```suggestion
            registry.registerBean('controllerBeanDefinitionsPostProcessor', ControllerBeanDefinitionsPostProcessor) {
                it.infrastructure().supplier {
                    new ControllerBeanDefinitionsPostProcessor(grailsApplication, useJsessionId)
                }
        }
```

grails-converters/src/main/groovy/org/grails/plugins/converters/ConvertersGrailsPlugin.groovy
```diff
@@ -48,22 +55,30 @@ class ConvertersGrailsPlugin extends Plugin {
...
-                marshaller = { JsonErrorsMarshaller om -> }
-                converterClass = JSON
+            registry.registerBean('errorsJsonMarshallerRegisterer', ObjectMarshallerRegisterer) { BeanRegistry.Spec<ObjectMarshallerRegisterer> spec ->
+                spec.supplier { BeanRegistry.SupplierContext context ->
+                    ObjectMarshallerRegisterer registerer = new ObjectMarshallerRegisterer()
+                    registerer.marshaller = context.bean('jsonErrorsMarshaller', JsonErrorsMarshaller)
+                    registerer.converterClass = JSON
+                    return registerer
+                }
             }
```
matrei:
```suggestion
            registry.registerBean('errorsXmlMarshallerRegisterer', ObjectMarshallerRegisterer) {
                it.supplier {
                    new ObjectMarshallerRegisterer(
                            marshaller: it.bean('xmlErrorsMarshaller', XmlErrorsMarshaller),
                            converterClass: XML
                    )
                }
            }

            registry.registerBean('errorsJsonMarshallerRegisterer', ObjectMarshallerRegisterer) {
                it.supplier {
                    new ObjectMarshallerRegisterer(
                            marshaller: it.bean('jsonErrorsMarshaller', JsonErrorsMarshaller),
                            converterClass: JSON
                    )
                }
            }
```

grails-converters/src/test/groovy/org/grails/plugins/converters/ConvertersGrailsPluginSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+        new BeanRegistryAdapter(beanFactory, new StandardEnvironment(), registrar.getClass()).register(registrar)
+    }
+
+    void "beanRegistrar registers the converters beans"() {
+        expect:
+        beanFactory.getBeanDefinition('jsonErrorsMarshaller').beanClassName == JsonErrorsMarshaller.name
+        beanFactory.getBeanDefinition('xmlErrorsMarshaller').beanClassName == XmlErrorsMarshaller.name
+        beanFactory.getBeanDefinition('convertersConfigurationInitializer').beanClassName == ConvertersConfigurationInitializer.name
+        beanFactory.containsBeanDefinition('errorsXmlMarshallerRegisterer')
+        beanFactory.containsBeanDefinition('errorsJsonMarshallerRegisterer')
```
matrei: Instances like these can be cleaned up with Spocks `with`-feature:

```suggestion
        with(beanFactory) {
            getBeanDefinition('jsonErrorsMarshaller').beanClassName == JsonErrorsMarshaller.name
            getBeanDefinition('xmlErrorsMarshaller').beanClassName == XmlErrorsMarshaller.name
            getBeanDefinition('convertersConfigurationInitializer').beanClassName == ConvertersConfigurationInitializer.name
            containsBeanDefinition('errorsXmlMarshallerRegisterer')
            containsBeanDefinition('errorsJsonMarshallerRegisterer')
        }
```

matrei (approved):
LGTM. I have added some comments that I think would be improvements. I did not have time to go through all of the files today, so if willing, consider applying the same patterns to the rest of the files.

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
codeconsole: Applied in 76bdd8c079.

grails-cache/src/main/groovy/grails/plugin/cache/GrailsCacheAutoConfiguration.groovy
codeconsole: Applied in 76bdd8c079.

grails-async/plugin/src/main/groovy/org/grails/plugins/web/async/ControllersAsyncGrailsPlugin.groovy
codeconsole: Yes — the converted descriptors now require the Grails 8-only `BeanRegistrar` API, so 76bdd8c079 bumps every descriptor this PR converts to `8.0.0-SNAPSHOT > *` (including gorm-graphql's `7.1.0 > *`), matching the hibernate and undertow plugins.

grails-cache/src/test/groovy/grails/plugin/cache/CacheGrailsPluginSpec.groovy
codeconsole: Applied across all specs in this PR in 76bdd8c079.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllerBeanDefinitionsPostProcessor.groovy
codeconsole: Applied in 76bdd8c079, with one tweak: `resolveControllerLazyInit` keeps the `as boolean` coercion since the class is `@CompileStatic` and `getPropertyValue` returns `Object`.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersGrailsPlugin.groovy
codeconsole: Applied in 76bdd8c079 — and swept the same style through all the converted descriptors so they stay consistent.

grails-converters/src/main/groovy/org/grails/plugins/converters/ConvertersGrailsPlugin.groovy
codeconsole: Applied in 76bdd8c079.

grails-converters/src/test/groovy/org/grails/plugins/converters/ConvertersGrailsPluginSpec.groovy
codeconsole: Applied in 76bdd8c079.

PR #15995, opened 2026-07-17
Improve unresolved plugin dependency diagnostics

jamesfredley:
## Summary

- replace the malformed single-line ERROR for an unresolved plugin with one concise WARN
- enumerate each missing or version-incompatible dependency with required and actual versions
- reuse existing resolution rules; no change to discovery, registration, failed-plugin tracking, or load order

## Verification

- :grails-core:test --tests PluginDiscoverySpec
- :grails-core:test
- git diff --check

This is an AI-generated starting point for richer plugin dependency diagnostics.

### DIFF
--- a/grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
+++ b/grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
@@ -429,23 +429,50 @@ public class DefaultPluginDiscovery implements PluginDiscovery {
         for (var name : plugin.getDependsOnNames()) {
             var requiredVersion = plugin.getMetadata().getDependentVersion(name);
             var dependency = findPlugin(name);
-            if (dependency == null) {
+            if (dependency != null) {
+                if (!GrailsVersionUtils.isValidVersion(dependency.getPluginVersion(), requiredVersion)) {
+                    unresolvedDependencies.add(
+                            "dependency [" + name + "] has version [" + dependency.getPluginVersion() +
+                                    "] but requires [" + requiredVersion + "]"
+                    );
+                }
+            } else if (getFailedPlugin(name) != null) {
+                unresolvedDependencies.add(
+                        "dependency [" + name + "] with required version [" + requiredVersion + "] failed to load"
+                );
+            } else if (isDelayed(name)) {
+                unresolvedDependencies.add(
+                        "dependency [" + name + "] with required version [" + requiredVersion +
+                                "] is still pending load"
+                );
+            } else {
                 unresolvedDependencies.add(
                         "dependency [" + name + "] with required version [" + requiredVersion + "] is missing"
                 );
-            } else if (!GrailsVersionUtils.isValidVersion(dependency.getPluginVersion(), requiredVersion)) {
-                unresolvedDependencies.add(
-                        "dependency [" + name + "] has version [" + dependency.getPluginVersion() +
-                                "] but requires [" + requiredVersion + "]"
-                );
             }
         }
-        LOG.warn(
+        LOG.error(
                 "Grails plug-in [{}] with version [{}] cannot be loaded: {}",
                 plugin.getName(),
                 plugin.getPluginVersion(),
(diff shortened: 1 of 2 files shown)


davydotcom (approved):
Looks good — approving.

Failure determination is unchanged: plugins still fail the same way via the existing dependency resolution / delayed-load path into `failedPlugins`. Discovery, registration, failed-plugin tracking, and load order are untouched.

What’s new here is the notification: replacing the malformed ERROR with a single concise WARN that enumerates missing vs version-incompatible dependencies. That’s a clearer alert for users without changing how unresolved plugins are decided.

grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
```diff
@@ -418,16 +418,36 @@ private void loadDelayedPlugins() {
...
+                        "dependency [" + name + "] with required version [" + requiredVersion + "] is missing"
+                );
+            } else if (!GrailsVersionUtils.isValidVersion(dependency.getPluginVersion(), requiredVersion)) {
+                unresolvedDependencies.add(
+                        "dependency [" + name + "] has version [" + dependency.getPluginVersion() +
+                                "] but requires [" + requiredVersion + "]"
+                );
+            }
+        }
+        LOG.warn(
```
jdaugherty: I disagree on this being a warning, there is a legitimate configuration issue.  This should be an error message

jdaugherty (changes requested):

grails-core/src/main/groovy/org/apache/grails/core/plugins/DefaultPluginDiscovery.java
davydotcom: Please keep this at ERROR rather than WARN.

Unresolved dependencies already put the plugin in `failedPlugins`; that is a load failure, and the previous code logged it with `LOG.error`. A clearer message is welcome, but the severity should stay ERROR so it isn’t lost when WARN is filtered.

davydotcom (changes requested):
Revising my earlier approval with one requested change.

The richer diagnostics look good, and failure determination is still unchanged — this is still just a clearer alert for an already-failed plugin. Please keep the log level at **ERROR**, not WARN.

The previous message was already `LOG.error`. Downgrading to WARN isn’t justified in the PR beyond “one concise WARN,” and unresolved/failed plugin dependencies are a real load failure. ERROR keeps the severity aligned with that outcome and with prior behavior; WARN can be filtered out of environments that only surface errors.

Requested change:
- use `LOG.error(...)` in `logUnresolvedDependencies`
- update the new `PluginDiscoverySpec` assertion accordingly (`ERROR` instead of `WARN`)

jdaugherty:
Reviewed the branch against current `8.0.x` and pushed two follow-up commits (a merge of `8.0.x` plus a fix) so this can go in.

**What was wrong**

- The log level is already back at ERROR from the earlier feedback, and that stays. Both new spec features failed once merged with `8.0.x`, though: they captured `System.err` expecting slf4j-simple output, but `DefaultPluginDiscovery` logging now goes through logback on the base branch, so the capture saw nothing (`errors.size() == 1` failed with `[]`, and the second feature hit an NPE on a null line).
- The `is still pending load` branch hid the real cause. A plugin only fails while a same-named dependency is still delayed when that delayed candidate does not satisfy the version constraint (`isDependentOn` checks name and version before re-queuing). So the message reported the dependency as merely pending, when the actual problem was the version.

**What changed**

- Rewrote both features on the existing `LogCapture` test fixture and assert on `Level.ERROR` plus the formatted message, matching the neighbouring duplicate-registration test.
- `isDelayed(name)` became `findDelayedPlugin(name)`, and the message now reads `dependency [x] with required version [2.0.0] is still pending load and only version [1.0.0] was found`. Added a feature that drives that path (dependent processed before its too-old, still-delayed dependency gives up).

**Verification**

- `:grails-core:test` full module: 657 tests, 0 failures
- `:grails-core:codeStyle` clean

jdaugherty:
The original issue has been addressed, so merging.

PR #15998, opened 2026-07-17
Add artefact name contract characterization spec

jamesfredley:
## Summary

- add a test-only characterization spec locking in public GrailsClass artefact naming behavior
- cover controllers, services, domains, and url-mappings, including acronym-heavy names, repeated accessor reads, and repeated wrapper construction
- assert the framework's actual observed names; no production changes and no timing thresholds

## Verification

- :grails-core:test --tests ArtefactNamingContractSpec
- :grails-core:test
- git diff --check

This is an AI-generated starting point that pins the naming contract ahead of any GrailsNameUtils precomputation work.


grails-core/src/test/groovy/org/grails/core/ArtefactNamingContractSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+                ],
+                [
+                        label: 'Acronym domain',
+                        artifactType: 'domain',
+                        wrapperClass: URLDomain
+                ]
+        ]
+    }
+
+    private GrailsClass grailsClassFor(String artifactType, Class<?> wrapperClass) {
```
jdaugherty, line 214: Constructing `DefaultGrailsControllerClass`/`DefaultGrailsServiceClass`/etc. directly bypasses artefact detection entirely — these constructors accept any class (`new DefaultGrailsControllerClass(URLDomain)` would work just as well). So this spec pins the naming metadata contract, but not the detection contract (`ArtefactHandler.isArtefact`), which is the part a precomputation refactor is most likely to disturb.

Consider adding cases that exercise the handlers, e.g.:
- an abstract `FooController` is rejected (`ControllerArtefactHandler` passes `allowAbstract=false`)
- `URLDomain` without `@Entity`/`@Artefact` is rejected by `DomainClassArtefactHandler.isArtefactClass` (domain detection is annotation/location-based; the name plays no role)
- a suffix match like `HTMLController` is accepted by `ControllerArtefactHandler.isArtefactClass`

grails-core/src/test/groovy/org/grails/core/ArtefactNamePrecomputationSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+package org.grails.core
+
+import grails.core.GrailsClass
+import spock.lang.Specification
+import spock.lang.Unroll
+
+/**
+ * Tests for deterministic GrailsClass naming metadata.
+ */
+class ArtefactNamePrecomputationSpec extends Specification {
```
jdaugherty: "Precomputation" in the spec name is aspirational — nothing here asserts precomputation; the repeated-read/repeated-construction tests pin stability only. That's a fine pre-refactor baseline, but consider a name that says what it locks in, e.g. `ArtefactNamingContractSpec`, so it doesn't read as if precomputed behavior already exists.

grails-core/src/test/groovy/org/grails/core/ArtefactNamingContractSpec.groovy
```diff
@@ -0,0 +1,213 @@
...
+                        label: 'Acronym controller',
+                        artifactType: 'controller',
+                        wrapperClass: HTMLController,
+                        name: 'HTML',
+                        shortName: 'HTMLController',
+                        fullName: "${BASE_PACKAGE}.HTMLController",
+                        packageName: BASE_PACKAGE,
+                        propertyName: 'HTMLController',
+                        logicalPropertyName: 'HTML',
+                        naturalName: 'HTMLC ontroller'
```
jdaugherty, line 68: For the record: expectations like `'HTMLC ontroller'` / `'JSONAPIS ervice'` / `'URLD omain'` are correct characterizations — GrailsNameUtils.getNaturalName genuinely splits acronym-prefixed names after the last uppercase letter. Worth a brief code comment noting these pin known-quirky behavior intentionally, so a future reader doesn't "fix" the expectations (or the algorithm) without realizing this spec exists to catch exactly that change.

PR #15999, opened 2026-07-17
Document continuous test mode

jamesfredley:
## Summary

- document the existing `grails test-app -continuous` continuous test mode
- cover the watched source directories, .groovy/.java triggers, no initial run, and interactive-CLI lifecycle/termination
- show the correct interactive sequence (start `grails`, then `test-app -continuous` at the prompt) and explain why a one-off invocation exits immediately
- distinguish the Grails DirectoryWatcher from Gradle --continuous; synopsis lists -continuous as an independent optional flag

## Verification

- :grails-doc:publishGuide -x aggregateGroovydoc
- git diff --check

This is an AI-generated starting point documenting existing continuous test behavior.

### DIFF
--- a/grails-doc/src/en/guide/testing/runningTests/withGrailsCli.adoc
+++ b/grails-doc/src/en/guide/testing/runningTests/withGrailsCli.adoc
@@ -59,6 +59,25 @@ open test-report
 You can also run your unit tests from within most IDEs.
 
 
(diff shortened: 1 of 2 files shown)


davydotcom (approved):
This is cool, looks good too! applied one recommended copilot fix that made sense to me.

PR #16000, opened 2026-07-17
Add optional artefact index reader for startup scanning

jamesfredley:
## Summary

- Add an optional internal artefact-index reader for `META-INF/grails/artefacts.idx`
- `ApplicationArtefactScanner` uses a valid application-root index (honoring `packageNames`) and safely falls back to classpath scanning when the index is absent, malformed, unresolvable, or linkage-invalid
- Deterministic ordering, deduplication, and transformed-class inclusion preserved

## Why

Today, artefact discovery (`ApplicationArtefactScanner` → `ClassPathScanner`) runs at every application boot: it resource-pattern-matches the classpath per package, reads each candidate `.class` file's metadata via ASM, and checks it for a `grails.*`-prefixed annotation. That's proportional to the number of classes on the classpath, not the number of actual artefacts, and it's pure runtime I/O + reflection-adjacent work that produces the same result on every single boot for a given build.

An artefact index moves that discovery from "boot time, every time" to "build time, once": a producer (not part of this PR — see Scope note) would emit the flat list of artefact class names once, at build time, and boot then becomes a plain file read plus loading exactly those classes — no classpath scanning, no per-class metadata reads. That's the startup-speed win, and it scales with the number of artefacts instead of the size of the classpath.

This is consistent with a direction the team is already moving in elsewhere: shifting checks/work that don't need to happen at runtime to compile/build time (e.g. the compile-time SQL injection prevention work). This PR proposes applying that same idea to artefact discovery.

## Scope note

This is a starter/seed: it adds only the reader and its safe fallback. The index producer is not included, so the optimization activates only once an `artefacts.idx` is emitted into the application code-source root; until then behavior is unchanged (fallback scanning).

## Verification

- `:grails-core:test --tests ApplicationArtefactScannerSpec`
- `:grails-core:test`
- `git diff --check`

This is an AI-generated starting point for build-time artefact indexing.

### DIFF
--- a/grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
+++ b/grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
@@ -1,0 +1,83 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package grails.boot.config;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import grails.io.IOUtils;
+
+/**
+ * Reads the optional application artefact index.
+ *
+ * <p>The index is UTF-8 text at {@value #RESOURCE_NAME}, with one fully qualified
+ * class name per nonempty line. Any unreadable or unresolvable entry rejects the
+ * complete index so callers can use their normal classpath scan.</p>
+ */
+final class ArtefactIndexReader {
+
+    static final String RESOURCE_NAME = "META-INF/grails/artefacts.idx";
+
+    private ArtefactIndexReader() {
+    }
+
+    static Collection<Class> read(Class<?> applicationClass, Collection<String> packageNames) {
+        try {
+            URL resource = new URL(IOUtils.findRootResource(applicationClass), RESOURCE_NAME);
+            Set<Class> classes = new LinkedHashSet<>();
+            return readResource(resource, applicationClass.getClassLoader(), packageNames, classes) ? classes : null;
+        } catch (IOException ignored) {
+            return null;
+        }
+    }
+
+    private static boolean readResource(URL resource, ClassLoader classLoader, Collection<String> packageNames, Set<Class> classes) {
+        try (InputStream inputStream = resource.openStream();
+             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+            String className;
+            while ((className = reader.readLine()) != null) {
+                if (className.isEmpty()) {
+                    return false;
+                }
+                if (isInPackage(className, packageNames)) {
+                    classes.add(classLoader.loadClass(className));
+                }
+            }
+            return true;
+        } catch (IOException | ClassNotFoundException | LinkageError ignored) {
+            return false;
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


davydotcom:
Trying to understand the intent behind this one. Is it for startup speed and if so, how does it help there. PR description just needs a little more to help me understand intent.

grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private ArtefactIndexReader() {
+    }
+
+    static Collection<Class> read(Class<?> applicationClass, Collection<String> packageNames) {
+        try {
+            URL resource = new URL(IOUtils.findRootResource(applicationClass), RESOURCE_NAME);
+            Set<Class> classes = new LinkedHashSet<>();
+            return readResource(resource, applicationClass.getClassLoader(), packageNames, classes) ? classes : null;
+        } catch (IOException ignored) {
```
jdaugherty: `IOUtils.findRootResource` throws **IllegalStateException**, not IOException, when the class resource can't be resolved (`targetClass.getResource(...)` returning null — e.g. an application class from a classloader that doesn't expose `.class` resources). That escapes this catch and would fail startup, where today the same situation just scans. Since the whole contract of this reader is "never make things worse than the fallback," this should catch that too (e.g. `catch (IOException | RuntimeException)`).

grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
```diff
@@ -0,0 +1,83 @@
...
+ * complete index so callers can use their normal classpath scan.</p>
+ */
+final class ArtefactIndexReader {
+
+    static final String RESOURCE_NAME = "META-INF/grails/artefacts.idx";
+
+    private ArtefactIndexReader() {
+    }
+
+    static Collection<Class> read(Class<?> applicationClass, Collection<String> packageNames) {
```
jdaugherty, line 68: Two operational concerns for production startup code:

1. **Silence.** Both outcomes are invisible — no log when the index is used, none when it's rejected. A corrupted index silently degrades to slow scanning and nobody finds out; a working index can't be confirmed either. Suggest debug-level logs for "index used (N entries)" and "index rejected, falling back".
2. **Stale-index hazard.** A valid but *incomplete* index is the dangerous case: developer adds an artefact, index isn't regenerated, and the new class is silently absent from the application — no error, no fallback. The producer side will need a freshness guarantee (or the index should carry a hash/marker the reader can validate), and a kill-switch system property to force scanning would be a cheap escape hatch worth adding in the seed.

grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
```diff
@@ -0,0 +1,83 @@
...
+                    classes.add(classLoader.loadClass(className));
+                }
+            }
+            return true;
+        } catch (IOException | ClassNotFoundException | LinkageError ignored) {
+            return false;
+        }
+    }
+
+    private static boolean isInPackage(String className, Collection<String> packageNames) {
```
jdaugherty, line 106: Worth documenting the semantic differences from the `ClassPathScanner` path this replaces, since the future index producer has to compensate for them:

- The scanner only returns classes carrying an annotation whose name starts with `grails.` (default `annotationFilter`); the reader trusts every listed entry with no annotation check, so a hand-edited or buggy index can inject arbitrary classes into the artefact set.
- The scanner skips `DEFAULT_IGNORED_ROOT_PACKAGES` (`com`, `org`, `net`, …) even when explicitly passed as packageNames; the reader honors them.

Both are fine if the producer mirrors scan semantics exactly, but that contract currently lives nowhere — a sentence in the class javadoc would pin it.

grails-core/src/main/groovy/grails/boot/config/ArtefactIndexReader.java
```diff
@@ -0,0 +1,83 @@
...
+            return null;
+        }
+    }
+
+    private static boolean readResource(URL resource, ClassLoader classLoader, Collection<String> packageNames, Set<Class> classes) {
+        try (InputStream inputStream = resource.openStream();
+             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+            String className;
+            while ((className = reader.readLine()) != null) {
+                if (className.isEmpty()) {
```
jdaugherty, line 92: Rejecting the entire index on an empty line is stricter than it needs to be — a trailing blank line is the most common artifact of text-file generation and concatenation (the spec's own `writeIndex` has to `.trim()` to avoid it). Skipping blank lines (`continue`) keeps the strict-reject behavior for genuinely malformed content while tolerating the boring case. If strictness is intentional as a whole-file integrity signal, the javadoc should say the producer must not emit blank lines, including trailing ones.

jdaugherty (changes requested):
The initial review is below, but I think this needs discussed more in the weekly since it's an architectural shift.  In general, I'm ok moving to more compile time based fixes for the performance benefit, but we need to agree as a team.  Putting a PR out here prior to a discussion is concerning, since it could be merged without a wider architectural discussion.

jdaugherty:
I'm going to move this out of 8.0 given that this is just a seed / not even fully implemented yet.

PR #16003, opened 2026-07-17
Characterize data-binding allowlist through public binding API

jamesfredley:
## Summary

- add a test-only characterization spec that pins the compile-time data-binding allowlist contract via OBSERVABLE behavior through the public binding API
- covers a domain (via `DataBindingUtils.bindObjectToInstance`) and a controller-action `@Validateable` command
- asserts which properties actually bind vs stay null, so the test tracks the generated allowlist's effect rather than an internal field

## Behavior characterized (current 8.0.x)

- Domain: `name` and nested `address.street` bind; `id`/`version`/`dateCreated`/`lastUpdated` and a declared non-allowlisted `Object` property stay null
- `@Validateable` command: declared special properties bind; the non-allowlisted property stays null

## Scope note

Test-only; no production changes. Exercises only public data-binding APIs (per the repo's test-through-public-API policy); it does not read the AST-generated allowlist field.

## Verification

- :grails-test-suite-web:test --tests DefaultDatabindingWhitelistBehaviorSpec
- :grails-test-suite-web:test
- git diff --check

This is an AI-generated starting point pinning the data-binding code-generation contract.


grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+@Entity
+class WhitelistDomain {
+    String name
+    WhitelistAddress address = new WhitelistAddress()
+    Long id
+    Long version
+    Date dateCreated
+    Date lastUpdated
+    Object ignored
```
jdaugherty: "Unlisted" / "non-allowlisted" slightly mischaracterizes why this property doesn't bind: `shouldFieldBeInWhiteList` (DefaultASTDatabindingHelper.java:277) excludes it specifically because its static type is `Object` without generics. There is no general "unlisted" category — every eligible typed, non-transient, non-special property is auto-whitelisted, so e.g. an extra `String` property here *would* bind. Suggest naming this `Object untypedProperty` (or a `def` field) and wording the test name as "excludes special and Object/def-typed properties" so the pinned rule reads as what it is.

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+    static hasOne = [address: WhitelistAddress]
```
jdaugherty: `hasOne` declares a bidirectional association whose foreign key lives on the child, and `WhitelistAddress` has no back-reference — GORM would reject this mapping if the domain were ever registered with a mapping context (it isn't here, since the spec uses ControllerUnitTest without DataTest, which is why nothing complains). The typed `address` property alone is what puts `address`/`address.*` in the generated whitelist, so `hasOne` adds nothing to the contract under test. Suggest dropping it (and possibly the field initializer tells the nested-binding story on its own).

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+import grails.artefact.Artefact
+import grails.persistence.Entity
+import grails.testing.web.controllers.ControllerUnitTest
+import grails.validation.Validateable
+import grails.web.databinding.DataBindingUtils
+import spock.lang.Specification
+
+class DefaultDatabindingWhitelistBehaviorSpec extends Specification implements ControllerUnitTest<WhitelistBehaviorController> {
+
+    void 'domain binding includes simple and association properties but excludes special and unlisted properties'() {
```
jdaugherty: Partial overlap: `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec` in this same package already pins the domain special-property exclusions (id/version via #15681, dateCreated/lastUpdated via GRAILS-11173), also through observable binding behavior. The genuinely new coverage here is the Object-typed exclusion, nested association binding, and the Validateable command *including* id/version/dateCreated/lastUpdated. Worth a cross-reference comment, or trimming this first test to the non-duplicated assertions, so the two specs don't drift apart when the contract changes.

matrei (approved):

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
jamesfredley: Addressed: the field is now `Object untypedProperty`, the test names talk about Object/def-typed exclusion rather than an "unlisted" category, and we also pin a `def untypedDefProperty` plus an extra `String extra` that *does* bind so the auto-whitelist for ordinary typed properties is observable too.

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
jamesfredley: Addressed in 1c34ff: `hasOne` was dropped. Nested binding is covered by the typed `address` field and its initializer alone.

grails-test-suite-web/src/test/groovy/org/grails/web/binding/DefaultDatabindingWhitelistBehaviorSpec.groovy
jamesfredley: Addressed: the domain test no longer re-asserts id/version/dateCreated/lastUpdated exclusion, and there is a cross-reference to `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec` (GRAILS-11173, #15681). This spec now only pins nested association binding, extra typed-property inclusion, and Object/def-typed exclusion.

PR #16005, opened 2026-07-17
docs: fix broken external and internal links in Groovydocs

borinquenkid:
Continues #15327 by @sanjana2505006 — rebases onto current 7.0.x and addresses @jdaugherty's outstanding review feedback:

- Removes the stray `TEST_FAILURES.md` that was accidentally committed (flagged in #15327, never addressed).
- Decouples `auditGroovydocLinks` from the default `docs` build so it can't fail every future docs build on a false positive. jdaugherty's last comment on #15327: *"The audit should be a task we run, but it shouldn't be run by default unless we're going to capture it in the workflow summary of the build."* It remains available on demand: `./gradlew :grails-doc:auditGroovydocLinks`.
- Extracts the link-violation-detection logic into a Gradle-API-free `GroovydocLinkAuditor` class with unit tests (`build-logic/docs-core` deliberately keeps Gradle classes off the test compile classpath, so the previous `DefaultTask`-only implementation had no path to test coverage).

Verified against a full `aggregateGroovydoc` run on current `7.0.x` (all 60+ modules): **0 violations** — the ~30 phantom-link false positives Sanjana observed earlier in the PR's life (from the old Python-script-based approach) no longer reproduce with the current native-Groovy audit logic.

Closes #15327.

Co-Authored-By: Sanjana <sanju250506@gmail.com>

### DIFF
--- a/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
+++ b/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
@@ -27,24 +27,17 @@ import java.util.regex.Matcher
 import java.util.regex.Pattern
 
 /**
- * A task that repairs malformed navigation links in generated Groovydocs.
+ * A task that audits generated Groovydocs for malformed navigation links.
  * Specifically targets 'phantom' links and relative path issues in navigation files
  * like overview-summary.html, deprecated-list.html, and help-doc.html.
  */
-abstract class FixGroovydocLinksTask extends DefaultTask {
-    
-    /**
-     * If true, the task will fail the build if any malformed links are found.
-     * If false, the task will attempt to repair the links (patching mode).
-     */
-    @org.gradle.api.tasks.Input
-    boolean auditMode = false
+abstract class AuditGroovydocLinksTask extends DefaultTask {
 
     @InputDirectory
     abstract DirectoryProperty getApiDocsDir()
 
     @TaskAction
-    void fixLinks() {
+    void auditLinks() {
         File apiDir = apiDocsDir.get().asFile
         if (!apiDir.exists()) {
             logger.warn "API documentation directory does not exist: ${apiDir.absolutePath}"
@@ -59,24 +52,20 @@ abstract class FixGroovydocLinksTask extends DefaultTask {
             (Pattern.compile(/href='([^']+?)\/overview-summary\.html'/)): "href='overview-summary.html'"
         ]
 
-        int totalFixes = 0
         List<String> violations = []
 
         apiDir.eachFileRecurse { File file ->
             if (file.name.endsWith(".html")) {
                 String content = file.text
-                boolean changed = false
                 
                 replacements.each { pattern, replacement ->
                     Matcher matcher = pattern.matcher(content)
                     if (matcher.find()) {
-                        content = matcher.replaceAll(replacement)
-                        changed = true
-                        violations << "Malformed nav link in ${file.name}"
+                        violations << "Malformed nav link in ${file.name}: ${matcher.group(0)}"
                     }
                 }
                 
-                // Fix specific inner class path issues like Query/Order.Direction.html -> Query.Order.Direction.html
+                // Check specific inner class path issues like Query/Order.Direction.html -> Query.Order.Direction.html
                 Pattern innerClassPattern = Pattern.compile(/href='([^']+?)\/([A-Z][A-Za-z0-9_]*?)\/([A-Z][A-Za-z0-9_.]*?\.html)'/)
                 Matcher innerMatcher = innerClassPattern.matcher(content)
                 while (innerMatcher.find()) {
@@ -88,30 +77,16 @@ abstract class FixGroovydocLinksTask extends DefaultTask {
                     Path targetPath = currentPath.resolve(relPath).resolve("${outer}.${inner}").normalize()
                     
                     if (Files.exists(targetPath)) {
-                        String newHref = "href='${relPath}/${outer}.${inner}'"
-                        content = content.replace(innerMatcher.group(0), newHref)
-                        changed = true
-                        violations << "Malformed inner class link in ${file.name}: ${innerMatcher.group(0)} -> ${newHref}"
-                    }
-                }
-
-                if (changed) {
-                    totalFixes++
-                    if (!auditMode) {
-                        file.text = content
+                        violations << "Malformed inner class link in ${file.name}: ${innerMatcher.group(0)}"
                     }
                 }
             }
         }
         
-        if (totalFixes > 0) {
-            if (auditMode) {
-                violations.take(10).each { logger.error(it) }
-                if (violations.size() > 10) logger.error("... and ${violations.size() - 10} more")
-                throw new org.gradle.api.GradleException("Found ${violations.size()} malformed links in Groovydoc. Please fix the source issue rather than patching. See logs for details.")
-            } else {
-                logger.lifecycle "Repaired ${totalFixes} HTML files in ${apiDir.absolutePath}"
-            }
+        if (!violations.isEmpty()) {
+            violations.take(10).each { logger.error(it) }
+            if (violations.size() > 10) logger.error("... and ${violations.size() - 10} more")
+            throw new org.gradle.api.GradleException("Found ${violations.size()} malformed links in Groovydoc. Please fix the source issue rather than patching. See logs for details.")
         } else {
             logger.lifecycle "No malformed Groovydoc links found in ${apiDir.absolutePath}"
         }
--- a/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
+++ b/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
@@ -22,9 +22,6 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.TaskAction
-import java.nio.file.*
-import java.util.regex.Matcher
-import java.util.regex.Pattern
 
 /**
  * A task that audits generated Groovydocs for malformed navigation links.
@@ -44,45 +41,8 @@ abstract class AuditGroovydocLinksTask extends DefaultTask {
             return
         }
 
-        // Patterns common to Groovydoc navigation failures
-        Map<Pattern, String> replacements = [
-            (Pattern.compile(/href='([^']+?)\/deprecated-list\.html'/)): "href='deprecated-list.html'",
-            (Pattern.compile(/href='([^']+?)\/help-doc\.html'/)): "href='help-doc.html'",
-            (Pattern.compile(/href='([^']+?)\/index-all\.html'/)): "href='index-all.html'",
-            (Pattern.compile(/href='([^']+?)\/overview-summary\.html'/)): "href='overview-summary.html'"
-        ]
+        List<String> violations = GroovydocLinkAuditor.findViolations(apiDir)
 
-        List<String> violations = []
-
-        apiDir.eachFileRecurse { File file ->
-            if (file.name.endsWith(".html")) {
-                String content = file.text
-                
-                replacements.each { pattern, replacement ->
-                    Matcher matcher = pattern.matcher(content)
-                    if (matcher.find()) {
-                        violations << "Malformed nav link in ${file.name}: ${matcher.group(0)}"
-                    }
-                }
-                
-                // Check specific inner class path issues like Query/Order.Direction.html -> Query.Order.Direction.html
-                Pattern innerClassPattern = Pattern.compile(/href='([^']+?)\/([A-Z][A-Za-z0-9_]*?)\/([A-Z][A-Za-z0-9_.]*?\.html)'/)
-                Matcher innerMatcher = innerClassPattern.matcher(content)
-                while (innerMatcher.find()) {
-                    String relPath = innerMatcher.group(1)
-                    String outer = innerMatcher.group(2)
-                    String inner = innerMatcher.group(3)
-                    
-                    Path currentPath = file.toPath().parent
-                    Path targetPath = currentPath.resolve(relPath).resolve("${outer}.${inner}").normalize()
-                    
-                    if (Files.exists(targetPath)) {
-                        violations << "Malformed inner class link in ${file.name}: ${innerMatcher.group(0)}"
-                    }
-                }
-            }
-        }
-        
         if (!violations.isEmpty()) {
             violations.take(10).each { logger.error(it) }
             if (violations.size() > 10) logger.error("... and ${violations.size() - 10} more")
--- a/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
+++ b/build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
@@ -21,33 +21,67 @@ package org.grails.doc.gradle
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.SetProperty
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.TaskAction
 
 /**
- * A task that audits generated Groovydocs for malformed inner-class links, where a
- * slash-separated 'phantom' path (e.g. {@code Outer/Inner.html}) was emitted instead
- * of the dotted Groovydoc naming convention ({@code Outer.Inner.html}).
+ * A task that audits generated Groovydocs for relative links that resolve to a page which
+ * was never generated. Links that Groovydoc itself is responsible for are reported as
+ * warnings; links that this repository can fix - by putting the type on the Groovydoc
+ * classpath or mapping its package to an external javadoc site - fail the build.
  *
  * @see GroovydocLinkAuditor
  */
 abstract class AuditGroovydocLinksTask extends DefaultTask {
 
+    private static final int MAX_REPORTED = 25
+
     @InputDirectory
     abstract DirectoryProperty getApiDocsDir()
+
+    /**
+     * Package prefixes with no published javadoc to link to. Links into these are reported
+     * as neither a warning nor a violation.
+     */
+    @Input
+    abstract SetProperty<String> getUnmappedPackages()
 
     @TaskAction
     void auditLinks() {
         File apiDir = apiDocsDir.get().asFile
 
-        List<String> violations = GroovydocLinkAuditor.findViolations(apiDir)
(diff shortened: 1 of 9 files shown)


gradle/docs-dependencies.gradle
```diff
@@ -56,12 +50,35 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
         }
         def springVersion = resolveProjectVersion('spring-core')
         if (springVersion) {
-            links << [packages: 'org.springframework.core.', href: "https://docs.spring.io/spring-framework/docs/${springVersion}/javadoc-api/"]
+            links << [packages: 'org.springframework.', href: "https://docs.spring.io/spring-framework/docs/${springVersion}/javadoc-api/"]
```
jdaugherty: Groovydoc resolves `links` first-match-wins in list order (`SimpleGroovyClassDoc.getDocUrl` iterates the links and returns on the first `type.startsWith(prefix)` hit). Broadening this entry to `org.springframework.` while it sits *before* the `org.springframework.boot.` entry means every Spring Boot class reference now matches here first and links into `docs.spring.io/spring-framework/...`, where the Boot packages don't exist — the Boot entry below becomes dead code and all Boot links 404.

Most-specific-first fixes it: move the `org.springframework.boot.` entry above this one.

gradle/docs-dependencies.gradle
```diff
@@ -56,12 +50,35 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
             links << [packages: 'org.springframework.boot.', href: "https://docs.spring.io/spring-boot/docs/${springBootVersion}/api/"]
         }
+        def hibernateVersion = resolveProjectVersion('hibernate-core')
+        if (hibernateVersion) {
+            def shortVersion = hibernateVersion.split('\\.').take(2).join('.')
+            links << [packages: 'org.hibernate.', href: "https://docs.jboss.org/hibernate/orm/${shortVersion}/javadocs/"]
+        }
+        def jakartaValidationVersion = resolveProjectVersion('jakarta.validation-api')
+        if (jakartaValidationVersion) {
+            links << [packages: 'jakarta.validation.', href: "https://jakarta.ee/specifications/bean-validation/3.0/apidocs/"]
```
jdaugherty: Minor: these three Jakarta links gate on the resolved artifact version but hardcode the spec version in the URL (`3.0`, `3.1`, `platform/10`), so they'll silently go stale on the next spec bump. Deriving the URL segment from the resolved version (as done for Hibernate above) would keep them consistent.

gradle/docs-dependencies.gradle
```diff
@@ -56,12 +50,35 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
+        }
+        def jakartaPersistenceVersion = resolveProjectVersion('jakarta.persistence-api')
+        if (jakartaPersistenceVersion) {
+            links << [packages: 'jakarta.persistence.', href: "https://jakarta.ee/specifications/persistence/3.1/apidocs/"]
+        }
+        def jakartaServletVersion = resolveProjectVersion('jakarta.servlet-api')
+        if (jakartaServletVersion) {
+            links << [packages: 'jakarta.servlet.', href: "https://jakarta.ee/specifications/platform/10/apidocs/"]
+        }
+        links << [packages: 'org.grails.datastore.', href: "https://gorm.grails.org/latest/api/"]
```
jdaugherty: Question on these three: since the data modules merged into this repo in 7.0, `org.grails.datastore.*` / `grails.gorm.*` classes are generated in this repo's own aggregate groovydoc. `gorm.grails.org/latest/api/` is the standalone pre-merge GORM site — is that the intended target, and are we sure these prefixes don't hijack references that should resolve locally within the combined docs?

gradle/docs-dependencies.gradle
```diff
@@ -56,12 +50,35 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
+        links << [packages: 'grails.gorm.', href: "https://gorm.grails.org/latest/api/"]
+        links << [packages: 'org.grails.gorm.', href: "https://gorm.grails.org/latest/api/"]
+        links << [packages: 'groovy.', href: "https://docs.groovy-lang.org/latest/html/gapi/"]
```
jdaugherty: `docs.groovy-lang.org/latest` is Groovy 5 while 7.0.x ships Groovy 4, so these links will point at the wrong major version's API docs. The rest of this block pins resolved versions — `GroovySystem.version` is available here to build the versioned URL (`https://docs.groovy-lang.org/docs/groovy-${GroovySystem.version}/html/gapi/`).

build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/GroovydocLinkAuditor.groovy
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Scans generated Groovydoc HTML for malformed navigation and inner-class
+ * links. Kept free of the Gradle API so it can be unit tested directly -
+ * {@code build-logic/docs-core} deliberately keeps Gradle classes off the
+ * test compile classpath.
+ */
+class GroovydocLinkAuditor {
+
+    // Patterns common to Groovydoc navigation failures
+    private static final Map<Pattern, String> NAV_LINK_PATTERNS = [
```
jdaugherty: These nav-link patterns can never match real Groovydoc output. The templates in `groovy-groovydoc` (checked 4.0.24, what 7.0.x uses) emit nav links **double-quoted with a legitimate relative prefix**: `href="${classDoc.relativeRootPath}deprecated-list.html"`. So the single-quoted patterns are dead code — and if the quoting were "fixed", they'd flag every nested class page's perfectly valid nav links, since the prefix is correct there. That also means a "0 violations" run partly reflects this vacuity rather than clean docs.

The inner-class check below is the genuinely valid part: class cross-reference links *are* single-quoted in the templates, which is where the `Outer/Inner.html` vs `Outer.Inner.html` phantom-link bug lives, and the only-flag-if-the-dotted-file-exists guard is a smart false-positive suppressor.

Suggest dropping these patterns (or reworking them to detect links whose *resolved target doesn't exist*, like the inner-class check does). Two related smells that go away with them: the map values (the `replacement` strings) are never read anywhere, and `findViolationsInFile` only records the first match per pattern per file, so the "Found N malformed links" total in the task's exception would undercount.

build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
```diff
@@ -0,0 +1,54 @@
...
+ */
+abstract class AuditGroovydocLinksTask extends DefaultTask {
+
+    @InputDirectory
+    abstract DirectoryProperty getApiDocsDir()
+
+    @TaskAction
+    void auditLinks() {
+        File apiDir = apiDocsDir.get().asFile
+        if (!apiDir.exists()) {
```
jdaugherty: This guard is unreachable: with `@InputDirectory`, Gradle fails input validation before the `@TaskAction` runs when the directory is missing.

build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/AuditGroovydocLinksTask.groovy
```diff
@@ -0,0 +1,54 @@
...
+            logger.warn "API documentation directory does not exist: ${apiDir.absolutePath}"
+            return
+        }
+
+        List<String> violations = GroovydocLinkAuditor.findViolations(apiDir)
+
+        if (!violations.isEmpty()) {
+            violations.take(10).each { logger.error(it) }
+            if (violations.size() > 10) logger.error("... and ${violations.size() - 10} more")
+            throw new org.gradle.api.GradleException("Found ${violations.size()} malformed links in Groovydoc. Please fix the source issue rather than patching. See logs for details.")
```
jdaugherty: Import `org.gradle.api.GradleException` instead of the inline FQCN — it's already on the compile classpath here, and the project convention is explicit imports.

grails-doc/build.gradle
```diff
@@ -125,6 +126,11 @@ combinedGroovydoc.configure { Groovydoc gdoc ->
     gdoc.outputs.dir(gdoc.destinationDir)
 }
 
+def auditGroovydocLinks = tasks.register('auditGroovydocLinks', AuditGroovydocLinksTask) {
+    dependsOn combinedGroovydoc
+    apiDocsDir = combinedGroovydoc.get().destinationDir
```
jdaugherty: `combinedGroovydoc.get()` realizes the task eagerly and requires the explicit `dependsOn` above. Provider wiring keeps it lazy and carries the dependency implicitly:

```groovy
apiDocsDir = project.layout.dir(combinedGroovydoc.map { it.destinationDir })
```

(and then the `dependsOn` line can go.)

gradle/docs-dependencies.gradle
jdaugherty: Actually, we host these directly now on our website, so we shoudlnt' refer to the older docs link

jdaugherty (changes requested):
I think there's more work to do on this one.

borinquenkid:
@jdaugherty nudge

gradle/docs-dependencies.gradle
```diff
@@ -54,14 +48,54 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
+        // since the data modules merged into grails-core in 7.0, those classes are generated
+        // directly in this repo's own aggregate Groovydoc and resolve locally. Mapping them to
+        // the older, standalone gorm.grails.org site would misdirect those in-repo links.
+        def groovyVersion = resolveProjectVersion('groovy')
+        if (groovyVersion) {
+            links << [packages: 'groovy.', href: "https://docs.groovy-lang.org/${groovyVersion}/html/gapi/"]
+            links << [packages: 'org.apache.groovy.', href: "https://docs.groovy-lang.org/${groovyVersion}/html/gapi/"]
+            links << [packages: 'org.codehaus.groovy.', href: "https://docs.groovy-lang.org/${groovyVersion}/html/gapi/"]
+        }
         if (it.ext.has('groovydocLinks')) {
```
jdaugherty: The curated link list never reaches the Groovydoc tasks, so none of the entries added or reordered above actually take effect.

`GroovydocEnhancerPlugin.resolveLinks(Groovydoc gdoc)` reads `gdoc.ext.groovydocLinks` off the **Groovydoc** task, but inside this `doLast` block `it` is the `configureGroovyDoc` task — so line 102 stores the list as an extra property on `configureGroovyDoc` itself, and the `has()` check here only ever sees what this same task set on an earlier run.

Reproduced on this branch:

```
./gradlew :grails-spring:groovydoc
grep -rho "href='https://[^']*'" grails-spring/build/docs/groovydoc | wc -l   # 0
```

Zero external links in the output. Every external reference comes out as a dangling relative link instead — 20 distinct missing targets in that one small module, e.g. `href='../../../org.springframework.context.ApplicationContext.html'`. The published docs show the same thing: `org/grails/web/servlet/mvc/GrailsWebRequest.html` on grails.apache.org/docs/latest/api links `HttpServletRequest` to `../../../../../jakarta.servlet.http.HttpServletRequest.html`, which doesn't exist.

Setting the same list on the Groovydoc task instead does produce external links, which confirms the mechanism works and the wiring is the only thing missing.

One thing worth knowing before restructuring this, because it changes what this PR can claim: Groovy's `links` only feeds `SimpleGroovyClassDoc.getDocUrl`, i.e. doc-comment references, `@see`, and unresolved string types. Signature types (superclass, interfaces, parameters, return types) go through `dolink` in groovydoc's `classDocName.html` template, which builds `classDoc.relativeRootPath + t.fullPathName + '.html'` unconditionally and never consults `links`. In the `:grails-spring` output only 1 of the 21 Spring references switched to an external URL once the wiring was corrected. So fixing this recovers a real but small slice; the bulk of the broken links needs a different fix.

gradle/docs-dependencies.gradle
```diff
@@ -54,14 +48,54 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
-        if (springVersion) {
-            links << [packages: 'org.springframework.core.', href: "https://docs.spring.io/spring-framework/docs/${springVersion}/javadoc-api/"]
-        }
+        // Groovydoc resolves 'links' first-match-wins in declaration order, so the more
+        // specific org.springframework.boot. entry must come before the broader
+        // org.springframework. entry or every Boot class link resolves against the
+        // Framework javadocs instead, where the Boot classes don't exist.
         def springBootVersion = resolveProjectVersion('spring-boot')
         if (springBootVersion) {
             links << [packages: 'org.springframework.boot.', href: "https://docs.spring.io/spring-boot/docs/${springBootVersion}/api/"]
```
jdaugherty: This URL scheme was retired at Boot 3.4, so with the version this branch resolves (3.5.16) it 404s:

| URL | status |
| --- | --- |
| `docs.spring.io/spring-boot/docs/3.2.12/api/org/springframework/boot/SpringApplication.html` | 200 |
| `docs.spring.io/spring-boot/docs/3.4.0/api/org/springframework/boot/SpringApplication.html` | 404 |
| `docs.spring.io/spring-boot/3.5.16/api/java/org/springframework/boot/SpringApplication.html` | 200 |

So the href wants to be `https://docs.spring.io/spring-boot/${springBootVersion}/api/java/`.

The entry predates this PR, but this is the change that moves it above the framework entry specifically so Boot types resolve here — worth having it point somewhere that exists. (The `spring-framework/docs/${version}/javadoc-api/` entry below is fine; verified 200 on 6.2.12.)

gradle/docs-dependencies.gradle
```diff
@@ -54,14 +48,54 @@ def configureGroovyDoc = tasks.register('configureGroovyDoc') {
...
+        }
+        def jakartaValidationVersion = resolveProjectVersion('jakarta.validation-api')
+        if (jakartaValidationVersion) {
+            def specVersion = jakartaValidationVersion.split('\\.').take(2).join('.')
+            links << [packages: 'jakarta.validation.', href: "https://jakarta.ee/specifications/bean-validation/${specVersion}/apidocs/"]
+        }
+        def jakartaPersistenceVersion = resolveProjectVersion('jakarta.persistence-api')
+        if (jakartaPersistenceVersion) {
+            def specVersion = jakartaPersistenceVersion.split('\\.').take(2).join('.')
+            links << [packages: 'jakarta.persistence.', href: "https://jakarta.ee/specifications/persistence/${specVersion}/apidocs/"]
```
jdaugherty: The Jakarta Persistence apidocs are published as *modular* javadoc, so the module name is a path segment. Groovydoc builds the href as `href + type.replace('.', '/') + '.html'` (`SimpleGroovyClassDoc.buildUrl`), which lands on:

- `.../persistence/3.1/apidocs/jakarta/persistence/Entity.html` → 404
- `.../persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/Entity.html` → 200

Same for 3.2. Appending `jakarta.persistence/` to the href fixes it.

Checked the neighbours while I was there: `bean-validation/3.1/apidocs/jakarta/validation/Validator.html` and `platform/11/apidocs/jakarta/servlet/http/HttpServletRequest.html` are both flat and both 200, so those two entries are correct as written. Hibernate `orm/{5.6,6.6,7.1}/javadocs/org/hibernate/Session.html` and `docs.groovy-lang.org/4.0.33/html/gapi/groovy/lang/MetaClass.html` are 200 as well.

build-logic/docs-core/src/main/groovy/org/grails/doc/gradle/GroovydocLinkAuditor.groovy
```diff
@@ -0,0 +1,79 @@
...
+ * help-doc.html, etc.): the bottom nav bar that {@code groovy-groovydoc} emits on
+ * package-summary pages omits the relative-root prefix that every other occurrence
+ * carries, which is a pre-existing quirk of that template rather than a content
+ * issue this repository can fix - checking for it would flag hundreds of instances
+ * of the same known, low-impact tool behavior on every doc build.</p>
+ */
+class GroovydocLinkAuditor {
+
+    // Inner class path issues like Query/Order.Direction.html -> Query.Order.Direction.html.
+    private static final Pattern INNER_CLASS_LINK_PATTERN =
```
jdaugherty: Scope question on the audit as a whole.

"0 violations on a full `aggregateGroovydoc` run" isn't evidence the docs are clean — this pattern only matches slash-separated inner-class paths, and that shape doesn't appear to be emitted by Groovy 4.0.x. Published inner-class links are correctly dotted (`org/grails/datastore/mapping/query/Query.html` links its nested types as `Query.Criterion.html` etc.), and running this exact regex over a locally generated `:grails-spring:groovydoc` tree matches nothing — while that same tree contains 20 distinct hrefs pointing at files that were never generated (`org.springframework.beans.MutablePropertyValues.html`, `org.apache.commons.logging.Log.html`, …).

So the task as written can essentially only report clean, which makes it weak insurance against exactly the class of breakage this PR is about.

`findViolationsInFile` already does the right thing structurally — resolve the href relative to the page's own directory and check the filesystem. Generalizing that to *every* relative `href` (flag when the resolved target doesn't exist, skipping absolute/anchor-only hrefs) would catch the phantom inner-class case, the dangling external-type case, and anything else, with less code than the current special-casing. That would also give the task something to say about whether the link config above is doing its job.

build-logic/docs-core/src/test/groovy/org/grails/doc/gradle/GroovydocLinkAuditorSpec.groovy
```diff
@@ -0,0 +1,69 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.doc.gradle
+
+import spock.lang.Specification
+import spock.lang.TempDir
+
+class GroovydocLinkAuditorSpec extends Specification {
```
jdaugherty, line 26: Non-blocking, and pre-existing: nothing runs these. `build-logic` is an included build, and no workflow invokes its `build` or `test` — a root `./gradlew build` only pulls the plugin jars out of it. The three specs already in this project have the same problem.

Verified they pass locally:

```
cd build-logic && ../gradlew :grails-docs-core:test --tests '*GroovydocLinkAuditorSpec*'
# 3 passed
```

Just flagging that the coverage this adds won't gate anything until build-logic's tests are wired into CI.

jdaugherty:
@borinquenkid I took a pass at fixing the remaining issues.  I believe this is ready to merge. If you agree, please hit merge and we'll work through anything that remains as it comes up.

jdaugherty:
We probably should back port these changes to 7.x too

jdaugherty (approved):
Thanks for picking this up — the previous round is addressed: link ordering is most-specific-first with a comment explaining why, the Jakarta/Hibernate/Groovy URLs derive from resolved versions, the GORM entries are gone, the dead nav-link patterns and the unreachable directory guard are gone, `GradleException` is imported, and `auditGroovydocLinks` is wired through a provider (`--dry-run` confirms `:grails-doc:aggregateGroovydoc` now precedes it in the graph, so it is runnable on a clean checkout).

One blocker and two URL corrections, plus a scope question on the audit itself. The blocker is pre-existing rather than introduced here, but it means the link list this PR curates currently has no effect on the generated docs — details inline.

PR #16009, opened 2026-07-19
Fix afterEvaluate race between applyGrailsBom and GrailsCliGradlePlugin

codeconsole:
## Problem

`GrailsGradlePlugin.applyGrailsBom()` applies `BomPropertyOverridesPlugin` from inside its own `project.afterEvaluate {}` callback. Since `BomPropertyOverridesPlugin.apply()` itself registers a *second*, nested `project.afterEvaluate {}` for the actual override-application logic, and Gradle appends newly-registered `afterEvaluate` listeners to the end of the notification queue that's currently being processed, that nested registration ends up running **after** every `afterEvaluate` callback other plugins registered synchronously during `apply()` — including `GrailsCliGradlePlugin`'s `configureApplicationCommands` (added in #15948), which eagerly resolves the `api`/`implementation`/`runtimeOnly` buckets via the `grailsCliDetect` probe configuration and locks them against further mutation.

In multi-project builds this surfaces as:

```
Cannot mutate the dependencies of configuration ':foo:api' after the configuration's child
configuration ':foo:grailsCliDetect' was resolved. After a configuration has been observed,
it should not be modified.
```

The trigger: resolving one project's `grailsCliDetect` (e.g. while a sibling project's own CLI probe reaches it through a `project(...)` dependency) forces that project to fully configure mid-resolution. If the forced configuration happens to land inside the window between the CLI probe locking its buckets and the re-queued BOM-override callback trying to mutate them, the build fails at configuration time — with any `grailsVersion`/BOM property override active (e.g. `-PgrailsVersion=...`) in a build with `project(...)` dependencies between Grails-plugin projects.

Reproduced against a real multi-project Grails 8 app; confirmed via `--stacktrace` that the failure is exactly this ordering race (`BomManagedVersions.applyTo()` → `DependencyConstraintSet.add()` on `api`, after `GrailsCliGradlePlugin.autoProvisionCliDependencies()` had already resolved `grailsCliDetect`, which extends `api`).

## Fix

Register the `bomPropertyOverrides` extension eagerly (not via `project.plugins.apply(BomPropertyOverridesPlugin)` inside the `afterEvaluate`), and call `BomPropertyOverridesPlugin.applyOverrides()` directly from the *existing*, correctly-ordered `afterEvaluate` callback in `applyGrailsBom`, immediately after the platform is injected — instead of re-deferring through `project.plugins.apply()`. This eliminates the extra deferral hop entirely while preserving the required "platform injected, then overrides computed" sequencing, and keeps `BomPropertyOverridesPlugin` unchanged for its documented standalone use case.

As a side benefit, the `bomPropertyOverrides {}` extension is now created eagerly at `apply()` time rather than inside `afterEvaluate`, so it's now actually usable from a consumer's own `build.gradle` (previously the extension didn't exist yet when the user's own script body ran).

## Testing

- `:grails-gradle-plugins:compileGroovy` — clean compile.
- Published the patched plugin to `mavenLocal`, pointed a real multi-project Grails 8 app's buildscript at it (scoped via `content { includeModule(...) }` so only this one artifact was affected), and reran the previously-failing `-PgrailsVersion=8.0.0-SNAPSHOT` build: every subproject now configures successfully (`BUILD SUCCESSFUL`), where it previously failed with the `InvalidUserCodeException` above.

### DIFF
--- a/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -18,6 +18,7 @@
  */
 package org.grails.gradle.plugin.core
 
+import java.util.function.Function
 import java.util.zip.ZipFile
 
 import grails.util.BuildSettings
@@ -59,6 +60,7 @@ import org.gradle.jvm.toolchain.JavaToolchainService
 import org.gradle.language.jvm.tasks.ProcessResources
 import org.gradle.process.JavaForkOptions
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry
+import org.grails.gradle.plugin.bom.BomPropertyOverridesExtension
 import org.grails.gradle.plugin.bom.BomPropertyOverridesPlugin
 import org.grails.gradle.plugin.commands.GrailsCliGradlePlugin
 import org.grails.gradle.plugin.exploded.ExplodedCompatibilityRule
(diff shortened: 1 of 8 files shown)


codeconsole:
Pushed a follow-up commit addressing review feedback on the first commit.

**What was wrong with the first commit:** it fixed the race by calling `BomPropertyOverridesPlugin.applyOverrides()` directly instead of `project.plugins.apply(BomPropertyOverridesPlugin)`. That broke the plugin's identity/idempotency contract - `project.plugins.findPlugin('org.apache.grails.gradle.bom-property-overrides')` returned `null` even when the override machinery was active (breaking all 4 `BomPlatformFunctionalSpec` cases), and manually recreating the `bomPropertyOverrides` extension could collide with the plugin's own extension creation if a build applied `org.apache.grails.gradle.bom-property-overrides` explicitly alongside a Grails app/plugin. It also shipped with no automated test reproducing the actual multi-project race - the existing suite is all single-project fixtures.

**Two alternative fixes investigated and discarded:**
- Deferring `GrailsCliGradlePlugin`'s classpath probe to `project.gradle.projectsEvaluated` - throws `IllegalResolutionException` ("attempted without an exclusive lock"); resolving a configuration there isn't safe.
- Switching `BomManagedVersions.applyTo()` to register constraints via `Configuration.withDependencies()` instead of eagerly - passes the entire existing suite, but does **not** actually fix the race: registering `withDependencies()` on an already-observed configuration is rejected the same way the eager mutation was, just at a different call site (`DefaultConfiguration.withDependencies`). Only caught by testing against the real multi-project reproduction, not the unit suite.

**The actual fix:** keep `project.plugins.apply(BomPropertyOverridesPlugin)` (restoring identity + idempotency), but call it as a top-level statement in `applyGrailsBom` instead of nesting it inside the method's own `afterEvaluate` callback. Since the `grails.bom = null` opt-out can only be determined once `afterEvaluate` runs (too late to conditionally gate this call without reintroducing the race), the plugin is now applied unconditionally - functionally inert when there's no BOM (auto-detection finds nothing to override), but `findPlugin()` now reports it as applied even when opted out. `BomOptOutFunctionalSpec` is updated to assert the functional outcome (zero override constraints) instead, with the tradeoff documented in both the code and the test for maintainers to weigh in on.

**New regression coverage:** `BomCliMultiprojectRaceFunctionalSpec` - a two-project fixture (`consumer` depending on `producer` via `project(...)`, `producer` carrying an active property-based BOM override) that reproduces the original crash exactly. Confirmed it fails against the pre-fix code with the same `Cannot mutate the dependencies of configuration ... after ... was resolved` error, and passes with the fix.

Full `:grails-gradle-plugins:test` suite (75 tests) passes. Re-verified against the original real-world reproduction (a Grails 8 multi-project app with a `project(...)` dependency chain and a `grailsVersion` property override) by publishing to `mavenLocal` and rebuilding.

jdaugherty:
Once these tests finish, I intend to merge this since it's identified a backwards compatibility issue @jamesfredley

PR #16026, opened 2026-07-20
fix(grails-data-hibernate7): stop conflating decimal and bit precision for numeric columns

borinquenkid:
## Summary

`NumericColumnConstraintsBinder` applied one hand-rolled default precision (`15`, or `126` for Oracle) to every `Number` subtype when no explicit precision/constraints were configured. That conflates two different meanings of "precision":

- **NUMERIC/DECIMAL** (`BigDecimal`/`BigInteger`) precision is a **decimal digit** count.
- **FLOAT/DOUBLE** precision is a **bit** count (IEEE-754) — Hibernate converts whatever decimal-digit value it's given into bits when rendering `float(n)` DDL.

Feeding a decimal-oriented default (Hibernate's own `Size.DEFAULT_PRECISION` = 19, or naively a dialect's already-bit-valued `getFloatPrecision()`/`getDoublePrecision()`) overflows H2/PostgreSQL's 53-bit ceiling and produces DDL those dialects reject — **silently**, since `hibernate.hbm2ddl.auto` only logs the failure rather than throwing, so the table is never created.

**Fix:** `Float`/`Double` now leave precision unset, letting Hibernate fall back to the dialect's own correct float/double DDL type directly. `BigDecimal`/`BigInteger` default from `Dialect.getDefaultDecimalPrecision()`, which every dialect already computes correctly — removing the need for the Oracle-specific special case entirely.

Along the way this also caught a real gap in `SimpleValueBinderSpec`'s test double (`JdbcEnvironment.getDialect()` was never stubbed, returning `null` — previously harmless, now exposed as an NPE since the fix calls `dialect.getDefaultDecimalPrecision()` directly). Fixed the mock rather than adding defensive null-handling in production code for a scenario a real Hibernate bootstrap can't produce.

## Root cause, reproduced with real DDL

Reverting to Hibernate's own `Size.DEFAULT_PRECISION` (19) against H2 produces:

```sql
create table numeric_type_message (
    big_decimal_value numeric(19,2) not null,
    double_value float(64) not null,
    float_value float(64) not null,
    ...
)
```
```
org.h2.jdbc.JdbcSQLSyntaxErrorException: Precision ("64") must be between "1" and "53" inclusive
```
logged as a `WARN` via `ExceptionHandlerLoggedImpl`, not thrown — the table is silently dropped from the schema.

## Test plan

- [x] New `GormNumericTypeColumnPrecisionSpec` (H2, no container): covers the default-precision behavior for `Float`/`Double`/`BigDecimal`, plus a live-DDL check that the table is actually created with valid columns.
- [x] New `GormNumericTypeColumnIntegrationSpec` (Postgres/MySQL/MariaDB via Testcontainers, Oracle excluded per the existing `RLikeHibernate7Spec` precedent): confirms valid, creatable DDL across dialects.
- [x] `NumericColumnConstraintsBinderSpec`: updated for the new `propertyType` parameter, added cases for the Float/Double/BigDecimal split and a parameterized dialect check (H2/Postgres/MySQL/Oracle) proving no per-dialect special case is needed.
- [x] Confirmed RED against unmodified code (H2: `float(64)` rejected; with my first, incorrect attempt at feeding `dialect.getFloatPrecision()` through `setPrecision()`: `float(80)`, still rejected) and GREEN after the fix (`float(24)`, `float(53)`, `numeric(38,2)`).
- [x] Full `grails-data-hibernate7-core` suite: 3020 tests, 0 failures, 24 pre-existing skips.
- [x] `codeStyle` (checkstyle + CodeNarc): clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/ColumnBinder.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/ColumnBinder.java
@@ -119,7 +119,7 @@ public class ColumnBinder {
                 stringColumnConstraintsBinder.bindStringColumnConstraints(column, mappedForm);
             } else if (type != null && Number.class.isAssignableFrom(type)) {
                 PropertyConfig mappedForm = property.getHibernateMappedForm();
-                numericColumnConstraintsBinder.bindNumericColumnConstraints(column, cc, mappedForm);
+                numericColumnConstraintsBinder.bindNumericColumnConstraints(column, cc, mappedForm, type);
             }
         }
 
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/NumericColumnConstraintsBinder.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/NumericColumnConstraintsBinder.java
@@ -25,7 +25,6 @@ import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 
 import org.hibernate.dialect.Dialect;
 import org.hibernate.dialect.H2Dialect;
-import org.hibernate.dialect.OracleDialect;
 import org.hibernate.mapping.Column;
 
 import org.grails.orm.hibernate.cfg.ColumnConfig;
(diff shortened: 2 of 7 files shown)


jdaugherty (approved):
- No test for the Float/Double + min/max constraint case — the behavior change I flagged above goes undocumented by a test. Add one asserting precision stays null in that scenario to lock in the intent.
- GormNumericTypeColumnIntegrationSpec starts three @Shared containers. With @Testcontainers, all shared containers are started for the whole spec regardless of the where: row, so the manual if (!container.isRunning()) container.start() guard is redundant, and you pay for Postgres+MySQL+MariaDB every run. Consider whether all three need to be @Shared/eagerly started, or drop the redundant guard.

PR #16031, opened 2026-07-21
Fix flaky PerTestRecordingSpec by waiting for recording files to stabilize

borinquenkid:
## Summary
`PerTestRecordingSpec > the recordings of the previous two tests are different` is
flaky (~1% of runs per apache/grails-core#16030), with 0 hard failures — same commit,
different outcome on rerun.

## Root cause
`waitForRecordingFiles` polled for candidate recording files by existence + name-match
+ count only. Testcontainers' `VncRecordingContainer.saveRecordingToFile()` copies the
video with a plain `Files.copy(...)` and no atomic temp-file+rename, so the destination
file becomes visible to a directory scan the instant the copy *starts* — a
still-writing (possibly 0-byte) file passes the existence check. Two still-partial
files can register as byte-identical via `Files.mismatch`, intermittently failing the
"recordings are different" assertion.

This builds on James's prior fix in c179aacdc0 ("Address flaky test"), which fixed an
earlier race in *which* files get matched but didn't check that a matched file had
finished being written.

## Fix
`waitForRecordingFiles` now tracks each candidate file's size across polls and only
accepts a file once its size is `> 0` **and** unchanged from the prior poll — proof the
copy has actually finished, not just that a file handle exists. Kept the existing
10s timeout / 500ms poll interval. This is a targeted readiness check, not a generic
retry — no `@Retry`, no blanket rerun.

## Testing
- Real Testcontainers run (Chrome + vnc-recorder, full recording lifecycle):
  `./gradlew :grails-test-examples-geb:integrationTest --tests "org.demo.spock.PerTestRecordingSpec"`
  — BUILD SUCCESSFUL, all 3 iterations passed including the previously-flaky assertion.
- No new unit test added: `waitForRecordingFiles` is a private helper embedded in the
  integration spec with no public-API surface to test independently; the existing
  integration spec is the only legitimate exercise path per the repo's public-API
  testing convention.
- CodeNarc/Checkstyle: clean (`aggregateStyleViolations`, no violations).

### DIFF
--- a/grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
+++ b/grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
@@ -1,0 +1,137 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package grails.plugin.geb
+
+import java.time.LocalDateTime
+
+import org.testcontainers.containers.BrowserWebDriverContainer
+import org.testcontainers.containers.VncRecordingContainer
+
+import geb.Browser
+import geb.test.GebTestManager
+import spock.lang.Specification
+
+import static org.testcontainers.containers.BrowserWebDriverContainer.VncRecordingMode
+
+class WebDriverContainerHolderSpec extends Specification {
+
+    WebDriverContainerHolder holder = new WebDriverContainerHolder(new GrailsGebSettings(LocalDateTime.now()))
+
+    void 'stop() resets container, browser and testManager on the happy path'() {
+        given: 'a holder with an initialized container'
+        def container = Mock(BrowserWebDriverContainer)
+        holder.container = container
+        holder.browser = Mock(Browser)
+        holder.testManager = Mock(GebTestManager)
+
+        when: 'the holder is stopped'
+        holder.stop()
+
+        then: 'the underlying container is stopped'
+        1 * container.stop()
+
+        and: 'all held state is cleared'
+        holder.container == null
+        holder.browser == null
+        holder.testManager == null
+        !holder.initialized
+    }
+
+    void 'stop() still resets all held state when container.stop() throws'() {
+        given: 'a holder whose container fails to stop cleanly'
+        def container = Mock(BrowserWebDriverContainer)
+        container.stop() >> { throw new IllegalStateException('boom') }
+        holder.container = container
+        holder.browser = Mock(Browser)
+        holder.testManager = Mock(GebTestManager)
+
+        when: 'the holder is stopped'
+        holder.stop()
+
+        then: 'the exception from stop() propagates'
+        thrown(IllegalStateException)
+
+        and: 'held state is still cleared, so a broken container is never reported as initialized'
+        holder.container == null
+        holder.browser == null
+        holder.testManager == null
+        !holder.initialized
+    }
+
+    void 'restartVncRecordingContainer() does nothing when recording is disabled'() {
+        given:
+        holder.settings.recordingMode = VncRecordingMode.SKIP
+        holder.settings.restartRecordingContainerPerTest = true
+        def container = Mock(BrowserWebDriverContainer)
+        holder.container = container
+
+        when:
+        holder.restartVncRecordingContainer()
+
+        then:
+        0 * container._
+    }
+
+    void 'restartVncRecordingContainer() does nothing when per-test restart is disabled'() {
+        given:
+        holder.settings.recordingMode = VncRecordingMode.RECORD_ALL
+        holder.settings.restartRecordingContainerPerTest = false
+        def container = Mock(BrowserWebDriverContainer)
+        holder.container = container
+
+        when:
+        holder.restartVncRecordingContainer()
+
+        then:
+        0 * container._
+    }
+
+    void 'restartVncRecordingContainer() does nothing when no container has been initialized'() {
+        given:
+        holder.settings.recordingMode = VncRecordingMode.RECORD_ALL
+        holder.settings.restartRecordingContainerPerTest = true
+        holder.container = null
+
+        expect: 'no exception is thrown even though there is nothing to restart'
+        holder.restartVncRecordingContainer()
+    }
+
+    void 'restartVncRecordingContainer() swallows a failure from the current recording container instead of propagating it'() {
+        given: 'a container whose active VNC recording container fails to stop'
+        holder.settings.recordingMode = VncRecordingMode.RECORD_ALL
+        holder.settings.restartRecordingContainerPerTest = true
+        def container = Mock(BrowserWebDriverContainer)
+        holder.container = container
+
+        def vncContainer = Mock(VncRecordingContainer)
+        vncContainer.stop() >> { throw new IllegalStateException('vnc container refused to stop') }
+        def vncField = BrowserWebDriverContainer.getDeclaredField('vncRecordingContainer')
+        vncField.accessible = true
+        vncField.set(container, vncContainer)
+
+        when: 'restarting the recording container'
+        holder.restartVncRecordingContainer()
+
+        then: 'the failure is logged and swallowed rather than breaking test execution'
+        noExceptionThrown()
+
+        and: "the field is left untouched - it still points at the container that just failed to stop, not a container that never started"
+        vncField.get(container).is(vncContainer)
+    }
+}
--- a/grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
+++ b/grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
@@ -23,27 +23,57 @@ import java.time.LocalDateTime
 import org.testcontainers.containers.BrowserWebDriverContainer
 import org.testcontainers.containers.VncRecordingContainer
 
+import org.openqa.selenium.WebDriver
+
 import geb.Browser
 import geb.test.GebTestManager
+import spock.lang.Shared
 import spock.lang.Specification
 
 import static org.testcontainers.containers.BrowserWebDriverContainer.VncRecordingMode
 
 class WebDriverContainerHolderSpec extends Specification {
 
+    private static final String GEB_SYSTEM_PROPERTY_PREFIX = 'grails.geb.'
+
+    // GrailsGebSettings' constructor parses live `grails.geb.*` system properties (forwarded
+    // from `grails.geb.*` project properties by gradle/test-config.gradle), so an externally
+    // supplied value - e.g. a malformed `grails.geb.recording.mode` in a developer's gradle
+    // properties - could otherwise make the `holder` field initializer below throw before any
+    // feature method runs. Clearing them for the duration of this spec keeps it hermetic.
+    @Shared
+    private Map<String, String> savedGebSystemProperties
+
+    def setupSpec() {
+        savedGebSystemProperties = System.getProperties().stringPropertyNames()
+                .findAll { it.startsWith(GEB_SYSTEM_PROPERTY_PREFIX) }
+                .collectEntries { [(it): System.getProperty(it)] }
+        savedGebSystemProperties.keySet().each { System.clearProperty(it) }
+    }
+
+    def cleanupSpec() {
+        savedGebSystemProperties.each { key, value -> System.setProperty(key, value) }
+    }
+
     WebDriverContainerHolder holder = new WebDriverContainerHolder(new GrailsGebSettings(LocalDateTime.now()))
 
     void 'stop() resets container, browser and testManager on the happy path'() {
         given: 'a holder with an initialized container'
         def container = Mock(BrowserWebDriverContainer)
+        def driver = Mock(WebDriver)
+        def browser = Mock(Browser)
+        browser.driver >> driver
         holder.container = container
-        holder.browser = Mock(Browser)
+        holder.browser = browser
         holder.testManager = Mock(GebTestManager)
 
         when: 'the holder is stopped'
         holder.stop()
 
-        then: 'the underlying container is stopped'
+        then: 'the driver session is quit while the container backing it is still up'
+        1 * driver.quit()
+
+        and: 'the underlying container is stopped'
         1 * container.stop()
 
         and: 'all held state is cleared'
@@ -68,6 +98,29 @@ class WebDriverContainerHolderSpec extends Specification {
         thrown(IllegalStateException)
 
         and: 'held state is still cleared, so a broken container is never reported as initialized'
+        holder.container == null
+        holder.browser == null
+        holder.testManager == null
+        !holder.initialized
+    }
+
+    void 'stop() still quits the driver and stops the container when quitting the driver throws'() {
+        given: 'a holder whose driver refuses to quit cleanly'
+        def container = Mock(BrowserWebDriverContainer)
+        def driver = Mock(WebDriver)
+        driver.quit() >> { throw new IllegalStateException('driver refused to quit') }
+        def browser = Mock(Browser)
+        browser.driver >> driver
+        holder.container = container
+        holder.browser = browser
+        holder.testManager = Mock(GebTestManager)
+
+        when: 'the holder is stopped'
+        holder.stop()
+
+        then: "the driver failure doesn't stop the container from being stopped or state from being reset"
+        noExceptionThrown()
+        1 * container.stop()
         holder.container == null
         holder.browser == null
         holder.testManager == null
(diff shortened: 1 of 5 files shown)


grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
```diff
@@ -123,21 +123,36 @@ class PerTestRecordingSpec extends ContainerGebSpec {
...
+
+            // Testcontainers copies each recording with a plain, non-atomic
+            // stream copy, so a file can appear in the directory scan above
+            // while still 0 bytes or only partially written. Only treat a
+            // recording as ready once its size is non-zero and unchanged
+            // since the previous poll, which means the copy has finished.
+            readyFiles = candidateFiles.findAll { File file ->
+                long size = currentSizes[file.absolutePath]
+                size > 0 && previousSizes[file.absolutePath] == size
+            }
```
jdaugherty: The premise in this comment doesn't hold, so this check is guarding against something that can't happen.

`GebRecordingTestListener.afterIteration` calls `container.afterTest(...)` synchronously on the test thread. That reaches `BrowserWebDriverContainer.retainRecordingIfNeeded`, which calls `vncRecordingContainer.saveRecordingToFile(recordingFile)`:

```java
try (InputStream inputStream = streamRecording()) {
    Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
}
```

and `streamRecording()` blocks on `execInContainer("ffmpeg", ...)` before the copy starts. Because the spec is `@Stepwise`, features 1 and 2 have both returned from `afterIteration` — and therefore from `Files.copy` — before this feature method runs. `Files.copy` lacking a temp-file+rename only matters when a reader can race a live writer; there is no live writer here.

Separately, even on its own terms "size unchanged since the previous poll" isn't proof a copy finished — a docker tar stream that stalls for more than the 500ms poll interval under CI I/O contention gives two equal readings mid-copy. If a readiness check is genuinely needed, it would want N consecutive stable polls rather than one.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
```diff
@@ -123,21 +123,36 @@ class PerTestRecordingSpec extends ContainerGebSpec {
             long pollIntervalMillis = 500L
     ) {
         long deadline = System.currentTimeMillis() + timeoutMillis
-        List<File> recordingFiles = []
+        Map<String, Long> previousSizes = [:]
```
jdaugherty: `previousSizes` starts empty, so on the first poll `previousSizes[path] == size` is `null == size` for every candidate and `readyFiles` is always empty — no matter how long the recordings have been sitting complete on disk.

That means every run of this spec now sleeps at least one full `pollIntervalMillis` (500ms) before it can succeed, where previously the first poll could break immediately. Seeding `previousSizes` from an initial scan before entering the loop would avoid the unconditional penalty if this check is kept.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
```diff
@@ -123,21 +123,36 @@ class PerTestRecordingSpec extends ContainerGebSpec {
...
 
-            if (recordingFiles.size() >= minFileCount) {
+            if (readyFiles.size() >= minFileCount) {
                 break
             }
+            previousSizes = currentSizes
             sleep(pollIntervalMillis)
         }
-        return recordingFiles
+        return readyFiles
```
jdaugherty: On timeout this returns `readyFiles`, and `candidateFiles` is discarded. That makes a real failure harder to diagnose than before: previously the returned list held whatever was on disk, so the `recordingFiles.size() >= minFileCount` failure message showed the files that were found. Now a timeout typically reports an empty list, with no signal distinguishing "no recording was ever written" from "two recordings existed but never satisfied the stability check" — and those point at different bugs.

If the check stays, please surface the candidates on timeout (return them, or at minimum log their names and sizes) so the next flaky run leaves behind something actionable.

jdaugherty (commented):
Thanks for digging into this one — the polling helper is definitely the right place to look, and the stability-tracking code itself is clean and readable.

My concern is with the diagnosis. I walked the recording lifecycle and I don't think the partial-write race described in the PR body can actually occur:

- `GebRecordingTestListener.afterIteration` calls `containerHolder.container.afterTest(...)` synchronously, on the test thread, in Spock's per-iteration callback.
- `BrowserWebDriverContainer.afterTest` → `retainRecordingIfNeeded` → `vncRecordingContainer.saveRecordingToFile(recordingFile)`.
- `saveRecordingToFile` is `try (InputStream in = streamRecording()) { Files.copy(in, file.toPath(), REPLACE_EXISTING); }`, and `streamRecording()` itself blocks on `execInContainer("ffmpeg", ...)` before the copy even begins.

So both `Files.copy` calls have fully returned before the third feature method starts — nothing is still writing to those files while `waitForRecordingFiles` scans. `Files.copy` not being atomic only matters if a reader can observe the file concurrently with the writer, and here the spec is `@Stepwise`, so there is no concurrent writer to observe. That makes me think the change doesn't remove the flake so much as change which assertion reports it.

Worth noting too that the linked flaky report shows `0/1419` hard failures and `1/1419` flaky, so there's no captured failure output in the issue — which means the root cause is still a hypothesis. Before landing a fix I'd want to know which assertion actually failed, because the two candidates want very different fixes:

1. **`recordingFiles.size() >= minFileCount` failed** — only one recording landed. `GebRecordingTestListener` swallows `NotFoundException` for `/newScreen.mp4` with a debug log, which is exactly what happens when the freshly restarted VNC container hadn't captured any frames yet for `ffmpeg` to re-encode. If that's the cause, the fix belongs in `restartVncRecordingContainer` (wait until the new container is actually capturing), and this PR makes the symptom worse: the wait now burns the full 10s before failing.
2. **`Files.mismatch(...) != -1` failed** — the two recordings really were byte-identical. Requiring `size > 0` rules out the both-files-empty case, but not the case that seems likelier to me: two degenerate near-blank captures that `ffmpeg -vcodec libx264 -movflags faststart` encodes to the same bytes. Those are non-zero and size-stable, so they sail through the new check. If this is the cause, the byte-comparison assertion is itself the flake and should be replaced with something that asserts the actual framework contract (a distinct, non-trivial recording per feature method) rather than raw byte inequality of ffmpeg output.

Suggestion: rather than land this, could you pull the failure output for the flaky run from the CI/Develocity build scan (or reproduce with the recording container restart forced to fail) so we can confirm which assertion broke? Happy to help chase it. I'd rather not add a readiness check for a race that the synchronous call chain rules out — it costs latency on every run and leaves the real cause in place.

One process note: `PerTestRecordingSpec` exists on `7.0.x` too, with an older `waitForRecordingFiles` that doesn't even have the per-run directory scoping from c179aacdc0. Whatever we land here, it'd be good to decide whether it goes to `7.0.x` first and merges forward, so the two copies of this helper don't keep diverging.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
borinquenkid: You're right, and thanks for walking through the source — confirmed the same thing against Testcontainers 2.0.5: `saveRecordingToFile`'s `Files.copy` has already returned by the time this feature method runs, so there's no concurrent writer for a directory scan to race against.

Reverted the stability-polling entirely in 24a0b59c and reworked the diagnosis around your more likely explanation (a VNC container that's connected but hasn't captured meaningful frames yet after a restart). That also surfaced a real bug in `WebDriverContainerHolder#restartVncRecordingContainer` — the `vncRecordingContainer` field was updated to the new container *before* `start()` succeeded, so a swallowed `start()` failure left it pointing at a container that never actually started. Fixed alongside.

I want to flag honestly: I didn't manage to pull the original CI/Develocity failure output to confirm which assertion actually broke (it's a <1% flake with no captured output in #16030). 24a0b59c hedges across both of your candidate diagnoses rather than landing a fix confirmed against the real failure — happy to hold this for a repro if you'd rather chase the actual signal first.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
borinquenkid: Good catch, and moot now — 24a0b59c reverts the stability-polling entirely (`previousSizes` and all), so `waitForRecordingFiles` is back to its original form and there's no first-poll penalty to worry about. Thanks for spotting it regardless.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
borinquenkid: Also moot after 24a0b59c reverted the stability-polling — `waitForRecordingFiles` returns the raw scan result again, so a timeout still reports whatever files were actually found, same as before this PR. Good instinct on wanting failure diagnostics preserved, though; I'll keep it in mind if we ever add readiness logic here again.

borinquenkid:
@jdaugherty thanks for the deep read on this one — replied inline on each of the three specific comments, and wanted to close the loop on the top-level review too.

**On the diagnosis:** you were right that the original stability-polling premise doesn't hold. Verified the same thing you did against the Testcontainers 2.0.5 source (`saveRecordingToFile`'s `Files.copy` has returned before this feature method ever runs), and reverted that change entirely in 24a0b59c.

**On confirming which assertion actually failed:** I wasn't able to pull the original CI/Develocity output — it's a <1% flake and #16030 has no captured failure info attached, so I don't have a build scan to chase. Rather than sit on this without that signal, 24a0b59c hedges across both of the candidate diagnoses you laid out instead of landing a fix confirmed against the real failure:

1. Fixed a real bug in `WebDriverContainerHolder#restartVncRecordingContainer` — the `vncRecordingContainer` field was updated to the new container *before* `start()` succeeded, so a swallowed `start()` failure left it silently pointing at a container that never started. This doesn't cover the "connected but not yet capturing frames" timing gap you described — I don't have a way to force that deterministically without a real repro.
2. Replaced the byte-difference-only check with a minimum-size floor per recording. Kept the byte-mismatch check on top rather than dropping it — it also catches two recordings accidentally resolving to the *same* file, which the size floor alone wouldn't. The floor isn't a guess anymore either: 08800e4 grounds `MIN_MEANINGFUL_RECORDING_BYTES` in two real local runs against the container, where every genuine recording measured 75KB-855KB (5,000 bytes stays more than an order of magnitude below the smallest of those).
3. Added `WebDriverContainerHolderSpec` (08800e4) covering `stop()`'s try/finally reset and `restartVncRecordingContainer()`'s guard clauses / exception-swallowing — no Docker required, since neither had unit coverage before.

I'd still take a real repro over this if one turns up — happy to hold or follow up further if you'd rather chase the actual signal first.

**On 7.0.x:** agreed the two copies of `waitForRecordingFiles` shouldn't keep diverging. Proposing we land this on 8.0.x first, and I'll follow up with a backport PR to 7.0.x that also picks up the per-run directory scoping from c179aacdc0, since 7.0.x doesn't have that either. Let me know if you'd rather sequence it the other way.

borinquenkid:
@jdaugherty nudge

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -440,13 +447,18 @@ class WebDriverContainerHolder {
...
+                // swallowed to avoid breaking test execution - so if the field were already
+                // pointing at newVncContainer by then, every subsequent saveRecordingToFile()
+                // would silently target a container that never actually started.
                 def newVncContainer = new VncRecordingContainer(container)
                         .withVncPassword('secret')
                         .withVncPort(5900)
                         .withVideoFormat(settings.recordingFormat)
-                field.set(container, newVncContainer)
                 newVncContainer.start()
+                field.set(container, newVncContainer)
```
jdaugherty, line 487: If `newVncContainer.start()` throws, the swallow below leaves this field still pointing at the **old** container — which `vncContainer.stop()` above has already stopped *and removed* (testcontainers' `GenericContainer.stop()` removes the container). The next `afterIteration` → `saveRecordingToFile()` then hits a docker `NotFoundException` whose message names the missing container id, not `/newScreen.mp4`, so `GebRecordingTestListener`'s guard rethrows it and a passing test gets reported as an error — the exact outcome the swallow is meant to prevent.

Consider clearing the field (`field.set(container, null)`) in the failure path so the save is skipped cleanly instead of throwing.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -440,13 +447,18 @@ class WebDriverContainerHolder {
...
+                // "Connected" wait strategy times out), the exception below is deliberately
+                // swallowed to avoid breaking test execution - so if the field were already
+                // pointing at newVncContainer by then, every subsequent saveRecordingToFile()
+                // would silently target a container that never actually started.
                 def newVncContainer = new VncRecordingContainer(container)
                         .withVncPassword('secret')
                         .withVncPort(5900)
                         .withVideoFormat(settings.recordingFormat)
-                field.set(container, newVncContainer)
                 newVncContainer.start()
```
jdaugherty: Related: when `start()` throws after the docker container has already been created (e.g. the `Connected` wait strategy times out), nothing references or stops `newVncContainer` anymore — a live ffmpeg recorder stays attached to the browser container's VNC port until Ryuk reaps it at JVM exit. The previous ordering self-healed: the field held the failed container, so the next restart's `vncContainer.stop()` cleaned it up.

A `try { newVncContainer.start(); field.set(container, newVncContainer) } catch (Exception e) { newVncContainer.stop(); throw e }` around just the start would cover both this and the stale-field issue above.

grails-geb/src/testFixtures/groovy/grails/plugin/geb/WebDriverContainerHolder.groovy
```diff
@@ -88,11 +88,18 @@ class WebDriverContainerHolder {
...
-        containerConf = null
+        try {
+            container?.stop()
+        } finally {
+            // Reset state even if stop() throws - otherwise isInitialized() keeps reporting
+            // true for a container that's actually broken, and a later reinitialize() call
+            // would see matchesCurrentContainerConfiguration() as a false positive without
+            // ever attempting to recover.
+            container = null
+            browser = null
```
jdaugherty, line 109: While touching `stop()`: the comment in `reinitialize()` says "The driver is explicitly quit by us in stop() method", but `stop()` never calls `browser?.driver?.quit()` — it only nulls the reference. Since `cacheDriver=false` and `quitDriverOnBrowserReset=false` were set precisely on that promise (disabling Geb's own cleanup), every container recycle abandons a `RemoteWebDriver` and its HTTP client/connection pool. Worth either quitting the driver here (inside the `try`, before `container?.stop()`) or correcting the comment.

grails-test-examples/geb/src/integration-test/groovy/org/demo/spock/PerTestRecordingSpec.groovy
```diff
@@ -85,12 +85,35 @@ class PerTestRecordingSpec extends ContainerGebSpec {
...
+        // WebDriverContainerHolder#restartVncRecordingContainer) is guaranteed to have
+        // connected, but not to have captured more than a frame or two by the time a fast
+        // iteration finishes. Two such near-blank captures can encode to identical,
+        // non-zero, stable-sized bytes via ffmpeg - passing a raw byte-difference check
+        // without actually being distinct, meaningful recordings. Requiring a sensible
+        // minimum size asserts the real framework contract - a real, played-out recording -
+        // rather than raw byte inequality of whatever ffmpeg happened to produce.
         def firstRecording = recordingFiles.find { it.name.contains('setup_running_a_test_to_create_a_recording') }
         def secondRecording = recordingFiles.find { it.name.contains('setup_running_a_second_test_to_create_another') }
+        firstRecording.length() > MIN_MEANINGFUL_RECORDING_BYTES
```
jdaugherty, line 109: Two concerns with the size check and its justification:

1. The comment misstates the old check's semantics: two identical near-blank captures make `Files.mismatch` return `-1`, which *failed* the pre-existing `!= -1` assertion — they couldn't "pass a raw byte-difference check".
2. In the exact near-blank scenario the comment describes (recorder just restarted, fast iteration finishes before more than a frame or two is captured), these size assertions also fail — so that flake is relabeled rather than eliminated. And the 5,000-byte floor is calibrated from only two local runs; a short, low-motion encode on a slow-to-restart CI iteration could legitimately land below it.

If the real contract is "a played-out recording exists per test", something like waiting for the file to stop growing / reach the floor (with a timeout) before asserting would be less flaky than a static minimum on whatever ffmpeg had flushed at iteration end.

grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+    void 'restartVncRecordingContainer() swallows a failure from the current recording container instead of propagating it'() {
+        given: 'a container whose active VNC recording container fails to stop'
+        holder.settings.recordingMode = VncRecordingMode.RECORD_ALL
+        holder.settings.restartRecordingContainerPerTest = true
+        def container = Mock(BrowserWebDriverContainer)
+        holder.container = container
+
+        def vncContainer = Mock(VncRecordingContainer)
+        vncContainer.stop() >> { throw new IllegalStateException('vnc container refused to stop') }
+        def vncField = BrowserWebDriverContainer.getDeclaredField('vncRecordingContainer')
```
jdaugherty, line 184: This test injects and asserts state through testcontainers' private `vncRecordingContainer` field via reflection, which runs against the repo test conventions (CLAUDE.md rule 9: test via public APIs, never bypass the public surface). It also pins a third-party private implementation detail — a field rename in a testcontainers upgrade breaks this test with no public-API contract explaining why. The production code admittedly uses the same reflection, but the assertion here could target observable behavior instead (e.g. that a subsequent recording save/restart still behaves after a failed stop) rather than re-reading the private field.

grails-geb/src/test/groovy/grails/plugin/geb/WebDriverContainerHolderSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+
+import geb.Browser
+import geb.test.GebTestManager
+import spock.lang.Specification
+
+import static org.testcontainers.containers.BrowserWebDriverContainer.VncRecordingMode
+
+class WebDriverContainerHolderSpec extends Specification {
+
+    WebDriverContainerHolder holder = new WebDriverContainerHolder(new GrailsGebSettings(LocalDateTime.now()))
```
jdaugherty, line 58: `GrailsGebSettings` reads live JVM system properties, and `gradle/test-config.gradle` forwards every `grails.geb.*` project property into test JVMs — so an externally supplied value (e.g. a malformed `grails.geb.recording.mode` or `.format` from a developer's gradle properties) makes this field initializer throw in `VncRecordingMode.valueOf` and errors all five features before any assertion runs. Constructing the settings hermetically here (explicit values, or saving/clearing the relevant system properties around the spec) would isolate the tests from the environment.

PR #16032, opened 2026-07-21
Fix flaky RunningApplicationProcessSpec by widening the stop() timeout budget

borinquenkid:
## Summary
`RunningApplicationProcessSpec > stop terminates a running process and removes the PID
file` is flaky (~2% of runs per apache/grails-core#16030), 0 hard failures — same
commit, different outcome on rerun.

## Root cause
`RunningApplicationProcess.stop()` already behaves correctly: it calls
`process.destroy()`, blocks on `ProcessHandle.onExit()`, falls back to
`destroyForcibly()`, and its final check queries the OS process table directly
(`!process.isAlive()`) — accurate regardless of reaper-thread scheduling. The bug is
purely a timing budget: the test's 15s total wasn't enough headroom for the JVM's
process-reaper notification under CI contention (many parallel Gradle forks each
spawning child processes), occasionally causing `stop()` to report `STILL_RUNNING`
instead of `STOPPED`.

This file was already patched twice before for a related Windows race
(6c76333ea9/a13c38c3fd) — those touched the test's tail assertion but never this
timeout budget.

## Fix
No production code changed — `awaitExit()`'s fallback is already correct, so hardening
it further would add complexity without addressing the actual bottleneck. Bumped the
test's `stop(pidFile, 15000)` call to `30000`ms, giving realistic headroom on a
contended runner. Smallest reviewable diff.

## Testing
- `:grails-shell-cli:test` (full module): 3 separate runs (forced-fresh,
  isolated-fresh-JVM `--rerun-tasks --no-daemon`, and cached) all BUILD SUCCESSFUL,
  target spec passing in each.
- No new test coverage required (no production change); reviewed that the existing
  test still fully exercises real stop-terminates-process behavior.
- CodeNarc/Checkstyle: clean, including a full repo-wide `aggregateStyleViolations`
  run (740 tasks).

### DIFF
--- a/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
+++ b/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
@@ -208,8 +208,8 @@ class RunningApplicationProcessSpec extends Specification {
         expect:
         RunningApplicationProcess.isRunning(pidFile)
 
-        when:
-        def result = RunningApplicationProcess.stop(pidFile, 15000)
+        when: "a generous timeout budget gives headroom for reaper-notification lag on a contended CI runner"
+        def result = RunningApplicationProcess.stop(pidFile, 30000)
 
         then:
         result == RunningApplicationProcess.StopResult.STOPPED
--- a/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
+++ b/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit
 
 import spock.lang.Specification
 import spock.lang.TempDir
+import spock.util.concurrent.PollingConditions
 
 /**
  * Tests for {@link RunningApplicationProcess}, the PID file based mechanism that lets
@@ -208,14 +209,21 @@ class RunningApplicationProcessSpec extends Specification {
         expect:
         RunningApplicationProcess.isRunning(pidFile)
 
-        when: "a generous timeout budget gives headroom for reaper-notification lag on a contended CI runner"
-        def result = RunningApplicationProcess.stop(pidFile, 30000)
+        when: "the application is stopped"
+        // Same budget the shipped stop-app command passes (grails-profiles/base/commands/stop-app.groovy)
+        def result = RunningApplicationProcess.stop(pidFile, 30_000)
 
         then:
         result == RunningApplicationProcess.StopResult.STOPPED
         !pidFile.exists()
-        // stop() observes exit via the ProcessHandle; the Process object's own reaper can lag
-        // behind on Windows, so wait for it rather than sampling isAlive() immediately
-        process.waitFor(10, TimeUnit.SECONDS)
+
+        and: "the JVM's own process bookkeeping catches up with the OS-level exit stop() already observed"
+        // Process.isAlive() queries OS process state directly, like stop()'s own awaitExit() fallback.
+        // Process.waitFor(timeout) instead blocks on the JVM's internal reaper thread, whose notification
+        // can lag under CI contention independently of whether the process has actually exited - polling
+        // the OS-backed check avoids that lag causing a spurious failure here.
+        new PollingConditions(timeout: 30, initialDelay: 0, delay: 0.1).eventually {
+            assert !process.isAlive()
+        }
     }
 }
--- a/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
+++ b/grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
@@ -218,12 +218,14 @@ class RunningApplicationProcessSpec extends Specification {
         !pidFile.exists()
 
         and: "the JVM's own process bookkeeping catches up with the OS-level exit stop() already observed"
-        // Process.isAlive() queries OS process state directly, like stop()'s own awaitExit() fallback.
-        // Process.waitFor(timeout) instead blocks on the JVM's internal reaper thread, whose notification
-        // can lag under CI contention independently of whether the process has actually exited - polling
-        // the OS-backed check avoids that lag causing a spurious failure here.
+        // process.isAlive() (java.lang.Process) is backed by the same async reaper-thread completion
+        // as process.waitFor(timeout) - it is NOT a live OS query, so polling it would still be exposed
+        // to the same notification lag under CI contention. process.toHandle().isAlive() delegates to
+        // ProcessHandle's native isAlive0() check instead, which queries the OS process table directly
+        // on every call, exactly like stop()'s own awaitExit() fallback - so it is unaffected by reaper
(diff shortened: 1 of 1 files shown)


grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
```diff
@@ -208,8 +208,8 @@ class RunningApplicationProcessSpec extends Specification {
         expect:
         RunningApplicationProcess.isRunning(pidFile)
 
-        when:
-        def result = RunningApplicationProcess.stop(pidFile, 15000)
+        when: "a generous timeout budget gives headroom for reaper-notification lag on a contended CI runner"
```
jdaugherty: The block description explains why the number is large rather than what the stimulus is, so the Spock report ends up reading `when: a generous timeout budget gives headroom for reaper-notification lag on a contended CI runner` for a step whose stimulus is simply "the application is stopped". It also undersells the change: `30000` is not an arbitrary generous value, it is exactly what the shipped `stop-app` passes (`grails-profiles/base/commands/stop-app.groovy:31`). Saying that makes this an alignment with production instead of a number we grew until the test stopped failing.

```groovy
when: "the application is stopped"
// Same budget the shipped stop-app command passes (grails-profiles/base/commands/stop-app.groovy)
def result = RunningApplicationProcess.stop(pidFile, 30_000)
```

Also `30_000` for consistency with the `3_600_000L` literals already in this file. If you want to go further, hoisting the value onto `RunningApplicationProcess` (`static final long DEFAULT_STOP_TIMEOUT_MILLIS = 30_000L`) and using it from both `stop-app.groovy` and here would leave one source of truth instead of two copies of the same magic number - fine to skip if you would rather keep the diff minimal.

grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
```diff
@@ -208,8 +208,8 @@ class RunningApplicationProcessSpec extends Specification {
...
+        def result = RunningApplicationProcess.stop(pidFile, 30000)
```
jdaugherty: This is the part I would like confirmed before merge.

`stop()` can only return `STILL_RUNNING` if `awaitExit()` returns false twice: once after the `timeoutMillis` wait on `onExit()`, and again 5s after `destroyForcibly()` (note `Math.min(timeoutMillis, 5000L)` - the second budget does not move with this change, so the effective total goes 20s -> 35s). Both of those returns fall through to `!process.isAlive()`, a direct OS liveness check. For the target here - `sleep 60` / `ping -n 60 127.0.0.1`, hit with `destroy()` and then `destroyForcibly()` - the OS still reporting it alive ~20s later is hard to credit. `NOT_RUNNING` is ruled out by the `expect: isRunning(pidFile)` immediately above. So I do not think the 15s budget is what produced the failure.

The candidate I would check first is the tail condition seven lines down, `process.waitFor(10, TimeUnit.SECONDS)`. It returns `false` - a failing Spock condition - when the test JVM's own bookkeeping for its child has not caught up within 10s, which is precisely the reaper-notification lag the description blames, and this PR leaves its 10s budget untouched. That tail is also what 6c76333ea9/a13c38c3fd were fixing. If that is the condition failing on CI, this change will not stop the flake and the fix belongs there instead - an unbounded `process.waitFor()`, or a `PollingConditions` wait on `!process.isAlive()`.

Worth ruling out as well: `stop()` discards the boolean returned by `destroy()`/`destroyForcibly()`, so a signal that never landed is indistinguishable from a slow shutdown. Not something to fix in this PR, but it is the other way `STILL_RUNNING` happens without any timeout being too short.

The Spock output from one flaky run settles which of these it is, since it prints the rendered condition and the actual `StopResult`.

jdaugherty (commented):
Initial AI Review: 

Thanks for chasing this one down. Aligning the test with the timeout the shipped command actually uses is the right instinct: `stop-app` passes `30000` (`grails-profiles/base/commands/stop-app.groovy:31`), so after this change the test exercises the same budget a user gets rather than an arbitrarily tighter one. That is a stronger justification than the CI-headroom framing, and I would like the code to say so.

My concern is the root-cause analysis, because it decides whether this fixes the flake or just widens the window. The description says `awaitExit()`'s fallback "queries the OS process table directly (`!process.isAlive()`) - accurate regardless of reaper-thread scheduling". If that is true, then a `sleep 60` / `ping -n 60` that was signalled at t=0 is long gone by t=15s, the fallback returns `true`, and `stop()` reports `STOPPED` - so a 15s budget could not have produced `STILL_RUNNING`. Either the fallback is not authoritative in this scenario (in which case the budget was not the bottleneck), or the condition that actually failed on CI was not `result == StopResult.STOPPED`. Details inline.

Could you pull the failure output from one of the flaky runs before this lands? The rows in #16030 link to build scans on develocity.apache.org, and Spock renders the failing condition along with the actual `StopResult`. That single line tells us whether this change is the fix or a mitigation.

One scoping question: `7.0.x` still has `stop(pidFile, 15000)` and, at the tail, the older `!process.isAlive()` assertion - neither 6c76333ea9 nor a13c38c3fd is an ancestor of `7.0.x`. Should 7.0.x get the same hardening, or are we deliberately only touching 8.0.x because that is all the dashboard tracks? (Minor: those two commits are the same change - identical diffstat, both only on 8.0.x - so "patched twice before" reads as two separate attempts at the problem.)

grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
borinquenkid: Done — the `when:` label now reads `"the application is stopped"`, and there's a comment noting `30_000` matches the value `stop-app.groovy:31` passes in production, rather than framing it as an arbitrary generous number. Kept the underscore style consistent with the `3_600_000L` literals already in the file. Left the `DEFAULT_STOP_TIMEOUT_MILLIS` hoist out per your "fine to skip" note, to keep the diff minimal.

grails-shell-cli/src/test/groovy/org/grails/cli/gradle/RunningApplicationProcessSpec.groovy
borinquenkid: You were right — the 15s→30s `stop()` budget wasn't the fix. Replaced the tail `process.waitFor(10, TimeUnit.SECONDS)` with a `PollingConditions` wait on the OS-backed liveness check, same idea as your suggested alternatives.

One more round worth flagging: my first pass at that polled `process.isAlive()` (`java.lang.Process`), reasoning it was OS-backed like `stop()`'s own fallback. Checked the JDK 21 source to confirm and that's wrong — `Process#isAlive()` is `!hasExited`, set only when the same async `ProcessHandleImpl` reaper-thread completion that backs `waitFor()`/`onExit()` resolves. It's not a live query, so polling it would've just re-widened the same budget from 10s to 30s rather than removing the dependency you flagged. Fixed to poll `process.toHandle().isAlive()` instead, which delegates to `ProcessHandleImpl`'s native `isAlive0()` — a genuine per-call OS check, matching what `awaitExit()`'s fallback actually relies on. That version is pushed now.

jdaugherty (approved):

PR #16033, opened 2026-07-21
Fix flaky grails-views-gson tests caused by shared static test state

borinquenkid:
## Summary
~20 test methods across `JsonViewHelperSpec`, `JsonViewTestSpec`, `JsonApiSpec`, and
`ExpandSpec` (module `:grails-views-gson:test`) show the same failure+flakiness
pattern (~2% failures, ~1% flakiness per apache/grails-core#16030). This PR started
from the hypothesis that this was caused by a class-keyed GORM cache leak across
specs that silently shared `@Entity` classes via unqualified same-package
resolution.

## Status of the root-cause hypothesis: not confirmed, and probably not the cause
jdaugherty's review went to verify the mapping-context-isolation premise directly and
found it doesn't hold: `JsonViewTest.mappingContext` is a plain (non-`@Shared`) trait
property, and Spock instantiates a spec fresh for every feature method — so **every
spec already builds a brand-new `KeyValueMappingContext` per feature regardless of
this PR**, and re-registers whatever entity classes it uses into it every time. A
throwaway probe spec confirmed this empirically (3 features → 3 distinct
`KeyValueMappingContext`s → 3 distinct `PersistentEntity` instances for the same
`Class`). So "no class is registered into two independently-built mapping contexts"
isn't an invariant this codebase has, or one a rename can establish — whether a
*second* spec also registers the class is N vs 2N of something that already happens
by design.

**No one has produced a stack trace from an actual failing run**, despite several
independent attempts (see Testing below). The dashboard in #16030 is the only source
for the failure counts, and its underlying per-test data isn't reachable via the
GitHub API or CI artifacts.

jdaugherty's alternative theory remains the more likely explanation and is
**unaddressed by this PR**: every flagged method *within* a given spec in #16030's
dashboard shares an identical failure count (e.g. all 8 `JsonViewHelperSpec` methods
at 20/2165), which looks like whole-spec/fixture-level failures (a `setup()` throw or
a fork-level failure) rather than independent per-assertion staleness.

## What this PR actually does
Given the above, this PR is a **test hygiene / naming-safety improvement**, not a
flakiness fix: unqualified same-package binding to another spec's fixture classes is
a real footgun independent of whether it explains #16030, and it's worth closing on
its own merits.

Per jdaugherty's design suggestion, each spec that needs its own copy of a shared
fixture (`Team`/`Player`/`PlayerWithAge`/`Person`/`Address`/`Author`) now gets it in
**its own sub-package**, keeping the original simple class names, rather than
prefixing the class names as earlier revisions of this PR did:

- `grails.plugin.json.view.expand` — `ExpandSpec`'s own `Team`/`Player`
- `grails.plugin.json.view.include` — `IncludeAssociationsSpec`'s own `Team`/`Player`
- `grails.plugin.json.view.halembedded` — `HalEmbeddedSpec`'s own
  `Team`/`Player`/`Person`/`Address`
- `grails.plugin.json.view.iterable` — `IterableRenderSpec`'s own `Team`/`Player`
- `grails.plugin.json.view.maprender` — `MapRenderSpec`'s own
  `Team`/`Player`/`PlayerWithAge`
- `grails.plugin.json.view.nullrendering` — `NullRenderingSpec`'s own `Team`/`Player`,
  plus a local `Child` POGO replacing the previously-unaddressed unqualified borrow of
  `PogoDeepRenderingSpec`'s `Child2`
- `grails.plugin.json.view.api.handleassociations` —
  `JsonApiHandleAssociationsSpec`'s own `Author`

This buys the isolation for almost none of the diff churn the earlier
prefix-rename approach caused: JSON API `type` and HAL `href` values are derived from
the entity's *simple* class name (`PersistentEntity.decapitalizedName`,
`GrailsNameUtils.getPropertyName(clazz)`), which is unchanged by moving a class to a
different package — so none of the expected-JSON assertions in these specs needed to
change. It's also self-enforcing: a future spec added to `grails.plugin.json.view`
that types `new Player(...)` no longer silently binds to another spec's fixture,
because the classes live in different packages now and same-package resolution can't
reach them.

**Two same-package borrows are intentionally left as-is**, not moved to a
sub-package:
- `TemplateInheritanceSpec` still uses `JsonViewHelperSpec`'s `Player`/`Circular`
  directly (documented with a comment on the class). The published
  `child2`/`child4`/`circular` `.gson` templates under
  `grails-views-gson/grails-app/views` import
  `grails.plugin.json.view.Player`/`.Circular` directly, so the model passed to
  `render()` in this spec has to be that exact class — there's no fixture-only
  workaround.
- `api/PaginationSpec` imports `grails.plugin.json.view.Book`
  (`JsonViewTemplateEngineSpec`'s `@Linkable` POGO) explicitly, rather than typing a
  bare `Book` that would silently resolve to `JsonApiSpec`'s own same-package `Book`.
  An explicit import isn't the silent-binding trap this PR closes elsewhere, so it's
  left alone per review feedback.

Also unchanged from earlier revisions:
- `JsonViewTest` (the published `grails-views-gson` trait) is back to its pre-PR
  shape — no production code changes. An earlier version of this PR added
  `cleanup()`/`cleanupSpec()` directly to it, which breaks any downstream spec
  implementing `JsonViewTest` that declares its own `cleanup()` (a Groovy trait
  method must be `public`; Spock's AST transform lowers the visibility of the
  fixture methods it generates — the two are irreconcilable).
- `JsonApiSpec` keeps the public `SuperHero.clearConstraintsMapCache()` API (in place
  of the earlier reflection hack into `Validateable`'s internal state) and its own
  `cleanup()`.

## Testing
- Full `:grails-views-gson:test` (178 tests): 0 failures, verified against this
  revision's sub-package restructuring.
- `codeStyle` (checkstyle + CodeNarc on `src/main`): clean. checkstyle/CodeNarc on
  test sources are skipped project-wide, unchanged by this commit.
- Prior rounds' reproduction attempts (none successful, kept for context — see
  earlier revisions of this description): all ~50 failed CI runs on `8.0.x` in the
  30 days before this PR was opened showed no `grails-views-gson` failure; brute-force
  repetition (single JVM/fork, 60 iterations, ~44 min) against the pre-PR base found
  0 failures; jdaugherty independently got 178/178 across 4 runs, with and without
  this PR's changes.

Given all of the above, **this PR should be evaluated as a robustness/naming
cleanup, not a fix for the flakiness in #16030**. That issue's root cause is still
open — jdaugherty's whole-spec-fixture-failure theory is the most promising lead and
would need a captured stack trace or a targeted repro to confirm.

### DIFF
--- a/grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
+++ b/grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
@@ -21,6 +21,7 @@ package grails.plugin.json.view.test
 
 import groovy.json.JsonSlurper
 import groovy.text.Template
+import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 
 import org.springframework.beans.factory.annotation.Autowired
@@ -41,6 +42,7 @@ import grails.web.mapping.LinkGenerator
 import grails.web.mime.MimeUtility
 import org.grails.datastore.mapping.keyvalue.mapping.config.KeyValueMappingContext
 import org.grails.datastore.mapping.model.MappingContext
+import org.grails.validation.ConstraintEvalUtils
 
 /**
  * A trait that test classes can implement to add support for easily testing JSON views
@@ -103,6 +105,43 @@ trait JsonViewTest {
         templateEngine.setJsonApiIdRenderStrategy(jsonApiIdRenderStrategy)
         return templateEngine
     }()
+
+    /**
+     * Resets the {@link ConstraintEvalUtils} default-constraints cache after every feature
+     * method so state from one test cannot leak into the next test that happens to run in the
+     * same JVM/fork.
+     *
+     * <p>{@link ConstraintEvalUtils} memoizes the default GORM constraints map in a single
+     * JVM-wide static field keyed by the identity of the last {@code Config} seen. Since each
+     * spec implementing this trait typically builds its own {@code GrailsApplication}/{@code
+     * Config}, a stale entry left behind by one spec can otherwise be picked up by another.
+     * This is cheap to recompute, so it is safe to clear after every feature.</p>
+     */
+    void cleanup() {
+        ConstraintEvalUtils.clearDefaultConstraints()
+    }
+
+    /**
+     * Tears down any {@code GrailsApplication} cached by {@code org.grails.testing.GrailsUnitTest}
+     * once the whole spec has finished, so a later spec running in the same JVM/fork always starts
+     * from a clean application context.
+     *
+     * <p>This is deliberately a {@code cleanupSpec()} (once per spec class), not a per-feature
+     * {@code cleanup()}: {@code GrailsUnitTest} intentionally builds its {@code GrailsApplication}
+     * lazily and reuses it across every feature of a spec (it is expensive to build and some
+     * traits, e.g. {@code DataTest}, register beans into it once per spec). Tearing it down after
+     * every feature would rebuild it before the next feature could see those spec-scoped beans.</p>
+     *
+     * <p>{@code GrailsUnitTest} lives in {@code grails-testing-support-core}, a test-only
+     * dependency this trait cannot reference directly since it ships as part of the main
+     * {@code grails-views-gson} artifact, so the call is made dynamically only when present.</p>
+     */
+    @CompileDynamic
+    void cleanupSpec() {
+        if (metaClass.respondsTo(this, 'cleanupGrailsApplication')) {
+            cleanupGrailsApplication()
+        }
+    }
 
     /**
      * Render a template for the given source
--- a/grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
+++ b/grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
@@ -21,7 +21,6 @@ package grails.plugin.json.view.test
 
 import groovy.json.JsonSlurper
 import groovy.text.Template
-import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 
 import org.springframework.beans.factory.annotation.Autowired
@@ -42,7 +41,6 @@ import grails.web.mapping.LinkGenerator
 import grails.web.mime.MimeUtility
 import org.grails.datastore.mapping.keyvalue.mapping.config.KeyValueMappingContext
 import org.grails.datastore.mapping.model.MappingContext
-import org.grails.validation.ConstraintEvalUtils
 
 /**
  * A trait that test classes can implement to add support for easily testing JSON views
@@ -105,43 +103,6 @@ trait JsonViewTest {
         templateEngine.setJsonApiIdRenderStrategy(jsonApiIdRenderStrategy)
         return templateEngine
     }()
-
-    /**
-     * Resets the {@link ConstraintEvalUtils} default-constraints cache after every feature
-     * method so state from one test cannot leak into the next test that happens to run in the
-     * same JVM/fork.
-     *
-     * <p>{@link ConstraintEvalUtils} memoizes the default GORM constraints map in a single
-     * JVM-wide static field keyed by the identity of the last {@code Config} seen. Since each
-     * spec implementing this trait typically builds its own {@code GrailsApplication}/{@code
-     * Config}, a stale entry left behind by one spec can otherwise be picked up by another.
-     * This is cheap to recompute, so it is safe to clear after every feature.</p>
-     */
-    void cleanup() {
-        ConstraintEvalUtils.clearDefaultConstraints()
-    }
-
-    /**
-     * Tears down any {@code GrailsApplication} cached by {@code org.grails.testing.GrailsUnitTest}
-     * once the whole spec has finished, so a later spec running in the same JVM/fork always starts
-     * from a clean application context.
-     *
-     * <p>This is deliberately a {@code cleanupSpec()} (once per spec class), not a per-feature
-     * {@code cleanup()}: {@code GrailsUnitTest} intentionally builds its {@code GrailsApplication}
-     * lazily and reuses it across every feature of a spec (it is expensive to build and some
-     * traits, e.g. {@code DataTest}, register beans into it once per spec). Tearing it down after
-     * every feature would rebuild it before the next feature could see those spec-scoped beans.</p>
-     *
-     * <p>{@code GrailsUnitTest} lives in {@code grails-testing-support-core}, a test-only
-     * dependency this trait cannot reference directly since it ships as part of the main
-     * {@code grails-views-gson} artifact, so the call is made dynamically only when present.</p>
-     */
-    @CompileDynamic
-    void cleanupSpec() {
-        if (metaClass.respondsTo(this, 'cleanupGrailsApplication')) {
-            cleanupGrailsApplication()
-        }
-    }
 
     /**
      * Render a template for the given source
--- a/grails-views-gson/src/test/groovy/grails/plugin/json/view/ExpandSpec.groovy
+++ b/grails-views-gson/src/test/groovy/grails/plugin/json/view/ExpandSpec.groovy
@@ -19,6 +19,7 @@
 package grails.plugin.json.view
 
 import tools.jackson.databind.json.JsonMapper
+import grails.persistence.Entity
 import grails.plugin.json.view.test.JsonRenderResult
 import grails.plugin.json.view.test.JsonViewTest
 import org.grails.datastore.mapping.core.Session
@@ -32,23 +33,23 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
     JsonMapper objectMapper = JsonMapper.builder().build()
 
     void setup() {
-        mappingContext.addPersistentEntities(Team, Player)
+        mappingContext.addPersistentEntities(ExpandTeam, ExpandPlayer)
     }
 
     void 'Test expand parameter allows expansion of child associations'() {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(Team, 1L) >> new Team(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 1L)
+        mockSession.retrieve(ExpandTeam, 1L) >> new ExpandTeam(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 1L)
 
-        def player = new Player(name: 'Cantona', team: teamProxy)
+        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
 
         def templateText = '''
-            import grails.plugin.json.view.*
-            
+            import grails.plugin.json.view.ExpandPlayer as Player
+
             @Field Player player
-            
+
             json g.render(player)
         '''
 
@@ -84,15 +85,13 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(Team, 1L) >> new Team(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 1L)
+        mockSession.retrieve(ExpandTeam, 1L) >> new ExpandTeam(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 1L)
 
-        def player = new Player(name: 'Cantona', team: teamProxy)
+        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
         def templateText = '''
-            import grails.plugin.json.view.*
-            
             @Field Map map
-            
+
             json g.render(map)
         '''
 
@@ -119,13 +118,13 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(Team, 1L) >> new Team(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 1L)
+        mockSession.retrieve(ExpandTeam, 1L) >> new ExpandTeam(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 1L)
 
-        def player = new Player(name: 'Cantona', team: teamProxy)
+        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
 
         def templateText = '''
-            import grails.plugin.json.view.*
+            import grails.plugin.json.view.ExpandPlayer as Player
             model {
                 Player player
             }
@@ -140,7 +139,7 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
             {
                 "_links": {
                     "self": {
-                        "href": "http://localhost:8080/player",
+                        "href": "http://localhost:8080/expandPlayer",
                         "hreflang": "en",
                         "type": "application/hal+json"
                     }
@@ -161,7 +160,7 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
                     "team": {
                         "_links": {
                             "self": {
-                                "href": "http://localhost:8080/team/1",
+                                "href": "http://localhost:8080/expandTeam/1",
                                 "hreflang": "en",
                                 "type": "application/hal+json"
                             }
@@ -171,7 +170,7 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
                 },
                 "_links": {
                     "self": {
-                        "href": "http://localhost:8080/player",
+                        "href": "http://localhost:8080/expandPlayer",
                         "hreflang": "en",
                         "type": "application/hal+json"
                     }
@@ -185,18 +184,18 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(Team, 9L) >> new Team(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 9L)
-        def player = new Player(name: 'Cantona', team: teamProxy)
+        mockSession.retrieve(ExpandTeam, 9L) >> new ExpandTeam(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 9L)
+        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
         player.id = 3
 
         when: 'The domain is rendered with expand parameters'
         JsonRenderResult result = render('''
-            import grails.plugin.json.view.*
+            import grails.plugin.json.view.ExpandPlayer as Player
             model {
                 Player player
             }
-            
+
             json jsonapi.render(player, [expand: 'team'])
         ''', [player: player])
 
@@ -204,7 +203,7 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
         objectMapper.readTree(result.jsonText) == objectMapper.readTree('''
             {
                 "data": {
-                    "type": "player",
+                    "type": "expandPlayer",
                     "id": "3",
                     "attributes": {
                         "name": "Cantona"
@@ -212,21 +211,21 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
                     "relationships": {
                         "team": {
                             "links": {
-                                "self": "/team/9"
+                                "self": "/expandTeam/9"
                             },
                             "data": {
-                                "type": "team",
+                                "type": "expandTeam",
                                 "id": "9"
                             }
                         }
                     }
                 },
                 "links": {
-                    "self": "/player/3"
+                    "self": "/expandPlayer/3"
                 },
                 "included": [
                     {
-                        "type":"team",
+                        "type":"expandTeam",
                         "id": "9",
                         "attributes": {
                             "titles": null,
@@ -241,11 +240,33 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
                             }
                         },
                         "links": {
-                            "self": "/team/9"
+                            "self": "/expandTeam/9"
                         }
                     }
                 ]
             }
         ''')
+    }
+}
+
+@Entity
+class ExpandTeam {
+    String name
+    ExpandPlayer captain
+    List players
+    List<String> titles
+    @SuppressWarnings('unused')
+    static hasMany = [players: ExpandPlayer]
+}
+
+@Entity
+class ExpandPlayer {
+    Long version
+    String name
+    @SuppressWarnings('unused')
+    static belongsTo = [team: ExpandTeam]
+
+    static constraints = {
+        name nullable: false
     }
 }
--- a/grails-views-gson/src/test/groovy/grails/plugin/json/view/ExpandSpec.groovy
+++ b/grails-views-gson/src/test/groovy/grails/plugin/json/view/ExpandSpec.groovy
@@ -19,7 +19,8 @@
 package grails.plugin.json.view
 
 import tools.jackson.databind.json.JsonMapper
-import grails.persistence.Entity
+import grails.plugin.json.view.expand.Player
+import grails.plugin.json.view.expand.Team
 import grails.plugin.json.view.test.JsonRenderResult
 import grails.plugin.json.view.test.JsonViewTest
 import org.grails.datastore.mapping.core.Session
@@ -33,20 +34,20 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
     JsonMapper objectMapper = JsonMapper.builder().build()
 
     void setup() {
-        mappingContext.addPersistentEntities(ExpandTeam, ExpandPlayer)
+        mappingContext.addPersistentEntities(Team, Player)
     }
 
     void 'Test expand parameter allows expansion of child associations'() {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(ExpandTeam, 1L) >> new ExpandTeam(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 1L)
+        mockSession.retrieve(Team, 1L) >> new Team(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 1L)
 
-        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
+        def player = new Player(name: 'Cantona', team: teamProxy)
 
         def templateText = '''
-            import grails.plugin.json.view.ExpandPlayer as Player
+            import grails.plugin.json.view.expand.Player
 
             @Field Player player
 
@@ -85,10 +86,10 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
         given: 'An entity with a proxy association'
         def mockSession = Mock(Session)
         mockSession.getMappingContext() >> mappingContext
-        mockSession.retrieve(ExpandTeam, 1L) >> new ExpandTeam(name: 'Manchester United')
-        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, ExpandTeam, 1L)
+        mockSession.retrieve(Team, 1L) >> new Team(name: 'Manchester United')
+        def teamProxy = mappingContext.proxyFactory.createProxy(mockSession, Team, 1L)
 
-        def player = new ExpandPlayer(name: 'Cantona', team: teamProxy)
+        def player = new Player(name: 'Cantona', team: teamProxy)
         def templateText = '''
             @Field Map map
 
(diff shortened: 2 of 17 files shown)


grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
```diff
@@ -104,6 +106,43 @@ trait JsonViewTest {
...
+     * method so state from one test cannot leak into the next test that happens to run in the
+     * same JVM/fork.
+     *
+     * <p>{@link ConstraintEvalUtils} memoizes the default GORM constraints map in a single
+     * JVM-wide static field keyed by the identity of the last {@code Config} seen. Since each
+     * spec implementing this trait typically builds its own {@code GrailsApplication}/{@code
+     * Config}, a stale entry left behind by one spec can otherwise be picked up by another.
+     * This is cheap to recompute, so it is safe to clear after every feature.</p>
+     */
+    void cleanup() {
```
jdaugherty: Blocking: this breaks downstream specs, and there is no workaround available to the user.

`JsonViewTest` is a published trait. Because a Groovy trait method becomes an interface method, an implementing class must declare it `public` — but Spock's AST transform lowers the visibility of fixture methods. The two requirements are irreconcilable, so any application spec that implements `JsonViewTest` and declares its own `cleanup()` no longer compiles. Reproduced on this branch by adding a spec with a `cleanup()` body to `grails-views-gson/src/test`:

```
> Task :grails-views-gson:compileTestGroovy FAILED
startup failed:
.../TmpUserCleanupProbeSpec.groovy: 29: The method cleanup should be public as it implements
the corresponding method from interface grails.plugin.json.view.test.JsonViewTest
. At [29:5]  @ line 29, column 5.
       void cleanup() {
       ^
```

`cleanup()` is one of the most commonly used Spock fixtures, and the user's only remedy is to delete theirs. The same applies to `cleanupSpec()` below. It's already biting inside this PR: `JsonApiSpec` had to give up its own `cleanup()`, not because the teardown was unnecessary but because it can no longer declare one.

If per-feature isolation is wanted for this module's specs, it belongs in a test-scoped fixture — a base spec or trait under `grails-views-gson/src/test`, or in `grails-testing-support-views-gson`, which is already a `testImplementation` dependency and is the natural home for test-lifecycle behaviour. That keeps the shipped API unchanged.

grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
```diff
@@ -104,6 +106,43 @@ trait JsonViewTest {
...
+        ConstraintEvalUtils.clearDefaultConstraints()
```
jdaugherty: Separately from the API concern: I don't think this clearing is load-bearing, because the cache is already reset once per spec class today.

`ConstraintEvalUtils`' static initialiser registers its own reset as a *preserved* shutdown operation:

```groovy
// grails-core/.../org/grails/validation/ConstraintEvalUtils.groovy:37
static {
    ShutdownOperations.addOperation({ clearDefaultConstraints() } as Runnable, true)
}
```

`true` is `preserveForNextShutdown`, so `ShutdownOperations.runOperations()` re-adds it after each run and it fires every time. `GrailsUnitTest.cleanupGrailsApplication()` calls `runOperations()` (`GrailsUnitTest.groovy:184`), and nothing in the repo calls `ShutdownOperations.resetOperations()` — the only thing that would drop a preserved operation. Since the cache can only be non-empty if `ConstraintEvalUtils` has been loaded, and loading is what registers the reset, any populated cache is guaranteed to be cleared at the end of the spec that populated it. So cross-spec leakage of this cache is already impossible for `GrailsUnitTest` specs; this change only moves it from per-spec to per-feature, and a single spec has a single `Config`.

The stated mechanism also doesn't fit the failure rate. `getDefaultConstraints(config)` recomputes unless `configId == System.identityHashCode(config)`, so returning a stale map requires an identity-hash collision between two *consecutive* `Config` instances — roughly 1 in 2^31, not ~1%.

If you have a reproduction that shows otherwise I'd like to see it, since that would point at a real bug in `ConstraintEvalUtils`' identity-hash keying — which would be worth fixing there (e.g. a `WeakHashMap` keyed on the `Config` itself) rather than papered over from a test trait.

grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy
```diff
@@ -104,6 +106,43 @@ trait JsonViewTest {
...
+     * lazily and reuses it across every feature of a spec (it is expensive to build and some
+     * traits, e.g. {@code DataTest}, register beans into it once per spec). Tearing it down after
+     * every feature would rebuild it before the next feature could see those spec-scoped beans.</p>
+     *
+     * <p>{@code GrailsUnitTest} lives in {@code grails-testing-support-core}, a test-only
+     * dependency this trait cannot reference directly since it ships as part of the main
+     * {@code grails-views-gson} artifact, so the call is made dynamically only when present.</p>
+     */
+    @CompileDynamic
+    void cleanupSpec() {
```
jdaugherty: This method is a no-op and can be dropped, along with the `@CompileDynamic` annotation and the `groovy.transform.CompileDynamic` import.

`cleanupGrailsApplication()` is already invoked after every spec class that implements `GrailsUnitTest`, via the global Spock extension in `grails-testing-support-core`:

```groovy
// org/grails/testing/spock/TestingSupportExtension.groovy:51 (registered in
// META-INF/services/org.spockframework.runtime.extension.IGlobalExtension)
if (GrailsUnitTest.isAssignableFrom(spec.reflection)) {
    spec.addCleanupSpecInterceptor(cleanupContextInterceptor)
}
```

`CleanupContextInterceptor` calls `cleanupGrailsApplication()` in a `finally` block, and Spock attaches cleanup-spec interceptors to a synthetic `CLEANUP_SPEC` `MethodInfo` that runs for every spec whether or not one is declared (`PlatformSpecRunner.createMethodForDoRunCleanupSpec`, spock-core 2.4). So for a `GrailsUnitTest` spec this trait method runs inside `invocation.proceed()`, nulls `_grailsApplication`, and the interceptor's own call then finds null and no-ops — net behaviour identical to 8.0.x. For a spec that doesn't implement `GrailsUnitTest`, `respondsTo` is false and nothing happens at all.

That matters beyond tidiness: this is the half of the change that adds a second breaking fixture method to the published trait, for no behavioural gain.

grails-views-gson/src/test/groovy/grails/plugin/json/view/ExpandSpec.groovy
```diff
@@ -32,23 +33,23 @@ class ExpandSpec extends Specification implements JsonViewTest, GrailsUnitTest {
     JsonMapper objectMapper = JsonMapper.builder().build()
 
     void setup() {
-        mappingContext.addPersistentEntities(Team, Player)
+        mappingContext.addPersistentEntities(ExpandTeam, ExpandPlayer)
```
jdaugherty: This is the part of the PR I'd keep — an unqualified same-package reference silently binding to another spec's `@Entity` classes is a real trap, and giving `ExpandSpec` its own fixtures is the right shape of fix.

But it's applied to one of four specs doing exactly this, and not to the two with the worst numbers in #16030. `Team` and `Player` are declared once, in `JsonViewHelperSpec.groovy:663` and `:672`, and after this PR they are still registered into three other independently-built `KeyValueMappingContext` instances:

- `IncludeAssociationsSpec` — `import grails.plugin.json.view.*` plus `addPersistentEntities(Player, Team)`
- `HalEmbeddedSpec` — same-package unqualified reference, `addPersistentEntities(Team, Player)`
- `IterableRenderSpec` — same-package unqualified reference, `addPersistentEntities(Player, Team)`, in five separate features

The same pattern holds for two other classes:

- `grails.plugin.json.view.api.Author` is declared in `api/JsonApiSpec.groovy:449` and also registered by `api/JsonApiHandleAssociationsSpec` (`addPersistentEntities(Author, PublishedBook, Publisher)`)
- `Person` is declared in `EmbeddedAssociationsSpec.groovy:180` and also registered by `HalEmbeddedSpec` (`addPersistentEntities(Person, Parent)`)

Per the dashboard the two worst specs are `JsonViewHelperSpec` (8 methods, 20 failures) and `JsonApiSpec` (7 methods, 20 failures) — and both still share entity classes with another spec after this change. `ExpandSpec` (19 failures) is the only one decoupled. So if class-keyed GORM state is the root cause, the flakiness should survive this PR.

Could you extend the same treatment to those specs? Given `Team`/`Player` are wanted by four specs, a shared read-only fixture file plus one mapping context, or per-spec copies as done here, would both work — the important thing is that no two independently-built mapping contexts see the same `Class`. A dedicated fixture source file would also make the ownership obvious and stop the next spec from picking them up by accident.

grails-views-gson/src/test/groovy/grails/plugin/json/view/api/JsonApiSpec.groovy
```diff
@@ -24,39 +24,23 @@ import grails.plugin.json.view.test.JsonRenderResult
...
     JsonMapper objectMapper = JsonMapper.builder().build()
 
     void setup() {
-        ConstraintEvalUtils.clearDefaultConstraints()
-        clearConstraintsMapCache(SuperHero)
+        // Resets SuperHero's own cached constraints map (see Validateable#clearConstraintsMapCache).
+        // JsonViewTest#cleanup() already resets the shared ConstraintEvalUtils cache after every
+        // feature, but SuperHero's cache is specific to this spec's own Validateable command
+        // object, so it is reset here too.
+        SuperHero.clearConstraintsMapCache()
```
jdaugherty, line 36: Good change — dropping the `Validateable$Trait$StaticFieldHelper` reflection in favour of the public `clearConstraintsMapCache()` is exactly right, and it matches how `grails-validation`'s own specs do it (`ValidateableTraitSpec:46`, `ValidateableTraitAdHocSpec:37`). It also brings this spec in line with the project rule about testing through public APIs.

One consequence worth being explicit about: with the local `cleanup()` gone, `SuperHero`'s cache is now only reset on the way *in*, so the last feature of this spec leaves it populated for the rest of the fork. That's harmless today — `SuperHero` is declared and used only in this file — but it's the opposite of the isolation-by-default goal, and note that you can't fix it by re-adding a `cleanup()` here as long as the trait declares one. Another argument for keeping the fixture methods out of the published trait.

jdaugherty (commented):
Before taking a look at this, I had AI take a look.  Here's it's comments: 

Thanks for digging into #16030 — the triage write-up is genuinely useful, and disproving the trait-static hypothesis with isolated probes rather than assuming was the right instinct. Two things I'd like to resolve before this lands.

**1. The `JsonViewTest` changes are in `src/main`.** The PR body says "No production code changed", but `grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy` ships in the `grails-views-gson` artifact and is the documented way applications test JSON views. Adding `cleanup()`/`cleanupSpec()` to it is a breaking API change for downstream specs (details inline — I reproduced the compile failure locally). I also believe both fixture methods are redundant with machinery that already exists in `grails-testing-support-core`; sources cited inline.

**2. The `ExpandSpec` decoupling looks right, but it's applied to one of four specs with the same problem** — and not to the two with the highest failure counts. Details inline on `ExpandSpec`.

On evidence: the counts in #16030 are worth a second look. Every flagged method within a spec has an *identical* count (all 8 `JsonViewHelperSpec` methods 20/2037, all 7 `JsonApiSpec` methods 20/2023, all 4 `ExpandSpec` methods 19/2034, `JsonViewTestSpec` 19/2029). Identical per-method counts across an entire spec is the signature of ~20 CI runs in which those specs failed *wholesale* — a fixture/`setup()` throw or a fork-level failure — rather than independent per-assertion flakiness. That's a different shape of bug than a cache returning stale data, and it's the strongest clue available. Could you pull the actual stack trace from one of those failing runs?

The reason I'm pushing on that: a green suite doesn't discriminate between hypotheses here. I ran `:grails-views-gson:test` on this branch (178/178) and then again three times with the `JsonViewTest` change reverted and only the `ExpandSpec` change kept, single JVM, `-PforkEveryUnitTest=0 -PtestBisect` to maximise shared state — 178/178 every time. 8.0.x passes ~99% of the time on its own, so neither result tells us whether the leak is closed.

What I'd suggest: land the `ExpandSpec` and `JsonApiSpec` changes (both are improvements on their own merits), extend the class-decoupling to the remaining specs, and drop the `JsonViewTest` trait change.

borinquenkid:
Thanks for the detailed review — all five points below are addressed in `a42b48780f`, but I want to be upfront that the central question you raised isn't resolved, only investigated further.

**The 5 inline points:**
1. Blocking (trait `cleanup()`/`cleanupSpec()` breaks downstream specs): reverted. `JsonViewTest` is back to its pre-PR shape.
2. `ConstraintEvalUtils` reset not load-bearing: agreed, moot now that point 1 is reverted — your `ShutdownOperations`/`preserveForNextShutdown` analysis was correct.
3. `cleanupSpec()` no-op: removed along with `@CompileDynamic`.
4. `ExpandSpec`-only decoupling incomplete: extended to `IncludeAssociationsSpec`, `HalEmbeddedSpec` (`Team`/`Player` and `Person`), `IterableRenderSpec`, `MapRenderSpec`, `NullRenderingSpec`, and `JsonApiHandleAssociationsSpec` (`Author`) — each gets its own spec-prefixed entity classes now, so no class is registered into two independently-built mapping contexts.
5. `JsonApiSpec`'s `SuperHero` cache left populated after the last feature: `cleanup()` restored (it only had to go in the first version because the trait declared one).

**On your stack-trace ask — still open.** I couldn't produce one either. What I checked:
- All ~50 failed CI runs on `8.0.x` in the last 30 days (the window #16030's dashboard covers): zero show a `grails-views-gson` failure.
- Ran the four originally-flagged specs together in a single JVM/fork (`-PmaxTestParallel=1 -PforkEveryUnitTest=0`) against the pre-this-PR base commit, 60 iterations (~44 min): 0 failures. Same result you got with your 4 runs.

So three independent attempts now (yours, and this one twice) have failed to reproduce it locally, and I couldn't find a corroborating CI job failure either. `testlens-app`'s dashboard is the only source for the failure counts, and its underlying per-test data isn't reachable via the GitHub API or CI artifacts — I couldn't get past the aggregate numbers to see what's actually throwing.

Given that, I've updated the PR description to stop asserting the class-cache-leak explanation as settled and instead flag it as the working hypothesis it is. The entity-decoupling change is worth keeping on its own merits (shared `Class` objects across independently-built mapping contexts is a real risk regardless), but I think your whole-spec-fixture-failure theory is still the more likely explanation for the *specific* pattern in #16030, and this PR doesn't confirm or rule it out either way. Open to suggestions on how to get real evidence here — happy to try something more targeted than brute-force repetition if you have an idea for what would actually trigger it.

grails-views-gson/src/test/groovy/grails/plugin/json/view/HalEmbeddedSpec.groovy
```diff
@@ -421,16 +421,17 @@ class HalEmbeddedSpec extends Specification implements JsonViewTest {
 
     void 'test hal embedded with associations that have GORM embedded properties'() {
         given: 'A domain class with embedded associations'
-        mappingContext.addPersistentEntities(Person, Parent)
-        def p = new Person(name: 'Robert')
+        mappingContext.addPersistentEntities(HalPerson, Parent)
+        def p = new HalPerson(name: 'Robert')
         p.homeAddress = new Address(postCode: '12345')
```
jdaugherty: The `Person` → `HalPerson` rename is right, but `Address` is still reaching into another spec by exactly the mechanism this PR is closing: it's declared in `EmbeddedAssociationsSpec` (line 190) and picked up here unqualified via same-package resolution.

It matters for the same reason `Person` did. `Address` is the embedded type of both `Person` and `HalPerson`, so `GormMappingConfigurationStrategy` calls `context.createEmbeddedEntity(Address)` — see `AbstractMappingContext#createEmbeddedEntity`, which builds a fresh `EmbeddedPersistentEntity(type, this)` bound to the calling context — once for this spec and once for `EmbeddedAssociationsSpec`. That's the identical `Class` object wrapped by two independently-built mapping contexts, which is the condition the rest of the PR eliminates.

Please give this spec its own `HalAddress` alongside `HalPerson`. It's a two-line change in a file you're already editing.

grails-views-gson/src/test/groovy/grails/plugin/json/view/HalEmbeddedSpec.groovy
```diff
@@ -330,14 +330,14 @@ class HalEmbeddedSpec extends Specification implements JsonViewTest {
 
     void 'test hal embedded method for one-to-many associations'() {
         when: 'A GSON view that renders hal.embedded(..) is rendered'
-        def player = new Player(id: 1L, name: 'Cantona')
+        def player = new HalPlayer(id: 1L, name: 'Cantona')
         player.id = 1L
-        def captain = new Player(name: 'Keane')
+        def captain = new HalPlayer(name: 'Keane')
         captain.id == 1L
```
jdaugherty: `==` rather than `=`, so this line does nothing — Spock only treats bare conditions as assertions in `then:`/`expect:`, and this is a `when:` block. Pre-existing, but you're changing the lines directly above and below it, so it's free to fix here.

Worth noting the expected JSON further down asserts `"href": "http://localhost:8080/halPlayer"` with no id, i.e. the captain genuinely has no id and the feature is passing for the right reason. So the fix is to delete this line rather than turn it into an assignment — unless you'd rather set the id and update the expected href to `/halPlayer/1`.

grails-views-gson/src/test/groovy/grails/plugin/json/view/NullRenderingSpec.groovy
```diff
@@ -92,12 +93,12 @@ class NullRenderingSpec extends Specification implements JsonViewTest {
...
             }
-            
+
             json g.render(obj)
         '''
 
         when:
-        mappingContext.addPersistentEntity(Player)
+        mappingContext.addPersistentEntity(NullRenderingPlayer)
         def renderResult = render(templateText, [obj: new Child2()])
```
jdaugherty: Same pattern as the `Team`/`Player` borrowing the rest of this PR fixes: `Child2` is declared in `PogoDeepRenderingSpec` and reached here unqualified.

Lower stakes than the entity cases — `Child2` is a plain POGO, so nothing registers it into a mapping context — but it's the same silent binding, in a file you're already changing. A `NullRenderingChild` local to this spec closes it.

grails-views-gson/src/test/groovy/grails/plugin/json/view/IterableRenderSpec.groovy
```diff
@@ -339,3 +340,25 @@ class IterableRenderSpec extends Specification implements JsonViewUnitTest {
         ex.message == 'Error rendering view: JSON API pagination arguments must contain resource and total'
     }
 }
+
+@Entity
+class IterableRenderTeam {
```
jdaugherty: Design question on the approach as a whole, anchored here because this file shows the cost most clearly.

Prefixing the class names forces every expected-JSON string in the spec to change, and that churn is most of the +363/−219. Per-spec sub-packages would buy the same isolation for almost none of it: the JSON API `type` comes from `PersistentEntity.decapitalizedName` (`DefaultJsonApiViewHelper:183`) and HAL hrefs from `GrailsNameUtils.getPropertyName(clazz)` (`TestLinkGenerator:72`) — both the *simple* name. So `grails.plugin.json.view.iterable.Player` renders byte-identically to today's `Player`, distinct `Class` object and all, and every assertion in the file stays untouched.

Two reasons I lean that way:

- Rewritten assertions lose their regression value. If name derivation itself regressed, the old strings would catch it; the new ones were written to match current output.
- It's self-enforcing. Nothing in this PR stops the next spec added to `grails.plugin.json.view` from typing `new Player(...)` and silently binding to `JsonViewHelperSpec` all over again. With per-spec packages that doesn't compile.

There's no template fallout to worry about: the module's only `.gson` fixtures live under `grails-app/views` and none of them are named for `player` or `team`.

This is a rework of a rename you've already done twice, so I'll leave the call to you. If you keep the prefixes, please add a line of comment above each duplicated fixture block saying why it's duplicated — otherwise someone will helpfully consolidate the seven copies back into one shared pair and reintroduce the problem.

grails-views-gson/src/test/groovy/grails/plugin/json/view/NullRenderingSpec.groovy
```diff
@@ -110,15 +111,37 @@ class NullRenderingSpec extends Specification implements JsonViewTest {
...
+class NullRenderingPlayer {
+    Long version
+    String name
+    @SuppressWarnings('unused')
+    static belongsTo = [team: NullRenderingTeam]
+
+    static constraints = {
+        name nullable: false
+    }
 }
```
jdaugherty: Nit: still missing the trailing newline at EOF, and this commit rewrites the tail of the file anyway.

grails-views-gson/src/test/groovy/grails/plugin/json/view/api/JsonApiHandleAssociationsSpec.groovy
```diff
@@ -72,3 +72,8 @@ class Publisher {
     String name
 }
 
+@Entity
+class HandleAssociationsAuthor {
+    String name
+}
+
```
jdaugherty: Nit: trailing blank line at EOF.

jdaugherty (commented):
Thanks for turning this round quickly — both asks from my last review are addressed. `git diff <merge-base>..a42b487 -- grails-views-gson/src/main` comes back empty, so the trait is byte-identical to base and "No production code changed" now holds, and the entity decoupling is extended to six more specs. I also appreciate how explicit the PR body is about what the evidence does and doesn't show.

Verified locally on `a42b487`: `DO_NOT_CACHE_TESTS=1 ./gradlew :grails-views-gson:test` → 178 tests, 0 failures. The branch is 314 commits behind `8.0.x`, but `8.0.x` has no changes under `grails-views-gson` since your merge base, so there are no conflicts to expect.

Two things I'd like resolved before this merges, plus one design question.

**1. The decoupling pass is still incomplete, including in files this PR edits.** Details inline on `HalEmbeddedSpec` (`Address`) and `NullRenderingSpec` (`Child2`). Beyond those two, `TemplateInheritanceSpec` still resolves `Player` and `Circular` out of `JsonViewHelperSpec` by the same unqualified same-package mechanism — so the commit message's claim that "JsonViewHelperSpec, EmbeddedAssociationsSpec and JsonApiSpec keep their original classes unchanged since those are no longer borrowed by anyone else" isn't accurate. There's a good reason those two are awkward to move: `grails-app/views/_child{2,3,4}*.gson` and `circular/_circular.gson` import `grails.plugin.json.view.Player` and `grails.plugin.json.view.Circular` directly, so renaming them means touching published-module templates. Please state that as the reason rather than claiming nothing borrows them. (`api/PaginationSpec` also imports `grails.plugin.json.view.Book` from `JsonViewTemplateEngineSpec`, but that one is an explicit import rather than a silent binding, so I'd leave it.)

**2. `captain.id == 1L`** in `HalEmbeddedSpec` — inline.

**3. Design question: per-spec sub-packages instead of name prefixes** — inline on `IterableRenderSpec`, where the churn is easiest to see.

Nits, none blocking:

- `HalEmbeddedSpec` imports `grails.gorm.annotation.Entity` while the other new fixture blocks use `grails.persistence.Entity`. Both work, and each file is internally consistent, so only worth aligning if it's cheap.
- Several of the copied fixtures carry fields the borrowing spec never touches (`IncludeAssociationsTeam.captain`/`titles`, `NullRenderingTeam` in its entirety). They're faithful copies of the originals, which is defensible; trimming is optional.
- Two EOF nits flagged inline.

On #16030: I'd land this on its own merits, but please don't close #16030 with it, and consider retitling the PR and branch to what's actually verifiable — something like "Isolate shared test entities in grails-views-gson specs". My own runs don't discriminate between hypotheses any better than yours do; 178/178 green tells us nothing about whether the leak is closed. The identical per-method failure counts within each spec are still the strongest lead, and I'd like the issue left open pointing at that rather than treated as resolved by association.

borinquenkid:
@jdaugherty nudge

grails-views-gson/src/test/groovy/grails/plugin/json/view/HalEmbeddedSpec.groovy
```diff
@@ -33,25 +33,25 @@ class HalEmbeddedSpec extends Specification implements JsonViewTest {
     JsonMapper objectMapper = JsonMapper.builder().build()
 
     void setup() {
-        mappingContext.addPersistentEntities(Team, Player)
+        mappingContext.addPersistentEntities(HalTeam, HalPlayer)
```
jdaugherty: This is the clearest place to raise it: I don't think the mapping-context rationale holds, and I checked rather than reasoned about it.

`JsonViewTest.mappingContext` is a plain trait property, not `@Shared`:

```groovy
// grails-views-gson/src/main/groovy/grails/plugin/json/view/test/JsonViewTest.groovy:67
@Autowired(required = false)
MappingContext mappingContext = {
    def ctx = new KeyValueMappingContext('test')
    ctx.setCanInitializeEntities(true)
    return ctx
}()
```

Groovy initialises trait properties from `$init$` in the implementing class's constructor, and Spock instantiates the spec once per feature method. So every feature gets its own `KeyValueMappingContext`, and this `setup()` re-registers the same two `Class` objects into a fresh one each time.

I confirmed it with a throwaway spec implementing `JsonViewTest` — three features, `setup()` calling `addPersistentEntities`, recording `System.identityHashCode` of both the context and the resulting `PersistentEntity`:

```
PROBE distinct mappingContexts=3 distinct PersistentEntities=3
```

Three features, three independently-built contexts, three distinct `PersistentEntity` instances for one `Class`.

So "no class is registered into two independently-built mapping contexts" isn't an invariant this codebase has, and it isn't one the renames can establish. Post-rename, `HalEmbeddedSpec` registers `HalTeam`/`HalPlayer` into nine independently-built contexts on its own (nine features, all through this `setup()`); `IterableRenderSpec`, `NullRenderingSpec` and `MapRenderSpec` do the same explicitly, per feature, in this very diff. Whether a *second spec* also registers the class isn't a categorical change — it's N versus 2N of something that already happens by design.

That moves the stated root cause from "unconfirmed", where the description currently leaves it, to "contradicted by the PR's own code". I'd rather the description drop the flakiness framing than leave a hypothesis in the permanent history that the diff itself disproves.

What still stands on its own merits is narrower: the template strings previously did `import grails.plugin.json.view.*` into a package holding 20+ specs' fixtures, and the specs referenced `Player`/`Team` unqualified. Both are genuinely fragile and worth closing. But that's a readability and robustness argument — which is exactly what the per-spec sub-package suggestion on `IterableRenderSpec` buys, at a fraction of the churn.

grails-views-gson/src/test/groovy/grails/plugin/json/view/IncludeAssociationsSpec.groovy
```diff
@@ -44,7 +45,29 @@ json jsonapi.render(players, [associations: false])
...
 
         then: "The result is an array"
-        renderResult.jsonText == '{"data":[{"type":"player","id":"1","attributes":{"name":"Cantona"}}],"links":{"self":"/foo"}}'
+        renderResult.jsonText == '{"data":[{"type":"includeAssociationsPlayer","id":"1","attributes":{"name":"Cantona"}}],"links":{"self":"/foo"}}'
 
     }
 }
+
+@Entity
+class IncludeAssociationsTeam {
```
jdaugherty: Two more borrows the sweep hasn't reached, beyond the `Address` and `Child2` cases I flagged last time:

- `TemplateInheritanceSpec` constructs `new Player(...)` in six places and `new Circular(...)` in `"test circular rendering is handled"`, both binding unqualified to `JsonViewHelperSpec`'s `@Entity` classes (`JsonViewHelperSpec:672` and `:690`). Those are the same `Team`/`Player` fixtures the rest of this PR is decoupling, so it's the last remaining `@Entity` case in the package.
- `PaginationSpec` (`grails.plugin.json.view.api`) imports `grails.plugin.json.view.Book`, which is `JsonViewTemplateEngineSpec`'s `@Linkable` POGO — while `JsonApiSpec` declares its own `Book` in `PaginationSpec`'s own package. Only the explicit import disambiguates the two; remove it and the reference silently flips to a different class. Lower stakes than the entity cases, but the same shape.

I'm not asking for six more renames on top of these — the opposite. Enumerating them is the argument for the sub-package approach, which would cover every case in the package mechanically and without touching a single expected-JSON string.

jdaugherty (commented):
One new finding from this pass, and it goes to the premise rather than the diff: I went to verify the mapping-context isolation argument and it doesn't survive contact with `JsonViewTest`. `mappingContext` is a per-instance trait property, so every Spock feature already builds its own `KeyValueMappingContext` and re-registers the same `Class` objects into it. Details and the probe output are on `HalEmbeddedSpec`.

That doesn't make the diff wrong — replacing `import grails.plugin.json.view.*` inside the template strings with an explicit aliased import is a real improvement, and unqualified same-package fixture references are a real trap. It does mean the value here is naming and robustness, not flakiness, which I think should be reflected in the description before this lands.

My earlier comments are unchanged and still open.

jdaugherty:
@borinquenkid

borinquenkid:
Thanks for pushing on this — the mapping-context finding is a good catch, and I agree it changes what this PR should claim.

**On the premise.** You're right that `mappingContext` being a plain (non-`@Shared`) trait property means every feature already gets its own `KeyValueMappingContext`, with or without this PR. I've updated the description to stop framing this as a flakiness fix and call it what it actually is: closing a real same-package-borrowing footgun, independent of whether it explains #16030. The whole-spec-fixture-failure theory is still the leading explanation for that issue and remains open.

**On the design question — done, in `d7ee26c4b2`.** Switched from prefixed class names to per-spec sub-packages, each keeping the original simple names:
- `grails.plugin.json.view.expand`, `.include`, `.halembedded`, `.iterable`, `.maprender`, `.nullrendering` for `Team`/`Player`/`PlayerWithAge`/`Person`/`Address`
- `grails.plugin.json.view.api.handleassociations` for `Author`

Since `type`/`href` values come from the entity's simple class name, none of the expected-JSON assertions had to change — the diff on the 8 touched spec files dropped from +363/−219 to +100/−99, plus 7 small new fixture files. And it's self-enforcing the way you described: a same-package unqualified reference can't resolve into another spec's fixture anymore, because they're in different packages.

**Both remaining gaps from your last pass are closed:**
- `HalEmbeddedSpec`'s `Address` now has its own copy alongside `Person`.
- `NullRenderingSpec` gets its own `Child` POGO instead of unqualified-borrowing `PogoDeepRenderingSpec`'s `Child2`.

**`TemplateInheritanceSpec` and `PaginationSpec` are intentionally left alone**, per your reasoning: the published `child2`/`child4`/`circular` `.gson` templates import `grails.plugin.json.view.Player`/`.Circular` directly, so `TemplateInheritanceSpec`'s model has to be that exact class — I added a comment on the class explaining that so it doesn't read as an oversight next time. `PaginationSpec`'s explicit `import grails.plugin.json.view.Book` is disambiguating, not silently binding, so per your comment I left it as-is.

Also fixed the `captain.id == 1L` no-op you flagged — deleted rather than turned into an assignment, since the expected HAL link asserts no id.

Verified: full `:grails-views-gson:test` (178 tests) 0 failures, `codeStyle` clean.

PR #16034, opened 2026-07-21
Fix flaky WhereQueryClosureCaptureSpec: key AstPropertyResolveUtils cache by ClassNode identity

borinquenkid:
## Summary
`WhereQueryClosureCaptureSpec` has two flaky feature methods (~2% flaky per
apache/grails-core#16030, 0 hard failures) verifying that closures generated by the
`where {}` query transform only capture the variables they actually reference.

## Root cause
`AstPropertyResolveUtils.cachedClassProperties` is a static, non-thread-safe `HashMap`
used by the `where`-transform's variable-scope recompute path
(`DetachedCriteriaTransformer`). It was already keyed by `ClassNode.getName()`
(qualified name), not a bare simple name as first suspected — but `ClassNode.equals()`/
`hashCode()` also compare by name, and this spec's own test deliberately compiles
identical source **twice** into two separate `GroovyClassLoader`s, producing two
distinct `ClassNode` instances that are `equals()`-identical and therefore collide as
the same cache entry. This is reachable by any two classes sharing a name across
separate compilations, not just this test — a prior partial fix (08bd0bfbeb) only
renamed this spec's fixtures to reduce collisions with sibling specs, without fixing
the cache itself.

## Fix
- Switched `cachedClassProperties` to a synchronized `IdentityHashMap<ClassNode, ...>`
  — keys compared by `==`, so two distinct `ClassNode`s can never collide regardless of
  shared naming.
- Fixed a related concurrency bug in the same method: a new cache entry was previously
  published into the shared map *before* its population loop finished, letting a
  concurrent reader observe a partially-populated entry. Population now completes into
  a local map first, then publishes atomically.

## Testing
- New `AstPropertyResolveUtilsSpec.groovy` (public-API only): includes a test that
  constructs two distinct `ClassNode`s sharing the exact same name and asserts each
  resolves independently — a direct, deterministic reproduction of the collision
  mechanism, not a rerun-and-hope test.
- `:grails-datamapping-core:test` (full module): BUILD SUCCESSFUL, no failures,
  including both previously-flaky feature methods.
- CodeNarc/Checkstyle: clean.

### DIFF
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
@@ -49,7 +49,8 @@ import org.grails.datastore.mapping.reflect.NameUtils;
 public class AstPropertyResolveUtils {
 
     /**
-     * Key under which the resolved property map is stashed via {@link ClassNode#getNodeMetaData(Object, java.util.function.Function)}.
+     * Key under which the resolved property map is stashed as {@code ClassNode} metadata via
+     * {@link #getPropertiesFromCache}.
      * <p>
      * Earlier versions of this class cached resolved properties in a single static, process-wide
      * {@code Map} keyed by class name (later by {@code ClassNode} identity). Both designs share a
@@ -65,12 +66,21 @@ public class AstPropertyResolveUtils {
      *     <li>No collision is possible between distinct {@code ClassNode} instances that happen to
      *     share a name (e.g. classes compiled without a package, or the same source compiled twice
      *     in separate {@code GroovyClassLoader}s) - each instance owns its own metadata storage, so
-     *     there is no shared key space to collide on in the first place.</li>
+     *     there is no shared key space to collide on in the first place. This holds for the resolved
+     *     <em>property map</em> built here.</li>
      *     <li>No leak is possible - the cached data is only reachable through the {@code ClassNode}
      *     it describes, so it becomes eligible for garbage collection at the same time as the node
      *     (and the compilation/classloader it belongs to) rather than being pinned forever by a
      *     static field.</li>
      * </ul>
+     * This does <strong>not</strong> extend to whether a given {@code ClassNode} is treated as a
+     * domain class in the first place: {@link AstUtils#isDomainClass(ClassNode)} is
+     * {@code @Memoized}, and Groovy's default memoize is keyed by argument <em>equality</em>
+     * ({@code ClassNode#equals}/{@code hashCode} both reduce to the class name), not identity. Two
+     * distinct, same-named {@code ClassNode}s can therefore still receive the same domain-class
+     * verdict from each other via that memo cache, independent of the property cache here. That is
+     * a narrower, pre-existing exposure in {@code isDomainClass} itself - this class neither causes
+     * nor fixes it, and it is tracked separately rather than folded into this cache's guarantees.
      * Per {@link ClassNode#getModule()}'s own convention, the cache is stored on
      * {@link ClassNode#redirect()} - the node a placeholder/generics-parameterized reference
      * ultimately stands in for - so that looking a class up through different reference nodes still
@@ -116,6 +126,24 @@ public class AstPropertyResolveUtils {
      * metadata key to the same shared node concurrently without also synchronizing on that node -
      * this class has no way to compel that. That residual risk belongs to {@code ClassNode}'s
      * metadata storage in general, not to anything specific to the cache here.
+     * <p>
+     * Those interned singleton nodes ({@link ClassNode#isPrimaryClassNode()} returns {@code false}
+     * for them, since {@link ClassHelper#make(Class)} builds them via the {@code ClassNode(Class)}
+     * constructor) are never written to by this class at all: {@link #getPropertiesFromCache}
+     * recomputes for them on every call instead of caching. Caching there would mean writing this
+     * class's own metadata key into a node no single compilation owns, growing the residual risk
+     * above from "someone else's key might race with a read" to "this class's own writes pollute a
+     * JVM-wide-shared node forever" - not a real cost, since the properties of a JDK type are cheap
+     * to recompute and there are only a handful of these nodes in practice.
+     * <p>
+     * {@link #computeProperties} can run arbitrary user code for a resolved domain class - it
+     * forces the class's static initializer and invokes user-written static getters via
+     * {@link ClassPropertyFetcher}. Running that while holding the per-node monitor would risk
+     * deadlocking against unrelated code the caller has no visibility into, so
+     * {@link #getPropertiesFromCache} computes outside the lock and only takes the monitor to check
+     * for and, if needed, publish the result. Two threads can therefore both compute for the same
+     * as-yet-uncached node; the second to acquire the monitor discards its own (equivalent) result
+     * and returns whatever the first published, so only one map is ever visible to callers.
      */
     private static final String PROPERTIES_CACHE_KEY = AstPropertyResolveUtils.class.getName() + ".properties";
 
@@ -165,8 +193,23 @@ public class AstPropertyResolveUtils {
 
     private static Map<String, ClassNode> getPropertiesFromCache(ClassNode classNode) {
         ClassNode cacheHolder = classNode.redirect();
+        if (!cacheHolder.isPrimaryClassNode()) {
+            return computeProperties(cacheHolder);
+        }
         synchronized (cacheHolder) {
-            return cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY, key -> computeProperties(cacheHolder));
+            Map<String, ClassNode> cached = cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY);
+            if (cached != null) {
+                return cached;
+            }
+        }
+        Map<String, ClassNode> computed = computeProperties(cacheHolder);
+        synchronized (cacheHolder) {
+            Map<String, ClassNode> cached = cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY);
+            if (cached != null) {
+                return cached;
+            }
+            cacheHolder.putNodeMetaData(PROPERTIES_CACHE_KEY, computed);
+            return computed;
         }
     }
 
@@ -174,8 +217,8 @@ public class AstPropertyResolveUtils {
         Map<String, ClassNode> newProperties = new HashMap<>();
         boolean isDomainClass = AstUtils.isDomainClass(classNode);
         if (isDomainClass) {
-            newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
-            newProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
+            newProperties.put(GormProperties.IDENTITY, ClassHelper.make(Long.class).getPlainNodeReference());
+            newProperties.put(GormProperties.VERSION, ClassHelper.make(Long.class).getPlainNodeReference());
         }
         ClassNode currentNode = classNode;
         while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
@@ -20,7 +20,9 @@
 package org.grails.datastore.gorm.transform;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -47,7 +49,28 @@ import org.grails.datastore.mapping.reflect.NameUtils;
  * @since 6.1
  */
 public class AstPropertyResolveUtils {
-    protected static Map<String, Map<String, ClassNode>> cachedClassProperties = new HashMap<>();
+
+    /**
+     * Cache of resolved properties per {@link ClassNode}.
+     * <p>
+     * Keyed by {@code ClassNode} identity rather than name. {@link ClassNode#equals(Object)} and
+     * {@link ClassNode#hashCode()} compare by {@link ClassNode#getText()} (essentially the class
+     * name), so a {@code Map} keyed by name - or even by {@code ClassNode} itself as the map key -
+     * treats any two distinct {@code ClassNode} instances that happen to share a name as the same
+     * cache entry. That collision is a real hazard for classes compiled without a package (common
+     * in tests and dynamically generated sources), and for the same source compiled more than once
+     * in separate {@code GroovyClassLoader}s: each compilation produces its own {@code ClassNode}
+     * instance that must never share cached property data with another compilation's instance of a
+     * same-named class. An {@link IdentityHashMap} avoids that collision entirely by comparing keys
+     * with {@code ==} instead of {@code equals()}.
+     * <p>
+     * Wrapped in {@link Collections#synchronizedMap(Map)} because AST transforms that populate and
+     * read this cache can run concurrently on multiple threads (e.g. parallel test execution within
+     * one JVM/fork); a plain, unsynchronized {@link HashMap} is not safe for concurrent structural
+     * modification and can corrupt its internal state under concurrent {@code put()} calls.
+     */
+    protected static final Map<ClassNode, Map<String, ClassNode>> cachedClassProperties =
+            Collections.synchronizedMap(new IdentityHashMap<>());
 
     /**
      * Resolves the type of of the given property
@@ -94,22 +117,25 @@ public class AstPropertyResolveUtils {
     }
 
     private static Map<String, ClassNode> getPropertiesFromCache(ClassNode classNode) {
-        String className = classNode.getName();
-        Map<String, ClassNode> cachedProperties = cachedClassProperties.get(className);
+        Map<String, ClassNode> cachedProperties = cachedClassProperties.get(classNode);
         if (cachedProperties == null) {
-            cachedProperties = new HashMap<>();
+            Map<String, ClassNode> newProperties = new HashMap<>();
             boolean isDomainClass = AstUtils.isDomainClass(classNode);
             if (isDomainClass) {
-                cachedProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
-                cachedProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
+                newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
+                newProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
             }
-            cachedClassProperties.put(className, cachedProperties);
             ClassNode currentNode = classNode;
             while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
-                populatePropertiesForClassNode(currentNode, cachedProperties, isDomainClass, !isDomainClass);
+                populatePropertiesForClassNode(currentNode, newProperties, isDomainClass, !isDomainClass);
                 currentNode = currentNode.getSuperClass();
             }
-        } return cachedProperties;
+            // Publish only once fully populated so a concurrent reader can never observe a
+            // partially-populated entry for this ClassNode.
+            cachedProperties = newProperties;
+            cachedClassProperties.put(classNode, cachedProperties);
+        }
+        return cachedProperties;
     }
 
     private static void populatePropertiesForClassNode(ClassNode classNode, Map<String, ClassNode> cachedProperties, boolean isDomainClass, boolean allowAbstract) {
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
@@ -20,9 +20,7 @@
 package org.grails.datastore.gorm.transform;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -51,26 +49,43 @@ import org.grails.datastore.mapping.reflect.NameUtils;
 public class AstPropertyResolveUtils {
 
     /**
-     * Cache of resolved properties per {@link ClassNode}.
+     * Key under which the resolved property map is stashed via {@link ClassNode#getNodeMetaData(Object, java.util.function.Function)}.
      * <p>
-     * Keyed by {@code ClassNode} identity rather than name. {@link ClassNode#equals(Object)} and
-     * {@link ClassNode#hashCode()} compare by {@link ClassNode#getText()} (essentially the class
-     * name), so a {@code Map} keyed by name - or even by {@code ClassNode} itself as the map key -
-     * treats any two distinct {@code ClassNode} instances that happen to share a name as the same
-     * cache entry. That collision is a real hazard for classes compiled without a package (common
-     * in tests and dynamically generated sources), and for the same source compiled more than once
-     * in separate {@code GroovyClassLoader}s: each compilation produces its own {@code ClassNode}
-     * instance that must never share cached property data with another compilation's instance of a
-     * same-named class. An {@link IdentityHashMap} avoids that collision entirely by comparing keys
-     * with {@code ==} instead of {@code equals()}.
+     * Earlier versions of this class cached resolved properties in a single static, process-wide
+     * {@code Map} keyed by class name (later by {@code ClassNode} identity). Both designs share a
+     * problem: a static map is never emptied, so every {@code ClassNode} ever looked up - and, for
+     * a primary node, the {@code GroovyClassLoader}/{@code CompileUnit} it pins via
+     * {@link ClassNode#getModule()} - is retained for the lifetime of the JVM. In a long-lived
+     * process that repeatedly compiles Groovy (a Gradle daemon reusing its Groovy compiler across
+     * builds, a dev-mode recompile loop), that is an unbounded classloader leak.
      * <p>
-     * Wrapped in {@link Collections#synchronizedMap(Map)} because AST transforms that populate and
-     * read this cache can run concurrently on multiple threads (e.g. parallel test execution within
-     * one JVM/fork); a plain, unsynchronized {@link HashMap} is not safe for concurrent structural
-     * modification and can corrupt its internal state under concurrent {@code put()} calls.
+     * Storing the resolved properties as metadata on the {@code ClassNode} itself instead avoids
+     * both hazards this class has previously had to fix:
+     * <ul>
+     *     <li>No collision is possible between distinct {@code ClassNode} instances that happen to
+     *     share a name (e.g. classes compiled without a package, or the same source compiled twice
+     *     in separate {@code GroovyClassLoader}s) - each instance owns its own metadata storage, so
+     *     there is no shared key space to collide on in the first place.</li>
+     *     <li>No leak is possible - the cached data is only reachable through the {@code ClassNode}
+     *     it describes, so it becomes eligible for garbage collection at the same time as the node
+     *     (and the compilation/classloader it belongs to) rather than being pinned forever by a
+     *     static field.</li>
+     * </ul>
+     * Per {@link ClassNode#getModule()}'s own convention, the cache is stored on
+     * {@link ClassNode#redirect()} - the node a placeholder/generics-parameterized reference
+     * ultimately stands in for - so that looking a class up through different reference nodes still
+     * shares one cache entry.
+     * <p>
+     * No explicit synchronization is used. A given {@code ClassNode} instance is built by, and only
+     * ever mutated by, the single compiler thread compiling its source unit; nothing in this
+     * codebase compiles the same {@code ClassNode} from two threads at once (Gradle's test workers
+     * are separate JVMs, not threads sharing this cache, and this module does not enable JUnit's
+     * in-JVM parallel execution). The realistic concurrency exposure - if it ever arises - is
+     * distinct {@code ClassNode}s being resolved concurrently by independent {@code compileGroovy}
+     * tasks, each populating its own node's metadata; that is naturally race-free since the nodes
+     * involved are different objects.
      */
-    protected static final Map<ClassNode, Map<String, ClassNode>> cachedClassProperties =
-            Collections.synchronizedMap(new IdentityHashMap<>());
+    private static final String PROPERTIES_CACHE_KEY = AstPropertyResolveUtils.class.getName() + ".properties";
 
     /**
      * Resolves the type of of the given property
@@ -117,25 +132,23 @@ public class AstPropertyResolveUtils {
     }
 
     private static Map<String, ClassNode> getPropertiesFromCache(ClassNode classNode) {
-        Map<String, ClassNode> cachedProperties = cachedClassProperties.get(classNode);
-        if (cachedProperties == null) {
-            Map<String, ClassNode> newProperties = new HashMap<>();
-            boolean isDomainClass = AstUtils.isDomainClass(classNode);
-            if (isDomainClass) {
-                newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
-                newProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
-            }
-            ClassNode currentNode = classNode;
-            while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
-                populatePropertiesForClassNode(currentNode, newProperties, isDomainClass, !isDomainClass);
-                currentNode = currentNode.getSuperClass();
-            }
-            // Publish only once fully populated so a concurrent reader can never observe a
-            // partially-populated entry for this ClassNode.
-            cachedProperties = newProperties;
-            cachedClassProperties.put(classNode, cachedProperties);
+        ClassNode cacheHolder = classNode.redirect();
+        return cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY, key -> computeProperties(cacheHolder));
+    }
+
+    private static Map<String, ClassNode> computeProperties(ClassNode classNode) {
+        Map<String, ClassNode> newProperties = new HashMap<>();
+        boolean isDomainClass = AstUtils.isDomainClass(classNode);
+        if (isDomainClass) {
+            newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
+            newProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
         }
-        return cachedProperties;
+        ClassNode currentNode = classNode;
+        while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
+            populatePropertiesForClassNode(currentNode, newProperties, isDomainClass, !isDomainClass);
+            currentNode = currentNode.getSuperClass();
+        }
+        return newProperties;
     }
 
     private static void populatePropertiesForClassNode(ClassNode classNode, Map<String, ClassNode> cachedProperties, boolean isDomainClass, boolean allowAbstract) {
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
@@ -74,16 +74,48 @@ public class AstPropertyResolveUtils {
      * Per {@link ClassNode#getModule()}'s own convention, the cache is stored on
      * {@link ClassNode#redirect()} - the node a placeholder/generics-parameterized reference
      * ultimately stands in for - so that looking a class up through different reference nodes still
-     * shares one cache entry.
+     * shares one cache entry. {@code redirect()} can in principle be reassigned after construction
+     * (e.g. a forward-reference placeholder later pointed at the real, resolved node by the
+     * compiler's resolve phase) - but {@link ClassNode#setRedirect(ClassNode)} explicitly refuses to
+     * do so for a <em>primary</em> node ({@code GroovyBugError}), and every {@code ClassNode} this
+     * utility's real callers pass in - one actively being compiled from source during an AST
+     * transform - is primary, so for them {@code redirect()} is simply {@code this} for the entire
+     * object's life; there is no reassignment to account for. The one place a non-primary,
+     * redirect-able node does appear ({@link ClassHelper#make(Class, boolean) ClassHelper.make(c,
+     * false)}) sets its redirect target eagerly, at construction, before any caller could observe or
+     * cache through it in an unredirected state. Even so, if some future, currently-unforeseen path
+     * ever did cache through a node before it was redirected, that would be harmless rather than a
+     * source of stale data: the entry becomes unreachable the moment the redirect is set (a later
+     * lookup through the same original reference resolves to the different, real node and computes
+     * fresh there instead), so it is simply never read again and becomes eligible for garbage
+     * collection with the abandoned node.
      * <p>
-     * No explicit synchronization is used. A given {@code ClassNode} instance is built by, and only
-     * ever mutated by, the single compiler thread compiling its source unit; nothing in this
-     * codebase compiles the same {@code ClassNode} from two threads at once (Gradle's test workers
-     * are separate JVMs, not threads sharing this cache, and this module does not enable JUnit's
-     * in-JVM parallel execution). The realistic concurrency exposure - if it ever arises - is
-     * distinct {@code ClassNode}s being resolved concurrently by independent {@code compileGroovy}
-     * tasks, each populating its own node's metadata; that is naturally race-free since the nodes
-     * involved are different objects.
+     * {@code classNode.isResolved()} gates one part of what gets cached (see
+     * {@link #populatePropertiesForClassNode}) and, in principle, its result can depend on when it
+     * is checked - it delegates through {@code redirect} (see {@link ClassNode#isResolved()}). Given
+     * the above, that delegation is immaterial for the primary nodes this cache actually serves: a
+     * terminal node's {@code isResolved()} reduces to {@code clazz != null} (or, for array/component
+     * types, {@code componentType.isResolved()}) - and {@code clazz} has no setter anywhere in
+     * {@code ClassNode} outside its {@code ClassNode(Class)} constructor (verified against the
+     * Groovy 5.0.7 sources), so that value is fixed for the object's entire life.
+     * <p>
+     * Access is synchronized per {@code ClassNode} ({@code synchronized (cacheHolder)} in
+     * {@link #getPropertiesFromCache}) because {@link ClassNode}'s node-metadata storage
+     * ({@code NodeMetaDataHandler}, backed by {@code ListHashMap}) is explicitly documented as not
+     * thread-safe, and while a given {@code ClassNode} representing a class actively being compiled
+     * is normally touched by only the one thread compiling it, that is not true of every node this
+     * utility can be called with: some callers resolve a property's declared type as a plain
+     * {@code ClassNode} for a common JDK type (e.g. an {@code Object}- or {@code def}-typed
+     * property), and {@link ClassHelper#make(Class)} returns a small, fixed set of interned,
+     * JVM-wide-shared singleton nodes for exactly those types (e.g. {@link ClassHelper#OBJECT_TYPE},
+     * {@code STRING_TYPE}). Two unrelated, concurrently-running compilations that both happen to
+     * resolve such a type would otherwise race on the same node's metadata map with no protection at
+     * all. Synchronizing per node makes concurrent calls into this class safe with respect to each
+     * other and keeps the common case (one thread, one node) effectively uncontended. It cannot, by
+     * itself, protect against unrelated code elsewhere in the compiler writing a <em>different</em>
+     * metadata key to the same shared node concurrently without also synchronizing on that node -
+     * this class has no way to compel that. That residual risk belongs to {@code ClassNode}'s
+     * metadata storage in general, not to anything specific to the cache here.
      */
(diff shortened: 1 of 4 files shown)


grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,101 @@
...
+ */
+class AstPropertyResolveUtilsSpec extends Specification {
+
+    void "property lookups for two same-named ClassNodes in different packages do not corrupt each other"() {
+        given: 'two distinct ClassNodes with the same simple name declared in different packages'
+        ClassNode first = new ClassNode('org.example.one.Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+        first.addProperty('color', Modifier.PUBLIC, ClassHelper.STRING_TYPE, null, null, null)
+
+        ClassNode second = new ClassNode('org.example.two.Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+        second.addProperty('weight', Modifier.PUBLIC, ClassHelper.Integer_TYPE, null, null, null)
```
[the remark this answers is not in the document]
borinquenkid: Good catch on the mechanics — you're right that this test's two ClassNodes use different fully-qualified names (`org.example.one.Widget` vs `org.example.two.Widget`), so it wouldn't have collided under the old name-/equals()-keyed cache either, and doesn't by itself guard the regression.

That guard is the next test below, `"property lookups for two distinct ClassNode instances with the exact same unqualified name do not corrupt each other"` — both ClassNodes there are named plain `Widget` with no package, so they *do* compare equal (`first == second`, matching `hashCode()`) exactly as the old cache's key would, and the test asserts they still resolve independently under the new identity-keyed cache. That's the one that fails under the old implementation and passes under this fix.

This first test is intentionally a different, narrower check — that two distinct, differently-named classes never get confused with each other, which is a correctness property worth keeping on its own regardless of the collision bug. Leaving both as-is: this one for general non-contamination across genuinely different classes, the next one for the actual same-name collision regression.

borinquenkid:
Thanks for the review! The scenario you're describing — two distinct `ClassNode` instances sharing the exact same fully-qualified name — is already covered by the test right below this one:

```groovy
void "property lookups for two distinct ClassNode instances with the exact same unqualified name do not corrupt each other"() {
    given: 'two distinct ClassNode instances - as produced by two separate compilations - sharing an identical unqualified name'
    ClassNode first = new ClassNode('Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
    first.addProperty('color', Modifier.PUBLIC, ClassHelper.STRING_TYPE, null, null, null)

    ClassNode second = new ClassNode('Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
    second.addProperty('weight', Modifier.PUBLIC, ClassHelper.Integer_TYPE, null, null, null)

    expect: 'the two ClassNode instances compare equal by name - the exact condition that would collide in a name-keyed or equals()-keyed cache'
    first == second
    first.hashCode() == second.hashCode()
    !first.is(second)
    ...
```

It goes a step further than the suggested edit by asserting `first == second` and matching `hashCode()`, which explicitly proves the collision condition a name-/equals()-keyed cache would hit — exactly the regression this fix guards against.

The test this comment is attached to intentionally covers a different case: two `ClassNode`s that share a simple name but differ by package (so they're *not* equal), which is a distinct scenario worth keeping separate. Given the exact-FQN-collision case is already exercised, I'll leave both tests as-is rather than introduce a near-duplicate.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -47,7 +49,28 @@
...
+     * same-named class. An {@link IdentityHashMap} avoids that collision entirely by comparing keys
+     * with {@code ==} instead of {@code equals()}.
+     * <p>
+     * Wrapped in {@link Collections#synchronizedMap(Map)} because AST transforms that populate and
+     * read this cache can run concurrently on multiple threads (e.g. parallel test execution within
+     * one JVM/fork); a plain, unsynchronized {@link HashMap} is not safe for concurrent structural
+     * modification and can corrupt its internal state under concurrent {@code put()} calls.
+     */
+    protected static final Map<ClassNode, Map<String, ClassNode>> cachedClassProperties =
+            Collections.synchronizedMap(new IdentityHashMap<>());
```
jdaugherty: Keying this **static** map by `ClassNode` identity does fix the collision, but it converts a bounded cache into a classloader leak.

Under the old `String` key the map held at most one entry per distinct class name. With `ClassNode` keys it holds one entry per *instance*, and a primary `ClassNode` transitively pins its entire compilation:

```
ClassNode.getModule()  ->  ModuleNode.getUnit()  ->  CompileUnit.loader  (GroovyClassLoader)
```

plus `ClassNode.clazz` directly for resolved nodes. Since nothing ever removes an entry and the field is now `static final`, every compilation that runs a GORM AST transform inside a long-lived JVM permanently retains that compilation's classloader and every class it loaded. That is not hypothetical: the Gradle Groovy compiler daemon is reused across `compileGroovy` tasks and across builds, and dev-mode / `GroovyClassLoader`-driven recompiles land in the same place. The map values are `ClassNode`s too, so they pin loaders as well — that part pre-existed, but it was O(distinct class names) and is now unbounded in the number of compilations.

Groovy already stores per-node state exactly this way (`ClassNode.getModule()` above is itself `getNodeMetaData(ModuleNode.class)`), so the cleanest fix is to drop the static map and hang the cache off the node:

```java
private static final String PROPERTIES_CACHE_KEY = AstPropertyResolveUtils.class.getName() + ".properties";

private static Map<String, ClassNode> getPropertiesFromCache(ClassNode classNode) {
    return classNode.getNodeMetaData(PROPERTIES_CACHE_KEY, cn -> computeProperties(classNode));
}
```

`getNodeMetaData(Object, Function)` is available in Groovy 5, is identity-scoped by construction (so the collision this PR fixes cannot arise at all), needs no global lock, and is collected together with the node. Worth deciding explicitly whether the holder should be `classNode` or `classNode.redirect()` — `getModule()` uses `redirect()`, and redirected nodes are the one case where the two differ.

If a process-wide map has to stay for some reason, it needs weak identity keys and/or an explicit eviction point, plus a note documenting the expected lifecycle.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -47,7 +49,28 @@
...
+     * instance that must never share cached property data with another compilation's instance of a
+     * same-named class. An {@link IdentityHashMap} avoids that collision entirely by comparing keys
+     * with {@code ==} instead of {@code equals()}.
+     * <p>
+     * Wrapped in {@link Collections#synchronizedMap(Map)} because AST transforms that populate and
+     * read this cache can run concurrently on multiple threads (e.g. parallel test execution within
+     * one JVM/fork); a plain, unsynchronized {@link HashMap} is not safe for concurrent structural
+     * modification and can corrupt its internal state under concurrent {@code put()} calls.
+     */
+    protected static final Map<ClassNode, Map<String, ClassNode>> cachedClassProperties =
```
jdaugherty: Two breaking changes to a `protected` member of a public class on one line: the key type changed (`String` -> `ClassNode`) and the field became `final`. Anything downstream that reassigned it — today the *only* way to clear this cache, which plugin AST transforms and test harnesses plausibly do — now fails to compile rather than merely behaving differently.

If it's being broken anyway, take it the rest of the way: make it `private static final` and expose an explicit, documented `clearCache()`. That shrinks the exposed surface and gives the retention problem above an escape hatch, instead of leaving `protected` visibility on a field nobody can usefully touch any more.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -47,7 +49,28 @@
...
+     * in tests and dynamically generated sources), and for the same source compiled more than once
+     * in separate {@code GroovyClassLoader}s: each compilation produces its own {@code ClassNode}
+     * instance that must never share cached property data with another compilation's instance of a
+     * same-named class. An {@link IdentityHashMap} avoids that collision entirely by comparing keys
+     * with {@code ==} instead of {@code equals()}.
+     * <p>
+     * Wrapped in {@link Collections#synchronizedMap(Map)} because AST transforms that populate and
+     * read this cache can run concurrently on multiple threads (e.g. parallel test execution within
+     * one JVM/fork); a plain, unsynchronized {@link HashMap} is not safe for concurrent structural
+     * modification and can corrupt its internal state under concurrent {@code put()} calls.
```
jdaugherty: This rationale is inaccurate in a way that will mislead the next reader. `maxParallelForks = configuredTestParallel` (`gradle/test-config.gradle:86`) forks **separate JVMs**, and each JVM gets its own copy of a `static` field — parallel test forks can therefore never race on this map. JUnit's in-JVM parallel execution isn't enabled anywhere in the build either.

The concurrency exposure that does exist is the compiler itself: multiple `compileGroovy` tasks running concurrently in a shared Gradle worker, and any embedded compilation driven from more than one thread. Worth rewording to that, otherwise the comment justifies the synchronization with a scenario that can't happen.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -94,22 +117,25 @@ public static List<String> getPropertyNames(ClassNode classNode) {
...
             while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
-                populatePropertiesForClassNode(currentNode, cachedProperties, isDomainClass, !isDomainClass);
+                populatePropertiesForClassNode(currentNode, newProperties, isDomainClass, !isDomainClass);
                 currentNode = currentNode.getSuperClass();
             }
-        } return cachedProperties;
+            // Publish only once fully populated so a concurrent reader can never observe a
+            // partially-populated entry for this ClassNode.
+            cachedProperties = newProperties;
+            cachedClassProperties.put(classNode, cachedProperties);
```
jdaugherty: The reorder-and-publish-last fix is correct, and `synchronizedMap` does supply the safe publication it depends on: `get` and `put` synchronize on the same mutex, so a reader that observes the entry also observes a fully-populated `HashMap`. Worth stating that explicitly in the inline comment, because "publish only once fully populated" is only sufficient *given* that happens-before edge — with a bare `HashMap` the reordering alone wouldn't have been enough.

One remaining wrinkle: the check-then-act across the `get` on line 120 and the `put` on line 136 is not atomic, so two concurrent callers can each build a complete map and the later `put` wins. Benign here (the maps are equivalent and never mutated after publication), but `cachedClassProperties.computeIfAbsent(classNode, ...)` on the synchronized wrapper would be atomic and shorter — the tradeoff being that it holds the mutex for the whole superclass walk.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -94,22 +117,25 @@ public static List<String> getPropertyNames(ClassNode classNode) {
...
+                newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
+                newProperties.put(GormProperties.VERSION, new ClassNode(Long.class));
             }
-            cachedClassProperties.put(className, cachedProperties);
             ClassNode currentNode = classNode;
             while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
-                populatePropertiesForClassNode(currentNode, cachedProperties, isDomainClass, !isDomainClass);
+                populatePropertiesForClassNode(currentNode, newProperties, isDomainClass, !isDomainClass);
                 currentNode = currentNode.getSuperClass();
             }
```
jdaugherty: Identity keying makes the cache immune to name collisions, but I don't think it removes the nondeterminism that a flake like #16030 needs. `populatePropertiesForClassNode` consults `ClassPropertyFetcher` only when `classNode.isResolved()` (line 172), so an entry is a snapshot of whatever resolution state the node happened to be in at the *first* lookup, and it is never refreshed afterwards. If the first lookup for a node lands at a different compilation phase between runs, the cached property set still differs between runs — identity keys or not.

That also weakens the root-cause story in the description: compiling the *same* source twice produces two `ClassNode`s with identical property sets, so a name-keyed collision between just those two is harmless and can't produce divergent bytecode on its own. The collisions that would actually diverge are with a *differently-shaped* same-named class, or with a same-named node whose entry was cached at a different resolution state.

Can you pin down which one you observed — e.g. the two colliding class names, or the flake reproducing in a loop pre-fix and not post-fix? The change is an improvement regardless, but if the resolution-state snapshot is the real driver then #16030 comes back and this gets recorded as already fixed.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,101 @@
...
+ * map keyed by {@link ClassNode}. Two distinct compilations (e.g. the same source parsed in two
+ * different {@code GroovyClassLoader}s, as happens for dynamically-generated sources and in
+ * tests) produce distinct {@code ClassNode} instances that can legitimately share the exact same
+ * name - {@code ClassNode#equals(Object)} compares by name, so a naive name- or equals()-based
+ * cache key would conflate them, corrupting the resolved properties of one class with those of
+ * an unrelated class that happens to share its name. This spec proves the cache keys strictly by
+ * {@code ClassNode} identity, so same-named-but-distinct class nodes never contaminate each
+ * other's cached property data.
+ */
+class AstPropertyResolveUtilsSpec extends Specification {
```
jdaugherty, line 54: Good to see this class get its first coverage. As written, though, it only exercises hand-built non-domain `ClassNode`s, so the branches that actually matter to the transform never execute: `AstUtils.isDomainClass` is false in all three features, which means the injected `id`/`version` entries, the `hasMany`/`belongsTo`/`hasOne` handling in `populatePropertiesForInitialExpression`, the `isResolved()` / `ClassPropertyFetcher` path, and the superclass walk are all untested. This PR rewrites the population loop feeding every one of those, so per the repo rule that a touched class gets its behaviour verified they should be covered here — an `@Entity`-annotated `ClassNode` with a `hasMany` initial expression and a domain superclass would reach most of it.

Also missing: a test for the concurrency fix the PR claims. Several threads calling `getPropertyNames` concurrently on the same node and on distinct nodes, asserting every returned list is complete, would exercise both the synchronized map and the publish-after-populate ordering. As it stands, reverting either half of the concurrency change leaves this spec green.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    void "getPropertyType resolves and caches the type of a declared property"() {
+        given: 'a class node with a declared property'
+        ClassNode classNode = new ClassNode('org.example.PropertyTypeWidget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+        classNode.addProperty('label', Modifier.PUBLIC, ClassHelper.STRING_TYPE, null, null, null)
+
+        expect: 'the resolved property type matches the declared type, both on first (cache-populating) and second (cache-hit) lookup'
+        AstPropertyResolveUtils.getPropertyType(classNode, 'label') == ClassHelper.STRING_TYPE
+        AstPropertyResolveUtils.getPropertyType(classNode, 'label') == ClassHelper.STRING_TYPE
```
jdaugherty: The name says "and caches", but neither assertion can distinguish a cache hit from a miss — on a miss `getPropertyType` falls through to `classNode.getProperty(propertyName)` and returns the same `STRING_TYPE`, so both lines pass with caching entirely disabled.

To actually pin the caching behaviour: resolve once, then add a second property to the `ClassNode` and assert `getPropertyNames` still returns the *stale* view. That snapshot-at-first-lookup semantics is what this class really guarantees and what the transform relies on, and it's currently unverified.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,101 @@
...
+        given: 'two distinct ClassNode instances - as produced by two separate compilations - sharing an identical unqualified name'
+        ClassNode first = new ClassNode('Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+        first.addProperty('color', Modifier.PUBLIC, ClassHelper.STRING_TYPE, null, null, null)
+
+        ClassNode second = new ClassNode('Widget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+        second.addProperty('weight', Modifier.PUBLIC, ClassHelper.Integer_TYPE, null, null, null)
+
+        expect: 'the two ClassNode instances compare equal by name - the exact condition that would collide in a name-keyed or equals()-keyed cache'
+        first == second
+        first.hashCode() == second.hashCode()
```
jdaugherty: These two assert Groovy's own `ClassNode` equality contract rather than anything about `AstPropertyResolveUtils`. They hold today — 5.0.7's `ClassNode.equals` compares `getText()` and `hashCode()` delegates to `getText().hashCode()` — but if Groovy ever moves `ClassNode` to identity equality this spec fails while the production behaviour it guards is still perfectly correct.

`!first.is(second)` on line 79 is the precondition the test actually needs. Consider keeping that and demoting the other two to a comment explaining *why* two same-named nodes used to collide.

jdaugherty (commented):
I had AI review this before taking a look, here's it's comments: 

Reviewed the identity-keyed cache change. The direction is right — `ClassNode.equals`/`hashCode` really do compare by `getText()` (confirmed in Groovy 5.0.7: `equals` falls through to `((ClassNode) that).getText().equals(getText())`), so a name- or `equals()`-keyed static cache genuinely conflates distinct compilations, and reordering so an entry is published only once fully populated is a real fix.

Two things I'd want resolved before this lands:

1. **Keying a `static` map by `ClassNode` turns a bounded cache into a classloader leak.** A primary `ClassNode` transitively pins its `CompileUnit` and therefore its `GroovyClassLoader`, and nothing ever evicts. Detail inline — the per-node `getNodeMetaData` alternative removes the collision *and* the leak, and drops the static state entirely. This is the part I'm most concerned about, since the whole point of the change is to make a process-wide static cache safe.
2. **The stated root cause doesn't explain a *flaky* failure.** The same source compiled twice yields two `ClassNode`s with identical property sets, so a name collision between just those two can't diverge. The genuinely order-dependent input is `classNode.isResolved()` at first-lookup time, which this PR doesn't change. Inline.

Smaller items, all inline: the `protected` field's type change plus new `final` is a breaking change worth converting into `private` + an explicit clear hook; the javadoc's concurrency rationale cites parallel test forks, which are separate JVMs and so cannot race on a static field; and the new spec doesn't cover the domain-class branches or the concurrency behaviour it claims to fix.

One follow-up outside the diff: `WhereQueryClosureCaptureSpec:35` and `WhereQueryEmbeddedBlockTransformSpec:37` both still carry the comment *"The domain class names must be unique across the test JVM because AstPropertyResolveUtils caches resolved properties statically by class name"*. That is no longer true after this change, and the fixture renaming in 08bd0bfbeb that it justifies is now unnecessary. Please update both comments (and note whether the renaming can be reverted) so those specs aren't left documenting the old behaviour.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Agreed, and went with your suggested direction: the static map is gone entirely, and the cache is now stored as `ClassNode` metadata (`classNode.redirect().getNodeMetaData(key, fn)`), so a cache entry is only reachable through the node it describes and is collected with it. See the updated javadoc on `cachedClassProperties`'s replacement (`PROPERTIES_CACHE_KEY`) for the full writeup, including why keying on `redirect()` is safe here (`setRedirect()` throws for a *primary* node - i.e. every real caller of this utility, since they're all mid-compilation - so `redirect()` is just `this` for the node's whole life in practice) and why access ended up needing explicit per-node synchronization (some real callers can resolve to interned singletons like `ClassHelper.OBJECT_TYPE` for a plain `Object`/`def`-typed property, and those are shared across every compilation in the JVM, not scoped to one thread the way a normal node is). Landed in 078ac9da53, tightened further in c3a4d9235b after an adversarial self-review turned up the synchronization gap.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Went further than private + `clearCache()` - the static map is gone entirely (078ac9da53), so there's no field left to expose or protect. That's still technically a breaking removal for anyone who referenced the old `protected` field directly, just a cleaner one than a silent type change. Given it's an internal implementation-detail field with no getter/documented extension use, I'm inclined to leave it out of the upgrade guide, but flagging it in case there's a reason to cover it there.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: You're right, fixed (078ac9da53). The rationale no longer claims parallel test forks race on this - confirmed against `gradle/test-config.gradle` that `maxParallelForks` really does fork separate JVMs and this build never enables JUnit's in-JVM parallel execution, so that was never a real race. Rewrote the javadoc around the actual exposure instead: interned `ClassHelper` singletons (`OBJECT_TYPE`, `STRING_TYPE`, etc.) that any concurrently-running compilation in the same JVM could resolve to and touch through this cache - which is also why access is now synchronized per-node (c3a4d9235b) rather than resting on "only one thread ever touches a given node," which turned out not to be universally true.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: `getNodeMetaData(key, fn)` uses `computeIfAbsent` internally (078ac9da53), and I additionally wrapped the call in `synchronized (cacheHolder)` (c3a4d9235b) after an adversarial self-review flagged that the backing `ListHashMap` is explicitly documented as not thread-safe, and that some real callers can land on interned singleton nodes (`ClassHelper.OBJECT_TYPE` etc.) that more than one concurrent compilation could reach - the "only one thread per node" assumption doesn't hold for those. See the javadoc for the precise scope of what the synchronization does and doesn't cover.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Wanted to actually test this rather than argue about it. I reverted to the pre-fix cache and ran the exact "same source compiled twice" scenario `WhereQueryClosureCaptureSpec`'s second test exercises, 40 times sequentially in one JVM (simulating many test classes sharing one fork's static state) - it never diverged, even with the old buggy cache. So I can't point to a repro that pins the ~1% CI flake specifically to the name/identity-collision mechanism, and you may well be right that it isn't the actual trigger. I'm keeping the identity-safety fix regardless, since it's an independently real, provable bug on its own terms (the spec's second test shows two distinct, differently-shaped `ClassNode`s that happen to share a name getting their properties conflated under the old cache) - but I want to be upfront that I haven't confirmed it's *the* flake's cause, only that it's *a* real bug. I'll be watching the flaky-test dashboard (#16030) after this merges to see whether `WhereQueryClosureCaptureSpec` actually clears.

On the `isResolved()` snapshot mechanism specifically, I checked the Groovy 5.0.7 source directly: `ClassNode.clazz` has no setter anywhere outside the `ClassNode(Class)` constructor, and `setRedirect()` throws a `GroovyBugError` if called on a *primary* node - which is what every real caller of this utility passes in, since they're all resolving a class mid-compilation. So for the nodes this cache actually serves, `redirect()` is simply the node itself for its entire life, and `isResolved()` can't flip after the node is first cached. I don't think the staleness you're describing can occur for this code path as it's actually used, though I agree the "cache once, forever" design would be fragile if it could - documented the reasoning (and the primary-node guarantee it leans on) in the class javadoc.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
borinquenkid: Added both. Domain-class coverage now includes identity/version injection and `hasMany`/`belongsTo`/`hasOne` resolved two ways: via AST initial expressions (a hand-built `ClassNode` with a crafted `MapExpression`, matching what `static hasMany = [...]` compiles to) and via reflection on an already-resolved class (compiled through `GroovyClassLoader.parseClass` then re-wrapped with `ClassHelper.make()`, since that's the only way to get `isResolved() == true` for that branch).

Also added two concurrency tests: many threads each resolving their own distinct, identically-named `ClassNode` (proves the identity-collision-freedom property holds under concurrent load, not just sequentially), and - after an adversarial self-review pointed out the first test can't exercise any real race, since nothing is shared between the threads - a second test where 32 threads resolve the exact same shared `ClassNode` concurrently, which is what the `synchronized` fix mentioned above actually protects. Worth being honest about that second test's limits: I checked whether it fails without the `synchronized` guard, and it didn't, in 8 runs - the cached computation is deterministic and idempotent, so a black-box return-value test can't reliably force the underlying race into an observably wrong result. Said that directly in the test's comment rather than overclaiming what it proves; the synchronization is justified by `ListHashMap`'s own "not thread-safe" documentation, not by this test having caught a live bug.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
borinquenkid: Replaced it. The new test adds a second property to the `ClassNode` after the first `getPropertyNames` lookup and asserts the second lookup doesn't see it - `getPropertyNames` has no live-fallback path (unlike `getPropertyType`, which falls through to a direct `classNode.getProperty()` lookup when the cache doesn't contain the key), so this actually proves the result was cached rather than recomputed. It also asserts, via a direct `classNode.getProperty('extra') != null` check, that the property really was added to the underlying node - so the test demonstrates staleness specifically, not just a lookup that happens to return nothing.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
borinquenkid: Done - trimmed to just `!first.is(second)`, with a comment explaining why the `equals()`/`hashCode()` equality matters (it's the exact collision condition a name- or `equals()`-keyed cache would hit) rather than asserting on Groovy's own `ClassNode` equality contract.

borinquenkid:
@jdaugherty

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -94,22 +164,25 @@ public static List<String> getPropertyNames(ClassNode classNode) {
...
-        } return cachedProperties;
+        ClassNode cacheHolder = classNode.redirect();
+        synchronized (cacheHolder) {
+            return cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY, key -> computeProperties(cacheHolder));
+        }
+    }
+
+    private static Map<String, ClassNode> computeProperties(ClassNode classNode) {
+        Map<String, ClassNode> newProperties = new HashMap<>();
+        boolean isDomainClass = AstUtils.isDomainClass(classNode);
```
jdaugherty, line 218: `AstUtils.isDomainClass(ClassNode)` is `@Memoized` (`AstUtils.groovy:428`), and that memo cache is keyed by `ClassNode` *equality*, not identity — `ClassNode.equals`/`hashCode` both reduce to `getText()` (`ClassNode.java:1283`/`1288` in 5.0.7). `@Memoized` also defaults to `maxCacheSize = 0`, so it is an unbounded `UnlimitedConcurrentCache` that never evicts.

That leaves both headline claims in the new javadoc false one level down: two distinct same-named `ClassNode`s do still collide here, and the first node seen per distinct name is retained for the life of the JVM, pinning its module, compile unit and `GroovyClassLoader` with it.

Verified against this branch with two hand-built nodes both named `probe.MemoWidget`, the first plain and the second genuinely carrying `@Entity`, resolved in that order:

```
plainProps            = [title]
entityProps           = [title]     <- @Entity node, no id/version
isDomainClass(entity) = false       <- despite the annotation
```

Control, the same `@Entity` node under a name nothing else had touched: `[id, title, version]`.

To be fair to the change: the `@Memoized` predates this PR and you are not introducing it. But this PR does newly route through it. Under the old name-keyed map, a same-named second node returned the first node's cached map and never called `isDomainClass` at all; now it computes its own properties and consults the memoized verdict. The corruption narrows from the entire property map to the domain-ness boolean, which is a genuine improvement — "no collision is possible" just overstates where it lands.

I am not asking you to fix `isDomainClass` here; that deserves its own issue rather than growing this PR. Just asking that the javadoc claims match what is actually true.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -94,22 +164,25 @@ public static List<String> getPropertyNames(ClassNode classNode) {
...
-            }
-            cachedClassProperties.put(className, cachedProperties);
-            ClassNode currentNode = classNode;
-            while (currentNode != null && !currentNode.equals(ClassHelper.OBJECT_TYPE)) {
-                populatePropertiesForClassNode(currentNode, cachedProperties, isDomainClass, !isDomainClass);
-                currentNode = currentNode.getSuperClass();
-            }
-        } return cachedProperties;
+        ClassNode cacheHolder = classNode.redirect();
+        synchronized (cacheHolder) {
```
jdaugherty, line 199: `synchronized (cacheHolder)` cannot deliver the safety the javadoc claims for it, because the nodes being locked are not always private to one compilation. `QueryStringTransformer.groovy:208` walks a property path by feeding each resolved type back in as the next receiver, so a JDK-typed property makes an interned `ClassHelper` singleton the cache holder.

Verified on this branch:

```
propType.is(ClassHelper.STRING_TYPE) = true
STRING_TYPE cache metadata before    = false
STRING_TYPE cache metadata after     = true
```

`ClassNode.getModule()` reads that same `ListHashMap` through `getNodeMetaData(ModuleNode.class)` without taking the node monitor, so this lock only excludes callers that opt into it. The javadoc frames the leftover exposure as risk that "belongs to `ClassNode`'s metadata storage in general", but before this change the utility never wrote to any `ClassNode` — it is this PR that makes GORM a writer to those JVM-wide maps.

Gating on `cacheHolder.isPrimaryClassNode()` and skipping the cache otherwise would confine writes to nodes the compilation actually owns, and the JDK types that get skipped are cheap to recompute.

Separately, `computeProperties` runs entirely inside this monitor, and for a resolved domain class it reaches `ClassPropertyFetcher.forClass(getTypeClass()).getPropertyValue(...)`, which forces the user class's static initialiser and invokes user-written static getters while the lock is held. Computing outside the lock and publishing afterwards would take alien code out from under it.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
```diff
@@ -94,22 +164,25 @@ public static List<String> getPropertyNames(ClassNode classNode) {
...
+            return cacheHolder.getNodeMetaData(PROPERTIES_CACHE_KEY, key -> computeProperties(cacheHolder));
+        }
+    }
+
+    private static Map<String, ClassNode> computeProperties(ClassNode classNode) {
+        Map<String, ClassNode> newProperties = new HashMap<>();
+        boolean isDomainClass = AstUtils.isDomainClass(classNode);
+        if (isDomainClass) {
+            newProperties.put(GormProperties.IDENTITY, new ClassNode(Long.class));
```
jdaugherty: Minor: both of these are redirect-less holder nodes, so under the new scheme each becomes its own cache holder and never shares with anything. Line 236 already uses the plain-reference convention — `ClassHelper.make(Long.class).getPlainNodeReference()` here would redirect to the interned `Long` node so id/version lookups share a single entry again.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/WhereQueryClosureCaptureSpec.groovy
```diff
@@ -31,8 +31,14 @@ import spock.lang.Specification
...
 class WhereQueryClosureCaptureSpec extends Specification {
 
-    // The domain class names must be unique across the test JVM because
-    // AstPropertyResolveUtils caches resolved properties statically by class name
+    // Historical note: these domain class names were made unique across the test JVM
+    // (ClosureCaptureBook/ClosureCaptureAuthor rather than the more generic Book/Author)
+    // because AstPropertyResolveUtils used to cache resolved properties in a single
+    // static map keyed by class name, so a same-named fixture in another spec could
+    // collide with this one. AstPropertyResolveUtils now caches per-ClassNode instance
+    // (see its javadoc), so that collision can no longer happen regardless of naming -
```
jdaugherty: Of everything here, this is the change I would most like reverted. Given the `@Memoized` `isDomainClass` collision (see the comment on `computeProperties`), uniqueness *is* still required for correctness here — the original warning was right.

The risk is concrete inside this module. `DirtyCheckTransformationSpec` compiles `@DirtyCheck class Book` and `class Author` from strings, which are not domain classes, while `QueryStringTransformerSpec` compiles `@Entity class Book` and `class Author`, which are. `forkEvery = 50` on CI (`gradle/test-config.gradle:88`) means up to fifty spec classes share one JVM, so both can land in the same fork in either order, and whichever runs first wins the memoized `isDomainClass` answer for that name.

A maintainer who takes this comment at face value and renames `ClosureCaptureBook` back to `Book` walks straight into an order-dependent flake of exactly the #16030 character.

Please keep the "must be unique" wording as a requirement. If you want to record the history, the accurate version is that the *property map* is now per-instance while the domain-class verdict is still name-keyed.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/WhereQueryEmbeddedBlockTransformSpec.groovy
```diff
@@ -33,8 +33,13 @@ import spock.lang.Specification
  */
 class WhereQueryEmbeddedBlockTransformSpec extends Specification {
 
-    // The domain class names must be unique across the test JVM because
-    // AstPropertyResolveUtils caches resolved properties statically by class name
+    // Historical note: these domain class names were made unique across the test JVM
+    // because AstPropertyResolveUtils used to cache resolved properties in a single
+    // static map keyed by class name, so a same-named fixture in another spec could
+    // collide with this one. AstPropertyResolveUtils now caches per-ClassNode instance
+    // (see its javadoc), so that collision can no longer happen regardless of naming -
```
jdaugherty: Same point as the equivalent comment in `WhereQueryClosureCaptureSpec`: "that collision can no longer happen regardless of naming" does not hold while `AstUtils.isDomainClass` is `@Memoized` by name, so the uniqueness requirement should stay stated as a requirement rather than a stylistic preference.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,280 @@
...
+        second.addProperty('weight', Modifier.PUBLIC, ClassHelper.Integer_TYPE, null, null, null)
+
+        and: 'they are genuinely different instances - the precondition a name- or equals()-keyed cache would get wrong'
+        // ClassNode#equals()/hashCode() compare by getText() (essentially the class name), so
+        // first == second and first.hashCode() == second.hashCode() both hold here even though
+        // these are two unrelated ClassNode instances with different declared properties. A cache
+        // keyed by name or by equals()/hashCode() would treat them as the same entry; only
+        // reference identity (!first.is(second)) tells them apart, which is exactly what the
+        // cache must key on.
+        !first.is(second)
```
jdaugherty: This precondition never runs as an assertion. It sits in an `and:` block continuing `given:`, and Spock applies implicit conditions only in `then:`/`expect:` blocks, so the expression is evaluated and discarded.

Confirmed on this branch by putting `false` and `x == 999` in an `and:` after `given:` in a scratch spec — the feature passed.

Needs `assert !first.is(second)`, or to move into an `expect:` block. Worth correcting because the comment directly above presents this line as the load-bearing premise of the feature.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
```diff
@@ -0,0 +1,280 @@
...
+                ClassNode node = new ClassNode('ConcurrentWidget', Modifier.PUBLIC, ClassHelper.OBJECT_TYPE)
+                String propertyName = "prop${i}".toString()
+                node.addProperty(propertyName, Modifier.PUBLIC, ClassHelper.STRING_TYPE, null, null, null)
+
+                List<String> names = AstPropertyResolveUtils.getPropertyNames(node)
+                names.contains(propertyName) && names.count { it.startsWith('prop') } == 1
+            } as Callable<Boolean>)
+        }
+        List<Boolean> outcomes = futures.collect { Future<Boolean> future -> future.get(30, TimeUnit.SECONDS) }
+        executor.shutdown()
```
jdaugherty: `executor.shutdown()` is the last statement of the `when:` block, so it is skipped on exactly the path where it matters: if any `future.get(30, SECONDS)` throws, the `collect` aborts and the pool is never shut down. `Executors.newFixedThreadPool` threads are non-daemon and do not time out, and with `forkEvery = 50` this JVM goes on to run more specs.

`barrier.await()` also has no timeout, so if one worker dies before reaching the barrier the rest park indefinitely and the only symptom is a 30-second timeout per future.

Suggest a `cleanup:` block calling `executor.shutdownNow()` in both concurrency features, plus the timeout overload of `await`. The `GroovyClassLoader` in the reflection feature is also never closed — it is `AutoCloseable`, and closing it would suit a spec whose subject is a classloader leak.

jdaugherty (commented):
Re-reviewed after the move to per-`ClassNode` metadata. Everything from the previous round is addressed and I have resolved those threads — the static map is gone, publication is safe, and the tests now reach the domain-class branches. The empirical reproduction you did settled the question of whether the flake mechanism was real, which was the right way to answer it.

Two major remain:

1. `AstUtils.isDomainClass` is `@Memoized` by name, so "no collision is possible" and "no leak is possible" are not accurate yet. I verified a genuinely `@Entity` node losing its `id`/`version` because a same-named plain node was resolved first. The fix is still a clear improvement — it narrows the corruption from the whole property map to a single boolean — but the javadoc and the two spec comments should say what is actually true. The retracted "names must be unique" warning is the one I most want restored, since `Book`/`Author` really are shared across specs in this module and `forkEvery = 50` puts them in one JVM.

2. `!first.is(second)` in the new spec sits in a `given:`/`and:` block, so it never asserts.

The rest is smaller: the cache now writes into interned `ClassHelper` singletons that `ClassNode.getModule()` reads without the lock, alien code runs under the node monitor, and the new spec leaks its executors and classloader on the failure path.

jdaugherty:
@borinquenkid

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Agreed, and fixed in 0d195f1493. Corrected the javadoc's "no collision is possible" claim to scope it explicitly to the resolved property map, and added a paragraph documenting the `isDomainClass` memo-cache exposure you found — noting it's a separate, pre-existing issue this class neither causes nor fixes. Also restored the "must stay unique" wording in both `WhereQuery*Spec` comment blocks accordingly (see replies below).

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Both points fixed in 0d195f1493.

`getPropertiesFromCache` now checks `classNode.isPrimaryClassNode()` first and skips the cache entirely for non-primary nodes — which is exactly what `ClassHelper.make(Class)`'s interned singletons are (verified against the 5.0.7 source: the `ClassNode(Class)` constructor sets `isPrimaryNode = false`). So this class no longer writes its metadata key into a node it doesn't own, closing the `QueryStringTransformer` path you traced. These nodes are recomputed on every call instead, which is cheap given there are only a handful of them.

Separately, `computeProperties` now runs outside the `synchronized (cacheHolder)` block — the lock is only taken to check for and then publish the result, so the static-initializer/getter execution via `ClassPropertyFetcher` no longer happens while holding the per-node monitor. Two threads can both compute for the same as-yet-uncached node; the second to acquire the lock discards its own (equivalent) result and returns the one already published.

Updated the javadoc to describe both.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtils.java
borinquenkid: Done — `IDENTITY`/`VERSION` now use `ClassHelper.make(Long.class).getPlainNodeReference()`, matching the convention already used for the `hasMany`/`belongsTo`/`hasOne` values a few lines down.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/WhereQueryClosureCaptureSpec.groovy
borinquenkid: You're right, reverted. Restored the "must stay unique" wording and spelled out why: the property-map fix doesn't cover `isDomainClass`'s own memoization, so a same-named, differently-shaped fixture in another spec sharing this JVM fork (`forkEvery = 50`) could still hand this class a stale domain-class verdict from that memo cache. Renaming back to `Book`/`Author` would risk exactly the order-dependent flake you described.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/WhereQueryEmbeddedBlockTransformSpec.groovy
borinquenkid: Same fix applied here — restored the uniqueness requirement with the same rationale as the equivalent comment in `WhereQueryClosureCaptureSpec`.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
borinquenkid: Confirmed and fixed — good catch, that precondition was silently never checked. Added an explicit `assert`.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/transform/AstPropertyResolveUtilsSpec.groovy
borinquenkid: Fixed both concurrency tests: `executor.shutdownNow()` moved into a `cleanup:` block so it runs even if a `future.get()` throws, and `barrier.await()` now takes the 30s timeout so a dead worker fails fast instead of parking the rest indefinitely. Also closed the `GroovyClassLoader` in the reflection test.

borinquenkid:
> @borinquenkid
Im confused

jdaugherty (approved):
Re-verified this round against 0d195f1493 — every item from the previous reviews is addressed in the code, not just in the replies: the javadoc's collision claim is now correctly scoped to the property map with the `isDomainClass` memo exposure documented as the separate, pre-existing issue it is; non-primary (interned/class-backed) nodes are no longer written to; `computeProperties` runs outside the per-node monitor with a double-checked publish; `IDENTITY`/`VERSION` use plain node references; the uniqueness requirements are restored in both Where specs with accurate rationale; and the test fixes (explicit assert for the silent precondition, `cleanup:` with `shutdownNow()`, timed `barrier.await`, closed `GroovyClassLoader`) are all in place.

CI is green across the board and `:grails-datamapping-core` passes locally on this branch, including all three touched specs.

Two follow-ups worth tracking outside this PR, neither blocking: (1) the `AstUtils.isDomainClass` name-keyed memoization deserves its own issue as previously discussed — please file it so the javadoc's "tracked separately" has something to point at; (2) if compile-time profiling ever shows the uncached recompute path mattering for precompiled (non-primary, class-backed) domain-class nodes, a `ClassValue` keyed on the resolved `Class` would be a targeted improvement over re-widening the cache.

PR #16040, opened 2026-07-22
Adopting Spring Boot Multipart Configuration for File Uploads

### LINKED ISSUE #15644: Adopting Spring Boot Multipart Configuration for File Uploads
### Feature description

Currently, file upload limits in Grails are configured as follows:

```
grails:
  controllers:
    upload:
      maxFileSize: 20000000
      maxRequestSize: 20000000
```

In Spring Boot, the equivalent configuration is:

```
spring:
  servlet:
    multipart:
      max-file-size: 200MB
      max-request-size: 200MB
```

I propose that we adopt the Spring Boot configuration properties instead.

This would provide several benefits:
- Avoid reinventing existing functionality
- Reduce redundant and framework-specific configuration code
- Improve readability by allowing the use of human-friendly size units (e.g., MB instead of raw byte values)

gsartori:
## Description

Fixes [#15644](https://github.com/apache/grails-core/issues/15644).

This change delegates multipart upload configuration to Spring Boot instead of creating and configuring a Grails-specific `MultipartConfigElement`.
Applications should now use the standard `spring.servlet.multipart` properties:

```yaml
spring:
    servlet:
        multipart:
            max-file-size: 200MB
            max-request-size: 200MB
```

The previous `grails.controllers.upload` configuration namespace is no longer supported. If legacy properties are detected, application startup fails with migration instructions instead of silently ignoring them.

The change also:

- Removes the obsolete upload constants from `grails.config.Settings`.
- Removes the corresponding Grails configuration metadata.
- Allows Spring Boot’s multipart auto-configuration to supply the `MultipartConfigElement`.
- Makes servlet registration tolerate multipart support being disabled or otherwise unavailable.
- Adopts Spring Boot’s defaults of 1 MB per file and 10 MB per multipart request.
- Adds a regression test covering detection of legacy configuration.
- Updates the uploading-files documentation and threat model with the new properties and defaults.
`GrailsHibernatePersistentEntity` is not changed by these commits.

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team: [#15644](https://github.com/apache/grails-core/issues/15644).
- [ ] This PR addresses the **complete scope** of the linked issue.
- [x] This PR targets the `8.0.x` major release branch, where breaking configuration changes are permitted.

### Code Quality

- [ ] I have run `./gradlew codeStyle` and resolved any violations.
- [x] This PR does **not** include mass reformatting, style-only changes, or unrelated large-scale refactoring.
- [ ] If generative AI tooling was used, a quality model was used to ensure consistency with the project’s quality standards.

### Licensing and Attribution

- [ ] All contributed code is provided under the Apache License 2.0.
- [ ] I have the necessary rights to submit this contribution and confirm it is my own original work.
- [ ] If generative AI tooling was used, I have followed the ASF policy and properly attributed its use.

### Documentation

- [x] The relevant user-facing documentation has been updated.
- [ ] The **What’s New** section of the Grails Guide has been updated.
- [ ] The Grails 8 **Upgrade Notes** have been updated to document migration from `grails.controllers.upload` to `spring.servlet.multipart`.


grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
```diff
@@ -54,6 +57,27 @@ class ControllersAutoConfigurationSpec extends Specification {
...
+        applicationContext.environment.propertySources.addFirst(new MapPropertySource('test', [
+                'grails.controllers.upload.maxFileSize': 20000000,
+        ]))
+        applicationContext.register(ControllersAutoConfiguration)
+
+        when:
+        applicationContext.refresh()
+
+        then:
+        BeanCreationException exception = thrown()
```
matrei: Consider using: `def exception = thrown(BeanCreationException)` (for readability)?

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
```diff
@@ -58,7 +63,15 @@
         after = {GrailsDomainClassAutoConfiguration.class}
 )
 @ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
-public class ControllersAutoConfiguration {
+public class ControllersAutoConfiguration implements EnvironmentAware {
+
+    private static final String LEGACY_MULTIPART_CONFIGURATION = "grails.controllers.upload";
+
+    static final String LEGACY_MULTIPART_CONFIGURATION_ERROR =
```
matrei: `private`?

grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
```diff
@@ -54,6 +57,27 @@ class ControllersAutoConfigurationSpec extends Specification {
...
+        exception.rootCause instanceof IllegalStateException
+        exception.rootCause.message == ControllersAutoConfiguration.LEGACY_MULTIPART_CONFIGURATION_ERROR
```
matrei: Consider not using constants in tests. Write out the expected value instead?

matrei (approved):

gsartori:
@martei I've added a paragraph in the upgrade guide here as well

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -136,6 +138,34 @@ spring:
...
+Applications that still define any property under `grails.controllers.upload` fail at startup with migration instructions.
+Move those settings to Spring Boot's `spring.servlet.multipart.*` namespace:
+
+[source,yaml]
+.application.yml - Before (Grails 7)
+----
+grails:
+    controllers:
+        upload:
+            maxFileSize: 200MB
```
matrei: Was the `200MB` syntax compatible with old properties?

gsartori: Nope, I've corrrected it

jdaugherty (commented):
This is a breaking change. I am ok with this but I would like to know if @jamesfredley is ok with this

gsartori:
@jdaugherty I think I need help with the failing tests since I didn't manage to understand the link between my code and the failures

gsartori:
@jdaugherty @jamesfredley I've let AI fix the issue with the logger. <ins>I need someone to review the changes</ins> since this time I let her do the job altogether. Changes are small but I have no experience on that area.

matrei:
@gsartori I reverted your last change. The tests are green now :heavy_check_mark:

grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
```diff
@@ -54,6 +57,31 @@ class ControllersAutoConfigurationSpec extends Specification {
...
+        exception.rootCause instanceof IllegalStateException
+        exception.rootCause.message ==
+                "Configuration properties under 'grails.controllers.upload' are no longer supported. " +
+                "Use Spring Boot's 'spring.servlet.multipart' configuration instead. For example, set " +
+                "'spring.servlet.multipart.maxFileSize=200MB' and " +
+                "'spring.servlet.multipart.maxRequestSize=200MB'."
+
+        cleanup:
+        applicationContext.close()
+    }
```
jdaugherty: `dispatcherServletRegistration` changed signature to `ObjectProvider<MultipartConfigElement>`, but nothing exercises it. Could this spec add three cases using the `WebApplicationContextRunner` already used below?

- Boot supplies the element and it reaches the registration — the behavior this PR is preserving
- `spring.servlet.multipart.enabled=false` leaves the registration without multipart config
- no legacy property present → context starts cleanly (negative case for the new guard)

Adding `MultipartAutoConfiguration` to `AutoConfigurations.of(...)` is enough to get the element, and `DispatcherServletRegistrationBean.getMultipartConfig()` exposes the result.

`grails-test-examples/app1/src/integration-test/groovy/functionaltests/fileupload/FileUploadSpec.groovy` would also be a good home for an end-to-end case: a file over `maxFileSize` is rejected, and raising `spring.servlet.multipart.maxFileSize` takes effect.

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
```diff
@@ -150,12 +151,11 @@ public FilterRegistrationBean<GrailsWebRequestFilter> grailsWebRequestFilter(Gra
         return registrationBean;
     }
 
-    @Bean
-    public MultipartConfigElement multipartConfigElement() {
-        if (uploadTmpDir == null) {
-            uploadTmpDir = System.getProperty("java.io.tmpdir");
+    @Override
+    public void setEnvironment(Environment environment) {
+        if (Binder.get(environment).bind(LEGACY_MULTIPART_CONFIGURATION, Bindable.mapOf(String.class, Object.class)).isBound()) {
```
jdaugherty: Two things on the guard:

1. The message surfaces as a nested cause inside a `BeanCreationException` stack trace — no `FailureAnalyzer` matches an `IllegalStateException` thrown from `setEnvironment`, so the migration instructions never reach Boot's `APPLICATION FAILED TO START` / `Action:` block where a user would look for them. Either register a `FailureAnalyzer`, or move the check into an `EnvironmentPostProcessor` (grails-core already registers those in `META-INF/spring.factories`) so it fires before context creation. Secondary point: `setEnvironment` is a wiring callback, and a validation that throws is unexpected there.

2. The check is namespace-wide, so it can't tell the user which key to fix — any property under `grails.controllers.upload` trips it, including keys that were never Grails settings. The bound map already carries the names:

```java
BindResult<Map<String, Object>> legacy = Binder.get(environment)
        .bind(LEGACY_MULTIPART_CONFIGURATION, Bindable.mapOf(String.class, Object.class));
if (legacy.isBound()) {
    throw new IllegalStateException(LEGACY_MULTIPART_CONFIGURATION_ERROR
            + " Found: " + legacy.get().keySet());
}
```

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -136,6 +138,34 @@ spring:
...
+.application.yml - After (Grails 8)
+----
+spring:
+    servlet:
+        multipart:
+            maxFileSize: 200MB
+            maxRequestSize: 200MB
+----
+
+Spring Boot also provides `spring.servlet.multipart.location`, `spring.servlet.multipart.fileSizeThreshold`, and `spring.servlet.multipart.enabled` as replacements for the corresponding legacy settings.
```
jdaugherty: §4.2 covers the namespace move but not the two runtime changes an upgrading operator would notice:

- **Temp directory.** `location` used to default to `System.getProperty("java.io.tmpdir")`; Boot leaves it unset, so the servlet container's temp directory is used instead (the resulting `MultipartConfigElement` has `location=""`). Worth a sentence for anyone who relied on uploads landing in `java.io.tmpdir`, or who runs with a constrained container temp dir.
- **Effective limits when nothing is configured.** An app that never set an upload limit now picks up Boot's `MultipartProperties` defaults — 1MB per file, 10MB per request. Stating that here, next to the advice to set an explicit limit, saves the reader a hop to `uploadingFiles.adoc`.

grails-doc/src/en/guide/theWebLayer/controllers/uploadingFiles.adoc
```diff
@@ -103,16 +103,19 @@ You can configure the limit in your `application.yml` as follows:
...
+            maxRequestSize: 200MB
 ----
 
-`maxFileSize` = The maximum size allowed for uploaded files.
+`maxFileSize` = The maximum size allowed for an uploaded file.
 
-`maxRequestSize` =  The maximum size allowed for multipart/form-data requests.
+`maxRequestSize` = The maximum size allowed for a multipart/form-data request.
+
+The `spring.servlet.multipart` namespace also supports Spring Boot's `location`, `fileSizeThreshold`, and `enabled` properties.
```
jdaugherty: Boot 4.1 also exposes `spring.servlet.multipart.resolveLazily` and `spring.servlet.multipart.strictServletCompliance` under this namespace. Either add them, or link Boot's multipart configuration docs so this list doesn't need maintaining.

jdaugherty (commented):
Remaining comments are about the migration surface and test coverage.  

It looks like cf2e542 was pulled into this PR and then reverted by Mattias - we should be sure to squash this PR when merged.

grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
matrei: Should now be addressed

grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
matrei: Should now be addressed

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
matrei: Should now be addressed

grails-doc/src/en/guide/theWebLayer/controllers/uploadingFiles.adoc
matrei: Should now be addressed

jdaugherty:
@matrei we merged https://github.com/apache/grails-core/pull/16149 - is this PR even needed still? I think we can just close it, yes?

matrei:
> @matrei we merged #16149 - is this PR even needed still? I think we can just close it, yes?

@jdaugherty No, it is still needed. #16149 changed how an *oversized* upload fails, not where the limits come from. On `8.0.x` today `ControllersAutoConfiguration` still builds its own `MultipartConfigElement` from `grails.controllers.upload.*`, with the 128000-byte defaults, and #15644 (adopt `spring.servlet.multipart`) is still open. That is exactly what this PR does, and nothing in #16149 overlaps with it.

It does need a rebase though. Two files conflict, and a few things #16149 added now reference the old properties and have to move over with it:

- `ControllersAutoConfiguration.java` — the `dispatcherServlet` bean now takes an `Environment` for the hidden-method resolution, and the `multipartConfigElement` bean this PR removes sits right next to it.
- `THREAT_MODEL.md` / `threat-model.yaml` — the upload-limit row now describes `ControllersAutoConfiguration` overriding Boot's `MultipartProperties`; it should describe Boot's 1 MB / 10 MB defaults instead.
- Upgrade guide §53.1 (oversized uploads) and its Undertow note name `grails.controllers.upload.maxFileSize` / `maxRequestSize` and the 128000-byte defaults.
- `FileUploadSpec` on `8.0.x` gained an "upload exceeding the configured limit" test that posts 200000 bytes against the 128000 default. With this PR's app1 config (`spring.servlet.multipart.maxRequestSize: 3KB`) it still trips the limit, but the `given:` label should be updated.

@gsartori are you able to rebase, or would you like me to push the rebase onto your branch?

jamesfredley:
This still needs a rebase onto current `8.0.x` before it can merge.

Once rebased, please also move the leftover `grails.controllers.upload` references that landed in #16149:

- `ControllersAutoConfiguration.java` — keep the current `dispatcherServlet(Environment)` / hidden-method wiring; only drop the Grails `multipartConfigElement` bean and use Boot's `MultipartConfigElement` via `ObjectProvider`
- `THREAT_MODEL.md` / `threat-model.yaml` — describe Boot's 1 MB / 10 MB defaults instead of Grails overriding them
- Upgrade guide oversized-upload notes — point at `spring.servlet.multipart` instead of `grails.controllers.upload` and the 128000-byte defaults
- `FileUploadSpec` — update the over-limit test's `given:` label

I approve this after that rebase and cleanup are done. Please squash on merge.

matrei:
@jamesfredley 

I have merged `8.0.x` into the branch and resolved the overlap with #16149:

- `ControllersAutoConfiguration.java` — kept the `Environment`-aware `dispatcherServlet` bean from #16149, dropped the Grails `MultipartConfigElement` bean as this PR intends.
- `THREAT_MODEL.md` — kept the new P10 XXE row next to the reworded P9.
- Upgrade guide §53.1 and its Undertow note now name `spring.servlet.multipart.*` and Boot's 1 MB / 10 MB defaults, and link to the new §4.2 migration section.
- The oversized-upload specs #16149 added to the app1, app4, jetty and undertow example apps relied on the old 128000-byte default; with Boot's defaults a 200000-byte payload would no longer trip the limit. Those apps now configure `spring.servlet.multipart` limits and the spec labels say so.

Verified locally: `:grails-controllers:test`, the four upload-limit integration specs on Tomcat, Jetty and Undertow (26 tests, all green), and `aggregateViolations` reports no Checkstyle, CodeNarc, PMD or SpotBugs issues.

codeconsole (changes requested):
I am not a fan of introducing a `GrailsControllersEnvironmentPostProcessor` just to maintain a legacy configuration.

This is 8.0.x, let's just get rid of legacy configuration OR deprecate the temporary post processor for removal in 9.0.x. Having the post processor prohibits converting the ControllerAutoConfiguration to pure bean dsl.

matrei:
@codeconsole Thanks for the review. One clarification first: `GrailsControllersEnvironmentPostProcessor` does not maintain the legacy configuration. `grails.controllers.upload.*` is fully removed in this PR (the `Settings` constants, the configuration metadata and the Grails `MultipartConfigElement` bean are all gone). The post processor only *rejects* those properties at startup with migration instructions; nothing is translated or honored.

I would prefer not to drop it outright, for two reasons:

- Without it the removed properties are silently ignored, and an application that had set, say, `maxFileSize: 128000` quietly moves to Boot's 1 MB / 10 MB limits. The upload limits are a security-relevant knob in `THREAT_MODEL.md`, so a silent loosening on upgrade is the worst way for this to fail.
- The guard was added at @jdaugherty's request in the earlier review round, as an `EnvironmentPostProcessor` specifically so the message surfaces before context creation.

So I took your second option: the class is now `@Deprecated(forRemoval = true, since = "8.0")`, its Javadoc states it is a temporary migration aid that goes away in 9.0 without replacement, and the upgrade guide (§4.2) says the same.

On the bean DSL conversion: the post processor is registered on its own in `META-INF/spring.factories` and `ControllersAutoConfiguration` does not reference it, so it should not stand in the way of converting the auto-configuration. If I am missing a coupling there, point me at it and I will sort it out.

jdaugherty:
@codeconsole The post processor won't do anything if you use the correct settings.  As Mattias stated this is only a bridge for 8.0.

codeconsole (approved):

codeconsole:
> @codeconsole The post processor won't do anything if you use the correct settings. As Mattias stated this is only a bridge for 8.0.

That's why I wanted the deprecation tag so we don't forget to remove it.

PR #16042, opened 2026-07-22
Clean up GlobalGrailsClassInjectorTransformation

matrei:


borinquenkid:
@matrei I pushed a commit adding unit test coverage for the paths codecov flagged in `GlobalGrailsClassInjectorTransformation` (patch coverage was at 55.68%) — should be resolved now once CI/codecov re-runs.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -347,17 +303,104 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+                    resourceNode.replaceNode {}
                 }
             }
         }
     }
 
-    public static final ClassNode ARTEFACT_CLASS_NODE = new ClassNode(Artefact)
+    @CompileDynamic
+    private static Writable createMarkup(GPathResult node) {
+        new StreamingMarkupBuilder().mkp.yield(node)
```
jdaugherty: This throws every time it is called. `mkp` is not a property of `StreamingMarkupBuilder` — it is a pseudo-namespace that only resolves against the builder's delegate *inside* a `bind {}` closure, so accessing it on the instance fails:

```
groovy.lang.MissingPropertyException: No such property: mkp for class: groovy.xml.StreamingMarkupBuilder
Possible solutions: qt
```

(verified against groovy-xml 5.0.7; `bind { mkp.yield(node) }` on the same node returns the expected `<plugin name='foo'>...</plugin>`.)

Because `updatePluginXml` wraps everything in `catch (ignored)`, the failure is swallowed and the method falls through to `writePluginXml`. Two consequences:

- For the `pluginClassNode == null` branch — `generatePluginXml` → `updatePluginXml(null, ...)`, which is the path every non-descriptor source unit takes — the fallback `writePluginXml(null, ...)` is a no-op because of its `if (pluginClassNode)` guard. Artefact resources discovered in source units compiled after the `*GrailsPlugin` class are therefore silently dropped from `META-INF/grails-plugin.xml`, and `pendingPluginClasses` is never cleared.
- For the `pluginClassNode != null` branch the descriptor is fully regenerated from the AST on every pass rather than merged, so anything already in the file that is not reproducible from the AST is lost.

Keeping the `bind {}` form preserves the original behaviour and still gets you the extracted helper:

```groovy
@CompileDynamic
private static Writable createMarkup(GPathResult node) {
    new StreamingMarkupBuilder().bind { mkp.yield(node) }
}
```

(`bindNode(node)` — defined as `bind { out << node }` — may also fit, but it is a different rendering call than `mkp.yield`, so it would need verifying against the descriptor output.)

Worth a test that fails on the current code: put an element into an existing `plugin.xml` that the AST cannot reproduce (or call `updatePluginXml(null, '1.0', file, ['Foo'])`) and assert it survives the update.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -347,17 +303,104 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+    }
 
-    CompilationUnit compilationUnit
+    private static Object resolveProjectVersion(ClassNode classNode) {
+        def projectVersion = classNode.getNodeMetaData('projectVersion')
+        if (projectVersion == null) {
+            projectVersion = getClass().package.implementationVersion
```
jdaugherty: Moving this expression out of the instance method `visit()` and into a `static` helper changes what `getClass()` resolves to. In a static Groovy method the implicit receiver is the class object, so `getClass()` returns `java.lang.Class` — not `GlobalGrailsClassInjectorTransformation`:

```groovy
@CompileStatic
class Probe {
    static Object fromStatic()   { getClass() }   // -> class java.lang.Class
    Object fromInstance()        { getClass() }   // -> class Probe
}
```

`Class.class.getPackage().getImplementationVersion()` is `null` (java.base carries no manifest attributes), whereas the previous instance-context lookup read `Implementation-Version` off the grails-core jar — which `CompilePlugin.groovy` does set to `grailsVersion`. So this fallback now always yields `null` instead of the Grails version.

That propagates: `pluginVersion` becomes `null`, and `addPluginVersionProperty` calls `pluginVersion.toString()` with no null guard (line 349), so a `*GrailsPlugin` class compiled without `projectVersion` node metadata — plain `groovyc`, Groovy-Eclipse, anything that is not the Grails Gradle plugin stamping the customizer — now NPEs during CANONICALIZATION rather than getting the framework version.

Either keep the lookup anchored to the class explicitly:

```groovy
projectVersion = GlobalGrailsClassInjectorTransformation.package.implementationVersion
```

or pass the resolved version in. Same latent issue exists in `writePluginXml`/`writePluginXmlProperties`, but those were already static before this PR, so it is pre-existing there rather than introduced.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -285,60 +273,28 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
-            pluginXmlFile.withWriter(StandardCharsets.UTF_8.name()) { Writer writer ->
-                writable.writeTo(writer)
+            pluginXmlFile.withWriter(StandardCharsets.UTF_8.name()) {
+                createMarkup(pluginXml).writeTo(it)
             }
 
             pendingPluginClasses.clear()
 
-        } catch (e) {
+        } catch (ignored) {
```
jdaugherty: This catch-all is what turns the `createMarkup` defect above into a silent, test-invisible fallback: it catches every `Exception`, not just the "XML is corrupt" case the comment describes, and discards it without a trace.

Given the transform runs inside the compiler, a log line (or narrowing to the parse/IO failures that actually mean "corrupt, recreate") would have surfaced the `MissingPropertyException` immediately. Naming the parameter `ignored` also asserts the exception is uninteresting, which is the opposite of true here.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -347,17 +303,104 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
     protected static void handleExcludes(GPathResult pluginXml) {
         if (pluginExcludes) {
-
             def antPathMatcher = new AntPathMatcher()
-            pluginXml.resources.resource.each { res ->
-                if (pluginExcludes.any() { String exc -> antPathMatcher.match(exc, res.text().replace('.', '/')) }) {
-                    res.replaceNode {}
+            pluginXml.resources.resource.each {
+                def resourceNode = it as GPathResult
+                if (pluginExcludes.any() { antPathMatcher.match(it, resourceNode.text().replace('.', '/')) }) {
```
jdaugherty: The inner closure's implicit `it` shadows the outer `each`'s `it`, so this line reads as if the resource were being matched against itself. It is in fact correct — I confirmed the inner `it` binds to the exclude pattern and `resourceNode` still captures the resource — but the previous `{ String exc -> antPathMatcher.match(exc, res.text()...) }` said so plainly, and this is a readability step backwards in a refactoring PR. Keeping the explicit parameter costs nothing:

```groovy
pluginXml.resources.resource.each { res ->
    def resourceNode = res as GPathResult
    if (pluginExcludes.any { String exc -> antPathMatcher.match(exc, resourceNode.text().replace('.', '/')) }) {
        resourceNode.replaceNode {}
    }
}
```

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -212,14 +196,18 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+                'META-INF/grails.factories',
+                ['src/main/resources/META-INF/grails.factories']
+        )
     }
 
-    static LinkedHashSet<String> pendingPluginClasses = []
-    static Collection<String> pluginExcludes = []
-
-    protected static void generatePluginXml(ClassNode pluginClassNode, String pluginVersion, Set<String> transformedClasses, File pluginXmlFile) {
+    protected static void generatePluginXml(ClassNode pluginClassNode, Object pluginVersion, Set<String> transformedClasses, File pluginXmlFile) {
```
jdaugherty: `pluginVersion` widened from `String` to `Object` here and on the two `public static` methods below. Nothing in this repo calls them (only the spec does), but they are public/protected static entry points on 8.0.x, so this is a binary-incompatible signature change for any external plugin build compiled against them, for no gain — the type is immediately narrowed again by `.toString()` at five separate call sites.

The cleaner shape is to normalise once where the value originates, in `visit()`, and leave these signatures as `String`:

```groovy
String pluginVersion = null
...
pluginVersion = projectVersion?.toString()
```

That also gives you the natural place to decide what a missing version means, instead of deferring it to an unguarded `pluginVersion.toString()` in `addPluginVersionProperty` (see the `resolveProjectVersion` comment).

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -221,4 +228,160 @@ class FooGrailsPlugin {
...
+''', targetDirectory, [projectName: 'plain', projectVersion: '1.0'])
+
+        then: "no @GrailsPlugin metadata is stamped on the class"
+            classNode.getAnnotations(ClassHelper.make(GrailsPlugin)).isEmpty()
+
+        cleanup:
+            projectDir.deleteDir()
+    }
+
+    void "Test that plugin dot xml excludes are honoured and metadata refreshed when updating an existing file"() {
```
jdaugherty: This test does not exercise the path its name claims. Because `createMarkup` throws (see the main-file comment), `updatePluginXml` aborts before writing and `catch (ignored)` re-runs `writePluginXml`, which regenerates the descriptor from the AST — and since `writePluginXml` applies `pluginExcludes` to `artefactClasses` itself and reads `grailsVersion` from the same plugin properties, it produces exactly the asserted `version == '2.0'`, `grailsVersion == '3.0 > *'`, `resource*.text() == ['KeptThing']`. The assertions pass whether the update path works or is entirely broken.

The pre-existing "plugin dot xml file is updated when the plugin dot xml does exist" test has the same blind spot. To pin the update semantics, assert on something only the merge path can produce — e.g. seed the existing `plugin.xml` with a resource that is *not* in `artefactClasses` and require it to survive, or drive `updatePluginXml(null, ...)` where the rewrite fallback is a no-op.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -221,4 +228,160 @@ class FooGrailsPlugin {
...
+        SourceUnit capturedSourceUnit = null
+        cu.addPhaseOperation(new CompilationUnit.PrimaryClassNodeOperation() {
+            @Override
+            void call(SourceUnit source1, GeneratorContext context, ClassNode cn) throws CompilationFailedException {
+                capturedClassNode = cn
+                capturedSourceUnit = source1
+            }
+        }, Phases.CANONICALIZATION)
+        cu.compile(Phases.CANONICALIZATION)
+        [capturedClassNode, capturedSourceUnit]
```
jdaugherty: `capturedSourceUnit` is returned but never used — all three callers destructure it into a `sourceUnit` variable that is then dead. Dropping it lets `compileToFile` return the `ClassNode` directly and removes the awkward `source1` parameter naming in both inner classes.

Also worth adding the `pendingPluginClasses` / `pluginExcludes` reset to these three specs' `cleanup:` blocks, as the excludes test below already does. Compiling for real drives `generatePluginXml`, which mutates both static collections, and with `maxParallelForks > 1` that state outlives the feature method inside the fork.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -20,16 +20,23 @@ package org.grails.compiler.injection
...
+import org.codehaus.groovy.control.customizers.CompilationCustomizer
 import spock.lang.Specification
 import spock.util.environment.RestoreSystemProperties
 
+import grails.plugins.metadata.GrailsPlugin
+import grails.util.GrailsNameUtils
+import org.apache.grails.common.compiler.GroovyTransformOrder
+
 /**
  * Created by graemerocher on 19/09/14.
```
jdaugherty: Might as well remove this since we're removing authors too.

jdaugherty (commented):
I had AI take a pass at reviewing this before I dug in, I think it found some real issues: 

Mostly a welcome tidy-up of a long-neglected transform — hoisting the constants/fields to the top of the class, extracting the intent-revealing predicates (`shouldVisit`, `isGrailsPluginDescriptorClass`), and collapsing the duplicated property-writing into helpers all make this much easier to follow.

Two of the extractions changed runtime behaviour, though, and both are silent:

1. `createMarkup()` throws `MissingPropertyException` unconditionally — `mkp` is not a property of `StreamingMarkupBuilder`, it only exists as a pseudo-namespace inside a `bind {}` closure. That makes the whole happy path of `updatePluginXml` dead code; every call now lands in `catch (ignored)` and rewrites the descriptor from scratch instead of merging.
2. `resolveProjectVersion()` moved `getClass()` from an instance method into a `static` one, where Groovy resolves it to `java.lang.Class` rather than the transform class — so the implementation-version fallback is now always `null`.

Neither is caught by the suite: the pre-existing update test and the new excludes test both assert values that the `writePluginXml` fallback happens to reproduce, so they pass either way. Details inline.

Smaller things:

- Widening `pluginVersion` from `String` to `Object` across `generatePluginXml` / `writePluginXml` / `updatePluginXml` is binary-incompatible for anything outside this repo compiled against 8.0.x, and it pushes `.toString()` calls down into five call sites. Converting once at the top of `visit()` and keeping `String` on the signatures reads better and is the safer change.
- `writePluginXmlProperties` / `writePluginXmlResources` are helpers for `updatePluginXml`, not for `writePluginXml`, so the shared `write*` prefix reads as if they belonged to the create path. `updatePluginXmlProperties` / `updatePluginXmlResources` would remove the ambiguity.
- `resolveProjectName()` is a one-line passthrough over `getNodeMetaData('projectName')`; inlining it would be clearer than the indirection.

I ran `:grails-core:test --tests "*GlobalGrailsClassInjectorTransformationSpec"` and `:grails-core:codeStyle` on 5e9ce01 — both green, which is precisely the problem for the first two points.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -56,140 +58,158 @@ import org.grails.io.support.GrailsResourceUtils
...
+    public static final String ISOLATED_BUILD_PROPERTY = 'grails.isolated.build'
+
+    public static final ClassNode ARTEFACT_CLASS_NODE = new ClassNode(Artefact)
     public static final ClassNode ARTEFACT_HANDLER_CLASS = ClassHelper.make('grails.core.ArtefactHandler')
     public static final ClassNode TRAIT_INJECTOR_CLASS = ClassHelper.make('grails.compiler.traits.TraitInjector')
 
+    private static final AntPathMatcher ANT_PATH_MATCHER = new AntPathMatcher()
+
+    private final LinkedHashSet<String> pendingPluginClassNames = []
+    private final Collection<String> pluginExcludePatterns = []
```
jdaugherty, line 90: Good change, and I wanted to confirm the assumption it rests on rather than guess at it. Groovy's `ASTTransformationVisitor` creates **one instance of a global transform per `CompilationUnit`** and drives `visit()` on that instance for each source unit in order — I verified with a throwaway global transform registered via `META-INF/services`:

```
total instances created: 3
  instance=2 visit#1 source=SrcA.groovy
  instance=2 visit#2 source=SrcB.groovy
  instance=2 visit#3 source=SrcC.groovy
  instance=3 visit#1 source=SrcA.groovy     <- second CompilationUnit -> new instance
  instance=3 visit#2 source=SrcB.groovy
```

So instance scope == compilation scope, which is precisely the lifetime `pendingPluginClassNames` and `pluginExcludePatterns` want: carried across source units within one compile task, isolated between compile tasks. It also answers the Copilot concurrency thread — the phase operation is applied sequentially over source units, so the unsynchronised `LinkedHashSet` is fine and a `ThreadLocal`/`CompilationUnit`-keyed map would be strictly worse.

The corollary is that the `clear()` at line 286 no longer has a job to do; see that comment.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
-        pluginClasses.addAll(pendingPluginClasses)
-
-        // if the class being transformed is a *GrailsPlugin class then if it doesn't exist create it
-        if (pluginClassNode && !pluginClassNode.isAbstract()) {
+        def pluginClasses = [] as LinkedHashSet<String>
+        pluginClasses.addAll(transformedClassNames)
+        pluginClasses.addAll(pendingPluginClassNames)
+        // Reset excludes from a previous source unit so that patterns declared by one plugin
+        // do not leak into a subsequent compilation within the same Gradle worker.
+        pluginExcludePatterns.clear()
```
jdaugherty: This is a regression, and it silently turns `pluginExcludes` off for most of a plugin build.

The comment's rationale no longer applies. `pluginExcludePatterns` is now an instance field, and a global transform gets a fresh instance per `CompilationUnit` (verified — see my note on line 89), so patterns cannot leak into "a subsequent compilation within the same Gradle worker". There is nothing left for this `clear()` to guard against.

What it *does* do is discard the descriptor's excludes for every **later source unit in the same compilation**:

- SU1 = `FooGrailsPlugin.groovy` → `writePluginXml`/`updatePluginXml` records `pluginExcludePatterns`.
- SU2 = any other artefact → `generatePluginXml` clears the patterns, then `updatePluginXml(null, ...)` runs with `pluginClassNode == null`, so line 412 never repopulates them, and `handleExcludes` at line 421 is a no-op.

Excluded resources therefore survive into `META-INF/grails-plugin.xml`. That is not cosmetic: `PluginXmlHandler` collects `<resource>` into `providedClasses`, which `BinaryGrailsPlugin.initializeProvidedArtefacts` registers as the plugin's artefacts — so classes an author deliberately excluded get loaded in consuming applications.

Here is a spec that **passes on `8.0.x` and fails on this branch**:

```groovy
void "excludes recorded by the descriptor still apply to a later source unit"() {
    given: 'an existing descriptor listing a resource the plugin excludes'
        def pluginXml = new File(tempDir, 'carry-over-plugin.xml')
        def seed = '''
            <plugin name="carryOver" version="1.0" grailsVersion="1.0 > *">
                <type>CarryOverGrailsPlugin</type>
                <resources>
                    <resource>ExcludedThing</resource>
                    <resource>KeptThing</resource>
                </resources>
            </plugin>
        '''
        pluginXml.text = seed
        def classNode = compilePlugin('''
            class CarryOverGrailsPlugin {
                def pluginExcludes = ['Excluded*']
            }
        ''')

    when: 'source unit 1 is the plugin descriptor'
        transformation.generatePluginXml(classNode, '1.0', ['KeptThing'] as Set, pluginXml)

    then:
        new XmlSlurper().parse(pluginXml).resources.resource*.text() == ['KeptThing']

    when: 'source unit 2 is an ordinary artefact source'
        pluginXml.text = seed
        transformation.generatePluginXml(null, null, ['AnotherThing'] as Set, pluginXml)

    then: 'the excludes recorded in source unit 1 are still honoured'
        new XmlSlurper().parse(pluginXml).resources.resource*.text() == ['KeptThing', 'AnotherThing']
}
```

On this branch the second `then:` yields `[ExcludedThing, KeptThing, AnotherThing]`.

The fix is to delete the three lines. Worth adding the test above regardless — no existing spec drives two successive `generatePluginXml` calls, which is why the whole static-to-instance change is currently uncovered in the direction that matters.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+
+        // Create or update grails-plugin.xml when a concrete plugin class is present; otherwise,
+        // update an existing descriptor or defer resource names until the descriptor is compiled.
+        if (pluginClassNode && !pluginClassNode.abstract) {
+            if (!pluginVersion) {
+                throw new IllegalStateException(
+                        "Unable to generate '${pluginXmlFile}' because plugin class '${pluginClassNode.name}' " +
+                                'does not define a plugin version.'
+                )
+            }
```
jdaugherty: Failing fast here is the right instinct, but throwing out of `visit()` produces a diagnostic that blames Groovy for a plugin misconfiguration. Compiling `class UnversionedGrailsPlugin {}` on this branch gives the author:

```
org.codehaus.groovy.GroovyBugError: BUG! exception in phase 'canonicalization' in source unit
'.../UnversionedGrailsPlugin.groovy' Unable to generate '.../META-INF/grails-plugin.xml' because
plugin class 'UnversionedGrailsPlugin' does not define a plugin version.
```

"BUG!" is Groovy's wrapper for an unexpected `RuntimeException` escaping a phase operation; it invites a bug report against the wrong project, and it carries no source location.

`GrailsASTUtils.error(SourceUnit, ASTNode, String, boolean)` already exists for this, in this package (`GrailsASTUtils.java:181`). Routing the failure through the error collector reports it as an ordinary compilation error against the descriptor's line/column:

```groovy
GrailsASTUtils.error(source, pluginClassNode,
        "Unable to generate '${pluginXmlFile}' because plugin class '${pluginClassNode.name}' " +
        'does not define a plugin version.', true)
return
```

That needs `generatePluginXml` to take the `SourceUnit` (or the check to move up into `visit()`, which has it) — either is fine.

Same applies to the `IllegalStateException` in `resolveCompilationTargetDirectory`, though that one predates this PR.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
-                def excludes = pluginProperties.get('pluginExcludes')
-                if (excludes instanceof List) {
-                    pluginExcludes.clear()
-                    pluginExcludes.addAll(excludes)
-                }
+            writePluginXmlWithDescriptor(pluginClassNode, pluginVersion, pluginXml, artefactClassNames)
+        } else {
+            writePluginXmlWithoutDescriptor(pluginXml, artefactClassNames)
+        }
+        pendingPluginClassNames.clear()
```
jdaugherty, line 474: Dropping the old `if (pluginClassNode)` guard means `writePluginXml(null, ...)` is no longer a no-op, and that changes what happens to a corrupt descriptor. I ran `updatePluginXml(null, null, file, ['Foo'])` against `<plugin><resources>` on both branches:

- `8.0.x`: file left untouched — `<plugin><resources>`
- this branch: file replaced with `<plugin><resources><resource>Foo</resource></resources></plugin>`

It does answer the Copilot thread about `pendingPluginClassNames` never being cleared, but the replacement is a descriptor with no `name`, no `version` and no `<type>`. `PluginXmlHandler` will happily parse it and hand back `providedClasses` that belong to no plugin, so under incremental compilation — where the `*GrailsPlugin` source unit is not recompiled — the corrupt file is swapped for a well-formed but meaningless one, and the failure stops being visible.

Two alternatives that both keep the pending-list fix without inventing a nameless descriptor: delete the unusable file and fall back to the `pendingPluginClassNames.addAll(...)` deferral path, or keep writing but log at WARN that the descriptor is being written without plugin identity. Whichever you pick, the "recreates safely when the existing descriptor is malformed" test should assert the resulting *content*, not just that no exception escaped — right now it would pass for either behaviour.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+                pluginExcludePatterns.clear()
+                pluginExcludePatterns.addAll(pluginExcludes)
+            }
 
-                    for (entry in pluginProperties) {
-                        delegate."$entry.key"(entry.value)
-                    }
+            // if the plugin class doesn't define a grailsVersion, use the version of the grails-core jar
+            def grailsVersion = pluginProperties['grailsVersion'] ?:
+                    GlobalGrailsClassInjectorTransformation.package.implementationVersion + ' > *'
```
jdaugherty: Two notes on this expression, which is duplicated verbatim at line 620.

First, the `getClass()` fix has a side effect worth calling out in the PR description. On `8.0.x` this resolved against `java.lang.Class`, so a plugin that did not declare `grailsVersion` got `grailsVersion="null > *"` written into its descriptor. Now it gets the real value — `CompilePlugin.groovy:83` sets `Implementation-Version` to `grailsVersion` — so published descriptors will start carrying `grailsVersion="8.0.0 > *"`. As far as I can tell nothing reads that attribute back (`PluginXmlHandler` only collects `<type>` and `<resource>`; `BinaryGrailsPlugin` only uses `getProvidedClasses()` and the descriptor `Resource`), so the blast radius looks like zero — but it is a change in generated output that is currently untested and unmentioned.

Second, when the implementation version *is* unavailable — grails-core on the classpath as class directories, i.e. IDE and this project's own test run — the string concatenation still produces the literal `"null > *"`. I hit it in a probe:

```
<plugin name='probeTwo' version='1.0' grailsVersion='null > *'>
```

Given this PR now hard-fails on a missing *plugin* version, silently emitting `null > *` for a missing *grails* version is inconsistent. Since the expression is needed twice, extracting it is the natural place to decide:

```groovy
private static String resolveGrailsVersionRange(Map pluginProperties) {
    def declared = pluginProperties['grailsVersion']
    if (declared) return declared.toString()
    def frameworkVersion = GlobalGrailsClassInjectorTransformation.package.implementationVersion
    frameworkVersion ? "${frameworkVersion} > *" : null
}
```

and then only write the attribute when it is non-null.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+     *
+     * @param pluginClassNode the plugin descriptor class
+     * @param pluginVersion the plugin version
+     * @param pluginXml the parsed plugin descriptor
+     * @return the plugin properties extracted from the class
+     */
     @CompileDynamic
-    protected static void handleExcludes(GPathResult pluginXml) {
-        if (pluginExcludes) {
+    private static Map writePluginXmlProperties(
```
jdaugherty: Re-raising this from the last round since it got more confusing rather than less. `writePluginXmlProperties` and `writePluginXmlResources` are helpers for **`updatePluginXml`** — they mutate a parsed `GPathResult` — but this PR also introduced `writePluginXmlWithDescriptor` / `writePluginXmlWithoutDescriptor`, which really are helpers for `writePluginXml`. Four `writePluginXml*` methods, split across two unrelated code paths, is the opposite of what the rest of this refactor achieves.

`updatePluginXmlProperties` / `updatePluginXmlResources` removes the ambiguity for free.

While you're in here: these two sit at the bottom of the class, ~200 lines from their only caller at line 413, with `handleExcludes`, `createMarkup`, `resolveProjectVersion` and the whole `addXxx` family in between. Moving them next to `updatePluginXml` would make the create-path/update-path split readable at a glance.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -211,153 +231,419 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
...
+                        }
+                    }
+                }
+            }
+        }
+    }
 
+    /**
+     * Updates an existing plugin descriptor with plugin metadata and newly discovered artefact
+     * resources. If the descriptor cannot be read or written, it is recreated.
```
jdaugherty: "If the descriptor cannot be read or written, it is recreated" no longer matches the code. The catch on line 429 covers parse and I/O failures only; a `GroovyRuntimeException` from the `@CompileDynamic` GPath manipulation in `writePluginXmlProperties` now propagates and fails the build.

That narrowing is the right call — it is what would have surfaced the `mkp` defect immediately — but the contract in the doc should say so, e.g. "If the descriptor cannot be parsed or written, it is recreated; any other failure aborts compilation."

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -184,41 +169,475 @@ class FooGrailsPlugin {
...
+
+        when:
+            compileToFile(sourceFile, 'class UnversionedGrailsPlugin {}', targetDir)
+
+        then:
+            def exception = thrown(GroovyBugError)
+            with(exception) {
+                cause instanceof IllegalStateException
+                cause.message.contains('does not define a plugin version')
+            }
```
jdaugherty: This test locks in the diagnostic rather than questioning it. `GroovyBugError` is Groovy's wrapper for an unexpected `RuntimeException` escaping a phase operation, so what this asserts is "a plugin author with a misconfigured descriptor is told they hit a compiler bug" — see my comment on the `throw` site.

If the failure is routed through `GrailsASTUtils.error(...)` instead, this becomes a much better test *and* a much better error: assert `MultipleCompilationErrorsException` and that the collected message names the descriptor class, which also pins the source location the error collector attaches.

Reaching through `thrown(GroovyBugError)` to `cause.message` is also fragile against Groovy changing how it wraps phase-operation failures, which is a second reason not to depend on it.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -184,41 +169,475 @@ class FooGrailsPlugin {
...
+                getMember('name').text == GrailsNameUtils.getPropertyNameForLowerCaseHyphenSeparatedName('foowidget')
+                getMember('version').text == '2.0'
+            }
+    }
+
+    void "the global transform processes a Grails service as an artefact"() {
+        given: "a service source under grails-app"
+            def sourceFile = new File(tempDir, 'grails-app/services/FooService.groovy')
+            def targetDir = new File(tempDir, 'build/classes/groovy/main')
+            TraitInjectionUtils.@traitInjectors = []
```
jdaugherty: `TraitInjectionUtils.traitInjectors` is a `private static` field in a Java class (`TraitInjectionUtils.java:48`), so `.@traitInjectors` is reaching past the public surface into implementation state — the thing AGENTS.md rule 9 asks tests not to do. It appears three times in this spec (299, 547, 571), each with a matching `cleanup:` reset, so it is at least disciplined, but it is global mutable state being toggled from a test and it will break silently the day that field is renamed or initialised eagerly.

If real trait injection has to be suppressed for these tests to be fast/deterministic, that is worth an actual seam. Otherwise it may be simpler to let the real injectors load: the three affected features compile a trivial `FooService`/`AnnotatedService`, and I would expect the resolved injector list to be short.

Either way, please add a comment explaining *why* it is being nulled out — right now the next reader has no way to tell whether it is a performance shortcut or a correctness requirement.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -184,41 +169,475 @@ class FooGrailsPlugin {
...
+            transformation.generatePluginXml(
+                    null,
+                    null,
+                    ['NewThing'] as Set,
+                    pluginXml
+            )
+
+        then:
+            new XmlSlurper().parse(pluginXml).resources.resource*.text() == ['ExistingThing', 'NewThing']
+    }
```
jdaugherty, line 504: This is the closest the spec gets to the "descriptor already compiled, ordinary artefact compiled later" flow, but it drives `generatePluginXml` exactly once, on a fresh `transformation`, so no state is ever carried from one source unit to the next.

That gap is why the `pluginExcludePatterns.clear()` regression is invisible: every feature here is a single-source-unit scenario, and `pluginExcludes` only breaks on the *second* call. The `pendingPluginClassNames` direction is covered ("artefact class names are deferred..." at line 474), so it is specifically the excludes carry-over that has no test.

Adding the two-call spec from my comment on `GlobalGrailsClassInjectorTransformation.groovy:286` closes it, and would also stand as the regression test for the static-to-instance change generally — right now nothing asserts that state *does* persist across `visit()` calls, only that it exists.

grails-core/src/test/groovy/grails/plugins/DefaultGrailsPluginManagerSpec.groovy
```diff
@@ -87,83 +90,93 @@ class DefaultGrailsPluginManagerSpec extends Specification {
...
+            getGrailsPlugin('betaProbe') != null
+        }
 
         and: 'the per-plugin loaded-successfully messages are not emitted at INFO'
-        captured.toString().readLines()
-                .findAll { it.contains('loaded successfully') }
-                .every { !it.contains('INFO') }
+        logCapture.events
+                .findAll { it.formattedMessage.contains('loaded successfully') }
+                .every { it.level.toString() != 'INFO' }
```
jdaugherty: Comparing `it.level.toString()` against a string literal gives up the type safety that moving to logback just bought. `ILoggingEvent.getLevel()` returns `ch.qos.logback.classic.Level`, and `LogCapture`'s own javadoc example and `GlobalGrailsClassInjectorTransformationSpec:466` both use the enum:

```groovy
.every { it.level != Level.INFO }
```

Same at line 161 and at `PluginDiscoverySpec:375`. Worth making consistent across all three specs in this PR.

Separately: `findAll { ... }.every { ... }` is vacuously true if nothing matched, so this assertion cannot distinguish "the messages are at DEBUG" from "the messages were never logged". Since `LogCapture` defaults to `TRACE`, the DEBUG events genuinely are captured now — which means this is finally checkable, unlike the old slf4j-simple version. Worth spending one line to say so:

```groovy
def loadedMessages = logCapture.events.findAll { it.formattedMessage.contains('loaded successfully') }
loadedMessages.size() == 2
loadedMessages.every { it.level == Level.DEBUG }
```

grails-core/src/test/groovy/org/grails/exception/reporting/StackTraceFiltererSpec.groovy
```diff
@@ -64,55 +65,48 @@ class StackTraceFiltererSpec extends Specification {
...
+                ['java.lang.reflect.Method', 'invoke', 'Method.java', 580]
             ] as StackTraceElement[]
 
         when: 'the exception is filtered'
             quietFilterer.filter(exception)
 
         then: "no 'Full Stack Trace:' entry is emitted by the filterer"
-            System.err.flush()
-            !baos.toString().contains('Full Stack Trace:')
+            logCapture.events.every { !it.formattedMessage.contains('Full Stack Trace:') }
```
jdaugherty: `every { !... }` on an empty list is `true`, so this passes whether the filterer stayed quiet or the logger was never wired up at all. Same at line 210.

The old `!baos.toString().contains(...)` had the same hole, so this is not a regression — but the migration is the moment to close it, since the positive counterpart at line 82/180 already proves events *are* captured for this logger. Either assert emptiness directly:

```groovy
logCapture.events.empty
```

or, if other events on the `StackTrace` logger are expected, `logCapture.events.count { it.formattedMessage.contains('Full Stack Trace:') } == 0`, which mirrors the `== 2` assertion at line 180 and reads as a deliberate zero rather than an accident.

grails-core/src/test/groovy/org/grails/testing/support/LogCapture.groovy
```diff
@@ -0,0 +1,86 @@
...
+ *     def logCapture = new LogCapture('org.some.Logger')
+ *
+ * then:
+ *     logCapture.events.any { it.level == Level.WARN }
+ *
+ * cleanup:
+ *     logCapture.stop()
+ * </pre>
+ */
+class LogCapture {
```
jdaugherty: Nice extraction — asserting on structured `ILoggingEvent`s instead of scraping `System.err` is a real improvement, and restoring the previous level in `stop()` is the part people usually forget.

Two things, given the commit message calls this "reusable":

1. It lives in `src/test`, so it is reusable only within `grails-core`. If the intent is for other modules to adopt it, `java-test-fixtures` + `testFixturesApi 'ch.qos.logback:logback-classic'` would make it actually consumable. If the intent is grails-core-only for now, the commit message is overselling it and the class javadoc should say it is module-local.
2. `stop()` is not idempotent and there is no `Closeable`/`AutoCloseable`, so every caller needs a `cleanup:` block and a forgotten one leaks an appender plus a mutated level into every subsequent test in the fork. Implementing `AutoCloseable` (with `close()` delegating to `stop()`) would at least let callers use `withCloseable`, and makes the leak-on-forgetting harder.

Minor: `LogCapture(Class loggerClass, ...)` is raw — `Class<?>` avoids the rawtypes warning.

grails-core/build.gradle
```diff
@@ -95,21 +95,20 @@ dependencies {
     api 'org.slf4j:jcl-over-slf4j'
 
     // Testing
-    testImplementation 'org.slf4j:slf4j-simple'
+    testImplementation 'ch.qos.logback:logback-classic' // Used for checking log output in tests
```
jdaugherty: This makes `grails-core` the only module in the repo on logback for tests — 18 others still declare `testImplementation 'org.slf4j:slf4j-simple'`. That is justified here (logback is what makes `ListAppender` assertions possible), but it is worth a word in the PR description so the next person adding a module does not copy the wrong pattern, and so the eventual "standardise on one test logging backend" decision is a conscious one rather than drift.

There is no `logback-test.xml` under `grails-core/src/test/resources`, so logback falls back to its default configuration (root at DEBUG, console appender). I measured the actual impact on `:grails-core:test` and it is small — 2 DEBUG lines across 62 test classes — so this is not urgent. But it is unpinned: the noise level of the whole module's test output now depends on whatever the classes under test happen to log at DEBUG. A three-line `logback-test.xml` with root at `WARN` makes it deterministic, and `LogCapture` still works because it raises the level on the specific logger it captures.

jdaugherty (commented):
I let AI review this again given it's critical nature to Grails Apps, review follows:

Re-reviewed at c3b60c8 (last pass was 5e9ce01). All three defects from the previous round are fixed, and I verified each rather than taking the diff's word for it:

- `createMarkup` now uses `bindNode(node)`. I rendered the same `GPathResult` through `bindNode(node)` and through `bind { mkp.yield(node) }` under groovy-xml 5.0.7 — byte-identical output. The merge path in `updatePluginXml` is live again.
- `resolveProjectVersion` and both `grailsVersion` fallbacks are anchored to `GlobalGrailsClassInjectorTransformation.package` instead of `getClass()`, so the implementation-version fallback reads the grails-core jar again.
- `catch (ignored)` is now `catch (IOException | ParserConfigurationException | SAXException e)` with a `log.warn` and a test. That narrowing is what makes the rest of this file reviewable at all.

Moving `pendingPluginClassNames` / `pluginExcludePatterns` from `static` to instance fields is the right fix, and better than the `ThreadLocal` / `CompilationUnit`-keyed suggestion in the Copilot thread. I confirmed the lifecycle assumption with a probe global transform: Groovy instantiates a global transform **once per `CompilationUnit`** and calls `visit()` on that one instance, sequentially, per source unit:

```
total instances created: 3
  instance=2 visit#1 source=SrcA.groovy
  instance=2 visit#2 source=SrcB.groovy
  instance=2 visit#3 source=SrcC.groovy
  instance=3 visit#1 source=SrcA.groovy     <- second CompilationUnit, new instance
  instance=3 visit#2 source=SrcB.groovy
```

That is exactly the scope this state wants, and it also means plain `LinkedHashSet` is safe here.

Two things to resolve before merge:

1. **`pluginExcludePatterns.clear()` at line 286 is a regression** — it makes `pluginExcludes` a no-op for every source unit compiled after the descriptor. I have a test that passes on `8.0.x` and fails on this branch; details inline. This is user-visible: `<resource>` entries feed `BinaryGrailsPlugin.initializeProvidedArtefacts`, so excluded classes get registered as provided artefacts in consuming applications.
2. **The new `IllegalStateException` reaches plugin authors as `BUG! exception in phase 'canonicalization'`** — it reads as a Groovy bug rather than a plugin misconfiguration. `GrailsASTUtils.error(...)` already exists in the same package for this.

Everything else below is smaller. Ran `:grails-core:test` (green) and `:grails-core:codeStyle` (clean) at c3b60c8.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
matrei: Fixed the `IllegalStateException` added by this PR but leaves the existing one for another day.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
matrei: Went the "delete the unusable file" path.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
matrei: Now fall backs to no `grailsVersion` attribute when it cannot be resolved.

jdaugherty (approved):
Incredible work @matrei Looks like everything has been addressed, so if the tests pass, I'm good to merge.

PR #16047, opened 2026-07-23
Automate configuration metadata for @ConfigurationProperties modules

### LINKED ISSUE #15469: Migrate configuration metadata to @ConfigurationProperties with annotation processor for Grails 8

jamesfredley:
## What this does

Closes #15469 for the five modules that already use `@ConfigurationProperties`:

- `grails-cache`
- `grails-databinding`
- `grails-views-gson`
- `grails-views-markup`
- `grails-web-url-mappings` (CORS owner)

Each publishes standard Spring Boot metadata at `META-INF/spring-configuration-metadata.json` during the module build. IDEs, `config-report`, and the Application Properties reference consume that resource.

This is **not** Spring Boot's `spring-boot-configuration-processor`. That approach was tried in #15566 and closed because Java annotation processors break incremental Groovy compilation. This PR uses a Grails-owned path instead:

1. **Groovy:** a semantic-analysis AST transform embeds a private synthetic metadata payload on each compiled `@ConfigurationProperties` class (no shared processor output).
2. **Java:** ASM reads compiled bytecode without classloading.
3. **Build:** a cacheable Gradle task merges class payloads after compilation, applies curated overlays, and writes one standard metadata file into resources/jars.

## Migration details

| Module | Change |
| --- | --- |
| Cache / Data Binding / JSON Views | Hand-written `spring-configuration-metadata.json` renamed to `additional-spring-configuration-metadata.json` (overlay). Overlay values win for matching identities. |
| Markup Views | No prior metadata file on `8.0.x`; generation now publishes bindable Markup Views properties. |
| CORS | Moved from `grails-web-core` into `grails-web-url-mappings` with `GrailsCorsConfiguration`. Non-CORS web-core metadata stays hand-maintained in `grails-web-core`. |

Generated defaults only include compile-time constants. Dynamic Groovy defaults are omitted unless an overlay supplies an authoritative value.

## Compatibility check (against `origin/8.0.x`)

Re-verified on exact PR HEAD after regenerating metadata:

| Source | Result |
| --- | --- |
| Cache | 1 group / 6 properties preserved, no changes |
| Data Binding | 1 group / 5 properties preserved, no changes |
| JSON Views | 1 group / 9 properties preserved, no changes; **+9** inferred template/base properties |
| CORS | 1 group / 9 properties preserved and relocated to URL Mappings, no changes |
| Web-core non-CORS | 10 groups / 24 properties still present in `grails-web-core` |
| Markup Views | New generated group/properties (no prior `8.0.x` file) |

Each migrated jar contains exactly one standard metadata resource plus any curated additional metadata file.

## Scope notes

- Completes issue Phase 1 for the five existing `@ConfigurationProperties` modules.
- Does **not** migrate Groovy-DSL-only config (e.g. Spring Security `DefaultSecurityConfig`, general `application.groovy` merging). Those remain later work.
- Immutable constructor-bound properties are supported when Boot-compatible constructor selection applies; curated overlays remain the fallback for non-inferable metadata.

## Verification

Local:

- `:grails-configuration-metadata:test` + `codeStyle`
- full `ConfigurationMetadataPluginSpec` (clean, incremental, edit, delete, overlay, duplicate, immutable, generic-constructor safety)
- tests for Cache, Data Binding, JSON Views, Markup Views, URL Mappings, and affected Web Core
- `ConfigReportCommandSpec`
- clean Checkstyle / CodeNarc / PMD / SpotBugs aggregates
- semantic zero-loss comparison + jar resource inspection

CI on this PR: core builds (Linux/macOS), style/analysis/RAT/CodeQL/coverage, Forge, functional, security, Redis, MongoDB, and Hibernate suites are green. A few long jobs (Windows core, joint Groovy validation, selected functional reruns) were still finishing at description update time.

## Commits

1. Compiler module + AST transform
2. Build-logic Gradle plugin + TestKit
3-7. Per-module migrations (cache, databinding, gson, markup, URL mappings/CORS)
8-9. Docs reference inputs + guide prose
10. Immutable constructor-bound metadata support

## Stacked follow-up

[#16048](https://github.com/apache/grails-core/pull/16048) is stacked on this PR and should be reviewed and merged afterward. It handles Groovy DSL parsing and the remaining in-repository metadata migrations tracked by [#15469](https://github.com/apache/grails-core/issues/15469). This PR remains limited to the five existing `@ConfigurationProperties` modules and the shared generator correctness fixes required by its review; #16048 does not replace those fixes.


jdaugherty:
We had discussed creating a custom solution and not requiring ConfigurationProperties in the meeting, why the reversal here now?

grails-doc/build.gradle
```diff
@@ -292,7 +292,9 @@ generateConfigReference.configure { Task it ->
             project(':grails-cache').tasks.named('jar'),
             project(':grails-data-hibernate5-dbmigration').tasks.named('jar'),
             project(':grails-data-mongodb').tasks.named('jar'),
-            project(':grails-views-gson').tasks.named('jar')
+            project(':grails-views-gson').tasks.named('jar'),
+            project(':grails-views-markup').tasks.named('jar'),
```
jdaugherty, line 346: Wiring these two jars into `generateConfigReference` publishes the generated metadata straight into the Application Properties reference, and `grails-views-markup` has no curated overlay. Generating the reference from this branch:

- 33 of 187 rows now have an empty Description **and** an empty Default: all 24 `grails.views.markup.*`, plus the 9 newly inferred `grails.views.json.*` entries (`baseTemplateClass`, `cache`, `enableReloading`, `extension`, `packageImports`, `packageName`, `staticImports`, `templatePath`, `useAbsoluteLinks`). Every pre-existing row in that file has a description, so all 33 blanks are new.
- The markup rows land inside the existing "Views & GSP" section, interleaved with fully documented `grails.views.gsp.*` rows, so they read as holes in an otherwise complete table.

Every one of these is settable, so every one needs documenting. Can we add `grails-views-markup/src/main/resources/META-INF/additional-spring-configuration-metadata.json` with a group description plus a description and default for each of the 24, and descriptions for the 9 new JSON Views entries, before these jars feed the reference?

grails-configuration-metadata/src/main/groovy/org/apache/grails/configuration/metadata/ConfigurationMetadataTransformation.groovy
```diff
@@ -0,0 +1,277 @@
...
+import static java.lang.reflect.Modifier.PRIVATE
+import static java.lang.reflect.Modifier.FINAL
+import static java.lang.reflect.Modifier.STATIC
+
+/**
+ * Embeds configuration metadata in each annotated Groovy class. Aggregation is deliberately
+ * deferred to the Gradle task so incremental Groovy compilation never writes shared output.
+ */
+@CompileStatic
+@GroovyASTTransformation(phase = CompilePhase.SEMANTIC_ANALYSIS)
```
jdaugherty, line 51: `SEMANTIC_ANALYSIS` runs before `@Delegate` composes (`DelegateASTTransformation` is `CANONICALIZATION`), so delegate-generated setters are invisible to `collectSetterProperties`. CORS is exactly that case: running `:grails-web-url-mappings:generateConfigurationMetadata` on this branch, the generated half contributes only `grails.cors.enabled` and `grails.cors.mappings`. `allowedOrigins`, `allowedMethods`, `allowedHeaders`, `exposedHeaders`, `maxAge` and `allowCredentials` — all bindable through the `@Delegate GrailsDefaultCorsConfiguration` and Spring's `CorsConfiguration` setters — survive only because the new overlay lists them by hand.

That is a reasonable limitation to accept, but today it fails silently: add a delegated property and the metadata simply will not have it, with nothing in the build to say so. Can we state the limitation in the class documentation, and ideally have the transform warn (or the task fail) when a `@ConfigurationProperties` class carries a `@Delegate` field, so whoever adds one is told to extend the overlay?

grails-configuration-metadata/src/main/groovy/org/apache/grails/configuration/metadata/ConfigurationMetadataTransformation.groovy
```diff
@@ -0,0 +1,277 @@
...
+            String propertyName = decapitalize(method.name.substring(3))
+            ClassNode propertyType = method.parameters[0].type
+            bindable.putIfAbsent(propertyName,
+                    [propertyName: propertyName, type: typeName(propertyType), classNode: propertyType, nested: true])
+        }
+        collectSetterProperties(node.superClass, bindable, visited)
+        node.interfaces.each { ClassNode interfaceNode -> collectSetterProperties(interfaceNode, bindable, visited) }
+    }
+
+    private static boolean isNested(ClassNode type) {
```
jdaugherty: `isNested` treats every type outside `java.*`/`groovy.*` as a nested configuration group, and `collectSetterProperties` walks superclasses and interfaces without applying that filter at all — it only stops at `Object`. The two hardcoded names on line 174 are what keeps that from exploding today: `JsonViewConfiguration` and `MarkupViewConfiguration` implement the `GenericViewConfiguration` trait, which extends `GrailsApplicationAware`, so without the `setGrailsApplication` exclusion the transform would recurse into `GrailsApplication`.

Boot's own processor keys nesting off inner classes and `@NestedConfigurationProperty` rather than a package-name heuristic, and the ASM path in this PR is already narrower — it nests only into types found in the same module's output (`models[property.rawType()]`). Could the Groovy path use the same restriction, and honour `@NestedConfigurationProperty`? As written, the next config class with a setter taking a framework or third-party bean type silently emits a nested group tree, and the remedy will be another name added to that blocklist.

grails-configuration-metadata/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+ *  under the License.
+ */
+
+plugins {
+    id 'groovy'
+    id 'java-library'
+    id 'org.apache.grails.buildsrc.properties'
+    id 'org.apache.grails.buildsrc.dependency-validator'
+    id 'org.apache.grails.buildsrc.compile'
+    id 'org.apache.grails.buildsrc.publish'
```
jdaugherty: Applying `buildsrc.publish` (with the matching `publish-root-config.gradle` entry) puts this module on Maven Central, and because `grails-bom/base` adds every published subproject as a constraint it lands in `grails-bom` too. It is also the only new module that deliberately skips `gradle/docs-config.gradle`, so it ships as a published artifact with no API docs. It is `compileOnly` on the five framework modules, so it never reaches their POMs, and nothing outside this build can use it without the unpublished `build-logic` plugin.

It also registers a **global** transform via `META-INF/services/org.codehaus.groovy.transform.ASTTransformation`. Once published, anyone who ends up with it on a compile classpath silently gets `__grailsConfigurationMetadata` injected into their own `@ConfigurationProperties` classes, and a hard compile error if they happen to declare a field with that name. Is there a consumer that needs the artifact? If not, dropping the publish/sbom/vulnerability-scan plugins and the `publishedProjects` entry keeps it internal to the build.

Separately, `compileOnly 'org.springframework.boot:spring-boot'` on line 39 looks unused — the transform matches `ConfigurationProperties` and `ConstructorBinding` by name string and imports nothing from Spring.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
```diff
@@ -0,0 +1,557 @@
...
+        classesDirs.files.findAll { File file -> file.isDirectory() }.sort { File file -> file.absolutePath }.each {
+            File directory ->
+            Stream<java.nio.file.Path> paths = Files.walk(directory.toPath())
+            try {
+                paths.filter { java.nio.file.Path path -> Files.isRegularFile(path) && path.fileName.toString().endsWith('.class') }
+                        .sorted()
+                        .forEach { java.nio.file.Path path ->
+                            ClassModel model = GenerateConfigurationMetadataTask.readClass(Files.readAllBytes(path))
+                            ClassModel previous = models.put(model.name, model)
+                            if (previous != null && previous != model) {
```
jdaugherty: `ClassModel` does not implement `equals`, so `previous != model` is identity comparison and is always true for two distinct instances. This reduces to "fail on any duplicate class name across the input directories", which is not what the guard reads as. Either give `ClassModel` value equality (or compare prefix/payload) so a genuinely identical duplicate is tolerated as intended, or drop the second condition so the code says what it does.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
```diff
@@ -0,0 +1,557 @@
...
+            if (merged[name] instanceof Map && value instanceof Map) {
+                merged[name] = new LinkedHashMap((Map) merged[name]) + (Map) value
+            } else {
+                merged[name] = value
+            }
+        }
+        merged.keySet().sort().collect { String name -> merged[name] }
+    }
+
+    private static Map<String, Object> indexByName(List source, String category, String sourceName) {
```
jdaugherty, line 392: `indexByName` enforces name uniqueness per category. That is right for `properties`, but Boot's format allows repeated group names distinguished by `sourceType`/`sourceMethod`, and `generate()` emits one group per `@ConfigurationProperties` class (line 128) plus one per nested holder. Two config classes sharing a prefix in the same module, or a nested prefix reachable from two classes, would fail the build with `Conflicting generated groups metadata for '...'` on metadata that is legal. Worth keying groups on name + `sourceType`, or merging instead of rejecting.

grails-web-core/src/main/resources/META-INF/spring-configuration-metadata.json
```diff
@@ -20,10 +20,6 @@
             "name": "grails.logging",
             "description": "Web & Controllers"
         },
-        {
-            "name": "grails.cors",
-            "description": "CORS"
-        },
         {
             "name": "grails.mime",
```
jdaugherty, line 24: With CORS moved out, this file still carries 10 groups and 24 properties for prefixes spanning URL mappings, content negotiation, static resources and scaffolding. Not a blocker for this PR, but is the intent to keep relocating those to the modules that own them as they gain `@ConfigurationProperties` classes? Worth capturing in #15469 so this file does not stay the central fallback by default.

grails-configuration-metadata/src/test/groovy/org/apache/grails/configuration/metadata/ConfigurationMetadataTransformationSpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+package org.apache.grails.configuration.metadata
+
+import groovy.json.JsonSlurper
+import org.codehaus.groovy.control.MultipleCompilationErrorsException
+import spock.lang.Specification
+
+import java.lang.reflect.Field
+import java.lang.reflect.Modifier
+
+class ConfigurationMetadataTransformationSpec extends Specification {
```
jdaugherty, line 42: Two feature methods for a 277-line transform leaves most of the logic that decides what gets published unexercised. The TestKit spec covers the Java/ASM half thoroughly, but the Groovy half is only tested through one happy-path class, so the branches below can change behaviour without any test noticing:

- `@Delegate` field exclusion, and the `setGrailsApplication`/`setMetaClass` exclusions.
- `collectSetterProperties` walking superclasses and interfaces — a config class extending a base class and implementing a trait is the shape both view configurations actually have, and it is the source of most generated properties.
- `constructorBoundProperties` selection: `@ConstructorBinding` on one of several constructors, the "no no-arg constructor plus exactly one candidate" rule, and private/synthetic constructors being ignored.
- A non-constant `prefix` expression, which makes `addPayload` bail out entirely and silently hands the class to the bytecode bean-scan path instead — worth pinning, since the two paths do not infer the same set.
- `typeName`/`genericTypeName` for arrays, wildcards (`? extends` / `? super`) and type variables.
- The hand-rolled `toJson`/`escapeJson`. A constant default containing a quote, backslash, newline or control character is written into a class-file string constant and parsed back by `JsonSlurper` in the Gradle task; if the escaping is ever wrong the failure surfaces as a JSON parse error during someone else's module build.
- An interface annotated `@ConfigurationProperties` (skipped by `!node.interface`), and a prefix-less `@ConfigurationProperties`.

Each is a few lines of `parseClass` in this spec, and they are what makes the mechanism safe to change later.

jdaugherty (commented):
I had AI review this and then iterated with the review. I think the `@Delegate` finding is rather important.  Review follows.

Read through the compiler module, the Gradle plugin and the five module migrations, then ran `generateConfigurationMetadata` for all five modules plus `:grails-doc:generateConfigReference` on this branch to diff the actual output. The mechanism works: the ASM path and the Groovy payload path both produce correct metadata for Java and Groovy config classes, and cache / data binding / JSON views / CORS come through with no loss. The TestKit spec covers clean, incremental, edit, delete and overlay behaviour well.

One thing I would want resolved before this goes in: **the Application Properties reference regresses.** 33 of 187 rows come out with an empty description *and* an empty default (24 `grails.views.markup.*`, 9 newly inferred `grails.views.json.*`). Every other row in that file has a description, so this PR introduces all 33.

The remaining comments are about how quietly the generated half can drift from what Boot actually binds (the `@Delegate` blind spot and the `isNested` heuristic), how thinly the transform itself is covered, and a question about publishing the compiler module.

This review is on the implementation as written; it does not address the separate open question about whether the approach should require `@ConfigurationProperties`.

grails-configuration-metadata/src/test/groovy/org/apache/grails/configuration/metadata/ConfigurationMetadataTransformationSpec.groovy
jamesfredley: Addressed in 7c10347322. `ConfigurationMetadataTransformationSpec` now covers every requested branch: delegated/framework setter exclusions; superclass, interface, and trait setter discovery; annotated, fallback, private, and pre-semantic synthetic constructor selection; non-`ConstantExpression` prefix deferral; arrays, bounded/unbounded wildcards, and type variables; JSON escaping for quotes, backslashes, newlines, and control characters; annotated interface skipping; and prefix-less properties.

The synthetic-constructor fixture uses a CONVERSION-phase compilation customizer, so it would fail if the production synthetic filter were removed. All nine spec features pass, the full `grails-configuration-metadata` module check and dependency validation pass, and the dual Oracle/Codex review is green.

jamesfredley: Follow-up: merged the current `8.0.x` base in 74587e2906. The expanded transformation coverage remains intact, and all nine `ConfigurationMetadataTransformationSpec` features pass on the merged branch.

matrei:
## Review

Thanks for this, generating the configuration metadata instead of hand-maintaining it is a good direction.

I built `generateConfigurationMetadata` for `grails-databinding`, `grails-views-gson`, `grails-views-markup` and `grails-web-url-mappings` and inspected the generated JSON (I did not build `grails-cache`). The metadata shipped for the converted modules looks correct today, because the curated overlays cover every generated entry. The points below matter as soon as other modules adopt the plugin without a complete overlay, so I think the first three should be fixed before merge, along with the failing test in point 10.

### `ConfigurationMetadataPlugin.groovy` (bytecode scan path)

**1. Constructor parameter types are misaligned (around line 243)**

The parameter index counts synthetic and mandated parameters, but `argumentTypeNames` comes from the generic `Signature` attribute, which omits them. For a Java non-static inner class with a generic constructor parameter, e.g. `Inner(List<String> names, int size)`, `MethodParameters` reports `[this$0 (mandated), names, size]` while the signature list is `[List<String>, int]`. `names` gets type `int` and `size` gets `null`. The null then reaches `PropertyModel.rawType()` (`type.replaceFirst`) and fails the task with an NPE. When no NPE occurs, wrong types are published silently.

**2. The bytecode fallback publishes a bogus `<prefix>.metaClass` property (around line 256)**

When a Groovy class uses a non-literal prefix, e.g. `@ConfigurationProperties(PREFIX)` as in `grails-mail`'s `MailConfigurationProperties`, the AST transform returns without a payload and the task falls back to scanning bytecode. That path does not exclude `getMetaClass`/`setMetaClass` (`javap` confirms they are plain `ACC_PUBLIC`, not synthetic), nor `setGrailsApplication`. The AST path filters both. The result would be `grails.mail.metaClass` of type `groovy.lang.MetaClass` in the published metadata.

**3. Getter-only nested objects are dropped (around line 298)**

The per-class filter keeps only writable, collection/map or constructor-bound properties, so this standard Spring Boot idiom is removed before nested-group resolution in `addProperties` can see it:

```java
private final Pool pool = new Pool();
public Pool getPool() { return pool; }
```

Neither the `prefix.pool` group nor any `prefix.pool.*` property is generated, and there is no warning. Spring binds these at runtime, and the Groovy AST path does handle getter-only nested properties, so the two paths disagree.

**4. The reported property type depends on method order (around line 307)**

`addAccessor` overwrites the property type whenever a later accessor has a generic signature. With `setPackageImports(String[])` and `setPackageImports(List<String>)`, or a getter returning `Collection<String>` with a setter taking `List<String>`, the last generic-bearing method visited wins. The published type may not match the binding setter and can change when methods are reordered. A non-generic overload visited later never corrects a generic type set earlier.

### `ConfigurationMetadataTransformation.groovy` (AST path)

**5. Primitive types are emitted unboxed (around line 268)**

`boolean enabled = true` with no overlay entry produces `"type": "boolean"`. Spring Boot's own processor and every curated overlay entry in this repo use the wrapper (`java.lang.Boolean`). The bytecode path's `fieldType` does the same. This is currently hidden only because the overlays override the type for every shipped property.

**6. The JSON payload ships in the runtime class (around line 92)**

The payload is stored as a `static final String` constant, so every `@ConfigurationProperties` class in the released jars carries the `__grailsConfigurationMetadata` constant (confirmed with `javap`). It is build-time-only data living permanently in the runtime constant pool. It is also limited to the 65535-byte class-file constant size, so a large config tree with deep `@NestedConfigurationProperty` expansion inlined into the root payload would fail compilation with a cryptic "UTF8 string too long" error.

A side-car file per class, or stripping the field after aggregation, would avoid both problems. This is a design decision that is much cheaper to settle now than after a release.

**7. The `@Delegate` warning cannot be suppressed (around line 161)**

It fires on every compilation, even when the overlay already documents the delegated properties. `GrailsCorsConfiguration` has `@Delegate GrailsDefaultCorsConfiguration` and its overlay already supplies `allowedOrigins`, `maxAge` and the rest, yet every build of `grails-web-url-mappings` would emit a warning asking for metadata that already exists. That is permanent noise, and it could fail a build that treats warnings as errors. The aggregation task can see the overlay and could check coverage there; the transform cannot.

### Tests

**8. `JsonViewGrailsPluginSpec.groovy:73` reads a private synthetic field reflectively**

`JsonViewConfiguration.getDeclaredField('__grailsConfigurationMetadata')` with `accessible = true` ties the test to an internal compiler artefact, which goes against the "test via public APIs" rule in the agent guide. The user-visible contract is the jar's `META-INF/spring-configuration-metadata.json`, and that file is not checked. The test also breaks if the payload mechanism changes (see point 6).

**9. `ConfigurationMetadataPluginSpec.groovy:46` hardcodes fixture versions**

The TestKit fixtures pull `org.apache.groovy:groovy:5.0.7` and `spring-boot:4.1.0` from Maven Central instead of the versions in `dependencies.gradle`. The transform is therefore tested against Groovy 5.0.x rather than the 5.1.x it ships with, so a difference in AST or phase behaviour (timing of `@Delegate`, trait composition) would go unnoticed. The tests also need network access and will drift as the BOM is bumped. Separately, the file is ~600 lines of inline fixture strings repeating the same build script, which could be factored out.

**10. `ConfigurationMetadataPluginSpec` "generates canonical metadata across clean incremental edit and source deletion builds" fails on this branch**

Running `cd build-logic && ./gradlew :build-logic:test --tests '*ConfigurationMetadataPluginSpec'` on the unmodified branch gives:

```
metadata.groups*.name == ['fixture.groovy', 'fixture.groovy.nested', 'fixture.java', 'fixture.java.nested', 'overlay.only']
                         [fixture.groovy, fixture.java, fixture.java.nested, overlay.only]
```

The fixture declares `GroovyNested` as a top-level class and expects a `fixture.groovy.nested` group. The AST transform deliberately treats a property as nested only when its type is an inner class of the configuration class or it carries `@NestedConfigurationProperty` (that rule has its own test in `ConfigurationMetadataTransformationSpec`, "only recurses into inner and explicitly nested configuration properties"). So the fixture contradicts the transform, not the other way round. Making `GroovyNested` a static inner class of `GroovyConfiguration`, and expecting the type `fixture.GroovyConfiguration$GroovyNested`, makes the test pass.

Related: the same fixture treats any Java type compiled by the module as nested (`JavaNested` is top-level and becomes a group), so the Java and Groovy paths currently apply different nesting rules. It would be good to decide which one is intended.

### Summary

- **Should fix before merge:** 1, 2, 3, 10 (a red test)
- **Worth deciding now:** 6
- **Can follow up:** 4, 5, 7, 8, 9

Points 2, 5 and 6 were confirmed against compiled output or the generated JSON, and point 10 by running the test. The rest come from reading the code, so please push back if I have misread something.

matrei (approved):

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
```diff
@@ -0,0 +1,760 @@
...
+     * Payloads are honoured only for a compiled class that still carries the annotation, so a payload left
+     * behind by a deleted or no longer annotated source never leaks into the metadata.
+     */
+    private void readPayloads(Map<String, ClassModel> models) {
+        classesDirs.files.collect { File directory -> new File(directory, PAYLOAD_DIRECTORY) }
+                .findAll { File directory -> directory.isDirectory() }
+                .sort { File directory -> directory.absolutePath }.each { File directory ->
+            directory.listFiles().findAll { File file -> file.isFile() && file.name.endsWith('.json') }
+                    .sort { File file -> file.name }.each { File file ->
+                ClassModel model = models[file.name - '.json']
```
sbglasius: **The payload lookup strips the first `.json`, not the extension.**

Groovy's `String.minus(String)` removes the *first* occurrence, so for `grails.plugin.json.view.JsonViewConfiguration.json` this evaluates to `grails.plugin.view.JsonViewConfiguration` — the `.json` package segment is eaten. `models[...]` is then `null` and `applyPayload` is never called, so the payload is silently discarded.

This is happening right now on this branch: the payload file is written correctly to `grails-views-gson/build/classes/groovy/main/META-INF/grails-configuration-metadata/grails.plugin.json.view.JsonViewConfiguration.json`, but it never reaches the model. The gson output only looks right because the ASM fallback happens to cover that class — I confirmed the generated file is byte-identical after fixing the lookup.

What is being skipped for any such class:
- the payload's compile-time `defaultValue`s,
- the `@NestedConfigurationProperty` this PR adds to `JsonViewConfiguration.generator` (the nesting currently comes from the ASM heuristic instead, so the annotation is untested here),
- all `@Delegate` handling — `addDelegatedProperties` and the "not entirely compiled by this project" warning never run.

There is also a correctness hazard rather than just a gap: with both `a.b.json.C` and `a.b.C` compiled in one module, the payload of the former is applied to the model of the latter, overwriting its prefix, groups and properties.

Suggested fix:

```groovy
ClassModel model = models[file.name.substring(0, file.name.length() - '.json'.length())]
```

Worth adding a `ConfigurationMetadataPluginSpec` fixture in a package containing a `json` segment — none of the current cases would catch this.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
```diff
@@ -0,0 +1,760 @@
...
+class ConfigurationMetadataPlugin implements Plugin<Project> {
+
+    @Override
+    void apply(Project project) {
+        project.plugins.withId('java') {
+            JavaPluginExtension java = project.extensions.getByType(JavaPluginExtension)
+            project.tasks.withType(JavaCompile).configureEach { JavaCompile task ->
+                if (!task.options.compilerArgs.contains('-parameters')) {
+                    task.options.compilerArgs.add('-parameters')
+                }
```
sbglasius, line 70: **`-parameters` never reaches the sources that need it, and is redundant where it does.**

All five migrated modules keep their `.java` sources under `src/main/groovy` (none of them has a `src/main/java`), so those files are joint-compiled by `GroovyCompile`, not `JavaCompile`. `groovyOptions.parameters = true` only covers Groovy sources; the javac side of joint compilation takes its flags from `GroovyCompile.options.compilerArgs`.

Measured on this branch:

```
$ javap -v .../groovy/main/org/grails/plugins/databinding/DataBindingConfigurationProperties.class | grep -c MethodParameters
0
$ javap -v .../groovy/main/grails/plugin/cache/CachePluginConfiguration.class | grep -c MethodParameters
7
```

So for a Java `@ConfigurationProperties` class in these modules, `visitParameter` never fires, `constructor.properties` stays empty, `bindingConstructor` resolves to `null`, and every constructor-bound property is silently omitted — which is exactly the "immutable constructor-bound properties are supported" case in the PR description.

Separately, this block is a no-op even where `JavaCompile` does run: `CompilePlugin.groovy:101` already adds `-parameters` to every `JavaCompile` in the build, and every one of these modules applies `org.apache.grails.buildsrc.compile`.

If constructor binding is meant to work for joint-compiled Java, this needs to configure `GroovyCompile.options.compilerArgs` instead (ideally in `CompilePlugin`, next to the existing `groovyOptions.parameters`).

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
```diff
@@ -0,0 +1,760 @@
...
+    }
+
+    private static Map<String, Object> indexByName(List source, String category, String sourceName) {
+        Map<String, Object> indexed = [:]
+        source.each { Object entry ->
+            String name = entry instanceof Map ? ((Map) entry).name as String : entry as String
+            if (!name) {
+                throw new IllegalArgumentException("${category} entry has no name")
+            }
+            if (indexed.containsKey(name) && indexed[name] != entry) {
```
sbglasius, line 568: **Two config classes sharing a prefix abort the build on the properties path.**

`mergeGroups` deliberately supports a shared prefix — there is a spec for it, `"preserves repeated group names from two configuration classes sharing a prefix"`. The properties path does not: entries carry `sourceType`, so two generated properties with the same `name` but different `sourceType` are unequal and this throws.

Concretely, if classes `A` and `B` both use prefix `grails.foo` and both expose `enabled`, `generate()` emits

```groovy
[name: 'grails.foo.enabled', type: ..., sourceType: 'A']
[name: 'grails.foo.enabled', type: ..., sourceType: 'B']
```

and the build fails with `Conflicting generated properties metadata for 'grails.foo.enabled'` — a message that reads like an authoring error in `additional-spring-configuration-metadata.json`, pointing the reader at the wrong file.

The same shape arises without a literal duplicate prefix: class `A` with prefix `grails.x` and a nested property `b`, alongside class `B` annotated `@ConfigurationProperties('grails.x.b')`.

Either de-duplicate generated properties by identity the way `mergeGroups` does (name + sourceType), or keep it fatal but say which two source types collided.

sbglasius (commented):
_This is an AI assisted review._

Reviewed the generator end to end: I built and ran `generateConfigurationMetadata` for all five migrated modules and compared the output against `8.0.x`. No group or property is lost, and each jar carries exactly one `spring-configuration-metadata.json` with no payload leakage — the zero-loss claim holds.

Three issues in the Gradle plugin, inline below. The first one is reproducible today on this branch.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
matrei: Good catch, confirmed. `String.minus` ate the `.json` package segment, so the `JsonViewConfiguration` payload was written but never applied, and the ASM fallback masked it.

Fixed by stripping only the extension:

```groovy
ClassModel model = models[file.name.replaceFirst(/\.json$/, '')]
```

Added the fixture you suggested: `"applies the payload of a Groovy class whose package contains a json segment"` compiles `fixture.json.view.SegmentConfiguration` and `fixture.view.SegmentConfiguration` in the same module and asserts that each one gets its own payload, using the compile-time `defaultValue` as the evidence since the bytecode fallback cannot produce it. That covers both the silently dropped payload and the cross-application hazard. The spec fails against the old lookup.

The generated metadata for all five migrated modules is byte-identical before and after, which matches what you saw — but gson now gets there through the payload, so the `@NestedConfigurationProperty` on `JsonViewConfiguration.generator` and the `@Delegate` handling are actually exercised.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
matrei: Confirmed, I reproduced the `0` for `DataBindingConfigurationProperties`. `groovyOptions.parameters` only covers the Groovy sources; the javac side of joint compilation reads `GroovyCompile.options.compilerArgs`.

`CompilePlugin` now adds `-parameters` there, next to the existing `groovyOptions.parameters`, as you suggested. After the change that class carries 6 `MethodParameters` entries.

I kept the block in the metadata plugin rather than removing it, and extended it to `GroovyCompile` as well. It is redundant wherever `org.apache.grails.buildsrc.compile` is applied, but the metadata plugin reads constructor binding from `MethodParameters`, so I would rather it not depend silently on another plugin being present — its own TestKit fixture applies it standalone. Both sides are guarded with a `contains` check, so the flag is never passed twice regardless of plugin order.

New coverage: `"binds the constructor of a Java class that is joint-compiled from the Groovy source directory"` (a constructor-bound `.java` under `src/main/groovy`), and a `CompilePluginSpec` case asserting the `GroovyCompile` compiler args.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/ConfigurationMetadataPlugin.groovy
matrei: Agreed, and the inconsistency with `mergeGroups` was the real problem. I went with your first option: properties are now identified the same way groups are. `mergeGroups` became `mergeSourced` and handles both, keyed by name + `sourceType` + `sourceMethod`. The metadata format explicitly allows repeated names, so both entries are kept. Overlay semantics carry over unchanged: a name-only entry augments every generated entry of that name, a source-qualified one targets its own.

A conflict within the same identity is still fatal, and the message now names the source type:

```
Conflicting generated properties metadata for 'grails.foo.enabled' from source type 'A'
```

Both shapes you described are covered: `"preserves a property name repeated by two configuration classes sharing a prefix"` (also asserts the name-only vs. source-qualified overlay behaviour) and `"preserves a nested property that another configuration class binds by its own prefix"`. Both failed with the old code.

One note: the docs generator in `grails-doc/build.gradle` and `ConfigReportCommand` index properties by name only, so a repeated name would render as two rows. No migrated module has one today, so I left those consumers alone for this PR.

jdaugherty (approved):

jdaugherty:
Once @sbglasius approves, we can merge.

sbglasius (approved):
All three review findings are addressed, and I verified the fixes by building rather than by reading:

- the payload key is now anchored with `replaceFirst(/\.json$/, '')`, so a `.json` package segment no longer eats the lookup;
- `-parameters` reaches joint-compiled Java via `GroovyCompile.options.compilerArgs`, with a `CompilePluginSpec` assertion to keep it there;
- `mergeSourced` identifies properties by name + sourceType + sourceMethod, so a shared prefix no longer aborts the build.

I re-ran `generateConfigurationMetadata` for the three modules with the largest hand-written files and compared name, type, defaultValue and description against the pre-review revision: grails-core 16/16 groups and 45/45 properties preserved (+16 inferred), grails-spring-security 14/14 and 145/145 preserved, grails-web-core 10/10 and 20/24 — the four missing `grails.controllers.upload.*` entries are already gone on `8.0.x` from #16040, not from this PR.

Nice piece of work. LGTM.

PR #16048, opened 2026-07-24
Generate configuration metadata from Groovy DSL sources

jamesfredley:
## Summary

- parse explicitly registered Groovy configuration DSL sources with `SourceUnit` without executing scripts, resolving dependencies, or running AST transforms
- merge DSL-derived metadata, typed configuration metadata, and curated overlays with deterministic field-level precedence
- generate Spring Security metadata from `DefaultSecurityConfig` and route every remaining in-repo metadata module through one canonical generated JAR resource
- retain descriptions, hints, legacy properties, and indirect properties in `additional-spring-configuration-metadata.json`

## Relationship

This is a stacked follow-up to #16047 and should be reviewed and merged after it.

It completes the remaining in-repo work for #15469. Asset Pipeline metadata remains in its external repository and existing upstream pull request.

## Verification

- `build-logic`: `check`
- affected module test suites: `grails-core`, `grails-web-core`, `grails-gsp`, `grails-data-mongodb`, Hibernate 5/7 core and dbmigration, and `grails-spring-security` - 2,990 tests passed
- `:grails-test-examples-config-report:integrationTest`
- `:grails-spring-security:codeStyle`
- focused transform-safety, parser edge-case, incremental, and Spring Security metadata contract specs
- direct Spring Security JAR inspection: one canonical metadata entry, 14 groups, and 145 properties


build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,207 @@
...
+        String name = "${prefix}.${segments.join('.')}"
+        Inference inference = infer(assignment.rightExpression)
+        Map<String, Object> property = [name: name]
+        if (inference.type != null) {
+            property.type = inference.type
+        }
+        if (!conditional && inference.literal) {
+            property.defaultValue = inference.value
+        }
+        properties << property
```
jdaugherty, line 179: Duplicate assignments to the same property produce distinct entries here, and the DSL index in `mergeNamed` (`indexByName`) fails the task when two entries for one name aren't map-equal. Two patterns that are valid at runtime under ConfigSlurper's last-write-wins semantics trip this:

```groovy
foo = 'default'
if (Environment.current == Environment.TEST) {
    foo = 'test' // conditional entry has no defaultValue -> conflicts with the unconditional entry
}
```

and if/else branches whose literals infer different types (e.g. `= null` in one branch, `= true` in the other). `DefaultSecurityConfig.groovy` avoids both today — the `password` branches assign the same types on both sides — but the next DSL source registered could fail the build on a perfectly legitimate config. Consider collapsing same-name entries before returning (union the type the way `sharedType` does, keep the unconditional `defaultValue`), or at least add a spec pinning the conflict failure so it's an explicit contract.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,207 @@
...
+            parseStatements(statement.elseBlock, prefix, true, properties)
+        } else if (statement instanceof ExpressionStatement) {
+            Expression expression = statement.expression
+            if (expression instanceof BinaryExpression && expression.operation.type == Types.ASSIGN) {
+                addAssignment(expression, prefix, conditional, properties)
+            } else if (expression instanceof MethodCallExpression) {
+                ClosureExpression closure = closureArgument(expression)
+                String nestedName = expression.methodAsString
+                if (closure != null && nestedName != null) {
+                    parseStatements(closure.code, "${prefix}.${nestedName}", conditional, properties)
```
jdaugherty, line 141: These sources are ConfigSlurper scripts at runtime (`SpringSecurityUtils` loads `DefaultSecurityConfig` with `new ConfigSlurper(Environment.current.name).parse(...)`), so an `environments` block is valid in them. This branch would treat it as an ordinary nested section and silently emit names like `<prefix>.environments.production.foo` instead of `<prefix>.foo`. Nothing hits this today since `DefaultSecurityConfig.groovy` uses `Environment.current` if-checks instead, but it may be worth handling an `environments` call like `IfStatement` (recurse into each environment closure with the unchanged prefix and `conditional = true`) — or failing loudly when one is encountered.

grails-spring-security/plugin/src/test/groovy/grails/plugin/springsecurity/ConfigurationMetadataSpec.groovy
```diff
@@ -0,0 +1,62 @@
...
+class ConfigurationMetadataSpec extends Specification {
+
+    void 'generates merged configuration metadata from DefaultSecurityConfig'() {
+        given:
+        Map metadata = configurationMetadata()
+        Map<String, Map> properties = metadata.get('properties').collectEntries { Map property -> [(property.get('name')): property] }
+
+        expect: 'the complete curated metadata contract is retained'
+        metadata.get('groups').size() == 14
+        properties.size() == 145
```
jdaugherty: These exact counts mean every future addition to `DefaultSecurityConfig.groovy` or the curated overlay fails this spec with only a number mismatch. If a deliberate tripwire is the intent that's fine, but consider deriving the expectation from the packaged `META-INF/additional-spring-configuration-metadata.json` in the same JAR (assert every curated group/property name appears in the canonical file) — the spec then stays self-maintaining and a failure names the missing entry instead of reporting `146 != 145`.

jdaugherty (commented):
I realized this review never got submitted, apologies: 

Reviewed the DSL parser, the merge precedence changes, the module wiring, and the metadata moves. Verified the grails-core metadata consolidation is lossless (all groups, properties, and hints from both prior files are present in the merged additional file). A few comments below, mostly about robustness of the parser against ConfigSlurper semantics for future DSL registrations.

jdaugherty:
@jamesfredley nudge

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
borinquenkid: Fixed in dd1ab42936 + 6c3328e3df. Same-name entries are now grouped and reconciled before they reach `indexByName`: unconditional entries are checked for uniqueness among themselves specifically (order-independent, so a conditional entry sitting between two conflicting unconditional assignments can't mask the conflict), the surviving unconditional entry's `defaultValue` is kept, and the type is unioned across every entry for the name — covers both examples above (differing literal types, and an unconditional default overridden under `Environment.TEST`). Added specs for both, plus a null-branch variant and a "still fails on two genuinely conflicting unconditional assignments" case.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
borinquenkid: Fixed in 6a30571fa3 — an `environments{}` call is now recognized the way `IfStatement` already is: each named environment's closure is parsed under the unchanged prefix (not folded in as `<prefix>.environments.<name>.*`) and marked conditional, so it reconciles through the same merge logic as if/else branches. Added a spec covering `environments { production { ... } test { ... } }` asserting the property lands at the unprefixed path and no `.environments.` name is emitted.

grails-spring-security/plugin/src/test/groovy/grails/plugin/springsecurity/ConfigurationMetadataSpec.groovy
borinquenkid: Fixed in 013ba6c143 — replaced the hardcoded counts with an assertion that every group/property name in the curated overlay is present in the canonical generated metadata, reading the overlay off the classpath the same way the generated file already is (same disambiguation-by-declared-group pattern, since several modules now publish a same-named overlay resource). A missing entry now shows up by name in a `(list - list).empty` failure instead of a bare count mismatch. Verified two ways: it still passes against the real `DefaultSecurityConfig.groovy`, and it doesn't break when I temporarily added a 146th curated property (which the old `properties.size() == 145` would have failed on) — reverted after confirming.

matrei:
Deriving metadata from the DSL defaults instead of duplicating them by hand is a nice idea. The parser handles the shapes in `DefaultSecurityConfig.groovy` well, but it is quite permissive about what it treats as configuration and quite silent about what it skips. The points below are about those two edges, plus the state of the branch.

I ran the parser on small sample inputs for points 1, 2, 4, 5 and 7, and checked the merge state for point 8. The rest come from reading the code.

### Wrong metadata (`GroovyDslConfigurationMetadataParser.groovy`)

**1. Local variable declarations become configuration properties (around line 133)**

```groovy
security {
    def tmp = 'x'
    String other = 'y'
}
```

emits `<prefix>.tmp` and `<prefix>.other` with String types and default values. `DeclarationExpression` extends `BinaryExpression` with an assignment operation, so it passes the assignment check. Skipping `expression instanceof DeclarationExpression` fixes it. Any helper local in a registered DSL file currently ends up in the published `spring-configuration-metadata.json`.

**2. Any method call with a closure argument becomes a nested section (around line 140)**

```groovy
security {
    items.each { item -> visited = true }
}
```

emits `<prefix>.each.visited` (Boolean, default `true`). The receiver of the call is never checked, so `with {}`, `tap {}` and `someList.each {}` all fabricate property paths. Only calls with an implicit-this receiver (`expression.implicitThis`) should be treated as sections.

**3. Type merging ignores branches whose type could not be inferred (around line 80)**

```groovy
if (x) { timeout = 'PT30S' } else { timeout = computeTimeout() }
```

is published as `java.lang.String`, although the other branch may yield an Integer or a Duration. The same happens for an unconditional `foo = 5` followed by an environment override `foo = bar()`. Null literals and genuinely unknown expressions are conflated here. I think only null constants should be ignored; any other uninferable branch should drop the type.

**4. `System.getenv()` without arguments is typed `java.lang.String` (around line 218)**

Every `System.getenv(...)` / `System.getProperty(...)` call is typed as String without looking at the arguments, but `System.getenv()` returns a `Map<String, String>`. Requiring at least one argument before inferring String is enough.

**5. Map literals are emitted as `defaultValue` JSON objects (around line 256)**

*Edit: on reflection I would leave this as it is.* `m = [k: [1, 2]]` produces `"defaultValue": {"k": [1, 2]}`, and the Spring Boot metadata format only describes a scalar or an array of scalars for `defaultValue`. However, the curated overlays in this repo already use object defaults in several places (for example `"defaultValue": {}` for `grails.gorm.custom.types` in grails-core), so the DSL path is consistent with current practice. If object defaults turn out to be a problem for IDEs, that is better handled for curated and generated entries together in a separate change.

### Silently missing metadata

**6. Missing DSL files and unmatched root names are ignored (around line 52)**

An explicitly registered file that does not exist is dropped by `file.isFile()`, and a `dslRootPrefixes` key that matches no top-level call is skipped without a diagnostic. If `DefaultSecurityConfig.groovy` is moved or renamed, or the path in `dslConfigurationFiles.from(...)` has a typo, the task still succeeds and all DSL-derived properties vanish from the published metadata. The plugin spec currently asserts this silent behaviour when the DSL source is deleted. Since these files are registered explicitly, I think a missing file, or a root name that matches nothing, should fail the task (or at least warn).

**7. Only top-level `root { ... }` calls are parsed (around line 99)**

```groovy
security.top = 5

environments {
    production {
        security { envOnly = 1 }
    }
}
```

emits neither property, and nothing is reported. Both are idiomatic ConfigSlurper forms. The `environments` and `if` handling added in `parseStatements` is unreachable at the top level of the file. `parseFile` could route top-level statements through the same statement walker, with the root-prefix mapping applied.

### Branch state and design

**8. The branch no longer merges into its base**

It is 508 commits behind `feature/automated-configuration-metadata`, and `git merge-tree` reports two conflicts:

- A content conflict in `ConfigurationMetadataPlugin.groovy`. The base now reads per-class side-car payloads (`PAYLOAD_DIRECTORY` instead of `PAYLOAD_FIELD`), reads the overlay once up front, and added `addDelegatedProperties`.
- A modify/delete conflict on `grails-core/src/main/resources/META-INF/spring-configuration-metadata.json`. The base added `grails.gorm.defaultIdType` (String, default `long`) to that file, and this PR deletes it. Resolving by keeping the deletion would drop that property from the grails-core metadata, so it needs to be ported into grails-core's `additional-spring-configuration-metadata.json`.

**9. The overlay silently overrides DSL-derived defaults (`ConfigurationMetadataPlugin.groovy`, around line 405)**

I compared the parser output for `DefaultSecurityConfig` with the curated overlay. All 143 DSL properties are already fully specified in the overlay, including `type` and `defaultValue`, so for Spring Security the DSL parsing currently contributes nothing to the published result. For `providerNames`, `voterNames` and `logout.handlerNames` the two even disagree: the DSL default is `[]`, while the overlay publishes non-empty lists such as `[daoAuthenticationProvider, ...]`. The build neither warns nor fails on the mismatch.

That means a future edit to a default in `DefaultSecurityConfig` will not reach the metadata unless the overlay is edited by hand as well, which is the duplication this PR set out to remove. Two options: drop the redundant `type` and `defaultValue` fields from the overlay and keep only descriptions and hints there, or keep them and warn when a DSL default and an overlay default differ. The three list properties probably need a decision either way, since the effective runtime default is not what the DSL literal says.

**10. `config.adoc` (around line 128) documents a build-internal mechanism**

The new paragraph says "Framework modules can also register legacy Groovy configuration DSL sources for static metadata discovery". That describes `org.apache.grails.buildsrc.configuration-metadata`, an unpublished convention plugin in `build-logic`, so application and plugin authors cannot use it. It also goes against the "no internal APIs in docs" rule in the agent guide. I would drop the paragraph from the user guide, or reduce it to a statement of what the published metadata covers.

### Summary

- **Should fix before merge:** 1, 2, 6, 8
- **Worth deciding now:** 9
- **Can follow up:** 3, 4, 7, 10
- **No change needed:** 5

sbglasius (commented):
Code review of the Groovy DSL metadata parser. The merge-precedence rework in `ConfigurationMetadataPlugin` looks behaviour-preserving for the no-DSL case (`mergeNamed([], typed, overlay, ...)` reduces exactly to the old two-arg form), the overlay renames lose no properties or groups, and driving `SourceUnit` only to CONVERSION does genuinely keep `@Grab`/AST transforms out. Four findings, all in the new parser.

Also checked and found clean: all 21 overlay defaults in `additional-spring-configuration-metadata.json` against `DefaultSecurityConfig.groovy` — `rejectRestatedDslDefaults` will not fire. The three list defaults (`voterNames`, `providerNames`, `logout.handlerNames`) differ from the DSL's `[]`, but they correctly document the runtime fallback in `SpringSecurityCoreGrailsPlugin`.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,339 @@
...
+    private static boolean isAssignment(Expression expression) {
+        expression instanceof BinaryExpression && !(expression instanceof DeclarationExpression) &&
+                expression.operation.type == Types.ASSIGN
+    }
+
+    /**
+     * Only a call on the script itself opens a configuration section: {@code items.each { }} or
+     * {@code value.with { }} take a closure as well, but their receiver is ordinary code.
+     */
+    private static boolean isSectionCall(Expression expression) {
```
sbglasius, line 216: **Dotted section headers are silently dropped.**

`isSectionCall` requires `implicitThis`, but the ConfigSlurper dotted-section idiom is a call on a receiver:

```groovy
security.userLookup {
    userDomainClassName = 'User'
}
```

This parses as a `MethodCallExpression` named `userLookup` whose `objectExpression` is `security`, so `implicitThis` is false and `isSectionCall` returns false. `parseTopLevel` then falls through with no match: no property is emitted and no error is raised.

The stale-registration guard in `parse()` does not catch it either — if `security { ... }` appears anywhere else in the same file, `matchedRoots` is already populated, so `unmatchedRoots` is empty. The settings just vanish from the published `spring-configuration-metadata.json`.

Given the rest of the class is deliberately fail-loud about stale registrations, this silent path seems worth closing (or at least detecting and rejecting).

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,339 @@
...
+                [declaration.variableExpression.name]
+    }
+
+    /**
+     * A ConfigSlurper {@code environments { production { ... } } } block picks exactly one named
+     * environment closure at runtime; unlike an ordinary nested section, its name is not part of
+     * the property path. Each environment closure is parsed under the unchanged prefix, marked
+     * conditional the same way an if/else branch is.
+     */
+    private static List<ClosureExpression> environmentClosures(Statement statement) {
```
sbglasius, line 236: **`environmentClosures` is both too narrow and too wide.**

Too narrow — it only recurses into `BlockStatement`, so a conditional inside an environments block is silently dropped:

```groovy
security {
    environments {
        if (someFlag) {
            production { foo = 1 }
        }
    }
}
```

The `IfStatement` is not a `BlockStatement`, and `methodCall()` returns null for it, so `foo` produces no metadata and no error.

Too wide — unlike `isSectionCall`, it accepts *any* `MethodCallExpression` carrying a closure, with no `implicitThis` check:

```groovy
security {
    environments {
        candidates.each { c -> bar = 1 }
    }
}
```

`candidates.each` is treated as an environment closure, so `bar` is emitted as a real configuration property even though it is ordinary loop code. Reusing the `implicitThis` predicate here would keep the two paths consistent.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,339 @@
...
+                                    List<Map<String, Object>> properties) {
+        Inference inference = infer(value)
+        Map<String, Object> property = [name: name, conditional: conditional]
+        if (inference.type != null) {
+            property.type = inference.type
+        } else if (!inference.literal) {
+            property.dynamic = true
+        }
+        if (!conditional && inference.literal) {
+            property.defaultValue = inference.value
```
sbglasius, line 258: **An unconditional `= null` emits `"defaultValue": null` instead of omitting the key.**

`infer()` returns `literal=true, value=null` for a null `ConstantExpression`, so this line sets `property.defaultValue = null`. `mergeGroup` then copies it through via `containsKey('defaultValue')`, and `JsonOutput` writes an explicit `"defaultValue": null` into the shipped metadata.

`DefaultSecurityConfig.groovy` assigns null unconditionally in six places, so six properties are affected:

- `grails.plugin.springsecurity.ajaxCheckClosure`
- `grails.plugin.springsecurity.logout.targetUrlParameter`
- `grails.plugin.springsecurity.rememberMe.cookieDomain`
- `grails.plugin.springsecurity.rememberMe.persistentToken.domainClassName`
- `grails.plugin.springsecurity.rememberMe.useSecureCookie`
- `grails.plugin.springsecurity.x509.subjectDnClosure`

Everywhere else this class signals "no default" by omitting the key (and the specs assert on `containsKey('defaultValue')`), and Spring Boot's own `JsonMarshaller` omits null defaults. As written, IDEs will render a literal `null` default for these six.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
```diff
@@ -0,0 +1,339 @@
...
+        stripped.remove('conditional')
+        stripped.remove('dynamic')
+        stripped
+    }
+
+    /**
+     * The top level of a ConfigSlurper script may open a root section, assign a dotted path that starts
+     * with a root, or wrap either of those in an if statement or an environments block.
+     */
+    private static void parseTopLevel(Statement statement, Map<String, String> rootPrefixes, boolean conditional,
```
sbglasius, line 118: **`parseTopLevel` has no local-variable guard, unlike `parseStatements`.**

`parseStatements` threads a `locals` set and skips assignments whose first segment is a declared local. `parseTopLevel` takes no such parameter, so a script-level local that shadows a registered root leaks into the metadata:

```groovy
def security = [:]
security.enabled = true
```

`leftHandPath` yields `[security, enabled]`, `rootPrefixes['security']` matches, and a spurious `grails.plugin.springsecurity.enabled` property is published. The identical statement inside a `security { ... }` block is correctly skipped, so the two paths disagree.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
matrei: I checked this one against ConfigSlurper before changing anything, and `security.userLookup { ... }` is not a section idiom — it does not parse at runtime. ConfigSlurper only intercepts closure calls on the script itself; `security` resolves to a `ConfigObject`, and the call on it fails:

```
new ConfigSlurper('production').parse("security.userLookup {\n userDomainClassName = 'User'\n}")
=> MissingMethodException: No signature of method: userLookup for class: groovy.util.ConfigObject
```

Same result with a `security { }` block elsewhere in the file, and for the nested form (`security { remember.me { x = 1 } }`), on both Groovy 4.0.31 and 5.1.2. Only dotted *assignments* (`security.userLookup.userDomainClassName = 'User'`) are valid, and those are handled.

So there is no working configuration whose settings vanish here: a source written this way throws the first time the plugin loads it. The fail-loud checks in `parse()` are for the case the runtime cannot catch — a build registration drifting away from a source that still works. Inside a section the shape is also statically indistinguishable from `items.each { }`, which the parser has to keep ignoring. I left this path as it is.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
matrei: Both confirmed, both fixed in 4d49894f05.

Too narrow: ConfigSlurper does accept an `if` inside `environments` (your example yields `[security:[foo:1]]` under `production`), so `environmentClosures` now recurses into both branches of an `IfStatement`.

Too wide: it now uses the same `isSectionCall` predicate as the other two paths, so `candidates.each { ... }` is no longer taken for an environment. That made the private `methodCall` helper unused, so it is gone.

Covered by `"parses an environment declared inside an if statement and ignores ordinary calls in an environments block"` (an environment in each branch of an if/else, plus the `candidates.each` case).

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
matrei: Confirmed, and it was wider than six: `DefaultSecurityConfig.groovy` has 14 unconditional null assignments, and all 14 shipped an explicit `"defaultValue": null` (the six you listed plus `userLookup.userDomainClassName`, `userLookup.authorityJoinClassName`, `authority.className`, `authority.groupAuthorityNameField`, `requestMap.className`, `dao.reflectionSaltSourceProperty`, `switchUser.targetUrl` and `switchUser.switchFailureUrl`).

Fixed in 4d49894f05: `addProperty` now records a default only for a non-null literal. A semantic diff of the regenerated Spring Security metadata shows exactly those 14 keys removed and nothing else changed (145 properties, 14 groups before and after); `"defaultValue": null` no longer occurs in the file. Nulls *inside* a list or map default (`[null, 'present']`) are still preserved, since there they are part of the value.

`"preserves null literals and omits map defaults with non-string keys"` now also asserts that `nothing = null` produces an entry with only a `name`.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
matrei: Confirmed. ConfigSlurper yields an empty config for `def security = [:]; security.enabled = true`, while the parser published `grails.plugin.springsecurity.enabled` — and, worse, counted it as a declaration of the root, so it also defeated the stale-registration check.

Fixed in 4d49894f05: `parseTopLevel` now threads a `locals` set the same way `parseStatements` does (block-scoped, fed by declarations), skips dotted assignments and section calls whose first segment is a local, and hands the script-level locals on to the sections it opens, since those are visible inside the closures. For symmetry, `parseStatements` also no longer treats a call on a local (`helper { ... }`) as a nested section.

New coverage: `"ignores a top-level local variable that shares its name with a registered root"`, `"does not take a top-level local variable for the declaration of a registered root"` (fails with the unmatched-root message), and a `helper { viaLocal = true }` line in the existing local-variable spec.

All five new/extended features fail against the previous parser and pass now. Ran fresh with `--no-build-cache`: `build-logic` `check` (11 specs, `ConfigurationMetadataPluginSpec` 34/34) and `:grails-spring-security:test --tests '*ConfigurationMetadataSpec'`.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
sbglasius: You're right and my premise was wrong. I reproduced your result before replying: `security.userLookup { ... }` throws `MissingMethodException` on both Groovy 4.0.31 and 5.1.0 — plain, with a `security { }` block elsewhere in the same file, and for the nested `security { remember.me { x = 1 } }` form. Only the dotted *assignment* parses, and that path is already handled.

So there is no working configuration whose settings vanish here, and the "statically indistinguishable from `items.each { }`" point stands on its own. Withdrawn — leave the path as it is.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GroovyDslConfigurationMetadataParser.groovy
sbglasius: Fix looks right. Worth noting for anyone reading later that the guard is `inference.value != null` rather than a Groovy truth check, so `false`, `0` and `''` defaults are still recorded — that was the trap here and it's avoided. Nulls inside a list or map value are untouched, as you say.

One leftover, since the principle you state — *a null default is no default* — only got applied to the DSL-derived side. The curated overlays are passed through verbatim by `mergeNamed`, so at this PR's head 17 entries still ship an explicit `"defaultValue": null` into the generated canonical file:

- `grails-data-hibernate5/dbmigration` — 4: `updateOnStartDefaultSchema`, `contexts`, `excludeObjects`, `includeObjects`
- `grails-data-hibernate7/dbmigration` — the same 4
- `grails-data-mongodb/grails-plugin` — 9: `databaseName`, `username`, `password`, `options.readPreference`, `options.writeConcern`, `options.readConcern`, `options.retryWrites`, `options.retryReads`, `options.applicationName`

Pre-existing content that this PR only renames, so not a blocker — but this PR is what makes those three modules publish a generated canonical file, so it seems like the moment to drop them.

sbglasius (approved):
Approving. All four findings from my earlier pass are resolved: three fixed in 4d49894f05, and the fourth was my mistake — `security.userLookup { ... }` is not a working ConfigSlurper form, so there was nothing to fix.

What I verified against the fix rather than taking on trust:

- The null guard is `inference.value != null`, not a Groovy truth check, so `false`, `0` and `''` defaults survive and only genuine nulls are dropped. The regenerated Spring Security metadata keeps 145 properties and 14 groups with no `"defaultValue": null` left.
- 14 unconditional null assignments in `DefaultSecurityConfig.groovy` confirmed (my original six were an artefact of a too-strict grep that missed trailing comments).
- `environmentClosures` now recurses into both branches of an `IfStatement` and reuses `isSectionCall`, so `implicitThis` is required and `candidates.each { }` is ordinary code again.
- The new `locals` guards do not over-reach. I checked the case where a local does not obviously shadow a *method call*: `def helper = { }` followed by `security { helper { x = 1 } }` is a script that runs fine and yields `[security:[:]]` — no settings — and the parser now agrees, where before it published `...springsecurity.helper.x`. The String-local and Map-local cases throw or yield an empty config respectively, and are skipped.
- Withdrawn finding re-checked on Groovy 4.0.31 and 5.1.0, plain / with a `security { }` block elsewhere / nested form: all three throw `MissingMethodException`.

One non-blocking follow-up left on the null thread: 17 entries in the three curated overlays this PR renames still carry an explicit `"defaultValue": null`, which `mergeNamed` passes straight into the generated file. Pre-existing content, happy for it to be a separate change.

jdaugherty:
@matrei are you ok with this PR now?  From reading yours + @sbglasius reviews, I think we can merge if you are.

matrei (approved):

PR #16052, opened 2026-07-25
Fix DDL failure for hasMany collections of enums

### LINKED ISSUE #16051: [Hibernate7G8M4] DDL fails if hasMany on enums
### Expected Behavior

The app should launch without exception, join table should be created

### Actual Behaviour

Join tables are not created. The domain class:
```groovy
package tc

import grails.compiler.GrailsCompileStatic

@CompileStatic
enum SomeEnum {
    FOR_SURE, MAYBE, MAYBE_NOT, DONT_KNOW
}

@GrailsCompileStatic
class SomeDomain {

    Set<SomeEnum> someResponses

    static constraints = {

    }

    static hasMany = [
            someResponses: SomeEnum
    ]
}
```

Here is the wrong DDL:

```bash
$ ./gradlew bootRun
# [. . .]
Hibernate: drop table if exists some_domain cascade 
Hibernate: drop table if exists some_domain_some_responses cascade 
Hibernate: create table some_domain (id bigint generated by default as identity, version bigint not null, tc_some_enum enum ('DONT_KNOW','FOR_SURE','MAYBE','MAYBE_NOT') not null, primary key (id))
Hibernate: create table some_domain_some_responses (some_domain_id bigint not null, primary key (some_domain_id, tc_some_enum))
2026-07-25T07:51:10.624+02:00  WARN 127224 --- [  restartedMain] o.h.t.s.i.ExceptionHandlerLoggedImpl     : GenerationTarget encountered exception accepting command : Error executing DDL "create table some_domain_some_responses (some_domain_id bigint not null, primary key (some_domain_id, tc_some_enum))" via JDBC [Colonne "TC_SOME_ENUM" non trouvée
Column "TC_SOME_ENUM" not found;]

org.hibernate.tool.schema.spi.CommandAcceptanceException: Error executing DDL "create table some_domain_some_responses (some_domain_id bigint not null, primary key (some_domain_id, tc_some_enum))" via JDBC [Colonne "TC_SOME_ENUM" non trouvée
Column "TC_SOME_ENUM" not found;]
	at org.hibernate.tool.schema.internal.exec.GenerationTargetToDatabase.accept(GenerationTargetToDatabase.java:91)
	at org.hibernate.tool.schema.internal.Helper.applySqlString(Helper.java:218)
	at org.hibernate.tool.schema.internal.Helper.applySqlStrings(Helper.java:204)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.createTables(SchemaCreatorImpl.java:404)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.createSequencesTablesConstraints(SchemaCreatorImpl.java:320)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.createFromMetadata(SchemaCreatorImpl.java:215)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.performCreation(SchemaCreatorImpl.java:151)
# [. . .]
```

### Steps To Reproduce

1. bootRun the following test-case

[tc-enum-set.tar.gz](https://github.com/user-attachments/files/30368407/tc-enum-set.tar.gz)

### Environment Information

```bash
❯ uname -a
Linux cachyos-auo-portable 7.1.4-1-cachyos #1 SMP PREEMPT_DYNAMIC Sat, 18 Jul 2026 22:23:17 +0000 x86_64 GNU/Linux

~
❯ java -fullversion
openjdk full version "26.0.1"
```

### Example Application

_No response_

### Version

8.0.0-M4

borinquenkid:
## Summary
- `HibernateBasicProperty.getTable()` always resolved to the owning entity's table instead of the collection's join table, so a `hasMany`-of-enum property's element column was bound to the wrong table: it appeared (wrongly) on the owner and was missing from the join table's own `CREATE TABLE`, breaking schema generation.
- `HibernateToManyProperty.joinTableColumName()` derived the enum element column name from the enum's fully-qualified class name instead of its simple name.
- Introduces `HibernateBasicEnumProperty` so `hasMany`-of-enum elements are recognized as `HibernateEnumProperty` like their singular counterparts, collapsing `EnumTypeBinder` down to a single `bindEnumType()` entry point and letting each `HibernateEnumProperty` implementation supply its own table, column name, nullability, and enum-storage configuration instead of the binder branching on property shape.

Fixes #16051

## Test plan
- [x] Added `EnumHasManyDdlSpec` reproducing the issue: join table gets its element column, owner table gets no spurious column, and a `hasMany`-of-enum can be saved and reloaded end-to-end.
- [x] Updated `EnumTypeBinderSpec` and `BasicCollectionElementBinderSpec` for the collapsed `EnumTypeBinder` API.
- [x] Full `:grails-data-hibernate7-core:test` suite green (3009 tests, 0 failures).

### DIFF
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/GrailsPropertyBinder.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/GrailsPropertyBinder.java
@@ -75,7 +75,11 @@ public class GrailsPropertyBinder {
 
         Value value;
 
-        if (currentGrailsProp instanceof HibernateEnumProperty hibernateEnumProperty) {
+        if (currentGrailsProp instanceof HibernateEnumProperty hibernateEnumProperty &&
+                !(currentGrailsProp instanceof HibernateToManyProperty)) {
+            // A hasMany-of-enum property is also a HibernateEnumProperty, but it must still go
+            // through collectionBinder.bindCollection() below so its join table gets created;
+            // EnumTypeBinder only binds its element column, from BasicCollectionElementBinder.
             value = enumTypeBinder.bindEnumType(hibernateEnumProperty, path);
         } else if (currentGrailsProp.isUserButNotCollectionType()) {
             value = simpleValueBinder.bindBasicValue(currentGrailsProp, parentProperty, path);
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/GrailsPropertyBinder.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/GrailsPropertyBinder.java
@@ -76,7 +76,7 @@ public class GrailsPropertyBinder {
         Value value;
 
         if (currentGrailsProp instanceof HibernateEnumProperty hibernateEnumProperty &&
-                !(currentGrailsProp instanceof HibernateToManyProperty)) {
+                !hibernateEnumProperty.isCollectionElement()) {
             // A hasMany-of-enum property is also a HibernateEnumProperty, but it must still go
             // through collectionBinder.bindCollection() below so its join table gets created;
             // EnumTypeBinder only binds its element column, from BasicCollectionElementBinder.
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
@@ -1,0 +1,56 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
+
+import java.beans.PropertyDescriptor;
+
+import org.grails.datastore.mapping.model.MappingContext;
+import org.grails.orm.hibernate.cfg.PersistentEntityNamingStrategy;
+import org.grails.orm.hibernate.cfg.domainbinding.util.ColumnNameForPropertyAndPathFetcher;
+
+/**
+ * Hibernate basic collection element property whose element type is an enum. Created by {@link
+ * HibernateMappingFactory#createBasicCollection} when the collection's element type is an enum.
+ */
+public class HibernateBasicEnumProperty extends HibernateBasicProperty implements HibernateEnumProperty {
+
+    public HibernateBasicEnumProperty(
+            GrailsHibernatePersistentEntity entity, MappingContext context, PropertyDescriptor property) {
+        super(entity, context, property);
+    }
+
+    @Override
+    public Class<?> getEnumType() {
+        return getComponentType();
+    }
+
+    @Override
+    public String resolveEnumColumnName(
+            PersistentEntityNamingStrategy namingStrategy,
+            ColumnNameForPropertyAndPathFetcher columnNameForPropertyAndPathFetcher,
+            String path) {
+        return joinTableColumName(namingStrategy);
+    }
+
+    /** A hasMany element column is always nullable, matching the non-enum sibling binding path. */
+    @Override
+    public boolean isEnumColumnNullable() {
+        return true;
+    }
+}
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
@@ -20,6 +20,9 @@ package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
 
 import java.beans.PropertyDescriptor;
 
+import org.hibernate.mapping.Collection;
+import org.hibernate.mapping.Table;
+
 import org.grails.datastore.mapping.model.MappingContext;
 import org.grails.orm.hibernate.cfg.PersistentEntityNamingStrategy;
 import org.grails.orm.hibernate.cfg.domainbinding.util.ColumnNameForPropertyAndPathFetcher;
@@ -57,5 +60,21 @@ public class HibernateBasicEnumProperty extends HibernateBasicProperty implement
     @Override
     public boolean isCollectionElement() {
         return true;
+    }
+
+    /**
+     * For an enum collection element, the table to bind the element column against is the
+     * collection's join table rather than the owning entity's table. Before the collection
+     * table has been assigned (e.g. while it is itself being computed), falls back to the
+     * owning entity's table, matching the pre-collection-binding default. Scoped to the enum
+     * subclass because only {@link org.grails.orm.hibernate.cfg.domainbinding.binder.EnumTypeBinder}
+     * binds through {@code getTable()}; the non-enum element binding reads the collection
+     * table directly.
+     */
+    @Override
+    public Table getTable() {
+        Collection collection = getHibernateCollection();
+        Table collectionTable = collection != null ? collection.getCollectionTable() : null;
+        return collectionTable != null ? collectionTable : getPersistentClass().getTable();
     }
 }
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
@@ -53,4 +53,9 @@ public class HibernateBasicEnumProperty extends HibernateBasicProperty implement
     public boolean isEnumColumnNullable() {
         return true;
     }
+
+    @Override
+    public boolean isCollectionElement() {
+        return true;
+    }
 }
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicProperty.java
@@ -21,7 +21,6 @@ package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
 import java.beans.PropertyDescriptor;
 
 import org.hibernate.mapping.Collection;
-import org.hibernate.mapping.Table;
 
 import org.grails.datastore.mapping.model.MappingContext;
 import org.grails.datastore.mapping.model.types.mapping.BasicWithMapping;
@@ -45,17 +44,5 @@ public class HibernateBasicProperty extends BasicWithMapping<PropertyConfig> imp
     @Override
     public void setHibernateCollection(Collection collection) {
         this.collection = collection;
-    }
-
-    /**
-     * For a basic (scalar or enum) collection element, the property's table is the
-     * collection's join table rather than the owning entity's table. Before the collection
-     * table has been assigned (e.g. while it is itself being computed), falls back to the
-     * owning entity's table, matching the pre-collection-binding default.
-     */
-    @Override
-    public Table getTable() {
-        Table collectionTable = collection != null ? collection.getCollectionTable() : null;
-        return collectionTable != null ? collectionTable : getPersistentClass().getTable();
     }
 }
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicProperty.java
@@ -21,6 +21,7 @@ package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
 import java.beans.PropertyDescriptor;
 
 import org.hibernate.mapping.Collection;
+import org.hibernate.mapping.Table;
 
 import org.grails.datastore.mapping.model.MappingContext;
 import org.grails.datastore.mapping.model.types.mapping.BasicWithMapping;
@@ -44,5 +45,17 @@ public class HibernateBasicProperty extends BasicWithMapping<PropertyConfig> imp
     @Override
     public void setHibernateCollection(Collection collection) {
         this.collection = collection;
+    }
+
+    /**
+     * For a basic (scalar or enum) collection element, the property's table is the
+     * collection's join table rather than the owning entity's table. Before the collection
+     * table has been assigned (e.g. while it is itself being computed), falls back to the
+     * owning entity's table, matching the pre-collection-binding default.
+     */
+    @Override
+    public Table getTable() {
+        Table collectionTable = collection != null ? collection.getCollectionTable() : null;
+        return collectionTable != null ? collectionTable : getPersistentClass().getTable();
     }
 }
--- a/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateEnumProperty.java
+++ b/grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateEnumProperty.java
@@ -18,18 +18,46 @@
  */
 package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
 
+import org.grails.orm.hibernate.cfg.PersistentEntityNamingStrategy;
+import org.grails.orm.hibernate.cfg.domainbinding.util.ColumnNameForPropertyAndPathFetcher;
+
 /**
  * Marker interface for Hibernate persistent properties whose Java type is an enum.
  *
- * <p>Two concrete subtypes exist, corresponding to the two creation paths in {@link
+ * <p>Three concrete subtypes exist, corresponding to the three creation paths in {@link
  * HibernateMappingFactory}:
  *
  * <ul>
  *   <li>{@link HibernateSimpleEnumProperty} — plain enum with no custom type marshaller
  *   <li>{@link HibernateCustomEnumProperty} — enum backed by a custom type marshaller
+ *   <li>{@link HibernateBasicEnumProperty} — enum element of a {@code hasMany} basic collection
  * </ul>
  *
  * <p>Use {@code instanceof HibernateEnumProperty} instead of {@code isEnumType()} to branch on
- * enum properties at binding time.
+ * enum properties at binding time. Each implementation resolves its own enum class and column
+ * name so {@link org.grails.orm.hibernate.cfg.domainbinding.binder.EnumTypeBinder} can bind any
+ * of them through a single code path.
  */
-public interface HibernateEnumProperty extends HibernatePersistentProperty {}
+public interface HibernateEnumProperty extends HibernatePersistentProperty {
+
+    /** The enum class to bind: the property's own type, or a basic collection's element type. */
+    default Class<?> getEnumType() {
+        return getType();
+    }
+
+    /** Resolves the column name to bind the enum value under. */
+    default String resolveEnumColumnName(
+            PersistentEntityNamingStrategy namingStrategy,
+            ColumnNameForPropertyAndPathFetcher columnNameForPropertyAndPathFetcher,
+            String path) {
+        return columnNameForPropertyAndPathFetcher.getColumnNameForPropertyAndPath(this, path, null);
+    }
+
+    /**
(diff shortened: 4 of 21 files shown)


grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.grails.datastore.mapping.model.MappingContext;
+import org.grails.orm.hibernate.cfg.PersistentEntityNamingStrategy;
+import org.grails.orm.hibernate.cfg.domainbinding.util.ColumnNameForPropertyAndPathFetcher;
+
+/**
+ * Hibernate basic collection element property whose element type is an enum. Created by {@link
+ * HibernateMappingFactory#createBasicCollection} when the collection's element type is an enum.
+ */
+public class HibernateBasicEnumProperty extends HibernateBasicProperty implements HibernateEnumProperty {
```
jdaugherty, line 34: Making this type a `HibernateEnumProperty` silently changes behaviour in `PropertyBinder`, which is outside this diff:

```java
// PropertyBinder#bindProperty
if (persistentProperty instanceof Association<?> association &&
        !(persistentProperty instanceof HibernateEnumProperty)) {
    prop.setCascade(cascadeBehaviorFetcher.getCascadeBehaviour(association));
}
```

`HibernateBasicProperty` extends `BasicWithMapping` -> `Basic` -> `ToMany` -> `Association`, so a `hasMany`-of-enum **is** an `Association`. Before this PR the `!(... instanceof HibernateEnumProperty)` clause was unreachable (the only two implementations extended `HibernateSimpleProperty`/`HibernateCustomProperty`, neither of which is an `Association`); after this PR it fires, and `prop.setCascade(...)` is no longer called for enum collections.

Two concrete consequences:

1. `CascadeBehaviorFetcher.getImpliedBehavior()` has an explicit `if (association instanceof Basic) return ALL;` branch, so `Set<String>` still gets `cascade="all"` while `Set<SomeEnum>` now gets none — the two basic-collection shapes have diverged for no stated reason.
2. An explicitly configured `answers cascade: 'all-delete-orphan'` in the mapping block is read by `getDefinedBehavior()` and is now dropped on the floor for enum collections only.

For a collection of value types the runtime blast radius is small (there is no associated entity to cascade to), so this is unlikely to be what CI would catch — but it is an unintended side effect, and the same class of side effect applies to every `instanceof HibernateEnumProperty` site once a collection type joins that hierarchy. Please either narrow the `PropertyBinder` guard to the scalar enum types it was written for, or add an explicit test asserting the cascade for `hasMany`-of-enum matches `hasMany`-of-String.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateToManyProperty.java
```diff
@@ -226,12 +226,14 @@ default String joinTableColumName(PersistentEntityNamingStrategy namingStrategy)
         String columnName;
         if (present) {
             columnName = joinColumnMappingOptional.get().getName();
+        } else if (referencedType.isEnum()) {
+            // Use the enum's simple name, not its fully-qualified name, so the column
+            // isn't named after the enum's package.
+            columnName = namingStrategy.resolveColumnName(referencedType.getSimpleName());
```
jdaugherty, line 269: `HibernateToManyPropertySpec` is the direct unit test for this method and it was not touched. Its enum case currently asserts nothing useful:

```groovy
void "joinTableColumName returns derived column name for enum collection"() {
    given:
    def property = createTestHibernateToManyProperty(HTMPEntityWithEnum, "statuses")
    def namingStrategy = getGrailsDomainBinder().namingStrategy

    expect:
    property.joinTableColumName(namingStrategy) != null
}
```

That passed before this change and passes after it, so the behaviour you are fixing here has no unit-level guard. Since the sibling case two features down (`"joinTableColumName uses explicit join table column name when present"`) already asserts an exact string, please pin this one the same way — `== "htmp_status"` or whatever the strategy resolves for that enum's simple name. That also documents the Grails 7 parity this restores.

grails-data-hibernate7/core/src/test/groovy/grails/gorm/tests/EnumHasManyDdlSpec.groovy
```diff
@@ -0,0 +1,117 @@
...
+                "select column_name from information_schema.columns " +
+                "where table_name = 'SURVEY_RESPONSE_ANSWERS'"
+            ).executeQuery()
+        }
+        Set<String> columnNames = []
+        while (columns.next()) {
+            columnNames << columns.getString('column_name').toLowerCase()
+        }
+        columnNames.contains('survey_response_id')
+        columnNames.any { it.contains('answer') }
```
jdaugherty: `it.contains('answer')` is too loose to cover the naming half of the fix. Under the old FQN behaviour the column was `grails_gorm_tests_survey_answer`, which also satisfies `contains('answer')` — so this feature would have gone green against the buggy `joinTableColumName` as long as the *table* half of the fix were in place. Please assert the exact name:

```groovy
columnNames == ['survey_response_id', 'survey_answer'] as Set
```

Two smaller points on the same block:

- The `PreparedStatement`/`ResultSet` are never closed, and the `ResultSet` is drained after `doReturningWork` has returned. It happens to work because the session's connection is still open, but the idiomatic form is to do the whole read inside the callback and return the `Set<String>` — which also lets both features share one helper instead of duplicating the loop.
- The refactor moved `STRING`/`ORDINAL`/`IDENTITY` handling into `GrailsEnumType.configure` and nullability into `isEnumColumnNullable`, but every assertion here exercises the default (STRING) path. A case with `answers enumType: 'ordinal'` and one with an explicit `joinTable column: [name: '...']` would cover the branches this PR actually rewrote.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicEnumProperty.java
```diff
@@ -0,0 +1,56 @@
...
+    public String resolveEnumColumnName(
+            PersistentEntityNamingStrategy namingStrategy,
+            ColumnNameForPropertyAndPathFetcher columnNameForPropertyAndPathFetcher,
+            String path) {
+        return joinTableColumName(namingStrategy);
+    }
+
+    /** A hasMany element column is always nullable, matching the non-enum sibling binding path. */
+    @Override
+    public boolean isEnumColumnNullable() {
```
jdaugherty, line 56: This is a behaviour change that the PR description does not call out. Previously `bindEnumTypeForColumn` reached the shared `column.setNullable(property.isNullable())`, so a non-nullable enum collection element got a `NOT NULL` column; now it is unconditionally nullable.

I think the new behaviour is the right one — it matches the `true` that `BasicCollectionElementBinder` passes to `simpleValueColumnBinder.bindSimpleValue(...)` on the non-enum branch, and it matches how Hibernate treats `@ElementCollection` element columns — so this reads as an intentional alignment rather than an accident. But it does change generated DDL for anyone who declared `answers nullable: false`, so it deserves a line in the PR body and a test pinning the resulting column's nullability.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/binder/GrailsPropertyBinder.java
```diff
@@ -75,7 +75,11 @@ public Value bindProperty(
 
         Value value;
 
-        if (currentGrailsProp instanceof HibernateEnumProperty hibernateEnumProperty) {
+        if (currentGrailsProp instanceof HibernateEnumProperty hibernateEnumProperty &&
+                !(currentGrailsProp instanceof HibernateToManyProperty)) {
```
jdaugherty: This guard is correct but states the condition indirectly — it reads as "enums that are not to-many" when what you mean is "not the new basic-collection variant". It also silently depends on `HibernateBasicProperty implements HibernateToManyCollectionProperty`, which is not obvious from here.

`!(currentGrailsProp instanceof HibernateBasicEnumProperty)` says the same thing directly and is the condition that stays true if another `HibernateEnumProperty` implementation is added later. Alternatively, put the decision on the interface (e.g. a `bindsOwnValue()`/`isCollectionElement()` default) so the binder does not have to enumerate implementations at all — that is the pattern the rest of this PR follows for table, column name and nullability.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateEnumProperty.java
```diff
@@ -18,18 +18,46 @@
  */
 package org.grails.orm.hibernate.cfg.domainbinding.hibernate;
 
+import org.grails.orm.hibernate.cfg.PersistentEntityNamingStrategy;
+import org.grails.orm.hibernate.cfg.domainbinding.util.ColumnNameForPropertyAndPathFetcher;
+
 /**
  * Marker interface for Hibernate persistent properties whose Java type is an enum.
```
jdaugherty: This opening line is now wrong twice over: the interface is no longer a marker (it carries three default methods), and `HibernateBasicEnumProperty`'s Java type is a `Collection`, not an enum. Something like "Contract for Hibernate persistent properties that bind an enum value — either the property's own type or a basic collection's element type" would match what the interface has become, and the list below it already spells out the three implementations.

jdaugherty (commented):
I had AI review this and think it's comments are relevant.  Take a look and then I'll take another pass.

jdaugherty:
@borinquenkid have you finished working through the feedback on this PR?  Can I take another look yet?

borinquenkid:
> @borinquenkid have you finished working through the feedback on this PR? Can I take another look yet?

@jdaugherty yes

grails-data-hibernate7/core/src/test/groovy/grails/gorm/tests/EnumHasManyDdlSpec.groovy
```diff
@@ -0,0 +1,207 @@
...
+
+        when:
+        def response = new OrdinalSurveyResponse(respondent: "Bob")
+        response.addToAnswers(SurveyAnswer.FOR_SURE)
+        response.save(flush: true)
+        response.discard()
+        def reloaded = OrdinalSurveyResponse.get(response.id)
+
+        then:
+        reloaded.answers == [SurveyAnswer.FOR_SURE] as Set
```
jdaugherty: This feature does not test what its name claims. I ran it against the PR with the only change being `enumType: 'ordinal'` -> `enumType: 'string'` on `OrdinalSurveyResponse`, and it still passes:

```
EnumHasManyDdlSpec > a hasMany of enum with enumType ordinal stores the ordinal, not the name PASSED
Results: SUCCESS (6 tests, 6 successes, 0 failures, 0 skipped)
```

That is expected from the two assertions:

- `columnNamesFor(...)` compares column *names*, which are identical for `STRING` and `ORDINAL`. The `expect:` label even says "the element column is a numeric ordinal column, not a string one", but nothing in the expression looks at the type.
- The round trip reads through the same mapping it wrote through, so it succeeds under either style.

This is the same shape as the `contains('answer')` issue from the last round — the feature would go green against a build where `GrailsEnumType.ORDINAL.configure()` was never reached, which is exactly the branch this PR rewrote.

`IdentityEnumTypeSpec` in this same package is the established pattern for pinning storage representation — it selects the raw column and asserts the stored value:

```groovy
ResultSet resultSet = ds.getConnection().prepareStatement('select status from enum_entity_domain').executeQuery()
```

Something equivalent here would actually discriminate, e.g. asserting the raw `survey_answer` value is `0` for `FOR_SURE`, or reading `data_type` from `information_schema.columns` and asserting it is numeric rather than character. As written the `ORDINAL` and `IDENTITY` arms of the new `GrailsEnumType.configure()` have no coverage on the collection path at all.

Minor, same block: `expect:` followed by `when:`/`then:` puts two unrelated assertions in one feature. Splitting the DDL check from the round trip would let each fail with its own message.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateToManyProperty.java
```diff
@@ -226,12 +263,14 @@ default String joinTableColumName(PersistentEntityNamingStrategy namingStrategy)
...
```
jdaugherty, line 269: Repeating this from the last round because it is unchanged: `HibernateToManyPropertySpec` is the direct unit test for this method and it still asserts nothing about the value.

```groovy
void "joinTableColumName returns derived column name for enum collection"() {
    ...
    expect:
    property.joinTableColumName(namingStrategy) != null
}
```

I confirmed the gap is real. Reverting just this line to `referencedType.getName()` and running both specs:

```
EnumHasManyDdlSpec > join table for a hasMany of enum is created with the element column FAILED
EnumHasManyDdlSpec > a hasMany of enum with enumType ordinal ... FAILED
EnumHasManyDdlSpec > the hasMany enum element column stays nullable ... FAILED
HibernateToManyPropertySpec > joinTableColumName returns derived column name for enum collection PASSED
```

So the integration spec now guards the fix (good — that is an improvement over last round), but the unit spec for the changed method still passes against the bug. The sibling feature two down, `"joinTableColumName uses explicit join table column name when present"`, already asserts `== "tag_val"`; please pin this one the same way.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateMappingFactory.groovy
```diff
@@ -167,10 +167,13 @@ class HibernateMappingFactory extends AbstractGormMappingFactory<Mapping, Proper
             PersistentEntity entity, MappingContext context, PropertyDescriptor property, Class collectionType) {
         if (entity instanceof GrailsHibernatePersistentEntity) {
             GrailsHibernatePersistentEntity ghpEntity = (GrailsHibernatePersistentEntity) entity
-            HibernateBasicProperty basic = new HibernateBasicProperty(ghpEntity, context, property)
+            boolean isEnumCollection = collectionType != null && collectionType.isEnum()
+            HibernateBasicProperty basic = isEnumCollection
+                    ? new HibernateBasicEnumProperty(ghpEntity, context, property)
+                    : new HibernateBasicProperty(ghpEntity, context, property)
```
jdaugherty, line 173: `HibernateMappingFactorySpec` has three features covering exactly this method, and none of them can tell whether this branch is present. Reducing it back to `new HibernateBasicProperty(ghpEntity, context, property)` leaves the spec fully green:

```
Results: SUCCESS (29 tests, 29 successes, 0 failures, 0 skipped)
```

The three features are:

- `"createBasicCollection produces HibernateBasicProperty for a basic element collection"`
- `"createBasicCollection sets custom marshaller for enum hasMany"`
- `"createBasicCollection uses Enum base marshaller when no specific marshaller for enum collection type"`

The latter two build entities whose collections *are* enums, and all three assert `instanceof HibernateBasicProperty` — the supertype, which stays true either way.

The latter two should assert `HibernateBasicEnumProperty`; the first should assert the negative (`!(sectionsProp instanceof HibernateBasicEnumProperty)`) so the split is pinned from both sides. `EnumHasManyDdlSpec` does catch the regression at boot, so this is a weak-assertion problem rather than uncovered behaviour — but it is a one-word fix in each case.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateBasicProperty.java
```diff
@@ -45,4 +46,16 @@ public Collection getHibernateCollection() {
...
+     * For a basic (scalar or enum) collection element, the property's table is the
+     * collection's join table rather than the owning entity's table. Before the collection
+     * table has been assigned (e.g. while it is itself being computed), falls back to the
+     * owning entity's table, matching the pre-collection-binding default.
+     */
+    @Override
+    public Table getTable() {
+        Table collectionTable = collection != null ? collection.getCollectionTable() : null;
+        return collectionTable != null ? collectionTable : getPersistentClass().getTable();
+    }
```
jdaugherty: Two things about placing this override on `HibernateBasicProperty` rather than on `HibernateBasicEnumProperty`.

**Only the enum path needs it.** `BasicCollectionElementBinder`'s non-enum branch reads `collection.getCollectionTable()` directly and never calls `property.getTable()`, so scalar basic collections gain nothing here and only inherit the risk. Scoping it to the enum subclass would match where `resolveEnumColumnName` and `isEnumColumnNullable` were placed.

**It changes what `TableForManyCalculator.getJoinTableSchema()` reads.**

```java
String owningTableSchema = property.getTable().getSchema();
```

For a basic collection that expression no longer means what the variable is named. It still returns the right value, but only because of an ordering coincidence in `CollectionBinder.bindCollection()`:

1. `collectionHolder.create(property)` -> `CollectionType.create()` does `coll.setCollectionTable(owner.getTable())`
2. `property.setCollection(collection, path)` -> the field here becomes non-null, so `getTable()` starts returning the collection table
3. `bindCollectionTable()` calls `getJoinTableSchema()`, which now reads the collection table — still the owner's table from step 1
4. `collection.setCollectionTable(<real join table>)`

The correct schema survives only because of the seed in step 1. The javadoc says the fallback covers "before the collection table has been assigned", but by the time `getJoinTableSchema()` runs the collection table *is* assigned — to the owner's table. Anything that reorders steps 1 and 3 silently changes the schema of every basic join table.

Either scope the override to the enum subclass, or have `getJoinTableSchema()` ask the owner directly (`property.getPersistentClass().getTable().getSchema()`) so it stops depending on this ordering.

`EnumHasManyDdlSpec` covers the override end-to-end, so there is no coverage hole — but a feature in `HibernateBasicPropertySpec` pinning both arms of the ternary would be worth adding alongside whichever fix you pick.

grails-data-hibernate7/core/src/test/groovy/org/grails/orm/hibernate/cfg/domainbinding/GrailsPropertyBinderSpec.groovy
```diff
@@ -250,6 +256,37 @@ class GrailsPropertyBinderSpec extends HibernateGormDatastoreSpec {
...
+            ClassMapping getClassMapping() { null }
+            PropertyConfig getMappedForm() { new PropertyConfig() }
+        })
+
+        expect: "no type: mapping means the isUserButNotCollectionType() branch cannot intercept it"
+        !customProp.isUserButNotCollectionType()
+
+        when:
+        Value value = propertyBinder.bindProperty(customProp, null, EMPTY_PATH)
+
```
jdaugherty, line 313: The new branch in `GrailsPropertyBinder` — `&& !hibernateEnumProperty.isCollectionElement()` — has no test in this spec. The only thing currently standing behind it is `EnumHasManyDdlSpec` booting successfully, which will not tell you which side of the condition broke.

Two features in the same style as the ones already here would pin it at the level it was written:

- a scalar enum property yields the `BasicValue` from `enumTypeBinder`
- a `hasMany`-of-enum property falls through to `collectionBinder` and yields a `Collection`

The second is the one that regresses silently if `isCollectionElement()` ever returns the wrong thing.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/domainbinding/hibernate/HibernateEnumProperty.java
```diff
@@ -18,18 +18,57 @@
...
+    }
+
+    /**
+     * Whether the enum column should allow NULL. Subclass properties in a table-per-hierarchy
+     * strategy must be nullable; otherwise this follows the property's own nullable constraint.
+     */
+    default boolean isEnumColumnNullable() {
+        return getHibernateOwner().isTablePerHierarchySubclass() || isNullable();
+    }
+
```
jdaugherty, line 64: The debug log that used to accompany the table-per-hierarchy case in `EnumTypeBinder` is gone:

```java
LOG.debug("[GrailsDomainBinder] Sub class property [{}] for column name [{}] forced to nullable", ...)
```

That message is the only signal a user gets that their `nullable: false` on a `table-per-hierarchy` subclass was deliberately overridden. `isEnumColumnNullable()` returns a boolean so it has nowhere natural to log; keeping the message at the `EnumTypeBinder` call site would preserve it.

jdaugherty (approved):
Approving — the fix is correct and now properly guarded. Reverting `joinTableColumName` to `getName()` fails three `EnumHasManyDdlSpec` features, so the regression is genuinely pinned. The round-one items all landed and I've resolved those threads: `EnumHasManyDdlSpec` asserts exact column names and reads inside `doReturningWork` with try-with-resources, the `joinTable column:` and `nullable: false` cases are there, the `GrailsPropertyBinder` guard moved onto the interface as `isCollectionElement()`, and the `HibernateEnumProperty` javadoc is accurate. The extra `HibernateCustomProperty` coverage in `GrailsPropertyBinderSpec` is a good addition too.

Everything below is non-blocking follow-up — none of it is a correctness problem and none of it should hold up the fix. Findings are verified by mutation: I broke the production code and checked whether the suite noticed, rather than inferring from reading.

### Three tests that don't assert what they're named for

1. **`EnumHasManyDdlSpec`, the ordinal feature.** Substituting `enumType: 'string'` for `'ordinal'` leaves it green. Column names are identical under both styles, and the round trip reads through the mapping it wrote.
2. **`HibernateMappingFactorySpec`.** Deleting the `HibernateBasicEnumProperty` branch outright leaves it 29/29 green — all three `createBasicCollection` features assert the supertype.
3. **`HibernateToManyPropertySpec`** (carried over from round one, still unchanged — I've left that thread open). Its `joinTableColumName` feature asserts `!= null` and passes against the bug it is named for.

None of these are uncovered behaviour — the suite catches all three regressions elsewhere, which is why they stayed green. They're tests that wouldn't fail if the thing they're named for broke, and each is a small fix. Happy for them to land as a follow-up rather than here.

### One design question

`HibernateBasicProperty.getTable()` is overridden on the shared base class, but only the enum path reads it, and it changes what `TableForManyCalculator.getJoinTableSchema()` sees. That still yields the right schema, but only because `CollectionType.create()` seeds `collectionTable` to the owner's table before `getJoinTableSchema()` runs. Details inline — worth deciding deliberately rather than inheriting, but not a blocker.

borinquenkid:
@jdaugherty nudge

jdaugherty (approved):

jdaugherty:
Given how long this has been out there and the critical nature of it, I'm going to merge it.

PR #16054, opened 2026-07-26
Address review feedback on databinding whitelist characterization spec

borinquenkid:
Stacked on #16003.

Addresses review comments from @jdaugherty:
- Renames the Object-typed field to `untypedProperty` and rewords test names — the property is excluded because it is raw Object-typed (`DefaultASTDatabindingHelper#shouldFieldBeInWhiteList`), not because it is generically "unlisted".
- Drops the misleading `hasOne` association, which added nothing to the tested contract and declares an incomplete mapping (no `belongsTo` back-reference on `WhitelistAddress`).
- Trims the domain test to its non-duplicated assertions and cross-references `DefaultASTDatabindingHelperDomainClassSpecialPropertiesSpec`, which already pins id/version/dateCreated/lastUpdated exclusion for domain classes.

## Verification
- `:grails-test-suite-web:test --tests DefaultDatabindingWhitelistBehaviorSpec`
- `./gradlew codeStyle`


borinquenkid:
Folding this directly into #16003 instead — no need for a separate stacked PR.

PR #16065, opened 2026-07-29
fix: preserve full transaction attribute state in GrailsTransactionAttribute (web) copy constructors

borinquenkid:
# fix: preserve full transaction attribute state in GrailsTransactionAttribute copy constructors

Follow-up to #16063 (`fix/customizable-rollback-tx-attribute-copy`), which fixed the same bug in `org.grails.datastore.mapping.transactions.CustomizableRollbackTransactionAttribute`. That PR's "known follow-up" section named this class as carrying the identical pattern; this PR is that fix.

## Problem

`org.grails.transaction.GrailsTransactionAttribute` (grails-core) had the same lossy copy constructors as its GORM-module namesake and the CRTA class:

- The `(TransactionAttribute)` overload copied only the five `TransactionDefinition` fields — **rollback rules, qualifier, labels, descriptor, and `timeoutString` were silently dropped**.
- The `(TransactionDefinition)` overload had the same gap.
- The `(RuleBasedTransactionAttribute)` overload called `super(other)` (copying definition fields and rules) but still dropped qualifier, labels, descriptor, and `timeoutString`.
- Two trace log statements used GString-style placeholders (`"$ex"`, `"$winner"`) that never interpolate in `.java` source, and a third trace statement logged unconditionally without an `isTraceEnabled()` guard.

This class has no in-repo constructor callers — `grep`ing `new GrailsTransactionAttribute(` across the whole repo turns up only `GrailsTransactionTemplate.groovy:55`, which resolves to the same-package GORM twin (`grails.gorm.transactions.GrailsTransactionAttribute`, fixed separately in #16064), not this class. This class is public API consumed by downstream plugins and applications that construct or extend it directly.

## Fix

Same pattern as the CRTA fix (`4042a87d54`):

- `(TransactionAttribute)` now delegates to `(TransactionDefinition)`.
- `(TransactionDefinition)` copies the five definition fields, then recovers the dynamic type via pattern-matching `instanceof`: `copyAttributeState` for `TransactionAttribute` sources, and a rollback-rules snapshot through a temporary `new RuleBasedTransactionAttribute(other).getRollbackRules()` — never the source's lazy getter, so the source is never mutated.
- `(RuleBasedTransactionAttribute)` does `super(other)` + `copyAttributeState` + `copyGrailsState`.
- New private helpers `copyAttributeState` (descriptor/timeoutString when source is `DefaultTransactionAttribute`, qualifier, defensive `new ArrayList<>(labels)` copy) and `copyGrailsState` (`inheritRollbackOnly`, this class's only Grails-specific field — it has no `connection` field, unlike the CRTA class).
- Log placeholders fixed to string concatenation; the previously-unguarded trace call is now behind `isTraceEnabled()`.

No other production files touched.

Same shape as the CRTA change: downstream code that passes its own rule-bearing `TransactionAttribute` into this class's copy constructors will now have those rules **honored** — an exception matching a `NoRollbackRuleAttribute` commits instead of rolling back. Previously the rules were silently dropped and every exception rolled back. One-directional: no scenario turns a commit into a rollback. Since this class has no in-repo callers, the practical impact is scoped to plugin/application code outside this repository.

## Tests

`GrailsTransactionAttributeSpec` (13 tests, modeled on `CustomizableRollbackTransactionAttributeSpec`): rule-list and labels deep-copy independence in both directions with explicit non-mutation-of-source assertions, qualifier/labels/`inheritRollbackOnly` preservation, all five definition fields, descriptor/`timeoutString`, a plain `RuleBasedTransactionAttribute` source, rollback-on behavior (no-rollback rule honored, deepest rule wins, rollback-everything default), and the statically-dispatched `TransactionDefinition`/`TransactionAttribute` entry points.

- `:grails-core:test` — 475 tests, 0 failures.
- `:grails-core:codeStyle` — clean.

### DIFF
--- a/grails-core/src/main/groovy/org/grails/transaction/GrailsTransactionAttribute.java
+++ b/grails-core/src/main/groovy/org/grails/transaction/GrailsTransactionAttribute.java
@@ -19,12 +19,15 @@
 
 package org.grails.transaction;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.interceptor.DefaultTransactionAttribute;
 import org.springframework.transaction.interceptor.NoRollbackRuleAttribute;
 import org.springframework.transaction.interceptor.RollbackRuleAttribute;
 import org.springframework.transaction.interceptor.RuleBasedTransactionAttribute;
@@ -52,12 +55,7 @@ public class GrailsTransactionAttribute extends RuleBasedTransactionAttribute {
     }
 
     public GrailsTransactionAttribute(TransactionAttribute other) {
-        super();
-        setPropagationBehavior(other.getPropagationBehavior());
-        setIsolationLevel(other.getIsolationLevel());
-        setTimeout(other.getTimeout());
-        setReadOnly(other.isReadOnly());
-        setName(other.getName());
+        this((TransactionDefinition) other);
     }
 
     public GrailsTransactionAttribute(TransactionDefinition other) {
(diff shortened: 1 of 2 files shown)


borinquenkid:

borinquenkid:
The CI failures on this PR (`Build Grails-Core` failing on ubuntu-latest/macos-latest/windows-latest, "Rerunning all Tasks") are pre-existing on `8.0.x` and unrelated to this change. Both failures are the same root cause:

Both assert on a `System.err`-backed `ByteArrayOutputStream` capture of the stack trace filterer's output, but `8.0.x` now routes that output through a Logback appender instead of `System.err`, so the capture is empty.

borinquenkid:
@jamesfredley nudge

PR #16066, opened 2026-07-29
feat: GORM O(M+N) scaling — GormRegistry, SessionResolver infrastructure, and core-class tests (consolidates #15779, #15780, #15790)

borinquenkid:
# feat: GORM O(M+N) scaling — GormRegistry, SessionResolver infrastructure, and core-class tests (consolidates #15779, #15780, #15790)

## Why one PR

The previous 3-PR stack (#15779 infra → #15780 implementation → #15790 tests) generated review churn because the infrastructure PR added public API whose callers lived one PR downstream, so "no caller in this PR" objections and "the tests are elsewhere" objections could not both be answered at once. Per discussion with @jdaugherty, the stack is consolidated into this single PR: every new API lands next to its consumer and its tests.

Supersedes and closes #15779, #15780, #15790.

## Summary

Extracts all per-entity, per-qualifier GORM API state out of `GormEnhancer` into a `GormRegistry` singleton. `GormEnhancer` becomes a thin facade and delegates entity registration, API lookup, and datastore lifecycle to the registry. APIs are created by a pluggable `GormApiFactory` and looked up by `(entityClass, qualifier)` at call time — collapsing up-front API allocation from O(entityCount × tenantCount) to O(entityCount + tenantCount), with per-(entity, qualifier) APIs materialized lazily on first use.

Modules touched: `grails-datamapping-core` (registry + enhancer, bulk of the diff), `grails-datastore-core` (session-resolution infrastructure below), Hibernate 5/7, MongoDB, Simple adapters (minimal wiring; the full adapter migrations remain follow-up PRs), plus TCK/test-example updates.

### grails-datastore-core infrastructure (formerly #15779)

- `SessionResolver` + `TransactionSynchronizationSessionResolver`: a stateless view over the existing `SessionHolder`/`TransactionSynchronizationManager` state — one authoritative session store, no parallel bookkeeping. `resolve()` performs the same validation housekeeping as `doGetSession` (evicts disconnected sessions, unbinds a holder emptied by eviction unless a transaction owns it).
- `AbstractDatastore`: lazy resolver accessor; `destroy()` closes thread-bound sessions via `DatastoreUtils.closeSession`, skipping holders owned by an active transaction; `hasCurrentSession()` now agrees with `getCurrentSession()` (validated-session semantics); publisher wiring no longer routes through the deprecated `getApplicationContext()`.
- `DatastoreUtils`: new `executeWithNewSession(..)` overloads (used by `GormStaticApi`); `execute`/`doWithSession` now stack via `bindNewSession` and clean up via the one canonical `unbindSession`, so a bound-but-empty holder can never fail a later bind.
- `SessionHolder.getSessions()` (used by `destroy()`), `AbstractConnectionSourceFactory` fallback-settings extraction, `MultipleConnectionSourceCapableDatastore` (used by `GormApiResolver`/`AbstractGormApi`/`GormRegistry`), and an `AstUtils.copyAnnotations` dedup guard needed by `ServiceTransformation` (with `AstUtilsSpec` coverage).

Note on `SessionResolver.bind()`/`unbind()`: `resolve()` is what core consumes (via `hasCurrentSession()`); `bind`/`unbind` complete the SPI contract that the per-adapter follow-up PRs implement (e.g. a Hibernate resolver bridging native `SessionFactory`-keyed bindings). They are specified, tested, and small; flagging rather than hiding that their production callers arrive with the adapter PRs.

## Review response (2026-07-29 round on #15779)

1. **Scope**: `Query.java`, `Service.groovy`, `DefaultServiceRegistrySpec`, the `AbstractPersistentEntity.getTenantId()` fallback + `isMultiTenant` swap, and the `createPropertyResolver` cast are reverted. The `CustomizableRollbackTransactionAttribute` copy-semantics change is split into its own PR (#16063, branch `fix/customizable-rollback-tx-attribute-copy`) with the reworked implementation per review (Spring copy constructors, label independence, no lazy-getter mutation of the source, `timeoutString` preserved, the GString-in-`.java` log fixes) and behavior-level tests through `GrailsTransactionTemplate`/`DefaultTransactionService`. The same lossy-copy-constructor pattern was also found in two sibling classes and fixed in follow-up PRs #16064 (`grails.gorm.transactions.GrailsTransactionAttribute`) and #16065 (`org.grails.transaction.GrailsTransactionAttribute`).
2. **API duplication**: the unbind sequence now exists exactly once (`DatastoreUtils.unbindSession`); `executeWithNewSession`, `execute`, `doWithSession`, and `TransactionSynchronizationSessionResolver.unbind()` all delegate to it (also fixing the TSM-key mismatch for sessions owned by a child datastore — regression test added). The `AbstractDatastore` event-publisher machinery (third publisher implementation + reflective listener registration) is removed entirely rather than moved: nothing in this stack or the adapter follow-ups calls it; concrete datastores keep publishing through their own `ConfigurableApplicationEventPublisher`, as today. `getApplicationEventPublisher()` stays null when no context is configured (no per-query event allocation for bare datastores — unchanged from 8.0.x).
3. **Speculative API**: removed as caller-less — `Datastore.getSessionResolver()` default method (allocated per call), `AbstractConnectionSourceFactory.createSettings(PropertyResolver)`, the 3-arg `ConnectionSourceSettingsBuilder` constructor.

Inline-comment items: constructor `this`-escape fixed (lazy resolver); `bind()` rejects a session owned by a different datastore (spec added); `resolve()` uses validated sessions with disconnected-session and empty-holder specs; resolver class renamed since it holds no `ThreadLocal`; interface generics dropped; `@author` tags use a real name; `unbind()` keeps its name (matching `DatastoreUtils.unbindSession`, to which it now delegates) with the close-on-unbind contract stated explicitly in the interface javadoc.

## Fixes from the post-consolidation contrarian review

An adversarial multi-agent review of the consolidated branch surfaced and fixed:

- `DatastoreUtils.execute`/`doWithSession` could throw `IllegalStateException("Already value bound")` when a bound-but-empty `SessionHolder` remained on the thread; they now stack via `bindNewSession` and release via `unbindSession` (spec added).
- `TransactionSynchronizationSessionResolver.resolve()` left an emptied holder bound (poisoning later binds) and returned null when a valid session sat beneath a stale one; it now resolves down the stack and unbinds an emptied non-transactional holder (specs added, including the transaction-owned case).
- `GormStaticApi.withStatelessSession` had been rewired through `executeWithNewSession`, silently handing out **stateful** sessions and dropping the `UnsupportedOperationException` guard; baseline `connectStateless()` behavior restored.
- `GormStaticApi.saveAll` force-flushed mid-transaction, deviating from the 8.0.x baseline it claimed to restore; the flush is removed.
- `TenantDelegatingGormOperations.delete(instance, params)` called `save` instead of `delete` — a pre-existing 8.0.x data-integrity bug in a touched class; fixed with a delegation spec that would have caught it.
- Cross-datastore tenant bleed: three selector paths in `GormApiResolver` read the tenant via no-arg `CurrentTenantHolder.get()` (an arbitrary datastore's tenant); they now pass the datastore being evaluated.
- Registry races: `registerEntityDatastores` rebuilt an entity's routing map with remove-then-repopulate (a concurrent lookup could observe no routing and fall back to the wrong datastore) — it now publishes the rebuilt map atomically; `AbstractGormApiRegistry.getDirect` could cache a qualified API derived from a superseded default API indefinitely — it now re-validates after publishing and retracts if superseded.
- Memory: `GormRegistry.removeDatastore` now clears the normalization caches (`Class`-keyed keys retain classloaders across dev reloads; the qualifier cache grew per tenant-ID ever seen) once the last datastore is gone.
- The `ActiveSessionDatastoreSelector` ≤10-datastore fallback scan is now documented as the only discovery path for non-`Datastore`-keyed sessions (e.g. Hibernate's `SessionFactory` keying) and as intentionally bounded.

### Known limitations (deliberate, documented)

- Steady-state memory is still O(entities × qualifiers-actually-used) — lazily materialized, never eagerly allocated. The eager win is allocation/startup, not asymptotic worst-case residency.
- Beyond 10 registered datastores, unbound non-transactional Hibernate sessions are not discovered by the selector fallback (routing falls back to the entity's DEFAULT datastore); deployments at that scale should bind sessions explicitly (transactions / `Tenants.withId`).
- The six per-adapter follow-up branches predate this consolidation's resolver rework (renamed class, no setter, private field) and must be rebased onto it before review.

## Test plan

- [x] `./gradlew :grails-datastore-core:test :grails-datamapping-core:test :grails-data-simple:test` — 0 failures
- [x] `./gradlew :grails-data-hibernate5-core:test :grails-data-hibernate7-core:test :grails-data-mongodb-core:test` — 0 failures
- [x] `codeStyle` (Checkstyle + CodeNarc) clean


borinquenkid:

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
```diff
@@ -193,14 +211,15 @@ class Tenants {
...
+
     /**
      * Execute the given closure with given tenant id
      * @param tenantId The tenant id
      * @param callable The closure
      * @return The result of the closure
      */
-    static <T> T withId(Class<? extends Datastore> datastoreClass, Serializable tenantId, Closure<T> callable) {
-        Datastore datastore = GormEnhancer.findDatastoreByType(datastoreClass)
+    static <T> T withId(Class domainClass, Serializable tenantId, Closure<T> callable) {
```
jdaugherty, line 229: This replaces `withId(Class<? extends Datastore> datastoreClass, Serializable, Closure)` with `withId(Class domainClass, Serializable, Closure)`. Both erase to `withId(Class, Serializable, Closure)`, so every existing caller that passes a *datastore* class keeps compiling and silently changes meaning — it now goes through `datastoreLocator.getDatastoreForDomain(..)` and resolves the datastore class as if it were a domain class.

There are already ~40 call sites of exactly that shape in-tree: `RxGormStaticApi:596` (`Tenants.withId((Class<RxDatastoreClient>) datastoreClient.getClass(), tenantId) { .. }`) and `TenantDelegatingRxGormOperations`, which pass `datastoreClientClass`. `grails-datamapping-rx` is commented out in `settings.gradle`, so CI never compiles or runs them and won't catch this.

Please keep the datastore-class overload and give the domain-class variant a distinct name (`withIdForDomain`, or take the entity as a typed parameter). A silent same-erasure meaning change on a documented public multi-tenancy entry point isn't something users can be expected to notice.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormStaticApi.groovy
```diff
@@ -19,1019 +19,719 @@
...
+
+    @Override
+    Number deleteAll() {
+        execute({ Session session ->
+            session.deleteAll(new DetachedCriteria(persistentClass))
+        } as SessionCallback<Number>)
+    }
+
+    @Override
+    Number deleteAll(Map params) {
```
jdaugherty, line 523: `deleteAll(Map params)` discards `params` entirely and delegates to the unqualified `deleteAll()`. The interface javadoc added in `GormStaticOperations` says "Deletes all objects for the given arguments", so `Book.deleteAll(flush: true)` — or anything a user reasonably writes to *narrow* the delete — silently deletes every row instead.

Separately, I'd like explicit agreement on adding `deleteAll()`/`deleteAll(Map)` to `GormStaticOperations` and `GormEntity` at all. A no-argument "delete every row of this table" static on every domain class is a significant and destructive addition to the entity API surface; it isn't mentioned in the PR description and it isn't needed by the registry refactor. Either drop it from this PR and propose it on its own, or implement `params` properly (at minimum `flush`) and document both methods in `grails-doc`.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormEnhancer.groovy
```diff
@@ -540,162 +253,139 @@ class GormEnhancer implements Closeable {
...
-    @CompileStatic
-    List<FinderMethod> getFinders() {
-        if (finders == null) {
-            finders = Collections.unmodifiableList(createDynamicFinders())
+    @CompileDynamic
+    protected void addStaticMethods(PersistentEntity e) {
+        def cls = e.javaClass
+        ExpandoMetaClass mc = MetaClassUtils.getExpandoMetaClass(cls)
+        
+        mc.static.methodMissing = { String name, args ->
```
jdaugherty, line 290: `registerEntity` now calls `addStaticMethods`/`addInstanceMethods` unconditionally for every non-external entity, and those methods install catch-all `mc.static.methodMissing`, `mc.static.propertyMissing`, `mc.methodMissing` and both `mc.propertyMissing` variants on the entity's `ExpandoMetaClass`.

On 8.0.x this did not happen. `addStaticMethods`/`addInstanceMethods` were only reachable from `enhance(..)`, `enhance()` was gated on `dynamicEnhance`, the settings constructor hard-coded `this.dynamicEnhance = false`, and nothing in the tree calls `enhance(..)`. So this is new global metaclass mutation on the bootstrap path, not a like-for-like replacement.

Three consequences I'd like addressed:

- Assigning `mc.methodMissing` / `mc.propertyMissing` **overwrites** any handler an application or another plugin installed on that domain class. There's no merge or delegation to a previous handler.
- `GormEntity` already declares `methodMissing`, `propertyMissing` and `staticPropertyMissing` as real trait methods. An EMC closure takes precedence, so for Groovy entities the trait implementations become unreachable and the two dispatch paths can diverge.
- It forces an `ExpandoMetaClass` for every persistent entity at startup, and every dynamic-finder call now pays a metaclass miss plus `registry.findStaticApi(cls, null)` plus a dynamic `api.invokeMethod(..)`. `methodMissing` results aren't installed into the metaclass the way `registerInstanceMethod` was, so that cost is per call, not once. For a PR whose headline is scaling, please include numbers for the steady-state dynamic-finder path, not just startup allocation.

If the goal is only to stop eagerly allocating APIs, this metaclass change looks separable from that and I'd rather see it justified on its own.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/core/AbstractDatastore.java
```diff
@@ -170,8 +215,13 @@ public Session getCurrentSession() throws ConnectionNotFoundException {
...
     }
 
+    /**
+     * @return Whether {@link #getCurrentSession()} would return an existing session rather than
+     * opening a new one: a validated (still connected) session is bound to the current thread.
+     * Delegates to {@link #getSessionResolver()} so both methods read the same state.
+     */
     public boolean hasCurrentSession() {
-        return TransactionSynchronizationManager.hasResource(this);
+        return getSessionResolver().resolve() != null;
```
jdaugherty: `hasCurrentSession()` goes from a pure `TransactionSynchronizationManager.hasResource(this)` check to `getSessionResolver().resolve() != null`, and `resolve()` mutates thread-bound state: it evicts disconnected sessions from the `SessionHolder` and calls `unbindResourceIfPossible(datastore)` when the holder ends up empty and isn't transaction-synchronized.

So a predicate now has side effects, and it's called from places that are purely interrogative — `DatastoreUtils.execute`/`doWithSession`, and `ActiveSessionDatastoreSelector`, which calls `hasCurrentSession()` on **every** registered datastore (up to the documented 10) while deciding where to route. A routing probe for entity A can now unbind datastore B's holder as a side effect.

I'd like the validating/evicting behaviour kept in `resolve()` and out of `hasCurrentSession()`, or a separate non-mutating query used for the routing scan. If agreeing with `getCurrentSession()` really requires validation here, please say so explicitly in the javadoc — "this method may unbind thread-bound resources" — and add a spec covering the discovery-scan case.

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/tests/FirstAndLastMethodSpec.groovy
```diff
@@ -163,7 +163,7 @@ class FirstAndLastMethodSpec extends GrailsDataTckSpec {
     }
 
     @PendingFeatureIf(
-            value = { System.getProperty('hibernate5.gorm.suite') },
+            value = { System.getProperty('hibernate5.gorm.suite') || System.getProperty('hibernate7.gorm.suite') },
```
jdaugherty: This widens the `@PendingFeatureIf` from hibernate5 to hibernate5-or-hibernate7, i.e. a TCK test that passes on 8.0.x for Hibernate 7 is now excluded. The `reason` still reads "Was previously @Ignore", which no longer describes what's happening.

This looks like a direct consequence of removing the `first(Map)`/`last(Map)` overrides from `HibernateGormStaticApi` in the Hibernate 7 module, which sends composite-key `first`/`last` through the generic `GormStaticApi` implementation instead. Please fix the regression rather than widen the exclusion — the TCK is our contract for adapter parity, and quietly moving a passing test into pending hides the behaviour change from anyone reading the PR.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApi.groovy
```diff
@@ -20,76 +20,188 @@ package org.grails.datastore.gorm
...
+                // match the connection name instead of the real tenant.
+                boolean isConnectionQualifier = false
+                if (ds instanceof MultipleConnectionSourceCapableDatastore) {
+                    try {
+                        Datastore resolved = ((MultipleConnectionSourceCapableDatastore) ds)
+                                .getDatastoreForConnection(currentQualifier)
+                        if (resolved != null) {
+                            isConnectionQualifier = true
+                        }
+                    } catch (Exception ignored) {
```
jdaugherty: This uses an exception as control flow to decide whether the qualifier is a connection name or a tenant id, on **every** `execute()` call for a multi-tenant entity with a non-default qualifier. `getDatastoreForConnection` throws `ConfigurationException` for an unknown name (see `SimpleMapDatastore`, and Hibernate's equivalent), so in DISCRIMINATOR mode — where the qualifier is always a tenant id — every single query constructs and fills in a stack trace before proceeding.

That's a per-operation cost in the hot path of a performance PR. Please add a non-throwing lookup for this (`hasDatastoreForConnection(name)` on `MultipleConnectionSourceCapableDatastore`, or check the connection-source names) instead of probing with exceptions. It also removes the need for the bare `catch (Exception ignored)`.

Related, at line 145: `executeQualified(tenantId.toString(), callback)` flattens the tenant id to a string to use it as a qualifier. That's lossy for non-String tenant ids — `1L` and `1` (and any type whose `toString()` collides) become the same qualifier and therefore the same cached API. Please key on the tenant id itself, or document the constraint that tenant ids must be uniquely representable as strings.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
```diff
@@ -0,0 +1,992 @@
...
+    Datastore getDatastoreDirect(String normalizedClassName, String normalizedQualifier) {
+        if (normalizedClassName != null) {
+            Map<String, Datastore> mappedDatastores = entityDatastores.get(normalizedClassName)
+            if (mappedDatastores != null) {
+                Datastore ds = mappedDatastores.get(normalizedQualifier)
+                if (ds != null) {
+                    return ds
+                }
+                if (ConnectionSource.DEFAULT.equals(normalizedQualifier) && !mappedDatastores.isEmpty()) {
+                    return mappedDatastores.values().iterator().next()
```
jdaugherty: `return mappedDatastores.values().iterator().next()` picks an arbitrary datastore. `mappedDatastores` is a `ConcurrentHashMap`, so iteration order is unspecified and can change with the map's internal state.

For an entity mapped only to non-default connections (e.g. `datasource 'analytics'` plus `datasource 'reporting'`) with no DEFAULT entry, an unqualified operation therefore resolves to whichever of the two the iterator happens to hand back. That's a silent wrong-database read or write, and it can differ between runs of the same application.

This needs to be deterministic: pick the entity's first *declared* connection source (`ConnectionSourcesSupport.getConnectionSourceNames(entity)` order) and store the routing explicitly at registration time, or fail loudly. Same pattern needs fixing in `CurrentTenantHolder.get()`.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/CurrentTenantHolder.groovy
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /**
+     * @return Obtain the current tenant (fallback for any datastore)
+     */
+    static Serializable get() {
+        def map = currentTenantThreadLocal.get()
+        if (!map.isEmpty()) {
+            return map.values().iterator().next()
```
jdaugherty: The no-argument `get()` returns `map.values().iterator().next()` from a `HashMap`, i.e. an arbitrary tenant id when more than one datastore has a tenant bound on this thread. The PR description says the datastore-blind lookups in `GormApiResolver` were fixed to pass the datastore being evaluated — but this method survives as *public* API on a class in the `grails.gorm.multitenancy` package, so applications will find it and use it.

Either drop it (all in-tree callers can pass a datastore) or make it total: throw when more than one binding exists rather than returning a nondeterministic one. Returning "some tenant" from a multi-tenant API is the kind of thing that produces cross-tenant data leaks that only show up under load.

While you're here: this whole class is new public API in a `grails.*` package and needs a `grails-doc` section.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
```diff
@@ -0,0 +1,992 @@
...
+            }
+        }
+    }
+
+    /**
+     * Create a GormStaticApi instance.
+     * Uses a bound resolver (always returns the given datastore) at registration time so that
+     * tenant-resolving DatastoreResolvers are not invoked before any tenant context is active.
+     */
+    GormStaticApi createStaticApi(Class cls, Datastore datastore, DatastoreResolver resolver, String qualifier) {
```
jdaugherty: `createStaticApi`, `createInstanceApi` and `createValidationApi` all accept a `DatastoreResolver resolver` parameter and then ignore it, building `boundResolver` from `datastore` instead. `registerEntity` duly constructs a resolver via `createClassDatastoreResolver(cls)` and passes it into all three, where it's discarded — so that call, and the `normalizedClassName` local inside `createClassDatastoreResolver` (also unused), are dead.

If binding to the concrete datastore at registration time is the intended semantics — and the comment says it is — then drop the parameter. Leaving it in signatures that are described as public API for external plugins is a trap: a caller passing a tenant-aware resolver gets no error and no effect.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
```diff
@@ -0,0 +1,992 @@
...
+    }
+
+    /**
+     * Register constraints for all entities in a datastore.
+     * Delegates to the ConstraintsEvaluator if available in the mapping context.
+     *
+     * @param datastore The datastore containing the entities
+     */
+    @CompileDynamic
+    void registerConstraints(Object datastore) {
```
jdaugherty: This adds a second constraint-registration path. `GormEnhancer`'s constructor calls `registry.initializeDatastore(datastore, qualifier)` — which lands here and runs `ConstraintsEvaluator.evaluate(entity.javaClass)` for every persistent entity — and then calls `registerConstraints(datastore)`, which is the pre-existing `ConstraintRegistrar` path. So on a normal bootstrap constraints get evaluated twice, and the extra pass is O(entities) work added by a PR about reducing startup cost.

Both this method and the rewritten `GormEnhancer.removeConstraints()` also reach the evaluator by duck-typing (`factory.hasProperty('entityContext')`, `Class.forName(..)`) inside `catch (Throwable) { log.debug(..) }`. If the shape ever changes, constraint registration silently becomes a no-op and nothing fails. Note `Class.forName` never returns null, so the `if (cls != null)` guard in `removeConstraints` is dead.

Separately: `HibernateGormEnhancer`/`MongoGormEnhancer` now override `registerConstraints(Datastore)` to call `registry.registerApiFactory(..)`. Using the constraints hook to register API factories is going to confuse the next person, and it couples factory registration to a lifecycle step that has nothing to do with it — if a subclass ever skips or reorders `registerConstraints`, Mongo's `(MongoStaticApi) findStaticApi(..)` cast in `MongoEntity` breaks at runtime. Please give factory registration its own hook.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
```diff
@@ -254,39 +286,76 @@ class Tenants {
...
+            }
+        }
+        return CurrentTenantHolder.withTenant(multiTenantCapableDatastore, tenantId) {
             if (multiTenantCapableDatastore.getMultiTenancyMode().isSharedConnection()) {
                 def i = callable.parameterTypes.length
                 if (i == 2) {
                     return multiTenantCapableDatastore.withSession { session ->
-                        return callable.call(tenantId, session)
+                        def result = callable.call(tenantId, session)
+                        log.debug('Result from shared connection with 2 args: {}', result)
```
jdaugherty: These look like leftover debugging scaffolding rather than logging we want to ship: six near-identical `log.debug('Result from ... with N args: {}', result)` statements whose only purpose is to log the closure's return value on the way past, plus `log.debug('Inside withNewSession for tenantId {}', ..)`. They also restructure the switch arms (introducing a `result` local in each `case`) purely to make the logging possible.

The same pattern is in `GrailsTransactionTemplate.executeAndRollback` (lines 77-96), where it's worse: the messages are duplicated (`'... - finished action'` immediately followed by `'... action finished. status='`, and the same for the caught-exception and setRollbackOnly pairs), and they use string concatenation inside an `isDebugEnabled()` guard instead of SLF4J placeholders.

Please strip both back. If any of it is genuinely useful diagnostics, keep one parameterized statement per branch and drop the `isDebugEnabled` wrappers.

At line 293 in this file: `catch (Throwable e)` around `getDatastoreForTenantId(tenantId)` logging at debug is too broad — it swallows `Error`s, and a genuine misconfiguration (unknown tenant, missing connection) becomes invisible at default log levels. Catch the specific exception and log at warn.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/gorm/mongo/api/MongoStaticApi.groovy
```diff
@@ -63,8 +64,15 @@ import org.grails.datastore.mapping.multitenancy.MultiTenancySettings
 @CompileStatic
 class MongoStaticApi<D> extends GormStaticApi<D> implements MongoAllOperations<D> {
 
+    protected final PersistentEntity persistentEntity
```
jdaugherty: `protected final PersistentEntity persistentEntity` here shadows `AbstractGormApi.persistentEntity`, which this PR deprecates but keeps and keeps assigning in the `AbstractGormApi` constructor. There are now two fields with the same name on the same object holding independently-computed values: base-class code sees the base field, `MongoStaticApi` sees its own. That's a latent divergence with no compiler warning.

I realise the driver is that `GormStaticApi.multiTenancyMode` was removed and had to be re-declared here. Two asks:

1. Don't shadow. Use `getGormPersistentEntity()` (the replacement you added) in `wrapFilterWithMultiTenancy` and drop the field.
2. Removing `protected final multiTenancyMode` from `GormStaticApi` is a break for any out-of-tree `GormStaticApi` subclass. If it has to go, that needs an upgrade note; if it doesn't, keeping it avoids this whole re-declaration.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApi.groovy
```diff
@@ -20,76 +20,188 @@ package org.grails.datastore.gorm
...
-        'getExtendedMethods',
-        'wait',
-        'equals',
-        'toString',
-        'hashCode',
-        'getClass',
-        'notify',
-        'notifyAll',
-        'setTransactionManager'
+    protected static final List<String> EXCLUDES = [
```
jdaugherty: `EXCLUDES` quietly lost three entries — `getMethods`, `getExtendedMethods` and `setTransactionManager` — and the PR simultaneously adds `abstract PlatformTransactionManager getTransactionManager()` to this class. So `getMethods()` now reports `getMethods`, `getExtendedMethods` and `getTransactionManager` as part of each API's method catalogue.

The new `AbstractGormApiSpec` can't catch that, because it asserts `!methods.any { it.name in AbstractGormApi.EXCLUDES }` — it checks the filter against the very constant that drives the filter, so it passes for any content of `EXCLUDES`, including an empty list. Please assert the specific names that must never be exposed (`getClass`, `getMetaClass`, `invokeMethod`, `getMethods`, `getTransactionManager`, …) as literals.

Also, the visibility change from an implicit-public Groovy property to `protected static final` removes the generated accessor, which is a source/binary break for anything reading `AbstractGormApi.EXCLUDES` outside the hierarchy.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/core/connections/MultipleConnectionSourceCapableDatastore.java
```diff
@@ -0,0 +1,56 @@
...
+     * Real datastores manage an independent session per connection, so such an entity's default
+     * operations belong to that mapped connection's datastore. The in-memory mock used for unit
+     * testing manages a single session on this (parent) datastore and only fabricates isolated
+     * children for explicit connection access, so it overrides this to keep unqualified operations
+     * on the parent — otherwise they would target a child whose session the test harness never
+     * flushes.
+     *
+     * @return {@code true} to route unqualified operations to the mapped connection datastore
+     */
+    default boolean routesUnqualifiedToMappedConnection() {
```
jdaugherty: This adds a method to a public SPI whose javadoc defines the contract in terms of a test double: "The in-memory mock used for unit testing manages a single session on this (parent) datastore … otherwise they would target a child whose session the test harness never flushes." `SimpleMapDatastore` is the only override, and it returns `false` for exactly that reason.

Production SPI shouldn't exist to accommodate the unit-test datastore's session model. Either fix `SimpleMapDatastore` so its per-connection children behave like real ones (the harness flushes what it creates), or keep the special-casing inside `GormRegistry`/`SimpleMapDatastore` rather than putting it on the interface every third-party GORM implementation has to reason about.

If it does stay, it's a new public interface method and needs documenting, and the javadoc should describe the semantics abstractly rather than referring to our test harness.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateSession.java
```diff
@@ -389,30 +392,71 @@ public List retrieveAll(final Class type, final Iterable keys) {
...
         final String idName = persistentEntity.getIdentity().getName();
         final String hql = "from " + entityName + " as e where e." + idName + " in (:keys)";
+        final Class idType = persistentEntity.getIdentity().getType();
+        final ConversionService conversionService = getMappingContext().getConversionService();
+
+        // Convert each requested id to the entity's identifier type, preserving order and
+        // duplicates. getAll() must return entities in the supplied id order with a null slot
+        // for any id that does not resolve to a row, so order is driven by the request rather
+        // than the database.
+        final List<Serializable> requestedIds = new ArrayList<>();
```
jdaugherty, line 402: This changes `getAll()` semantics for Hibernate 7 (and the same change is made in Hibernate 5): results are now reassembled in requested-id order, duplicate ids are preserved, and unresolvable ids produce `null` slots instead of being absent. That's a real improvement over returning whatever order the database chose — but it's a user-visible behaviour change with nothing to do with the registry refactor, and there's no TCK coverage asserting the new contract.

Please pull it into its own PR with TCK tests (order preserved, duplicates preserved, missing id yields a `null` at the right index, empty input). The same goes for the `HibernateQuery` `disjunction()`/`conjunction()`/`negation()` overrides added in this PR — the bug you describe there (a factory junction added to the unused base `criteria` field, so `countByXOrY` loses its disjunction) sounds like a genuine data-correctness bug that deserves its own PR, its own test, and a backport decision.

One concrete thing to check in this implementation: `entitiesById` is keyed by `session.getIdentifier(entity)` but looked up by the converted `requestedId`. `convertToIdentifierType` falls back to returning the raw key when the `ConversionService` can't convert it, so an id that Hibernate itself would have coerced now misses the map and silently yields a `null` slot instead of the entity. Previously the raw key went into the query and Hibernate did the coercion.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApiRegistry.groovy
```diff
@@ -0,0 +1,179 @@
...
+
+    void removeDatastore(Datastore datastore) {
+        if (datastore == null) return
+        Iterator<Map.Entry<String, T>> it = apis.entrySet().iterator()
+        while (it.hasNext()) {
+            try {
+                if (it.next().value.getDatastore() == datastore) {
+                    it.remove()
+                }
+            } catch (Exception e) {
```
jdaugherty: `catch (Exception e) { it.remove() }` — on any exception from `getDatastore()` this silently evicts the entity's registered API, and the same pattern repeats at line 160 for the qualified map. `e` is unused, so nothing is logged either.

That turns a transient failure into permanent loss of the entity's API registration: the next lookup finds nothing and either throws "No GORM implementation configured" or falls back to a different datastore. Given `getDatastore()` now routes through `datastoreResolver.resolve()`, which for a tenant-aware resolver can throw `TenantNotFoundException` when no tenant is bound, this is reachable from an ordinary unresolved-tenant call rather than only from a corrupt-state edge case.

Please only remove on a positive identity match, and log at warn (with `e`) when the datastore can't be read.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/AstUtils.groovy
```diff
@@ -366,8 +366,10 @@ class AstUtils {
             String annotationClassName = node.getClassNode().getName()
             if ((excluded == null || !excluded.contains(annotationClassName)) &&
                     (included == null || included.contains(annotationClassName))) {
-                final AnnotationNode copyOfAnnotationNode = cloneAnnotation(node)
-                to.addAnnotation(copyOfAnnotationNode)
+                if (to.getAnnotations(node.getClassNode()).isEmpty()) {
```
jdaugherty: This makes `copyAnnotations` skip any annotation already present on the target, for every caller of this shared utility. Repeatable annotations legitimately appear more than once, and for those the guard now silently drops all but the first — the annotated node ends up with an incomplete set and there's no diagnostic.

The PR describes this as a dedup guard needed by `ServiceTransformation`. Please scope it there rather than changing the semantics of a general-purpose AST helper — either dedup at the `ServiceTransformation` call site, or add a boolean/`Set<String>` parameter so callers opt in. If it must be global, exclude annotations meta-annotated `@Repeatable`.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
```diff
@@ -210,13 +229,26 @@ class Tenants {
         }
     }
 
+    /**
+     * Execute the given closure with given tenant id for the given datastore. This method will create a new datastore session for the scope of the call and hence is designed to be used to manage the connection life cycle
+     * @param tenantId The tenant id
+     * @param callable The closure
+     * @return The result of the closure
+     */
+    static <T> T withTenant(Class domainClass, Serializable tenantId, Closure<T> callable) {
```
jdaugherty, line 252: The javadoc says "This method will create a new datastore session for the scope of the call and hence is designed to be used to manage the connection life cycle", but the implementation only nests two `CurrentTenantHolder.withTenant` calls — it sets thread-locals and never opens a session. Same for `withTenant(Serializable, Closure)` at line 68.

That's the documented difference between `Tenants.withId` and this method, so users following the javadoc will assume connection management they aren't getting. Please either implement it (delegate to `withId` after resolving the datastore) or rewrite the javadoc to say plainly that it only binds the tenant to the current thread and that the caller owns the session — and explain in `grails-doc` when to reach for `withTenant` versus `withId`, because the names give no hint.

jdaugherty (changes requested):
Initial AI Review seems to have found a few issues (including breaking tests):

Thanks for consolidating the stack — having the API, its consumer and its tests in one place does make the registry design reviewable in a way the 3-PR split wasn't.

I've gone through the whole diff. The GormRegistry / GormApiFactory / DatastoreResolver core is a sound shape and the lazy (entity, qualifier) materialization is the right call. Most of my comments below are about things that came along for the ride and about behaviour changes that aren't called out in the description.

Three things I'd like resolved before this can move forward:

1. No documentation. 175 files and not one line in grails-doc. This PR adds public API (grails.gorm.multitenancy.CurrentTenantHolder, Tenants.withTenant(..), Tenants.withId(Class, ..), GormEntity.deleteAll()/deleteAll(Map), four new GormStaticOperations.deleteAll overloads, SessionResolver, MultipleConnectionSourceCapableDatastore.routesUnqualifiedToMappedConnection()) and removes public API (GormEnhancer.enhance(..), GormEnhancer.finders/transactionManager/dynamicEnhance, the 1-arg and 5-arg GormEnhancer constructors, GormStaticApi.multiTenancyMode, Tenants.withId(Class<? extends Datastore>, ..)). Every settable/callable thing we ship has to be documented, and the removals need upgrade notes — GormEnhancer is the extension point every out-of-tree GORM implementation subclasses, so "it's a major version" isn't sufficient on its own.

2. Unrelated fixes bundled in. getAll() ordering/null-slot semantics (both Hibernate 5 and 7), the HibernateQuery junction overrides, CriteriaBuilder.call(Closure) + the ensureQueryIsInitialized() additions, and the removeConstraints() rewrite are all independent behaviour changes. Several look correct and worth having, which is exactly why they should land as their own reviewable/revertable PRs rather than inside a 17k-line refactor. Please split them out.

3. A TCK test is being excluded rather than fixed — see the FirstAndLastMethodSpec comment.

CI being green is necessary but not sufficient here: grails-datamapping-rx is commented out of settings.gradle, so the Tenants.withId overload change isn't compiled anywhere, and the multi-tenant routing paths added in GormApiResolver are mostly covered by specs that drive the internal selectors directly rather than by an app-level scenario.

jdaugherty:
I used AI to take a pass at this here: https://github.com/apache/grails-core/tree/fix/gorm-registry-review-feedback , I don't think this fully addresses everything but it may be a starting point to save time.

borinquenkid:
thanks for the branch, I will compare to my in flight solutions, which I am doing with a tooth comb.

Walter

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
borinquenkid: Traced the actual blast radius before picking a fix. The ~40 call sites cited as evidence (`RxGormStaticApi`, `TenantDelegatingRxGormOperations`) turned out to be a false positive — `grails-datamapping-rx` has its own, fully independent `grails.gorm.rx.multitenancy.Tenants` operating on `RxDatastoreClient`, untouched by this PR (grepped every import in that module to confirm zero references to the core `Tenants` class).

The real clash is in `TenantDelegatingGormOperations`'s ~100 `Tenants.withId((Class<Datastore>) datastore.getClass(), ...)` calls, but that class turns out to be dead code on this branch: `GormStaticApi.withTenant(Serializable)` used to construct it directly, and this PR's rewrite replaced that with `forQualifier(tenantId.toString())` instead — nothing else in the tree constructs it except its own unit test. Fixed anyway since it's cheap and removes the landmine if it's ever wired back up: added `requireMultiTenantCapableDatastore()` and switched all 100 sites to the existing, unambiguous `Tenants.withId(MultiTenantCapableDatastore, Serializable, Closure)` overload rather than a `withIdForDomain` rename — the domain-class overload has no live erasure conflict anywhere reachable, so a rename felt like a needless public API break for a theoretical case. `Tenants.withId(Class domainClass, ...)` itself is untouched.

Commit `4b116c2166`. Also verified the `withTenant()` decorator→`forQualifier` mechanism swap doesn't change runtime behavior — `GormRegistrySpec`'s DISCRIMINATOR-mode tests and `PartitionedMultiTenancySpec`'s DATABASE-mode assertion both exercise the no-closure chained form against it and pass.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormStaticApi.groovy
borinquenkid: Took the implement option, not drop — this is your own stated bar, and your `fix/gorm-registry-review-feedback` branch already met it, so matched it rather than reinventing: `deleteAll(Map params)` now honors `params?.flush` instead of discarding params, both `deleteAll()`/`deleteAll(Map)` are documented in `grails-doc` (new 'Deleting Every Instance' section in `basicCRUD.adoc` with an explicit destructive-semantics warning and a pointer to `.where{}.deleteAll()` for partial deletes), and `TenantDelegatingGormOperations` gets matching delegators via its own `requireMultiTenantCapableDatastore()` pattern (see the `Tenants.withId` thread) rather than a class-based fallback.

Commit `e011e0896c`. 3 new specs in `GormStaticApiSpec` (basic delete-everything, `Map`-returns-count, flush-honored via a `Session` mock) + 2 in `TenantDelegatingGormOperationsSpec`.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormEnhancer.groovy
borinquenkid: Restored the `dynamicEnhance`-gated bootstrap exactly as it is on 8.0.x: `registerEntity` no longer calls `addStaticMethods`/`addInstanceMethods` unconditionally, `enhance()`/`enhance(entity)` are back and gated on `dynamicEnhance` (hard-coded `false` in the settings constructor, matching 8.0.x's own quirk), and nothing in the tree calls `enhance(..)` — confirmed by grepping the whole real 8.0.x tree, not just this PR's diff, for any caller that ever flips `dynamicEnhance` true. `GormEntity`'s own trait `methodMissing`/`propertyMissing` hooks already cover all dispatch (83 spec classes stayed green with zero `ExpandoMetaClass` installed). Commit `da09401c4d`.

Separately adopted your instance-side clobber guard on `addInstanceMethods` (backs off with a debug log if a `methodMissing` handler is already installed) — commit `2dd18808ff`. I did not adopt the static-side guard (`getStaticMetaMethod('methodMissing', ...)`): wrote a few isolated `ExpandoMetaClass` probe scripts (no GORM involved) and confirmed `getStaticMetaMethod` can never see a static `methodMissing` handler installed via the standard `mc.static.methodMissing = {...}` idiom, regardless of install order — even though the handler is functionally live (`invokeStaticMethod` proves it dispatches). Your branch has the identical static-side check; it looks like it never actually fires there either. Happy to share the repro scripts if useful — didn't want to ship a check that looks protective but is dead code.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/core/AbstractDatastore.java
borinquenkid: Agreed the mutation shouldn't live in a predicate. Added `SessionResolver.hasResolvedSession()` — implemented by iterating `SessionHolder.getSessions()` (the genuinely read-only `Collections.unmodifiableCollection` view) checking `isConnected()`, same accuracy as `resolve()` but zero mutation. `AbstractDatastore.hasCurrentSession()` now delegates to that instead of `resolve() != null`. Commit `3707a82f63`.

Audited all 8 call sites first: the 2 `ActiveSessionDatastoreSelector` routing-scan sites needed the pure check; the other 6 all immediately call `getCurrentSession()` right after anyway, which independently re-validates via its own `getValidatedSession()`, so none of them were actually relying on the old eviction side effect. Added a regression spec proving a `hasCurrentSession()` probe on datastore A leaves datastore B's session binding completely untouched (present, not evicted).

grails-datamapping-tck/src/main/groovy/org/apache/grails/data/testing/tck/tests/FirstAndLastMethodSpec.groovy
borinquenkid: Fixed the regression instead of widening the exclusion. Root cause: `GormStaticApi.first(Map)`/`last(Map)` (shared, not H7-specific — H5 lost the same override at merge-base too, both now route through the generic method) force-applies `max:1`/`order` at the DB level even when a composite-key entity has no derivable identity to sort by, so it returned an arbitrary row instead of falling back to natural/insertion order. Fixed generically: only force max/order when a sort key exists (user-supplied or derived from a simple identity); otherwise fetch normally and index `[0]`/`[-1]`, matching what the removed H7-specific override used to do. Reverted the `@PendingFeatureIf` back to its original H5-only condition — H7's composite-key case now passes 12/12; H5's own separate pre-existing skip is correctly untouched. Commit `da09401c4d`.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApi.groovy
borinquenkid: Both fixed. Exception-as-control-flow: added `ConnectionSourceNameResolver.isConnectionSourceName(datastore, name)` — checks the datastore's declared connection-source names directly, no throwing — and swapped the try/catch for that non-throwing call. Commit `b07c5d7f95`, +5 new specs on `ConnectionSourceNameResolverSpec`.

On the `tenantId.toString()` flattening: traced the actual data flow through `GormRegistry.resolveStaticApi` before deciding what 'fixed' should mean here. In the hot-path DISCRIMINATOR-mode case, that flattened string never resolves to a real registered api (both priority-1/2 lookups miss), falls through to the entity's own DEFAULT api, and `executeQualified`'s `qualifiedApi != this` check short-circuits before ever re-entering with the flattened string — so in that mode it's a wasted `String` allocation, not a live correctness bug. The `1L` vs `"1"` collision risk you named is real, but only manifests in DATABASE/SCHEMA mode, where it's a property of `GormRegistry`'s entire String-keyed per-tenant registration surface (`entityDatastores`, `staticApiRegistry`, `instanceApiRegistry`, `validationApiRegistry`), not something unique to this one call site — widening key types across that whole surface is a materially bigger change than this item's scope. Documented the constraint in a comment at the call site instead. For what it's worth, your own fix branch left the identical line unchanged too, which is part of why I read this as genuinely out of scope for a targeted fix here rather than something I was skipping.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
borinquenkid: Traced this to the removal path specifically — registration already deterministically tracks the entity's first declared connection and assigns DEFAULT correctly; the arbitrary pick only happened after `removeDatastore`/`removeEntityDatastore` stripped an entry and left the map non-empty but DEFAULT-less. Reproduced first as a failing test (registered an entity on 3 declared connections, removed the one DEFAULT pointed at, asserted DEFAULT re-points at the next declared one — failed against unfixed code, landing on whichever entry `ConcurrentHashMap`'s hash order happened to hand back). Fixed with `entityConnectionOrder` (a declared-order list per entity) plus a `repairDefaultRouting` helper called from both removal paths. Commit `63c2c20021`.

Also removed the `mappedDatastores.values().iterator().next()` fallback itself as defense-in-depth on top of the repair (same change your branch made) — any not-yet-found edge case now fails loud instead of silently routing to a wrong datastore. Commit `b81bda3f73`. One thing I checked: your branch's version of this fix doesn't actually repair the removal-path bug — deleting the fallback alone just turns it into DEFAULT resolving to `null` post-removal rather than a wrong-but-nonnull datastore, which is a harder failure, not a fix for the underlying issue. Kept both pieces (the repair mechanism plus the loud-failure fallback) rather than just the latter.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/CurrentTenantHolder.groovy
borinquenkid: Made it total rather than dropping it — grepped every module (grails-datamapping-core/H5/H7/Mongo/Simple/Neo4j/datastore-core/rx/tck) for production callers of the no-arg `get()` and found none, only 2 test files, confirming in-tree callers can all pass a datastore already. `get()` now collects the distinct bound tenant values; throws `TenantException` (existing type, not new) naming the ambiguous tenants when more than one is bound, returns the single value otherwise, or `null` if nothing's bound — same-tenant-bound-to-multiple-datastores correctly does not throw. Commit `4909abfc67`. Reproduced the bug as a failing test first, before implementing. This matches your branch's fix for the same comment.

Also added the `grails-doc` coverage you asked for — this whole class is now documented in the new multiTenancy.adoc page.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
borinquenkid: Confirmed dead and removed — traced every call site first. `createStaticApi`/`createInstanceApi`/`createValidationApi` all discarded the passed-in `resolver` in favor of their own `boundResolver` closure; the `resolver` built in `registerEntity` and the `normalizedClassName` local inside `createClassDatastoreResolver` were both genuinely unused. `createClassDatastoreResolver` itself stays — it has real callers elsewhere (`GormStaticApiRegistry`, `GormInstanceApiRegistry`, `GormInstanceApi`, `GormValidationApiRegistry`) for actual tenant-resolving `qualify()` lookups. Commit `709d72a574`, matches your branch, no conflict. Pure dead-parameter removal with zero observable behavior change, so no new tests — existing `registerEntity` coverage across the whole suite already exercises the path.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/GormRegistry.groovy
borinquenkid: Both problems fixed, traced against the real 8.0.x baseline rather than assumed. `GormRegistry.registerConstraints(Object)` (the duck-typed `ConstraintsEvaluator` path) has no 8.0.x basis at all — `GormRegistry` doesn't exist there — and duplicates work the real `ConstraintRegistrar` path (`GormEnhancer.registerConstraints(Datastore)`) already does. Deleted entirely, including the dead `if (cls != null)` guard on `Class.forName` you flagged.

Also found `GormEnhancer.removeConstraints()` hadn't actually been restored to 8.0.x's real implementation (`ConstrainedProperty.removeConstraint('unique')`) in an earlier pass of this work — it was still the PR-invented rewrite, verified byte-for-byte against 8.0.x. Restored the real logic, but moved it onto `GormRegistry.removeConstraints()` (called from `removeDatastore()`) rather than back onto `GormEnhancer` — `removeDatastore()` already owns every other piece of datastore teardown (API deregistration, datastore mapping cleanup), so this fits there. Register-time `registerConstraints` stays on `GormEnhancer`, matching both 8.0.x and your branch; only the remove side moved.

On the API-factory/constraints-hook coupling: agreed and adopted your fix — added a dedicated `registerApiFactories()` hook on `GormEnhancer`, called from the constructor before `registerConstraints`; H5/H7/Mongo now register their factories there instead of piggybacking on the constraints override. Commit `d49ee5992d`.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
borinquenkid: Stripped it back — `Tenants.withId` down to plain `return callable.call(...)` per switch arm, keeping one informative entry log at the top of the method; `GrailsTransactionTemplate.executeAndRollback` loses the duplicate-message `isDebugEnabled()` blocks, kept a single parameterized `log.debug('Rolling back after the action threw', e)`. Commit `508cce361e`, matches your branch's shape here.

Narrowed the broad `catch (Throwable e)` around `getDatastoreForTenantId` to `catch (ConfigurationException | TenantException e)`, bumped to warn — checked the real adapter code first: H5/H7/Mongo's `getDatastoreForConnection` throws exactly `ConfigurationException` for an unknown name, and `TenantNotFoundException extends TenantException`, so the narrowed catch covers the real failure modes rather than a guess. This did break an existing test that used a generic `IllegalStateException` as its simulated failure (only worked under the old unconditional catch) — replaced with 3 tests using the real exception types, plus one proving a genuinely unrelated exception now correctly propagates instead of being silently swallowed.

grails-data-mongodb/core/src/main/groovy/org/grails/datastore/gorm/mongo/api/MongoStaticApi.groovy
borinquenkid: Took your option 1 — kept `GormStaticApi.multiTenancyMode` as the field rather than removing it, so no upgrade note or re-declaration is needed at all. Dropped `MongoStaticApi`'s shadowing `persistentEntity`/`multiTenancyMode` fields entirely; `wrapFilterWithMultiTenancy`/`preparePipeline` now use `getGormPersistentEntity()` and the inherited field.

One thing worth flagging on your own branch while I was here: your replacement `getMultiTenancyMode()` is commented 'an API is no longer bound to a single datastore for its lifetime.' I traced `GormRegistry.createStaticApi`'s `boundResolver = { datastore } as DatastoreResolver` closure on both branches (yours included — the binding mechanism itself is unchanged there, you only dropped the now-unused `resolver` param) and every `GormStaticApi` is built through that one path on both. An instance genuinely is pinned to one datastore for its whole lifetime; when a datastore changes, the registry replaces the API instance wholesale rather than mutating one in place. The rationale in that comment doesn't hold against your own diff — worth a look, might be worth just keeping the field on your branch too.

Also found and fixed a real, separate bug while restoring this: `MongoStaticApi`'s only constructor routed through `GormStaticApi`'s deprecated 4-arg super, which eagerly resolved the datastore and discarded whatever qualifier the registry actually asked for — every Mongo entity's API silently hardcoded `ConnectionSource.DEFAULT` regardless of its real qualifier. Added a primary constructor mirroring `GormStaticApi`'s non-deprecated signature; `MongoGormApiFactory` now calls that one directly. Commit `da09401c4d`.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApi.groovy
borinquenkid: Fixed, just now — you're right on all counts. Restored `getMethods`/`getExtendedMethods`/`setTransactionManager` (verified byte-for-byte against real 8.0.x), added `getTransactionManager` since that abstract getter is new on this branch and needed the same treatment as the setter, and reverted the field back to `static final` with no explicit visibility modifier so Groovy re-generates the public static accessor — `protected` had silently dropped it, exactly as you called out. Fixed `AbstractGormApiSpec`'s circular assertion too, replaced with a check against the literal excluded names, plus a dedicated new test asserting `EXCLUDES`' exact contents and the restored accessor. Commit `c9c57e6a21`.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/core/connections/MultipleConnectionSourceCapableDatastore.java
borinquenkid: Removed it from the public SPI entirely rather than fixing `SimpleMapDatastore`'s session model — agreed it shouldn't be third-party datastores' problem to reason about. Added a narrow `SingleSessionCapableDatastore` marker interface in `grails-datastore-core` (javadoc explicitly says it's not part of the public multi-connection contract, don't implement it for production datastores); `SimpleMapDatastore` implements it instead of overriding a method on the shared interface. `GormRegistry`'s one call site collapsed from a double-negative instanceof+method check to a single `instanceof SingleSessionCapableDatastore`. Commit `da09401c4d`.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateSession.java
borinquenkid: Kept these rather than splitting them out — traced each one against real git blame/TCK history before deciding, rather than against just the PR description's framing:

- `getAll()` order/null-slot semantics: the shared TCK's `GormEnhancerSpec` (zero diff from 8.0.x) already asserts 'Test getAll preserves the supplied id order' and 'returns a null slot for a missing id' — the old H5/H7 code was structurally incapable of passing either. This is a pre-existing 8.0.x contract violation the registry work surfaced and fixed, not new undocumented behavior — the coverage already existed, it just hadn't been cross-referenced against an already-large diff. I did fix the specific gap you named, though: `entitiesById` is now consistently keyed/looked-up by `String.valueOf(id)` on both sides, so a raw `requestedId` that `convertToIdentifierType` couldn't convert (but Hibernate itself would have coerced) now resolves instead of silently yielding a null slot. Applied to both H5 and H7.
- `HibernateQuery`'s `disjunction()`/`conjunction()`/`negation()`: small, self-contained, a genuine correctness bug — core's base implementations wrote to an unused field, silently dropping `countByXOrY`'s OR. Reverting would reintroduce silently-wrong query results. It didn't have a dedicated test before this; happy to add one if that's what would unblock keeping it here instead of splitting it out.
- `CriteriaBuilder`'s `ensureQueryIsInitialized()` guards ship with their own dedicated commit and regression test (a real NPE from a bare `createCriteria()` call); `getPersistentEntity()` is load-bearing for `DynamicFinder`'s own machinery in this same refactor, not optional; `list(Closure)`/`call(Closure)` formalize dispatch that already worked via `invokeMethod` pre-PR and is exercised today by pre-existing, unchanged tests.
- `removeConstraints()`: this is the undo-counterpart of the double-constraint-registration bug flagged separately on the `GormRegistry.groovy` thread — same fix, not a separate concern.

Genuinely open to splitting any one of these into its own PR if the trace above doesn't change your read — wanted to give you the actual evidence rather than just assert they're fine and ask you to trust it.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/AbstractGormApiRegistry.groovy
borinquenkid: Confirmed and fixed exactly as described. `removeDatastore()`'s `catch (Exception e) { it.remove() }` (both the default and qualified maps) could evict an unrelated entity's API registration just because reading its datastore happened to throw at that moment — reachable from an ordinary unresolved-tenant condition, since `getDatastore()` routes through a `DatastoreResolver` that can throw `TenantNotFoundException` when no tenant is bound. Reproduced first as a failing test (temporarily reverted the fix, confirmed both new specs failed as expected against the original catch-and-remove code). Extracted a `belongsTo(api, datastore, className, qualifier)` helper — only a positive identity match removes; on exception, logs at warn with the exception and which entity/qualifier was affected, returns `false` (kept, not evicted). Commit `efe62b0ca2`, matches your branch's fix. 3 new specs covering cross-entity eviction stays scoped correctly and both throwing-API survival cases.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/AstUtils.groovy
borinquenkid: Scoped it to the caller rather than changing the shared helper's semantics globally. Confirmed `ServiceTransformation` is the only real caller of this method in the whole repo (an earlier grep hit on `GrailsASTUtils.copyAnnotations` was a false positive — a completely different class in a completely different module). Added a 5th `skipExisting` boolean parameter defaulting to `false` via the existing convenience overloads; `ServiceTransformation`'s two call sites now explicitly pass `skipExisting=true` to preserve its own original behavior (avoiding double-adding annotations it may already have written). Commit `fda9664613`, matches your branch and your spec's shape — repeatable annotations copy fully by default, `skipExisting=true` dedupes, the omitted default keeps both copies.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
borinquenkid: Confirmed the javadoc was simply wrong, not just imprecise — `withTenant`'s body only nests two `CurrentTenantHolder.withTenant` calls (thread-local set/remove), no session anywhere in the call chain, unlike `withId` which genuinely does `withNewSession`/`withSession`. Rewrote both `withTenant` overloads' javadoc to say plainly: binds the tenant id only, opens no session, caller owns the session the closure runs in, with a pointer to the corresponding `withId` overload for when a real session is actually needed. Commit `ebad67a331`. The withTenant-vs-withId guidance for `grails-doc` you asked for here is in the new multiTenancy.adoc page — explicit section on when to reach for each.

borinquenkid:
Went through every comment individually (replies below on each thread) and also diffed this branch against your own AI-assisted pass at the same review (`fix/gorm-registry-review-feedback`, `4d2b832628`) for every case where the two of us landed on different fixes, so I wasn't re-deriving something you'd already solved or contradicting your own diff without checking it first. Full reasoning trail (including the cases where I kept my approach over yours, with the evidence) is written up as I went; happy to share that working doc if useful.

**Blocker 1 (no documentation)** — resolved. New `grails-doc/.../GORM/multiTenancy.adoc` covering `Tenants` and `CurrentTenantHolder` end to end, plus a new upgrade-note section in `upgrading80x.adoc`. I didn't adapt your upgrade-notes draft wholesale — ran a diff of `GormEnhancer`'s actual current surface against real `origin/8.0.x` first, since a few of the removals your draft describes (`enhance()`, `dynamicEnhance`, the constructors, `multiTenancyMode`) are restored on this branch and would be actively wrong to document as gone.

**Blocker 2 (unrelated fixes bundled in)** — not split out. Traced each of the four (`getAll()` ordering/null-slot semantics, the `HibernateQuery` junction overrides, `CriteriaBuilder`'s guards, `removeConstraints()`) against git blame and the shared TCK before deciding, rather than against the PR description's framing. Detail is on the `HibernateSession.java` thread below, but short version: three of the four are demonstrated pre-existing 8.0.x contract violations this refactor surfaced and fixed (with existing, unchanged TCK coverage proving the contract), and the fourth (`removeConstraints`) is the direct undo-counterpart of the double-constraint-registration bug you flagged separately. Genuinely open to splitting any one of them out if the trace doesn't change your mind.

**Blocker 3 (TCK exclusion instead of a fix)** — fixed for real. Root cause was the shared `GormStaticApi.first(Map)`/`last(Map)` losing an H7-specific override at merge-base (not H7-specific after all — H5 has the same gap). Fixed generically, exclusion reverted, composite-key H7 case passes 12/12 again.

**The CI-blind-spot note** — `TenantDelegatingGormOperations`'s ~100 class-based `Tenants.withId` call sites are fixed (detail on the `Tenants.groovy:229` thread) — though it turns out that class is dead code on this branch, nothing currently constructs it outside its own test. Confirmed `grails-datamapping-rx` has zero diff in this PR and uses its own fully independent `Tenants` class, so there's no actual compilation gap there. On `GormApiAllocationSpec` driving internal selectors: real, still-open gap on the DATABASE-mode qualified-routing case specifically — documented as a recommended follow-up rather than closing it with an unreviewed new spec.

Two things worth a look on your own branch, found while cross-checking (not blockers, just flagging): the `getMultiTenancyMode()` rationale (on the `MongoStaticApi.groovy` thread) doesn't hold against your own `GormRegistry.createStaticApi`; and the static-side `methodMissing` clobber guard (on the `GormEnhancer.groovy` thread) is unreachable dead code on both our branches — confirmed with a few isolated `ExpandoMetaClass` probes, happy to share them.

Separately: just merged current `8.0.x` into this branch (191 commits since the fork point) to keep it current before this goes further — one doc-section-numbering conflict, resolved; full suite green after.

grails-datamapping-core/src/main/groovy/grails/gorm/multitenancy/Tenants.groovy
jdaugherty: Thanks for tracing the RX module — you're right that `grails.gorm.rx.multitenancy.Tenants` is independent and my ~40-call-site count was wrong. The `TenantDelegatingGormOperations` migration to the `MultiTenantCapableDatastore` overload is a good change on its own merits.

What I still want addressed isn't in-tree callers, though — it's that `withId(Class<? extends Datastore>, Serializable, Closure)` is gone from a public, documented entry point and nothing anywhere says so. I traced what an existing out-of-tree caller does now, and it's quieter than I assumed when I first wrote this:

`Tenants.withId(SomeDatastore, tenantId) { }`
→ `Tenants:229` `withId(Class domainClass, ..)` — same erasure, still compiles
→ `datastoreLocator.getDatastoreForDomain(SomeDatastore)` → `GormApiResolver.findDatastore(SomeDatastore, null)`
→ no `entityDatastores` entry under that class name, so the preferred / qualified / active-session selectors all miss
→ `DefaultDatastoreSelector.select` → `getDatastoreByString(className, DEFAULT)` → `datastoresByQualifier.get(DEFAULT)`

That returns the application's DEFAULT datastore. It's non-null, so the `instanceof MultiTenantCapableDatastore` check passes and the closure runs — against the wrong datastore, with no exception and no log. On a single-datastore app it's indistinguishable from correct; on a multi-datastore app it's a silent cross-database read or write that surfaces only as bad data. The loud `stateException` at `GormApiResolver:89` fires only when there is no DEFAULT datastore at all, which isn't the case in any app that would hit this.

So either:

1. restore the datastore-class overload and give the domain-class variant a distinct name — still my preference, it costs one method; or
2. keep it as-is and document the removal in `upgrading80x.adoc`. Section 45 covers `GormEnhancer`'s removed members and `CurrentTenantHolder.get()` thoroughly, but says nothing about `Tenants` — this is exactly the class of change that section exists for.

Either works for me. What I can't sign off on is it landing undocumented.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/HibernateSession.java
jdaugherty: The `getAll()` trace holds up — I checked, and `GormEnhancerSpec` has a zero-byte diff against 8.0.x, so "preserves the supplied id order" and "returns a null slot for a missing id" genuinely do pre-date this PR. That makes it a pre-existing contract violation being fixed rather than new undocumented behaviour, and I'm happy for it to stay. Thanks also for the `String.valueOf` fix on both sides of `entitiesById` — confirmed in H5 (`:212`/`:219`) and H7 (`:442`/`:449`).

On `disjunction()`/`conjunction()`/`negation()`: yes please, add the test — that's what unblocks keeping it here rather than splitting it out. What I'm after is one that actually fails against core's base implementations, so it needs to reach the junction factory (the path you describe as `countByXOrY` losing its disjunction) rather than only exercising a builder that never calls it, and it should assert the result set rather than the shape of the criteria object:

- a `countByXOrY` over two rows where each matches exactly one side — expect 2; with the junction dropped onto the unused base `criteria` field this returns the full table count
- the negation and conjunction equivalents, since all three overrides have the same root cause and only one of them is covered by the finder above

H7 only — H5's `HibernateQuery` extends `AbstractHibernateQuery` and never had this bug — so worth a line in the commit message noting the asymmetry, otherwise it reads like a missed backport. `grails-data-hibernate7/core/src/test/groovy/grails/gorm/tests/hibernatequery/` looks like the right home.

With that in place I'm fine with all of these staying in this PR.

gradle/test-config.gradle
```diff
@@ -33,6 +33,38 @@ dependencies {
...
+ *
+ * Uses Gradle 9 compatible path lookup (proj.project(pd.getPath())) because
+ * ProjectDependency.getDependencyProject() was removed in Gradle 9.
+ *
+ * @param proj the project to inspect
+ * @param targetProjectName the name of the target dependency project (e.g. 'grails-datamapping-core')
+ * @param visited set of already visited projects to prevent infinite recursion in cyclic dependency configurations
+ * @return true if the project depends on the target project
+ */
+@groovy.transform.CompileStatic
```
jdaugherty: Does the compilestatic really make an impact here?

gradle/test-config.gradle
```diff
@@ -83,7 +115,11 @@ tasks.withType(Test).configureEach {
         showStackTraces = true
     }
     excludes = ['**/*TestCase.class', '**/*$*.class']
-    maxParallelForks = configuredTestParallel
+    
+    // Selectively isolate GORM (grails-datamapping-core) dependent tests to prevent GormRegistry conflicts
+    def isGormProject = dependsOnProject(project, 'grails-datamapping-core')
```
jdaugherty: I'm confused, the forking gradle does is using a different jvm, why would there be a conflict?

jdaugherty (commented):
Some minor comments, I'd like to understanding the parallelism restriction.  Overall, I think this design is workable and it should be a noticeable memory reduction in multi-tenant environments.  I'd really like to see others review it to see if they agree with the approach.

gradle/test-config.gradle
borinquenkid: Moot now — removed `dependsOnProject` entirely along with the `isGormProject` gate it fed. See the general comment below for why.

gradle/test-config.gradle
borinquenkid: You were right to push on this. Verified empirically rather than just arguing it: ran a full CI pass with the isolation removed — `configuredTestParallel` forks for `grails-datamapping-core` and every GORM-dependent module, all running concurrently on the same shared `macos-latest` runner this PR's CI already uses — and it passed clean, including `GormRegistryConcurrencySpec` under real concurrent-fork load. Each Gradle fork is an independent JVM with its own `GormRegistry` singleton instance; there's no shared heap between forks, so there was never a genuine cross-fork data hazard here. Removed the whole mechanism in 9e3e06ddf4.

Run: https://github.com/apache/grails-core/actions/runs/32087151762

borinquenkid:
On the parallelism restriction you asked about: fair ask, and it exposed that the restriction wasn't earning its keep. `GormRegistry` is a per-JVM singleton by design — it models exactly one running application instance, same as production, where you'd never run multiple redundant forks of the same app sharing a registry. That's real design intent, but it doesn't imply Gradle's test forks need isolating from each other: each fork already gets its own independent registry instance with zero shared state.

The `maxParallelForks=1` gate for GORM-dependent modules was a proactive safeguard added without an observed failure behind it. Tested that assumption directly: ran a full CI pass with the isolation removed (`configuredTestParallel` forks for `grails-datamapping-core` and every dependent module, running concurrently on the same shared `macos-latest` runner) — it passed clean, including `GormRegistryConcurrencySpec` under real concurrent-fork load. Removed the whole `isGormProject`/`dependsOnProject` mechanism in 9e3e06ddf4; tests now run at full parallelism like everything else.

Run: https://github.com/apache/grails-core/actions/runs/32087151762

jdaugherty (approved):
@matrei can you help review this so we can get it merged?

PR #16069, opened 2026-07-30
Improve project dependency substitution logic

matrei:
This cuts the Gradle configuration time roughly in half.

### DIFF
--- a/gradle/functional-test-config.gradle
+++ b/gradle/functional-test-config.gradle
@@ -16,21 +16,25 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-def skippedProjects = testProjects + docProjects + cliProjects
-def substitutableProjects = rootProject.subprojects
-        .findAll { !(it.name in skippedProjects) }
-        .collect { subproject ->
-            // Ensure the target project is evaluated before we attempt to substitute it.
-            // This is necessary because it may not have been evaluated yet, and we need to
-            // ensure that the artifactId and cliArtifactId properties are available for substitution.
-            project.evaluationDependsOn(subproject.path)
-            return [
-                    project: subproject,
-                    artifactId: subproject.findProperty('pomArtifactId') ?: subproject.name,
-                    cliArtifactId: subproject.findProperty('cliArtifactId')
-            ]
-        }
 
+if (!rootProject.ext.has('substitutableFunctionalTestProjectsCache')) {
+    def skippedProjects = testProjects + docProjects + cliProjects
+    rootProject.ext.set('substitutableFunctionalTestProjectsCache', rootProject.subprojects
+            .findAll { !(it.name in skippedProjects) }
+            .collect { subproject ->
+                // Ensure the target project is evaluated before we attempt to substitute it.
+                // This is necessary because it may not have been evaluated yet, and we need to
+                // ensure that the artifactId and cliArtifactId properties are available for substitution.
+                project.evaluationDependsOn(subproject.path)
+                return [
+                        project: subproject,
+                        artifactId: subproject.findProperty('pomArtifactId') ?: subproject.name,
+                        cliArtifactId: subproject.findProperty('cliArtifactId')
+                ]
+            }
+    )
+}
+def substitutableProjects = rootProject.ext.get('substitutableFunctionalTestProjectsCache')
 configurations.configureEach {
     resolutionStrategy.dependencySubstitution {
         // Test projects will often include dependencies from local projects. This will ensure any dependencies
--- a/gradle/functional-test-config.gradle
+++ b/gradle/functional-test-config.gradle
@@ -16,15 +16,18 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-
+def skippedProjects = testProjects + docProjects + cliProjects
 def substitutableProjects = rootProject.subprojects
-        .findAll { !(it.name in testProjects) && !(it.name in docProjects) && !(it.name in cliProjects) }
-        .collect {
-            project.evaluationDependsOn(it.path)
+        .findAll { !(it.name in skippedProjects) }
+        .collect { subproject ->
+            // Ensure the target project is evaluated before we attempt to substitute it.
+            // This is necessary because it may not have been evaluated yet, and we need to
+            // ensure that the artifactId and cliArtifactId properties are available for substitution.
+            project.evaluationDependsOn(subproject.path)
             return [
-                    project: it,
-                    artifactId: it.findProperty('pomArtifactId') ?: it.name,
-                    cliArtifactId: it.findProperty('cliArtifactId')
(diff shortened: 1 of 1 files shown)


gradle/functional-test-config.gradle
```diff
@@ -17,49 +17,53 @@
...
-rootProject.subprojects
+def substitutableProjects = rootProject.subprojects
         .findAll { !(it.name in testProjects) && !(it.name in docProjects) && !(it.name in cliProjects) }
-        .each { project.evaluationDependsOn(it.path) }
+        .collect {
+            project.evaluationDependsOn(it.path)
+            return [
+                    project: it,
+                    artifactId: it.findProperty('pomArtifactId') ?: it.name,
+                    cliArtifactId: it.findProperty('cliArtifactId')
```
jdaugherty: This moves the `cliArtifactId` read from lazy (per configuration, when the substitution rules are registered) to eager (immediately after `evaluationDependsOn`). That matters because `cliArtifactId` is only set inside an `afterEvaluate` in `GrailsCliArtifactGradlePlugin`, which `gradle/cli-companion-bom-constraints.gradle` documents as the reason it uses a lazy `withDependencies` window.

It is correct as written — `evaluationDependsOn` flushes the target's `afterEvaluate` before returning, and I confirmed the snapshot matches the late value for all 134 substitutable projects across all 104 applying test projects with zero mismatches.

The only thing I would add is a short comment recording that invariant. The snapshot is now silently order-dependent: if `cliArtifactId` ever moves to a hook later than `afterEvaluate`, or gets set cross-project, the cli substitutions would quietly stop being registered rather than fail loudly, which is an unpleasant failure mode to debug.

gradle/functional-test-config.gradle
```diff
@@ -17,49 +17,53 @@
  *  under the License.
  */
 
-rootProject.subprojects
+def substitutableProjects = rootProject.subprojects
         .findAll { !(it.name in testProjects) && !(it.name in docProjects) && !(it.name in cliProjects) }
-        .each { project.evaluationDependsOn(it.path) }
+        .collect {
```
jdaugherty: Optional follow-up, not a blocker. This snapshot is rebuilt independently by each of the 104 test projects that apply this script, so the `findAll` plus 134 `evaluationDependsOn`/`findProperty` calls all run 104 times over.

Computing it once and sharing it (keeping a cheap per-project `evaluationDependsOn` loop so evaluation ordering is unchanged) measured a further **~10% off configuration time, 3.69s → 3.32s**, with identical resolution output.

For completeness, I also tried precomputing the `"$group:$artifactId"` coordinate strings into the snapshot so the loop body does no GString work at all — that was only 3.69s → 3.53s, near noise, so it is not worth doing for performance alone. The bulk of the win is already in this PR.

jdaugherty (commented):
Verified this is behavior-preserving and that the performance claim holds.

**Configuration time** (`./gradlew help`, warm daemon, 3 runs each, `--profile` → "Configuring Projects"): **7.95s → 3.69s (−54%)**; total wall time 17.3s → 10.8s. "Roughly in half" checks out.

**Equivalence checks**

- *Snapshot vs. lazy read* — instrumented the precomputed snapshot and compared each entry against the value the previous code would have observed later, at resolution time: 134 substitutable projects × 104 applying test projects, **zero mismatches** (9 projects carry a `cliArtifactId`, 7 a custom `pomArtifactId`).
- *Resolved graph* — `:grails-test-examples-app1:dependencies --configuration testRuntimeClasspath` is identical. A first cross-checkout comparison showed `grails-core-cli` and `grails-console` gaining a requested version, but a controlled same-directory A/B (base version of this file swapped into the PR worktree) reproduced the PR output exactly, so that delta was checkout/build state, not this change.
- *Published BOM* — `:grails-bom:generatePomFileForMavenPublication` output is byte-identical, including all 9 `*-cli` constraints.
- *Flat paths* — `project(":$name")` stays safe: all 134 substitutable projects satisfy `path == ":$name"`, so hoisting the filter introduces no nested-path regression.

No correctness issues found. Two non-blocking notes inline.

jdaugherty (approved):

gradle/functional-test-config.gradle
```diff
@@ -17,49 +17,53 @@
...
-                            handler.requireCapability("${substitutedArtifact}-test-fixtures" as String)
-                        }
-                    }
-
-                    def replacement = it.variant(project(":$possibleProject.name")) { v ->
-                        v.capabilities { it.requireCapability("${substitutedArtifact}-test-fixtures") }
+        for (def substitution : substitutableProjects) {
+            def possibleProject = substitution.project
+            def substitutedArtifact = "$possibleProject.group:${substitution.artifactId}"
+            //TODO: This does not handle libraries that are both test fixtures & a libraries like grails-data-mongodb,
```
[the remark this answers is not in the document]
matrei: TODO comment was not changed in this PR

gradle/functional-test-config.gradle
```diff
@@ -17,49 +17,53 @@
...
         .findAll { !(it.name in testProjects) && !(it.name in docProjects) && !(it.name in cliProjects) }
-        .each { project.evaluationDependsOn(it.path) }
+        .collect {
+            project.evaluationDependsOn(it.path)
+            return [
+                    project: it,
+                    artifactId: it.findProperty('pomArtifactId') ?: it.name,
+                    cliArtifactId: it.findProperty('cliArtifactId')
+            ]
+        }
```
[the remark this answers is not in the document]
matrei: I think it helps readability here.

gradle/functional-test-config.gradle
matrei: Comment added.

gradle/functional-test-config.gradle
matrei: Cached the snapshot so gain the additional 10%

jamesfredley (approved):
LGTM - nice speedup, and the hoisting of the per-configuration `findProperty` calls out of `dependencySubstitution` is clearly the right fix.

One optional follow-up on the `rootProject.ext` cache: the cached maps hold live `Project` instances, but only `name`, `group` and `path` are ever read from them. Caching plain Strings instead would drop long-lived `Project` references out of root ext, keep the cached value trivially serializable, and be friendlier to configuration cache / the Isolated Projects direction:

```groovy
return [
    name         : subproject.name,
    group        : subproject.group as String,
    artifactId   : subproject.findProperty('pomArtifactId') ?: subproject.name,
    cliArtifactId: subproject.findProperty('cliArtifactId') as String
]
```

The `def substitutableProject = substitutionInfo.project as Project` cast then goes away too, and the substitution branches just read `substitutionInfo.name` / `substitutionInfo.group`.

Not a blocker - happy for this to land as is.

jdaugherty (approved):

PR #16071, opened 2026-07-30
Add JMH performance regression reporting for pull requests

jamesfredley:
## Why

Today a performance regression in Grails just ships. There is no gate, no signal, no attribution. It surfaces months later as "our app got slower after upgrading", by which point nobody can point at the commit responsible.

This PR adds opt-in PR-level JMH comparison for Grails, inspired by Apache Groovy's approach, with a paired same-runner design that avoids cross-hardware calibration.

## What it does

Label a PR `performance` and a sticky advisory comment appears with per-benchmark verdicts.

A benchmark is only REGRESSED or IMPROVED when **both** hold:

1. the effect is at least 10%, **and**
2. the JMH confidence intervals for base and head are disjoint.

Everything else is "no clear change". Allocation (`gc.alloc.rate.norm`) is advisory. Pure-JDK ruler benchmarks are a runner-stability check only - each ruler is judged individually, never geomean-averaged - and are excluded from verdicts.

### Design choices

- **Paired same-runner A/B**: both revisions are built, then measured back-to-back on the same runner. A 2-shard matrix runs base-then-head and head-then-base so ordering bias cancels. Only complete same-shard pairs are pooled; incomplete shards are dropped rather than cross-paired.
- **Rulers as stability, not calibration**: if any ruler moves more than 5% between halves, the report warns that numbers are unreliable.
- **Per-benchmark verdicts**: names the benchmark and marks group geomeans as descriptive only.
- **Advisory only**: never fails the build for a performance finding.
- **Groovy comparison tooling**: comparison, Markdown rendering, and sticky comment posting live in a `report` source set under `grails-benchmarks` (`:grails-benchmarks:jmhCompare`). No Python.

### What is benchmarked

13 benchmarks over per-request / per-object paths, without a Spring context:

- urlmappings, databinding, gsp, interceptors, views
- ruler (CPU + memory stability probes only)

Benchmarks are Java; Groovy is confined to fixtures. `@Setup` guards fail the run rather than publish silent no-ops when a fixture stops matching or binding/rendering correctly.

### Cost and triggers

Expensive, so **opt-in only**:

- runs when the PR has the `performance` label
- on `labeled` events, only when the label just added is `performance` (other labels on an already-tagged PR do not re-run)
- doc-only path changes are ignored
- Gradle daemons are stopped before measurement so idle build JVMs do not compete with JMH forks
- if a run produces no HEAD results, the sticky comment is left unchanged (no empty overwrite)

### Security

Uses `pull_request`, never `pull_request_target`. Fork PRs get job summaries and artifacts but no comment. Only same-repo PRs reach the reporting job with `pull-requests: write`.

### Module layout

- `grails-benchmarks` - JMH suite, fixtures, Groovy report source set, Spock tests (including golden fixtures)
- `.github/workflows/benchmark.yml` - paired shard workflow + sticky comment posting
- build-time only, never published (JMH is GPLv2-with-Classpath-Exception / Category X). Code-style and vulnerability-scan are enabled; publish/sbom/jacoco stay off.

### Known gap

`Validateable.validate()` is not included: measured variance was too high for intervals to separate. Documented in `README.adoc`.

### Verification

- `:grails-benchmarks:test` green (golden fixtures + comparator/CLI/posting coverage)
- style checks clean for the module
- fixture guards and sticky-comment protection covered by the design above

### Out of scope

Long-term trend history, dashboard, gating on regressions, restoring validation benchmarks.


grails-benchmarks/build.gradle
```diff
@@ -0,0 +1,238 @@
...
+                                            }
+                                        }
+                                    }
+                                } else if (lineMetadataPrefixes.any { String prefix -> path.startsWith(prefix) }) {
+                                    Set<String> lines = lineEntries.computeIfAbsent(path) { new TreeSet<>() }
+                                    zipFile.getInputStream(entry).getText('UTF-8').readLines().each { String line ->
+                                        if (line) {
+                                            lines.add(line)
+                                        }
+                                    }
```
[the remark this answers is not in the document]
jamesfredley: Good catch, fixed in b06d51784c.
This was a genuine inconsistency rather than a theoretical one: of the three places `mergeJmhClasspathMetadata` reads a zip entry, the two `Properties.load` branches already wrapped the stream in `withCloseable`, and only the line-metadata branch called `.getText('UTF-8')` on an unclosed stream. That branch is the one that runs most often, since it handles every `META-INF/services/**` and `META-INF/groovy/**` entry across the whole jmh runtime classpath.

It now matches the other two:

```groovy
zipFile.getInputStream(entry).withCloseable { input ->
    input.getText('UTF-8').readLines().each { String line ->
        if (line) {
            lines.add(line)
        }
    }
}
```

Verified the fix changes nothing about the merge output: a full `--rerun-tasks` rebuild of `:grails-benchmarks:jmhJar` still produces the merged `META-INF/services/org.codehaus.groovy.runtime.ExtensionModule` with all 9 extension classes, and `META-INF/spring.factories` with 24 merged keys. All 13 benchmarks still execute cleanly from the rebuilt jar.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,404 @@
...
+              printf 'HEAD benchmark JAR was not built.\n\n'
+            fi
+          } > "$report_file"
+          if [ ! -f "$RESULT_DIR/head.json" ]; then
+            printf 'HEAD benchmark result was not produced.\n' >> "$report_file"
+          elif [ "${BASE_BUILD_OK:-false}" = "true" ] && [ -f "$RESULT_DIR/base.json" ]; then
+            python3 .github/scripts/jmh_compare.py --head "$RESULT_DIR/head.json" --base "$RESULT_DIR/base.json" >> "$report_file"
+          else
+            python3 .github/scripts/jmh_compare.py --head "$RESULT_DIR/head.json" >> "$report_file"
+          fi
```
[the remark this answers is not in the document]
jamesfredley: Agreed, and fixed in b06d51784c - this was a real gap. The 39 tests were the only thing standing behind the comparison logic, and nothing in CI ran them.

I've added a `🧪 Verify JMH comparison script` step that runs `python3 .github/scripts/test_jmh_compare.py`, with one deliberate change from your suggestion: rather than placing it immediately before the report is generated, it runs right after checkout and **before** the JDK/Gradle setup and the build-and-measure steps. A broken reporter should fail in seconds rather than after roughly an hour of building and benchmarking two revisions, and the shard job is the right home because it runs for fork PRs too, which never reach the reporting job.

The step is also deliberately allowed to **fail** the job. That is an intentional exception to this workflow's "never fails a build" rule, and the YAML carries a comment saying so, because the rule governs performance *regressions* (which stay advisory) whereas a broken comparison script invalidates every number the run produces. Without the note, the obvious later "fix" would be to add `continue-on-error: true` and silently defeat the guard.

Worth noting these tests are not decorative - the same suite already caught several defects during development, including shard results being silently overwritten instead of pooled, and a benchmark receiving a verdict after its JMH mode changed.

paulk-asert:
Just FYI, we find the graphs Groovy has super useful, the per PR/commit regression stats not so much because of the wide variance of GitHub runner speed. It does some "ruler adjusting" to try to account for runner speed but even then it isn't super accurate - we just use it as a guide to then check things manually on a dedicated machine.

jdaugherty:
Why are we using python in this PR?

jamesfredley:
The implementation was based on https://github.com/apache/groovy/blob/master/subprojects/performance/dashboard/jmh-summary.py, didn't notice the Python, but will make it Groovy instead.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,381 @@
...
+# limitations under the License.
+
+name: "JMH Benchmark Comparison"
+
+# SECURITY: This workflow deliberately uses pull_request, never pull_request_target.
+# Pull requests run untrusted code, and Apache Infra policy forbids exposing tokens
+# to that code through a privileged pull_request_target workflow.
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
jdaugherty, line 23: `labeled` fires for *any* label, and the job gate at line 46 only checks that `performance` is **present**, not that it was the label just added. So on a PR that already carries `performance`, adding `bug`, `deps`, or anything else kicks off another full two-shard run — two framework builds and two JMH suites per shard.

Given the CI budget argument in the description, worth narrowing:

```yaml
if: >-
  github.event_name == 'workflow_dispatch' ||
  (contains(github.event.pull_request.labels.*.name, 'performance') &&
   (github.event.action != 'labeled' || github.event.label.name == 'performance'))
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,381 @@
...
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [a, b]
+    env:
+      BASE_SHA: ${{ github.event.pull_request.base.sha }}
+      HEAD_SHA: ${{ github.sha }}
+      JMH_INCLUDE: '.*'
+      PR_NUMBER: ${{ github.event.pull_request.number || 0 }}
```
jdaugherty: `PR_NUMBER` and `REPOSITORY` are never referenced in this job — the compare step deliberately doesn't post, and the `report` job declares its own copies at lines 271-272. Dead as written; dropping them keeps it clear that this job has no comment-posting path.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,381 @@
...
+          # completeness from which files exist, so a partial file would be indistinguishable from
+          # a good one. Write to a staging path and publish it only on success, so a failed run
+          # leaves NO file rather than a plausible one.
+          run_benchmark() {
+            local revision_dir="$1"
+            local result_file="$2"
+            local staging_file="$result_file.partial"
+            rm -f "$staging_file" "$result_file"
+            if (
+              cd "$revision_dir" && ./gradlew :grails-benchmarks:jmh \
```
jdaugherty: Both revisions were built in the preceding step, so by the time measurement starts there are up to two Gradle daemons resident (3h default expiry) alongside this build's own, all on a 4-core runner, for the whole measurement window. Idle daemons still hold heap and run background GC.

The rulers exist to *detect* exactly this class of noise — cheaper to remove a known source of it. Either `--no-daemon` on the two measurement invocations, or a `./gradlew --stop` between the build step and this one, would leave the JMH fork as the only significant JVM competing for the runner.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,381 @@
...
+        if: always()
+        run: |
+          WORKTREE_ROOT="$RUNNER_TEMP/jmh-worktrees/$SHARD"
+          git worktree remove --force "$WORKTREE_ROOT/base" || true
+          git worktree remove --force "$WORKTREE_ROOT/head" || true
+
+  report:
+    name: "Publish JMH benchmark comparison"
+    needs: benchmark
+    if: ${{ always() && github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'performance') && github.event.pull_request.head.repo.full_name == github.repository }}
```
jdaugherty: `always()` also runs this job when the `benchmark` jobs failed outright. Concrete case: the "🧪 Verify JMH comparison tool" gate fails, every later step in the shard job is skipped, so no results are uploaded. The pooling step then falls through to the final `else` and posts *"No HEAD benchmark results were produced"* — which, because the comment is sticky, **replaces the previous good report on that PR with an empty one**. A tooling failure silently destroys the last known-good numbers.

Worth either skipping the post step when there are zero results (leaving the prior comment intact), or making the message say the run failed and the earlier report still stands.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/JmhCompare.groovy
```diff
@@ -0,0 +1,116 @@
...
+    private static Map<String, String> parse(String[] args) {
+        Set<String> values = ['head', 'base', 'threshold', 'repo', 'pr-number', 'expected-shards', 'output', 'post-file'] as Set<String>
+        Map<String, String> options = new LinkedHashMap<>()
+        for (int index = 0; index < args.length; index++) {
+            String option = args[index]
+            if (!option.startsWith('--')) {
+                throw new IllegalArgumentException("unknown option: ${option}")
+            }
+            String key = option.substring(2)
+            if (key == 'fail-on-regression') {
```
jdaugherty: `--fail-on-regression` is parsed and stored but never read — `run()` returns 0 regardless, and `JmhCompareSpec` line 113 ("fail on regression remains successful") locks that in.

Since gating is explicitly out of scope for this PR, I'd drop the flag and that test rather than ship a CLI option whose name promises behaviour it doesn't have. As it stands someone reaches for it later, sees it accepted without error, and assumes the build is gated.

grails-benchmarks/src/jmh/java/org/apache/grails/benchmarks/interceptors/UrlMappingMatcherBenchmark.java
```diff
@@ -0,0 +1,144 @@
...
+@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 2, jvmArgsAppend = {"-Xms1g", "-Xmx1g", "-XX:+UseG1GC"})
+public class UrlMappingMatcherBenchmark {
+
+    private UrlMappingMatcher matcher;
+    private UrlMappingInfo mappingInfo;
+
+    @Setup
+    public void setup() {
```
jdaugherty, line 59: `UrlMappingsBenchmark` guards its fixture in `@Setup` and fails the run rather than publish wrong numbers, and the description makes a strong case for why. This benchmark has the same failure mode with no guard: if `doesMatch` ever stops returning `true` for `/orders/42`, `matchUriPattern` silently becomes a second copy of `rejectNonMatchingUriPattern` and the report shows two plausible, nearly identical numbers with no error anywhere.

I ran a probe on this branch against `sourceSets.jmh.runtimeClasspath` and confirmed the fixtures are all fine today — `doesMatch` returns `true` then `false`, both binder targets come back fully populated, both templates render non-empty — so this is not a live defect. But the same one-line assertion applies here, in `SimpleDataBinderBenchmark` (`SimpleDataBinder.bind` skips properties it can't bind, silently), and in `ViewTemplateRenderingBenchmark`. Given the silent-no-op story is the lesson the description leads with, the guard is worth applying uniformly rather than only where the bug happened to be found.

grails-benchmarks/build.gradle
```diff
@@ -0,0 +1,268 @@
...
+    id 'org.apache.grails.buildsrc.compile'
+}
+
+version = projectVersion
+group = 'org.apache.grails'
+
+// This build-time-only benchmark harness is never published, so it deliberately omits
+// org.apache.grails.buildsrc.publish, org.apache.grails.buildsrc.sbom,
+// org.apache.grails.buildsrc.vulnerability-scan, org.apache.grails.gradle.grails-jacoco,
+// org.apache.grails.buildsrc.dependency-validator, and org.apache.grails.gradle.grails-code-style.
```
jdaugherty: Two of these omissions read differently from the rest. `publish`/`sbom`/`jacoco` follow straightforwardly from "never published".

`grails-code-style` doesn't. The trailing note says it stays off "until existing JMH source import-order and fixture-formatting violations can be remediated together", which leaves a new module permanently invisible to `./gradlew codeStyle` and to the four aggregate violation reports the contributor guide asks for before every commit. It's about ten files — fixing them now avoids establishing the precedent that a new module can opt out.

`vulnerability-scan` is worth a second look too: this module pulls JMH and its transitives (jopt-simple, commons-math3) into the default build graph, and nothing else in the repo scans them.

grails-benchmarks/build.gradle
```diff
@@ -0,0 +1,268 @@
...
+        if (mergeable && !details.file.absolutePath.startsWith(mergedMetadataPath)) {
+            details.exclude()
+        }
+    }
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+}
+
+tasks.named('check') {
+    // Benchmarks are not otherwise compiled by the standard lifecycle tasks.
+    dependsOn tasks.named('jmhClasses')
```
jdaugherty, line 288: Wiring `jmhClasses` into `check` means every root `./gradlew build`/`check` now compiles the JMH source set and resolves JMH, plus the six `implementation project(...)` dependencies, for a module only exercised by an opt-in workflow. That's a cost on every developer and CI build, which sits a little oddly beside the budget discipline argued for the workflow itself.

The alternative — letting the workflow's `jmhJar` be the only compile gate — risks the benchmarks rotting silently between labelled runs, so I don't think the current choice is wrong. But it deserves to be a recorded decision rather than a bare `dependsOn`; could the comment say why compile-on-every-build was preferred?

grails-benchmarks/src/test/groovy/org/apache/grails/benchmarks/report/GoldenReportSpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+class GoldenReportSpec extends Specification {
+    @TempDir
+    Path temporaryDirectory
+
+    @Unroll
+    def "#name report exactly matches the Python golden file"() {
```
jdaugherty: Leftover from the Python implementation this replaced — "the Python golden file" here, and "matches Python percent g rendering" in `JmhCompareSpec` line 81. There's no Python anywhere in the PR now, so these read as references to something that doesn't exist. The fixtures are still worth keeping as byte-exact locks; they're just locking the Groovy renderer's output.

jdaugherty (commented):
Initial AI Review: 

Reviewed the workflow, the `report` source set, the benchmarks and fixtures, and the golden tests. I checked the branch out and ran `:grails-benchmarks:test` (green, 43 tests, and the module is picked up by the aggregate test report), and I wrote a throwaway probe against `sourceSets.jmh.runtimeClasspath` to confirm every fixture actually does work today — `doesMatch` returns `true`/`false` as intended, both binder targets come back fully populated, and both templates render non-empty output. So nothing here is currently measuring a no-op.

The design holds up. Same-runner paired A/B removes the cross-hardware correction that weakens Groovy's version, per-ruler judgement rather than a geomean is the right call and the reasoning for it is correct, and declining to put p-values on ~10 JMH iterations that share a JVM is the honest choice. Verdict logic, shard pairing, the union-of-intervals pooling, and the sticky-comment path all read correctly, and the action pins and Develocity wiring match the other workflows.

Comments below are mostly cost, robustness, and one dead CLI flag. Two I'd want resolved before merge: the `labeled` re-trigger (a full two-shard run fires on *any* label change once `performance` is present) and the sticky comment being overwritten with an empty report when a run fails early. The rest are smaller.

.github/workflows/benchmark.yml
jamesfredley: Fixed in 202d206a6a.

Both the `benchmark` and `report` job gates now require:

```yaml
github.event.action != 'labeled' || github.event.label.name == 'performance'
```

So once a PR already carries `performance`, adding `bug`/`deps`/etc. no longer kicks off another two-shard run.

.github/workflows/benchmark.yml
jamesfredley: Fixed in 202d206a6a.

Dropped `PR_NUMBER` and `REPOSITORY` from the shard job env. They remain only on the `report` job, which is the sole comment-posting path.

.github/workflows/benchmark.yml
jamesfredley: Fixed in 202d206a6a.

Added a `Stop Gradle daemons before measurement` step between the paired `jmhJar` builds and the JMH runs. It runs `./gradlew --stop` in the main workspace and both worktrees so idle daemons are not competing with the JMH forks during the measurement window.

.github/workflows/benchmark.yml
jamesfredley: Fixed in 202d206a6a.

The pool/compare step now sets `should_post=true` only when complete pairs or HEAD-only results exist. The comment step is gated on that output. When nothing was produced, the job summary says the run failed and the previous sticky report was left unchanged - no empty overwrite, and the sticky marker is not written in that path.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/JmhCompare.groovy
jamesfredley: Fixed in 202d206a6a.

Removed the `--fail-on-regression` parse branch and the test that locked in the no-op success path. Passing the flag now fails CLI parsing (missing value / unknown option) instead of silently accepting it.

grails-benchmarks/src/jmh/java/org/apache/grails/benchmarks/interceptors/UrlMappingMatcherBenchmark.java
jamesfredley: Fixed in 202d206a6a.

Added `assertFixtureMatches()` guards called from `@Setup` in:

- `UrlMappingMatcherBenchmark` - match `/orders/42`, reject `/catalog/42`
- `SimpleDataBinderBenchmark` - both flat and type-converted binds populate all fields
- `ViewTemplateRenderingBenchmark` - JSON and markup renders contain expected content

Same fail-loud pattern as `UrlMappingsBenchmark`.

grails-benchmarks/build.gradle
jamesfredley: Fixed in 202d206a6a.

Enabled `org.apache.grails.gradle.grails-code-style` and `org.apache.grails.buildsrc.vulnerability-scan`. Import-order / formatting issues in the module were cleaned so style checks pass. Publish/sbom/jacoco/dependency-validator stay off because this harness is build-time-only and JMH is Category X.

grails-benchmarks/build.gradle
jamesfredley: Fixed in 202d206a6a.

Documented the decision on `check.dependsOn jmhClasses`: compile-on-every-build is preferred over workflow-only compilation so the suite cannot silently bit-rot between labelled runs; `jmhJar` remains the packaging gate.

grails-benchmarks/src/test/groovy/org/apache/grails/benchmarks/report/GoldenReportSpec.groovy
jamesfredley: Fixed in 202d206a6a.

Renamed the golden test to `#name report exactly matches the golden fixture` and the number-format test to `matches percent-g style rendering`. No Python references remain under `grails-benchmarks`. Also normalized CRLF in the golden comparison so Windows CI does not false-fail on line endings.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,396 @@
...
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        with:
+          cache-provider: basic # 'basic' uses the MIT-licensed, open-source cache provider; the default 'enhanced' provider (v6+) is proprietary (Gradle commercial Terms of Use)
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+      # Deliberately allowed to fail the job, and placed before the expensive work: a broken
+      # comparison tool makes every number produced here untrustworthy. That is a tooling
+      # failure rather than a performance finding, so regressions themselves stay advisory.
+      - name: "🧪 Verify JMH comparison tool"
+        run: ./gradlew :grails-benchmarks:test --max-workers=4
```
jdaugherty, line 79: This gate is placed first, and the comment above it explains why: a broken comparison tool should fail in seconds rather than after an hour of building and benchmarking two revisions. But `:grails-benchmarks:test` isn't seconds — it builds 33 framework projects before it can run a single spec.

The `test` source set inherits `main`'s output, and `main` is where the Groovy fixtures live, so the six `implementation project(...)` dependencies come along. Measured on this branch:

| gate | tasks | projects |
| --- | ---: | ---: |
| `:grails-benchmarks:test` (current) | 222 | 33 |
| isolated `reportTest` source set | 9 | 1 |

I probed the isolated version — a `reportTest` source set whose classpath is `sourceSets.report.output` plus Spock and `groovy-json`, with its own `Test` task — and every spec passes unchanged in 19s including compilation. No spec touches the fixtures, so nothing is lost by cutting `main` out of the classpath.

`README.adoc` already states this as the design intent: "The `report` source set depends on Groovy alone and on no Grails project." The *test* source set is what breaks that isolation.

One thing to watch if you split it: `grails-test-report` collects test tasks with `tasks.withType(Test).matching { it.name == phase }`, so a task named anything other than `test` drops out of the aggregate report and would need adding there.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/ReportRenderer.groovy
```diff
@@ -0,0 +1,148 @@
...
+                        : "worst ruler deviation: ${String.format(Locale.ROOT, '%.1f%%', deviation(worst.speedup) * 100D)}${worst.shard ? ' in ' + safe(worst.shard) : ''} (${safe(removePrefix(worst.identity, BenchmarkComparator.RULER_PACKAGE))})"
+        List<String> lines = ['### JMH Benchmark Report', '', "**Regressions:** ${regressions}".toString(), "**Improvements:** ${improvements}".toString(), "**Runner health:** ${health}".toString(), 'Ruler benchmarks are excluded from verdicts and group summaries; runner health is a stability check, not a calibration factor.']
+        if (comparison.missingShards) lines.addAll(['', '> **Warning:** No usable base/head pair was produced by ' + comparison.missingShards.collect { String item -> safe(item) }.join(', ') + ". This comparison rests on ${comparison.pairedShards.size()} shard pair(s) instead of the expected ${comparison.pairedShards.size() + comparison.missingShards.size()}, so the alternating measurement order did not fully cancel and the result is weaker than a normal run."])
+        if (deviations) {
+            lines.add('**Ruler movements:** ' + deviations.collect { RulerDeviation item ->
```
jdaugherty, line 90: Line 85 emits a `>` blockquote and this line follows it with no blank line, so GFM lazy continuation pulls "Ruler movements" *inside* the warning blockquote. Confirmed against GitHub's own renderer (`POST /markdown`, `mode: gfm`):

```html
<blockquote>
<p><strong>Warning:</strong> No usable base/head pair was produced by shard-c.json.<br>
<strong>Ruler movements:</strong> shard-a.json: CpuRuler.run: 0.89x, ...</p>
</blockquote>
```

This only fires when `missingShards` is non-empty — which is exactly the degraded run the warning exists to call out, and the one where the ruler numbers matter most. Prefixing the `lines.add` with `''` fixes it.

The two warnings below (lines 93 and 94) already do this correctly, so it's just this one.

Worth noting the golden fixtures can't catch this class of bug: they lock the Markdown *source* bytes (`expected-shards.md` lines 8-9 encode the current, broken layout), not the rendered output.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,396 @@
...
+        run: |
+          WORKTREE_ROOT="$RUNNER_TEMP/jmh-worktrees/$SHARD"
+          git worktree remove --force "$WORKTREE_ROOT/base" || true
+          git worktree remove --force "$WORKTREE_ROOT/head" || true
+
+  report:
+    name: "Publish JMH benchmark comparison"
+    needs: benchmark
+    if: >-
+      always() && github.event_name == 'pull_request' &&
```
jdaugherty, line 271: `github.event_name == 'pull_request'` here means a `workflow_dispatch` run can never reach the reporting job, so it produces artifacts and per-shard job summaries but no sticky comment. The PR description says a comment appears for `workflow_dispatch` too.

There's a second consequence: on dispatch there's no `pull_request` context, so `BASE_SHA` is empty and base resolution in "Prepare paired worktrees" falls through the merge-base branch to `git rev-parse "$HEAD_SHA^"`. A manual run therefore benchmarks HEAD against its own parent commit, not against a release branch.

Both behaviours are defensible, but neither is what the description promises. Either document what dispatch actually does, or drop the trigger.

grails-benchmarks/src/jmh/java/org/apache/grails/benchmarks/gsp/GroovyPageParserBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+ * are prepared. Parser regressions slow application startup and template reloads.
+ */
+@State(Scope.Benchmark)
+@Threads(1)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 2, jvmArgsAppend = {"-Xms1g", "-Xmx1g", "-XX:+UseG1GC"})
+public class GroovyPageParserBenchmark {
```
jdaugherty, line 51: This is the one benchmark left without a fixture guard — it has no `@Setup` at all. `UrlMappingsBenchmark`, `UrlMappingMatcherBenchmark`, `SimpleDataBinderBenchmark` and `ViewTemplateRenderingBenchmark` all now fail the run rather than publish a silent no-op.

The same failure mode applies here, and arguably more directly than anywhere else: the measured work *is* the volume of generated source. If `parse()` ever returned an empty or near-empty stream, both methods would keep producing plausible, self-consistent, meaningless numbers with no error anywhere.

A `@Setup` that parses both templates once and asserts the generated source is non-trivial (and, for `TAGGED_TEMPLATE`, that it actually contains the tag output) would close the last gap in the fail-loud story.

grails-benchmarks/build.gradle
```diff
@@ -0,0 +1,270 @@
...
+// Its JMH dependency is GPLv2 with the Classpath Exception (Category X) and must not be published.
+
+sourceSets {
+    report {
+        groovy.srcDirs = ['src/report/groovy']
+    }
+}
+
+// The Groovy fixtures under src/main/groovy build the framework objects whose APIs are
+// closure-based (the URL mappings DSL, the Validateable trait, view templates). The JMH
```
jdaugherty: `the Validateable trait` is stale — there is no Validateable fixture in the module. `README.adoc` line 93 carries the same claim in its Layout section, and is then contradicted by its own "Known gap: validation" section 28 lines later, which explains that the validation benchmark was written and withdrawn.

Same category as the Python leftovers from the last round: a reader looking for the Groovy fixture that needs the trait won't find one. The remaining two examples (the URL mappings DSL and view templates) carry the argument on their own.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/JmhResults.groovy
```diff
@@ -0,0 +1,139 @@
...
+            if (lower != null && upper != null) {
+                return [Math.min(lower, upper), Math.max(lower, upper)]
+            }
+        }
+        Double score = finiteNumber(metric.get('score'))
+        Double error = finiteNumber(metric.get('scoreError'))
+        return score != null && error != null && error >= 0D ? [score - error, score + error] : null
+    }
+
+    static String identity(String name, Map<String, Object> params) {
```
jdaugherty: `identity` keys on benchmark name plus params but not on `mode`, and `parseEntries` does a plain `benchmarks.put(benchmarkId, ...)`. A benchmark declaring more than one mode — `@BenchmarkMode({Mode.AverageTime, Mode.Throughput})` — emits one JSON entry per mode under the same name, so all but the last are silently dropped.

Nothing in the suite does this today, and `comparable()` plus `poolBenchmarks` correctly refuse to compare across modes, so it can't produce a *wrong* verdict. But it can silently shrink the comparison set, surfacing only as an anonymous "Malformed comparisons skipped: N" count — the same silent-shrink failure that `jmh.failOnError = true` was set to prevent on the measurement side.

Adding mode to the identity key would make it a visible `Only in head` / `Only in base` entry instead.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/CommentPoster.groovy
```diff
@@ -0,0 +1,111 @@
...
+
+@CompileStatic
+class GitHubComments implements CommentPoster {
+
+    static final int MAX_BODY = 65000
+    static final String TRUNCATION = '\n\n_Report truncated. The full report is available in the workflow artifacts._\n\n<!-- grails-jmh-benchmark -->'
+    private final HttpClient client = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(15)).build()
+
+    @Override
+    void post(String report, String repo, String prNumber, String token) {
```
jdaugherty, line 52: `post` itself is untested. The specs cover `commentBody` and `latestMatchingCommentId`, and everything through `JmhCompare` uses `Mock(CommentPoster)`, so the paginated GET loop and the create-versus-update decision on line 63 never execute in a test.

That's the sticky-comment mechanism — the thing whose failure mode is a duplicate report on every push rather than an updated one, which is the specific behaviour the marker exists to prevent. A `com.sun.net.httpserver.HttpServer` stub would cover the page-boundary walk (the `comments.size() < 100` break) and assert PATCH-versus-POST against a canned comment list.

Worth flagging that coverage tooling won't catch this for you: the module deliberately omits `grails-jacoco`, so the "all modified and coverable lines are covered by tests" report on this PR says nothing about `grails-benchmarks`.

Unrelated, while you're in the specs: `JmhCompareSpec`'s first two features ("speedup respects throughput and latency direction", "verdict requires threshold and disjoint intervals") duplicate five features in `BenchmarkComparatorSpec` exactly.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/CommentPoster.groovy
```diff
@@ -0,0 +1,111 @@
...
+            body = body.substring(0, MAX_BODY - TRUNCATION.length()) + TRUNCATION
+        }
+        return body
+    }
+
+    private static boolean markerComment(Map<String, Object> comment) {
+        Object body = comment.get('body')
+        Object user = comment.get('user')
+        Object login = user instanceof Map ? ((Map<String, Object>) user).get('login') : null
+        return body instanceof String && ((String) body).contains(ReportRenderer.MARKER) && (!(login instanceof String) || login == 'github-actions[bot]')
```
jdaugherty: The author check defaults the wrong way: `!(login instanceof String) || login == 'github-actions[bot]'` treats a comment whose `user` is absent or non-object as ours. If GitHub returns a comment without a usable `user` (deleted accounts are the usual way this happens) and that body contains the marker, it becomes a PATCH target.

The reporting job holds `pull-requests: write`, which is enough to edit any comment in the repository, so the failure mode is overwriting someone else's comment rather than a harmless 403.

Requiring the bot instead of assuming it — `login == 'github-actions[bot]'` — keeps the intended behaviour and makes the unknown-author case fall through to posting a fresh comment, which is the safe direction.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,396 @@
...
+          } >> "$GITHUB_ENV"
+      - name: "🛑 Stop Gradle daemons before measurement"
+        if: always()
+        run: |
+          ./gradlew --stop || true
+          if [ -n "${BASE_DIR:-}" ] && [ -d "$BASE_DIR" ]; then (cd "$BASE_DIR" && ./gradlew --stop) || true; fi
+          if [ -n "${HEAD_DIR:-}" ] && [ -d "$HEAD_DIR" ]; then (cd "$HEAD_DIR" && ./gradlew --stop) || true; fi
+      # Two forks, three warmup iterations, and five measurement iterations balance PR latency
+      # against confidence. Reversing shard order cancels first-versus-second runner-state bias.
+      - name: "🌡️ Run paired JMH benchmarks"
```
jdaugherty, line 150: Minor follow-on to the daemon fix: the `--stop` above clears the daemons left by the build phase, but not the one the *first* measurement starts. Whichever revision runs second in a shard is measured on a daemon already warmed and loaded by the first (Gradle reuses a compatible idle daemon across project directories, so it's the same JVM rather than a second one).

The alternating shard order cancels this on average, which is why it's a nit rather than a finding. But `--no-daemon` on the two `run_benchmark` invocations would make both halves start from identical daemon state and remove a source of noise the rulers would otherwise have to absorb.

jdaugherty (commented):
Re-reviewed at 202d206a. All eleven threads from the previous round are addressed, and I verified each rather than taking the replies at face value: the label gate is narrowed on both jobs, the dead shard-job env vars are gone, the daemon stop is in place, `should_post` protects the sticky comment from an empty overwrite, `--fail-on-regression` is gone along with the test that locked in its no-op, guards are in place on three benchmarks, `grails-code-style` and `vulnerability-scan` are on, the `check.dependsOn jmhClasses` decision is recorded, and no Python references remain.

I also confirmed locally that `:grails-benchmarks:codeStyle` is clean — zero CodeNarc violations across `main`, `jmh` and `report`, and the `report` source set is genuinely covered by the aggregate task, not just nominally enabled — and that `grails-test-report` picks up the new `test` task automatically, so the module lands in the aggregate test report.

What follows is the second pass. Two items are worth acting on before merge: the CI tooling gate does not actually fail fast, and the ruler-movements line renders inside the warning blockquote in exactly the degraded case that warning exists to flag. The rest are smaller, and the last one is a nit. Nothing here changes my view of the design — the paired same-runner approach, per-benchmark verdicts, and rulers-as-stability-not-calibration are the right calls.

grails-benchmarks/build.gradle
jamesfredley: Addressed in b8d3233184 and 22c45df65a. The stale Validateable fixture references were removed from the build and README while the withdrawn validation gap remains documented.

jamesfredley:
Merging this so that we can use it on a number of open Grails 8 PRs where we need to evaluate performance.

Adding 'performance' label triggers it.  

Can iterate on this later.

PR #16077, opened 2026-08-01
Isolate mavenBom import requirements of Spring DM test project

jdaugherty:
An alternative to https://github.com/apache/grails-core/pull/16073 since we now have the end to end tests - they also really only need to run in CI instead of locally.


end-to-end/spring-dependency-management/build.gradle
```diff
@@ -30,22 +30,26 @@
...
+    id 'org.apache.grails.gradle.grails-web'
+    id 'org.apache.grails.gradle.grails-gsp'
 }
 
 version = '0.1'
-group = 'functionaltests'
-
-apply plugin: 'org.apache.grails.gradle.grails-web'
-apply plugin: 'org.apache.grails.gradle.grails-gsp'
+group = 'springdm'
```
jamesfredley, line 46: **Optional nit:** `group` changed from `functionaltests` to `springdm`. No functional need jumps out for the move. Fine either way; only worth reverting if you want a pure rename/move diff.

end-to-end/spring-dependency-management/build.gradle
```diff
@@ -55,6 +59,8 @@ grails {
 
 dependencyManagement {
     imports {
+        // Resolved by Spring DM's own detached configuration, so it is a genuine Maven BOM import
+        // of the pom the core build published into build/local-maven - not a substituted project.
         mavenBom "org.apache.grails:grails-bom:${projectVersion}"
```
jamesfredley, line 64: This is the right thing to keep. The fixture exists to exercise a real Spring DM `mavenBom` import of `grails-bom`, which only a published pom can satisfy. Putting that behind `build/local-maven` + `exclusiveContent` is why end-to-end is the correct home (vs hand-managing versions from `dependencies.gradle` as in #16073).

end-to-end/spring-dependency-management/build.gradle
```diff
@@ -63,7 +69,7 @@ dependencyManagement {
 // the grails-bom import, so the security-patched logback isn't applied automatically. A migrated Grails 7
 // app hitting the same CVE would override the Spring-managed version property directly; reproduce that here,
 // sourcing the number from dependencies.gradle so it stays the single source of truth.
-apply from: rootProject.layout.projectDirectory.file('dependencies.gradle')
+apply from: rootProject.layout.projectDirectory.file('../dependencies.gradle')
 ext['logback.version'] = bomDependencyVersions['logback.version']
 // Same situation for the Jackson 3 security override (CVE-2026-59889) - see jackson3.version in dependencies.gradle.
 ext['jackson-bom.version'] = bomDependencyVersions['jackson3.version']
```
jamesfredley, line 75: **Optional:** Logback/jackson property overrides correctly document that Spring Boot version properties can win over the grails-bom import.

With the real `mavenBom` still here (unlike #16073's final approach), coordinate-level management from grails-bom should pin most other coords. A green `HelloControllerSpec` still will not notice silent property wins on e.g. groovy/log4j2/commons-codec.

If you want belt-and-suspenders without rewriting the fixture: a small `verify…` task on `check` asserting a few resolved versions (at least `org.apache.grails:*` → `projectVersion`, plus these logback/jackson pins). Not a merge blocker.

end-to-end/spring-dependency-management/build.gradle
```diff
@@ -81,6 +87,18 @@ dependencies {
...
+// project names, has no meaning in this build, which is driven by its own workflow.
 apply {
-    from rootProject.layout.projectDirectory.file('gradle/functional-test-config.gradle')
+    from rootProject.layout.projectDirectory.file('gradle/test-config.gradle')
+}
+
+tasks.withType(Test).configureEach {
+    // Allow extra headroom for slow start-up under CI load (the client default is 60s), matching
+    // what gradle/functional-test-config.gradle gave this application in the core build.
+    systemProperty('grails.http.client.timeout', '120')
```
jamesfredley, line 103: **Optional follow-up:** Timeout carry-over looks good.

Related optional idea from #16073: a tiny verification task that fails on version drift. The IT proves the app boots under Spring DM against the published BOM; it does not prove managed versions match the tree under test. Easy to add later if you prefer not to grow this PR.

end-to-end/gradle/test-config.gradle
```diff
@@ -0,0 +1,35 @@
...
+tasks.withType(Test).configureEach {
+    useJUnitPlatform()
+
+    // Honor DO_NOT_CACHE_TESTS=1 so developers can repeatedly invoke the same test command
+    // without --rerun-tasks (and without recompiling everything else). It matters more in this
+    // build than in the core one: republishing grails-core does not change these projects'
+    // inputs in any way Gradle can see, so re-running against freshly published artifacts is
+    // exactly the case that otherwise comes back UP-TO-DATE or FROM-CACHE.
+    outputs.cacheIf { !doNotCacheTests }
+    outputs.upToDateWhen { !doNotCacheTests }
```
jamesfredley, line 34: **Optional:** Matching core's `DO_NOT_CACHE_TESTS` pattern is the right call - republished `build/local-maven` is invisible to Gradle inputs, so FROM-CACHE is a real footgun locally.

For CI, consider also setting `DO_NOT_CACHE_TESTS=1` on the workflow `check` step. Not required on a clean runner that always publishes first; cheap insurance now that this workflow is the sole home of the Spring DM BOM-import fixture.

Also worth confirming the E2E job is a **required** status on `8.0.x` - if optional, this path can go red without blocking merge.

.github/workflows/end-to-end.yml
```diff
@@ -20,7 +20,10 @@
 # substitution - that is what makes the tests end-to-end. The repository is the same
 # build/local-maven that grails-forge points its generated applications at, populated by
 # publishAllPublicationsToTestCaseMavenRepoRepository, so the suite exercises real poms and
-# module metadata including the CLI companion artifacts.
+# module metadata including the CLI companion artifacts. That is a hard requirement for some of
+# these projects rather than a nicety: spring-dependency-management imports grails-bom as a Maven
+# BOM through io.spring.dependency-management, which resolves it in its own detached configuration
+# and so can only ever consume a published pom.
```
jamesfredley, line 26: Good callout in the workflow header - this is now a hard requirement for the suite, not a nicety.

**Optional:** on the `Run the end-to-end tests` step, set `DO_NOT_CACHE_TESTS: '1'` so a build-cache hit cannot skip tests after the publish steps above. Low probability on clean GHA runners; low cost.

end-to-end/settings.gradle
```diff
@@ -103,10 +104,26 @@ rootProject.name = 'grails-end-to-end'
...
         'legacy-commands-plugin',
+        // Belongs here rather than in grails-test-examples: it imports grails-bom as a Maven BOM
+        // through io.spring.dependency-management, which resolves imports in its own detached
+        // configuration. That bypasses any project substitution, so the import can only ever be
+        // satisfied by a published org.apache.grails:grails-bom:<projectVersion> - which is what
+        // the repository above is. In the core build it had to be excluded whenever nothing had
+        // been published yet (a reproducible release build, or a fresh release branch whose
+        // version has never been published), and otherwise resolved against whatever the Apache
+        // snapshot repository held rather than the working tree.
+        'spring-dependency-management',
```
jamesfredley, line 115: Agree with the placement and the comment. Correct home for a fixture that must import a published `grails-bom` pom through Spring DM's detached configuration.

**Optional PR-body note:** leaving core means this fixture no longer rides the Win/macOS/J25 functional matrix or SiteMesh 2 path filters. Acceptable for a published-artifact boundary test; just call the trade-off out so nobody assumes matrix parity with `grails-test-examples/*`.

end-to-end/settings.gradle
```diff
@@ -103,10 +104,26 @@ rootProject.name = 'grails-end-to-end'
...
 // org.apache.grails:grails-core for this repo's Groovy 5 project - which is exactly the
 // substitution the fixture exists to avoid. It stays a standalone build, run under its own JDK
 // (legacy-g7-command-plugin/.sdkmanrc), and legacy-commands consumes the jar it produces.
+
+// Due to https://github.com/gradle/gradle/issues/2986 , we can't change the global exclude for profiles, it must be done everywhere
+for (String pattern in DirectoryScanner.defaultExcludes) {
+    if (pattern.contains('gitignore') || pattern.contains('gitattributes')) {
+        DirectoryScanner.removeDefaultExclude(pattern)
+    }
+}
```
jamesfredley, line 129: **Optional / out of band:** Same Gradle #2986 `DirectoryScanner` workaround as root (also added in `build-logic` / `grails-gradle` / `grails-forge`). Legitimate for each independent settings script - Ant state is per-build - and not required for the Spring DM move itself.

Fine to keep for consistency. If you want a thinner review surface, the non-`end-to-end` copies could be a tiny chore PR; not worth blocking this one.

end-to-end/README.md
```diff
@@ -35,12 +35,21 @@ applications at via `GRAILS_REPO_URL`.
...
 the substitution the fixture exists to avoid — it must be compiled by a real Grails 7 toolchain for
 its trait-woven bytecode to prove anything.
 
+`spring-dependency-management` is here because Spring DM imports a BOM as a **Maven** BOM, resolving
+it in its own detached configuration. That bypasses any project substitution, so the import can only
+ever be satisfied by a published `org.apache.grails:grails-bom:<projectVersion>` — which is precisely
+what this build already provides. In the core build it had to be excluded whenever nothing had been
+published yet (a reproducible release build, or a fresh release branch whose version has never been
+published), and otherwise silently fell back to whatever the Apache snapshot repository happened to
+hold rather than the working tree.
```
jamesfredley, line 51: Docs read well - why it lives here, publish steps, and `DO_NOT_CACHE_TESTS` are exactly what a future reader needs. No change requested.

end-to-end/README.md
```diff
@@ -81,6 +90,13 @@ cd end-to-end
 
 Re-run the publish whenever you change something in the core build that these tests exercise;
 nothing here can detect that for you, because the whole point is that the build boundary is real.
+For the same reason a re-run against freshly published artifacts often comes back `UP-TO-DATE` —
+nothing Gradle can see about these projects changed. `DO_NOT_CACHE_TESTS` (honoured here exactly as
+in the core build, see `DEVELOPMENT.md`) forces the test tasks to run anyway:
+
+```shell
+DO_NOT_CACHE_TESTS=1 ./gradlew check
+```
```
jamesfredley, line 99: Good local guidance. **Optional companion:** set the same env var on the CI `check` step so the "republish then re-run" case cannot come FROM-CACHE in automation either.

settings.gradle
```diff
@@ -609,29 +609,6 @@ project(':grails-test-examples-jetty').projectDir = file('grails-test-examples/j
...
-// CI build, where the -SNAPSHOT BOM is available from the Apache snapshots repository.
-//
-// A newly created release branch is the same condition: the moment projectVersion is
-// bumped to a version that has never been published, the BOM import silently resolves
-// to nothing, every managed version comes back empty, and the whole build fails with
-// "Could not find <group>:<artifact>:" - before CI can ever publish the snapshot that
-// would fix it. Set -PbomSnapshotNotPublished (or add it to gradle.properties) on such
-// a branch to exclude the example until its first snapshot publish succeeds, then drop
-// the flag again. The flag is presence-based; its value is ignored.
-def bomSnapshotNotPublished = providers.gradleProperty('bomSnapshotNotPublished').isPresent()
```
jamesfredley, line 629: Removing this from core is the load-bearing part of the bootstrap fix - core no longer needs a published BOM (or `bomSnapshotNotPublished`) to configure.

**Merge-up (track explicitly):** when this change lands on `8.1.x` / `9.0.x`, delete `bomSnapshotNotPublished` from those branches' `gradle.properties` and any remaining comments. Until the example is gone from core on those lines the flag is still needed; after the forward-port it is dead config.

build-logic/settings.gradle
```diff
@@ -91,3 +92,10 @@ project(':build-logic').projectDir = file('plugins')
 
 include 'grails-docs-core'
 project(':grails-docs-core').projectDir = file('docs-core')
+
+// Due to https://github.com/gradle/gradle/issues/2986 , we can't change the global exclude for profiles, it must be done everywhere
+for (String pattern in DirectoryScanner.defaultExcludes) {
+    if (pattern.contains('gitignore') || pattern.contains('gitattributes')) {
+        DirectoryScanner.removeDefaultExclude(pattern)
+    }
+}
```
jamesfredley, line 101: **Optional:** Same note as on `end-to-end/settings.gradle` - fine for consistency across independent builds, not required for the Spring DM move. No objection to keeping.

jamesfredley (commented):
## Review (replacement for #16073)

**Direction: approve.** Moving the Spring DM fixture into `end-to-end/` is the correct replacement for #16073. It keeps the real `mavenBom` import (the point of the example) and sits behind the existing publish-then-consume boundary, so both original failures are fixed:

1. **Wrong BOM** - `exclusiveContent` on `org.apache.grails.*` forces `build/local-maven`; a remote snapshot cannot quietly win.
2. **Unpublished-version bootstrap** - removing it from root `settings.gradle` unblocks core/release builds; the E2E job publishes first, then runs `check`.

Prefer this over #16073, which stopped exercising the actual Maven BOM import path. CI is green including End to End.

Inline notes below are **optional** polish unless marked otherwise. None of them need to block merge if you want this in sooner.

### Please track (can be this PR or immediate follow-ups)

1. **Follow-up PR (do not mix into this one unless tiny):** carry the `BomPropertyOverridesPlugin` `ProjectDependency` skip + unit test from #16073 on its own. That defect (`grails.core.ROOT:grails-hibernate*-bom:unspecified`) is real and independent of this move; #16077 does not replace it.
2. **Merge-up when this lands on `8.1.x` / `9.0.x`:** drop dead `bomSnapshotNotPublished` from those branches' `gradle.properties` (and any remaining comments). It is still needed on those lines until the example is gone from core there; after this change forwards it becomes a no-op / dead property.

### Optional only

| Item | Notes |
|---|---|
| Thin resolved-version assertion on the fixture | IT proves boot; not managed-version drift |
| `DO_NOT_CACHE_TESTS=1` on the E2E `check` step | Defense in depth after republish |
| Confirm E2E is a **required** status check | Only if branch protection currently allows red E2E |
| Revert `group` `functionaltests` → `springdm` | Cosmetic |
| Matrix loss (Win / J25 / SiteMesh 2 paths) | Acceptable; one-line PR-body note is enough |
| DirectoryScanner copies outside `end-to-end` | Legitimate per-build; optional split for thinner review |

**Verdict:** Fine to merge as-is. Optional nits and the two tracked follow-ups above.

build-logic/settings.gradle
jdaugherty: It actually is required.  If using daemons, it causes an error because gradle doesn't match the daemon usage properly and these values mismatch.

end-to-end/spring-dependency-management/build.gradle
jdaugherty: I intentional changed this as I'm trying to enforce unique names where possible.

end-to-end/spring-dependency-management/build.gradle
jdaugherty: This is the downside of the spring dm plugin.  The purpose of this is to document compatibility and not try to revisit all of the problems with spring dm so I'm going to forego this.

end-to-end/settings.gradle
jdaugherty: This becomes a hard error if a daemon gets reused which is why I fixed it in this PR - I hit this locally & in CI a few times initially.

jdaugherty:
I've implemented all of @jamesfredley feedback.  Going to merge since this blocks releasing correctly.

PR #16044, opened 2026-07-22
Import log4j-bom before Spring Boot in published BOMs

jamesfredley:
## Description

Slim residual for **CVE-2026-49844** after Log4j `2.25.5` management already landed on `8.0.x`.

This PR only ensures every published Grails BOM **imports `log4j-bom` before Spring Boot** (and before inherited Grails / Micronaut platforms) so Maven consumers resolve the full Log4j family to `2.25.5`, including modules Grails does not constrain directly (for example `log4j-layout-template-json`).

Touched BOMs:

- `grails-gradle-bom`
- `grails-base-bom` / `grails-bom` (via base)
- `grails-hibernate5-bom`
- `grails-hibernate7-bom`
- `grails-micronaut-bom`
- `grails-hibernate5-micronaut-bom`
- `grails-hibernate7-micronaut-bom`

`dependencies.gradle` only adjusts `combinedPlatforms` ordering so docs / property extraction match the same Log4j-first platform order. Version pins and module constraints are unchanged.

### Out of scope (handled elsewhere)

**Jansi / Grails console / CLI classpath work is not in this PR.**

That work lives in **#16078** ([Remove Jansi and separate CLI from production classpath](https://github.com/apache/grails-core/pull/16078)), which:

- removes FuseSource Jansi entirely (does not swap to `org.jline:jansi`)
- moves console / CLI types onto the CLI tier
- enforces production classpath hygiene

Once #16078 merges, the Jansi half of the original #16044 scope is obsolete. This branch no longer documents or exempts Jansi CVEs.

### Stacking

Does **not** need to stack on #16078. The two changes are independent and can merge in either order.

## Verification

- Staged change is import-order only across the published BOM projects listed above.
- Base BOM skips `log4j2-bom` in the remaining platform loop to avoid a duplicate Maven import.
- No Jansi / vulnerability-exemption edits remain on this branch.

No user-facing API changes. No application docs required beyond what #16078 covers for CLI/Jansi.

> Generative AI tooling was used to assist with residual-scope analysis and PR preparation. The resulting changes were reviewed and verified by the submitter.

PR #16078, opened 2026-08-01
Remove Jansi and separate the CLI test tier from the production classpath

jdaugherty:
## Description

Removes Jansi from Grails entirely and separates developer tooling from what an application ships.

The PR started as a coordinate swap for **CVE-2026-8484** — a heap buffer overflow in Jansi's JNI
`ioctl()` wrapper, unfixed and unfixable: FuseSource Jansi is unmaintained, every release through
`2.4.3` is affected, and no patched version exists.

The obvious remedy was `org.jline:jansi`, the maintained fork. That turns out to be a false fix. The
fork's native `ioctl__IJ_3I` is byte-for-byte identical to FuseSource's — same
`GetIntArrayElements` with no length check before the syscall — differing only in the package inside
the JNI symbol name. The advisory names only the FuseSource coordinate, so migrating would have
silenced the scanner while shipping the same defect. `org.jline:jansi:4.3.1` still exposes
`public static native int ioctl(int, long, int[])`, and JLine has no issue or PR tracking it.

So Jansi is gone outright rather than replaced. The console renders its own escape sequences (a small,
fixed set of standard ones) and JLine remains only as the terminal library it already was.

Reviewing why Jansi was on application classpaths at all exposed the larger problem, and this PR fixes
that too: developer-only tooling was being shipped to every application. Before this change, an
application's production runtime classpath carried `grails-console`, `groovy-console`,
`groovy-groovysh`, `groovy-swing` and **13 JLine artifacts including Jansi** — a Swing console, a
Groovy shell and a terminal stack, none of which an application runs.

### Why this change is possible now

`8.0.x` is a major release branch, where the contribution policy permits breaking API changes. Moving
`grails.build.logging.GrailsConsole` and its collaborators to a companion artifact is such a change and
could not be made on a patch or minor branch. The CLI companion mechanism (`-cli` artifacts,
auto-provisioned onto `grailsCli`) already exists in 8.0.x, so the console moves onto established
infrastructure rather than new machinery.

### Issues addressed

This PR closes all five items raised in
[this comment](https://github.com/apache/grails-core/pull/16078#issuecomment-5153959171):

**1. The CLI split needs its own test source set so tests resemble production.**
`cli` output and CLI-only dependencies were being added to the plugin's `test` classpath, so unit tests
saw classes and libraries an application never receives — a test could pass against something that was
not there in production. Two dedicated source sets now exist, both created automatically by
`org.apache.grails.gradle.grails-plugin-cli` with nothing to declare:

| Source set | Purpose |
| --- | --- |
| `src/testCli` | Tests for CLI classes; runs as `testCli`, wired into `check` |
| `src/integration-test-cli` | CLI tests needing a booted application; runs as `integrationTestCli` |

`src/test` no longer sees the CLI tier at all. The direction is deliberately one-way: CLI tests may use
test fixtures, ordinary tests never see CLI classes. Roughly 70 existing specs moved accordingly.

**2. CLI-specific classes moved to the CLI source set.**
`AnsiConsoleUrlMappingsRenderer` moved to `grails-web-url-mappings`'s `src/cli` — its only consumer was
already the URL-mappings report command. An audit of every module's main sources for `GrailsConsole` and
`org.jline` references found one further cluster: the console itself. Nine classes moved from
`grails-bootstrap` to the new `grails-bootstrap-cli` companion. The app-facing `grails-bootstrap` jar
now contains **81 classes and no console classes**.

`ArtefactTypeAstTransformation` was the last main-source reference; it called
`GrailsConsole.getInstance().error(...)` on an AST-injector error path, wrapped in
`catch (Throwable) { ignore }` solely because the console could throw `NoClassDefFoundError`. It runs
inside the Groovy compiler and now uses a logger.

**3. JLine/Jansi eliminated from the production classpath — proven and tested.**
Proven by measurement and then enforced, so it cannot regress. A new `validateProductionClasspath` task
resolves `runtimeClasspath` and fails on any forbidden coordinate; it is wired into `check` and
configured on `grails-dependencies/starter-web`. It was negative-tested by deliberately reinstating
`grails-console` and confirming the failure lists every offending artifact.

**4. Jansi removed entirely rather than swapped.**
No third-party ANSI library remains. `ConsoleAnsi` (~130 lines, internal) emits the same standard
sequences, buffering style changes so consecutive ones collapse into a single SGR exactly as before.
Output is **byte-identical**: `ESC[1;31m| LABEL ESC[22;39mMSG ESC[m` for `error`, and
`ESC[1A ESC[1D ESC[0K ESC[1;33m| ESC[22;39mMSG ESC[m` for `outputMessage`. Emitting directly rather than
via JLine's `AttributedStringBuilder` was deliberate — that would have regrouped the SGR runs and made
cursor control depend on the terminfo database, which Jansi never did.

**5. `grails-console` is now CLI-only.**
It was exposed by `grails-dependencies/starter-web`, which put it on every application's runtime
classpath. It was already auto-provisioned onto `grailsCli` by the Grails Gradle plugin, so the starter
entry was both redundant and the cause. Removed. The `console` and `shell` tasks are unaffected — they
source it from `grailsCli`, which never reaches `runtimeClasspath`, so `bootRun`, `bootJar` and `bootWar`
are clean.

### Also found and fixed

- **`GrailsConsoleSpec` had never executed — anywhere.** It was gated on
  `@IgnoreIf({ !GrailsConsole.instance.isAnsiEnabled() })`, and a Gradle test worker has no tty, so the
  terminal is always `dumb` and the gate always skipped. Its own fixture also built a `dumb` terminal, so
  it would have emitted no ANSI even if it had run. It now drives an `ExternalTerminal` and runs 10
  assertions covering colour, cursor movement, reset and the disabled path. Un-gating it immediately
  surfaced two latent hazards it had been hiding.
- **ANSI configuration now uses Spring Boot's property.** Removing Jansi took its global switch with it.
  Rather than invent a Grails equivalent, `GrailsConsole` honours `spring.output.ansi.enabled`
  (`always`/`detect`/`never`), preferring what Boot bound to `AnsiOutput` and falling back to the system
  property for the CLI, which has no `Environment`. `always` is a new capability — Jansi's switch could
  only disable.
- **`withJansi` removed from 26 Logback configurations.** Logback's `ConsoleAppender` loads
  `org.fusesource.jansi.AnsiConsole` *by name*, so with Jansi gone it logged a `ClassNotFoundException`
  warning on every startup and fell back to the plain stream. Nothing is lost: colour comes from Spring
  Boot's `ColorConverter` (`%clr`), which honours the same property. Generated applications already omit
  it (#15663 — Jansi's global `System.out` replacement breaks console logging after a DevTools restart);
  only configurations carried forward were affected.
- **SBOM licence override.** `jline.version` moved to `3.30.9` to match the JLine version Groovy ships,
  which orphaned the version-pinned licence override and failed `cyclonedxDirectBom`. CycloneDX maps
  JLine's `"The BSD License"` to BSD-4-Clause (CycloneDX/cyclonedx-core-java#205); the POMs declare
  BSD-3-Clause. Override updated.
- **CLI test results were invisible.** `testCli`/`integrationTestCli` ran under `check` but were absent
  from the aggregate reports, hiding ~69 suites. Added to `grails-test-report`'s phases.

## Verification

- Full gate: `./gradlew clean aggregateViolations :grails-test-report:check --continue`
- Checkstyle, CodeNarc, PMD and SpotBugs aggregate reports: **no violations**
- **12,025 unit tests, 0 failures**; **3,618 integration tests, 0 failures**
- 67 `testCli` and 2 `integrationTestCli` suites present in the aggregate reports
- End-to-end suite against **published** artifacts: 7/7 `LegacyCommandCompatibilityIntegrationSpec`
  passing, with the Grails 7 fixture verified as built by the Grails 7 toolchain (Grails 7.0.14,
  Groovy 4.0.32, JDK 17)
- Legacy application's production `runtimeClasspath` confirmed free of JLine, Jansi, `grails-bootstrap-cli`
  and `grails-console`; its integration test classpath correctly resolves `grails-bootstrap-cli` from
  published metadata with no capability conflict

### Issue and Scope

- [ ] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. If no approved issue exists, please give background on why this change is necessary.  Tickets are preferred for release change log history.
- [x] This PR contains a **single, focused change**. Unrelated changes should be submitted as separate pull requests.

### Code Quality

### Licensing and Attribution

### Documentation

- [x] If this PR introduces breaking changes or changes that require user action during an upgrade, I have updated the **Upgrade Notes** for the corresponding version in the Grails Guide.

Upgrade notes cover: removing `runtimeOnly "org.fusesource.jansi:jansi"`; the
`spring.output.ansi.enabled` property with a before/after mapping from Jansi's own switch; removing
`withJansi` from Logback configuration; the move of `GrailsConsole` and its collaborators to
`grails-bootstrap-cli`; and the new CLI test source sets for plugin authors.

> Generative AI tooling was used to assist with scan-log analysis, CVE research, dependency resolution
> verification, and PR preparation. The resulting changes were reviewed and verified by the submitter.


jdaugherty:
This PR is attempting to adopt org.jline's fork, but additional issues found as part of this work: 
1. the cli split needs it's own test source set split so tests have a similar classpath to production 
2. any CLI specific class (i.e. AnsiConsoleUrlMappingsRenderer) needs moved to the CLI source set
3. jline/jansi needs eliminated completely from the production classpath as it's likely only used for the console / shell (this needs proven / tested)
4. It's very likely jansi can be entirely removed now since adopting the jline specific makes it a dependency of jline and we don't need a third party library 
5. the grails-console probably should be a cli only library and not on the production classpath

jdaugherty:
@jamesfredley I have proposed we hold the RC1 release until we can work through the issues in this [comment](https://github.com/apache/grails-core/pull/16078#issuecomment-5153959171)

jdaugherty:
I intend to push a change that tries to adjust the issues mentioned in the previous comments.

jdaugherty:
The biggest reason we have to do this change is we had 2 versions of jansi on our dependency list - groovy ships with one version, and then we had the old, unmaintained one.  This makes us adopt the jline version groovy uses and it ensures we don't have a CVE in our production classpath.

jamesfredley (approved):
https://github.com/apache/grails-core/pull/16044 will be stripped down to only the log4j parts.  This PR replaces everything else.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsDependencyValidatorPlugin.groovy
```diff
@@ -78,6 +87,53 @@ class GrailsDependencyValidatorPlugin implements Plugin<Project> {
                 task.onlyIf { Task t -> !shouldSkip(t.project) }
                 task.doLast { Task t -> validateDependencies(project) }
             }
+            project.tasks.register(VALIDATE_RUNTIME_TASK_NAME) { Task task ->
+                task.group = 'verification'
+                task.description = 'Fails if developer-only tooling reaches the production runtime classpath.'
+                task.onlyIf { Task t -> t.project.findProperty(FORBIDDEN_RUNTIME_EXT) != null }
+                task.doLast { Task t -> validateProductionClasspath(project) }
+            }
+            project.tasks.matching { Task candidate -> candidate.name == 'check' }.configureEach { Task check ->
```
matrei: I think `tasks.matching` is a problem for lazy configuration: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:eager_apis_to_avoid

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsDependencyValidatorPlugin.groovy
```diff
@@ -78,6 +87,53 @@ class GrailsDependencyValidatorPlugin implements Plugin<Project> {
                 task.onlyIf { Task t -> !shouldSkip(t.project) }
                 task.doLast { Task t -> validateDependencies(project) }
```
matrei: Accessing `project` at execution time. We should create a proper task class and set the values as providers.

build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/GrailsDependencyValidatorPlugin.groovy
```diff
@@ -78,6 +87,53 @@ class GrailsDependencyValidatorPlugin implements Plugin<Project> {
...
             }
+            project.tasks.register(VALIDATE_RUNTIME_TASK_NAME) { Task task ->
+                task.group = 'verification'
+                task.description = 'Fails if developer-only tooling reaches the production runtime classpath.'
+                task.onlyIf { Task t -> t.project.findProperty(FORBIDDEN_RUNTIME_EXT) != null }
+                task.doLast { Task t -> validateProductionClasspath(project) }
```
matrei: Accessing `project` at execution time.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -1940,3 +1940,193 @@ refreshes. Any context not started through `SpringApplication` — the CLI, plai
...
+
+==== 39. Jansi Removed and Developer Tooling Off the Application Classpath
+
+Grails no longer depends on a separate ANSI library. The console renders its escape sequences directly,
+and the FuseSource Jansi artifact - unmaintained, and the subject of an unfixed CVE - is gone from the
+dependency graph along with the `org.jline:jansi` fork that briefly replaced it.
+
+At the same time, developer-only tooling no longer ships with an application. `grails-console` (the Swing
+console, the Groovy shell, and the script and command runners) was previously exposed by the web starter,
+which put it - and everything it pulls in - on every application's production runtime classpath:
```
matrei: This is not true for the `grails-console`, `groovy-console`, `groovy-groovysh` and `groovy-swing`.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -1940,3 +1940,193 @@ refreshes. Any context not started through `SpringApplication` — the CLI, plai
...
+The `cli` tier now has its own test source sets:
+
+[cols="1,2"]
+|===
+|Source set |Purpose
+
+|`src/testCli`
+|Tests for CLI classes. Runs as the `testCli` task, wired into `check`.
+
+|`src/integration-test-cli`
```
matrei: Why are we diverging in the naming of these source sets? One use camelCase and one kebab-case. I think we should align on kebab-case as `integration-test` already uses it.

grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/InteractiveShell.java
```diff
@@ -62,7 +61,8 @@ public InteractiveShell(CommandLine commandLine,
     }
 
     public void start() {
-        AnsiConsole.systemInstall();
+        // no ansi stream installation: JLine's terminal enables virtual-terminal processing where the
+        // platform needs it, and picocli resolves ansi support itself via Help.Ansi.AUTO
```
matrei: This comment seems to be about a deleted line. Do we need it? Or can we re-phrase it?

grails-test-core/build.gradle
```diff
@@ -54,8 +54,7 @@ dependencies {
     api 'org.apache.groovy:groovy'
 
     // command line requirements
-    api  'org.jline:jline'
-    api 'org.fusesource.jansi:jansi'
+    compileOnly 'org.jline:jline'
```
matrei: Move to compileOnly ordered group?

matrei (approved):

jdaugherty:
assuming the tests pass, i think this is mergeable.

Issue #15497, opened 2026-03-11
Gradle Performance Improvements


jdaugherty:
### Issue description

See https://github.com/apache/grails-core/pull/15483#issue-4017789059

PR #16082, opened 2026-08-02
fix: route plugin src/main/scripts into companion -cli jars

### LINKED ISSUE #16035: Move src/main/scripts command resources out of runtime plugin artifacts
## Summary

PR #15948 successfully isolates `ApplicationCommand` implementations and their CLI-only dependencies in companion `-cli` artifacts, but the older plugin script-command channel still packages CLI-only command resources in each plugin's normal runtime JAR.

`GrailsPluginGradlePlugin.configurePluginResources` copies everything under `src/main/scripts` into the main `processResources` output at `META-INF/commands`. That main plugin JAR is then placed on the application's `runtimeClasspath` and inside deployed `bootJar`/`bootWar` artifacts.

This affects both command resource formats resolved from that directory:

- `*.groovy`, compiled at command execution time by `GroovyScriptCommandFactory`
- `*.yml` / `*.json`, parsed by `YamlCommandFactory` into multi-step commands

The CLI implementation dependencies themselves remain correctly isolated. The remaining problem is command-only code/resources in the deployed runtime plugin artifact.

Related: #15377 and #15948.

## Current behavior

The plugin convention still performs this main-source-set copy:

```groovy
TaskProvider<Copy> copyCommands = project.tasks.register('copyCommands', Copy) {
    from("${project.projectDir}/src/main/scripts")
    into("${processResources.destinationDir}/META-INF/commands")
}
processResources.dependsOn(copyCommands, copyTemplates)
```

`ClasspathCommandResourceResolver` later scans `classpath*:META-INF/commands/*.<extension>` from the command runner's context classloader. `runCommand` intentionally combines the application runtime classpath with `grailsCliClasspath`, so discovery currently depends on the command resources remaining in runtime plugin JARs.

## Reproduction

The existing legacy-command fixture contains:

```text
grails-test-examples/legacy-commands-plugin/src/main/scripts/hello-legacy-script.groovy
```

Build and inspect the plugin and deployable application archives:

```bash
./gradlew :grails-test-examples-legacy-commands:bootJar
jar tf grails-test-examples/legacy-commands-plugin/build/libs/grails-test-examples-legacy-commands-plugin-0.0.1.jar
jar tf grails-test-examples/legacy-commands/build/libs/grails-test-examples-legacy-commands-0.1.war
```

The plugin JAR contains:

```text
META-INF/commands/hello-legacy-script.groovy
```

The deployable WAR contains that plugin JAR:

```text
WEB-INF/lib/grails-test-examples-legacy-commands-plugin-0.0.1.jar
```

The same copy task accepts YAML and JSON files without filtering, and `YamlCommandFactory` resolves those formats from the same `META-INF/commands` location.

Control checks confirm that PR #15948's dependency isolation is working:

```bash
./gradlew :grails-test-examples-legacy-commands:dependencyInsight \
  --dependency grails-shell-cli --configuration runtimeClasspath
./gradlew :grails-test-examples-legacy-commands:dependencyInsight \
  --dependency grails-core-cli --configuration runtimeClasspath
```

Both report no matching runtime dependency. The deployable WAR also contains no `grails-shell-cli`, `grails-core-cli`, `grails-core-cli-legacy`, or `grails-profiles-*` JAR.

## Expected behavior

When a plugin publishes a companion CLI artifact, command resources from `src/main/scripts` should be placed in that companion artifact and excluded from the normal runtime plugin JAR. The companion remains available through `grailsCliClasspath`, preserving command discovery and execution without shipping command-only resources in deployed applications.

## Suggested coverage

Add a functional fixture containing both:

```text
src/main/scripts/example-script.groovy
src/main/scripts/example-multi-step.yml
```

Verify that:

1. The normal plugin JAR and consuming `bootJar`/`bootWar` do not contain either command resource.
2. The companion `-cli` JAR contains both resources under `META-INF/commands`.
3. `grails <command>` and `runCommand` still discover and execute both commands through `grailsCliClasspath`.
4. `grails-shell-cli`, profile artifacts, and companion CLI dependencies remain absent from `runtimeClasspath` and the deployed archive.

## Out of scope / verified isolated

- YAML commands packaged inside profile JARs under `META-INF/grails-profile/commands` are resolved by the CLI profile repository and were absent from the deployed application archive.
- `ServiceCommandFactory`, its `META-INF/services` registrations, and built-in profile command implementations remain in `grails-shell-cli`; that artifact was absent from the application's runtime classpath and deployed WAR.

jamesfredley:
## Summary

Addresses #16035 on top of the CLI split (#15948) and legacy command compatibility (#16011).

When a plugin applies `org.apache.grails.gradle.grails-plugin-cli`, `src/main/scripts` command resources (`*.groovy`, `*.yml`, `*.json`) are packaged into the companion `-cli` jar as `META-INF/commands` instead of the runtime plugin jar. That keeps command-only resources off `runtimeClasspath` / `bootJar` / `bootWar` while remaining available through `grailsCliClasspath`.

### Backwards compatibility

Plugins **without** a companion keep the historical behavior: scripts still ship in the runtime jar so unmigrated Grails 7 plugins and `legacyCommandSupport` consumers continue to discover them (including the end-to-end `legacy-commands` fixture).

### Incremental migration

After enabling `grails-plugin-cli`, a `cleanStaleRuntimeCommandResources` task wipes `META-INF/commands` under the configured `processResources` destination before main resources are processed. That removes prior `src/main/scripts` copies left by non-companion builds; `processResources` then restores any hand-authored `src/main/resources/META-INF/commands`.

## Test plan

- [x] `:grails-gradle-plugins:test --tests PluginScriptCommandPackagingSpec` (companion + non-companion jar contents, stale cleanup, hand-authored main resources, templates)
- [x] Related CLI specs: `CliAutoDiscoverySpec`, `CliCompanionPublishingSpec`, `LegacyCommandTaskDiscoverySpec`, `BomCliMultiprojectRaceFunctionalSpec`

## Docs

- `providingBasicArtefacts.adoc`
- `creatingCustomCommands.adoc`
- `upgrading80x.adoc` (CLI companion section)

Fixes #16035


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
```diff
@@ -244,22 +246,69 @@ class GrailsPluginGradlePlugin extends GrailsGradlePlugin {
...
+                // incremental jars do not keep shipping them; processResources then re-runs (its
+                // outputs changed) and restores any hand-authored src/main/resources/META-INF/commands.
+                // Use the configured processResources destination, not a hard-coded build path.
+                TaskProvider cleanStaleRuntimeCommands = project.tasks
+                        .register('cleanStaleRuntimeCommandResources') { Task cleanTask ->
+                            cleanTask.outputs.upToDateWhen { false }
+                            cleanTask.doLast {
+                                project.delete(new File(processResources.destinationDir, 'META-INF/commands'))
+                            }
+                        }
```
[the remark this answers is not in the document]
jdaugherty: When designing gradle tasks its advised to never force cleans. Instead set the output directories unique by tasks and gradle will ensure no stale state is left

jdaugherty (changes requested):
My only concern with this is the regression first found by copilot.  Clean tasks can't be forced out of date and always wired

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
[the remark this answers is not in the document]
jamesfredley: Addressed in 7a3049d707.

Removed `cleanStaleRuntimeCommandResources` entirely (no more `outputs.upToDateWhen { false }` wired into `processResources`/`jar`).

Instead:
1. `copyCommands` / `copyTemplates` are now **`Sync`** tasks with **unique** dirs under `build/tmp/`
2. `processResources` / `processCliResources` consume those via `from(...)` so Gradle owns each path
3. When a companion is present, the runtime `jar` filters `META-INF/commands` to only hand-authored `src/main/resources/META-INF/commands` entries - covers leftover files from prior non-companion packaging **without** a forced clean task

`PluginScriptCommandPackagingSpec` still asserts companion vs non-companion packaging, unique Sync output path, and that seeded stale runtime leftovers are excluded from the runtime jar while hand-authored commands remain.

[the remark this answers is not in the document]
jamesfredley: Agreed - forced clean tasks were the wrong tool. Fixed in 7a3049d707 per your guidance:

- Unique Sync output dirs for `copyCommands` / `copyTemplates` (`build/tmp/grails-plugin-*`)
- Downstream `process*Resources` consume via `from(...)` so Gradle tracks ownership
- No clean task forced out-of-date or wired into every build

For leftovers already sitting under `processResources.destinationDir` from the old side-write design, the runtime jar filters non-hand-authored `META-INF/commands` entries when a companion is present (packaging-time filter, not a clean task).

jamesfredley:
## Review feedback addressed (7a3049d707)

Responding to **@jdaugherty**, Copilot, and Bito on the forced-clean regression:

### Removed
- `cleanStaleRuntimeCommandResources` with `outputs.upToDateWhen { false }` wired into `processResources` / `jar`

### Replaced with Gradle-owned unique outputs
- `copyCommands` / `copyTemplates` are **`Sync`** tasks writing only to `build/tmp/grails-plugin-commands` and `build/tmp/grails-plugin-templates`
- Companion: `processCliResources.from(copyCommands) { into 'META-INF/commands' }`
- No companion: `processResources.from(copyCommands) { into 'META-INF/commands' }`
- Templates always: `processResources.from(copyTemplates) { into 'META-INF/templates' }`

### Migration leftovers (no forced clean)
When a companion is present, the runtime `jar` excludes `META-INF/commands/*` entries that are not hand-authored under `src/main/resources/META-INF/commands`. That drops old side-written script copies without disabling incremental builds.

### Verification
`:grails-gradle-plugins:test --tests PluginScriptCommandPackagingSpec` green (companion routing, non-companion legacy packaging, unique Sync dir, seeded stale leftovers excluded, hand-authored kept).

jamesfredley:
@jdaugherty I am also creating a PR with a Gradle skill based on the last 200 PRs, our gradle code standards and then lastly the Gradle 9 docs, so we can avoid gradle syntax concerns, hopefully, going forward.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
```diff
@@ -244,22 +248,71 @@ class GrailsPluginGradlePlugin extends GrailsGradlePlugin {
...
             processResources.setDuplicatesStrategy(DuplicatesStrategy.INCLUDE)
-            processResources.dependsOn(copyCommands, copyTemplates)
+            processResources.from(copyTemplates) { into 'META-INF/templates' }
+
+            if (hasCliCompanion) {
+                SourceSet cliSourceSet = project.extensions.getByType(SourceSetContainer)
+                        .getByName(GrailsCliArtifactGradlePlugin.CLI_SOURCE_SET_NAME)
+                ProcessResources commandResources = (ProcessResources) project.tasks
+                        .getByName(cliSourceSet.processResourcesTaskName)
+                commandResources.setDuplicatesStrategy(DuplicatesStrategy.INCLUDE)
```
jdaugherty: DuplicatesStrategy.INCLUDE is a strong smell here - it means we've not properly setup the input.  We should error by default as it's a gradle configuration problem

jdaugherty:
> @jdaugherty I am also creating a PR with a Gradle skill based on the last 200 PRs, our gradle code standards and then lastly the Gradle 9 docs, so we can avoid gradle syntax concerns, hopefully, going forward.

To be clear, this isn't a syntax issue.  It's a design issue.  I support such a skill though.

jamesfredley:
@jdaugherty "tomato, tomato", if Gradle had backported fixes, instead of only moving forward and made all these optimizations automatic, without syntax changes, we would have saved ~1,000 hrs on grails-core.  

The build system should take maybe 5% of our time, it currently takes 25-33%, which is crazy.

Hopefully the skill can solve this, since I see all of this as completely wasted time, that should be spent on real code in the framework.  

I still want the build to be 99% as fast as possible, just can't stand how hard Gradle make it to get there.

jdaugherty:
> @jdaugherty "tomato, tomato", if Gradle had backported fixes, instead of only moving forward and made all these optimizations automatic, without syntax changes, we would have saved ~1,000 hrs on grails-core.
> 
> The build system should take maybe 5% of our time, it currently takes 25-33%, which is crazy.
> 
> Hopefully the skill can solve this, since I see all of this as completely wasted time, that should be spent on real code in the framework.
> 
> I still want the build to be 99% as fast as possible, just can't stand how hard Gradle make it to get there.

I'm skeptical the skill will solve it.  The crux of this issue is Gradle decided to maintain backwards compatibility so it worked, even when the threading model changed.  Worse, most of the examples are mixed so it's never clear.  Gradle did put together this link though: https://docs.gradle.org/current/userguide/lazy_configuration.html  We can try it though.

jdaugherty:
@jamesfredley if you really wanted to fix this, we should adopt the fixes mentioned here: https://github.com/apache/grails-core/pull/15483#issue-4017789059  Having gated validations is better than instruction when it comes to AI's doing the right thing.  https://github.com/apache/grails-core/issues/15497 is the ticket I created to track this.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
```diff
@@ -244,22 +248,71 @@ class GrailsPluginGradlePlugin extends GrailsGradlePlugin {
...
+                    jarTask.eachFile { FileCopyDetails details ->
+                        String path = details.path.replace('\\', '/')
+                        if (path.startsWith('META-INF/commands/') && path != 'META-INF/commands/') {
+                            String relative = path.substring('META-INF/commands/'.length())
+                            if (!new File(handAuthoredCommands, relative).file) {
+                                details.exclude()
+                            }
+                        }
+                    }
+                }
```
jdaugherty: The unique `Sync` output for `copyCommands` is the right change. This filter isn't.

It runs for every entry in the runtime jar on every build and does a filesystem stat per `META-INF/commands` entry, and the only thing it accomplishes is hiding files that a *previous version* of this plugin left in `build/resources/main`. On a clean build there is nothing to hide — with a companion, `src/main/scripts` never reaches `build/resources/main` at all. I removed the filter (and the `seedStaleRuntimeCommands` fixture) and both `PluginScriptCommandPackagingSpec` cases still pass, `RUNTIME_HAS_HAND_AUTHORED=true` included: `src/main/resources/META-INF/commands/hand-authored.yml` ships in the runtime jar on its own because it is an ordinary main resource. So the filter's net effect is to exclude everything *else*.

Two problems with that:

1. It doesn't remove the stale state, it masks the archive. The leftovers stay in `build/resources/main`, which is `sourceSets.main.output`, so they remain on the plugin's own `test` / integration-test / `bootRun` classpath. I confirmed they survive the migration there.
2. It silently drops legitimate entries. Any `META-INF/commands` file not backed by a real file under `src/main/resources/META-INF/commands` is excluded, including command resources a plugin generates with its own task and wires in through `processResources.from(...)`. No warning and no way to opt out.

A `build/` directory holding output from an older version of the build logic is what `clean` is for. A packaging-time filter is a clean task in disguise, just paid on every build instead of once. Please drop it and note the one-time `clean` in `upgrading80x.adoc` next to the `grails-plugin-cli` instructions.

grails-gradle/plugins/src/test/resources/test-projects/plugin-script-commands/plugin-with-cli/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+        def dir = staleDir.get().asFile
+        dir.mkdirs()
+        new File(dir, 'example-script.groovy').text = "description('stale') {}\n"
+        new File(dir, 'removed-before-migration.groovy').text = "description('gone') {}\n"
+    }
+}
+
+tasks.named('jar', Jar).configure { jarTask ->
+    jarTask.dependsOn('seedStaleRuntimeCommands')
+}
```
jdaugherty: `seedStaleRuntimeCommands` writes into `build/resources/main`, which is `processResources`' output directory — the exact side-write this PR removes from the plugin itself. The fixture manufactures the only condition under which the runtime jar filter does anything, so `RUNTIME_HAS_SCRIPT=false` and `RUNTIME_HAS_REMOVED=false` are asserting the workaround rather than the packaging behavior.

Drop this block along with the filter. With both gone the remaining assertions still pass and cover what the change is actually for: scripts land only in the companion, and hand-authored `META-INF/commands` plus templates stay in the runtime jar.

jamesfredley:
Let's do https://github.com/apache/grails-core/issues/15497, but that might be 5-10% of Gradle build maintenance pain.

jdaugherty (changes requested):
Switching to sync is not a valid solution because it will break end apps (including the asset plugin usage).  Copy will ensure a clean base as long as gradle understands that each task has a unique output - the end result is the processResources task should be only a combination of unique outputs of other tasks.  The AI did a better job pointing the details out here so I'm posting it's review.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
jamesfredley: Addressed in the latest push - pure unique-`Copy` design, no packaging-time filter:

1. **`Copy` not `Sync`** - unique dirs under `build/tmp/grails-plugin-{commands,templates}`; end apps / asset-pipeline keep `tasks.named('copyTemplates', Copy)`
2. **`process*Resources` only composes unique outputs** via `from(...)` - no side-writes into `processResources.destinationDir`
3. **Dropped `DuplicatesStrategy.INCLUDE`** - overlapping paths fail as configuration errors
4. **Dropped the runtime jar `eachFile` filter** and `seedStaleRuntimeCommands` fixture
5. **`upgrading80x`**: one-time `./gradlew clean` after first applying `grails-plugin-cli` on a previously built tree

`PluginScriptCommandPackagingSpec` asserts companion vs non-companion packaging + unique Copy output path only.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
jamesfredley: Agreed - `DuplicatesStrategy.INCLUDE` was covering up overlapping inputs. Removed. With unique Copy outputs composed only via `from(...)`, duplicate paths should surface as a configuration error rather than being silently merged.

grails-gradle/plugins/src/test/resources/test-projects/plugin-script-commands/plugin-with-cli/build.gradle
jamesfredley: Dropped `seedStaleRuntimeCommands` and the filter it was justifying. The fixture now only asserts the packaging contract on a clean composition of unique Copy outputs.

jamesfredley:
## Design aligned with jdaugherty review + gradle skill principles

Latest commit applies the **unique task output composition** model (not Sync, not forced clean, not packaging-time filters):

| Concern | Resolution |
|---------|------------|
| Forced clean / always-run tasks | Never added back |
| `Sync` breaks end apps / asset plugin | **`Copy`** with unique `build/tmp/...` dirs |
| `DuplicatesStrategy.INCLUDE` smell | **Removed** - fail on overlap |
| Jar `eachFile` filter masks stale `build/` | **Removed** - document one-time `clean` |
| `seedStaleRuntimeCommands` side-writes | **Removed** |

`processResources` / `processCliResources` are only a combination of unique upstream Copy outputs via `from(...)`. That is the Gradle design point: ownership via unique outputs, not cleanup theatre.

Tests: `PluginScriptCommandPackagingSpec` green. `:grails-spring-security-ui:copyTemplates` configures as `Copy` again.

jdaugherty (commented):
I created https://github.com/apache/grails-core/pull/16086 to implement the feedback on this PR

jamesfredley:
Merged https://github.com/apache/grails-core/pull/16086 (`82ebbd775e`) into this branch.

**Review summary (acceptable):** jdaugherty's follow-up correctly fixes a real regression from routing templates through `processResources` - the `**/*.gsp` exclusion was silently emptying GSP templates from scaffolding and spring-security-ui jars.

Design matches the unique-output model we converged on, with the right refinements:
- Staging dirs attached via `sourceSet.output.dir(builtBy:)` so they never re-enter `processResources` filters
- Still `Copy` (not `Sync`) for consumer BC
- `doFirst` wipe of own destination only when the task runs (up-to-date preserved)
- Stronger TestKit coverage (GSP survival, staging outside process*Resources, delete-without-clean)

Local: all 5 `PluginScriptCommandPackagingSpec` features green.

jdaugherty (approved):
assuming no major test failures, I'm good to merge this.

PR #16086, opened 2026-08-02
fix: keep plugin templates out of the processResources filters

jdaugherty:
See https://github.com/apache/grails-core/pull/16082


jamesfredley:
Reviewed and merged into https://github.com/apache/grails-core/pull/16082 (fast-forward to `82ebbd775e`).

This is the right fix for GSP templates being dropped by the processResources `**/*.gsp` exclude. Thanks.

PR #16089, opened 2026-08-03
Fall back to the WatchService watcher instead of polling on macOS

codeconsole:
```
2026-08-03T13:20:35.127-07:00  INFO 8719 --- [  restartedMain] [                                                 ] org.grails.io.watch.DirectoryWatcher     : Exception while trying to load WatchServiceDirectoryWatcher (this is probably Java 6 and WatchService isn't available). Falling back to PollingDirectoryWatcher.

java.lang.reflect.InvocationTargetException
        at org.grails.io.watch.DirectoryWatcher.<init>(DirectoryWatcher.java:62)
        at grails.boot.GrailsApp.enableDevelopmentModeWatch(GrailsApp.groovy:219)
        at grails.boot.GrailsApp.run(GrailsApp.groovy:122)
        at grails.boot.GrailsApp.run(GrailsApp.groovy:464)
        at grails.boot.GrailsApp.run(GrailsApp.groovy:453)
        at com.l3me.Application.main(Application.groovy:55)
Caused by: java.lang.NoClassDefFoundError: io/methvin/watchservice/MacOSXListeningWatchService
        at org.grails.io.watch.MacOsWatchServiceDirectoryWatcher.<init>(MacOsWatchServiceDirectoryWatcher.java:71)
        ... 6 common frames omitted
Caused by: java.lang.ClassNotFoundException: io.methvin.watchservice.MacOSXListeningWatchService
        ... 7 common frames omitted
```

Java 6 reference is pretty ancient
> (this is probably Java 6 and WatchService isn't available)

## Problem

`DirectoryWatcher` probes for JNA to decide whether the native macOS FSEvents watcher is usable:

```java
Class.forName("com.sun.jna.Pointer");
```

JNA is not what that watcher needs. `MacOsWatchServiceDirectoryWatcher` needs `io.methvin:directory-watcher`, which `grails-bootstrap` declares `compileOnly` and therefore never puts on an application's runtime classpath.

Applications routinely acquire JNA transitively — `org.testcontainers:testcontainers` → `com.github.docker-java:docker-java-transport-zerodep` → `net.java.dev.jna:jna` is a common path. On macOS those applications are sent down a load of `MacOsWatchServiceDirectoryWatcher` that cannot succeed. The resulting `NoClassDefFoundError` is caught by a single catch-all that drops straight to `PollingDirectoryWatcher`, so they silently poll every watched directory on a 1s timer instead of receiving file system events:

```
Exception while trying to load WatchServiceDirectoryWatcher (this is probably Java 6
and WatchService isn't available). Falling back to PollingDirectoryWatcher.
Caused by: java.lang.NoClassDefFoundError: io/methvin/watchservice/MacOSXListeningWatchService
```

The fallback ordering makes this worse than having no JNA at all: without JNA the `else` branch selects `WatchServiceDirectoryWatcher` and works correctly. Adding an unrelated dependency downgrades file watching.

## Change

Probe for `io.methvin.watchservice.MacOSXListeningWatchService` — the class that actually backs the native watcher — and degrade in steps: native macOS → JDK `WatchService` → polling.

Polling is now only reached if the JDK cannot supply a `WatchService`, which cannot happen on the Java 21 baseline. The message no longer claims Java 6. On macOS without the optional dependency the fallback is logged at warn: the JDK supplies no native `WatchService` there — `FileSystems.getDefault().newWatchService()` returns `sun.nio.fs.PollingWatchService`, and `WatchServiceDirectoryWatcher` registers without a sensitivity modifier, so it polls at the default 10s — which is a real reload latency cost with a one dependency fix, and the message names it. The `ClassNotFoundException` behind it stays at debug.

## Verification

Selected delegate, comparing this branch against `8.0.x` under three classpaths on macOS:

| classpath | before | after |
|---|---|---|
| JNA present, `directory-watcher` absent | `PollingDirectoryWatcher` | `WatchServiceDirectoryWatcher` |
| neither present | `WatchServiceDirectoryWatcher` | unchanged |
| both present | `MacOsWatchServiceDirectoryWatcher` | unchanged |

The broken case is fixed and both already-working cases are untouched.

`DirectoryWatcherSpec` covers the three outcomes and fails on the current implementation (2 of 3 cases). It also covers that the advice is emitted on macOS without the optional library and stays off platforms it cannot help. Reproducing the reported shape needs JNA present and `directory-watcher` absent, so JNA is added to the test runtime classpath.

`:grails-bootstrap:test` (145 tests), `codeStyle` and `rat` all pass.

## Note

An alternative fix would be promoting `io.methvin:directory-watcher` to a real runtime dependency of `grails-bootstrap`, giving every macOS application native watching. That adds JNA and a native library to all application classpaths, so I kept the dependency optional here and fixed only the selection logic. Happy to take that route instead if preferred.


jamesfredley (approved):
We could backport to 7.x.x, since it has the same code.

grails-bootstrap/src/main/groovy/org/grails/io/watch/DirectoryWatcher.java
```diff
@@ -45,32 +45,57 @@ public class DirectoryWatcher extends Thread {
...
+            // MacOsWatchServiceDirectoryWatcher delegates to io.methvin's MacOSXListeningWatchService,
+            // an optional dependency of this module. Probe for that class rather than for JNA: JNA is
+            // frequently present transitively (Testcontainers, docker-java, ...) without the watcher
+            // library, and using it as the signal sends those applications down a load that can't succeed.
+            // A LinkageError means the class is present but its own dependencies (JNA) are not, which is
+            // equally unusable, so treat both as simply unavailable.
+            Class.forName("io.methvin.watchservice.MacOSXListeningWatchService");
+        } catch (ClassNotFoundException | LinkageError e) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Native macOS file event watching is unavailable. Add 'io.methvin:directory-watcher' to the classpath for faster file watching.", e);
```
jdaugherty: Shouldn't this be warn?  Once you added the right library, doesn't this resolve itself?

codeconsole: Agreed — changed to warn in ab2b2afb3e.

Worth surfacing: the JDK supplies no native `WatchService` on macOS. `FileSystems.getDefault().newWatchService()` returns `sun.nio.fs.PollingWatchService`, and `WatchServiceDirectoryWatcher` registers without a sensitivity modifier, so the fallback polls at the default 10s. That is a real reload latency cost with a one dependency fix, so the message now states the consequence as well as naming the dependency — and yes, adding the library silences it.

The `ClassNotFoundException` stays at `LOG.debug` so the warn line remains a single actionable sentence rather than a stack trace, which is what made the message on `8.0.x` noisy.

`DirectoryWatcherSpec` now also covers that the advice is emitted on macOS without the library, and that it stays off platforms it cannot help.

jdaugherty (approved):

PR #16099, opened 2026-08-06
change cooldown for dependabot to 7 days

paulk-asert:
this buys a little extra time for GitHub actions to be further battle tested by other projects and to be added to the ASF infra allow list

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,8 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    cooldown:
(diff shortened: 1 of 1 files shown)


paulk-asert:
I only applied to the github actions and gradle ecosystems but feel free to apply further if you see benefit. I think GitHub themselves moved to a 3-day cooldown as default and I think there is some kind of expedited path for CVE/security dependencies but I couldn't find the source I read originally.

jdaugherty (approved):

PR #16102, opened 2026-08-06
Move i18n onto Spring Boot's MessageSource

codeconsole:
Grails registered its own `messageSource` bean (`PluginAwareResourceBundleMessageSource`), so Spring Boot's `MessageSourceAutoConfiguration` backed off. Grails then discovered bundles by scanning `classpath*:*.properties` at runtime — which GraalVM native images cannot do, and which costs start-up time on the JVM. `spring.messages.*` was ignored entirely as a result.

Spring Boot now owns the `messageSource` bean outright. Grails supplies only the one thing Boot cannot know: which base names the application and its plugins contribute. That is recorded at build time, so nothing scans the classpath at runtime.

## `spring.messages.*` now works

All of Boot's message-source properties apply, where previously they were silently ignored:

```yaml
spring:
    messages:
        encoding: UTF-8
        cache-duration: 5s
        fallback-to-system-locale: false
        use-code-as-default-message: true
```

Grails contributes its own defaults at the lowest precedence — `fallback-to-system-locale: false`, `spring.messages.encoding` from `grails.views.gsp.encoding`, and a short `cache-duration` in development so bundle edits still reload without a restart.

`grails.i18n.cache.seconds` and `grails.i18n.filecache.seconds` are removed in favour of `spring.messages.cache-duration`.

## Base names are discovered for you

You do not normally set `spring.messages.basename`. The Grails Gradle plugin records each artifact's bundles into `META-INF/grails/i18n.properties`, and an `EnvironmentPostProcessor` composes the effective list at boot: your application's base names first, then each plugin's.

Setting it yourself still works and is how you reach a bundle outside `grails-app/i18n` — whatever you declare is kept and outranks the discovered names:

```yaml
spring:
    messages:
        basename: config/i18n/custom
```

## Plugin bundles must be namespaced

A plugin's base names must be its plugin name, or that name followed by a hyphen:

```
grails-app/i18n/spring-security-core.properties
grails-app/i18n/spring-security-core_fr.properties
grails-app/i18n/spring-security-core-validation.properties
```

Spring resolves a base name to the first matching resource on the classpath, so two plugins sharing a base name would shadow one another rather than merge. The plugin build now rejects a bundle outside the plugin's namespace. Applications are unaffected — there is only one application, so its base names cannot collide.

Within this repository that meant renaming three bundles: `spring-security-oauth2`, `spring-security-ui` (whose dotted `messages.spring-security-ui` base name is unreachable under `ResourceBundle`, which maps `.` to `/`), and the `loadfirst` test plugin.

## Ambiguous file names

An application may use any base name, with one reservation: a name ending in a valid locale identifier is ambiguous, since `api_fr.properties` reads as base name `api` in French. Declare the base name when the inference is wrong:

```groovy
grails {
    i18n {
        basenames = ['api', 'api_errors']
    }
}
```

## Native image

Because the bundles are known ahead of time, Grails registers the GraalVM resource hints for them — covering plugin base names and any base name an application configures itself, neither of which Boot's own registrar handles (it registers two hardcoded `messages*` patterns and derives nothing from the configured base names).

## Limitations

- **Adding or removing a base name requires a restart.** Spring Boot reads the configured base names once, when it builds the message source. Editing values in an existing bundle, and adding a locale file for a base name that already exists, both still work without one.
- **A plugin shipping a root `messages.properties` no longer contributes messages.** It previously acted as a fallback behind the application's bundle; that behaviour depended on Grails' own merge tier and cannot be reproduced under first-match-wins resolution.
- **Every bundle needs a locale-independent file.** Boot's auto-configuration only activates when a configured base name has an unsuffixed bundle, and it contributes the whole `messageSource` bean — so an application shipping only `messages_de.properties` would have had no message source at all. The build now rejects that rather than letting it fail at runtime.
- **Only `grails-app/i18n` is indexed.** Bundles elsewhere need an explicit `spring.messages.basename`; they are still covered by the native-image hints, which read the effective property rather than the descriptors.
- **The GraalVM bundle-hint behaviour is not verified by a native test.** Hints use resource patterns rather than `registerResourceBundle`, following Spring Boot's own choice for this case; that is a design signal, not something this PR proves with a native image build.

Inter-plugin precedence follows the reverse of Grails' plugin topological order, preserving what the previous message source did when two plugins define the same code. An application's own bundle always overrides every plugin's.


codeconsole:
# Spring Boot MessageSource Migration: Comparison

Even without native-image/AOT support, the new implementation is a net improvement architecturally and operationally, with a few meaningful compatibility and development-workflow costs.

## Better than the previous implementation

### Standard Spring Boot behavior

Previously Grails replaced Boot's message source, so standard properties such as these were mostly irrelevant:

```properties
spring.messages.basename
spring.messages.encoding
spring.messages.cache-duration
spring.messages.fallback-to-system-locale
spring.messages.always-use-message-format
spring.messages.use-code-as-default-message
spring.messages.common-messages
```

The new implementation delegates resolution to Boot's stock `ResourceBundleMessageSource`, making Grails behave like a normal Boot application and reducing custom framework behavior users must learn.

### Much less Grails-owned code

It deletes roughly 1,000 lines from `PluginAwareResourceBundleMessageSource` and Grails' fork of `ReloadableResourceBundleMessageSource`. Those classes duplicated caching, merging, discovery, formatting, and plugin-specific loading behavior that Spring already handles.

The replacement is primarily integration metadata:

```text
bundle files → generated descriptor → Boot basename configuration
```

### No runtime wildcard scanning

Previously Grails searched for `classpath*:*.properties`, while binary plugins scanned their resource roots for `*.properties`. Gradle now records known bundles during the build, and runtime discovery uses the exact descriptor name:

```text
META-INF/grails/i18n.properties
```

Benefits beyond AOT include:

- lower startup work;
- unrelated properties files cannot accidentally become message basenames;
- no `searchClasspath` performance compromise;
- behavior no longer depends on resource-pattern resolver peculiarities.

### Deterministic discovery and precedence

The new implementation explicitly defines:

```text
user-configured basenames
→ application bundles
→ plugins in reverse topological order
```

It preserves the legacy winner for plugin message-code collisions while making the rule testable. Classloader enumeration order no longer determines behavior.

### Better plugin isolation

Plugins now use namespaced bundles such as:

```text
spring-security-core.properties
spring-security-core-validation.properties
```

rather than all shipping `messages.properties`. This prevents silent shadowing and makes bundle ownership obvious. Multiple logical bundles per plugin remain supported.

### Consistent encoding

The old application and plugin paths did not necessarily use the same encoding. Binary plugin messages could use the platform encoding while application messages used configured UTF-8. Boot now applies one `spring.messages.encoding` value consistently to all basenames.

### Better configuration capabilities

Boot supplies useful behavior Grails did not cleanly expose, particularly:

```properties
spring.messages.common-messages
spring.messages.always-use-message-format
spring.messages.use-code-as-default-message
```

Applications can also add explicit basenames outside `grails-app/i18n` without replacing the generated application and plugin list.

For migration convenience, `grails.i18n.cache.seconds` remains temporarily supported in reload mode. Grails translates it to `spring.messages.cache-duration`, emits a deprecation warning, and lets an explicitly configured Boot property win. This preserves an existing development-time tuning without retaining the old message-source implementation.

### Available locales match actual resolution

Previously locale discovery independently scanned resources and could disagree with the effective plugin set. Now basename composition and `AvailableLocaleResolver` use the same descriptor model. Filtered, evicted, failed, or deliberately excluded plugins cannot advertise locales whose messages will not resolve.

### Earlier failure for invalid packaging

The Gradle build now rejects:

- locale-only bundles without a base file;
- colliding plugin basenames;
- malformed or ambiguous locale naming;
- duplicate plugin or application descriptors.

Previously many such problems appeared only as missing or shadowed messages at runtime.

### Stronger tests

The new implementation covers real Gradle application and plugin wiring, descriptor regeneration, plugin-name derivation, custom declarations, namespace enforcement, precedence, formatted and unformatted lookups, plugin filtering, reload behavior, and configuration precedence.

## Worse or more restrictive

### Existing plugin artifacts must be rebuilt and renamed

This is the biggest cost. A plugin that currently ships `messages.properties` must move to a namespaced basename. Old plugin artifacts without generated descriptors no longer contribute messages automatically.

Backward binary compatibility is explicitly out of scope, but this is still an ecosystem migration.

The source migration is less restrictive than the artifact break suggests: namespaced bundles also work with Grails 7's classpath-scanning implementation. A plugin can therefore rename its bundles once and build the same source for both Grails 7 and Grails 8; only the Grails 8 artifact gains and requires the generated descriptor.

### Bundle naming is more constrained

The build must infer whether an underscore introduces a locale. For example, `api_fr.properties` could mean basename `api` in French or unsuffixed basename `api_fr`. The new implementation reserves valid locale suffixes and requires explicit configuration for ambiguous cases:

```groovy
grails {
    i18n {
        basenames = ['api_fr']
    }
}
```

This is deterministic, but requires more ceremony than the previous permissive scanning model.

### More build-time coupling

Message availability now depends on the Grails Gradle plugin generating and packaging the descriptor. Nonstandard build systems must reproduce the descriptor format, and malformed descriptors fail startup. This trades runtime flexibility for build-time correctness.

### Structural reload is weaker

Editing an existing bundle still reloads automatically. However:

- a new locale requires descriptor regeneration before appearing in `<g:localeSelect>`;
- a new basename requires restart;
- removing a basename may stop resolution after cache invalidation while its metadata remains until restart.

A newly added basename already effectively required reinitialization previously, so the available-locale descriptor dependency is the clearest development-time regression.

### More startup bootstrap integration

The custom message source disappears, but the new system introduces an `EnvironmentPostProcessor`, descriptor parsing, effective plugin filtering, a Gradle descriptor task, and an AOT processor. This is less semantic duplication overall, but behavior now spans build time, bootstrap, and application context.

### Strict descriptor failures may expose unusual classpaths

The new code rejects multiple application descriptors, duplicate plugin descriptors, and unsupported descriptor versions. This improves correctness but may expose ambiguous layered applications or unusual test fixtures that previously started.

### `ResourceBundleMessageSource` is less reload-oriented

Boot uses `ResourceBundleMessageSource`, not `ReloadableResourceBundleMessageSource`. Grails compensates by setting a short cache duration in reload mode and clearing JDK bundle caches. This is tested, but less direct than calling `clearCache()` on a reloadable source.

## Overall comparison

| Area | Previous | New implementation |
|---|---|---|
| Message implementation | Grails-specific | Standard Boot/Spring |
| Discovery | Runtime wildcard scans | Build-generated descriptors |
| Plugin collision handling | Runtime merge | Unique basenames |
| Configuration | Grails properties/custom behavior | `spring.messages.*` |
| Encoding | Potential app/plugin mismatch | Unified |
| Startup cost | Classpath/resource scanning | Exact descriptor lookup |
| Error detection | Often runtime or silent | Build/startup failure |
| Existing plugin compatibility | Broad artifact compatibility | Rebuild required; migrated source can support Grails 7 and 8 |
| Content hot reload | Supported | Supported |
| Structural hot reload | Limited but more dynamic locale scan | Descriptor/restart dependent |
| Nonstandard builds | More tolerant | Must generate metadata |
| Maintenance | Large custom message-source stack | Smaller integration layer |

## Verdict

Excluding AOT, the new implementation is still substantially better for framework maintainability, startup determinism, configuration consistency, and failure diagnostics.

Its main disadvantages are the deliberate old-artifact compatibility break and the shift from permissive runtime discovery to stricter build-time metadata. The temporary cache-property bridge and cross-version-compatible source layout reduce upgrade friction without carrying the old message-source implementation forward. For a major-version modernization where backward plugin artifact compatibility is explicitly not required, that is a favorable trade.

sbglasius:
This is a yellow flag ⚠️ **Existing plugin artifacts must be rebuilt and renamed** as we are still struggling to keep up with plugin changes.

Would it be possible for the code to automatically prefix bundles coming from a plugin with the plugin-name. That way it might be possible to use plugins without rebuilding?

grails-i18n/src/main/groovy/org/grails/plugins/i18n/EffectiveI18nDescriptors.java
```diff
@@ -0,0 +1,141 @@
...
+                pluginDescriptors.put(PluginUtils.normalizePluginName(descriptor.name()), descriptor);
+            }
+        }
+
+        List<I18nDescriptor> effectivePlugins = new ArrayList<>();
+        if (includePluginBundles) {
+            for (String pluginName : pluginNamesInTopologicalOrder) {
+                I18nDescriptor descriptor = pluginDescriptors.get(PluginUtils.normalizePluginName(pluginName));
+                if (descriptor != null) {
+                    effectivePlugins.add(descriptor);
```
sbglasius, line 99: **Plugin descriptors are keyed by normalized name in a plain HashMap via put(), so two discovered plugins whose names normalize to the same key (e.g. differing only in hyphenation/case) silently overwrite one another instead of erroring.**

Two plugins on the classpath (e.g. "springSecurityCore" reported by PluginDiscovery vs. a differently-cased/hyphenated variant recorded in another plugin's descriptor) normalize to the same key; the second `put()` silently drops the first plugin's `basenames/locales` from message resolution and the language selector, with no warning — unlike `I18nDescriptors.rejectAmbiguousClasspath`, which only compares raw (unnormalized) names and would not catch this case.

sbglasius (approved):

grails-i18n/src/main/groovy/org/grails/plugins/i18n/EffectiveI18nDescriptors.java
codeconsole: Good catch — fixed in 0221bb01d8.

The two sites used different key forms: this class keys plugins by normalized name, while `I18nDescriptors.rejectAmbiguousClasspath` counted raw ones. So `spring-security-core` and `springSecurityCore` passed the ambiguity check as distinct, then collapsed to one key here and lost a plugin's basenames and locales to a silent `put()` — the classpath-order-dependent outcome that check exists to prevent. It now counts by normalized name and reports both raw spellings.

One detail: case alone doesn't collide. `normalizePluginName` only rewrites names containing a hyphen and returns others untouched, so `SpringSecurityCore` stays distinct from `springSecurityCore`. The reachable collision is hyphenated vs camel-case.

Severity is low in practice — the Gradle plugin always emits the hyphenated form via `GrailsNameUtils.getPluginName`, so two same-named plugins already collided on the raw name and were caught. Reaching this needs a hand-edited or foreign-tool descriptor. Worth fixing regardless, since the check's javadoc claims it rejects "two plugins sharing a name", and under the matching semantics that actually apply, these do.

Covered by two tests: the collision throws naming both spellings, and genuinely distinct plugins still load.

borinquenkid (dismissed):
│ Note: These technical observations were identified during detailed code analysis by Antigravity and reviewed and endorsed by @borinquenkid as mandatory requirements prior to merging PR #16102 into    
  │ 8.0.x.                                                                                                                                                                                                  
  ──────                                                                                                                                                                                                    
  ### Required Changes Checklist                                                                                                                                                                            
                                                                                                                                                                                                            
  #### 1. Reduce Development Reload Cache Duration to 1s                                                                                                                                                    
                                                                                                                                                                                                            
  • File: I18nEnvironmentPostProcessor.java:3491                                                                                                                                                            
  • Requirement: Change DEVELOPMENT_CACHE_DURATION from "5s" to "1s".                                                                                                                                       
  • Rationale: A 1s TTL provides near-instant hot-reloading feedback when editing .properties files during grails run-app, while still preventing redundant ResourceBundle.getBundle() disk/classpath       
  lookups within a single HTTP request render.                                                                                                                                                              
                                                                                                                                                                                                            
    // Change in I18nEnvironmentPostProcessor.java                                                                                                                                                          
    private static final String DEVELOPMENT_CACHE_DURATION = "1s";                                                                                                                                          
  ──────                                                                                                                                                                                                    
  #### 2. Handle Multiple Application Descriptors Gracefully (Merge & Warn)                                                                                                                                 
                                                                                                                                                                                                            
  • File: I18nDescriptors.java:3358-3369                                                                                                                                                                    
  • Requirement: Replace the fatal IllegalStateException when applications.size() > 1 with a logger.warn(...) and merge all discovered application basenames.                                               
  • Rationale: Multi-module Gradle projects, enterprise monorepos, and test fixture JARs (java-test-fixtures) frequently place more than one application descriptor on the test classpath. Crashing the JVM 
  breaks legitimate multi-project test suites. Merging basenames sequentially into EffectiveI18nDescriptors preserves translations without dropping keys.                                                   
  ──────                                                                                                                                                                                                    
  #### 3. Soften BootstrapContext Restrictions for Standalone Test Environments                                                                                                                             
                                                                                                                                                                                                            
  • File: I18nEnvironmentPostProcessor.java:3638-3646                                                                                                                                                       
  • Requirement: If PluginDiscovery is not registered in bootstrapContext (such as in @SpringBootTest, ApplicationContextRunner, or lightweight Spock contexts), issue a warning and sort plugin descriptors
  alphabetically rather than throwing a fatal IllegalStateException.                                                                                                                                        
  • Rationale: Framework users and plugin authors test components using standard Spring Boot test runners where GrailsBootstrapRegistryInitializer is not active. Throwing an unhandled exception prevents  
  isolated unit/integration testing.                                                                                                                                                                        
  ──────                                                                                                                                                                                                    
  #### 4. Establish a Legacy Plugin Compatibility / Migration Strategy                                                                                                                                      
                                                                                                                                                                                                            
  • Files: I18nDescriptors.java, I18nEnvironmentPostProcessor.java                                                                                                                                          
  • Requirement: Provide a fallback mechanism or explicit opt-in setting (e.g. grails.i18n.legacy-classpath-scan=true or automatic fallback detection) for un-indexed plugin JARs compiled against Grails 7 
  or earlier.                                                                                                                                                                                               
  • Rationale: Without a fallback, every pre-existing published plugin JAR in the Grails ecosystem will have its message bundles silently omitted when consumed in Grails 8 applications.                   
  ──────                                                                                                                                                                                                    
  #### 5. Provide End-to-End Native Image Verification                                                                                                                                                      
                                                                                                                                                                                                            
  • Location: end-to-end/ test suite                                                                                                                                                                        
  • Requirement: Add an automated native-image test verifying that I18nRuntimeHintsProcessor successfully registers namespaced plugin resource bundles in a compiled GraalVM binary.                        
  • Rationale: Native image support is cited as the core architectural justification for eliminating runtime classpath scanning; this capability must be verified empirically with a native build test prior
  to merge.

codeconsole:
Thanks @borinquenkid  — I went through all five.  Here is the rationale behind all 5.  Please LMK if you think they are still valid based on my response.

**1. Cache duration 5s → 1s — I'd keep 5s.** It isn't arbitrary: it's the historical Grails default. `I18nGrailsPlugin` set `field('cacheSeconds', int).value(Settings.I18N_CACHE_SECONDS, '5')`, and this PR deliberately preserves that so an upgrading application sees the same reload timing it had. It's also what the deprecated `grails.i18n.cache.seconds` alias documents. Anyone who prefers tighter feedback sets `spring.messages.cache-duration: 1s`, which now works where it silently did nothing before.

**2. Multiple application descriptors → warn and merge — I'd keep the error.** Merging doesn't remove the ambiguity, it hides it: two applications both shipping `messages` would have their precedence decided by classpath enumeration order, which is exactly the nondeterminism this design exists to remove (and what @sbglasius flagged for plugin names earlier in this PR). The scenario also isn't demonstrated — the full suite, 16k+ tests including every multi-module test-example, passes with the check in place. The error already names both descriptors and the likely cause. If a real multi-application classpath turns up, I'd rather fix it with evidence in hand.

**3. Soften the BootstrapContext requirement — I don't think this case is reachable, and the fallback would be worse.** `GrailsBootstrapRegistryInitializer` is registered in `grails-core`'s `META-INF/spring.factories` as a `BootstrapRegistryInitializer`, so `SpringApplication` loads it automatically — `@SpringBootTest` included. `ApplicationContextRunner` doesn't invoke `EnvironmentPostProcessor`s at all, so the branch never runs there either. More importantly, the proposed fallback of sorting alphabetically would silently produce *wrong* precedence: correctness here depends on reverse topological order, which is what reproduces the behaviour of the message source this replaces. A warning plus wrong messages is worse than a clear failure.

**4. Legacy plugin fallback — out of scope, deliberately.** A Grails 7 plugin JAR already doesn't work in Grails 8 for reasons unrelated to i18n; the upgrade guide states it directly in several places — "the plugin must be rebuilt against the typed metadata API", "the plugin must be rebuilt for Grails 8", "the plugin's trait must be recompiled with Groovy 5", and plugin authors "must compile against Spring Framework 7". Adding a runtime classpath scan to preserve message bundles for artifacts that can't load anyway would reintroduce the exact scan this PR removes — the one thing that makes the i18n path AOT-hostile — for no working use case. Plugins are recompiled for 8 regardless; a namespaced bundle works unchanged under both 7 and 8 (documented in the upgrade section), so the migration is a one-time rename.

**5. Native image end-to-end verification — not currently possible.** There is no native build path for a Grails application on `8.0.x` yet: no `org.graalvm.buildtools` plugin, no `SpringBootAotPlugin`/`processAot` wiring, and no native target in `end-to-end/`. That's what #16094 is for, and it's still open. This PR doesn't claim native support works end to end — the Limitations section says so explicitly, and calls out that the resource-pattern choice follows Spring Boot's own precedent rather than being proven by a native build. Once #16094 lands, an end-to-end native test is the right follow-up, and this PR is what makes the i18n side ready for it.

borinquenkid:
> Thanks @borinquenkid — I went through all five. Here is the rationale behind all 5. Please LMK if you think they are still valid based on my response.
> 
> **1. Cache duration 5s → 1s — I'd keep 5s.** It isn't arbitrary: it's the historical Grails default. `I18nGrailsPlugin` set `field('cacheSeconds', int).value(Settings.I18N_CACHE_SECONDS, '5')`, and this PR deliberately preserves that so an upgrading application sees the same reload timing it had. It's also what the deprecated `grails.i18n.cache.seconds` alias documents. Anyone who prefers tighter feedback sets `spring.messages.cache-duration: 1s`, which now works where it silently did nothing before.
Works for me
> 
> **2. Multiple application descriptors → warn and merge — I'd keep the error.** Merging doesn't remove the ambiguity, it hides it: two applications both shipping `messages` would have their precedence decided by classpath enumeration order, which is exactly the nondeterminism this design exists to remove (and what @sbglasius flagged for plugin names earlier in this PR). The scenario also isn't demonstrated — the full suite, 16k+ tests including every multi-module test-example, passes with the check in place. The error already names both descriptors and the likely cause. If a real multi-application classpath turns up, I'd rather fix it with evidence in hand.
Work for me
> 
> **3. Soften the BootstrapContext requirement — I don't think this case is reachable, and the fallback would be worse.** `GrailsBootstrapRegistryInitializer` is registered in `grails-core`'s `META-INF/spring.factories` as a `BootstrapRegistryInitializer`, so `SpringApplication` loads it automatically — `@SpringBootTest` included. `ApplicationContextRunner` doesn't invoke `EnvironmentPostProcessor`s at all, so the branch never runs there either. More importantly, the proposed fallback of sorting alphabetically would silently produce _wrong_ precedence: correctness here depends on reverse topological order, which is what reproduces the behaviour of the message source this replaces. A warning plus wrong messages is worse than a clear failure.
Works for me
> 
> **4. Legacy plugin fallback — out of scope, deliberately.** A Grails 7 plugin JAR already doesn't work in Grails 8 for reasons unrelated to i18n; the upgrade guide states it directly in several places — "the plugin must be rebuilt against the typed metadata API", "the plugin must be rebuilt for Grails 8", "the plugin's trait must be recompiled with Groovy 5", and plugin authors "must compile against Spring Framework 7". Adding a runtime classpath scan to preserve message bundles for artifacts that can't load anyway would reintroduce the exact scan this PR removes — the one thing that makes the i18n path AOT-hostile — for no working use case. Plugins are recompiled for 8 regardless; a namespaced bundle works unchanged under both 7 and 8 (documented in the upgrade section), so the migration is a one-time rename.
Works for me.
> 
> **5. Native image end-to-end verification — not currently possible.** There is no native build path for a Grails application on `8.0.x` yet: no `org.graalvm.buildtools` plugin, no `SpringBootAotPlugin`/`processAot` wiring, and no native target in `end-to-end/`. That's what #16094 is for, and it's still open. This PR doesn't claim native support works end to end — the Limitations section says so explicitly, and calls out that the resource-pattern choice follows Spring Boot's own precedent rather than being proven by a native build. Once #16094 lands, an end-to-end native test is the right follow-up, and this PR is what makes the i18n side ready for it.
Open an Issue in Grails for that.

borinquenkid (approved):
Only open the ISSUE on the native test

codeconsole:
Correcting my earlier reply on point 5: #16094 has since merged, so the premise there ("no native build path exists") is out of date. A Grails application can now apply `org.graalvm.buildtools.native` and the Gradle plugin cooperates with it, and `processAot` is wired up.

That makes the ask feasible rather than impossible, so it deserved a better answer than the one I gave. Filed as #16176.

I still don't think it belongs in this PR, for two reasons. Nothing in the repository invokes `nativeCompile`, `nativeRun` or `nativeTest` today, and no CI workflow provisions GraalVM — `grails-test-examples/aot` verifies `processAot` output and a JVM startup check, both on a normal JDK. So this isn't "add a test", it's "establish native-image CI", which would serve the AOT cache work, GORM, GSP and anything else shipping hints, not just i18n. #16094 itself, which added the native support, didn't add a native-compile test either — reasonably, for the same reason.

The issue captures the specific i18n assertions worth making once that infrastructure exists, including the one open question this PR does carry: `I18nRuntimeHintsProcessor` registers resource patterns rather than calling `registerResourceBundle`, following Spring Boot's own choice for this case, and only a native build can confirm that was the right call.

PR #16111, opened 2026-08-06
Configure generated applications for component scanning

### LINKED ISSUE #14914: Update Forge and Profiles to configure component-scan for base package
### Issue description

annotate Application with `@ComponentScan(value = 'com.mycompany.myapp.custom')`

This removes the need to define the beans one by one for the base application package in `grails-app/conf/spring/resources.groovy`

jamesfredley:
## Summary

- add explicit base-package Spring component scanning to Forge-generated applications
- keep base and plugin profile Application skeletons aligned with Forge output
- verify all application types and standard Spring component injection without `resources.groovy`
- document the generated scanning default separately from Grails artefact scanning

## Verification

- `:grails-forge-core:check :test-core:check`
- `:grails-profiles-base:check :grails-profiles-plugin:check`
- `GrailsApplicationSpec`: 18 tests, 0 failures
- `CreateAppSpec` generated-source matrix: 4 tests, 0 failures
- mandatory Oracle and Codex review gate: GREEN

Closes #14914

### DIFF
--- a/grails-doc/src/en/guide/conf/applicationClass/customizing.adoc
+++ b/grails-doc/src/en/guide/conf/applicationClass/customizing.adoc
@@ -23,7 +23,9 @@ There are several ways in which you can customize the `Application` class.
 === Customizing Scanning
 
 
-By default Grails will scan all known source directories for controllers, domain class etc., however if there are packages in other JAR files you wish to scan you can do so by overriding the `packageNames()` method of the `Application` class:
+Newly generated Grails applications include `@ComponentScan(value = 'your.application.package')` on the `Application` class. This scans standard Spring components in the application's base package and its subpackages.
+
+Grails scans its known source directories separately for Grails artefacts such as controllers and domain classes. To add packages to that Grails artefact scan, override the `packageNames()` method of the `Application` class:
 
 [source,groovy]
 ----
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/lang/groovy/application.rocker.raw
@@ -31,10 +31,12 @@ import groovy.transform.CompileStatic
 import io.micronaut.spring.boot.starter.EnableMicronaut
 
 }
+import org.springframework.context.annotation.ComponentScan
 @if(features.contains("spring-boot-starter-security")) {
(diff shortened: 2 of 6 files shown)


codeconsole:
Not sure if adding component scanning is a good idea.   It's definitely not needed for anything created by the app.

The new bean dsl is a replacement for `resources.groovy`.  You can now just specify whatever beans you want in `Application.groovy` in the beans closure.

matrei:
- I would like to keep the starter templates as lean as possible.
- Is scanning needed by any Forge feature?
- Even though I'm sure people use it, this is not the typical documented Grails way of defining beans in an application, and it's easy enough to add if you want it.

codeconsole:
I agree with @matrei I would like the starter app as lean as possible, but I also think it would be useful to expose all the possible settings as a feature.   The base app should focus on doing as much as possible purely by convention. I am a huge proponent of convention over configuration. 

@jamesfredley I think a good feature that you might like is like an opt-in "kitchen sink" feature that adds this and every possible default to `application.yml`

jamesfredley:
Moving back to draft since there is not agreement.  It seemed that https://github.com/apache/grails-core/issues/14914 was cleared, but that is not the case.

borinquenkid:
This PR closes the loop on #14914 by adding `@ComponentScan` to Forge-generated and profile-skeleton `Application` classes, with test coverage across all application types (`WEB`, `REST_API`, `PLUGIN`, `WEB_PLUGIN`) plus an end-to-end TestKit test proving a plain `@Component` bean outside Grails' artefact conventions gets autowired.

Progress on the underlying gating concern (finishing the move off `doWithSpring`/legacy bean loading before turning on component-scan by default) is further along than it may appear: `@GrailsBeans` now compiles bean-wiring into real `@AutoConfiguration` classes, and `grails-core`, `grails-domain-class`, `grails-cache`, `grails-mail`, `grails-sitemesh3`, `grails-url-mappings`, `grails-i18n`, and `grails-databinding` have all migrated.

Still on the legacy `doWithSpring` mechanism: `grails-spring-security` (core, cas, ldap, acl, oauth2, rest-* variants), the Hibernate/Mongo/Neo4j datastore plugins, `grails-quartz`, `grails-gsp`, and `grails-scaffolding`. Worth confirming none of those register a `@Component`/`@Service`-style bean that would now get picked up twice (once via their `doWithSpring` block, once via the new base-package scan) before this leaves draft.

PR #16113, opened 2026-08-07
Speed up Gradle builds by reusing cacheable outputs in CI

jamesfredley:
## Summary

Reduces Gradle build time on CI and locally without dropping any test family, matrix leg, or validation.

Opened as a draft so the timing can be measured against the historical baseline before asking for review. If the CI wall-clock does not actually improve, this PR will be closed.

## Why

Two sources of avoidable work were measured on `8.0.x`:

1. **CI discarded the remote cache.** Most jobs passed `--rerun-tasks`, which forces *every* selected task, not just tests. A sampled normal job logged 500+ `FROM-CACHE` outcomes; the forced job logged none. The flag was throwing away the Develocity benefit purely to guarantee tests actually ran.
2. **A dedicated forced-rerun job sat on the critical path.** `buildRerunTasks` cost a **78.4 minute median**. Across **253 completed `8.0.x` runs** it produced no confirmed unique correctness catch (the one apparent catch was a runner shutdown).

Baseline for comparison: **148.6 min median** for the last 20 successful push runs, **106.7 min median** across the recent PR cohort.

## Changes

**1. Limit the default local test fork count**

The local default was `availableProcessors * 3/4`, i.e. 15 concurrent JVMs on a 20-thread machine, which oversubscribed CPU and RAM. Now defaults to 4. Measured on a 14-core / 20-thread workstation with identical warm inputs and tests forced:

| Suite | 15 forks | 4 forks |
|---|---|---|
| `:grails-test-suite-web:test` | 114.5s | **79.4s** |
| `:grails-gradle-plugins:test` | 232.0s | **217.7s** |

(2 forks was slower than 4 at 89.1s.) CI defaults are unchanged, and `-PmaxTestParallel=N` still overrides.

**2. Reuse cacheable build outputs in normal CI runs**

- CI test jobs use `DO_NOT_CACHE_TESTS=1` instead of `--rerun-tasks`: tests are still forced to execute, but compilation and resource processing stay cacheable.
- `gradle/spring-security-test-config.gradle` now honours that variable, so its suites are forced like every other test family. Verified: two consecutive runs both executed the `Test` task while 209 other tasks stayed up-to-date.
- The whole-build forced rerun moves to a new weekly + `workflow_dispatch` `gradle-build-validation.yml` running the identical command, so nothing stops being validated - it just leaves every PR and push.
- Aggregate Groovydoc runs once on the Ubuntu / JDK 25 leg instead of all four core legs. JDK 25 builds the full graph including the Micronaut island that a sub-25 JDK prunes, so the documented surface is a superset of what JDK 21 covered. A Groovydoc failure still fails that leg and still blocks publishing.

**3. Document the controls** in `DEVELOPMENT.md`.

## Verification

- `./gradlew validateActions` - all 24 workflows compliant.
- `./gradlew clean aggregateViolations :grails-test-report:check --continue` - Checkstyle, CodeNarc, PMD and SpotBugs all report **no violations**.
- Post-change scoped runs green: web suite 78.7s, `grails-gradle` 204.7s, `:grails-redis:test`, and `:grails-spring-security:test -PonlySpringSecurityTests`.
- Two full-suite failures were triaged and reproduced in isolation, and **neither is caused by this branch**:
  - `grails-test-examples-mail` - pre-existing Windows-only CRLF assertion (`Hello\r\nWorld!` vs `Hello\nWorld!`), reproduces identically in isolation.
  - `grails-test-examples-spring-security-ui-extended` - Selenium deleted the shared session after 32s of inactivity under full-repo load; passes all 63 tests in isolation.

## Note for a follow-up

Unrelated to this PR: the existing `-PonlyRedisTests` selector composes conflicting `onlyIf` predicates, so the `grails-redis` plugin module's unit tests are skipped in the Redis CI lane even under `--rerun-tasks`. Left alone here because fixing it would *add* coverage and confound this timing experiment.

### DIFF
--- a/.github/workflows/gradle-build-validation.yml
+++ b/.github/workflows/gradle-build-validation.yml
@@ -1,0 +1,52 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Gradle Build Validation"
+on:
+  schedule:
+    - cron: '0 4 * * 1'
+  workflow_dispatch:
+permissions:
+  contents: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+jobs:
+  validate:
+    name: "Forced Rerun Build Validation"
+    runs-on: ubuntu-24.04
+    steps:
+      - name: "📥 Checkout repository"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: "☕️ Setup JDK"
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          distribution: liberica
+          java-version: 21
+      - name: "🐘 Setup Gradle"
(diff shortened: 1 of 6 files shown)


jamesfredley:
CI completed successfully, but the performance acceptance criterion was not met. The main CI run took 106m35s versus the recent PR cohort median of 106.7m, an approximately seven-second (0.1%) difference that is well within normal run-to-run noise. The unchanged Windows/JDK 25 core build remained the critical path at 106m31s. Some non-critical jobs and the local fork benchmarks improved, but this experiment required a legitimate end-to-end improvement rather than isolated wins. Closing the draft as a failed experiment, as planned.

PR #16114, opened 2026-08-07
Fix Groovy compilation on Gradle 9.7 and upgrade to 9.7.1

codeconsole:
## Problem

`GrailsGradlePlugin` assigned `groovyOptions.configurationScript` from a `doFirst`. Gradle finalizes task properties before any task action runs, so from **Gradle 9.7** — where `GroovyCompileOptions` [became a lazy property](https://docs.gradle.org/9.7.0/release-notes.html) — every Groovy compilation in every Grails project fails:

```
> The value for task ':compileGroovy' property 'groovyOptions.configurationScriptFile' is final and cannot be changed any further.
```

There is no user-side workaround and no Gradle opt-out flag.

## Fix

A `GrailsCompilerConfigScriptTask` produces the combined script, and the compile task depends on it. The script is built from `grails { }` state and reads no classpath, so the task takes the Grails half as one `@Input` string and the build's own `configurationScript`, when there is one, as an `@InputFile`: up-to-date checked, validated by Gradle when that file is missing, and it stores and reuses a configuration cache entry. Every `GroovyCompile` is covered — the source-set ones get a generator up front, and one a build registers directly gets it when its dependencies are resolved.

`GrailsGroovyCompilerConfigSpec` covers the lifecycle: a `GroovyCompile` registered after the project is evaluated, one no source set owns, a `configurationScript` the build assigns from any configuration callback, one produced by another task, one that does not exist, a generator per source set, and a plugin's version and name staying inputs of its compile task.

## Behaviour changes

Both are in the star imports, which stay opt-in — `importGrailsCommonAnnotations` and `starImports` default to off.

- `importGrailsCommonAnnotations` now adds `grails.gorm.annotation` and `grails.plugin.scaffolding.annotation` without first probing the compile classpath. A star import of a package that is not on the classpath contributes no classes and is not an error in Groovy.
- `importJavaTime` is removed. Groovy 5 imports `java.time` by default, so it did nothing on Grails 8. A build that sets it fails with `Could not set unknown property 'importJavaTime'`; drop the line.

Both are documented in the Grails 8 upgrade guide, section 53.

## Gradle 9.7.1

Grails builds on 9.7.1: `.sdkmanrc`, `gradleToolingApiVersion`, and the wrapper for root, `build-logic`, `grails-gradle`, `grails-forge`, `end-to-end`, plus the `grails-shell-cli` gradle-sample fixture. `end-to-end/legacy-g7-command-plugin` stays on Gradle 8.14.5, the version Grails 7 pins for that fixture.

Applications created by the forge and the profile CLIs get a 9.7.1 wrapper as well. They can only move to 9.7 together with the fix above; without it the first `compileGroovy` fails.


grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -204,33 +199,63 @@ class GrailsGradlePlugin implements Plugin<Project> {
...
+        // GroovyCompileOptions became a lazy property — "The value for task ':compileGroovy' property
+        // 'groovyOptions.configurationScriptFile' is final and cannot be changed any further." Once
+        // the property is assigned during configuration, Gradle also treats the script as an input
+        // file that has to exist before the compile task runs, which a producing task guarantees
+        // across a `clean build` and a doFirst cannot.
+        //
+        // Wiring happens after evaluation so a configurationScript set by the build script is already
+        // in place and gets folded into the combined file rather than clobbered. Names are read via
+        // TaskCollection.names, which does not realize the tasks.
+        project.afterEvaluate {
```
jamesfredley: This one-time snapshot regresses the previous live `configureEach` behavior. A `GroovyCompile` registered from a later `projectsEvaluated` callback is absent from `names` and receives no generator/configuration script; I reproduced this against this head (`LATE_GROOVY_UNCONFIGURED=true`, with no `generateLateGroovyGrailsCompilerConfig` task). Please retain live wiring without registering a generator from inside the task-container callback (for example, a deferred task rule plus string `dependsOn`) and add a late-task TestKit case.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -204,33 +199,63 @@ class GrailsGradlePlugin implements Plugin<Project> {
...
+
+                TaskProvider<Task> generateGroovyCompilerConfig = project.tasks.register("generate${compileTaskName.capitalize()}GrailsCompilerConfig") { Task t ->
+                    t.description = "Generates the Grails Groovy compiler configuration script for ${compileTaskName}"
+                    t.outputs.file(groovyCompilerConfigFile)
+                    // Generating the script needs the resolved compile classpath, and declaring that as
+                    // an input would pull the runtimeClasspath into this task's up-to-date check. Since
+                    // the inputs to this task would effectively be the runtimeClasspath, dependency
+                    // problems can arise if another task changes the runtimeClasspath. Generating the
+                    // script is cheap, so skip state tracking and regenerate on every build instead.
+                    t.doNotTrackState('Depends on the resolved compile classpath; cheap to regenerate')
```
jamesfredley: `doNotTrackState` suppresses validation but does not establish execution ordering. The `doLast` calls `getGroovyCompilerScript`, which scans `compileTask.classpath`, while this generator has no dependency on tasks producing those classpath entries. I forced a classpath producer to run after the generator on a clean build; the generator omitted `grails.gorm.annotation`, then `compileGroovy` failed with `unable to resolve class CreatedDate`. Please model the classpath/base script as inputs and depend on their producer tasks.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -204,33 +199,63 @@ class GrailsGradlePlugin implements Plugin<Project> {
...
+                        combinedFile.write(combinedScripts)
+                    }
+                }
 
-                    ${configuredScript?.trim() ?: ''}
-                """
-                combinedFile.write(combinedScripts)
-                c.groovyOptions.configurationScript = combinedFile
+                compileTask.configure { GroovyCompile c ->
+                    userConfigurationScript[0] = c.groovyOptions.configurationScript
```
jamesfredley: This capture/replace is not stable for later configuration. When a user assigns `configurationScript` from a later `projectsEvaluated` callback, that assignment overwrites the combined script after this action. Reproduced result: the user's `LocalDate` import worked, but the generated Grails `CreatedDate` import disappeared and compilation failed. Please capture the final configured script at a graph-safe point (or wire it provider-first) and cover a later-callback assignment in TestKit.

jamesfredley (commented):
I reproduced three lifecycle failures against this exact head. Inline comments include the commands/observable failures and focus only on correctness gaps.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: Confirmed and fixed in 8b8d063316. `TaskCollection.names` is an immutable snapshot, so a `GroovyCompile` registered after the `afterEvaluate` got no generator and no configuration script, silently.

Generator tasks are now registered from `sourceSets.configureEach`, which is live — source sets are what create `GroovyCompile` tasks, so a source set added from `projectsEvaluated` is covered.

One correction on the suggested remedy: a task rule plus string `dependsOn` does not work. Rules are not consulted when resolving a `dependsOn` name — the build fails at graph construction with `Task with name 'generateCompileGroovyGrailsCompilerConfig' not found`. And registering the generator from inside the task-container callback is what the snapshot was avoiding in the first place: it throws `TaskCreationException: Could not create task ':compileGroovy'`. Hooking the source-set container sidesteps both.

Test: `GrailsGroovyCompilerConfigSpec` — *a GroovyCompile registered after the project is evaluated still gets a generator*.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: Right that `doNotTrackState` establishes no ordering — the generator declared no dependencies at all while its action resolves `compileTask.classpath` and reads jar entries from it. Fixed in 8b8d063316: it now depends on the classpath's own build dependencies.

I used `dependsOn` rather than declaring the classpath as an input, because modelling it as an input reintroduces exactly what `doNotTrackState` is there to prevent — it drags the runtimeClasspath into the up-to-date check. `dependsOn` gets the ordering without the tracking.

For the record on severity: I could reproduce the missing edge, but not the failure. In the natural task graph Gradle schedules the producing jar before the generator, and the two probed classes normally come from external jars already in the module cache. That matches your note that you had to force the ordering. Still worth closing — it was a real latent hazard.

Test: `GrailsGroovyCompilerConfigSpec` — *the generator runs after the tasks that produce the compile classpath*.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: Confirmed and fixed in 8b8d063316. Capturing during `afterEvaluate` meant an assignment from any later callback simply overwrote the combined file and the Grails imports disappeared with no error.

Capture and assignment now happen at `taskGraph.whenReady` — the last point before execution and after every configuration callback has run, so the user's final value is what gets folded in. The property is still assignable there on 9.7.

Scope note for anyone reading later: the ordinary path — assigning `configurationScript` directly in the build script — was already merging correctly. Only assignment from a later callback was affected.

Test: `GrailsGroovyCompilerConfigSpec` — *a configurationScript assigned from a later callback is folded in, not clobbered*.

codeconsole:
Thanks — all three were real. Fixed in 8b8d063316 with a regression test each in `GrailsGroovyCompilerConfigSpec`; each fails against the previous commit.

Common root cause: the fix had to move from execution-time reads to configuration-time snapshots, because Gradle 9.7 forbids assigning `groovyOptions.configurationScript` from `doFirst`. The snapshots were narrower than the `doFirst` they replaced. The wiring is now live (`sourceSets.configureEach`), ordered against the classpath it reads, and captures the user's script at `taskGraph.whenReady` instead of during `afterEvaluate`.

`:grails-gradle:test`, `codeStyle`, and a full root `build -PskipTests` are green on 9.7.0.

sbglasius (approved):

jdaugherty:
@codeconsole Looking back the code history: we used a separate task previously and the commit message where we reverted this was: 

            fix: rework the configuration for the compileGroovy task to prevent task dependency errors if the runtimeclasspath is changed

How does this change then solve this?

jdaugherty:
Do we know why the compile configuration can no longer be changed at execution?  This seems like an artificial limitation by Gradle and I'm wondering if we should instead open a bug ticket on this.

codeconsole:
**runtimeClasspath.** The generator reads no classpath and declares no inputs:

```
GENERATOR_DEPENDS_ON_COMPILE_CLASSPATH=false
GENERATOR_DEPENDS_ON_RUNTIME_CLASSPATH=false
GENERATOR_DECLARED_INPUT_FILES=0
```

9a529d1996 declared `configurations.runtimeClasspath` as a task input, so it was resolved and fingerprinted for the up-to-date check. This one takes the finished script as a single `@Input` string, built from `grails { }` state rather than by probing jars, so no configuration reaches its dependency chain or its up-to-date check. Unchanged build is `UP-TO-DATE`; changing the imports regenerates.

**Setting it at execution.** Not artificial: `getConfigurationScript()` is annotated `@ReplacedBy("configurationScriptFile")`, and the replacement is a `RegularFileProperty`. Lazy task properties finalize before execution so inputs can be fingerprinted for up-to-date checks and the configuration cache. The branch uses that property.

**Behaviour changes**, both in the star imports — opt-in, and the flags default to off:

- `importGrailsCommonAnnotations` adds `grails.gorm.annotation` and `grails.plugin.scaffolding.annotation` without probing the compile classpath first. A star import of an absent package contributes no classes and is not an error in Groovy.
- `importJavaTime` is removed. Groovy 5 imports `java.time` by default, so it did nothing on Grails 8. Upgrade guide section 45.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -2256,7 +2362,278 @@ resolved unambiguously before can become ambiguous.
 Removing the packages from `grails.spring.bean.packages`, or the stray classes from those packages, restores
 the previous set of beans.
 
-==== 45. `importJavaTime` Removed
+==== 45. GORM API Resolution Moves to a Registry
```
sbglasius, line 2448: Where does all these lines come from and why are they related to this PR? Seems odd.

sbglasius (dismissed):

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: They come from `8.0.x`, not this PR. Section 45 arrived in #16066 (`2926a0f644`), and 46/47 likewise predate this branch — all three sit at this same line on `upstream/8.0.x` today. The branch picked them up when it was brought up to date.

This PR's net change to the file is 39 added lines — section 48, `importJavaTime` Removed:

```
git diff upstream/8.0.x HEAD -- grails-doc/src/en/guide/upgrading/upgrading80x.adoc
 1 file changed, 39 insertions(+)
```

The Files changed tab shows that one hunk. "Changes since your last review" will include the `8.0.x` merge, since the approval predates it.

codeconsole:
https://github.com/gradle/gradle/issues/38930

sbglasius (approved):

matrei (commented):
AI review:

# Review Findings

The current head `7714cded41` addresses the previously reported lifecycle issues, but the following regressions remain.

## [P2] Standalone `GroovyCompile` tasks lose Grails compiler configuration

**File:** `grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy:306-312,335-339`

Generator tasks are registered only from `SourceSetContainer` entries:

```groovy
sourceSets.configureEach { SourceSet sourceSet ->
    registerGroovyCompilerConfigGenerator(project, sourceSet.getCompileTaskName('groovy'))
}
```

The plugin still configures every `GroovyCompile` task at lines 314-327, but a standalone task that is not associated with a source set never gets a corresponding generator. The `whenReady` callback then skips it because the generated task name is absent.

This regresses the previous behavior, where the `doFirst` compiler-script wiring was attached to every `GroovyCompile` task. A manually registered task such as:

```groovy
tasks.register('compileCustomGroovy', GroovyCompile) {
    source('src/custom/groovy')
    destinationDirectory.set(layout.buildDirectory.dir('classes/custom'))
    classpath = sourceSets.main.compileClasspath
}
```

will not receive `starImports`, `importGrailsCommonAnnotations`, a plugin's compiler metadata, or a user-provided `configurationScript` merged with the Grails script. This is particularly significant for `GrailsPluginGradlePlugin`, whose override adds `projectVersion`, `projectName`, and `isPlugin` metadata to the compiler script.

Please preserve compiler-script wiring for standalone `GroovyCompile` tasks, while avoiding task-container mutation during task configuration. Add a TestKit regression test for a standalone task.

## [P2] Missing user compiler scripts are silently ignored

**File:** `grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy:391-394`

The combined script reads a captured user script only when the file exists:

```groovy
String configuredScript = userScript?.exists() ? (userScript.text?.trim() ?: null) : null
```

At lines 345-352, the plugin replaces the original `configurationScriptFile` with the generated combined file. Therefore, Gradle no longer validates the originally configured user file. If a build configures a nonexistent script:

```groovy
tasks.named('compileGroovy', GroovyCompile) {
    groovyOptions.configurationScript = file('missing-config.groovy')
}
```

the generator silently emits only the Grails script and compilation proceeds without the requested compiler configuration. Before this change, the execution-time `.text` access failed for a missing file, making a typo or deleted file visible to the user.

Please retain the producer-task support while ensuring that an explicitly configured user script which remains unavailable fails the build instead of silently disabling compiler behavior. Add coverage for a missing configured script.

## Verification

The following focused checks passed:

- `./gradlew :grails-gradle-plugins:test --tests 'org.grails.gradle.plugin.core.GrailsGroovyCompilerConfigSpec' --no-daemon`
- `./gradlew :grails-gradle-plugins:test --no-daemon`
- `./gradlew :grails-gradle-plugins:compileGroovy :grails-gradle-plugins:compileTestGroovy --no-daemon`

The wrapper audit also confirmed the intended Gradle 9.7.1 mainline locations and the intentional Gradle 8.14.5 legacy fixture split.

codeconsole:
@matrei 

Both confirmed and fixed in 818d22b56f, each with a regression case in `GrailsGroovyCompilerConfigSpec` that fails against the previous head.

**Standalone `GroovyCompile`.** The generator is now created on demand: the compile task's `dependsOn` resolves to `groovyCompilerConfigGenerator(project, name)`, which registers the task if no source set already did. That resolution happens while the task graph is built — after configuration, when the task container is mutable again — so it is not a registration from inside a task configuration action, which is what throws. The source-set hook stays, so the usual generators are still listed and a late source set is still covered. The `whenReady` capture now walks `graph.allTasks` instead of every `GroovyCompile` name in the project: only the tasks about to run are taken over, and nothing is realized just to be looked at.

Test: *a GroovyCompile that no source set owns is wired like the source set ones* — a directly registered `compileCustomGroovy` with a `configurationScript` of its own compiles with the combined script, carrying both the Grails imports and its own. The same fixture then runs twice with `--configuration-cache`, so the on-demand generator is shown to be part of the cached graph.

**Missing user script.** The build's script is now an `@InputFile @Optional` of the generator rather than text folded into the `@Input` string, so Gradle's own validation fails the build before the generator runs:

```
A problem was found with the configuration of task ':generateCompileGroovyGrailsCompilerConfig' (type 'GrailsCompilerConfigScriptTask').
Input file does not exist
  ... property 'configurationScript' specifies file '.../missing-config.groovy' which doesn't exist
```

Producer support is unchanged: the ordering edge to the task that writes the script is still declared, and the classpath decoupling assertions still hold (`GENERATOR_DECLARED_INPUT_FILES=0` when no script is configured).

Test: *a configurationScript that does not exist fails the build instead of being dropped*.

matrei (approved):

jdaugherty:
I originally raised a concern on this PR in this [comment](https://github.com/apache/grails-core/pull/16114#issuecomment-5280513306).  From what I can tell, it was never addressed or answered, yet this change was merged.  I'm frustrated this was merged.

jdaugherty:
From what I recall of the original problem: Add a secondary task dependency on compileGroovy, by adding an intermediate task, now that task dependency must also depend on compileGroovy. This has caused issues in our builds enough for us to revert to a `doFirst { }` configure in compileGroovy instead

codeconsole:
@jdaugherty I understand the concern about reintroducing task-ordering failures. I should have followed up with you to establish whether the change covered your actual build wiring before this was merged.

My [earlier explanation](https://github.com/apache/grails-core/pull/16114#issuecomment-5289016409) addressed the runtime-classpath dependency in the old implementation: the merged generator no longer reads either classpath. That addresses that particular dependency, but does not settle every ordering case you described.

I have now reproduced a narrower case in an execution-level test: a build that assigns a plain `configurationScript` file and separately adds its producing task to `compileGroovy.dependsOn` does not order that producer before Grails prepares the combined script. Requesting script preparation first exposes the missing edge and fails because the file does not yet exist. Wiring `configurationScriptFile` to the producing task's output provider makes the same task request compile and run successfully, including after the generated script changes.

I'm preparing a follow-up with that regression test and upgrade documentation for the provider wiring. It keeps the generator independent of unrelated compilation prerequisites; copying all of those dependencies could introduce cycles. Returning to the old `doFirst` assignment would also restore the finalized-property failure on Gradle 9.7.

Could you share the actual task registration and `compileGroovy` wiring from the affected build, including whether the prerequisite writes the compiler script, changes its contents, or only supplies source/classpath entries? That will let us check your specific case rather than assume the reproduced one is the same.

Issue #14005, opened 2025-02-06
Update/optimize dot notation navigation/NavigableMap configuration access


jamesfredley:
### Issue description

Was restored for Grails 7 on https://github.com/grails/grails-core/pull/13869

Address the underlying concerns:

https://github.com/grails/grails-core/issues/10188
https://github.com/grails/grails-core/pull/10280

PR #10280, opened 2016-10-27
Start working on support for replacing config values

jeffscottbrown:
This is not ready to merge and is just a starting point to explore ideas.

Issue #10188, opened 2016-09-29
doWithConfig in unit test merges map instead of replacing


rlovtangen:
### Steps to Reproduce

Add config to `application.groovy`:

```
foobar.foo = "1"
foobar.bar = ["1"]
foobar.baz = ["1" : "1"]
baz = ["1" : "1"]
```

`grails create-service Foo` and add these tests:

```
@TestFor(FooService)
class FooServiceSpec extends Specification {

    static doWithConfig(c) {
        c.foobar.foo = "2"
        c.foobar.bar = ["2"]
        c.foobar.baz = ["2": "2"]
        c.baz = ["2": "2"]
    }

    void "test string"() {
        expect:
            grailsApplication.config.foobar.foo == "2"
    }

    void "test list"() {
        expect:
            grailsApplication.config.foobar.bar == ["2"]
    }

    void "test map one level"() {
        expect:
            grailsApplication.config.baz == ["2": "2"]
    }

    void "test map two levels"() {
        expect:
            grailsApplication.config.foobar.baz == ["2": "2"]
    }

}
```
### Expected Behaviour

All tests should pass
### Actual Behaviour

The test "test map two levels" fails:

```
grailsApplication.config.foobar.baz == ["2": "2"]
|                 |      |      |   |
|                 |      |      |   false
|                 |      |      [1:1, 2:2]
|                 |      [foo:2, bar:[2], baz:[1:1, 2:2]]
```

Note that the test "test map one level" runs fine, which means there are differences based on how deep the property is nested.
### Environment Information
- **Operating System**: OS X El Capitan
- **Grails Version:** 3.2.0
- **JDK Version:** 1.8.0_77
- **Container Version (If Applicable):** 
### Example Application

https://github.com/rlovtangen/bugreport-grails-core-doWithConfig

rlovtangen:
This also results in a merge with existing map instead of replacing:
````
grailsApplication.config.foobar.baz = ["3": "3"]
````

**Comment by jeffbrown** [comment]
Do we know if `3.0.x` and/or `3.1.x` behaved the same way?

**Comment by rlovtangen** [comment]
At least `3.1.11`. Haven't tried `3.0.x`

**Comment by jeffbrown** [comment]
I have added a relevant test at https://github.com/grails/grails-core/commit/7a6ffd96d5400e5ac927a2b0510d461c7dd3c742.

I think the issue is related to how we handle `Map` values in config.

This:

```
cfg.foo.bar = ['one': '1']
[...]

PR #16130, opened 2026-08-11
Memoize system environment resolution in NavigableMapConfig

codeconsole:
## What

`Config.getProperty` resolves every key against the process environment from scratch on each call.

`findInSystemEnvironment` asks `resolvePropertyName` for the environment spelling of the key, and `checkPropertyName` probes `System.getenv` up to four times per casing — as-is, dots replaced with underscores, hyphens replaced with underscores, both replaced — then repeats the whole sequence against the uppercased key:

```java
private String checkPropertyName(String name) {
    if (containsKey(name)) return name;                       // containsKey == System.getenv(name) != null
    String noDotName = name.replace('.', '_');
    if (!name.equals(noDotName) && containsKey(noDotName)) return noDotName;
    String noHyphenName = name.replace('-', '_');
    ...
}
```

That is up to eight `getenv` probes and six intermediate strings per lookup, and the answer cannot change — the process environment is fixed for the lifetime of a config.

This memoizes the resolution per config instance, along with the token list a dotted key splits into:

```groovy
config.getProperty('grails.views.gsp.encoding', String)   // resolves the environment once
config.getProperty('grails.views.gsp.encoding', String)   // subsequent calls reuse it
```

Property **values** are not cached, so configuration mutated at runtime is still observed:

```groovy
config.getProperty('some.nested.value')          // 'original'
config.merge(['some.nested.value': 'updated'])
config.getProperty('some.nested.value')          // 'updated'
```

The cache is per-instance rather than static on purpose. `SystemEnvironmentConfigSpec` installs environment variables reflectively and then builds a fresh config, which must observe the environment as it stands at that point; a static cache would leak a stale answer across those specs.

## Why it matters

Callers resolve configuration inside render loops. On a scaffolded page rendering 100 rows, `FormFieldsTemplateService` (`getShouldCache`, `findTemplate`, `getTemplateFor`) and asset-pipeline resolve configuration per rendered property, so this sits on a hot path that scales with rows × properties.

## Measurements

| | before | after |
|---|---|---|
| `getProperty`, 2M lookups over a 4-key set | 355.3 ns/op | 91.6 ns/op |
| `NavigableMapConfig` share of JFR execution samples, 100-row scaffolded page under load | 7.01% | 1.16% |

## Limitations

- **End-to-end request latency improvement was not demonstrated.** The CPU work is measurably removed (both figures above), but on the machine used for this the run-to-run variance across JVM restarts was larger than the effect, so wall-clock A/B could not resolve it. The dominant remaining costs on that page are `ExpandoMetaClass` read-lock contention (~9–10% of samples) and reflective tag dispatch, neither of which this touches.
- The caches are bounded at 2048 entries each. Configuration keys come from a small fixed set of source literals; the bound only guards against a caller synthesising unbounded key strings at runtime.

## On the `@Deprecated` marker

`NavigableMapConfig` carries a class-level `@Deprecated` since #11554 (May 2020), pointing at `grails.config.Config`. That target is the *interface* this class implements, and no alternative implementation exists in the tree, so the note reads as "code against the interface" rather than "a replacement is available".

The class is on the live path today:

- `PropertySourcesConfig extends NavigableMapConfig` and is not itself deprecated.
- `AbstractGrailsApplication` builds a `PropertySourcesConfig`, so this is `grailsApplication.config` for every running application.
- Profiling a running 8.0.0-M5 application put `NavigableMapConfig` at 7.01% of execution samples, entered through `getProperty`.

If the `NavigableMap`-backed implementation is replaced, this change goes with it. It is offered as a cost reduction on the path applications actually execute for the six years the marker has been in place, and is easy to drop if a replacement is in flight.

### DIFF
--- a/grails-core/src/test/groovy/org/grails/config/SystemEnvironmentConfigSpec.groovy
+++ b/grails-core/src/test/groovy/org/grails/config/SystemEnvironmentConfigSpec.groovy
@@ -73,6 +73,56 @@ property-with_mixed.symbols: from-yml
         'property-with_mixed.symbols' | 'PROPERTY_WITH_MIXED_SYMBOLS' | 'from-env'
     }
 
+    void 'a property read more than once keeps resolving to the system environment value'() {
+        given: 'configuration that is overridden by the environment'
+        def config = configFor('property.with.period: from-yml')
+        modifiableSystemEnvironment.put('PROPERTY_WITH_PERIOD', 'from-env')
+
+        expect: 'every read resolves to the environment value, not just the first'
+        config.getProperty('property.with.period') == 'from-env'
+        config.getProperty('property.with.period') == 'from-env'
+        config.getProperty('property.with.period') == 'from-env'
+
+        cleanup:
+        modifiableSystemEnvironment.remove('PROPERTY_WITH_PERIOD')
+    }
+
+    void 'a property that has already been read still reflects later configuration changes'() {
+        given: 'a property that has been read once'
+        def config = configFor('some.nested.value: original')
+        assert config.getProperty('some.nested.value') == 'original'
+
+        when: 'the configuration is changed'
+        config.merge(['some.nested.value': 'updated'])
+
+        then: 'the new value is returned rather than the previously resolved one'
+        config.getProperty('some.nested.value') == 'updated'
+    }
+
+    void 'a config created after an environment variable is installed observes it'() {
+        given: 'a config created and read before the variable exists'
+        def before = configFor('late.bound.property: from-yml')
+        assert before.getProperty('late.bound.property') == 'from-yml'
+
+        when: 'the variable is installed and a new config is created'
+        modifiableSystemEnvironment.put('LATE_BOUND_PROPERTY', 'from-env')
+        def after = configFor('late.bound.property: from-yml')
+
+        then: 'the new config resolves to the environment value'
+        after.getProperty('late.bound.property') == 'from-env'
+
+        cleanup:
+        modifiableSystemEnvironment.remove('LATE_BOUND_PROPERTY')
+    }
+
+    private static PropertySourcesConfig configFor(String yaml) {
+        def yamlPropertiesSource = new YamlPropertySourceLoader()
+                .load('application.yml', new ByteArrayResource(yaml.bytes, 'test.yml'), null)
+        def propertySources = new MutablePropertySources()
+        propertySources.addFirst(yamlPropertiesSource.first())
+        new PropertySourcesConfig(propertySources)
+    }
+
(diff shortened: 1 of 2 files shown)


grails-core/src/test/groovy/org/grails/config/SystemEnvironmentConfigSpec.groovy
```diff
@@ -73,6 +73,56 @@ property-with_mixed.symbols: from-yml
...
+        config.getProperty('property.with.period') == 'from-env'
+
+        cleanup:
+        modifiableSystemEnvironment.remove('PROPERTY_WITH_PERIOD')
+    }
+
+    void 'a property that has already been read still reflects later configuration changes'() {
+        given: 'a property that has been read once'
+        def config = configFor('some.nested.value: original')
+        assert config.getProperty('some.nested.value') == 'original'
```
sbglasius, line 93:
```suggestion
        expect:
        config.getProperty('some.nested.value') == 'original'
```

grails-core/src/test/groovy/org/grails/config/SystemEnvironmentConfigSpec.groovy
```diff
@@ -73,6 +73,56 @@ property-with_mixed.symbols: from-yml
...
+        config.merge(['some.nested.value': 'updated'])
+
+        then: 'the new value is returned rather than the previously resolved one'
+        config.getProperty('some.nested.value') == 'updated'
+    }
+
+    void 'a config created after an environment variable is installed observes it'() {
+        given: 'a config created and read before the variable exists'
+        def before = configFor('late.bound.property: from-yml')
+        assert before.getProperty('late.bound.property') == 'from-yml'
```
sbglasius, line 105:
```suggestion
        expect:
        before.getProperty('late.bound.property') == 'from-yml'
```

sbglasius (approved):

codeconsole:
@jdaugherty suggests doing merge at the start and remove deprecated.

jdaugherty:
@codeconsole I think we need to look at the histoyro f the changes on the navigablemap. from our call yesterday: 

2016 – #10188: buggy/awkward nested-map replacement semantics are identified; Jeff investigates and adds a failing test.
2016 – #10280: Jeff starts an experimental implementation to fix the semantics. It never gets merged.
2020 – #11554: NavigableMap is deprecated, making Jeff's unfinished approach ostensibly unnecessary.
2024 – #13869: Grails reverses course enough to restore dot-notation navigation for Grails 7.
2025 – #14005: the project explicitly resurrects #10188/#10280 as the underlying problems that should now actually be fixed.

I don't think memoized is the solution here b/c it's a cache and with large configuration you're going to be using a lot of memory.  Can't we look at what Jeff did originally and see if we can fix it that way?

jdaugherty:
I've moved this to 8.1 for now.  It will likely land in 9 though.

PR #16135, opened 2026-08-11
Re-enable grails-datamapping-rx module

borinquenkid:
## Summary
- Restores the grails-datamapping-rx module to the active build (was disabled in both settings.gradle and gradle/publish-root-config.gradle pending a port to the current Grails release)
- Fixes 7 real bugs uncovered while re-enabling it:
  - RxSchedule.java: an @interface illegally declared a `default` annotation member with a method body instead of a constant, and referenced a removed GroovyTransformOrder constant
  - grails.gorm.rx.DetachedCriteria.clone() illegally narrowed visibility from its public parent
  - RxEntity.groovy was missing `import rx.Observable`
  - RxScheduleIOTransformation.groovy referenced the same stale GroovyTransformOrder constant
  - FindByFinder/FindAllByFinder (rx) called a 3-arg buildQuery(...) overload that no longer exists after the finders' Groovy->Java conversion; inlined the steps the current 2-arg buildQuery performs internally
- Brings the module's Checkstyle/CodeNarc compliance up to date with the rest of the codebase (it had never been linted while disabled)

## Test plan
- [x] `:grails-datamapping-rx:compileGroovy` — BUILD SUCCESSFUL
- [x] `:grails-datamapping-rx:test` — 8/8 passing
- [x] `:grails-datamapping-rx:codeStyle` — Checkstyle + CodeNarc both clean


jdaugherty (commented):
1. the gradle project 'grails-data-docs/guide-rx' was the documentation for this module and it should be restored as well.

2. since you're already touching all of these files, can you please remove the author tags? 

3. there are whole tests commented out in grails-datamapping-core-test around this functionality & the associated module include.  The reason we had commented this out is there were some test failures we couldn't figure out.  Are the tests you wrote completely new?

borinquenkid:
Thanks for the review, @jdaugherty — addressed all three points:

1. **`grails-data-docs/guide-rx`** — re-included in `settings.gradle` (also fixed a stale `projectDir` path left over from a prior restructuring). `:grails-data-docs-guide-rx:asciidoctor` builds clean. (1d8f3e7)

2. **`@author` tags** — removed from all 56 touched files under `grails-datamapping-rx/src/main`. (1d8f3e7)

3. **Are the tests completely new? Yes** — and since you mentioned the commented-out tests were disabled over unresolved failures, I went back and root-caused them instead of leaving them commented:
   - Restored the `grails-datamapping-rx` test dependency and un-commented `RxServiceImplSpec` in `grails-datamapping-core-test`.
   - Root-caused the underlying `GroovyCastException` bug: `SingleResultAdapter`/`ObservableResultAdapter` never actually narrowed the generated method's declared return type before delegating to the wrapped implementer, so the automatic `@Transactional` decoration built its wrapping cast against the Rx-wrapped type (e.g. `Single<Number>`) instead of the type the synchronous body actually returns — an always-throwing cast. (9eb7887)
   - Found and fixed a second, related bug: the RX-delegating closure let Groovy's static compiler infer its return type from context, injecting an invalid narrowing cast whenever the delegate returned an `Iterable` to flatten (the multi-result/`Observable` path). (10516d794)
   - Along the way, fixed two of the test's own bugs from 2017 (a `.first()`/`.value()` mixup on the wrong `Single`/`Observable` overload, and a stale "cannot implement" comment blocking a property projection that compiles fine today). (770448e4, 9a276e66c)
   - All 7 previously-commented/pending cases in `RxServiceImplSpec` now pass.

Also caught and fixed an earlier Copilot review pass (2026-08-11) that had gone unaddressed — two check-then-act races on `ConcurrentHashMap` (`QueryState.addLoadedEntity`, `RxGormEnhancer.registerEntity`) fixed with `computeIfAbsent`/`putIfAbsent` plus regression tests, and a message typo in `MultiTenantEventListener`. (506228bd)

Full history: 2632f6fc6c..506228bd92

jdaugherty (approved):

PR #16134, opened 2026-08-11
Resolve tag calls at compile time instead of through the metaclass

codeconsole:
## What

Tag libraries are described as they are compiled, and that description resolves tag calls in code compiled afterwards. A call whose namespace and tag are known becomes a direct invocation instead of being dispatched through the metaclass, and nothing is installed onto a metaclass to make dispatch work.

### Defining tags

```groovy
class GreetingTagLib {
    static namespace = 'greet'

    def hello(Map attrs) {
        out << "Hello ${attrs.name}"
    }

    def wrapped(Map attrs, Closure body) {
        out << '<div>' << body() << '</div>'
    }
}
```

### Calling tags

In a tag library or a controller:

```groovy
class BookController {
    def index() {
        String markup = g.createLink(controller: 'book')   // compiled into a direct invocation
        String other  = greet.hello(name: 'Grails')        // likewise
        String third  = createLink(controller: 'book')     // likewise, when nothing else answers to the name
    }
}
```

A call that names its namespace is compiled the same way inside a closure — a tag body, a `withFormat` block, anything taking a block — as outside one, as is one in a constructor or a field initialiser. A call written *without* a namespace inside a closure is not: a closure is handed a delegate when it runs, and a name the delegate answers to is the delegate's rather than a tag's. `request.withFormat { form multipartForm { } }` is the case that settles it — `form` there is a format in a DSL, not the `g:form` tag.

The tag is still selected by name when the call runs, through the same lookup dynamic dispatch uses. A tag library that overrides another, one registered while the application is running, and the order tag libraries are registered in all decide the outcome exactly as before. Nothing is bound to a particular tag library class, so a tag declared by more than one of them, and a tag declared as a `Closure` field, are compiled the same way.

A call written **without** a namespace is not compiled unless the build asks for it with `grails { compileStatic { unqualifiedTagCalls = true } }`. Whether a bare name is a tag depends on what else answers to it, and not all of that is visible when compiling — a method Groovy gives every object, a delegate an enclosing closure is handed, an overload the tag library also declares — so by default such a call is dispatched exactly as before.

The attributes and body are passed straight through where the call says what they are; where it does not — a map held in a variable, a single value the tag reads under its own name — the arguments are forwarded as written and adapted by the same rules dynamic dispatch applies.

### Tags in pages

A page resolves a name against the model it was rendered with before it reaches a tag library, and that model is not known when the page compiles. A page therefore keeps resolving its tags as it always has, unless it declares `compileStatic`:

```html
[...]

PR #16139, opened 2026-08-12
Compile GSP pages statically across an application

codeconsole:
Makes GSP static compilation something an existing application can turn on. Measured on an application with 73 `<g:set>` across its views and a page reading Grails internals reflectively: **415 compilation errors to 0**, with two of its own pages edited.

### Turning it on

```yaml
grails:
    views:
        gsp:
            compileStatic: true
```

The same thing can now be asked for from the build, beside the artefact opt-ins:

```groovy
grails {
    compileStatic {
        gsp = true
    }
}
```

The build option states the same `grails.views.gsp.compileStatic` setting as a system property, so it reaches both places a page is compiled: the forked page compiler, and the JVM running the application, which compiles a page again when it changes. A page compiled one way by the build and another way while being developed would be worse than not offering the option. A page still decides for itself with `<%@ page compileStatic="false" %>`.

`gsp` is not part of `compileStatic { all = true }`. The artefact opt-ins fail on code that is doubtful anyway; this one fails on a page reading a model variable it has not declared, which describes most pages in an application that has never declared one.

### What a page has to declare

A page that declares a model states what it is rendered with, so a name outside it is a mistake and is reported:

```html
<%@ page model="Book book" %>
<h1>${book.title}</h1>      <%-- ${publisher} here is an error: it is not in the model --%>
```

A page that declares nothing has stated nothing, and reads what it is rendered with the way a page that is not compiled statically reads it.

This is the decision the rest follows from. Compiling statically is worth ~1.5x whether or not a page declares anything (below), so requiring every model to be declared first would cost most applications the benefit to gain checking they had not asked for. Declaring buys checking, compiling statically buys speed, and they can be adopted separately. To hold every page to what it declares:

```groovy
grails {
    compileStatic {
        gsp = true
        strictGsp = true
    }
}
```

### Names the framework binds

These no longer need declaring, and most carry their real types, so they are checked:

| Name | Type |
| --- | --- |
| `params` | `TypeConvertingMap`, so `params.id` and `params.int('max')` both resolve |
| `flash` | `FlashScope` |
| `request`, `response`, `session` | `HttpServletRequest`, `HttpServletResponse`, `HttpSession` |
| `application`, `servletContext` | `ServletContext` — the same object under both names |
| `webRequest` | `GrailsWebRequest` |
| `controllerName`, `actionName`, `namespace` | `String` |

Written into the page rather than onto the class it extends: a page is compiled with the application's classpath and that class is not, so `FlashScope` and `GrailsWebRequest`, which live in a module depending on the one holding that class, can be named there and not there. It also leaves a page free to declare one of these names in its own model and get its own type.

`servletContext` is newly bound; only `application` was, which is the same object.

Being typed, they are checked: `${request.contextPathTypo}` is an error, and the converters from #15715 resolve, since `session.string('x')` only compiles where the session carries its type.

`grailsApplication` and `applicationContext` are the only two deliberately **not** typed. What pages read from them is not declared anywhere: `grailsApplication.controllerClasses` is matched at runtime against `(\w+)(Classes)` and answered from the artefact handlers, an open set no interface can enumerate. Typing them only changes which type the failure names, so they are read dynamically, which lets them work.

### Names a page introduces

The `var` and `status` attributes of the tags a page calls are names the page has declared as plainly as it can, and no longer need declaring again:

```html
<%@ page compileStatic="true" %>
<g:set var="total" value="${books.size()}"/>
<g:each in="${books}" var="book" status="i">${i}. ${book.title}</g:each>
Total: ${total}
```

What such a name holds is decided by the tag at render time, so it is read dynamically. To apply an operator to it the page has to say what it is, with either tag:

```html
<g:def type="int" var="total" value="${books.size()}"/>   <%-- a local --%>
<g:set type="int" var="total" value="${books.size()}"/>   <%-- a local, and written to the scope --%>
${total + 1}
```

`g:set` gained the `type`: the declaration is written first and the tag is then called with the declared variable as its value, so the scope write still happens and `scope` still decides where it goes. Untyped `g:set` is unchanged. A `type` is accepted only alongside `value` — with a body or `bean` the value is produced when the tag runs, so there is nothing to declare from, and that is rejected rather than ignored.

`g:def` emitted `int.cast(...)` for a primitive, which throws whatever it is handed, so a page using one compiled and then failed to render. Both tags cast through the wrapper now.

Each tag is used where it fits. The welcome page's counts and collections are read only by the page that declares them, so it uses `g:def`, which is a local and nothing more. `g:set type=` is for a value that does need to be in a scope, which previously could not be typed at all.

### Holding an artefact type back from `all`

`compileStatic { all = true; services = false }` enabled services anyway: the shortcut was folded in as `all || services`, and every flag carried a convention of `false`, so a type set to `false` could not be told apart from one never set. An artefact type now falls back to `all` rather than being OR-ed with it.

### What it is worth

One page body, compiled three ways, 200 rows, 800 renders, median of 7:

| | |
| --- | --- |
| dynamic | 1.00x |
| static, nothing declared | 1.55x – 1.70x |
| static, model declared | 1.38x – 1.62x |

The gain comes from compiling the page statically at all. **Declaring a model buys type checking, not speed** — the two static rows are indistinguishable here, and their ordering flips between runs. Round-to-round spread is roughly 2x, so these are directional: in-process renders, no HTTP, no JMH, one machine.

### Limitations

- **An operator applied to a value of no known type is reported, and has to be.** `${rows[0]}` where nothing said what `rows` holds is an error rather than being read dynamically: a subscript is written into the class as `getAt`, and the writer that does it fails with a `GroovyBugError` rather than an error when handed a receiver whose type was never established. The report says what the page can do — declare it, or give it a type with `<g:def type="...">`. It covers a closure parameter too, which nothing declares and nothing resolves.
- **Templates.** A `_foo.gsp` rendered via `<g:render template="foo" model="[...]"/>` takes its model from the caller, so it has to declare every attribute it may be passed, and nothing checks that the caller and the template agree.

### Relationship to #16134

#16134 is merged, and this branch carries it. Its change to `GroovyPageTypeCheckingExtension` seeds the allowed namespaces from the compiled tag library index, so `grails.views.gsp.compileStaticConfig.taglibs` and the `taglibs` page directive are no longer needed for a tag library on the compile classpath. The guide says so, and says what they are still for: a namespace filled in while the application runs, which nothing describes when the page is compiled.

Three files were touched by both. `GroovyPageTypeCheckingExtension` gains the operators reported here beside the index read there, and the namespaces each seeds are added to the same set. `GrailsCompileStaticOptions` keeps the artefact properties without a convention, which is what lets an explicit value be held back from `all`, alongside that PR's three tag settings. `GroovyPagePlugin` keeps both sets of methods.


grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
```diff
@@ -568,30 +568,30 @@
                                  [code: 'welcome.binding.formatted', beans: applicationContext.getBeansOfType(grails.databinding.converters.FormattedValueConverter)],
                                  [code: 'welcome.binding.structured', beans: applicationContext.getBeansOfType(grails.databinding.TypedStructuredBindingEditor)],
                                  [code: 'welcome.binding.listeners', beans: applicationContext.getBeansOfType(grails.databinding.events.DataBindingListener)]]}"/>
-                <g:set var="numBindingBeans" value="${bindingGroups.sum { g -> g.beans.size() } ?: 0}"/>
-                <g:set var="mimeTypeProviders"
+                <g:def type="int" var="numBindingBeans" value="${(int) (bindingGroups.sum { g -> g.beans.size() } ?: 0)}"/>
```
jdaugherty: Can't we fix the set tag to handle types?  We can always set it to Object and it should still work, no?  Why the def?

jdaugherty: @codeconsole I think we need to work on the syntax here before proceeding.  The set tag could take a type so the compiler knows the type or the def tag could be improved to autocast.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
```diff
@@ -0,0 +1,216 @@
...
+----
+<%@ page model="Book book; List<Review> reviews" %>
+----
+
+Because a declared model is what makes a page type checkable, the `model` directive enables static compilation on its own — `compileStatic="true"` is implied and does not need to be given as well.
+
+Declaring a model also states that the model is complete, so reading a name outside it is reported rather than left to the render:
+
+----
+The variable [autor] is undeclared.
```
matrei: Why is `[autor]` a typo. It should be reported anyway if spelled correctly as it was not declared as part of the model, right? So using a typo in this example is redundant?

codeconsole: Any name outside the model is reported whether it's spelled correctly or not. Changed the example to `[publisher]`, which is simply not in the model.

grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
codeconsole: `g:set` takes a `type` now (`a794e2a`): the declaration is written first and the tag is still called, so the scope write happens, `scope`/`bean` are unaffected, and untyped `g:set` is unchanged. `type` is only valid alongside `value` — with a body or `bean` there's no expression to declare from.

`g:def` needed the autocast too: it emitted `int.cast(...)` for a primitive, which throws whatever it's handed, so a page compiled and then failed to render. Casts through the wrapper now (`43dcdc4`).

The welcome page stays on `g:def` — those twelve values are read only by the page that declares them, so the scope write is unused. Say the word if you'd rather it demonstrate `g:set type=` instead.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
```diff
@@ -200,6 +236,28 @@ public class GroovyPageParser implements Tokens {
...
+     * Matches the {@code var} and {@code status} attributes of a namespaced tag, which is how a page
+     * names something it introduces: {@code <g:set var="total"/>}, {@code <g:each var="book"
+     * status="i">}, {@code <g:eachError var="error">}.
+     *
+     * <p>Read from the page source rather than from the parsed attributes because attributes are
+     * parsed only on the pass that writes the class, by which point the annotation carrying these
+     * names has already been written. Matching a name that turns out not to be a page scope variable
+     * costs only that the name resolves dynamically, so the pattern errs towards matching.</p>
+     */
+    private static final Pattern PAGE_SCOPE_VARIABLE_PATTERN = Pattern.compile(
```
sbglasius, line 271: `PAGE_SCOPE_VARIABLE_PATTERN`'s `[^>]*?` stops at the first `>`, so `<g:each in="${l.findAll { it.x > 1 }}" var="book">` never registers `book`; under a declared model (strict implied) or `strictGsp` that's a spurious "variable [book] is undeclared".

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
```diff
@@ -89,17 +142,65 @@ class GroovyPageTypeCheckingExtension extends GroovyTypeCheckingExtensionSupport
...
+        if (currentScope.undeclaredDynamicVariables.add("operator:$described")) {
+            typeCheckingVisitor.addStaticTypeError(
+                    "The type of ${described} is not known here, and [${expression.operation.text}] cannot be " +
+                            'applied to it. Declare it in the model directive, or give it a type where it is ' +
+                            'introduced.', expression.leftExpression)
+        }
+    }
+
+    private static String describe(Expression expression) {
+        expression instanceof VariableExpression ? "[${((VariableExpression) expression).name}]" : 'this value'
```
sbglasius, line 247: `describe()` returns the constant `'this value'` for non-variable expressions and the dedup key is built from it, so only the first operator-on-unknown-type error in a page is ever reported.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
```diff
@@ -1235,6 +1396,64 @@ private void flushBufferedWhiteSpace() {
...
+                    pageName, getCurrentOutputLineNumber());
+        }
+        Object value = attrs.get("\"value\"");
+        if (value == null) {
+            throw new GrailsTagException("Tag [set] can only be given a [type] together with a [value]; " +
+                    "the body and the [bean] attribute are produced when the tag runs, so there is nothing " +
+                    "to declare the variable from", pageName, getCurrentOutputLineNumber());
+        }
+        attrs.remove("\"" + TYPE_ATTRIBUTE + "\"");
+        out.println(type + " " + var + " = " + castingTypeFor(type) + ".cast(" + getExpressionText(value.toString()) + ")");
```
sbglasius: I don't know if this is a feature or a bug, but typed `g:set` emits a real local declaration, so two typed sets of the same var in one block scope fail with "scope already contains a variable"; the untyped form of `g:set` rebinds happily.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
```diff
@@ -39,40 +41,91 @@ import org.codehaus.groovy.transform.stc.StaticTypesMarker
...
 class GroovyPageTypeCheckingExtension extends GroovyTypeCheckingExtensionSupport.TypeCheckingDSL {
 
+    /**
+     * Names the framework binds into every page whose members are answered at runtime rather than
+     * declared: {@code grailsApplication.controllerClasses} is matched against a name and answered
+     * from the artefact handlers, an open set no interface enumerates, and what pages read from the
+     * application context belongs to an implementation rather than to the interface. Every other name
+     * the framework binds is declared on the page with the type it holds, and is checked normally.
+     */
+    /**
```
sbglasius: What Javadoc block should stay and which should go?

sbglasius (dismissed):
Reviewed with the help of Claude

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
codeconsole: Confirmed and fixed. Quoted attribute values are consumed whole now, so a `>` inside one is part of the value. (Your `g:each` example survives either way — it is a syntax tag and emits a real loop variable — but a non-syntax tag with `var` after the `>` did fail; that case is now a test.)

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
codeconsole: Confirmed and fixed — two property receivers in one page reported one error. Non-variable receivers are keyed by source position now; plain names still dedupe by name, so one error per variable rather than one per use.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
codeconsole: Bug, and fixed. `g:set` writes a scope and writing it twice is ordinary, so the first occurrence declares and a later one assigns — matching the untyped form.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
codeconsole: Both stay — a merge left the `FRAMEWORK_DYNAMIC_NAMES` block above the wrong field. Each is now on the field it describes.

sbglasius:
Re-checked `1f63d1a784` by compiling probe pages against `86c4f060ef` and the fix commit rather than reading the diff.

**Confirmed fixed:** the `var`/`status` regex (`<g:set value="${1 > 0}" var="flag"/>` failed with *"The variable [flag] is undeclared."* before, compiles now — and the quote-aware alternation is disjoint on its first character, so no backtracking risk); the operator-error dedup; the misplaced javadoc.

One correction to my earlier list: my "same expression reported twice" point was wrong. `<g:set var="rows" value="${[1,2]}"/>${rows[0]}` emits exactly one error both before and after. Disregard it.

**New regression from the typed-`g:set` fix.** `typedSetVariables` is a flat per-page set with no notion of the generated block scope, so a second typed `g:set` in a sibling scope emits a bare assignment where the variable was never declared — it resolves through the page's `setProperty` and the type is silently lost:

```gsp
<g:if test="${true}"><g:set type="int" var="n" value="${1}"/>${n + 1}</g:if>
<g:else><g:set type="int" var="n" value="${2}"/>${n + 1}</g:else>
```

Compiles at `86c4f060ef`; at `1f63d1a784` the else branch fails with *"The type of [n] is not known here, and an operator cannot be applied to it."* Declaring inside a `<g:if>` and typing again at top level fails the same way. Keying by emission depth (`tagMetaStack` size / `closureLevel`) and only reusing the assignment form when the earlier declaration encloses the current position would handle both and keep the same-scope case the new test covers.

**On framework names as model keys** Two things I checked that shape the note:

A model key does shadow the implicit variable today: `TemplateVariableBinding.getVariable` checks the page's own binding (where `GroovyPageWritable.createBinding` puts the model) before the parent `WebRequestTemplateVariableBinding`. With a real web request, `application`, `session`, `namespace`, `servletContext`, `params`, `flash` and `response` all returned the model value.

But the exposure is narrower: on the base, `<%@ page compileStatic="true" %>${application}` already fails with *"The variable [application] is undeclared."* for all seven, model key or not. So this is a pure improvement for pages already on `compileStatic="true"`; only dynamic pages newly opted in via `grails.views.gsp.compileStatic` are affected.

Worth calling out in the note: `${application.name}` is a compile error the build catches, but bare `${application}` compiles clean and throws `GroovyCastException` at render — `compileGroovyPages` stays green.

Also: the guide's GSP list is `application, applicationContext, flash, grailsApplication, out, params, request, response, session, webRequest`. `controllerName`, `actionName`, `namespace` and `servletContext` are documented only for tag libraries. Being `String`, the first three only fail on a non-String model value.

Suggested addition to the existing NOTE under "Enabling Static Compilation for Every Page":

> A model variable sharing a name with one the framework supplies (`params`, `request`, `response`, `session`, `application`, `servletContext`, `flash`, `webRequest`, `controllerName`, `actionName`, `namespace`) is no longer read as the model supplied it — the page reads the framework value's type. Reading a member the framework type does not have is a compilation error; passing a value of another type fails at render with a `GroovyCastException`, so it will not be caught by precompilation. Rename the model key, or declare it in the `model` directive to keep your own type.

**Minor, still open:** the `compileOnly` comment in `grails-gsp/core/build.gradle` sits above an empty block (only `testImplementation` entries were added); `GrailsGspCompileStaticProvider.isStrict()` is `@Input` but is ignored by `asArguments()` when `gsp` is off, so toggling `strictGsp` invalidates tasks for no change in output.

jdaugherty:
@codeconsole can we add a functioanl test app that actually uses the static compilation?

codeconsole:
@jdaugherty 

Added — `grails-test-examples/gsp-compile-static` (32d19da). A build that sets `grails.compileStatic.gsp`, pages reading a declared model, the framework-bound names, a typed local and a tag library in its own namespace, rendered through a running server.

Rendering is not on its own evidence, since a page renders the same either way, so the test also checks the class each page compiled to. Worth separating: a `<%@ page model %>` directive implies static compilation on its own, so those pages stay static with the flag off — only the page that declares nothing answers whether the build setting did anything. Verified by flipping the flag: that page is the one that flips.

codeconsole:
@sbglasius 

All four addressed in 7a5df8a.

**The regression is real and is fixed.** Your repro reproduced exactly, and so did the top-level variant. `typedSetVariables` is now one frame per emitted block, pushed when a tag opens a body and popped when it closes: a name an enclosing frame declared is assigned to, a name only a sibling declared is declared again. Both your cases plus a loop-body sibling and the enclosing-then-inner case are pinned in `GspCompileStaticConfigSpec`.

**The note** is in, close to your wording — all eleven names match what the parser writes.

**Both minors** fixed: the stale comment is gone, and `isStrict()` now reports `isCompileStaticGsp() && strictGsp`, so toggling strictness while `gsp` is off no longer invalidates the task. Pinned in `GrailsGspCompileStaticProviderSpec`.

On `${application}` compiling clean and throwing at render — agreed, and that is why the note says precompiling will not catch it.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
```diff
@@ -1235,6 +1430,67 @@ private void flushBufferedWhiteSpace() {
...
+        if (value == null) {
+            throw new GrailsTagException("Tag [set] can only be given a [type] together with a [value]; " +
+                    "the body and the [bean] attribute are produced when the tag runs, so there is nothing " +
+                    "to declare the variable from", pageName, getCurrentOutputLineNumber());
+        }
+        attrs.remove("\"" + TYPE_ATTRIBUTE + "\"");
+        // A name typed once is declared; typing it again assigns to what was declared. Declaring it
+        // twice would not compile, where the untyped tag simply writes the scope again.
+        String declaration = declareTypedSetVariable(var) ? type + " " + var : var;
+        out.println(declaration + " = " + castingTypeFor(type) + ".cast(" + getExpressionText(value.toString()) + ")");
```
jdaugherty: `Class.cast` does not coerce, so this is the same defect the `int.cast(...)` fix addressed, just narrowed rather than removed. Swapping the primitive for its wrapper only fixes exact-type boxing; every other conversion Groovy would have performed still fails, and it fails at render after compiling clean.

Measured on this branch, all of these compile with no error and then throw `ClassCastException` when the page renders:

| tag | outcome |
| --- | --- |
| `<g:set type="long" var="n" value="${2}"/>` | `Long.cast(Integer)` → CCE at render |
| `<g:set type="double" var="d" value="${2}"/>` | `Double.cast(Integer)` → CCE at render |
| `<g:set type="String" var="s" value="Total: ${n}"/>` | `String.cast(GString)` → CCE at render |
| `<g:set type="java.util.List" var="l" value="${s.split(',')}"/>` | `List.cast(String[])` → CCE at render |

A declared `long` or `double` initialised from an integer literal is about as ordinary as this tag gets, and `type="String"` over a GString is the most natural thing a page author would write. Worth noting that the passing cases in `GspCompileStaticConfigSpec` are exactly the shapes that survive `Class.cast` — `type="long"` is tested as `${2L}` and `type="double"` as `${1.5d}`, so the suffixes are load-bearing and the test reads as agreeing with the bug rather than pinning the behaviour.

Emitting a Groovy cast fixes it, and is what `writeFrameworkSuppliedAccessors()` in this same PR already does for the framework-supplied names:

```java
out.println(declaration + " = (" + type + ") " + getExpressionText(value.toString()));
```

I ran this: `long`/`${2}` renders `2`, `double`/`${2}` renders `2.0`, and the `String[]`-into-`List` case becomes an honest compile-time `Inconvertible types: cannot cast java.lang.String[] to java.util.List` instead of a clean compile and a render-time crash — which is the better outcome, since the whole point of the option is to move failures to compile time. All 71 existing `GspCompileStaticConfigSpec` tests still pass with the change. `PRIMITIVE_WRAPPERS` and `castingTypeFor` then become unnecessary here.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/tags/GroovyDefTag.java
```diff
@@ -59,10 +71,18 @@ public void doStartTag() {
         if (typeName.equals("def") || typeName.equals("Object")) {
             out.println(expr);
         } else {
-            out.println(typeName + ".cast(" + expr + ")");
+            // Cast through the wrapper for a primitive: Class.cast on a primitive class throws
+            // whatever it is handed, so int.cast(1) fails where Integer.cast(1) is what was meant.
+            // The declared type stays primitive, and the result unboxes into it.
+            out.println(castingTypeFor(typeName) + ".cast(" + expr + ")");
```
jdaugherty: Same `Class.cast` issue as the typed `g:set` — `(" + typeName + ") (" + expr + ")"` is the coercing form, and the primitive-wrapper map goes away with it.

Separately, and pre-existing rather than introduced here: `expr` is not parenthesised on this path, so a GString value emits malformed code. `<g:def type="String" var="s" value="Total: ${1 + 1}"/>` produces `String s=(String) Total: ${1 + 1}` and fails with `Cannot find matching method java.lang.Class#call(...)`. Since this PR is putting `g:def type=` in front of users as the recommended way to give a page-local value a type — and the welcome page now leans on it — it seems worth fixing in the same pass.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
```diff
@@ -0,0 +1,218 @@
...
+An individual page always wins over either, and can opt back out:
+
+[,xml]
+----
+<%@ page compileStatic="false" %>
+----
+
+[NOTE]
+====
+Enabling this for an existing application is not a transparent change: every page that references an undeclared model variable will fail to compile until it declares its model. Introduce the setting once the views it covers declare their models, and use `compileStatic="false"` for the pages that are not ready yet.
```
jdaugherty: This reads as a leftover from before the non-strict default landed, and it now contradicts the page itself. Line 97 says a page that declares no model "reads what it is rendered with exactly as a dynamically compiled page does", and `GroovyPageTypeCheckingExtension` only reports undeclared names when `strict` is set — so referencing an undeclared model variable does *not* fail to compile by default, and the advice to "introduce the setting once the views it covers declare their models" undersells how adoptable the feature actually is.

The two things that genuinely are not transparent are worth naming in its place, since they are what an existing application will actually hit:

- an operator applied to a value of no known type (`${a + b}`, `${rows[0]}`), which is reported regardless of `strict` — this is the one that will account for most of the work, and `+` on an undeclared name is common in real views;
- a member that does not exist on a framework-supplied name, now that those carry their real types — which the second paragraph of this note already covers well.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
```diff
@@ -0,0 +1,218 @@
...
+        gsp:
+            compileStaticConfig:
+                taglibs:
+                    - myns
+                    - othertags
+----
+
+The declared namespaces are added to the defaults rather than replacing them.
+
+TIP: Static compilation of GSP pages is independent of the `grails { compileStatic { } }` build options described in link:staticTypeCheckingAndCompilation.html#grailsCompileStatic[GrailsCompileStatic], which apply to controllers, services and tag library classes.
```
jdaugherty: This TIP is no longer true, and it contradicts the "Enabling Static Compilation for Every Page" section 40 lines above it, which documents `grails { compileStatic { gsp = true } }`. GSP static compilation is now *reached through* the `grails { compileStatic { } }` block; what remains true is the narrower point that `all = true` does not include `gsp`, which the NOTE at line 211 already makes.

Suggest either dropping it or rewriting it to say that the artefact opt-ins in that block apply to controllers, services and tag library classes, while `gsp` applies to pages and is not covered by `all`.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
```diff
@@ -0,0 +1,218 @@
...
+Rather than adding the directive to each page, static compilation can be made the default for an entire application with the `grails.views.gsp.compileStatic` setting:
+
+[source,yaml]
+.grails-app/conf/application.yml
+----
+grails:
+    views:
+        gsp:
+            compileStatic: true
+----
```
jdaugherty, line 175: The `type` attribute is missing from the guide entirely. The section explains that a name introduced by a tag "is read dynamically rather than reported", but never mentions that `type` is how a page opts out of that — even though the compiler error users will hit tells them to do exactly that: *"Declare it in the model directive, or give it a type where it is introduced."* As written, someone who reads that message and comes here finds nothing about giving a name a type.

The PR description covers this well (the `g:def` vs `g:set type=` distinction, and that `type` is only valid alongside `value`); it just has not made it into the guide.

`grails-doc/src/en/ref/Tags - GSP/set.adoc` also needs updating — it carries an explicit attribute list (`var`, `value`, `bean`, `scope`) and `type` is a new public attribute on a documented tag, so per the project's docs requirement it should be listed there with the value-only restriction noted.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
```diff
@@ -54,6 +55,26 @@ class GroovyPagePlugin implements Plugin<Project> {
...
+    /**
+     * Whether pages are compiled statically, from {@code grails { compileStatic { gsp } }}.
+     *
+     * <p>Resolved when the task runs rather than now: this plugin is applied on its own as well as
+     * alongside the one that registers the {@code grails} extension, and there is no ordering between
+     * them. Where the extension is absent, pages compile the way configuration alone says.</p>
+     */
+    private static Provider<Boolean> resolveCompileStaticPages(Project project) {
+        project.provider {
+            project.extensions.findByType(GrailsExtension)?.compileStatic?.gsp?.getOrElse(false) ?: false
```
jdaugherty: These two providers capture `project` and read `project.extensions` when they are resolved, and they feed `compileGroovyPages.compileStatic` / `compileStaticStrict`, which are `@Input`. That makes `Project` reachable from a task input, which is the pattern the configuration cache rejects.

It is latent today — `org.gradle.configuration-cache` is commented out in `gradle.properties` — but it is inconsistent with the rest of this PR and with the surrounding code: `GrailsGspCompileStaticProvider` deliberately holds only the `GrailsCompileStaticOptions` object, and `GrailsAppBaseDirProvider` carries a comment explaining that it exists to keep a plain `File` on the task rather than anything project-scoped.

Since the reason for deferring is plugin-application ordering, wiring it when the extension appears keeps the laziness without capturing the project:

```groovy
project.plugins.withType(GrailsGradlePlugin) {
    GrailsCompileStaticOptions options = project.extensions.getByType(GrailsExtension).compileStatic
    compileGroovyPages.configure { it.compileStatic.set(options.gsp.orElse(false)) }
    // ... and compileStaticStrict / compileWebappGroovyPages likewise
}
```

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
```diff
@@ -116,16 +222,99 @@ class GroovyPageTypeCheckingExtension extends GroovyTypeCheckingExtensionSupport
...
         }
-        expression.getNodeMetaData(StaticTypesMarker.DYNAMIC_RESOLUTION) != null
+        if (currentScope.pageScopeVariables.contains(expression.name)) {
+            return false
+        }
+        true
+    }
+
+    /**
+     * Whether the type of what a member is read from is not known, which is what {@code Object} means
```
jdaugherty: This Javadoc block describes `isUnknownReceiver`, but it is stacked directly above the Javadoc for `collectOperatorReceivers`, so `collectOperatorReceivers` has two doc comments and `isUnknownReceiver` (line 290) has none — it carries an inline comment instead. The `FRAMEWORK_DYNAMIC_NAMES` instance of this got fixed; this one is still stranded. Moving it down onto `isUnknownReceiver` finishes it.

grails-test-examples/gsp-compile-static/src/integration-test/groovy/gspstatic/GspCompileStaticSpec.groovy
```diff
@@ -0,0 +1,104 @@
...
+        then:
+        body.contains('<li id="book-0">Dune has 412 pages</li>')
+        body.contains('<li id="book-1">Emma has 474 pages</li>')
+        body.contains('<p id="count">2</p>')
+        body.contains('<p id="longest">Emma</p>')
+
+        and: 'a tag library in its own namespace is reached'
+        body.contains('<p id="shout">QUIET</p>')
+    }
+
```
jdaugherty, line 95: Good addition — an application that actually turns the option on is much stronger evidence than the unit specs alone.

One gap worth closing while it is fresh: nothing here asserts that the *forked page compiler* produced statically compiled classes. `pageClassFor` goes through `templateEngine.createTemplate(view)`, which resolves through the page locator at test time, so which class it hands back depends on whether `compileGroovyPages` output is on the integration-test runtime classpath rather than on anything the test states. And on the build side, `GroovyPagePluginFunctionalSpec` only asserts that the task's `compileStatic` input property is `true`, not that the fork honoured it.

Given that the central claim of the build option is that it "reaches both places a page is compiled", it would be good to have one assertion against a class the `compileGroovyPages` task actually wrote — loading the precompiled `gsp_...` class from the task's output directory and checking `CompileStaticGroovyPage.isAssignableFrom(...)` would do it. As it stands both existing assertions can be satisfied by the runtime path alone.

Also a smaller note: the two cases in `'a page that declares a model is compiled statically because it declared one'` are, as the `where:` comment says, static "whatever the build asked for" — so `frameworkNames.gsp` is the only page here that tests the build flag at all.

jdaugherty (commented):
Re-reviewed after the latest round. The three items from the previous pass hold up, and the new example application is a genuine improvement to the evidence for this feature. One correctness issue below is the same defect class this PR claims to have fixed, and two statements in the new guide page now contradict the feature as built.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
codeconsole: Confirmed and fixed in 16c7624 — all four of your cases threw at render, exactly as listed. Both tags emit a Groovy cast now, which is what `writeFrameworkSuppliedAccessors` already did.

Three of the four render correctly. `List` from `String[]` is now reported at compile time as `Inconvertible types` rather than throwing at render — better, but not silently coerced; a page wanting that writes `as List` in the value.

You were right about the tests agreeing with the bug: `${2L}` and `${1.5d}` are gone, replaced with `${2}` for both, plus GString and multi-expression cases.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/tags/GroovyDefTag.java
codeconsole: Fixed in the same commit. Both the cast and the parenthesisation.

On the pre-existing one: parenthesising alone was not enough — `calculateExpression` returns the raw text for a mixed value, so it was not Groovy to begin with. The value is emitted as a GString literal where it is neither a lone `${...}` nor plain text, which leaves `<g:def var="x" value="someVariable"/>` reading a variable as before. `${a} and ${b}` was broken the same way and is covered too.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
codeconsole: Right, that was left over from before the non-strict default. Rewritten in 9572cf6 to lead with the operator case as the one that accounts for most of the work, and to say that a page reading an undeclared variable still compiles.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
codeconsole: Rewritten rather than dropped: the block applies to controllers, services and tag libraries, `gsp` applies to pages, and `all = true` is the one that does not include it.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
codeconsole: Added in 9572cf6 — in the section the error message sends people to, with `g:def` beside it and the value-only restriction. `set.adoc` now lists `type` in its attribute list with the same restriction.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
codeconsole: Fixed in 16c7624, wired as you suggested. The tasks hold the `GrailsCompileStaticOptions` object rather than a provider reaching through the project, with `false` set at registration so the plugin applied on its own still has a value.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
codeconsole: Moved onto `isUnknownReceiver` in 16c7624, with the inline comment folded into it.

grails-test-examples/gsp-compile-static/src/integration-test/groovy/gspstatic/GspCompileStaticSpec.groovy
codeconsole: Valid, and worse than you described. Added in 5c45f23: the test reads the classes in the task output directory directly.

Wiring it up showed the gap was real — `integrationTest` has no dependency on `compileGroovyPages`, so on a clean build that directory is empty and the assertion would have been vacuous. My earlier runs only passed because I had run the task by hand. The dependency is stated now, and I checked the assertion both ways from a cleaned output directory: it fails with `gsp = false` and passes with it on.

That also answers the propagation question directly — with the flag off, `frameworkNames_gsp` compiles dynamically while `declared_gsp` stays static from its `model` directive, so the flag does reach the fork.

On the smaller note: agreed, and the test says so — the two model-directive pages are split into their own feature from the one that tests the flag.

sbglasius (approved):

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/tags/GroovyDefTag.java
```diff
@@ -59,10 +59,35 @@ public void doStartTag() {
...
+        if ((text.startsWith("\"") && text.endsWith("\"")) || (text.startsWith("'") && text.endsWith("'"))) {
+            text = text.substring(1, text.length() - 1).trim();
+        }
+        if (!text.contains("${")) {
+            return text;
+        }
+        if (text.startsWith("${") && text.endsWith("}") && text.indexOf("${", 2) < 0) {
+            return text.substring(2, text.length() - 1).trim();
+        }
+        return '"' + text.replace("\\", "\\\\").replace("\"", "\\\"") + '"';
```
jdaugherty: `text.replace("\\", "\\\\").replace("\"", "\\\"")` escapes the whole attribute, including the code inside its `${...}` — but by the time this runs the parser has already made the attribute valid Groovy. `populateMapWithAttributes` strips a lone `${...}` down to the expression it holds and wraps everything else in quotes, so what arrives here is either an expression or a (G)String literal. Re-escaping it breaks two shapes:

| page | emitted | outcome |
| --- | --- | --- |
| `<g:def type="String" var="s" value="a ${[1].collect { "x$it" }} b"/>` | `String s=(String) ("a ${[1].collect { \"x$it\" }} b")` | `Unexpected character: '\'` — does not compile |
| `<g:def var="s" value="cost: \${1}"/>` | `("cost: \\${1}")` | renders `cost: \1` |

The first fails to compile; the second is silently wrong output — the page escaped the dollar to keep `${1}` literal and gets a backslash plus an interpolation instead. Both go through the branch this line is in, and the untyped tag reaches it too, since `groovyExpressionFor` runs before the type is looked at.

`g:set type=` has neither problem, because it hands the parser's literal to `getExpressionText` untouched. Doing the same here fixes both:

```java
private static String groovyExpressionFor(String value) {
    String text = value.trim();
    if (text.contains("${")) {
        // Already Groovy: the parser stripped a lone ${...} to the expression it holds and quoted
        // anything else, so this is either that expression or a GString literal.
        return text;
    }
    if ((text.startsWith("\"") && text.endsWith("\"")) || (text.startsWith("'") && text.endsWith("'"))) {
        text = text.substring(1, text.length() - 1).trim();
    }
    return text;
}
```

I ran this: the nested-quote page renders `a [x1] b`, `value="cost: \${1}"` renders `cost: ${1}`, and `value='a ${["k": "v"]["k"]} b'` renders literally — a single-quoted attribute does not interpolate, which is what every other GSP attribute does with one, and better than the compile error it produces today. `GspCompileStaticConfigSpec` stays green, including the two mixed-value `g:def` rows, which round-trip through the new branch unchanged.

Those two rows (`"Total: ${1 + 1}"`, `"${1 + 1} and ${2 + 2}"`) are also exactly the shapes the escaping happens to survive, so a nested-quote value and an escaped `\${...}` would be worth adding beside them.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
```diff
@@ -39,40 +41,91 @@ import org.codehaus.groovy.transform.stc.StaticTypesMarker
...
+     * The operators the class writer emits directly, rather than leaving to a call site.
+     *
+     * <p>These are the ones that cannot be handed a receiver of no known type. A comparison or a
+     * logical operator is not among them: those report a type error of their own, which is an answer
+     * a page can act on, so the receiver of one is still resolved the way anything else is.</p>
+     */
+    private static final Set<Integer> WRITTEN_INTO_THE_CLASS = [
+            Types.LEFT_SQUARE_BRACKET,
+            Types.PLUS, Types.MINUS, Types.MULTIPLY, Types.DIVIDE, Types.INTDIV, Types.MOD, Types.POWER,
+            Types.PLUS_EQUAL, Types.MINUS_EQUAL, Types.MULTIPLY_EQUAL, Types.DIVIDE_EQUAL] as Set
```
jdaugherty: Three gaps in this set, each producing the outcome it exists to prevent.

`Types.MOD` is not the token this sees for `%`: with it in the set `${n % 2}` is still missed, and adding `Types.REMAINDER` is what catches it. So `%` is listed here and not covered. (`Types.INTDIV` looks to be in the same position — nothing in the language emits it.)

The shift and bitwise operators are absent, and the static writer emits them the same way it emits `getAt`:

| page | error |
| --- | --- |
| `${n % 2}`, `<% n %= 2 %>` | `Cannot access method: remainder() of class: java.lang.Object` |
| `${l << 1}`, `<% l <<= 1 %>` | `Cannot access method: leftShift() of class: java.lang.Object` |
| `${n >> 1}`, `${n >>> 1}` | `rightShift()`, `rightShiftUnsigned()` |
| `${n & 1}`, `${n \| 1}`, `${n ^ 1}` | `and()`, `or()`, `xor()` |
| `<% n **= 2 %>` | `power()` |

That message comes from `org.codehaus.groovy.classgen.asm.sc.StaticInvocationWriter`, i.e. the class writer — the failure `'an operator on a closure parameter of no known type is reported by type checking'` asserts against with `!message.contains('Cannot access method')`.

`<=>` is worse than an unhelpful message. It crashes the compiler:

```
General error during canonicalization: Index 0 out of bounds for length 0
  at org.codehaus.groovy.transform.sc.transformers.BinaryExpressionTransformer.transformRelationComparison(BinaryExpressionTransformer.java:310)
```

for `<g:set var="n" value="${1}"/>${n <=> 2}`, for `${undeclared <=> 2}`, and under a declared model too. Giving the receiver a type — `<g:set type="int">` or a `model` declaration — makes every row above compile and render, which is what puts the cause on the receiver rather than the operator.

Adding the missing tokens turns all of them into the page-level message:

```groovy
            Types.PLUS_EQUAL, Types.MINUS_EQUAL, Types.MULTIPLY_EQUAL, Types.DIVIDE_EQUAL,
            Types.MOD_EQUAL, Types.POWER_EQUAL, Types.REMAINDER, Types.REMAINDER_EQUAL,
            Types.LEFT_SHIFT, Types.RIGHT_SHIFT, Types.RIGHT_SHIFT_UNSIGNED,
            Types.BITWISE_AND, Types.BITWISE_OR, Types.BITWISE_XOR,
            Types.LEFT_SHIFT_EQUAL, Types.RIGHT_SHIFT_EQUAL, Types.RIGHT_SHIFT_UNSIGNED_EQUAL,
            Types.BITWISE_AND_EQUAL, Types.BITWISE_OR_EQUAL, Types.BITWISE_XOR_EQUAL,
            Types.COMPARE_TO] as Set
```

I ran this: every row reports `The type of [n] is not known here, and [%] cannot be applied to it`, the `<=>` crash becomes the same message, and `GspCompileStaticConfigSpec` stays green. `<`, `>`, `<=`, `>=` and `==` still compile on a receiver of no known type, so what the doc comment says about comparisons holds for them — `<=>` is the one that is not left to a call site.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
```diff
@@ -0,0 +1,245 @@
...
+grails {
+    compileStatic {
+        gsp = true
+    }
+}
+----
+
+This states the same `grails.views.gsp.compileStatic` setting, for both the pages the build compiles ahead of time and the application it runs, so a page compiles the same way in either. Where the build states it, it takes precedence over `application.yml` — the same order a system property takes over configuration in a running application. Set it in one place or the other rather than both.
+
+NOTE: `gsp` is not included in the `compileStatic { all = true }` shortcut described in link:staticTypeCheckingAndCompilation.html#grailsCompileStatic[GrailsCompileStatic]. The artefact opt-ins there fail on code that is doubtful anyway, whereas this one fails on any page reading a model variable it has not declared, so it is never enabled as a side effect of asking for everything.
```
jdaugherty: This NOTE still carries the pre-`strict` rationale that 9572cf6 corrected eleven lines below it. "this one fails on any page reading a model variable it has not declared" is the statement the new NOTE at line 204 contradicts — "A page that reads a model variable it has not declared still compiles" — and that line 54 contradicts as well.

The conclusion is still right, only the reason is stale. Something closer to: the artefact opt-ins fail on code that is doubtful anyway, whereas this one reports an operator applied to a value whose type nothing established, which most applications that never declared a page model have somewhere. Or keep the fact and drop the clause after the comma.

grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePluginFunctionalSpec.groovy
```diff
@@ -62,4 +62,34 @@ class GroovyPagePluginFunctionalSpec extends GradleSpecification {
...
+
+    def "pages compile the way configuration says where the opt-in is not set"() {
+        given:
+        setupTestResourceProject('gsp-compile-classpath')
+
+        when:
+        def result = executeTask('inspectGspCompileClasspath')
+
+        then: 'the project applies grails-gsp without the grails extension and still configures'
+        result.output.contains('HAS_GSP_COMPILE_CONFIGURATION=false')
```
jdaugherty: This asserts the same single line as `"plugin does not register a gspCompile configuration"` at the top of the class, and that line is about the removed `gspCompile` configuration — nothing in it is about the opt-in the test is named for. It passes whatever `compileStatic` ends up as.

The behaviour is worth a test, though, and this is the project that exercises it: `wireCompileStaticOptions` only reads the extension inside `plugins.withType(GrailsGradlePlugin)`, so a project applying `grails-gsp` on its own is the one where the extension never appears and the values set at registration have to stand. Printing them from the same task states it:

```groovy
// gsp-compile-classpath/build.gradle, inside inspectGspCompileClasspath
println "PAGE_COMPILER=${tasks.named('compileGroovyPages').get().compileStatic.get()}"
println "WEBAPP_PAGE_COMPILER=${tasks.named('compileWebappGroovyPages').get().compileStatic.get()}"
```

with `result.output.contains('PAGE_COMPILER=false')` here, which is the claim the name makes.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
```diff
@@ -175,6 +208,7 @@ public class GroovyPageParser implements Tokens {
     public static final String CONFIG_PROPERTY_GSP_GRAILS_LAYOUT_PREPROCESS = "grails.views.gsp.layout.preprocess";
     public static final String CONFIG_PROPERTY_GSP_COMPILESTATIC = "grails.views.gsp.compileStatic";
     public static final String CONFIG_PROPERTY_GSP_ALLOWED_TAGLIB_NAMESPACES = "grails.views.gsp.compileStaticConfig.taglibs";
+    public static final String CONFIG_PROPERTY_GSP_COMPILESTATIC_STRICT = "grails.views.gsp.compileStaticConfig.strict";
```
jdaugherty, line 211: This is a new settable configuration value, and it is not in `grails-gsp/plugin/src/main/resources/META-INF/spring-configuration-metadata.json` — the file `grails-doc/build.gradle` reads out of the `grails-gsp` jar to generate the configuration reference. As it stands the setting is described in prose on the new guide page but missing from the reference and from IDE completion.

Its siblings are already there (`grails.views.gsp.encoding`, `grails.views.gsp.layout.preprocess`, `grails.views.gsp.codecs.*`), so it is one entry:

```json
{
  "name": "grails.views.gsp.compileStaticConfig.strict",
  "type": "java.lang.Boolean",
  "description": "Whether a statically compiled page is held to the names it declares, rather than only pages that declare a model.",
  "defaultValue": false
}
```

`grails.views.gsp.compileStatic` and `grails.views.gsp.compileStaticConfig.taglibs` are absent too. Those predate this PR, but this is the PR that makes them the documented way to use the feature, so it looks like the moment to add all three.

grails-gsp/grails-web-taglib/src/main/groovy/org/grails/web/taglib/WebRequestTemplateVariableBinding.java
```diff
@@ -72,6 +72,13 @@ public Object evaluate(GrailsWebRequest webRequest) {
                 return webRequest.getServletContext();
             }
         });
+        // The same object as `application`, under the name the servlet API calls it. A page reading
+        // servletContext got nothing at all before, since nothing bound the name.
+        m.put("servletContext", new LazyRequestBasedValue() {
```
jdaugherty, line 77: This is the name the PR newly binds, and nothing exercises it. `WebRequestTemplateVariableBinding` has no spec of its own, `'the servlet scopes carry their own types'` covers `request`, `response`, `session` and `application` but not `servletContext`, and the example app's `frameworkNames.gsp` reads `controllerName`, `actionName`, `flash` and `params`. `webRequest` — the other name typed here for the first time — is in the same position.

Both do work: I compiled `${servletContext.serverInfo}` and `${webRequest.currentRequest}` statically and the accessors resolve with the types the guide's table gives them. A row each in that `where:` block would cover the typing, and `frameworkNames.gsp` is the natural place to prove the binding at render time — it is the same object as `application`, so `${application.serverInfo}` and `${servletContext.serverInfo}` in one page is the assertion.

Worth noting for the risk side of it: the new entry cannot shadow anything, since `getVariable` checks the model and the request attributes before the lazy map.

grails-test-examples/gsp-compile-static/grails-app/views/demo/index.gsp
```diff
@@ -0,0 +1,35 @@
...
+  specific language governing permissions and limitations
+  under the License.
+--%>
+<%@ page model="String title; java.util.List<gspstatic.Book> books" %>
+<!doctype html>
+<html>
+<head><title>${title}</title></head>
+<body>
+<h1>${title}</h1>
+<g:def type="int" var="total" value="${0}"/>
```
jdaugherty: `total` is declared here and never read. In the page that is the worked example of the feature it reads as a leftover — either use it, or drop it and leave the typed-local case to `declared.gsp`, which reads what its `g:set type=` declares.

jdaugherty (commented):
Third pass, over the three commits since the last round and the two merges after them.

The previous round's items hold up. I re-ran the four `Class.cast` cases: `long` and `double` from `${2}` render, `String` over a GString renders, and `String[]` into `List` is now an honest `Inconvertible types` at compile time. The page compilers hold the options object rather than a provider that reaches through the project, the stranded javadoc is on `isUnknownReceiver`, `type` is in the guide and in `set.adoc`, and the integration test now reads the classes `compileGroovyPages` wrote with the task dependency that makes that assertion mean something. The last two merges leave the tree identical to their upstream parent, so nothing was lost in restoring what the earlier merge dropped.

Tests are green as the branch stands: `:grails-gsp-core:test`, `:grails-gsp:test`, `:grails-web-gsp:test`, `:grails-web-taglib:test`, and the gradle plugin specs.

Two new correctness findings, both of the same shape this PR is otherwise careful about — something of no known type, or something already escaped, reaching a phase that cannot report it properly. `<=>` on an untyped value crashes the compiler outright, and a `g:def` value mixing text with an expression that contains a quote does not compile, while an escaped `\${...}` in one renders wrongly. Then one statement the docs commit missed, one new test that does not assert what its name says, a new config key missing from the metadata the configuration reference is generated from, and two names typed here that nothing exercises.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/tags/GroovyDefTag.java
codeconsole: Both reproduce, and your fix is the right one — taken as written in 05e11cc. The nested-quote page failed to compile and the escaped dollar rendered `cost: \1`; both render correctly now, and the mixed-text, two-expression and plain-variable cases still do. Two rows added to the render table.

You are right about the cause: I was escaping a value the parser had already made Groovy.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageTypeCheckingExtension.groovy
codeconsole: All of it reproduces — `%` giving `remainder()`, the shifts, the bitwise operators, the compound assignments, and `<=>` crashing canonicalization. Fixed in 05e11cc by listing the tokens rather than the symbols: `REMAINDER`, `COMPARE_TO`, the three shifts, the three bitwise operators and the compound forms. `MOD` and `INTDIV` stay, harmlessly, since nothing emits them.

One correction to the last line of your comment. Adding `COMPARE_TO` does not turn `<=>` into the page-level message — it puts that message *first*, and the transformer still fails afterwards with the same index out of bounds. Reporting the receiver does not stop the transformer that reads it from running. A test states that ordering rather than asserting a clean failure.

Thirteen operators are pinned reporting against an untyped receiver, and twelve compiling once it has a type. `**=` is the exception either way: with a type it fails with `Cannot assign value of type java.lang.Number`, which is ordinary static-compilation behaviour since `power()` returns `Number`, not over-reporting.

grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc
codeconsole: Right — the fact stands, the reason was stale. Reworded in c74ae38 to say it reports an operator applied to a value whose type nothing established, which most applications that never declared a page model have somewhere.

grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePluginFunctionalSpec.groovy
codeconsole: Fixed in c204890. The task now prints `PAGE_COMPILER_STATIC` and `WEBAPP_PAGE_COMPILER_STATIC`, and the test asserts both are false — which is the claim its name makes, and the one project where `wireCompileStaticOptions` never finds the extension.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
codeconsole: Added all three in c74ae38 — `grails.views.gsp.compileStatic`, `.compileStaticConfig.strict` and `.compileStaticConfig.taglibs`, beside `layout.preprocess` in `grails-gsp/plugin`.

grails-gsp/grails-web-taglib/src/main/groovy/org/grails/web/taglib/WebRequestTemplateVariableBinding.java
codeconsole: Both covered in c204890. `${servletContext.serverInfo}` and `${webRequest.currentRequest}` are rows in that `where:` block, and `frameworkNames.gsp` reads `servletContext.serverInfo == application.serverInfo` so the binding is proved rather than assumed.

grails-test-examples/gsp-compile-static/grails-app/views/demo/index.gsp
codeconsole: Dropped. `declared.gsp` already reads what its `g:set type=` declares, so the typed-local case is covered without a leftover in the worked example.

PR #16142, opened 2026-08-12
Record a GSP source checksum instead of a modification time

### LINKED ISSUE #16131: Precompiled GSP classes are not reproducible, defeating the build cache for every consumer

codeconsole:
Builds on @maczikasz's commit from #16132, preserved at the base of this branch. That commit diagnosed the problem and fixed the reproducibility half; the checksum layer on top is what keeps runtime reloading working.

## The problem

`GroovyPageCompiler` baked the `.gsp` source's modification time into every generated class as a `LAST_MODIFIED` constant. Git stores no modification times, so every fresh clone or CI checkout gives each source a new one, and byte-for-byte identical sources compile to different classes.

Because the value is a compile-time constant it belongs to the class's ABI and is inlined into callers, so the difference survives even Gradle's `COMPILE_CLASSPATH` normalization. Every task downstream of a jar carrying precompiled GSPs misses the build cache — #16132 measured roughly 616 hours of avoidable CI task re-execution over one week.

Release artifacts are unaffected: `release.yml` normalizes every file's modification time to a recorded `SOURCE_DATE_EPOCH` before building, and `verify-reproducible.sh` replays it, so published jars reproduce today. That normalization is release-only — ordinary CI runs and developer clones don't get it, which is where the cache misses come from.

## Why the timestamp couldn't simply be zeroed

`LAST_MODIFIED` is read at runtime. `GroovyPageMetaInfo.checkIfReloadableResourceHasChanged` compares it against the live source to decide whether a precompiled page is stale, and `DefaultGroovyPageLocator` installs a resource callable for *any* precompiled page when reloading is enabled — only the binary-plugin path nulls it.

Zeroing the constant alone switches off reloading for an application's own precompiled pages, which is documented behaviour ("GSP reloading is supported for precompiled GSPs since Grails 1.3.5"), and it fails silently — the page renders stale with nothing logged.

## The change

`GroovyPageParser` emits a `SOURCE_CHECKSUM` recording what the source *is* rather than when it was touched, and pages compiled at runtime record one too, so both paths decide staleness the same way. Identical sources now compile to identical bytes on every machine.

Reload detection gets more accurate than the timestamp it replaces: a page copied or checked out afresh is no longer treated as changed, and an edit is caught however close together two writes fall.

A `(mtime, length)` stamp is remembered when the source last matched, and the read is skipped while neither has moved, so steady state stays at one stat per page per check interval.

The recorded modification time is removed rather than left at `0`: every page Grails 8 produces has a checksum, so the constant was always `0` and the branch comparing it was unreachable.

## Branch

Targets `8.0.x` per @matrei. It adds public API — `GroovyPageParser.checksumOf`, plus `getSourceChecksum`/`setSourceChecksum` — which is what ruled out a patch branch. `checksumOf` can't be avoided: the compiler and `GroovyPageMetaInfo` are in different packages and must not disagree on how a source is digested.

## Notes

- **Removed API**: `GroovyPageMetaInfo.getLastModified()`, `setLastModified`, `applyLastModifiedFromResource`, and `GroovyPageParser.getLastModified()`/`setLastModified`. Generated pages no longer declare `LAST_MODIFIED`. `getSourceChecksum()` is the replacement. Covered in the Grails 8 upgrade guide.
- **Removed setting**: `grails.gsp.reload.granularity` absorbed rounding differences between modification times, which content comparison doesn't need. `grails.gsp.reload.interval` is unchanged.
- A GSP precompiled by Grails 7 inside a binary plugin carries no checksum and is never reload-checked. That was already true — the locator nulls the resource callable for views resolved out of a plugin jar, so the timestamp those classes carry was never consulted.
- `GroovyPageCompiler`'s own up-to-date check at line 214 is untouched. `GroovyPageForkCompileTask` is a `@CacheableTask` with content-hashed inputs, so that check is unreachable for anything built through Gradle.

Fixes #16131


sbglasius (approved):

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
```diff
@@ -215,11 +215,26 @@ class GroovyPageCompiler {
...
+                // `static final long`, so it belongs to the class's ABI and is inlined into callers, which
+                // even Gradle's COMPILE_CLASSPATH normalization cannot see past. Git stores no modification
+                // times, so every checkout gave each .gsp a new one and identical sources compiled to
+                // different bytes, costing every downstream consumer of the jar its build cache.
+                //
+                // SOURCE_CHECKSUM answers what the timestamp was only ever a proxy for -- has the source
+                // changed? -- and answers it identically on every machine. GroovyPageMetaInfo prefers it and
+                // falls back to LAST_MODIFIED for pages compiled by earlier versions, so the zero here means
+                // "no timestamp recorded", never "never reload".
+                gpp.lastModified = 0L
```
jdaugherty: Isn't this a breaking change?  I'm fine with it being removed long term, but setting this value here could cause downstream adopters to break.  @matrei are you ok with such a change in 7.x?

grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageMetaInfo.java
```diff
@@ -474,10 +518,21 @@ public Resource checkIfReloadableResourceHasChanged(final PrivilegedAction<Resou
...
+                        String currentChecksum = establishChecksum(resource);
+                        return (currentChecksum != null && !sourceChecksum.equals(currentChecksum)) ? resource : null;
+                    }
+                    // Pages compiled before SOURCE_CHECKSUM existed, and pages compiled at runtime, still carry a
+                    // real timestamp. A lastModified of 0 means none was recorded, so there is nothing to compare.
                     long currentLastmodified = establishLastModified(resource);
                     // granularity is required since lastmodified information is rounded somewhere in copying & war (zip) file information
                     // usually the lastmodified time is 1000L apart in files and in files extracted from the zip (war) file
-                    if (currentLastmodified > 0 && Math.abs(currentLastmodified - lastModified) > LASTMODIFIED_CHECK_GRANULARITY) {
+                    if (currentLastmodified > 0 && lastModified > 0 &&
```
jdaugherty: `establishLastModified` returns `-1` when the resource's timestamp cannot be read (`IOException`/`FileNotFoundException`), and `File.lastModified()` returns `0` on I/O error — `applyLastModifiedFromResource` stores whichever it got. A runtime-compiled page (no checksum) that recorded such a value used to self-heal here: `|current − (−1)|` always exceeded the granularity, so the first check that could read a real mtime reloaded the page once and re-recorded a valid timestamp. With `lastModified > 0`, that page can never reload until restart, and the failure is silent.

The trigger is narrow (a transient I/O failure while the meta info is built, or a URL resource that reports no timestamp), but the precompiler's deliberate sentinel is only ever `0`, so `-1` can keep its old meaning — e.g. `lastModified != 0` preserves the recovery path while still treating `0` as "nothing recorded".

grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageMetaInfo.java
```diff
@@ -474,10 +518,21 @@ public Resource checkIfReloadableResourceHasChanged(final PrivilegedAction<Resou
             public Resource call() {
                 Resource resource = resourceCallable.run();
                 if (resource != null && resource.exists()) {
+                    // A page compiled by GroovyPageCompiler records a checksum of its source rather than the
+                    // source's modification time, which git does not preserve across a checkout. Comparing
+                    // content answers the question directly: a page that was merely touched is not stale, and
+                    // an edit is caught however close together the writes fall.
+                    if (sourceChecksum != null) {
+                        String currentChecksum = establishChecksum(resource);
```
jdaugherty: Each staleness check for a checksum-bearing page now opens and fully reads the source to hash it, where the timestamp path cost one stat. It is gated per page per `grails.gsp.reload.interval` (5 s), but in the reload-enabled deployed scenario this feature targets — many views, sources on NFS/shared storage — that is a full file read per hot page per interval on request threads, indefinitely, even when nothing changes.

A cheap pre-check would restore stat-level steady-state cost: remember the `(lastModified, contentLength)` observed when the checksum was last computed and only re-hash when either moves. The only edit that slips through is one preserving both mtime and length — still strictly better than the old 2000 ms granularity window, and the spec's "edit within the granularity window" feature keeps passing since its edit changes the length. Not blocking, just worth weighing before this meets an app with hundreds of views.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
jdaugherty: A compatibility note worth capturing in the release notes: `LAST_MODIFIED = 0` only means "nothing recorded" to a runtime that contains this change. A pre-7.0.16 `grails-gsp` reading a class precompiled by this compiler evaluates the old condition — `|currentLastmodified − 0| > granularity`, true for any real mtime — so with reload enabled every such page is declared stale on its first check and recompiled from source at runtime (e.g. a plugin precompiled with 7.0.16+ consumed by an app still resolving an older 7.0.x). Self-correcting and reload-only, but surprising when it hits.

Related: `GroovyPageMetaInfo`'s constructor still reads `LAST_MODIFIED` unguarded (`GroovyPageMetaInfo.java:132`), unlike the null-guarded `SOURCE_CHECKSUM` read — so the constant is now permanent ABI that must keep being emitted even though its value is always `0`. A sentence on the constant would keep a future cleanup from dropping it.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
```diff
@@ -1380,6 +1391,54 @@ public void setLastModified(long lastModified) {
...
+     * Both the compiler that writes the constant and the runtime that compares against it use this method, so
+     * that the two can never disagree on how a GSP source is digested. The stream is read to the end but is
+     * not closed; closing it remains the caller's responsibility.
+     *
+     * @param source the raw bytes of the GSP source
+     * @return the checksum as a lower-case hex string
+     * @throws IOException if the source cannot be read
+     * @since 7.0.16
+     */
+    public static String checksumOf(InputStream source) throws IOException {
```
jdaugherty: The javadoc already describes the parameter as "the raw bytes of the GSP source" — and both callers actually hold the bytes: `GroovyPageCompiler` reads `gspfile.bytes` and wraps them in a `ByteArrayInputStream` solely to fit this signature, and `establishChecksum` could use `resource.getContentAsByteArray()` (spring-core 6.0.5+). A `checksumOf(byte[])` delegating to `MessageDigest.digest(byte[])` would drop the manual read loop, both stream wrappers in the compiler, and the "closing a ByteArrayInputStream is a no-op" comment — and shrink the new public surface to something harder to misuse.

Whichever shape stays, it is worth spelling out that the input must be the raw stored bytes of the page — not the decorated/re-encoded source the runtime parse path works with — since that is the invariant that keeps a compile-time checksum comparable with `establishChecksum`'s raw read at reload time.

grails-doc/src/en/guide/theWebLayer/gsp/makingChangesToADeployedApplication.adoc
```diff
@@ -46,8 +46,10 @@ There are also some system properties to control GSP reloading:
...
-|grails.gsp.reload.granularity|the number of milliseconds leeway to give before deciding a file is out of date. this is needed because different roundings usually cause a 1000ms difference in lastmodified times|1000
+|grails.gsp.reload.interval|interval between checks of the gsp source file, unit is milliseconds|5000
+|grails.gsp.reload.granularity|the number of milliseconds leeway to give before deciding a file is out of date. this is needed because different roundings usually cause a 1000ms difference in lastmodified times. Applies only to pages compared by modification time — see below|2000
 |===
 
 GSP reloading is supported for precompiled GSPs since Grails 1.3.5.
+
+A precompiled GSP records a checksum of the page it was compiled from, and is reloaded when the source no longer matches that checksum. Comparing content rather than modification times means a page that was copied or checked out afresh — and so carries a new modification time but the same content — is not needlessly recompiled, and an edit is detected however close together two writes fall. Pages compiled at runtime are still compared by modification time, and so is any page precompiled by Grails 7.0 or earlier.
```
jdaugherty: This ships in a 7.0.x release, so "precompiled by Grails 7.0 or earlier" reads as including the very version that introduces the checksum. Suggest bounding it at the actual version:

```suggestion
A precompiled GSP records a checksum of the page it was compiled from, and is reloaded when the source no longer matches that checksum. Comparing content rather than modification times means a page that was copied or checked out afresh — and so carries a new modification time but the same content — is not needlessly recompiled, and an edit is detected however close together two writes fall. Pages compiled at runtime are still compared by modification time, and so is any page precompiled by a version of Grails earlier than 7.0.16.
```

grails-gsp/core/src/test/groovy/org/grails/gsp/compiler/GroovyPageCompilerReproducibilitySpec.groovy
```diff
@@ -0,0 +1,109 @@
...
+    void setup() {
+        this.viewsDir = new File(this.tempDir, 'views')
+        this.page = new File(this.viewsDir, 'index.gsp')
+        this.page.parentFile.mkdirs()
+        this.page.text = PAGE_CONTENT
+    }
+
+    void 'a source compiled at two different modification times produces identical classes'() {
+        when: 'the same page is compiled twice, as two checkouts of one commit would'
+        this.page.setLastModified(1_000_000_000_000L)
```
jdaugherty: `File.setLastModified()` fails silently by returning `false` on filesystems that reject explicit mtime changes (container overlay FS, some CI volumes). If that happens here and at line 57, both compilations see the same mtime and `first == second` passes vacuously — the exact regression this spec exists to catch would ship undetected. Asserting the return value makes the environment problem loud instead:

```suggestion
        assert this.page.setLastModified(1_000_000_000_000L)
```

(same at line 57)

grails-gsp/core/src/test/groovy/org/grails/gsp/compiler/GroovyPageCompilerReproducibilitySpec.groovy
```diff
@@ -0,0 +1,109 @@
...
+        File targetDir = new File(this.tempDir, name)
+        Map results = compile(targetDir)
+        new File(targetDir, "${results.values().first()}.class").bytes
+    }
+
+    private GroovyPageMetaInfo compileToMetaInfo(String name) {
+        File targetDir = new File(this.tempDir, name)
+        Map results = compile(targetDir)
+        ClassLoader loader = new URLClassLoader([targetDir.toURI().toURL()] as URL[], getClass().classLoader)
+        new GroovyPageMetaInfo(loader.loadClass(results.values().first() as String))
```
jdaugherty: `URLClassLoader` is `Closeable`, and each call here leaks one loader (with its open file handles) for the life of the test JVM. On Windows those handles can block `@TempDir` cleanup, failing an otherwise green spec. Both uses of the loader complete before the closure returns, so it can be closed immediately:

```suggestion
        new URLClassLoader([targetDir.toURI().toURL()] as URL[], getClass().classLoader).withCloseable { URLClassLoader loader ->
            new GroovyPageMetaInfo(loader.loadClass(results.values().first() as String))
        }
```

grails-gsp/core/src/test/groovy/org/grails/gsp/GroovyPageMetaInfoReloadSpec.groovy
```diff
@@ -0,0 +1,180 @@
...
+        expect:
+        metaInfo.shouldReload(callableFor(resource))
+    }
+
+    void 'a source that was touched but not edited is not reported as stale'() {
+        given: 'a page whose source carries a modification time nothing like the one it was compiled at'
+        Resource resource = sourcePage()
+        GroovyPageMetaInfo metaInfo = new GroovyPageMetaInfo()
+        metaInfo.sourceChecksum = checksumOf(resource)
+        resource.getFile().setLastModified(resource.getFile().lastModified() + 86_400_000L)
```
jdaugherty: Same `setLastModified()` caveat as in the reproducibility spec, but here it erodes the premise rather than the assertion: this feature (and the pins back to `originalTimestamp` at lines 113 and 126) still pass via the checksum path if the call silently returns `false`, but then they no longer exercise the scenario their names document — "touched but not edited" runs with an unmoved mtime, and "edit within the granularity window" runs with an mtime that moved well past the granularity, which the old timestamp path already caught. An `assert` on all three keeps the coverage honest.

jdaugherty (dismissed):
The mechanism is sound and the test coverage is unusually thorough — the fallback matrix in `GroovyPageMetaInfoReloadSpec` covers the cases that matter. The inline comments are mostly about hardening edges: the `-1` error sentinel caught by the new guard, cross-version behavior of `LAST_MODIFIED = 0`, steady-state I/O of the checksum check, and two test-hygiene items that could silently hollow out the new specs.

Two thoughts that don't attach to a diff line:

- `buildPageMetaInfo` already materializes the full source when compiling at runtime (`GroovyPagesTemplateEngine.java:570`), so recording a checksum on that path too would be nearly free and would unify the two staleness mechanisms — dev-mode edits inside the 2000 ms granularity window are exactly the misses the checksum was built to catch, and it is the precondition for ever retiring the timestamp branch. Fine as a follow-up.
- `getLastModified()` returning `0` for newly precompiled pages is disclosed in the description, but the getter itself carries no documentation of the new convention; a javadoc line there pointing at `getSourceChecksum()` would cover external callers better than a release note alone.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
codeconsole: Retargeted to `7.1.x`.

grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageMetaInfo.java
codeconsole: Fixed in `45e2452cb9`. Guard is now `lastModified != 0`, so `-1` keeps self-healing and only `0` means "nothing recorded". Added a spec that fails if reverted to `> 0`.

grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPageMetaInfo.java
codeconsole: Implemented in `12fed6255c`. Remembers the `(mtime, length)` the source had when it last matched, and skips the read while neither has moved.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
codeconsole: Cross-version note added to the 7.0 → 7.1 upgrade guide. Comment added at the unguarded `LAST_MODIFIED` read explaining it is now permanent ABI.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageParser.java
codeconsole: Done in `e4055c2cb7` and `7920a56816`. `checksumOf(byte[])`, and both stream wrappers are gone — the `InputStream` constructor just delegates to the `String` one, which I'd missed. Raw-bytes invariant is in the javadoc.

grails-doc/src/en/guide/theWebLayer/gsp/makingChangesToADeployedApplication.adoc
codeconsole: Applied, bounded at 7.1.6.

grails-gsp/core/src/test/groovy/org/grails/gsp/compiler/GroovyPageCompilerReproducibilitySpec.groovy
codeconsole: Applied at both sites.

grails-gsp/core/src/test/groovy/org/grails/gsp/compiler/GroovyPageCompilerReproducibilitySpec.groovy
codeconsole: Applied — now `withCloseable`.

grails-gsp/core/src/test/groovy/org/grails/gsp/GroovyPageMetaInfoReloadSpec.groovy
codeconsole: Applied at all three sites.

matrei:
@codeconsole Can you give me the TLDR of why this is going into 7 and not 8?

> Targets 7.1.x rather than a patch branch...

How would this not be going into a patch version?

codeconsole:
@matrei

**Why 7 and not 8:** the bug is shipping from 7.x now — jars carrying precompiled GSPs don't reproduce, and `verify-reproducible.sh` has no GSP exclusion. 8.0.x isn't released, and 7.x merges up to it anyway, so landing here covers both. Landing only on 8.0.x would leave 7.x broken.

**Why not a patch:** it adds public API — `GroovyPageParser.checksumOf`, plus `getSourceChecksum`/`setSourceChecksum`. The checklist says patch branches take bug fixes only, no API changes, which is what @jdaugherty flagged. `checksumOf` can't be avoided: the compiler and `GroovyPageMetaInfo` are in different packages and must not disagree on how a source is digested.

If you're fine with the API addition on a patch branch, 7.0.x works and I'll move it back.

matrei:
> **Why not a patch:** it adds public API

@codeconsole I don't understand, if I'm not totally mistaken, this would land in 7.1.6, which is a patch release.
You would need a 7.3.0 release for it to not be a patch release.

At this point, I think it would be wise to put this in 8.

codeconsole:
@matrei the original target was 7.0.x.   because of the pushback, I just bumped it on a less iterated version.  I think it should go into 7.0.x, but I really don't care as long as it gets into 8.0.x.   

Not having it is obviously causing a lot of build issues, so if we are going to keep doing releases for 7.x, it would be a good idea to get it in there unless you think it is breaking.

Where do you recommend we put it?

matrei:
> GroovyPageCompiler baked the .gsp source's modification time into every generated class as a LAST_MODIFIED constant.

@jdaugherty Wasn't `8.0.0-M5` verified as fully reproducible?

> Where do you recommend we put it?

@codeconsole 8.0.0

codeconsole:
@matrei review the original PR this built on top of https://github.com/apache/grails-core/pull/16132 

> Over 2026-07-29 → 2026-08-05 on develocity.apache.org, roughly **616** hours of avoidable CI task
re-execution in this project trace to jars carrying precompiled GSPs

I moved to `8.0.x`  Rebased there, upgrade note moved into the Grails 8 guide, version references now 8.0.0. 
@maczikasz's commit is still at the base. (16132)

jdaugherty:
@matrei I think the AI is overstating the reproducible problem. I think with the SOURCE_EPOCH settings technically the build is reproducible (which is why it works for us).  I think this is meant to address a build problem - where the gradle cache is never hit due to unique gsp files.  This seems like a real issue.  I'm ok with this going into 8, but if it does, should we even keep the last modified date?  It seems like a checksum is a better value to keep instead of a date? Then it's cache stable always.

codeconsole:
@jdaugherty correct, the word reproducibility is being used in a different context here and no the same context we are using to verify builds.

codeconsole:
Removed the recorded modification time entirely (`f39a7f6e1e`), per @jdaugherty. Every Grails 8 page carries a checksum, so `LAST_MODIFIED` was always `0` and the branch reading it was unreachable. Gone with it: the field and accessors on both classes, `applyLastModifiedFromResource`, and `grails.gsp.reload.granularity`, which only existed to absorb timestamp rounding. `establishLastModified` stays — the stamp pre-check still uses it. Upgrade guide updated with the removed API and setting.

jdaugherty:
@matrei, I think i'm ok to merge this.  Are you?  @jamesfredley do you have concerns?

matrei:
> @matrei, I think i'm ok to merge this.  Are you?  @jamesfredley do you have concerns? 

I have not looked at it yet, I'll have a look first thing tomorrow if not merged by then.

grails-gsp/core/src/test/groovy/org/grails/gsp/GroovyPagesTemplateEngineChecksumSpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+import spock.lang.Specification
+
+import org.springframework.core.io.ByteArrayResource
+
+import org.grails.gsp.compiler.GroovyPageParser
+
+/**
+ * A page compiled at runtime records a checksum of its source as well as a modification time, so that it is
+ * checked for staleness the same way a precompiled page is.
```
matrei: Is this sentence correct?

matrei (approved):
Looks solid! Just one small javadoc review question?

grails-gsp/core/src/test/groovy/org/grails/gsp/GroovyPagesTemplateEngineChecksumSpec.groovy
codeconsole: No, it was wrong — thanks. It said the page records a checksum "as well as a modification time", which was true when I wrote it and stopped being true when the timestamp was removed. Fixed in `32f1eaf935`.

codeconsole:
> Wasn't `8.0.0-M5` verified as fully reproducible?

@matrei I checked the published artifact rather than guess. `grails-fields-8.0.0-M5.jar` contains precompiled GSP classes, and one of them decompiles to:

```
public static final long LAST_MODIFIED = 1785699288000l;
```

That decodes to 2026-08-02 19:34:48 UTC — the wall-clock time of the release builder's checkout. Anyone rebuilding from their own checkout gets a different value in that constant, so the class bytes and therefore the jar differ.

So M5 is not byte-reproducible for the jars carrying precompiled GSPs. Either the verification didn't cover those, or the diff was put down to the platform note in `RELEASE.md` ("Grails is officially built on Linux so if there are differences they may be due to the OS platform"). Worth re-checking against M5 if you want independent confirmation — `grails-fields` and `grails-spring-security` are the ones to look at.

Also merged latest `8.0.x` in rather than rebasing, so your `test: cleanup` commit is preserved rather than force-pushed over.

matrei:
> > Wasn't `8.0.0-M5` verified as fully reproducible?
> 
> @matrei I checked the published artifact rather than guess. `grails-fields-8.0.0-M5.jar` contains precompiled GSP classes, and one of them decompiles to:
> 
> ```
> public static final long LAST_MODIFIED = 1785699288000l;
> ```
> 
> That decodes to 2026-08-02 19:34:48 UTC — the wall-clock time of the release builder's checkout. Anyone rebuilding from their own checkout gets a different value in that constant, so the class bytes and therefore the jar differ.
> 
> So M5 is not byte-reproducible for the jars carrying precompiled GSPs. Either the verification didn't cover those, or the diff was put down to the platform note in `RELEASE.md` ("Grails is officially built on Linux so if there are differences they may be due to the OS platform"). Worth re-checking against M5 if you want independent confirmation — `grails-fields` and `grails-spring-security` are the ones to look at.

@codeconsole The release workflow creates a BUILD_DATE file and sets that date on all the files before building. The date stored in that file is then used in the verification script when building the jars again before comparing the checksums.

codeconsole:
You're right, that was an overstatement, but the original intention of this PR was 

  - Build-cache misses — #16132's ~616 hours, measured on ordinary CI, unaffected by release-only normalization
  - Reload accuracy — touched-but-unchanged pages no longer recompile; sub-granularity edits are now caught
  - Less machinery — LAST_MODIFIED and grails.gsp.reload.granularity both gone

I traced it: the M5 jar's `LAST_MODIFIED = 1785699288000` is exactly commit `3d129b7140` (2026-08-02 19:34:48 UTC), not a build wall-clock time. `release.yml:70-72` derives `SOURCE_DATE_EPOCH` from `git log -1 --pretty=%ct` and touches every file to it before building, and `verify-reproducible.sh:50` replays the same epoch from the published `BUILD_DATE`. So M5 is byte-reproducible and release verification is not affected by this. I'd read that script looking for GSP exclusions and missed the normalization a few lines above.

I've corrected the PR description.

What the change still buys is the build-cache side, which is where #16132's measurement came from: the mtime normalization is release-only, so ordinary CI runs and developer clones still produce differing `gsp-classes`, and every task downstream of those jars re-keys. A checksum also means the GSP case no longer depends on that normalization being applied.

PR #16143, opened 2026-08-12
Add test coverage for AbstractDetachedCriteria and rx DetachedCriteria

borinquenkid:
## Summary
- Adds mock-based unit specs for `AbstractDetachedCriteria` (via `grails.gorm.DetachedCriteria`) and for the reactive `grails.gorm.rx.DetachedCriteria`, taking both from ~0% to full line/method coverage without needing a real datastore.
- Fixes two real bugs surfaced while writing the rx specs:
  - `buildQueryableCriteria()` cast the built `DetachedCriteria` to `QueryableCriteria`, but the rx class never implemented that interface, so every closure-based subquery (`in`, `inList`, `notIn`, `eqAll`/`gtAll`/`ltAll`/`geAll`/`leAll`, `gtSome`/`geSome`/`ltSome`/`leSome`) threw a `ClassCastException` at runtime. Fixed with a small `SubqueryAdapter` that extends the shared `AbstractDetachedCriteria` base directly, since the reactive class's own `find()`/`list()` return `Observable` and can't coexist with `QueryableCriteria`'s `T`/`List<T>` signatures on the same type.
  - `prepareQuery()` applied fetch strategies (`join`/`select`) twice: once via `DynamicFinder.applyDetachedCriteria()`, then again via a redundant hand-rolled loop that also ignored custom `JoinType`s. Removed the dead duplicate.
- Fixes a handful of definite-assignment/shadowing/raw-`getAt` warnings in `AbstractDetachedCriteria` (uninitialized `prop` in `createAlias`, a local variable named `criteria` shadowing the instance field of the same name in `clone()`, and negative-index `List` access replaced with `getLast()`).

## Test plan
- [x] `:grails-datamapping-core:test` — 71 new tests in `AbstractDetachedCriteriaSpec`, all pass
- [x] `:grails-datamapping-rx:test` — 45 new tests across `DetachedCriteriaSpec` and `DetachedCriteriaQuerySpec`, all pass
- [x] `:grails-datamapping-core:codeStyle` and `:grails-datamapping-rx:codeStyle` clean
- [x] Verified via `jacocoTestReport` that `AbstractDetachedCriteria` and `grails.gorm.rx.DetachedCriteria` (plus its new `SubqueryAdapter`) are at ~100% line/method coverage


jdaugherty:
Let's just combine this PR with the other one?  It looks like the tests that are commented were never uncommented and we can do all of this in one PR. @borinquenkid

borinquenkid:
> Let's just combine this PR with the other one? It looks like the tests that are commented were never uncommented and we can do all of this in one PR. @borinquenkid

@jdaugherty I would prefer to keep them separate so they are as focused as possible

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/query/criteria/AbstractDetachedCriteria.groovy
```diff
@@ -216,7 +216,7 @@ abstract class AbstractDetachedCriteria<T> implements Criteria, Cloneable {
             }
         }
         if (junctions)  {
-            junctions[-1].add(criterion)
+            junctions.getLast().add(criterion)
```
jdaugherty: -1 is an accepted way to get the last when the array is defined, why change it?

grails-datamapping-rx/src/test/groovy/org/grails/gorm/rx/api/DetachedCriteriaQuerySpec.groovy
```diff
@@ -0,0 +1,295 @@
...
+            getJavaClass() >> QueryTestEntity
+            getName() >> QueryTestEntity.name
+        }
+        def staticApi = Mock(RxGormStaticApi) {
+            getDatastoreClient() >> datastoreClient
+        }
+        datastoreClient.createStaticApi(persistentEntity, ConnectionSource.DEFAULT) >> staticApi
+        datastoreClient.createInstanceApi(persistentEntity, ConnectionSource.DEFAULT) >> Mock(RxGormInstanceApi)
+        datastoreClient.createValidationApi(persistentEntity, ConnectionSource.DEFAULT) >> Mock(RxGormValidationApi)
+        RxGormEnhancer.registerEntityWithConnectionSource(persistentEntity, ConnectionSource.DEFAULT, ConnectionSource.DEFAULT, datastoreClient)
```
jdaugherty: Almost all of this spec duplicates coverage that `grails.gorm.rx.DetachedCriteriaSpec` already provides on `8.1.x` (find/findAll/get/toList/list/getCount/updateAll/deleteAll/toQuery/toObservable/subscribe, the prepareQuery max/offset and fetch-strategy paths, and additional-criteria merging). The only new behavior covered here is the custom `JoinType` passthrough.

That existing spec also shows this setup works through the public `RxGormEnhancer.registerEntity(entity, datastoreClient)` with a stubbed `ConnectionSources`, so the package-local placement to reach the protected `registerEntityWithConnectionSource` is not needed.

Suggest moving the "prepareQuery passes a custom JoinType through to the query" feature method into the existing `DetachedCriteriaSpec` and dropping this file.

This is the reason I think this should have been merged into the other PR

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/criteria/AbstractDetachedCriteriaSpec.groovy
```diff
@@ -0,0 +1,1167 @@
...
+import spock.lang.Specification
+
+/**
+ * Exercises {@link AbstractDetachedCriteria} through its concrete subclass {@link DetachedCriteria}.
+ */
+class AbstractDetachedCriteriaSpec extends Specification {
+
+    void "eq adds an Equals criterion"() {
+        given:
+        def criteria = new DetachedCriteria(TestEntity)
```
jdaugherty, line 41: `TestEntity` here resolves to the class declared at the bottom of `DetachedCriteriaCloneSpec.groovy`. Depending on a fixture defined inside another spec's file is fragile - refactoring that spec would silently break this one. Suggest either moving `TestEntity` into its own file in this package or declaring a dedicated fixture class in this file.

PR #16144, opened 2026-08-13
Support escaping a GSP expression with a backslash  \${}

codeconsole:
### Summary

A backslash immediately before a GSP expression now escapes it: `\${...}` renders a literal `${...}` with the backslash removed, instead of the expression being evaluated on the server.

GSP previously had no escape syntax at all — `\${x}` rendered a stray `\` followed by the evaluated expression. The motivating case is JavaScript template literals, whose interpolation placeholders share the `${}` syntax and are today either evaluated against the page binding (an unbound name silently renders as empty) or blow up the whole page render:

```html
<script>
    const name = 'Grails';
    console.log(`Hello \${name}`);
</script>
```

The browser now receives the template literal with its `${name}` placeholder intact, to be interpolated by JavaScript.

### Implementation

Escaping is resolved entirely at parse time, in two small pieces:

- `GroovyPageScanner` no longer starts an expression token when the `${` is immediately preceded by a backslash in template text — the sequence stays in the surrounding HTML token.
- `GroovyPageParser.html()` rewrites `\${` to `${` when writing the HTML token out.

Because this happens before code generation, behavior is identical in development, production, and `compileStatic` modes, and escaped text is invisible to static type checking — an escaped `${undeclaredName}` under `<%@ page compileStatic="true" %>` neither fails compilation nor evaluates (covered by a new test in `GspCompileStaticSpec`).

### Semantics

- Only the backslash performing the escape is removed: `\\${x}` renders `\${x}`.
- Each expression is escaped individually.
- Expressions in the attributes of GSP tags are scanned in a different lexer state and cannot be escaped this way.

### Also in this PR

`grails-gsp-core` declared `junit-jupiter-api` but never the Jupiter engine, so the JUnit Platform silently skipped every Jupiter-annotated test class in the module — all of `GroovyPagesTemplateEngineTests` (23 tests) never ran. Added the repo-standard `testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'`; all resurrected tests pass. The other engine-less modules with Jupiter tests (`grails-gsp/plugin`, `grails-gsp/grails-taglib`) receive the engine transitively and were already running them.

### Tests and docs

- `ScanTests`: escaped `${` stays in the HTML token; unescaped expressions still tokenize; mixed escaped/unescaped side by side.
- `GroovyPagesTemplateEngineTests`: literal render, escaped next to evaluated, JavaScript template literal end-to-end, double-backslash semantics.
- `GspCompileStaticSpec`: escaped expressions are invisible to static compilation.
- Guide: escaping documented in the GSP Basics "Expressions" section with the JavaScript template literal example.

Full `:grails-gsp-core:test` and `:grails-gsp:test` suites plus downstream `:grails-taglib`, `:grails-sitemesh3`, `:grails-layout`, and `:grails-web-gsp` all pass; `codeStyle` is clean on the touched modules.

https://claude.ai/code/session_01WtCxMe9pfmy7tiHCUr5THN

### DIFF
--- a/grails-doc/src/en/guide/theWebLayer/gsp/GSPBasics/expressions.adoc
+++ b/grails-doc/src/en/guide/theWebLayer/gsp/GSPBasics/expressions.adoc
@@ -30,4 +30,16 @@ In GSP the `<%= %>` syntax introduced earlier is rarely used due to the support 
 
 However, unlike JSP EL you can have any Groovy expression within the `${..}` block.
 
+A backslash immediately before an expression escapes it: the expression is not evaluated, and the page renders a literal `${..}` without the backslash. This matters most for JavaScript template literals, whose interpolation placeholders share the `${..}` syntax and would otherwise be evaluated on the server:
+
+[source,xml]
+----
+<script>
+    const name = 'Grails';
+    console.log(`Hello \${name}`);
+</script>
+----
+
+The browser receives the template literal with its `${name}` placeholder intact, to be interpolated by JavaScript. Only the backslash performing the escape is removed, and each expression must be escaped individually. Expressions in the attributes of GSP tags cannot be escaped this way.
+
 WARNING: Embedding data received from user input has the risk of making your application vulnerable to a Cross Site Scripting (XSS) attack. Please read the documentation under link:security.html#xssPrevention[XSS prevention] for information on how to prevent XSS attacks.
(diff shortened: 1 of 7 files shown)


sbglasius (approved):

matrei (commented):
Can we add a note in the upgrade section as it could break someones gsp output?

jdaugherty (approved):

jdaugherty:
@codeconsole I had merged this before seeing @matrei  comment. can you please update the upgrade section?

PR #16146, opened 2026-08-13
Fix 16145: Oversized multipart uploads cannot be handled by Grails application code

### LINKED ISSUE #16145: [Grails 8] Oversized multipart uploads cannot be handled by Grails application code
### Issue description

matrei:
Closes #16145

### DIFF
--- a/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
+++ b/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
@@ -106,7 +106,7 @@ public class ControllersAutoConfiguration {
     @ConditionalOnMissingBean(value = MultipartFilter.class, parameterizedContainer = FilterRegistrationBean.class)
     public FilterRegistrationBean<Filter> multipartFilter() {
         var registrationBean = new FilterRegistrationBean<>();
-        var multipartFilter = new MultipartFilter();
+        var multipartFilter = new GrailsMultipartFilter();
         multipartFilter.setMultipartResolverBeanName(GrailsApplication.MULTIPART_RESOLVER_BEAN);
         registrationBean.setFilter(multipartFilter);
         registrationBean.addUrlPatterns(Settings.DEFAULT_WEB_SERVLET_PATH);
--- a/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
+++ b/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
@@ -28,6 +28,7 @@ import jakarta.servlet.MultipartConfigElement;
 
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBooleanProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
 import org.springframework.boot.servlet.autoconfigure.HttpEncodingAutoConfiguration;
@@ -103,6 +104,7 @@ public class ControllersAutoConfiguration {
     }
 
     @Bean
+    @ConditionalOnBooleanProperty(name = "spring.servlet.multipart.enabled", matchIfMissing = true)
     @ConditionalOnMissingBean(value = MultipartFilter.class, parameterizedContainer = FilterRegistrationBean.class)
     public FilterRegistrationBean<Filter> multipartFilter() {
         var registrationBean = new FilterRegistrationBean<>();
--- a/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
+++ b/grails-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersAutoConfiguration.java
@@ -112,9 +112,10 @@ public class ControllersAutoConfiguration {
         multipartFilter.setMultipartResolverBeanName(GrailsApplication.MULTIPART_RESOLVER_BEAN);
         registrationBean.setFilter(multipartFilter);
         registrationBean.addUrlPatterns(Settings.DEFAULT_WEB_SERVLET_PATH);
-        // Resolve multipart requests before HiddenHttpMethodFilter and security filters access parameters.
+        // Resolve multipart requests after character encoding is applied but before HiddenHttpMethodFilter
+        // and security filters access parameters.
         // Otherwise, the servlet container parses the parts first and oversized uploads bypass the multipart resolver.
-        registrationBean.setOrder(GrailsFilters.FIRST.getOrder());
+        registrationBean.setOrder(GrailsFilters.CHARACTER_ENCODING_FILTER.getOrder() + 1);
         return registrationBean;
     }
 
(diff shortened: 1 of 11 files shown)


jdaugherty (approved):

codeconsole:
I think I fixed this in one of my apps. I need to see how I solved it there.
As is, I think this is silently dropping `_method`

codeconsole (changes requested):
| Finding | Verdict | Notes |
|---|---|---|
| Multipart `_method` regression | **Valid — blocking** | `HiddenHttpMethodFilter` now ignores `_method` for every `multipart/*` request. Grails upload forms using `method="PUT"`, `PATCH`, or `DELETE` therefore remain `POST`. |
| Later request wrappers are discarded | **Valid — blocking** | `getNativeRequest(..., MultipartHttpServletRequest)` returns the inner multipart wrapper. Storing it as the current request bypasses later wrappers, including `HttpMethodRequestWrapper` and Spring Security's request wrapper. |
| `spring.servlet.multipart.enabled=false` is defeated | **Valid — blocking** | When Boot removes its multipart resolver, Spring's `MultipartFilter` falls back to its private `StandardServletMultipartResolver`, effectively re-enabling multipart processing. |
| Multipart parsing occurs before character encoding | **Valid — blocking** | Registering the multipart filter at `FIRST` causes eager parsing before `CharacterEncodingFilter`. Calling `setCharacterEncoding` afterward is too late for multipart fields. |
| Async multipart cleanup happens too early | **Valid — blocking** | `MultipartFilter` cleans multipart resources when the initial filter chain returns. For asynchronous MVC requests, temporary upload files may be deleted before the async worker consumes them. |
| `GrailsWebRequestFilter` assignment is ineffective | **Valid — non-blocking** | It is redundant when the request is directly a `MultipartHttpServletRequest` and does not handle multipart requests hidden beneath later wrappers. |
| Missing end-to-end regression coverage | **Valid — non-blocking** | The tests do not exercise an oversized upload reaching application error handling, a valid security-wrapped upload, multipart `_method`, encoding, disabled multipart support, or asynchronous file consumption. |
| Registration test is incomplete | **Valid — non-blocking** | The test compares the order with `GrailsFilters.FIRST` but does not verify relative ordering against character encoding, hidden-method, and security filters or test conditional backoff. |
| Fully qualified `WebUtils` call | **Style only** | An import alias would improve readability but has no behavioral effect. |
| Missing documentation | **Advisory** | Documentation would be useful, but this bug fix does not necessarily need to be blocked solely under the rule requiring documentation for new features. |

codeconsole:
I haven't dug deep into this yet so I am just thinking out loud, but:

Do we still need all this 10+ year old infrastructure code? Is there an option to rewire more lean and push all the multipart handling onto Spring and use as little custom code as possible?  Is there anything we can get rid of?

Can we get rid of GrailsMultipartFilter?

https://github.com/apache/grails-core/pull/16040

codeconsole:
I have an alternative for the request plumbing: discover the multipart request from the wrapper chain rather than substituting it, so the security and method-override  wrappers survive and the adapter isn't needed. It doesn't cover the oversize half — that still needs your filter.

Let me know what you think, we could merge mine first, then your filter + the `HiddenHttpMethodFilter` guard on top?
https://github.com/apache/grails-core/pull/16149

It eliminates the need for `GrailsMultipartHttpServletRequest`

codeconsole:
I really don't like the idea of circumventing the Dispatcher servlet and parsing the multipart upstream all to get the `HiddenHttpMethodFilter`   This is a lot of infrastructure code just for `HiddenHttpMethodFilter`.   

It was a legacy workaround. Should we even be using it anymore???  We should at least explore disabling it by default like Spring.

The convention (originally Rails') is: the form POSTs with a hidden _method=PUT field, and a filter wraps the request so getMethod() returns PUT. That's what <g:form method="PUT"> and `<g:uploadForm  resource="book" action="update"`  emit, and it's how Grails' scaffolded RESTful forms work.

Spring has one — and Boot turns it off

`org.springframework.web.filter.HiddenHttpMethodFilter` exists in spring-web 7.0.8. But Boot's own configuration metadata says:

```
  name:     spring.mvc.hiddenmethod.filter.enabled
  default:  False
```

Spring Boot ships it disabled. It's been off by default since Boot 2.2 — the reasoning at the time being that it's a legacy workaround, browsers aren't the dominant client any more, and anything using fetch/XHR can just send the real method.

Grails, by contrast, registers its own version enabled by default, and Grails' version is broader: it accepts any method and also honours the X-HTTP-Method-Override header, where Spring's is narrower.

codeconsole:
It seems like there is still 1 issue with oversized on this PR.   I came up with a proposal that doesn't require any extra infrastructure.  It just ignores the multipart and kicks it down the line to where it is normally handled in the Spring Dispatcher Servlet.

Unfortunately, the PR had to be stacked on top of my other performance PR so you can't see a clean diff, but if you compare the changes between the two it will be more obvious.  Feel free to take whatever you want from It, expand on it, or bring any of the ideas over here.

I think we should consider exploring default disabling the Hidden Method Filter like Spring Boot does. What are your thoughts?


# Test Results


What an application actually receives when a multipart upload exceeds the configured limit
(200 KB payload against the default 128000-byte `maxRequestSize`).

- **No fix** — `8.0.x` @ `a83f87480e`
- **#16146** — `fix/issue-16145` @ `6fe63b8486` (early `MultipartFilter`)
- **#16152** — `fix/multipart-lean-8.0.x` (tolerate the parse failure, defer to `checkMultipart`)

Every row is same-app, same-config, same probe — only the framework code differs.

| Container | Case | No fix | #16146 | #16152 |
|---|---|---|---|---|
| Tomcat 11 | Oversized, no security, `"413"` mapping | `500` raw Tomcat HTML | `500` raw Tomcat HTML | **`413`** `{"error":"Content Too Large","handledBy":"errors.tooLarge"}` |
| Tomcat 11 | Oversized, Spring Security on the chain | `413` raw Tomcat HTML | `413` raw Tomcat HTML | **`413`** application's handler |
| Tomcat 11 | Normal upload | `200` | `200` | `200` |
| Tomcat 11 | `<g:uploadForm method="PUT">` | `200`, method `PUT` | `200`, method `PUT` | `200`, method `PUT` |
| Jetty 12 | Oversized | `400` "Bad Request" — Boot's generic error, indistinguishable from a malformed request | *not measured* | **`413`** application's handler |
| Undertow 2.4 | Oversized | `413`, empty body | *not measured* | `413`, empty body — **unchanged** |

`#16146` was run on Tomcat only. Its Jetty and Undertow cells are blank because they were not measured,
not because they are known to be unchanged.

Undertow applies the limit while reading the request entity, so no filter and no servlet run. Nothing the
framework does can influence it; the limitation is documented and pinned by a spec.

codeconsole:
@matrei. I am a little concerned with the performance implications of not using the DispatcherServlet (My alternate fix also has this issue).   These have always been present (even before this PR).

`GrailsMultipartFilter` still eagerly resolves every multipart request in the filter chain, ahead of `DispatcherServlet` — for every request, regardless of whether the app needs `_method` support. That's the same shape of thing Spring Boot moved away from in 2019.
    
From the Boot issue that led to `HiddenHttpMethodFilter` being disabled by default
(https://github.com/spring-projects/spring-boot/issues/16953):
    
> The endpoints that were affected by this were accepting `MultipartFile` request
> parameter - we didn't look that deep in since disabling the filter basically fixed
> the problem and improved performance significantly but I assume this was related to
> size of the request as `HiddenHttpMethodFilter` does `ServletRequest#getParameter`.

matrei:
@codeconsole I think you are right. We should go with your PRs #16149, #16152.

codeconsole:
@matrei I've explored a few more options with this.   We should discuss.  They all have trade offs.

PR #16148, opened 2026-08-13
Add test coverage and fix compiler crashes in GORM AST transforms

borinquenkid:
## Summary
- Substantially raises test coverage across grails-datamapping-core's compile-time AST
  transform classes (the `query/transform` package plus `DirtyCheckTransformation`,
  `JpaGormEntityTransformation`, `AbstractGormASTTransformation`,
  `AbstractTraitApplyingGormASTTransformation`, `OrderedGormTransformation`,
  `TenantTransform`, `ServiceTransformation`) — several classes move from 0-70%
  to 80-100% instruction coverage.
- Fixes three real compiler crashes surfaced while writing tests: `@Tenant` and
  `OrderedGormTransformation`'s error-reporting paths threw an internal
  NullPointerException/GroovyBugError instead of a clean compile error on malformed
  input, both traced to the same root cause (an inherited `sourceUnit` field that was
  never populated).
- Cleans up static-analysis warnings on `DirtyCheckingTransformer`,
  `DirtyCheckTransformation`, and `GormEntityTransformation` (unused code, redundant
  qualifiers, `equals()` vs `==`, methods that can be static), and extracts a
  duplicated `visit()` guard shared across three transform classes into a new
  `LocalTransformationSupport` helper.
- Documents the `protectSqlInjectionAttacks` kill-switch system property in the
  security guide.

## Test plan
- [x] Full `grails-datamapping-core` test suite green throughout
      (`./gradlew :grails-datamapping-core:test`)
- [x] `codeStyle` + `codenarcMain`/`codenarcTest` clean
- [ ] CI


grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/multitenancy/transform/TenantTransform.groovy
```diff
@@ -136,7 +138,11 @@ class TenantTransform extends AbstractDatastoreMethodDecoratingTransformation {
                 return makeDelegatingClosureCall(tenantServiceVar, 'withId', args(tenantIdVar), params(param(serializableClassNode, VAR_TENANT_ID)), originalMethodCallExpr, variableScope)
             }
             else {
-                addError('@Tenant value should be a closure', annotationNode)
+                sourceUnit.getErrorCollector().addErrorAndContinue(
```
jdaugherty: This is exactly `AstUtils.error(SourceUnit, ASTNode, String)`, which already exists (`grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/AstUtils.groovy:818`) and builds the same `SyntaxErrorMessage`/`SyntaxException` pair from the node's positions. Please call it instead of hand-rolling it here:

```groovy
AstUtils.error(sourceUnit, annotationNode, '@Tenant value should be a closure')
```

That also drops the two new imports.

grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/transform/OrderedGormTransformation.groovy
```diff
@@ -60,6 +60,8 @@ class OrderedGormTransformation extends AbstractASTTransformation implements Com
             throw new RuntimeException("Internal error: wrong types: ${astNodes[0].getClass()} / ${astNodes[1].getClass()}")
         }
 
+        init(astNodes, source)
```
jdaugherty, line 59: `init(...)` re-checks the exact condition the hand-rolled guard three lines above already rejected — `AbstractASTTransformation.init` throws `GroovyBugError` when `nodes[0]` is not an `AnnotationNode` or `nodes[1]` is not an `AnnotatedNode`. Now that `init` is being called, the manual `instanceof` check and its `RuntimeException` are dead weight; move `init(astNodes, source)` to the top of the method and delete the guard.

grails-datamapping-core/src/main/groovy/org/grails/compiler/gorm/GormEntityTransformation.groovy
```diff
@@ -114,7 +114,6 @@ class GormEntityTransformation extends AbstractASTTransformation implements Comp
     private static MethodNode REMOVE_FROM_METHOD_NODE = GORM_ENTITY_CLASS_NODE.getMethods('removeFrom').get(0)
     private static MethodNode GET_ASSOCIATION_ID_METHOD_NODE = GORM_ENTITY_CLASS_NODE.getMethods('getAssociationId').get(0)
     public static final Parameter[] ADD_TO_PARAMETERS = [new Parameter(AstUtils.OBJECT_CLASS_NODE, 'obj')] as Parameter[]
-    public static final ClassNode SERIALIZABLE_CLASS_NODE = ClassHelper.make(Serializable).getPlainNodeReference()
```
jdaugherty, line 117: `SERIALIZABLE_CLASS_NODE` is `public static final` on a shipped class, so removing it is a source- and binary-compatibility break for anything outside this repo (there are no in-repo callers). Same category of change in `DirtyCheckingTransformer`: `weaveIntoExistingSetter`, `createMarkDirtyMethodCall`, `getGetterAndSetterForPropertyName` and `isAnnotatedWithJavaValidationApi` go from `protected` to `protected static` (subclasses that override them stop compiling), and `GetterAndSetter` goes from an inner class to a static nested class (its constructor signature changes).

If that is intentional for 8.1.x, fine — but it is a behaviour-neutral cleanup commit carrying an API break, so please say so in the description rather than leaving it in a "fix IntelliJ warnings" commit.

grails-datamapping-core/src/test/groovy/grails/gorm/annotation/multitenancy/TenantTransformSpec.groovy
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+new BookService()
+
+''')
+        when: 'the list books method is invoked'
+        bookService.listBooks()
+
+        then: 'an exception was thrown because GORM is not setup, proving the delegating call was generated and reached'
+        thrown(IllegalStateException)
```
jdaugherty: This assertion does not prove what the `then:` label claims. `buildDelegatingMethodCall` emits the `ServiceRegistry.targetDatastore(...).getService(TenantService)` declaration as the *first* statement of the rewritten body on every branch, so the `IllegalStateException` is raised before the closure is ever cloned or called. The spec would pass identically if the `@Tenant` closure branch were never taken — including on the non-closure error path this PR just changed.

Assert something that only the closure branch produces: inspect the rewritten `MethodNode` (you already have `compileAndFindMethod` in this spec) for the `$tenantResolver` / `$tenantId` declarations, or check the generated method's parameter list. Same for the class-level test at line 81.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/query/transform/WhereQueryAssociationPathSpec.groovy
```diff
@@ -0,0 +1,181 @@
...
+        GroovyClassLoader gcl = new GroovyClassLoader()
+
+        when:
+        gcl.parseClass(SINGLE_LEVEL_SOURCE)
+
+        then:
+        noExceptionThrown()
+
+        and: 'one closure for the outer where-block and one nested closure for the association segment walked'
+        List<Class<?>> queryClosures = findQueryClosures(gcl, 'findByAuthorName').sort { it.name.count('$_closure') }
```
jdaugherty: These assertions are pinned to Groovy's generated closure class-naming scheme — the `_<methodName>_` substring and the number of `$_closure` occurrences in the class name. That is compiler-internal naming, not output of `DetachedCriteriaTransformer`: a change to how Groovy names nested closures silently breaks these tests, and a change to the transform that stops synthesizing association closures could still leave the counts intact.

The other specs in this PR (e.g. `WhereQueryEmbeddedPropertyPathSpec`) inspect the transformed AST at canonicalization instead — please use that approach here so the assertion is about the nested `delegate.<association> { ... }` calls actually generated.

grails-datamapping-core/src/test/groovy/org/grails/compiler/gorm/LocalTransformationSupportSpec.groovy
```diff
@@ -0,0 +1,66 @@
...
+    void "returns null when the annotation type does not match"() {
+        given:
+        ClassNode targetClass = new ClassNode('com.example.Target', 0, ClassHelper.OBJECT_TYPE)
+        AnnotationNode annotationNode = new AnnotationNode(OTHER_ANNOTATION_TYPE)
+
+        expect:
+        LocalTransformationSupport.resolveAnnotatedClassOrNull([annotationNode, targetClass] as ASTNode[], ANNOTATION_TYPE) == null
+    }
+
+    void "returns null when the annotated node is not a class"() {
```
jdaugherty: Two things here.

This branch cannot occur in production: `@DirtyCheck`, `@Entity` and `@JpaEntity` are all declared `@Target([ElementType.TYPE])`, so Groovy never dispatches these local transforms on a `FieldNode`. That sits oddly beside commit `53e1767`, which removed the malformed-`astNodes` guard from this same class precisely because it was structurally unreachable — the same argument applies to the `!(parent instanceof ClassNode)` half of the check this spec exists to cover.

More generally, the whole spec drives an extracted internal helper directly, which is what CLAUDE.md rule 9 asks us to avoid; the guard is already exercised end-to-end whenever a source annotated with `@DirtyCheck` / `@Entity` / `@JpaEntity` is compiled. Also worth fixing the class javadoc: it names `DirtyCheckTransformation` and `JpaGormEntityTransformation`, but `GormEntityTransformation` uses the helper too.

jdaugherty (commented):
Coverage work here is welcome and the two compiler-crash fixes are real. A few things to tighten before this goes in.

One carry-over from the review of #16135 (which this stack sits on): the same question about overlap with existing tests applies here. `grails-datamapping-core-test/src/test/groovy/grails/gorm/tests/WhereMethodSpec.groovy` is a 2000-line functional spec over the same where-query surface, and this PR adds ten new `WhereQuery*Spec` files against the transform. Please call out in the description what these cover that `WhereMethodSpec` (and the existing `WhereQueryClosureCaptureSpec` / `WhereQueryEmbeddedBlockTransformSpec`) does not, so we are not maintaining two parallel suites for the same behaviour.

The other theme is the same one raised on #16140: several of the new specs assert against internals (the extracted helper, protected/static transform methods, generated closure class names) rather than the behaviour a user's compilation actually produces. Details inline.

jdaugherty (approved):

Issue #9926, opened 2016-05-10
RestfulController should support updates via POST in addition to PUT or allow additional configuration.


codeconsole:
AngularJS by default uses POST requests to .$save objects.  It does not use PUT requests.

If you create a User object and a scaffolded UserController, then create a resources URLMapping.  There is no way to allow allow POST updates.

A working simple Grails 3.1.6 example is here:
https://github.com/codeconsole/grails3bugs

Simply clone that github app, run, and click the Demonstrate Bug button on the index page.

Fix should be to default RestfulController to
    
    static allowedMethods = [save: "POST", update: ["POST", "PUT"], delete: "DELETE"]

or allow overriding.

According to burt [9:01 PM] 
I looked at this in a debugger and it's hard to tell because it's mostly generated code, but the controller ends up being a subclass of `RestfulController` and there's code generated by an AST that does the check for allowed methods and that's where the 405 status is being set. it looks to me like it's ignoring the `allowedMethods` map defined in the controller and using the one from the `RestfulController` base class with the standard values

jeffbrown:
> it looks to me like it's ignoring the allowedMethods map defined in the controller and using the one from the RestfulController base class with the standard values

That is correct.  At present the code that deals with `allowedMethods` restrictions is added directly the controller action methods at compile time so when `RestfulController` is compiled, the `allowedMethods` property in that class are what is used.  That could be modified to support a subclass' `allowedMethods`, but that is not what currently happens.

codeconsole:
So you would classify this as a bug, no?  Maybe at the very least we can open up POSTs to updates. I don't really see a security problem with this, do you?

codeconsole:
The more I think about it, wouldn't it be better if RestfulController didn't even have static allowedMethods?  What is the point of it having it anyhow?  Just because someone want's to expose rest services on their controller, doesn't mean it should be locked down from a security stand point.

jeffbrown:
> The more I think about it, wouldn't it be better if RestfulController didn't even have static allowedMethods? 

I don't agree.  I think `RestfulController` should have sensible restrictions in it, and you should be allowed to override those.

PR #16149, opened 2026-08-15
Reduce per-request work, and rework hidden HTTP method and multipart handling

### LINKED ISSUE #16145: [Grails 8] Oversized multipart uploads cannot be handled by Grails application code

codeconsole:
## Description

One PR covering the Grails 8 request path: per-request cost, hidden HTTP method handling, and multipart. This supersedes #16150, #16152 and #16183, which are merged in here with their history intact rather than squashed — they were interdependent enough that reviewing them separately meant tracking a merge order, which is not a reasonable thing to ask of a reviewer.

> **⚠️ This changes a default for every application on upgrade.** See *The hidden HTTP method filter is now off by default* below, and in particular the authorization consequence, which is the one change here that fails silently. If the team would rather ship the property and defer the default flip to 9.0, that is a one-line change — say so and I will make it.

### 1. Per-request work (was #16149)

Reduces allocation and repeated lookups across the request path: cached per-context collaborators, shared stateless interceptors, `UrlPathHelper.defaultInstance` instead of per-request instances, and a `GrailsApplicationAttributes` cached per servlet context rather than rebuilt per request. Benchmarks are in `grails-benchmarks`.

### 2. URL mapping names from the match (was #16150)

`UrlMappingInfo` resolves controller and action names from the match rather than re-reading request parameters.

### 3. Oversized uploads reach the application's error handling (was #16152, closes #16145)

Reading any parameter of a `multipart/form-data` request makes the container parse the body, and a body past the configured limit fails that parse inside the filter chain, where no `HandlerExceptionResolver` can see it. Worse, once part parsing has failed **every** later parameter read on that request fails too — including the reads the error handling itself performs.

Framework parameter reads now go through one tolerant helper, so the failure is left for `checkMultipart` to raise during dispatch where Spring routes it to the exception resolvers.

| Container | Before | After |
|---|---|---|
| Tomcat 11 | `500`, raw container HTML | `413`, handled by the application |
| Jetty 12 | `400`, indistinguishable from a malformed request | `413`, handled by the application |
| Undertow 2.4 | `413`, empty body | unchanged — the limit is applied before any filter or servlet runs |

### 4. `request` is always the outermost request (was #16149)

Grails used to substitute the resolved `MultipartHttpServletRequest` for the request it exposes, which discarded wrappers contributed by other filters. Multipart capability is now discovered by unwrapping, with an attribute for the case where the wrapper sits above the request Grails bound. `GrailsWebRequest.getCurrentRequest()` is deprecated and returns `getRequest()`.

### 5. The hidden HTTP method filter is now off by default (was #16183)

Grails registered a servlet filter rewriting `POST` into `PUT`/`PATCH`/`DELETE` from a `_method` parameter or an `X-HTTP-Method-Override` header, ahead of the dispatcher and ahead of the Spring Security chain, with no way to turn it off.

Reading a request parameter before the dispatcher is not free: the container resolves the dispatcher's `MultipartConfigElement` at filter time, so reading `_method` on a multipart POST parses the whole body — writing its temporary files — **before the request has been routed or authenticated**. The filter is order `-170`; the security chain is `-100`. Spring Boot disabled its equivalent in 2.2 for the same reason.

The override is not lost. It moves into the dispatcher, resolved after multipart handling and after the filter chain, and published as a request attribute so `allowedMethods` and URL mapping resolution agree on the method. Restore the old behaviour with `grails.web.hiddenmethod.filter.enabled: true`.

Two supporting changes:

- **`POST /$controller/$id` → `update`** is generated by a `resources` mapping. `RestfulController` has declared `update: ['PUT', 'POST']` since [#9926](https://github.com/apache/grails-core/issues/9926) — raised because AngularJS `$resource` and the clients modelled on it POST to the member URL to save an existing object — but no route was ever generated, so the permission has been unreachable. It is also what lets a form submit without a `_method` parameter.
- **`<g:form method="PUT">` stops emitting `_method`**, submitting a plain POST to the same URL. `method="DELETE"` still emits it, because `delete` and `update` share a URL. In the new default `_method` appears in exactly one place: a delete form.

**The consequence to review before upgrading.** Servlet filters and the Spring Security chain now see a form delete as a bare `POST /books/1`. A rule matching `DELETE /books/**` will no longer fire, and since the URL is unchanged, `update` and `delete` are not distinguishable by path either. This is the only change here that fails silently.

### 6. A startup failure new in 8.0

Boot's `WebMvcAutoConfiguration` registers its filter under the same `hiddenHttpMethodFilter` bean name, so `spring.mvc.hiddenmethod.filter.enabled=true` failed startup with a `BeanDefinitionOverrideException`. Grails' registration moved to its own auto-configuration ordered after Boot's, detecting Boot's filter rather than assuming it from a property — and contributing its own when Boot cannot, which is the case under `@EnableWebMvc`.

## Review notes

- Every commit from the superseded PRs is preserved; nothing was squashed or dropped.
- Review feedback already addressed: @jdaugherty and @codeconsole on `GrailsWebRequest`, @sbglasius on `CachedBean` remembering a lookup miss, @matrei on the filter-mode flag being derived from only one of the two properties.

## Documentation

`grails-doc` upgrade guide sections 52–54, and the REST guide's *Linking to Resources* page.


matrei:
Just as a FYI, this is breaking Grails 7 plugin interceptors:

`... does not define or inherit an implementation of the resolved method 'abstract java.util.concurrent.atomic.AtomicReference grails_artefact_controller_support_ResponseRedirector__responseRedirector$get()' of interface grails.artefact.controller.support.ResponseRedirector$Trait$FieldHelper`

At this point, I'm not sure how much that matters, but I wanted to call it out.

codeconsole:
@matrei agreed, wasn't benefiting the overall design.  I think better without it.  I revised the trait. Thanks

grails-web-common/src/main/groovy/org/grails/web/servlet/mvc/GrailsWebRequest.java
```diff
@@ -214,14 +260,14 @@ public FlashScope getFlashScope() {
 
     /**
      * @return The currently executing request
+     *
+     * @deprecated as of 8.0, use {@link #getRequest()} instead. This used to return the resolved
+     *             multipart request in place of the request Grails was bound to; that substitution is
+     *             gone, so the two are now the same object.
      */
+    @Deprecated(since = "8.0")
```
jdaugherty, line 286: Why are we renaming & deprecating this?

jdaugherty:
@codeconsole  there are a lot of PRs related to the multipart problems, is this PR still valid?  I'm trying to understand what should be reviewed and what shouldn't.

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
```diff
@@ -774,4 +800,65 @@ private static Object unwrapGString(Object value) {
...
+            this.beanType = beanType;
+        }
+
+        private T get(ApplicationContext applicationContext) {
+            if (!resolved) {
+                bean = applicationContext.containsBean(beanName) ? applicationContext.getBean(beanName, beanType) : null;
+                resolved = true;
+            }
+            return bean;
+        }
```
sbglasius: If a `bean` isn't resolved yet, it never will be since a miss sets `bean = null` and `resolved = true` together, so `!resolved` is `false` on every later call regardless of whether the bean shows up afterward.

```
            if (bean == null) {
                bean = applicationContext.containsBean(beanName) ? applicationContext.getBean(beanName, beanType) : null;
            }
            return bean;
        }
```

and drop the resolved field entirely (no replacement needed since nothing else reads it). `bean` is already volatile, so keying the check on `bean == null` keeps the same safe-publication guarantee the class doc calls out, while letting a bean registered after the first miss be picked up on a later call

My answer is AI assisted, aka it found the potential bug, I described it.

sbglasius (dismissed):

grails-web-common/src/main/groovy/org/grails/web/servlet/mvc/GrailsWebRequest.java
codeconsole: Because there is no point of having 2 methods that do the same thing.  We can't get rid of getRequest() that is inherited and without the multipart check (which is no longer needed) getCurrentRequest() is the same as getRequest()

before
```groovy
    public HttpServletRequest getCurrentRequest() {
        if (multipartRequest != null) {
            return multipartRequest;
        }
        else {
            return getRequest();
        }
    }
```

after this PR
```groovy
    public HttpServletRequest getCurrentRequest() {
            return getRequest();
    }
```

what is the point in keeping it?

grails-web-databinding/src/main/groovy/grails/web/databinding/DataBindingUtils.java
codeconsole: Good catch — fixed in cfc47a55fe.

Dropped `resolved` and keyed on the bean, so one registered after a miss is picked up. Only writing the field when a bean is found avoids a redundant volatile write and keeps Checkstyle happy about the ternary.

Hot path is unchanged: one volatile read instead of two once resolved.

Added a regression test and confirmed it fails against the old implementation.

grails-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRedirector.groovy
```diff
@@ -122,7 +122,7 @@ trait ResponseRedirector implements WebAttributes {
             throw new IllegalArgumentException("Invalid arguments for method 'redirect': $argMap")
         }
 
-        grails.web.mapping.ResponseRedirector redirector = new grails.web.mapping.ResponseRedirector(grailsLinkGenerator)
+        grails.web.mapping.ResponseRedirector redirector = new grails.web.mapping.ResponseRedirector(getGrailsLinkGenerator())
```
sbglasius: not something to fix, but what is wrong with Groovy property accessors (`grailsLinkGenerator` vs `getGrailsLinkGenerator()`) 

It is changed here, but not in `ResponseRenderer`

sbglasius (approved):

matrei:
# AI Review

Compared the checked-out branch at `a192c75725` with `origin/8.0.x` at `4309d5b876`.

## Findings

### 1. High: Dispatcher-level hidden-method overrides are not propagated to `GrailsWebRequest`

References:

- `grails-web-mvc/src/main/groovy/org/grails/web/servlet/mvc/GrailsDispatcherServlet.groovy:115-122`
- `grails-web-common/src/main/groovy/org/grails/web/servlet/mvc/GrailsWebRequest.java:269-270`
- `grails-web-common/src/main/groovy/grails/web/api/ServletAttributes.groovy:47-49`

`checkMultipart()` returns a wrapper reporting `PUT`, `PATCH`, or `DELETE`, but the `GrailsWebRequest` was already bound to the original request and continues returning it. The override is only exposed through an attribute consumed by `HiddenHttpMethod.effectiveMethod()`.

As a result:

- URL mapping resolution can select `PUT`, `PATCH`, or `DELETE`.
- `allowedMethods` sees the overridden method.
- `controller.request.method` still reports `POST`.
- `ServletRenderContext.getHttpMethod()` still reports `POST`.
- Command-object initialization still branches on `POST`.
- Method-keyed mapping names still use `POST`.

This contradicts the implementation comments and documentation claiming that the Grails request API agrees with the selected route. The existing test already expects `request.method == 'PUT'` in `grails-test-examples/app1/src/integration-test/groovy/functionaltests/fileupload/FileUploadSpec.groovy:334-337`, but the current lifecycle supplies the original request to the controller.

This should block approval until the request method authority is unified, or the behavior is explicitly redesigned and all affected APIs are updated.

### 2. High: Multipart form fields can be lost from `params`

Reference:

- `grails-web-common/src/main/groovy/grails/web/servlet/mvc/GrailsParameterMap.java:99-117`

The new code reads ordinary parameters from the outer request, then discovers the inner `MultipartHttpServletRequest` but only merges its uploaded files. It does not merge the multipart wrapper's parameter map.

Spring's `StandardMultipartHttpServletRequest` explicitly documents that servlet containers are not guaranteed to expose multipart text parts through the native request's `getParameterMap()`. Its own `getParameterMap()` merges those parts. Therefore, with a container exhibiting that behavior, fields such as `description`, `category`, or `_method` disappear from Grails `params`.

The existing `uploadWithMetadata` integration test expects multipart text fields to be present, but coverage uses only the normal container path and does not verify the outer-wrapper/native-parameter-map combination.

`GrailsParameterMap` should merge parameters from the resolved multipart request as well as files, while retaining the tolerant behavior for an unparseable oversized request.

### 3. High: The URL-mapping fallback resolves hidden methods during forwards and includes

References:

- `grails-web-mvc/src/main/groovy/org/grails/web/servlet/mvc/GrailsDispatcherServlet.groovy:93,115`
- `grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/UrlMappingsHandlerMapping.groovy:89-94,178`

The dispatcher deliberately disables hidden-method resolution for forwards/includes. However, `UrlMappingsHandlerMapping.resolveHttpMethod()` still reads `_method` for every mapping lookup whenever `resolveHiddenHttpMethod` is enabled.

A forwarded or included `POST` carrying `_method=DELETE` can therefore be matched as `DELETE`, even though the dispatcher did not apply the override. This can produce inconsistent routing and `allowedMethods` behavior, and may let inherited request parameters influence an internal dispatch's action selection.

The handler mapping should apply the same dispatch-type guard as the dispatcher, or the dispatcher and mapping behavior should be made intentionally consistent.

### 4. High: Generic PUT/PATCH forms no longer preserve their requested method

Reference:

- `grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy:519-525`

With the filter disabled, `_method` is suppressed for every PUT/PATCH form, not just forms targeting a resource member URL. This changes:

```groovy
<g:form url="/admin/update" method="PUT">
```

from a POST containing `_method=PUT` into a plain POST. Existing custom PUT/PATCH-only mappings consequently stop matching.

The newly generated POST member route only compensates for `resources` mappings. It cannot preserve arbitrary application mappings. The form tag either needs to retain `_method` for non-resource forms or determine that the target is covered by the new resource fallback.

### 5. High: Bare member POST introduces a new mutation route with potentially different authorization semantics

References:

- `grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingEvaluator.java:147-155,960-967`
- `grails-rest-transforms/src/main/groovy/grails/rest/RestfulController.groovy:44`

The branch adds:

```text
POST /books/$id -> update
```

for every `resources` mapping when filter mode is disabled. A bare POST to a member URL can now invoke `update`, whereas it did not previously.

The PR rationale says this does not widen authorization because security sees both requests as POST. However, applications commonly distinguish collection POST and member POST, and may have authorization rules that permit collection POST while unintentionally allowing member POST. This is a new write route and should be treated as a security-sensitive behavior change.

The upgrade documentation currently understates this consequence.

### 6. Medium: `SpringSecurityUtils.isAjax()` still performs an unsafe multipart parameter read

Reference:

- `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/SpringSecurityUtils.groovy:303-310`

The method directly calls:

```groovy
request.getParameter('ajax')
```

before the tolerant multipart handling. For an oversized multipart request, this can fail inside the Spring Security filter chain, before `DispatcherServlet.checkMultipart()` can route the multipart exception through Grails error handling.

The direct parameter read should use `WebUtils.readParameter()`, and Spring Security integration coverage should include oversized multipart requests handled by authentication and access-denied flows.

### 7. Medium: `HttpMethodOverrideDetector` remains inconsistent with the tolerant parameter path

Reference:

- `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/filter/HttpMethodOverrideDetector.groovy:46-52`

This Spring Security extension point still calls `request.getParameter()` directly. If configured ahead of dispatcher multipart handling, an oversized multipart request can fail in the filter chain rather than reaching the intended application error pipeline.

No default production call site was found, so this is primarily an extension compatibility risk, but it should either use the tolerant helper or be documented as intentionally outside the new behavior.

### 8. Medium: Interceptor post-processing now mutates request state in place

Reference:

- `grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/GrailsInterceptorHandlerInterceptorAdapter.groovy:118-120`

The previous implementation created a reversed copy of the matched-interceptor list. The new implementation reverses the request attribute in place.

A second `postHandle()` call now reverses the list back, changing callback order between invocations. The existing spec already invokes `postHandle()` twice, but only checks the model/view result, not the callback order. Nested dispatches, forwards/includes, async redispatches, or custom lifecycle invocations can expose this difference.

The list should be copied before reversal, or the lifecycle should explicitly guarantee that `postHandle()` is called only once.

## Additional review notes

- The branch is clean and is 53 commits ahead of `origin/8.0.x` at the refreshed head `a192c75725`.
- The GORM/native-ID changes visible in the branch are already present in `origin/8.0.x`; they are not additional PR changes.
- The captured CI state was not fully green. CodeQL reported 11 new high-severity alerts, and the captured merge state was unstable.
- Patch coverage was approximately 69.7%, with 134 changed lines not covered.
- The most important missing coverage is a full request-path test combining filter ordering, multipart resolution, dispatcher method wrapping, URL mapping, `GrailsWebRequest`, controller request access, `allowedMethods`, command-object binding, REST rendering, and Spring Security.

codeconsole:
Thanks — worked through all eight.

**1. `request.method` reports POST — intended; the docs were wrong.** The override is resolved after the filter chain, so POST is the method on the wire, the one Spring Security saw, and the one the access log records. Routing and `allowedMethods` use the effective method. The `FileUploadSpec` expectation you cite was already updated in 423f2bd. The dispatcher comment and upgrade note still claimed the request agreed — corrected in a2ee534.

**2. Multipart text fields — valid, fixed.** `GrailsParameterMap` now merges the multipart wrapper's parameter map as well as its files, without displacing the request's own values. Test covers the outer-request/native-map shape. a2ee534

**3. Forward/include — valid, fixed.** `resolveHttpMethod` applies the dispatcher's guard. Error dispatches never reach it (matched by status code first), so it doesn't mention them. Tests for both. a2ee534

**4. Generic PUT/PATCH forms — valid, fixed.** `<g:form>` emits `_method` for any non-GET/non-POST again. Two cases beyond the one you listed were also broken: PATCH resource forms (the POST route reaches `update`, but PATCH maps to the separate `patch` action) and singular `resource` mappings (no POST route is generated). 7689098

**5. Bare member POST — valid as a docs gap, fixed.** The route is intentional (#9926), but the note described it as a fix without saying it is a route that did not exist. It now says so, and to review rules granting POST by prefix. a2ee534

**6. `isAjax()` — valid, fixed.** Tolerant read. Pre-existing upstream, but this PR changes when it is first hit. a2ee534

**7. `HttpMethodOverrideDetector` — not valid.** No call site, as you note. Dead code; removing it is its own change.

**8. Interceptor reversal — not valid.** Upstream reversed into a copy and wrote that copy back to the attribute, so a second `postHandle` re-reverses either way. I ran the same order-asserting test against both implementations — identical. Nothing else holds the list: the attribute is private to the adapter, and the only other reader is a benchmark. You were right that the spec asserted nothing about order, so it does now: d3bfaa7.

On the Grails 7 plugin interceptor error — tracked it down, and it is not the head of this branch. abc0316b0e adds a `private final AtomicReference responseRedirector` field to the `ResponseRedirector` **trait**, and 785d4e5f41 takes it out again; both are in this PR. A field on a Groovy trait generates `ResponseRedirector$Trait$FieldHelper` carrying exactly the `...__responseRedirector$get()` method your trace names, so anything compiled while that field existed expects it at runtime. The head is clean — no changed trait gains a field relative to 8.0.x — so a rebuild against the current head should clear it. The intermediate commit is the only binary-incompatible point in the range; say the word if you would rather those two were squashed so no commit in the history has it.

matrei (commented):
AI review:

# Review Findings

The previous findings are mostly addressed, but I found the following remaining issues on the current head `0e0d5993be`.

## [P1] Apply the effective method to command-object and request-body binding

**Files:**

- `grails-controllers/src/main/groovy/grails/artefact/Controller.groovy:415-449`
- `grails-web-databinding/src/main/groovy/org/grails/web/databinding/bindingsource/AbstractRequestBodyDataBindingSourceCreator.groovy:46-52`

Dispatcher-mode hidden method handling records `_method` as an attribute while leaving `request.getMethod()` as the wire method (`POST`). Routing, `allowedMethods`, reverse mapping, and REST rendering now use `HiddenHttpMethod.effectiveMethod()`, but command-object initialization and request-body binding still use the raw method.

This means a `POST` carrying `_method=DELETE` can still:

- instantiate and bind a domain command object as a POST;
- bind a request body even though DELETE is an ignored request-body method;
- take the POST construction path for a domain command object without an identifier.

Please either use `HiddenHttpMethod.effectiveMethod(request)` in these binding paths or explicitly document and test that command/body binding intentionally follows the wire method. The current tests cover ordinary POST/PUT/DELETE requests, but not dispatcher-mode requests with an `_method` attribute.

## [P1] Clear the dispatcher override before internal forwards/includes

**Files:**

- `grails-web-mvc/src/main/groovy/org/grails/web/servlet/mvc/GrailsDispatcherServlet.groovy:107-115`
- `grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/UrlMappingsHandlerMapping.groovy:81-90`
- `grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/AbstractUrlMappingInfo.java:151-154`
- `grails-controllers/src/main/groovy/grails/artefact/controller/support/RequestForwarder.groovy:120-141`

The dispatcher publishes `HiddenHttpMethod.OVERRIDDEN_METHOD_ATTRIBUTE` on the original request. The handler mapping now correctly ignores `_method` while resolving a forward or include, but the published attribute remains on the request.

As a result, an initial `POST` with `_method=PUT` can forward to another URL and produce two different method decisions during the internal dispatch:

1. `UrlMappingsHandlerMapping.resolveHttpMethod()` matches the internal dispatch as `POST`.
2. `AbstractUrlMappingInfo` then uses `HiddenHttpMethod.effectiveMethod()` and sees the stale `PUT` attribute when resolving a method-keyed action/controller name.

The same stale attribute is visible to other effective-method readers during the internal dispatch, including `ServletRenderContext` and `DefaultRequestStateLookupStrategy`.

Please scope the override attribute to the original dispatch and restore/remove it around forwards and includes. Add a method-keyed mapping test that performs a real internal dispatch after an overridden request.

## [P2] The resolved-multipart branch in `SpringSecurityUtils.isAjax()` still reads parameters directly

**File:** `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/SpringSecurityUtils.groovy:311-315`

The initial `request.getParameter('ajax')` read now correctly uses `WebUtils.readParameter()`, but after discovering a resolved multipart request the code still calls:

```groovy
multipart?.getParameter('ajax')
```

If that multipart wrapper performs lazy parsing, or its parameter parsing fails, this read can still throw inside the Spring Security filter chain before the dispatcher can route the multipart failure through Grails error handling.

Please use the same tolerant helper for the resolved multipart request, or otherwise ensure this branch cannot reintroduce the early multipart failure. Add coverage with a multipart request whose resolved wrapper throws from `getParameter()`.

## Addressed Previous Findings

- Multipart text parameters are now merged from the resolved multipart wrapper in `GrailsParameterMap`.
- Hidden-method parameters are ignored during forward/include mapping resolution.
- Generic PUT/PATCH forms emit `_method` again.
- The new bare member POST route is documented as a new route with an authorization warning.
- The initial `SpringSecurityUtils.isAjax()` parameter read is tolerant.
- Repeated interceptor callback ordering is covered and matches the previous behavior.

## Verification

The following focused suites passed:

- `:grails-web-common:test`
- `:grails-web-mvc:test`
- `:grails-web-url-mappings:test`
- `:grails-interceptors:test`
- `:grails-controllers:test`
- `:grails-web-databinding:test`
- `:grails-web-core:test`

The latest captured PR CI reports 81,909 tests passing and CodeQL reports no new alerts. Codecov still reports approximately 71.2% patch coverage, with 127 changed lines uncovered.

codeconsole:
All three valid. Fixed in 2234fe9 — though one of them I fixed the other way round, and it is worth saying why.

**Command object and request body binding — fixed.** Both read `effectiveMethod` now. The body case was the real one: a `POST` naming `DELETE` was having its body bound where the filter would have skipped it, since `ignoredRequestBodyMethods` is keyed on the method. Covered by two cases in `AbstractRequestBodyDataBindingSourceCreatorSpec`. The command object branch only diverges for a domain command object reached without an identifier, which needs a GORM-backed controller to exercise, so it rests on the rule and the upgrade table rather than a test of its own.

**Stale override on a forward — real, but clearing it is the wrong repair.** Under the servlet filter the wrapper reports the overridden method for the whole request, forwards included. Clearing the attribute would make dispatcher mode disagree with filter mode, which is the one thing moving the override was meant not to do. The inconsistency was my forward/include guard from the last round being too broad: it discarded an override the dispatcher had already published, so routing said `POST` while the method-keyed reader, `RenderContext` and link generation all still said `PUT`. The mapping now honours a published override wherever it applies, and refuses only to *derive* a fresh one from inherited parameters on an internal dispatch — which is the case your earlier finding was actually about. Tests cover a forward and an include carrying a resolved override, asserting the route and the other readers agree.

**`isAjax()` multipart read — fixed.** Tolerant on the wrapper too, which is the read likelier to force the failing parse.

On coverage: agreed there is no single end-to-end request-path test, and that remains the honest gap.

codeconsole:
The end-to-end test now exists: 9763091 adds `RequestPathSpec` in app1, driving a real server through the servlet filter chain.

One round trip asserts the whole chain agrees: a multipart form POST naming PUT is resolved by the dispatcher after multipart parsing, the mapping picks the PUT route, `allowedMethods` admits it by the effective method, the multipart text field and the uploaded file both reach `params`, a command object binds from them, the response renders — and `request.method` still reports the POST it arrived as.

Five more cases pin the surrounding behaviour: a bare POST to the member URL reaches `update` (#9926, and only routable while the filter is off); a form POST naming DELETE reaches `delete`; one naming PATCH reaches `patch` rather than `update`, which is what the suppressed `_method` used to break; a POST naming GET stays a POST; a real DELETE routes as itself.

Spring Security is the one dimension on your list this cannot cover — app1 does not have it on the classpath.

grails-controllers/src/main/groovy/grails/artefact/controller/support/ResponseRedirector.groovy
codeconsole: Nothing wrong with them - that was leftover. The explicit call arrived with a trait field cache this branch added and then removed again, and the removal left the call site behind. There is no field named `grailsLinkGenerator`, so the property resolves to the same getter either way.

Put back in 250034dc, which leaves the trait matching 8.0.x apart from the `getCurrentRequest` change. The other call site, at `resolveNamespace`, is explicit on 8.0.x already, so I left it alone.

grails-test-examples/benchmarks/README.adoc
jdaugherty, line 1: I know you did not add this module, but it looks like it was added as a test style project and yet it didn't use our naming convention.  This needs to either exist under grails-test-examples or it needs to be excluded under the root build file (see testProjectsStartWith in the root build file).  Since this PR is significantly expanding, can you please do either of these?

jdaugherty:
@codeconsole why did you rename currentRequest -> request throughout the code?  What's the reasoning for this change?

codeconsole:
> @codeconsole why did you rename currentRequest -> request throughout the code? What's the reasoning for this change?

@jdaugherty `getCurrentRequest()` does exactly 1 thing now.  It just returns `getRequest()`.  Why unnecessarily proxy a method?

jdaugherty:
> > @codeconsole why did you rename currentRequest -> request throughout the code? What's the reasoning for this change?
> 
> @jdaugherty `getCurrentRequest()` does exactly 1 thing now. It just returns `getRequest()`. Why unnecessarily proxy a method?

Because it's been there since 2009 and has a distinguished name.  It was clearly used throughout the code base too and resulted in a significant diff to remove it.

jdaugherty:
> > > @codeconsole why did you rename currentRequest -> request throughout the code? What's the reasoning for this change?
> > 
> > 
> > @jdaugherty `getCurrentRequest()` does exactly 1 thing now. It just returns `getRequest()`. Why unnecessarily proxy a method?
> 
> Because it's been there since 2009 and has a distinguished name. It was clearly used throughout the code base too and resulted in a significant diff to remove it.

Actually, 2009 is just when 1.1 was merged.  It's been there since the start of Grails.

codeconsole:
> > > > @codeconsole why did you rename currentRequest -> request throughout the code? What's the reasoning for this change?
> > > 
> > > 
> > > @jdaugherty `getCurrentRequest()` does exactly 1 thing now. It just returns `getRequest()`. Why unnecessarily proxy a method?
> > 
> > 
> > Because it's been there since 2009 and has a distinguished name. It was clearly used throughout the code base too and resulted in a significant diff to remove it.
> 
> Actually, 2009 is just when 1.1 was merged. It's been there since the start of Grails.

@jdaugherty 
but it was only there for handling multipart requests and it is no longer needed after this PR.  We don't need a bunch of legacy calls all over the place that do nothing?  The method still exists. If you really want to use it, you will get a deprecated warning.

### OLD 2009 IRRELEVANT CODE:
```groovy
public HttpServletRequest getCurrentRequest() {
    if (multipartRequest != null) {
      return multipartRequest;
    }
    else {
        return getRequest();
    }
}
```

replaced with

``` groovy
@Deprecated(since = "8.0")
public HttpServletRequest getCurrentRequest() {
      return getRequest();
}
```

grails-test-examples/benchmarks/README.adoc
codeconsole: Taken the second option in 93c8ffd - `grails-benchmarks` is now named in `testProjectsStartWith`.

The two places that key off those prefixes are the BOM enumeration and the functional-test coordinate substitution, and neither has anything to take from it: the module deliberately omits the publish plugin, since its JMH dependency is Category X, so the BOM was already skipping it on the published-project check further down. Naming it says that on purpose instead of leaving it to that second check.

Renaming it under `grails-test-examples` looked like the worse of the two: the directory is referenced from the JMH workflow, the comparison tooling and the README, and that tree is for applications exercising Grails as an application rather than for a build-time harness. Happy to move it instead if you would rather the convention hold without exception.

jdaugherty:
> > > > > @codeconsole why did you rename currentRequest -> request throughout the code? What's the reasoning for this change?
> > > > 
> > > > 
> > > > @jdaugherty `getCurrentRequest()` does exactly 1 thing now. It just returns `getRequest()`. Why unnecessarily proxy a method?
> > > 
> > > 
> > > Because it's been there since 2009 and has a distinguished name. It was clearly used throughout the code base too and resulted in a significant diff to remove it.
> > 
> > 
> > Actually, 2009 is just when 1.1 was merged. It's been there since the start of Grails.
> 
> @jdaugherty but it was only there for handling multipart requests and it is no longer needed after this PR. We don't need a bunch of legacy calls all over the place that do nothing? The method still exists. If you really want to use it, you will get a deprecated warning.
> ### OLD 2009 IRRELEVANT CODE:
> 
> ```groovy
> public HttpServletRequest getCurrentRequest() {
>     if (multipartRequest != null) {
>       return multipartRequest;
>     }
>     else {
>         return getRequest();
>     }
> }
> ```
> 
> replaced with
> 
> ```groovy
> @Deprecated(since = "8.0")
> public HttpServletRequest getCurrentRequest() {
>       return getRequest();
> }
> ```

You're viewing it legacy because of an implementation detail, while I'm saying it's been a part of the public api since Grails inception and I'm objecting to that removal.

codeconsole:
> You're viewing it legacy because of an implementation detail, while I'm saying it's been a part of the public api since Grails inception and I'm objecting to that removal.

@jdaugherty it's not removed, it's deprecated. The implementation detail is it no longer has an implementation.  It is literally dead code. If it was only created for multi-part requests, how long do you want to keep it for?  Do you have another use case you want to morph it into in the future?

grails-test-examples/benchmarks/README.adoc
codeconsole: Correcting myself - I took the exclusion first and it was the wrong call. Moved instead, in 14ae491: `grails-test-examples-benchmarks`, under `grails-test-examples/benchmarks`, and the `testProjectsStartWith` entry reverted with it.

You were right that it is a rogue module. It was declared in the framework `include(...)` block between `grails-bootstrap` and `grails-beans-dsl` while being unpublishable, so three separate mechanisms that read the prefix had to be told about it individually - and its own build file already listed four plugins it omits for the same reason. My exclusion would have made five. The prefix now carries the classification instead.

Cost was five files: settings, root build, the workflow's task paths, the RAT exclusion for the golden report fixtures, and the module README. One thing to know: the workflow probes the base revision for the project directory before comparing against it, so a base from before this commit falls back to head-only mode, which it already handles and reports.

`compileJmhJava`, the module's tests, `rat`, and the BOM check all pass.

jdaugherty:
> > You're viewing it legacy because of an implementation detail, while I'm saying it's been a part of the public api since Grails inception and I'm objecting to that removal.
> 
> @jdaugherty it's not removed, it's deprecated. The implementation detail is it no longer has an implementation. It is literally dead code. If it was only created for multi-part requests, how long do you want to keep it for? Do you have another use case you want to morph it into in the future?

@codeconsole We've been removing deprecations every release.  I do not agree with it's deprecation or removal.  I'm objecting to changing a public API that's been there since Grails started.  I'm asking to revert the caller renames to simplify this diff since they are unrelated to the issue being fixed.

codeconsole:
> @codeconsole We've been removing deprecations every release. I do not agree with it's deprecation or removal. I'm objecting to changing a public API that's been there since Grails started. I'm asking to revert the caller renames to simplify this diff since they are unrelated to the issue being fixed.

I am trying to understand why don't you agree with its deprecation or removal.  Give me a use case. Please help me understand why you want something other than "just because it has always been there" 

If the only reason of it being added in 2009 was for multi-part and this PR addresses request performance and eliminates the multi-part path, how is it unrelated?   Is it performant to call a proxy method that does nothing?

Please just give me a single use case how keeping it valuable or settle on a compromise of keeping the rename and removing the `@deprecated`. I am fine with that. We can keep the method for another 10 years, but I don't see the reason why we can't have the other callers go direct instead of proxying for something they no longer need?

Is that a good compromise?  remove `@deprecated` but allow rest of code to call direct?

grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/ReflectionUtils.groovy
```diff
@@ -206,7 +206,7 @@ class ReflectionUtils {
 
     static UrlMappingInfo[] matchAllUrlMappings(UrlMappingsHolder urlMappingsHolder, String requestUrl,
                                                 GrailsWebRequest grailsRequest, HttpServletResponseExtension extension) {
-        String method = grailsRequest.currentRequest.method
+        String method = grailsRequest.request.method
```
jdaugherty: **Effective-method split between security and dispatch (privilege-escalation surface).** With the filter off (the new default), the security chain still matches URL mappings against `grailsRequest.request.method` (a plain `POST`), while `GrailsDispatcherServlet` resolves the `_method` override and `UrlMappingsHandlerMapping`/`AllowedMethodsHelper` route and authorize on `HiddenHttpMethod.effectiveMethod(...)`. Combined with the new `POST /$controller/$id -> update` route, a browser form `POST /book/1` with `_method=DELETE` is authorized by Spring Security as the `update` mapping, then executed by the dispatcher as `delete`. If `update` carries a weaker `@Secured`/rule than `delete`, that is an authorization bypass.

The PR upgrade notes acknowledge that `update` and `delete` are no longer distinguishable by path; flagging the concrete escalation path so reviewers weigh it explicitly. Consider publishing the resolved method to the request before the security chain runs (or documenting a required security-rule change) so the authorized method and the executed method agree.  Otherwise, this is a CVE that can't be merged.

grails-web-common/src/main/groovy/org/grails/web/util/WebUtils.java
```diff
@@ -552,4 +564,95 @@ public static boolean isForwardOrInclude(HttpServletRequest request) {
...
+     * @param request The request
+     * @param read The read to perform
+     * @param fallback The value to use when the parameters are unreadable
+     * @return The read value, or {@code fallback} when the parameters are unreadable
+     */
+    private static <T> T readTolerantly(HttpServletRequest request, Supplier<T> read, T fallback) {
+        try {
+            return read.get();
+        }
+        catch (RuntimeException e) {
```
jdaugherty, line 650: **`readTolerantly` swallows *every* `RuntimeException` when the Content-Type is multipart, not only container part-parse failures.** The only discriminator is `isMultipartContentType(request)`, so an exception raised for an unrelated reason on a multipart request (a filter-contributed wrapper vetoing `getParameterMap()`, an already-consumed/aborted body, etc.) is downgraded to a DEBUG log and an empty result. If `checkMultipart` then succeeds (e.g. lazy resolution with the body within limits, or a different failure mode), the controller runs with silently-empty params and form fields bind as `null` instead of the request failing. Consider narrowing the catch to the multipart exception types Spring actually raises for a rejected body (e.g. `MultipartException` / `MaxUploadSizeExceededException`) rather than all `RuntimeException`.

grails-controllers/src/main/groovy/org/grails/compiler/web/ControllerActionTransformer.java
```diff
@@ -640,14 +665,23 @@ protected void wrapMethodBodyWithExceptionHandling(final ClassNode controllerCla
...
-        tryBlock.addStatement(codeToHandleAllowedMethods);
+        if (!codeToHandleAllowedMethods.isEmpty()) {
+            tryBlock.addStatement(codeToHandleAllowedMethods);
+        }
         tryBlock.addStatement(methodBody);
 
         final TryCatchStatement tryCatchStatement = new TryCatchStatement(tryBlock, new EmptyStatement());
         tryCatchStatement.addCatch(catchStatement);
 
+        if (codeToHandleAllowedMethods.isEmpty()) {
```
jdaugherty: **Behavior change: controllers that declare no (or an empty) `allowedMethods` map no longer write the `ALLOWED_METHODS_HANDLED` request attribute.** Previously every action generated the `if (attr not set) { ...; set attr }` wrapper unconditionally, so a restricted action invoked programmatically or via forward from an *unrestricted* controller had its method check suppressed (the attribute was already set). With this early-return the attribute is never set for such controllers, so a forward/chain from an unrestricted controller into a method-restricted action now runs that action's `allowedMethods` check against the original request method and can produce an unexpected `405` mid-request. Worth a test for cross-controller invocation from an `allowedMethods`-free controller.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingEvaluator.java
```diff
@@ -135,6 +136,21 @@ public class DefaultUrlMappingEvaluator implements UrlMappingEvaluator, ClassLoa
...
+     * RestfulController has permitted POST for update since #9926 — raised because AngularJS $resource, and
+     * the clients modelled on it, POST to the member URL to save an existing object rather than sending a
+     * PUT — but no mapping was ever generated for it, leaving that permission unreachable.
+     *
+     * Generated only while the hidden HTTP method filter is disabled. In that mode the filter chain already
+     * sees a form's PUT as a bare POST to this URL, so the route adds no request shape security had been
+     * able to distinguish; it does add a member URL that answers POST, which the upgrade notes call out.
+     */
+    private boolean isPostUpdateVariantEnabled() {
+        return grailsApplication != null && !HiddenHttpMethod.isServletFilterMode(grailsApplication.getConfig());
```
jdaugherty, line 157: **Filter mode is derived only from the two config properties, not from whether a hidden-method filter bean is actually registered.** The old registration used `@ConditionalOnMissingBean(HiddenHttpMethodFilter.class)`, so an application supplying its own `HiddenHttpMethodFilter` bean was a supported extension point. `isServletFilterMode()` now checks only `grails.web.hiddenmethod.filter.enabled` / `spring.mvc.hiddenmethod.filter.enabled`, so such an app is treated as *dispatcher* mode: `isPostUpdateVariantEnabled()` returns true and every `resources` mapping silently gains a `POST /$controller/$id -> update` route the app never contemplated (and `resolveHiddenHttpMethod` is enabled alongside its filter). Consider keying the mode on the actual presence of a filter bean, or documenting that a self-registered filter must also set the property.

grails-web-common/src/main/groovy/org/grails/web/servlet/mvc/GrailsWebRequest.java
```diff
@@ -118,13 +162,15 @@ public GrailsWebRequest(HttpServletRequest request, HttpServletResponse response
...
+     * resolution happens after params were already read.
+     * See <a href="https://github.com/apache/grails-core/issues/13837">gh-13837</a>.
      *
-     * @param multipartRequest The multipart request
+     * @since 8.0
      */
-    public void setMultipartRequest(HttpServletRequest multipartRequest) {
-        this.multipartRequest = multipartRequest;
-        this.originalParams = null; // originalParams will need to be re-initialized. See https://github.com/apache/grails-core/issues/13837
+    public void multipartRequestResolved() {
```
jdaugherty, line 172: **Public `setMultipartRequest(HttpServletRequest)` was removed outright** — unlike `getCurrentRequest()` it is neither `@Deprecated` nor mentioned in the upgrade guide, and its replacement `multipartRequestResolved()` has different semantics (it takes no request and only nulls the cached params). A plugin or test harness that previously installed a resolved multipart request via `webRequest.setMultipartRequest(resolved)` now fails to compile/link with no drop-in replacement, and even after switching to `multipartRequestResolved()` must additionally set `WebUtils.MULTIPART_HTTP_SERVLET_REQUEST_ATTRIBUTE` for the request to be discoverable. Consider keeping a deprecated shim or calling this out in the upgrade notes alongside `getCurrentRequest()`.

This is a breaking API change with no prior release deprecating the method and needs fixed.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/UrlMappingsHandlerMapping.groovy
```diff
@@ -59,11 +59,46 @@ class UrlMappingsHandlerMapping extends AbstractHandlerMapping {
...
+     */
+    protected String resolveHttpMethod(HttpServletRequest request) {
+        if (!resolveHiddenHttpMethod) {
+            return request.getMethod()
+        }
+        String resolved = HiddenHttpMethod.effectiveMethod(request)
+        if (resolved != request.getMethod() || WebUtils.isForwardOrInclude(request)) {
+            return resolved
+        }
+        HiddenHttpMethod.resolveOverride(request) ?: resolved
```
jdaugherty: **Fallback override is used for routing but never published as `HiddenHttpMethod.OVERRIDDEN_METHOD_ATTRIBUTE`.** When the dispatcher did not resolve the override (e.g. a stock `DispatcherServlet` replaces `GrailsDispatcherServlet`, or the handler mapping is driven standalone), this branch derives `resolveOverride(request)` and matches mappings against it, but `effectiveMethod(request)` / `AllowedMethodsHelper` still see the raw `POST`. The request is then routed to the `update`/`delete` mapping yet fails its generated `allowedMethods` check with `405`. If this fallback is meant to fully stand in for the dispatcher, it should also set the overridden-method attribute so the two agree.

grails-web-common/src/main/groovy/org/grails/web/util/HiddenHttpMethod.java
```diff
@@ -0,0 +1,144 @@
...
+    public static String resolveOverride(HttpServletRequest request) {
+        if (!HttpMethod.POST.name().equalsIgnoreCase(request.getMethod())) {
+            return null;
+        }
+        String requested = request.getParameter(DEFAULT_METHOD_PARAM);
+        if (requested == null || requested.isBlank()) {
+            return null;
+        }
+        String candidate = requested.toUpperCase(Locale.ROOT);
+        return OVERRIDABLE_METHODS.contains(candidate) ? candidate : null;
```
jdaugherty, line 115: **Silent narrowing of override handling in the new default mode.** `resolveOverride` only honours a `_method` parameter on a `POST`, restricted to `PUT`/`PATCH`/`DELETE`; the previously-registered Grails filter also honoured the `X-HTTP-Method-Override` header and any method name. In default (dispatcher) mode the header is now ignored and an out-of-set `_method` is silently dropped rather than applied. This is called out in the upgrade guide, but it is a silent runtime change for header-based REST clients (e.g. a proxy that rewrites `DELETE` to `POST + X-HTTP-Method-Override`) — such a client's `POST /books/1` now matches the new `POST -> update` route and returns `200` from `update` while the caller believes it issued a `DELETE`. Worth ensuring this is prominent in the migration notes.

jdaugherty (dismissed):
I'm highly concerned about the vulnerabilities introduced in this PR, the removal of previous public methods, & the deprecation of the getCurrentRequest() method.

grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/ReflectionUtils.groovy
codeconsole: Confirmed and fixed in a39a3502. The mechanism is as you describe: `AnnotationFilterInvocationDefinition` resolves the URL to a controller/action inside the chain via `matchAllUrlMappings`, which matched on `grailsRequest.request.method`. So `POST /book/1` with `_method=DELETE` resolved to the `update` mapping - reachable at all only because this branch generates that POST route - was authorized against `update`, then executed as `delete`.

Matching now resolves the override first, so the action security authorizes is the action that runs. It closes in the safe direction: adding `_method` can only select the stricter rule, never a weaker one, because both sides read the same resolution. Under the servlet filter the request already reports the overridden method and it resolves to that. Two tests in `ReflectionUtilsSpec` assert the method the mapping lookup is handed, with and without an override.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/mvc/UrlMappingsHandlerMapping.groovy
codeconsole: Fixed in a39a3502 - the fallback now publishes what it resolves, so `effectiveMethod` and `allowedMethods` agree with the route it picked. Without it the request reached `delete` and was then refused a 405 for the POST it arrived as, exactly as you describe. Test added.

grails-controllers/src/main/groovy/org/grails/compiler/web/ControllerActionTransformer.java
codeconsole: You are right, and it is fixed in a39a3502 - every action writes the attribute again. The optimization looked safe from inside the controller being compiled, which is the flaw: the action that *reads* the attribute is in whichever controller is entered second, so it is not knowable from the one being compiled.

grails-web-common/src/main/groovy/org/grails/web/servlet/mvc/GrailsWebRequest.java
codeconsole: Fixed in a39a3502. `setMultipartRequest(..)` is back, `@Deprecated(since = "8.0")`, publishing its argument as `WebUtils.MULTIPART_HTTP_SERVLET_REQUEST_ATTRIBUTE` and discarding the cached params, so an existing caller keeps working. Upgrade guide 54.4 now covers it beside `getCurrentRequest()`, with the attribute-plus-`multipartRequestResolved()` form for callers who want to move off it.

grails-web-common/src/main/groovy/org/grails/web/util/WebUtils.java
codeconsole: Kept the broad catch but made it non-silent, in a39a3502. Narrowing to `MultipartException` would lose the cases this exists for - the container fails the parameter read itself, and Tomcat and Jetty do not all surface that as a Spring exception. So the fallback still applies to any `RuntimeException` on a multipart request, but only a `MultipartException` is the expected case: anything else is logged at warn rather than debug, so an unrelated failure is visible instead of passing as empty params.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultUrlMappingEvaluator.java
codeconsole: Documented rather than detected, in a39a3502. Keying the mode on bean presence means resolving a filter bean from three call sites during evaluation, one of which runs while bean definitions are still being registered, so I would rather not make mode resolution order-dependent in this PR. The upgrade note now says plainly that an application registering its own `HiddenHttpMethodFilter` must set the property too, otherwise it gets its filter *and* dispatcher mode, including the generated POST member route.

grails-web-common/src/main/groovy/org/grails/web/util/HiddenHttpMethod.java
codeconsole: Made prominent in a39a3502. The note now spells out the silent case rather than only stating the narrowing: a proxy or SDK that rewrites `DELETE` to `POST` plus `X-HTTP-Method-Override` has its `POST /books/1` answered by `update` with a 200 while the caller believes it issued a `DELETE`, and the remedy - enable the filter for an application such a client talks to - is stated with it.

grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
```diff
@@ -200,6 +202,16 @@ class ControllersAutoConfigurationSpec extends Specification {
                 }
     }
 
+    private WebApplicationContextRunner hiddenMethodContextRunner() {
```
matrei: Unused?

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -3034,7 +3150,254 @@ job is disabled with `jobEnabled = false`, or the class is not a job artefact of
...
+[source,groovy]
+----
+// Before
+def uri = webRequest.currentRequest.requestURI
+
+// After
+def uri = webRequest.request.requestURI
+----
+
+Nothing about the value returned has changed, so this is a rename, not a behaviour change. Code that
```
matrei: I would not call this a rename. `getRequest()` was always there.

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
```diff
@@ -1575,6 +1580,7 @@ class FormTagLib implements ApplicationContextAware, InitializingBean, TagLibrar
     }
 
     @Override
+    @GrailsCompileStatic
```
matrei: Why `@GrailsCompileStatic` and not `@CompileStatic`?

matrei (commented):
# Review Round 2 — Mergeable

Current head: `533f7e2cfb` (branch `codeconsole/perf/request-path-8.0.x`).

The three findings from the previous round are addressed, and the new work adds a
real security fix and end-to-end coverage. I would merge.

## Previous P1/P2 findings — all resolved

### P1a — Command-object and request-body binding read the effective method

Both call sites now go through `HiddenHttpMethod.effectiveMethod(request)`:

- `grails-controllers/src/main/groovy/grails/artefact/Controller.groovy:416`
- `grails-web-databinding/src/main/groovy/org/grails/web/databinding/bindingsource/AbstractRequestBodyDataBindingSourceCreator.groovy:51`

`AbstractRequestBodyDataBindingSourceCreatorSpec` now asserts a `POST` naming
`DELETE` leaves the body unbound while a plain `POST` still binds
(`AbstractRequestBodyDataBindingSourceCreatorSpec.groovy:122-143`). The
command-object branch is only reachable with a GORM-backed domain, so it is
covered by the rule and the upgrade note rather than by a test.

### P1b — Dispatcher override across forwards/includes

The fix picks the opposite of what the previous comment suggested — instead of
clearing the attribute for internal dispatches, it now honours the published
override wherever it applies and refuses only to *derive a fresh one* from
inherited parameters
(`UrlMappingsHandlerMapping.groovy:91-109`). This aligns dispatcher mode with the
servlet filter, whose request wrapper reports the overridden method for the
whole request. It is a defensible choice and is documented in the method's
Javadoc. Both readers agree during an internal dispatch:

- `an override the dispatcher resolved survives an internal dispatch`
  (`HiddenHttpMethodHandlerMappingSpec.groovy:97-114`) drives a forward and an
  include and asserts both the mapping and `HiddenHttpMethod.effectiveMethod`
  return `PUT`.
- `the parameter does not select a route for an internal dispatch` still holds
  when no override was published (`.groovy:81-95`).

### P2 — Multipart branch in `SpringSecurityUtils.isAjax()`

Now reads through the tolerant helper:
`SpringSecurityUtils.groovy:313 → WebUtils.readParameter(multipart, 'ajax')`,
whose `readTolerantly` swallows `RuntimeException` on multipart requests and
returns the fallback (`WebUtils.java:646-666`). The isAjax-specific test with a
throwing multipart wrapper the previous comment asked for is not present, but
the behaviour is exercised by `WebUtilsSpec` (`the parameter reads yield their
fallback when a multipart body cannot be parsed`).

## New work in this round — worth calling out

- **Security parity fix (`ReflectionUtils.matchAllUrlMappings`, a39a3502)** —
  Spring Security's URL matching now resolves `_method` before matching so
  authorization applies to the action the dispatcher will actually run. This
  closes a real gap: on this branch a `POST /book/1` with `_method=DELETE` was
  authorized against `update` and then executed as `delete`. Covered by two new
  cases in `ReflectionUtilsSpec.groovy:224-263`. The framework invariant that
  either the filter rewrites the method or dispatcher-mode resolves it
  (`UrlMappingsGrailsPlugin.groovy:130` + `GrailsHiddenHttpMethodFilterAutoConfiguration`)
  keeps this safe — reading `_method` here cannot select a route the dispatcher
  will not execute.

- **`ALLOWED_METHODS_HANDLED` now written by every action
  (`ControllerActionTransformer.java:530-580`)** — Fixes a 405 mid-request when
  an unrestricted controller forwards into a restricted action in another
  controller. The compile-time optimisation of skipping the attribute for an
  unrestricted controller was unsafe: the reader is in whichever controller is
  entered second. Existing coverage exercises intra-controller forwarding
  (`ControllerActionTransformerAllowedMethodsSpec`); a cross-controller forward
  from an unrestricted controller to a restricted one is not directly asserted,
  but the change is a uniform simplification and the behaviour rests on the
  attribute contract.

- **`GrailsWebRequest.setMultipartRequest` restored as a deprecated shim
  (`GrailsWebRequest.java:186-192`)** — Backwards-compat for external callers,
  bridges to `MULTIPART_HTTP_SERVLET_REQUEST_ATTRIBUTE` publication.

- **`WebUtils.readTolerantly` distinguishes expected vs unexpected multipart
  failures (`WebUtils.java:646-666`)** — `MultipartException` at DEBUG,
  everything else at WARN. Improves observability without changing behaviour.

- **End-to-end `RequestPathSpec`** — Drives the whole chain (filter chain,
  multipart resolution, dispatcher override, mapping, allowedMethods, params,
  command binding, response) in one round trip for six cases. This is the test
  the earlier rounds kept asking for.

## Minor observations (non-blocking)

- `RequestPathController` defines `forwarding` and `forwarded` actions
  (`RequestPathController.groovy:55-61`) that no spec exercises. Either add a
  forward-through-override case to `RequestPathSpec` or drop the two actions.
  A functional test asserting the forward case would strengthen coverage of the
  P1b design decision (that a forward inherits the resolved override) in a
  running container.
- No `isAjax`-specific spec covers a multipart wrapper that throws from
  `getParameter()`. The tolerance is covered by `WebUtilsSpec`, so this is
  optional.
- The `ALLOWED_METHODS_HANDLED` cross-controller forward scenario (unrestricted
  → restricted, in different controllers) is described in the commit message
  but not asserted by a test. The uniform rule the commit installs makes this
  hard to regress accidentally.

## Verdict

Mergeable. The security bypass fix in particular is worth landing; the
end-to-end spec closes the long-standing "test the whole path together" ask; the
design of the effective-method rule now agrees across every reader.

matrei (approved):

grails-controllers/src/test/groovy/org/grails/plugins/web/controllers/ControllersAutoConfigurationSpec.groovy
codeconsole: Yes - no callers. A fixture for tests that ended up written another way. Removed in 89b3890.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Agreed, that was wrong. `getRequest()` has always been on `ServletRequestAttributes`; what changed is that `getCurrentRequest()` no longer returns anything different from it. Reworded in 89b3890 to say that rather than "rename".

grails-gsp/plugin/src/main/groovy/org/grails/plugins/web/taglib/FormTagLib.groovy
codeconsole: It arrived with the trait field cache this branch added and then removed, same as the explicit `getGrailsLinkGenerator()` call you and sbglasius spotted earlier. The rest of the file uses plain `@CompileStatic` and 8.0.x annotates this method not at all, so in 89b3890 the annotation and its import are gone and the method matches 8.0.x exactly.

codeconsole:
Thanks. Inline comments answered on their threads; the three review observations are in c8bbcb8.

**Unexercised forward actions** — kept and put to work rather than dropped, since they cover the P1b decision in a running container. The forward now targets the member URL instead of naming an action, so the mappings route it: `delete` for a POST naming DELETE, `update` for the same POST without it.

**Cross-controller `ALLOWED_METHODS_HANDLED`** — added. A new `RequestPathForwarderController` restricts nothing and forwards into the DELETE-only `delete`; it is admitted because the first action recorded that it began handling.

**`isAjax` multipart test** — written, then removed. It passed against the direct `multipart?.getParameter('ajax')` read it was meant to catch, so the branch is not reached under that harness and it asserted nothing. `WebUtilsSpec` keeps the coverage.

`RequestPathSpec` is nine cases now.

jdaugherty (dismissed):
I'm a -1 on merging this until the getCurrentRequest() deprecation is removed. This API has been there since the inception of Grails and I disagree with the removal.  I'm ok leaving the renames.

codeconsole:
Done in 5fa7a5e - the `@Deprecated` is off `getCurrentRequest()` and nothing else changed. The renames stay, as you said.

The method still returns what `getRequest()` returns, since the multipart substitution that used to make the two differ is gone; that is now just two accessors answering identically rather than one being steered away from. The spec asserted the annotation was present, so it asserts the behaviour instead, and upgrade note 54.5 says the two are the same object rather than telling readers to move off it.

`setMultipartRequest(..)` keeps its deprecation - that is the shim you asked for in the earlier round, where the method had been removed outright.

PR #16152, opened 2026-08-15
Let oversized uploads reach the application's error handling

### LINKED ISSUE #16145: [Grails 8] Oversized multipart uploads cannot be handled by Grails application code

codeconsole:
> **Stacked on #16149.** This branch builds on the multipart discovery work there —
> `GrailsParameterMap` uses `WebUtils.resolveMultipartRequest`, which #16149 introduces — so GitHub
> shows this PR against `8.0.x` with #16149's commits included. **The change here is 6 commits and
> 27 files**; the rest belongs to #16149. Review after that merges, at which point this rebases to
> just its own commits.

Closes #16145.

An oversized multipart upload never reaches the application. Grails reads request parameters
before the `DispatcherServlet` runs — `HiddenHttpMethodFilter` looks for `_method`, and
`GrailsWebRequestFilter` builds `GrailsParameterMap` — and reading any parameter of a
`multipart/form-data` request makes the container parse the body. A body past the configured
limit fails that parse, inside the filter chain, where no `HandlerExceptionResolver` can see it.

Spring resolves multipart in `DispatcherServlet.checkMultipart`, inside `doDispatch`, which is
why a plain Spring MVC application handles this correctly out of the box. Grails only differs
because it reads parameters first.

So the framework's parameter reads now tolerate a multipart body the container refuses to parse,
and the failure is left for `checkMultipart` to raise during dispatch, where Spring routes it to
the exception resolvers and the application's error handling runs. Any non-multipart parameter
failure still propagates unchanged.

## Measured

Same 200 KB upload against a 128000-byte limit, same probe, on three containers:

| Container | Before | After |
|---|---|---|
| Tomcat 11 | `500`, raw Tomcat HTML page | `413`, `{"error":"Content Too Large","handledBy":"errors.tooLarge"}` |
| Jetty 12 | `400` "Bad Request" — indistinguishable from a malformed request | `413`, handled by the application |
| Undertow 2.4 | `413`, empty body | `413`, empty body — **unchanged** |

Undertow applies the limit while reading the request entity, so no filter or servlet runs and
there is nothing the framework can influence. That limitation is documented and pinned by a spec
so a change in it is noticed.

## Also fixed

An error handler that itself fails was forwarded to indefinitely. With a `"413"` or `"500"`
controller mapping, the error dispatch re-entered `DispatcherServlet`, resolved the same mapping,
faulted on the same unreadable request and forwarded again — 78,120 nested frames,
`StackOverflowError`, then `OutOfMemoryError` while logging the unwind. `GrailsExceptionResolver`
no longer forwards to an error handler while a forward to it is already in progress. This is
independent of multipart: any error handler that throws would have looped.

## Notes

- `<g:uploadForm method="PUT">` (multipart plus `_method`) keeps working; there is a test.
- During an error dispatch for this failure the action sees `dispatcherType=ERROR`, empty `params`
  and no `request.getFile(..)`, since the body was never readable. Documented.
- Covers Tomcat, Jetty and Undertow with integration tests in `grails-test-examples`.


matrei:
Problem and solution in a nutshell:

```console
EARLY SPRING / GRAILS
=====================

HTML only GET/POST
       |
       v
_method convention
       |
       v
HiddenHttpMethodFilter
       |
       +-- calls getParameter()
               |
               v
        problem for multipart


SPRING'S HISTORICAL ADVICE
==========================

MultipartFilter
       |
       v
HiddenHttpMethodFilter
       |
       v
DispatcherServlet


SPRING BOOT, 2019
=================

Eager/automatic HiddenHttpMethodFilter work found expensive
       |
       v
Disabled by default


GRAILS HISTORICAL BEHAVIOR
==========================

HiddenHttpMethodFilter
GrailsWebRequestFilter
       |
       +-- read parameters before DispatcherServlet
               |
               v
       oversized multipart parsed too early
               |
               v
       container owns failure response


#16146
======

Move multipart resolution earlier
       |
       v
Solves some ordering/access problems
but introduces filter lifecycle,
wrapper, cleanup and performance complexity


#16149
======

Keep outermost request
       |
       v
Discover MultipartHttpServletRequest through wrapper chain


#16152
======

Do NOT move multipart parsing earlier
       |
       v
Tolerate premature parameter failures
for unreadable multipart requests
       |
       v
DispatcherServlet.checkMultipart()
       |
       v
MultipartException
       |
       v
HandlerExceptionResolver
       |
       v
APPLICATION CONTROLS RESPONSE
```

codeconsole:
@matrei yeah, that's a good outline of what is going on.   The only issue is performance on apps that use multipart requests frequently.  Performance is only applicable to those requests.   I have a few other solutions I will present shortly.

codeconsole:

PR #16156, opened 2026-08-16
Add test coverage and fix real bugs in grails-datamapping-core's services package

borinquenkid:
## Summary
- Adds unit test coverage for `DefaultTenantService` and `DefaultTransactionService` (`org.grails.datastore.gorm.services`), both previously at 0% coverage. Pure Spock-mock specs, no live datastore needed.
- Adds coverage for a batch of under-tested `@Service` AST-transform implementer classes (`UpdateOneImplementer`, `DeleteWhereImplementer`, `FindOneInterfaceProjectionWhereImplementer`, `FindAllPropertyProjectionImplementer`, `FindOnePropertyProjectionImplementer`, the `findById(id)` shortcut in `FindOneByImplementer`, and the `AbstractWriteOperationImplementer.enhance()` path for abstract-class services with pre-existing concrete write methods), all via the existing `GroovyClassLoader().parseClass(...)` + `@Implemented(by=...)` structural-testing convention already used by `ServiceTransformSpec`/`WhereConnectionRoutingSpec`.
- **Real bug fix**: `AbstractSaveImplementer`'s invalid-argument error path referenced `newMethodNode.declaringClass.module` (null on the synthesized implementation method), crashing the compiler with a `NullPointerException`/`GroovyBugError` instead of reporting a clean compile error. Fixed to use `abstractMethodNode.declaringClass.module.context`, matching the pattern used elsewhere in this package (e.g. `FindAllByImplementer`).
- A handful of IntelliJ-inspection cleanups surfaced while reviewing these classes:
  - Suppressed (not "fixed") warnings on `getDetachedCriteriaType`/`lookupById`/`findMethodToInvoke`/`getFindMethodToInvoke` where the flagged parameter/method is a legitimate `protected` polymorphic extension point on a class with real subclasses — changing the signature or making it `static` would silently break overriding.
  - Removed genuinely dead code: `bindParametersAndSave`'s unused `newMethodNode` parameter (a side effect of the `AbstractSaveImplementer` fix above), and three unused helper methods on `AbstractServiceImplementer` (`transactionalDatastore`, `transactionService`, `findInstanceApiForConnectionId`) with zero call sites anywhere in the repo.
  - Deduplicated a 13-line interface-projection property-compatibility check that was copy-pasted between `InterfaceProjectionBuilder` and `IterableInterfaceProjectionBuilder` into a shared `hasCompatibleProperties(...)` method.
  - Replaced an array range-slice (`parameters[1..-1] as Parameter[]`) with `Arrays.copyOfRange(...)` in `UpdateOneImplementer`, since IntelliJ's stricter `@CompileStatic` checker can't resolve the `getAt` overload for it even though `groovyc` compiles it fine.

## Test plan
- [x] `./gradlew :grails-datamapping-core:test` — full module suite passes
- [x] `./gradlew :grails-datamapping-core:codeStyle` — 0 Checkstyle/CodeNarc violations
- [x] `./gradlew :grails-datamapping-rx:compileGroovy` — sibling module unaffected by `AbstractServiceImplementer` cleanup

### DIFF
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractDetachedCriteriaServiceImplementor.groovy
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractDetachedCriteriaServiceImplementor.groovy
@@ -153,6 +153,9 @@ abstract class AbstractDetachedCriteriaServiceImplementor extends AbstractReadOp
      * Whether lookup by id is allowed by this implementation
      * @return True if it is
      */
+    // Not static: AbstractProjectionImplementer overrides this to return false, and
+    // doImplement() dispatches on it polymorphically. A static method would break that override.
+    @SuppressWarnings('MethodMayBeStatic')
     protected boolean lookupById() {
         return true
     }
--- a/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractDetachedCriteriaServiceImplementor.groovy
+++ b/grails-datamapping-core/src/main/groovy/org/grails/datastore/gorm/services/implementers/AbstractDetachedCriteriaServiceImplementor.groovy
@@ -123,6 +123,9 @@ abstract class AbstractDetachedCriteriaServiceImplementor extends AbstractReadOp
         }
     }
 
+    // domainClassNode is unused here, but kept so subclasses can override this as a
+    // polymorphic extension point and pick a DetachedCriteria type based on the domain class
+    @SuppressWarnings(['unused', 'MethodMayBeStatic'])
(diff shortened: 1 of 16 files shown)


jdaugherty:
@borinquenkid if this is a real bug, why aren't we targeting 8.0.x?

borinquenkid:
Good catch — the `newMethodNode.declaringClass.module` NPE bug in `AbstractSaveImplementer` is present identically on 7.0.x, 8.0.x, and 8.1.x, so it's not new to this branch. Retargeted this PR to 7.0.x so the fix (and the new test coverage) flows forward through the normal branch merges instead of being scoped only to 8.1.x.

jdaugherty (approved):

jdaugherty:
I had suggested 8.0.x for this, but I'm ok merging all the way back to 7.0.x

PR #15558, opened 2026-04-05
Grails 9 on Groovy 6.0.0-beta-2

jamesfredley:
## Purpose

Move the Grails 9 line onto Apache Groovy `6.0.0-beta-2` so Grails 9 snapshots can be built and published against Groovy 6. Groovy 6 RC1 is expected shortly, and Grails and Groovy are sister Apache projects - Grails 9 needs to be tracking Groovy 6 rather than discovering the work later.

Base branch: **`9.0.x`** (refreshed with `8.0.x` -> `8.1.x` -> `9.0.x` merges).

Groovy 5 baseline behavior from #15557 is accepted as-is and is not relisted here.

## Workaround inventory

**The catalogue of Groovy 6 changes, why each exists, and when each can be deleted lives in #16157.** That issue is the documentation of record; this description tracks branch state only.

## Baseline

| Component | Version |
|---|---|
| Apache Groovy | `6.0.0-beta-2` |
| Spring Boot / Framework | `9.0.x` line (Boot 4.1.0, Framework 7.0.8) |
| Gradle | `9.6.0` (matches trunk) |
| JDK | 21+ |
| Spock | `2.4-groovy-5.0` (see below) |

## Two accepted limitations

Both are deliberate, both have tracking issues, and neither blocks producing a Groovy 6 snapshot.

### 1. Spock runs via its Groovy version-check opt-out (#16157)

Spock has no `groovy-6.0` build, so the build passes `-Dspock.iKnowWhatImDoing.disableGroovyVersionCheck=true`.

This is not cosmetic and cannot simply be deleted. Spock registers `SpockTransform` as a **global** AST transform through `META-INF/services`, so the Groovy compiler loads it for any compile whose classpath contains Spock. Verified by removing the flag and building: 99 modules compiled fine, then

```
> Task :grails-test-core:compileGroovy FAILED
Could not instantiate global transform class org.spockframework.compiler.SpockTransform
  ... IncompatibleGroovyVersionException: Spock 2.4.0-groovy-5.0 is not compatible with Groovy 6.0.0-beta-2
```

Thirteen modules put Spock on a **main** classpath - `grails-test-core` and `grails-testing-support-views-gson` declare it `api`, several `grails-testing-support-*` modules declare it `implementation` - so without the opt-out those published artifacts do not compile at all, and a Groovy 6 snapshot would ship without Grails' Spock-based testing support. The flag is applied wherever it is required to compile, test and execute, and it goes away when Spock ships a `groovy-6.0` artifact.

[...]

PR #16160, opened 2026-08-16
fix(config): bind nested settings maps under Spring 7

### LINKED ISSUE #16159: ConfigurationBuilder fails to bind nested settings under Spring 7
Spring Framework 7 no longer converts a configuration `Map` into a type annotated `@Builder(builderStrategy = SimpleStrategy)`. `ConfigurationBuilder` relies on that conversion to bind nested settings from `application.yml` / `application.groovy`, so those settings now fail with:

```
org.springframework.core.convert.ConverterNotFoundException
```

**Grails 8 is the first version affected**, because it is the first to ship Spring 7: `8.0.x` pins Spring Boot `4.1.0`, which resolves `spring-core` **7.0.8**. `9.0.x` inherits the same versions, and `ConfigurationBuilder.groovy` plus its spec are byte-identical on both branches, so a fix landed on `8.0.x` flows up the maintenance chain (`8.0.x` -> `8.1.x` -> `9.0.x`) and covers both lines.

This has nothing to do with the Groovy version. Both branches are on Groovy `5.0.8`, and the affected code carries its own note that it is "independent of the Groovy version".

## Reproducing

With the current `ConfigurationBuilder` and the spec added in #16160, six scenarios fail with:

```
Expected exception of type 'ConfigurationException',
  but got 'org.springframework.core.convert.ConverterNotFoundException'
```

## Impact

Nested configuration that binds through a `SimpleStrategy` builder does not apply. Real consumers include the Hibernate and connection-source settings trees. Because the failure surfaces during configuration binding, it shows up at application startup rather than as a test failure.

## Fix

#16160 targets `8.0.x`. `ConfigurationBuilder` now instantiates the target type and populates it from the Map when, and only when, Spring genuinely has no converter.

Two independent reviewers examined it over five rounds and found nine distinct defects in the handling, all of which are fixed and covered by regression tests in that PR:

| # | Defect | Consequence if unguarded |
|---|---|---|
| 1 | Fallback engaged without checking the cause chain for `ConverterNotFoundException` | A converter that deliberately rejects a Map could be bypassed |
| 2 | `catch (Throwable)` swallowed `ConfigurationException` | Unknown-key and malformed-value failures were masked |
| 3 | Raw-lookup failures logged and discarded | Configuration whose lookup failed was silently accepted via the fallback |
| 4 | Fresh instance built without copying the fallback | Every unspecified field was lost, breaking per-field inheritance for named connection settings |
| 5 | Nested resolution passed a null fallback | Inheritance worked only at the first level; deeper children were reset |
| 6 | Map entries assigned without conversion | `multiTenancy.mode: database` was rejected where `DATABASE` worked |
| 7 | `Class` entries resolved via the property resolver | `hibernate.configClass` and other application classes were left unbound, because that converter resolves against the framework class loader |
| 8 | Every key required to be a declared bean property | `HibernateSettings extends LinkedHashMap` exists to carry arbitrary keys such as `hibernate.hbm2ddl.auto`; these were rejected |
| 9 | Flattened descendant keys rejected as unknown | The resolver flattens nested config, so anything nested more than one level failed to build |

Plus a null-varargs fix so an explicit null can clear an inherited value.

Strictness is preserved where it belongs: a dotted key whose first segment is unknown is still rejected, and non-`Map` types still reject unknown keys, both covered by guard specs.

## Status

Fixed in #16160 (targets `8.0.x`). `ConfigurationBuilderSpec` goes from 10 to 22 specs and the module total from 108 to 120, all passing on the `8.0.x` base.

One known limitation is documented in that PR: a `PropertyResolver` exposing only an aggregate map, without its entries addressable as dotted properties, can still yield null for a configured scalar. Grails' own `DatastoreUtils.createPropertyResolver` flattens and is unaffected.

## Provenance

The fix was originally written on the Groovy 6 canary branch (#15558), which is simply where the failure was first noticed. It does not belong there, and #16160 moves it to `8.0.x` so both the Grails 8 and Grails 9 lines get it.

jamesfredley:
Fixes #16159.

## The problem

Spring Framework 7 no longer converts a configuration `Map` into a type annotated `@Builder(builderStrategy = SimpleStrategy)`. `ConfigurationBuilder` depends on that conversion to bind nested settings from `application.yml` / `application.groovy`, so those settings now fail with `ConverterNotFoundException` at configuration-binding time - that is, at application startup.

**Grails 8 is the first affected version**, because it is the first on Spring 7: `8.0.x` pins Spring Boot `4.1.0`, which resolves `spring-core` **7.0.8**.

This targets **`8.0.x`** so the fix flows up the maintenance chain (`8.0.x` -> `8.1.x` -> `9.0.x`) instead of landing only on the newest line. Both branches pin the same Spring Boot and Groovy versions, and `ConfigurationBuilder.groovy` plus its spec are byte-identical on each, so one fix covers both.

The gap is demonstrable rather than theoretical: with the previous `ConfigurationBuilder` and only this PR's spec applied, six scenarios fail with

```
Expected exception of type 'ConfigurationException',
  but got 'org.springframework.core.convert.ConverterNotFoundException'
```

## The fix

`ConfigurationBuilder` now instantiates the target type and populates it from the Map when, and only when, Spring genuinely has no converter.

Two independent reviewers examined the handling over five rounds and found nine distinct defects. Every guard below exists because removing it produced an observable failure, and each has a regression test:

| Guard | Failure without it |
|---|---|
| Engage only when the cause chain contains `ConverterNotFoundException` | A converter that deliberately rejects a Map could be bypassed |
| Never suppress `ConfigurationException` | Unknown-key and malformed-value failures were masked by the original exception |
| Throw on raw-lookup failure instead of falling back | Configuration whose lookup had failed was silently accepted |
| Inherit from the fallback before applying overrides | Overriding one field discarded every unspecified field |
| Pass each property's fallback child into nested conversion | Inheritance worked only at the first level; deeper children were reset |
| Convert values to the target property type | `multiTenancy.mode: database` was rejected where `DATABASE` worked |
| Resolve `Class` entries via the thread context class loader | `hibernate.configClass` and other application classes were left unbound, because that converter resolves against the framework class loader |
| Let `Map`-backed types keep arbitrary entries | `HibernateSettings extends LinkedHashMap` exists to carry keys such as `hibernate.hbm2ddl.auto`; strict binding rejected them |
| Bind flattened descendant keys through their parent | The resolver flattens nested config, so anything nested more than one level failed to build |
| Invoke setters with an explicit single-element argument array | The Java null-varargs pitfall meant an explicit null could not clear an inherited value |

Strictness is preserved where it belongs: a dotted key whose **first** segment is unknown is still rejected, and non-`Map` types still reject unknown keys. Both are covered by guard specs so a future change cannot quietly relax them.

## Testing

`ConfigurationBuilderSpec` grows from **10 to 22** specs; the module total goes from **108 to 120**, all passing.

```
:grails-datastore-core:test --rerun-tasks   BUILD SUCCESSFUL, 46 tasks executed
tests=120 failures=0 errors=0
ConfigurationBuilderSpec tests=22 failures=0
```

Verified on the `8.0.x` base, not only on the branch this was originally written against.

Some specs use a narrow `PropertyResolver` proxy to reproduce the Spring 7 failure boundary, because the real resolver auto-converts the top-level bean and would bypass the handler entirely. Raw-map shape and dotted lookups still go through the real `DatastoreUtils.createPropertyResolver`.

## Known limitation

A `PropertyResolver` that exposes only an aggregate map, without its entries addressable as dotted properties, can still yield null for a configured scalar. Grails' own `DatastoreUtils.createPropertyResolver` flattens and is unaffected. Binding the raw value unconditionally was considered and rejected because it would bypass the type conversion and case-insensitive enum handling above; both reviewers agreed that would be a net regression. A follow-up could use the raw value only when it is already assignable.

## Provenance

This was first written on the Groovy 6 canary branch (#15558), which is simply where the failure was noticed. It is not Groovy-version-specific - both branches are on Groovy `5.0.8` and the code carries its own note to that effect - so it belongs here. Once this reaches `9.0.x` through the merge chain, the same change is dropped from #15558.

### DIFF
--- a/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/ConfigurationBuilder.groovy
+++ b/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/ConfigurationBuilder.groovy
@@ -18,6 +18,9 @@
  */
 package org.grails.datastore.mapping.config
 
+import java.beans.Introspector
+import java.beans.PropertyDescriptor
+import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
 
@@ -28,6 +31,7 @@ import groovy.transform.builder.SimpleStrategy
 import groovy.util.logging.Slf4j
 
 import org.springframework.core.convert.ConversionFailedException
+import org.springframework.core.convert.ConverterNotFoundException
 import org.springframework.core.env.PropertyResolver
 import org.springframework.util.ClassUtils
 import org.springframework.util.ReflectionUtils
@@ -393,23 +397,12 @@ abstract class ConfigurationBuilder<B, C> {
                         try {
                             value = propertyResolver.getProperty(propertyPathForArg, argType, fallBackValue)
                         } catch (ConversionFailedException e) {
-                            if (argType.isEnum()) {
-                                value = propertyResolver.getProperty(propertyPathForArg, String)
-                                if (value != null) {
-                                    try {
-                                        value = Enum.valueOf((Class) argType, value.toUpperCase())
-                                    } catch (Throwable e2) {
-                                        // ignore e2 and throw original
-                                        throw new ConfigurationException("Invalid value for setting [$propertyPathForArg]: $e.message", e)
-                                    }
-                                }
-                                else {
-                                    throw new ConfigurationException("Invalid value for setting [$propertyPathForArg]: $e.message", e)
-                                }
-                            }
-                            else {
-                                throw new ConfigurationException("Invalid value for setting [$propertyPathForArg]: $e.message", e)
-                            }
+                            value = handleConversionException(e, argType, propertyPathForArg, fallBackValue)
+                        } catch (ConverterNotFoundException e) {
+                            // Spring 7 nested-map conversion fallback: handle types with
+                            // @Builder(builderStrategy = SimpleStrategy) where Spring cannot
+                            // auto-convert from Map. Independent of the Groovy version.
+                            value = handleConverterNotFoundException(e, argType, propertyPathForArg, fallBackValue)
                         }
                         if (value != null) {
                             log.debug('Resolved value [{}] for setting [{}]', value, propertyPathForArg)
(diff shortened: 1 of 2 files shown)


grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/ConfigurationBuilder.groovy
```diff
@@ -393,23 +397,12 @@ abstract class ConfigurationBuilder<B, C> {
...
-                                }
-                            }
-                            else {
-                                throw new ConfigurationException("Invalid value for setting [$propertyPathForArg]: $e.message", e)
-                            }
+                            value = handleConversionException(e, argType, propertyPathForArg, fallBackValue)
+                        } catch (ConverterNotFoundException e) {
+                            // Spring 7 nested-map conversion fallback: handle types with
+                            // @Builder(builderStrategy = SimpleStrategy) where Spring cannot
+                            // auto-convert from Map. Independent of the Groovy version.
```
jdaugherty, line 404: `@Builder` is RUNTIME-retained on Groovy 5.0.8, so an argType annotated with `@Builder(builderStrategy = SimpleStrategy)` is intercepted by the `argType.getAnnotation(Builder)` branch earlier in `buildRecurse` and never reaches this `getProperty` call. The types that actually land here are ones with no runtime-visible `@Builder` (like the spec's synthetic settings classes). Two asks:

1. Reword this comment (and the `handleConverterNotFoundException` javadoc) so it doesn't name `@Builder(SimpleStrategy)` types as the case being handled here.
2. #16159 states the Hibernate and connection-source settings trees are real consumers, but `HibernateConnectionSourceSettings`, its nested types, `ConnectionSourceSettings`, and `MultiTenancySettings` are all annotated and bind through the recursion branch (the pre-existing specs in this file exercise that path and passed on `8.0.x` before this change). Which shipped configuration reproduces the startup failure on `8.0.x`? Worth capturing in the issue/PR so the affected surface is clear.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/ConfigurationBuilder.groovy
```diff
@@ -463,4 +456,196 @@ abstract class ConfigurationBuilder<B, C> {
...
+            throw new ConfigurationException("Invalid class name [$className] for setting [$propertyPath]: ${e.message}", e)
+        }
+    }
+
+    private Object resolveMapValue(Class propertyType, String propertyPath, Object fallBackValue, Object rawValue) {
+        // Class-typed entries must use the same thread context class loader route as the
+        // top-level Class handling above, because the resolver's String->Class converter
+        // resolves against the framework class loader and silently leaves an
+        // application-defined class (hibernate.configClass, for example) unbound.
+        if (propertyType == Class) {
```
jdaugherty, line 629: This branch and `resolveClassValue` aren't exercised by the new specs: `MapBackedSettings.configClass` is declared as `String`, and no test settings type has a `Class`-typed property, so the guard listed in the PR table ("Resolve `Class` entries via the thread context class loader") has no regression test. Could you add coverage for a nested `Class`-typed property — a `Class` literal value, a String class name, and an invalid class name?

While adding that, worth pinning the fallback behavior: the top-level `argType == Class` handling falls back to `fallBackValue` when no usable value is found, but `resolveClassValue` returns null, so a value that is neither a `Class` nor a `CharSequence` would silently clear an inherited fallback through the setter instead of retaining it (or erroring).

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/config/ConfigurationBuilder.groovy
```diff
@@ -463,4 +456,196 @@ abstract class ConfigurationBuilder<B, C> {
...
+            return resolveClassValue(propertyPath)
+        }
+        if (rawValue instanceof Map && !propertyType.isInstance(rawValue)) {
+            return handleConverterNotFoundException(null, propertyType, propertyPath, fallBackValue, rawValue)
+        }
+        try {
+            Object value = propertyResolver.getProperty(propertyPath, propertyType)
+            Object rawPropertyValue = propertyResolver.getProperty(propertyPath, Object)
```
jdaugherty, line 637: The raw lookup runs even when the typed lookup on the previous line already returned a non-null value, where its result is discarded. It's only needed in the `value == null` case, so it can move inside that branch and save a resolver pass per nested scalar.

jdaugherty:
Pushed 8aee638. It addresses the four open review comments and a defect they surfaced.

## Which shipped configuration reproduces this

`hibernate.hibernateEventListeners`. Configured as

```yaml
hibernate:
  hibernateEventListeners:
    listenerMap:
      pre-load: someListener
```

Spring 7 throws at binding time:

```
ConverterNotFoundException: No converter found capable of converting from type
[java.util.LinkedHashMap<?, ?>] to type [org.grails.orm.hibernate.HibernateEventListeners]
```

`HibernateEventListeners` is a plain bean with no runtime `@Builder`, which is why it reaches
this branch. The review comment on the `@Builder` retention is correct — `javap` on
`groovy-5.0.8.jar` confirms `groovy.transform.builder.Builder` is `RUNTIME`-retained, so
`argType.getAnnotation(Builder)` intercepts every `SimpleStrategy` type in the recursion above
and those never reach the new handler. The Hibernate and connection-source trees named in
\#16159 bind through the recursion, not here. The code comment and javadoc are corrected to
describe plain settings beans instead.

## The fallback did not bind it

With the previous revision applied, that configuration produced a `HibernateEventListeners`
with a null `listenerMap` — a loud startup exception became silent data loss.

This is the documented "known limitation", and the claim that
`DatastoreUtils.createPropertyResolver` is unaffected by it does not hold.
`createFlatConfig` only emits dotted keys for `ConfigObject` values; a plain nested `Map` is
stored as a leaf. So `getProperty('hibernate.hibernateEventListeners.listenerMap', Map)`
returns null while the value sits in the very map `resolveMapValue` was handed and discarded.

`resolveMapValue` now binds the entry in hand when the typed lookup finds nothing at the path
and the value is already assignable — the follow-up proposed in the description. Values needing
conversion still route through the enum, `Class` and nested-map branches, so the case-insensitive
enum handling is untouched.

## Review comments addressed

| Comment | Change |
|---|---|
| Comment and javadoc name `@Builder(SimpleStrategy)` as the case handled | Reworded to plain settings beans; the recursion handles annotated types |
| `Class` branch and `resolveClassValue` have no regression test | Specs for a class literal, a class name |
| `resolveClassValue` returns null where the top-level handling falls back | Returns the fallback when  no class, with a spec |
| Raw lookup runs even when the typed lookup succeeded | Moved inside the `value == null` branch, one ralar |
| Map-backed entries stored under the original key object | Stored under the normalized `String`, with a spec using a `GString` key |

Two further defects found while verifying:

- A flattened descendant key passed the **descendant's** value as the parent property's raw
  value. Not reachable through the real resolver's key ordering, but it would bind a
  grandchild map onto its parent. The descendant is now resolved from its path.
- A setter rejecting a value reported the **parent's** path. It now reports the property's own
  path and the expected type, rather than surfacing `argument type mismatch`.

## Testing

`ConfigurationBuilderSpec` goes from 22 to 29. Five of the seven new specs fail against the
previous revision; the two `Class`-name specs pass on both and exist to close the coverage gap
rather than pin new behaviour. `HibernateConnectionSourceSettingsBuilderSpec` gains the real
shipped-configuration case above.

```
:grails-datastore-core:test                                  SUCCESS
:grails-datamapping-core:test                                SUCCESS
:grails-data-hibernate7-core:test + :grails-data-hibernate5-core:test
                                    3036 tests, 0 failures, 24 skipped
:grails-data-mongodb-core:test  --tests *MongoConnectionSource*  SUCCESS
:grails-datastore-core:codeStyle :grails-data-hibernate7-core:codeStyle  SUCCESS
```

Neo4j is not in `settings.gradle` on `8.0.x`, so `Neo4jDriverConfigBuilder` was not exercised.
PMD and SpotBugs tasks are not registered on this branch, so `codeStyle` (CodeNarc + Checkstyle)
is the analysis gate.

## Description needs an update

The "Known limitation" paragraph is now wrong on two counts — the flattening resolver *is*
affected, and the limitation is fixed rather than deferred. The impact table in #16159 also
still attributes the failure to the `@Builder` settings trees. Both want rewording before merge.

jdaugherty:
@jamesfredley I'm going to go ahead and merge so we can clear some of these back log PRs.  Lets open a PR if you feel we need to iterate on this.

PR #16169, opened 2026-08-18
feat: add app-level indy A/B benches to the performance workflow

jamesfredley:
## Purpose

Teach CI to measure **app-level** indy vs classic callsite, not just JMH micros.

This does **not** change the Grails plugin indy default. That remains #16165.

## What landed

- `AppHttpBench` helper + unit tests (rejects empty sample sets, clamps fork count)
- Gated `AppBench*` specs for latency, app1, and gsp-layout
- `gsp-layout` now depends on `grails-testing-support-http-client`
- Native Gradle orchestrator: `./gradlew :grails-benchmarks:appIndyBench`
- `benchmark.yml` `app-bench` job on the `performance` label calls that single task (warmup 80 / samples 300 / forks 2)
- No PowerShell, shell, or Python runner

## Local usage

```
./gradlew :grails-benchmarks:appIndyBench
```

Optional: `-PappBenchWarmup`, `-PappBenchSamples`, `-PappBenchForks`, `-PappBenchOutDir`.

## Local numbers already posted on #16165

Same harness, same settings: latency **0.98x**, app1 **0.99x**, no clear change. gsp-layout was skipped there because that module lacked the HTTP client dep; this PR adds it.

## Test plan

- [x] `AppHttpBenchSpec`
- [x] `AppIndyBenchSpec`
- [ ] CI `performance` label runs the new job


grails-benchmarks/scripts/run-app-indy-bench.ps1
```diff
@@ -0,0 +1,118 @@
...
+#
+# Experimental: app-level indy vs non-indy HTTP benches against grails-test-examples.
+# Run from repo root. Rebuilds framework+app with -PgrailsIndy, runs gated AppBench* specs,
+# then compares JMH-shaped JSON with :grails-benchmarks:jmhCompare.
+#
+# Usage:
+#   pwsh grails-benchmarks/scripts/run-app-indy-bench.ps1
+#   pwsh grails-benchmarks/scripts/run-app-indy-bench.ps1 -Samples 500 -Warmup 100
+
+param(
```
jdaugherty: Why are we using powershell for this?

jamesfredley: Agreed. The PowerShell runner is gone. Local and CI now use a single Gradle task: `./gradlew :grails-benchmarks:appIndyBench`. Nested Gradle is started via the wrapper JAR (`org.gradle.wrapper.GradleWrapperMain`), not `gradlew` / `gradlew.bat` / pwsh.

grails-benchmarks/scripts/run-app-indy-bench.ps1
```diff
@@ -0,0 +1,118 @@
...
+        [hashtable]$App
+    )
+    $label = if ($Indy -eq 'true') { 'indy' } else { 'noindy' }
+    $outFile = Join-Path $outDir "$($App.OutName)-$label.json"
+    if (Test-Path -LiteralPath $outFile) {
+        Remove-Item -LiteralPath $outFile -Force
+    }
+
+    Write-Host "=== $($App.Name) grailsIndy=$Indy -> $outFile ==="
+    & "$root\gradlew.bat" 'clean' "$($App.Project):integrationTest" `
```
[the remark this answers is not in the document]
jamesfredley: Obsolete. `run-app-indy-bench.ps1` has been deleted. The orchestrator is `:grails-benchmarks:appIndyBench` and does not call `gradlew.bat`.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,221 @@
...
+     * Warm up, sample, and append one JMH-shaped benchmark entry to {@code out}.
+     *
+     * @param benchmark fully-qualified-style name, e.g. {@code appbench.latency.FastPing.httpGet}
+     * @param request closure that performs one successful request
+     */
+    static void measureAndWrite(String benchmark, Path out, Closure<?> request) {
+        int warmup = warmupCount()
+        int samples = sampleCount()
+        int forks = Math.max(1, forkCount())
+
```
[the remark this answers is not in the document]
jamesfredley: Fixed. `measureAndWrite` now fails fast when `samples < 1`.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,221 @@
...
+            for (int i = offset; i < end; i++) {
+                chunk.add(values[i])
+            }
+            if (chunk.isEmpty() && !rawData.isEmpty()) {
+                chunk.addAll(rawData.get(rawData.size() - 1))
+            }
+            rawData.add(chunk)
+            offset = end
+        }
+
```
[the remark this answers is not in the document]
jamesfredley: Fixed. `toJmhEntry` rejects an empty `values` array and clamps `forkCount` to `values.length` so rawData chunks are never synthesized by copying the last fork.

gradle/app-bench-config.gradle
```diff
@@ -0,0 +1,43 @@
...
+    test.systemProperty('app.bench.forks', appBenchForks.get())
+    if (appBenchOut.get()) {
+        test.systemProperty('app.bench.out', appBenchOut.get())
+        test.outputs.file(appBenchOut.get())
+    }
+    if (appBenchEnabled.get() == 'true') {
+        test.outputs.upToDateWhen { false }
+        test.outputs.cacheIf { false }
+    }
+}
```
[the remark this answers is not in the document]
jamesfredley: Fixed. `appBench` is parsed with `Boolean.parseBoolean` once and that boolean drives both the `app.bench` system property and the cache-disable path, so `TRUE`/`true` behave the same.

jdaugherty:
I asked Scott to fix this in his other PR, but 'grails-benchmarks' should either be under grails-test-examples *or* it should be considered a test project in the root build.gradle where we set testProjectsStartWith variable. I think Scott just moved it in his PR.  FYI: his PR expands on this significantly.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    static int run(String[] args, GradleRunner runner, CommentPoster poster, Map<String, String> environment) {
+        try {
+            Options options = parse(args)
+            Path noindyDir = recreateDirectory(options.outputDir.resolve('noindy'))
+            Path indyDir = recreateDirectory(options.outputDir.resolve('indy'))
+
+            ['false', 'true'].each { String indy ->
+                Path modeDir = indy == 'true' ? indyDir : noindyDir
```
jdaugherty: The A/B always runs all `noindy` builds first and all `indy` builds second on the same runner, so first-vs-second runner-state drift (thermal throttle, page/dependency caches, CPU frequency) lands systematically on the indy side. The JMH job in this same workflow alternates shard order specifically to cancel this bias (see the header comments in benchmark.yml). Over the ~hour between the paired measurements, monotonic drift shows up in the report as an indy regression/improvement with no cancellation mechanism. Consider alternating mode order per app or interleaving repeats so ordering bias cancels.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    static Map<String, Object> toJmhEntry(String benchmark, double[] values, int forks) {
+        if (values == null || values.length == 0) {
+            throw new IllegalArgumentException('values must contain at least one sample')
+        }
+        double mean = mean(values)
+        double stdev = stdev(values, mean)
+        double scoreError = stdev * 1.96d / Math.sqrt((double) values.length)
+
```
jdaugherty, line 128: `stdev * 1.96 / sqrt(n)` is a 95% standard-error band over N=300–1000 sequential request timings from one JVM — samples that are strongly autocorrelated (GC pauses, JIT phases, connection reuse). That is far tighter than the JMH iteration-mean confidence intervals `BenchmarkComparator` was designed around: with scoreError around 0.1% of the mean, the CI-disjointness test in the comparator is effectively always satisfied, so any ≥10% drift (realistic given the fixed noindy→indy run order) produces a confident REGRESSED/IMPROVED verdict. The app path also emits no ruler benchmarks (`RULER_PACKAGE` never matches `appbench.*`), so the report prints 'Runner health: not measured' and none of the instability gates can downgrade a verdict. Consider computing the error across batch/fork means (t-interval) the way JMH does, and emitting a ruler-equivalent so the health gates apply.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,223 @@
...
+        entry.put('secondaryMetrics', Collections.emptyMap())
+        return entry
+    }
+
+    static void appendEntry(Path out, Map<String, Object> entry) {
+        List<Object> entries = new ArrayList<>()
+        if (Files.exists(out)) {
+            String existing = Files.readString(out, StandardCharsets.UTF_8).trim()
+            if (existing.startsWith('[')) {
+                Object parsed = new groovy.json.JsonSlurper().parseText(existing)
```
jdaugherty: Three issues in the existing-file path: content that doesn't start with `[` is silently discarded and overwritten; a truncated array (JVM killed mid-write) throws an uncaught JsonException; and in the documented manual flow (`-PappBench=true` with the default out file, `upToDateWhen { false }` forcing reruns) each rerun appends another entry with the same benchmark name — `JmhResults.parseEntries` keys by identity, so only the last entry wins and earlier runs are silently ignored while the file looks pooled. Only the orchestrator's `recreateDirectory` avoids this. Suggest replacing the file rather than appending, or failing loudly on unexpected existing content.

gradle/app-bench-config.gradle
```diff
@@ -0,0 +1,43 @@
...
+
+tasks.withType(Test).configureEach { Test test ->
+    test.systemProperty('app.bench', String.valueOf(appBenchEnabled.get()))
+    test.systemProperty('app.bench.warmup', appBenchWarmup.get())
+    test.systemProperty('app.bench.samples', appBenchSamples.get())
+    test.systemProperty('app.bench.forks', appBenchForks.get())
+    if (appBenchOut.get()) {
+        test.systemProperty('app.bench.out', appBenchOut.get())
+        test.outputs.file(appBenchOut.get())
+    }
```
jdaugherty: This registers the single `-PappBenchOut` file as a declared output of every `Test` task (unit `test` and `integrationTest`) in all three projects. With `org.gradle.parallel=true` (gradle.properties), the header's documented manual invocation from the root runs the three apps' tasks concurrently: `appendEntry`'s unlocked read-parse-rewrite drops entries (last writer wins) and Gradle flags multiple tasks sharing one declared output (overlapping-outputs warnings, disabled caching). The `systemProperty` wiring above also runs unconditionally, so `app.bench.*` values become inputs of ordinary test runs and any `-PappBench*` change invalidates unrelated tests' up-to-date state. Suggest guarding the whole block behind `appBenchEnabled` and wiring only `integrationTest`.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,223 @@
...
+ *   <li>split the series into {@code forks} equal chunks to mimic JMH multi-fork rawData shape</li>
+ *   <li>report mean ns/op with a simple standard-error-based scoreError</li>
+ * </ul>
+ *
+ * <p>Enable gated specs with {@code -PappBench=true}. Optional properties:
+ * {@code appBenchWarmup}, {@code appBenchSamples}, {@code appBenchForks}, {@code appBenchOut}.
+ */
+@CompileStatic
+final class AppHttpBench {
+
```
jdaugherty, line 50: This module is published, so `AppHttpBench` — including the `app.bench.*` system-property contract and the JMH-shaped JSON emitter — ships as public API to every consumer of grails-testing-support-http-client; renaming properties or fixing the statistics later becomes a breaking change. grails-benchmarks deliberately never publishes its bench tooling, and grails-geb uses testFixtures for shared test machinery. Suggest moving this to a testFixtures source set or the unpublished benchmarks module — or, if it's intentionally public, it needs grails-doc coverage.

.github/workflows/benchmark.yml
```diff
@@ -264,6 +264,40 @@ jobs:
...
+      - name: "Setup Gradle"
+        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        with:
+          cache-provider: basic # 'basic' uses the MIT-licensed, open-source cache provider; the default 'enhanced' provider (v6+) is proprietary (Gradle commercial Terms of Use)
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+      - name: "Verify JMH comparison tool"
+        run: ./gradlew :grails-benchmarks:test --max-workers=4
+      - name: "Run app-level indy A/B benches"
+        timeout-minutes: 180
+        run: ./gradlew --no-daemon :grails-benchmarks:appIndyBench -PappBenchWarmup=80 -PappBenchSamples=300 -PappBenchForks=2 -PappBenchOutDir="$RESULT_DIR" --max-workers=4
```
jdaugherty, line 296: The workflow header says results are advisory and a detected regression never fails the workflow, and the JMH compare step runs with `if: always()` + `continue-on-error: true` plus a fallback summary. Here, any one of the six nested builds failing makes `AppIndyBench` exit 2 before any comparison or summary is produced — a flaky nested integrationTest turns the job red with zero diagnostics in the step summary (only the artifact upload is `if: always()`). Consider `continue-on-error` on this step, and/or having AppIndyBench compare whatever results exist and emit a fallback summary.

.github/workflows/benchmark.yml
```diff
@@ -264,6 +264,40 @@ jobs:
...
+        with:
+          distribution: liberica
+          java-version: 21
+      - name: "Setup Gradle"
+        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        with:
+          cache-provider: basic # 'basic' uses the MIT-licensed, open-source cache provider; the default 'enhanced' provider (v6+) is proprietary (Gradle commercial Terms of Use)
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+      - name: "Verify JMH comparison tool"
+        run: ./gradlew :grails-benchmarks:test --max-workers=4
```
jdaugherty, line 289: This runs without `--no-daemon`, so a persistent daemon (`org.gradle.jvmargs=-Xmx5G`, 3h idle timeout) stays resident on the runner throughout the measurement step. The `benchmark` job has an explicit 'Stop Gradle daemons before measurement' step for exactly this reason; during sampling this runner additionally hosts the outer single-use daemon, the nested build's daemon, and the forked Spring Boot test JVM — memory pressure and daemon GC noise skew the latency deltas the job exists to measure. Suggest adding the same daemon-stop step before the bench run.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+                APPS.each { App app ->
+                    Path out = modeDir.resolve(app.name + '.json')
+                    runner.run(options.projectDir, gradleArgs(options, app, indy, out))
+                    if (!Files.isRegularFile(out)) {
+                        throw new IllegalStateException("Missing result file: ${out}")
+                    }
```
jdaugherty, line 73: When the gated spec is silently skipped, the nested build exits 0 and this bare 'Missing result file' points nowhere near the cause. That happens with any inherited `skipTests`/`skipFunctionalTests`/`onlyCoreTests` property — the functional-test-config `onlyIf` is `hasProperty`-based, so even `-PskipTests=false` or a value in `~/.gradle/gradle.properties` disables the Test task — or if the `-PappBench=true` → `app.bench` sysprop chain ever breaks the `@IgnoreIf` gate. Suggest naming these likely causes in the exception message, or detecting that zero tests ran.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+        void run(Path projectDir, List<String> args) {
+            Path javaHome = Path.of(System.getProperty('java.home'))
+            List<String> command = commandLine(javaHome, projectDir, args)
+            ProcessBuilder processBuilder = new ProcessBuilder(command)
+            processBuilder.directory(projectDir.toFile())
+            processBuilder.inheritIO()
+            processBuilder.environment().put('JAVA_HOME', javaHome.toString())
+            Process process = processBuilder.start()
+            int exit = process.waitFor()
+            if (exit != 0) {
```
jdaugherty, line 267: `waitFor()` has no timeout, and the nested build runs in the same checkout and GRADLE_USER_HOME as the still-executing outer build. A wedged nested build (cache-lock contention with the live outer daemon, or a hung test JVM) hangs `appIndyBench` forever locally — CI is only saved by `timeout-minutes: 180`. Suggest `waitFor(timeout, unit)` + `destroyForcibly()`, and consider a separate `--project-cache-dir` for the nested builds so they don't contend for the outer build's execution-history/file-hash locks (the JMH job sidesteps this entirely by measuring in separate worktrees).

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+            }
+            if (!Files.isRegularFile(wrapperJar)) {
+                throw new IllegalStateException("Gradle wrapper jar not found: ${wrapperJar}")
+            }
+            List<String> command = new ArrayList<>()
+            command.add(java.toString())
+            command.add('-cp')
+            command.add(wrapperJar.toString())
+            command.add('org.gradle.wrapper.GradleWrapperMain')
+            command.addAll(args)
```
jdaugherty, line 293: Invoking `GradleWrapperMain` directly (rather than the gradlew script) means `DEFAULT_JVM_OPTS`/`JAVA_OPTS`/`GRADLE_OPTS` are never spliced into the launcher JVM — the gradlew script is what does that — so environment-based configuration is silently ignored for all six nested builds. Our own OOM guidance is `export GRADLE_OPTS="-Xms2G -Xmx5G"`, and proxy settings are commonly passed the same way; nested builds will behave differently from the documented `./gradlew` invocation with no hint the env var was dropped (build-JVM heap currently survives only because gradle.properties pins `org.gradle.jvmargs`). If avoiding the shell scripts is the goal, consider forwarding those env vars onto the wrapper JVM explicitly.

grails-benchmarks/src/reportTest/groovy/org/apache/grails/benchmarks/report/AppIndyBenchSpec.groovy
```diff
@@ -0,0 +1,225 @@
...
+                [GITHUB_STEP_SUMMARY: summary.toString()]
+        )
+
+        then:
+        exit == 0
+        invocations.size() == 6
+        invocations[0].contains(':grails-test-examples-latency:integrationTest')
+        invocations[0].contains('latencyapp.AppBenchFastPingSpec')
+        invocations[0].contains('-PgrailsIndy=false')
+        invocations[0].contains('-PappBench=true')
```
jdaugherty, line 88: The suite never correlates each invocation's `-PgrailsIndy` value with the mode directory in its `-PappBenchOut` path, and both modes get identical dummy scores — so an inverted `modeDir` ternary (or swapped `--base`/`--head`) would pass every test while flipping the sign of every production verdict. Suggest asserting that invocations carrying `-PgrailsIndy=false` write under `/noindy/` (and `true` under `/indy/`), or giving the two modes distinguishable scores and asserting the verdict direction.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
```diff
@@ -0,0 +1,223 @@
...
+    private AppHttpBench() {
+    }
+
+    static boolean enabled() {
+        Boolean.getBoolean('app.bench') || Boolean.parseBoolean(System.getProperty('appBench', 'false'))
+    }
+
+    static int warmupCount() {
+        Integer.getInteger('app.bench.warmup', Integer.getInteger('appBenchWarmup', 200))
+    }
```
jdaugherty, line 62: `Integer.getInteger` silently returns the default on an unparsable value and decodes leading-zero values as octal, and app-bench-config forwards the raw `-PappBench*` strings without validation. So `-PappBenchSamples=1O00` (letter O) silently benches with the default 1000, and `-PappBenchSamples=010` benches with 8 samples — no warning either way. The orchestrator path validates with `parseInt`; the documented direct-Gradle path should too (either in these getters or in app-bench-config.gradle).

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+                Path.of(summary),
+                Files.readString(report, StandardCharsets.UTF_8),
+                StandardCharsets.UTF_8,
+                StandardOpenOption.CREATE,
+                StandardOpenOption.APPEND
+        )
+    }
+
+    static Path recreateDirectory(Path directory) {
+        if (Files.exists(directory)) {
```
jdaugherty, line 193: This recursively deletes `<outputDir>/noindy` and `<outputDir>/indy` for whatever directory the user supplies via `--output-dir`/`-PappBenchOutDir`, with no containment or sanity check — pointing it at a directory that happens to contain unrelated `noindy`/`indy` subtrees destroys them before benching starts. Meanwhile a stale `indy-vs-noindy.md` in outputDir itself is never cleaned, so a rerun that fails before JmhCompare leaves yesterday's report sitting next to today's JSON. (Also, GDK `deleteDir()` already implements the walk + reverse-sort deletion — see PublishGuideTask in build-logic.)

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+            Options options = parse(args)
+            Path noindyDir = recreateDirectory(options.outputDir.resolve('noindy'))
+            Path indyDir = recreateDirectory(options.outputDir.resolve('indy'))
+
+            ['false', 'true'].each { String indy ->
+                Path modeDir = indy == 'true' ? indyDir : noindyDir
+                APPS.each { App app ->
+                    Path out = modeDir.resolve(app.name + '.json')
+                    runner.run(options.projectDir, gradleArgs(options, app, indy, out))
+                    if (!Files.isRegularFile(out)) {
```
jdaugherty, line 70: This is six sequential `--no-daemon` nested builds (app × mode), each paying full configuration of the 63-project build (functional-test-config forces `evaluationDependsOn` across subprojects) — hence the 180-minute CI timeout. Two builds — one per mode, passing all three task paths with per-task `--tests` filters — would do the same work at roughly a third of the cold-start/configuration cost. The single-file `-PappBenchOut` routing is the only blocker; per-app default out files (which `AppHttpBench.outputPath` already produces) would absorb it. A combined invocation should cap workers so the three integrationTests don't contaminate each other's latency samples.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
```diff
@@ -0,0 +1,244 @@
...
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.StandardOpenOption
+
+@CompileStatic
+class AppIndyBench {
+
+    static final List<App> APPS = [
+            new App('latency', ':grails-test-examples-latency:integrationTest', 'latencyapp.AppBenchFastPingSpec'),
```
jdaugherty, line 33: This roster duplicates knowledge the example projects own (project path, spec FQCN) into a compiled class in another module, and each benched app must separately remember to apply `gradle/app-bench-config.gradle` — adding or renaming an app takes lockstep edits in three places with no drift detection, and a miss surfaces only as the late 'Missing result file' failure deep in a nested-build run. The specs already follow an `AppBench*` naming convention and every example app applies functional-test-config.gradle; discovery by convention (or folding the ~10 lines of sysprop wiring into functional-test-config.gradle) would keep the roster where the apps live.

jdaugherty (commented):
The following review was AI generated for more feedback:

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Modes now interleave per app (even: noindy then indy, odd: reverse), so first-vs-second drift is not always on the indy side.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
jamesfredley: Implemented in c17dc649c3. scoreError is a 95% t-interval over fork means; JSON also emits org.apache.grails.benchmarks.ruler.AppBenchCpu.measure so runner-health gates apply.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
jamesfredley: Implemented in c17dc649c3. measureAndWrite replaces the output file. Truncated JSON and non-array existing content fail loudly.

gradle/app-bench-config.gradle
jamesfredley: Implemented in c17dc649c3. app-bench-config now wires only integrationTest and only when -PappBench=true, so ordinary tests do not share AppBenchOut or pick up app.bench.* inputs.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
jamesfredley: Implemented in c17dc649c3. AppHttpBench moved to grails-testing-support-http-client test fixtures; example apps depend on testFixtures(project(':grails-testing-support-http-client')).

.github/workflows/benchmark.yml
jamesfredley: Implemented in c17dc649c3. Nested failures no longer abort before compare. We compare whatever JSON exists, write a fallback GITHUB_STEP_SUMMARY, and the workflow measure step is continue-on-error so artifacts still upload.

.github/workflows/benchmark.yml
jamesfredley: Implemented in c17dc649c3. Added a Stop Gradle daemons step before app-bench measurement, matching the JMH job.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Missing-result message now names skipTests, skipFunctionalTests, onlyCoreTests, ignored gated spec, and --tests mismatch.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Nested Gradle waitFor now times out after 90 minutes and destroyForcibly; each nested build uses a unique --project-cache-dir under the owned run directory.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. DEFAULT_JVM_OPTS, JAVA_OPTS, and GRADLE_OPTS are forwarded onto the wrapper JVM before -cp.

grails-benchmarks/src/reportTest/groovy/org/apache/grails/benchmarks/report/AppIndyBenchSpec.groovy
jamesfredley: Implemented in c17dc649c3. Specs now assert grailsIndy=false writes under run/noindy and true under run/indy, with distinguishable dummy scores so an inverted ternary fails.

grails-testing-support-http-client/src/testFixtures/groovy/org/apache/grails/testing/http/client/bench/AppHttpBench.groovy
jamesfredley: Implemented in c17dc649c3. Integer.getInteger is gone; properties parse as decimal ints and reject unparsable/octal-looking values.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Mode trees and the report live under an owned outputDir/run/ directory that we fully recreate. Caller-owned noindy/indy siblings at the output root are left alone. Stale indy-vs-noindy.md in run/ is removed.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Left as six nested builds for this pass so each integrationTest keeps its own AppBenchOut and worker cap; combining three apps in one Gradle invocation is a follow-up because it needs per-task output routing to avoid sample contamination.

grails-benchmarks/src/report/groovy/org/apache/grails/benchmarks/report/AppIndyBench.groovy
jamesfredley: Implemented in c17dc649c3. Roster stays explicit for now so the orchestrator fails fast on a missing named spec. Convention discovery (AppBench* plus folding wiring into functional-test-config) is a follow-up.

jamesfredley:
@jdaugherty Implemented in c17dc649c3. grails-benchmarks is now in root testProjectsStartWith next to grails-test-suite and grails-test-examples, so it is classified as a test project rather than a published module.

jdaugherty:
FYI: Scott's PR was merged prior to this, so we'll want to move this to grails-test-examples/benchmarks too

PR #16170, opened 2026-08-18
feat!: make callsite caching the default and publish invokedynamic under a classifier

codeconsole:
Rebased onto latest `8.0.x` (`d0c68dc`, which brings Groovy 5.1.0) and reworked to the design the Grails team asked for: **the main artifact is the callsite-caching flavour, and `invokedynamic` ships under the `indy` classifier.**

### What an application sees

Nothing, unless it opts in. `grails.indy` defaults to `false`, exactly as 7.x behaved, and every Grails module and plugin publishes its main artifact compiled that way.

An application that wants `invokedynamic` gets it for its own classes *and* its dependencies:

```groovy
grails {
    indy = true
    indyModules = ['org.apache.grails:grails-core', 'com.example:my-plugin']
}
```

Dependencies left off the list keep their main artifact. That is safe on the JVM — the two flavours interoperate on the same classpath — but a native image needs the `indy` flavour of everything, so an application targeting one lists every Grails module and plugin it uses.

### Plugins publish both

A project applying `org.apache.grails.gradle.grails-plugin` builds both flavours through the new `indyJar` task, so any application that wants indy can have it. `grails.indy` no longer affects how a plugin compiles its own sources; plugin authors need to change nothing, and a plugin build that sets `indy` can drop it.

### Why a classifier and not a variant

The previous revision of this PR published the `indy` jar as a secondary Gradle variant, and CI failed across the board:

```
Could not resolve all dependencies for configuration ':grails-data-hibernate5-core:tck'.
   > More than one variant of project ':grails-datamapping-tck' matches the consumer attributes:
```

A configuration that requests **no attributes at all** — the shape `tck` and the CLI companion probe use, here and in other people's builds — finds every variant equally good. No attribute rule repairs it: a disambiguation rule only receives the values candidates declare, so the variant *without* the attribute never appears among them (`consumer=null candidates=[false]`), and declaring it on both variants breaks every plain consumer instead.

A plain Maven classifier adds nothing to the published metadata, so the ambiguity is unreachable rather than fixed. Selection is assembled on the consuming side: an application that opts in registers a `ComponentMetadataRule` deriving a variant from the classifier for the modules it lists. That variant exists only inside the opted-in build, whose schema also carries the rules needed to choose.

### Verification

10 functional tests in `GrailsIndyVariantsSpec`, including the two shapes that broke CI last time — an artifact view constraining nothing, and a configuration with no attributes — both now resolving a single artifact.

On `grails-cache`, a real plugin module with an `ast` source set:

| | entries | classes | with `invokedynamic` | AST classes |
|---|---|---|---|---|
| main | 56 | 37 | 1 | 5 |
| `-indy` | 56 | 37 | 11 | 5 |

Entry lists identical; only the Groovy call-site bytecode differs. (The 1 in the main jar is `makeConcatWithConstants` in a `.java` source file — javac's string concatenation, not a Groovy call site.)

Suite: 188/191. The 3 `GroovyPageToolchainSpec` failures need a JDK 21 toolchain that is not installed on the machine I ran them on and fail identically on unmodified `8.0.x`. `codeStyle` clean.

### Known limitations

- **`indyModules` must be listed explicitly.** An unguarded rule fails hard trying to fetch e.g. `groovy-5.1.0-indy.jar`, and platforms fail the same way — `grails-bom` is a platform every application depends on — so there is no safe group-wide default. Modules advertise their classifier through `Grails-Indy-Artifact` (mirroring `Grails-Cli-Artifact`) so the list can later be discovered by a probe rather than written; that discovery is not implemented here.
- **Component metadata rules do not apply to project dependencies**, so a composite or multi-project build always resolves the main artifact.

### Reviewer notes

Full `./gradlew build` has not been run; `grails-cache` was built and inspected locally. The `indyJar` task assembles by unpacking the finished main jar rather than listing source directories — enumerating them omitted the AST classes and staged command/template resources, and reusing the main jar's copy spec imported its `duplicatesStrategy`. Both were caught only by building a real module, not by the functional tests.

### DIFF
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
@@ -55,6 +55,7 @@ class CompilePlugin implements Plugin<Project> {
         configureJars(project)
         configureCompiler(project)
         configureReproducible(project)
+        IndyVariants.configure(project)
     }
 
     private static void configureJavaVersion(Project project) {
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/CompilePlugin.groovy
@@ -84,8 +84,12 @@ class CompilePlugin implements Plugin<Project> {
                     'Implementation-Version': lookupPropertyByType(project, 'grailsVersion', String),
                     'Implementation-Vendor': 'grails.apache.org'
             )
-            // Explicitly fail since duplicates indicate a double configuration that needs fixed
-            jar.duplicatesStrategy = DuplicatesStrategy.FAIL
+            if (jar.name != IndyVariants.INDY_JAR_TASK_NAME) {
+                // Explicitly fail since duplicates indicate a double configuration that needs fixed.
+                // The indy jar is the exception: it deliberately layers its own Groovy classes over
+                // a copy of the main jar and drops the duplicates that produces.
+                jar.duplicatesStrategy = DuplicatesStrategy.FAIL
+            }
         }
     }
 
@@ -104,6 +108,11 @@ class CompilePlugin implements Plugin<Project> {
 
         project.plugins.withId('groovy') {
             project.tasks.withType(GroovyCompile).configureEach {
+                if (it.name != IndyVariants.INDY_COMPILE_TASK_NAME) {
+                    // The main artifact of every published module is the callsite-caching flavour;
+                    // IndyVariants supplies the invokedynamic one under a classifier.
+                    it.groovyOptions.optimizationOptions.put('indy', false)
+                }
                 // encoding needs to be the same since it's different across platforms
                 it.groovyOptions.encoding = StandardCharsets.UTF_8.name()
                 // Preserve method parameter names in Groovy/Java classes for IDE parameter hints & bean reflection metadata.
--- a/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/IndyVariants.groovy
+++ b/build-logic/plugins/src/main/groovy/org/apache/grails/buildsrc/IndyVariants.groovy
@@ -23,6 +23,9 @@ import groovy.transform.CompileStatic
 
 import org.gradle.api.Project
 import org.gradle.api.file.Directory
+import org.gradle.api.file.CopySpec
+import org.gradle.api.file.DuplicatesStrategy
+import org.gradle.api.file.RegularFile
 import org.gradle.api.plugins.BasePlugin
 import org.gradle.api.plugins.JavaPluginExtension
 import org.gradle.api.publish.PublishingExtension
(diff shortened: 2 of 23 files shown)


codeconsole:
Rebuilt on a different mechanism after CI failed. The original approach was not fixable.

### Why it failed

Publishing the `noindy` jar as a secondary variant of `apiElements`/`runtimeElements` made every module ambiguous to a configuration that requests **no attributes at all**:

```
Could not resolve all dependencies for configuration ':grails-data-hibernate5-core:tck'.
   > More than one variant of project ':grails-datamapping-tck' matches the consumer attributes:
       - Configuration ':grails-datamapping-tck:runtimeElements':
       - Configuration ':grails-datamapping-tck:runtimeElements' variant noindy:
```

Every unmatched line reads *"but the consumer didn't ask for it"*, including `org.gradle.usage`. The `tck` and CLI companion probe configurations declare nothing, so every variant is an equally good candidate.

No attribute rule repairs this. A disambiguation rule receives only the values candidates declare, so the variant *without* the attribute never appears among them — instrumenting the rule prints `RULE_INVOKED consumer=null candidates=[false]`, leaving no value meaning "the default one". Declaring the attribute on both variants makes the rule work but breaks every plain consumer of a Grails artifact, which is the problem that led to the asymmetry in the first place.

Worth noting the one existing precedent here, `GrailsExplodedPlugin`, creates its secondary variant only `if (grails.developmentRun)` — it exists during `bootRun` and never during normal resolution, sidestepping the general case rather than solving it.

### What replaces it

The `noindy` jar is now a plain Maven classifier that adds **nothing** to the published metadata — the module keeps exactly the variants it always had, so no consumer's resolution changes and the ambiguity is unreachable rather than fixed.

Selection is assembled on the consuming side. An application setting `indy = false` registers a `ComponentMetadataRule` that derives a variant from the classifier for the modules it names:

```groovy
grails {
    indy = false
    noindyModules = ['org.apache.grails:grails-core', 'com.example:my-plugin']
}
```

That variant exists only inside the opted-in build, whose schema also carries the disambiguation rule needed to choose between it and the main artifact.

Verified across five scenarios, including the two that broke CI: an artifact view constraining nothing, and an ad-hoc configuration with no attributes. Both now resolve a single artifact. A dependency publishing only the main artifact still resolves under either setting.

### Two limitations, both inherent

- **The module list is required.** Applying the rule unguarded fails hard — it tries to fetch `groovy-5.0.8-noindy.jar`. Platforms fail the same way, and the Grails BOM is a platform every app depends on, so there is no safe group-wide default. Modules advertise their classifier through a `Grails-Noindy-Artifact` manifest attribute (mirroring `Grails-Cli-Artifact`) so the list can later be discovered by a probe rather than written by hand; that discovery is not implemented.
- **Component metadata rules do not apply to project dependencies**, so a composite or multi-project build always resolves the main artifact. Documented in the guide.

### Standing recommendation unchanged

This makes the approach viable, not necessarily right. `noindy` artifacts still cannot appear in a native image, which was the motivation, and #16171 delivers the same native guarantee in 73 lines with no module list to maintain and green CI. I would still merge #16171 — but the choice is now a real trade-off rather than one option being unworkable.

jamesfredley:
@codeconsole https://github.com/apache/grails-core/pull/16178 and this PR have a lot of overlap for Grails 8 on Groovy 5.  Since Groovy 5 Indy is not fast enough in Grails I am not sure we need to publish the extra modules.

Then on Groovy 6, we will flip everything to Indy, which will be the first time Grails is on Indy, with the exception of the narrow window at the beginning of Grails 7.

PR #16174, opened 2026-08-19
fix(cas): repair proxy receptor and single sign-out, add integration tests

sbglasius:
## Summary

`grails-spring-security-cas` had no tests, and its only exerciser was a `bootRun`-only demo pointed at a hand-run CAS server. Two defects were living in that gap. This fixes both, makes a security-relevant default explicit, and puts the plugin under integration tests that run against a real Apereo CAS server.

## Fixes

**`proxyReceptorUrl` was set unconditionally.** Spring Security 7.1 rejects a null pattern, so an application that did not configure a proxy receptor — the default — failed to start with `IllegalArgumentException: pattern cannot be null`. It is now only set when configured, which is how `CasAuthenticationFilter` expresses that proxy support is off. Before Spring Security 7.1 the unguarded assignment produced a matcher for the literal path `/**null`, quietly giving unconfigured applications a live proxy receptor.

**Single sign-out never worked.** The plugin set `useSessionFixationPrevention = false` from `doWithSpring`, but it declares `loadAfter = ['springSecurityCore']`, so the core plugin had already defined `sessionAuthenticationStrategy` from the original value. The session was still replaced on login, so CAS logout requests matched no session and were silently dropped. The bean is now redefined from the CAS plugin as well, through the same `BeanTypeResolver` the core plugin uses, so an application's `sessionAuthenticationStrategyBeanClass` override still wins.

Overriding a core bean from a sibling security plugin is the established pattern in this module family — CAS already overrides `authenticationEntryPoint`, LDAP overrides `userDetailsService`, ACL overrides `passwordEncoder`/`permissionEvaluator`/`runAsManager`.

## Behaviour change

`cas.useSingleSignout` now defaults to `false`. Enabling it disables session fixation prevention, which an application should choose deliberately rather than inherit. The plugin logs and prints a warning at startup when it is enabled.

Because single sign-out did not actually work before, applications upgrading are unlikely to lose working behaviour. Documented in the CAS configuration reference and in a new section of the Grails 8 upgrade guide.

## Tests

`grails-test-examples-spring-security-cas-test1` now runs against a real `apereo/cas` server started with Testcontainers, gated on Docker availability, and no longer needs an externally managed CAS server for `bootRun` either.

An `EnvironmentPostProcessor` starts the container and supplies the CAS URLs before the context is built. The service URL and proxy callback URL depend on the port the embedded server binds, so they are set once the server is up but before it serves a request.

Three configurations run via the existing `TESTCONFIG` idiom, all added to CI:

| `TESTCONFIG` | Configuration |
|---|---|
| `cas` | proxy settings unset; login, validation, roles, single sign-out |
| `casProxy` | full proxy-granting-ticket round trip |
| `casNoSingleSignout` | asserts the new default |

Proxy coverage obtains a real `PT-` proxy ticket through `AttributePrincipal.getProxyTicketFor(...)`, so it exercises the whole callback round trip rather than inspecting internals. Logout requests are posted from a cookie-less client, as CAS does on its own connection — posting from the authenticated client would clear the session through failed authentication anyway, which is indistinguishable from single sign-out working.

Verified that reverting either fix turns the relevant specs red.

## Notes for reviewers

Two findings worth knowing, both documented in the code:

- Do **not** add `attributeReleasePolicy.authorizedToReleaseProxyGrantingTicket` to the CAS service definition. Despite the name it is a different feature, it needs a service public key, and without one CAS drops `<cas:proxyGrantingTicket>` from the validation response entirely, silently breaking proxy authentication.
- The stock `apereo/cas` image only authorises `https` services, so a service definition permitting `http` is copied into the container.

This branches from and targets `8.0.x` because the default change is documented in the Grails 8 upgrade guide. The two bug fixes are independently backportable to `7.0.x` if wanted.

Also included: `Grails BOM Hibernate7 Micronaut.adoc` is generated and gitignored but was missing from the rat exclusions, so `./gradlew rat` failed after any docs build. One line, unrelated to CAS — happy to split it out.


grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
```diff
@@ -87,10 +91,30 @@ class SpringSecurityCasGrailsPlugin extends Plugin {
...
+                // Session fixation prevention breaks single signout because the service ticket is
+                // mapped to the session id, which changes when the session is replaced on login.
+                // Disabling it is a security trade-off the application has opted into, so say so.
+                String message = '''
+    WARNING: cas.useSingleSignout is enabled, so session fixation prevention has been disabled.
+    CAS maps the service ticket to the HTTP session id, and a logout request cannot be matched to a
+    session that was replaced when the user authenticated. Set cas.useSingleSignout to false to keep
+    session fixation prevention and handle logout in the application instead.
+    '''
+                println message
```
bkoehm, line 101: Minor: Is the `println` necessary on top of the `log.warn`?

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/specs/AbstractCasSpec.groovy
```diff
@@ -0,0 +1,182 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package specs
```
bkoehm: Is this a package typically used for tests?

grails-test-examples/spring-security/cas/test1/src/main/groovy/spring/security/cas/test/CasTestConfig.groovy
```diff
@@ -0,0 +1,72 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package spring.security.cas.test
```
bkoehm: Again, not sure if this is following typical Grails test package naming conventions or not.

bkoehm (approved):
Just added a few minor comments (questions) that don't hold up my approval.  I think adding real tests is definitely an improvement.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
sbglasius: Since Grails Spring Security uses `println` I have removed `log.warn`

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/specs/AbstractCasSpec.groovy
sbglasius: Aligned to be `grails.plugin.springsecurity.cas.test`

grails-test-examples/spring-security/cas/test1/src/main/groovy/spring/security/cas/test/CasTestConfig.groovy
sbglasius: Aligned to be `grails.plugin.springsecurity.cas.test`

grails-test-examples/spring-security/cas/README.md
```diff
@@ -16,19 +16,73 @@ limitations under the License.
 
 This is a CAS-enabled test application.  To run it successfully, a CAS
 server is required.  The URL for the CAS server is configured in the
-[application.groovy](test1/grails-app/conf/application.groovy)
-file.  Setting up a CAS server is out of the scope of this document, but
-good places to start are [Apereo CAS GitHub](https://github.com/apereo/cas)
-and the [CAS Initializr](https://getcas.apereo.org/ui) service.
+This is a CAS-enabled test application. It no longer needs a hand-run CAS server: an
```
matrei: The seems to be some leftover here...

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
```diff
@@ -35,11 +36,14 @@ import org.springframework.security.cas.authentication.CasAuthenticationProvider
...
+import org.springframework.security.web.authentication.session.NullAuthenticatedSessionStrategy
 
+import grails.plugin.springsecurity.BeanTypeResolver
 import grails.plugin.springsecurity.SecurityFilterPosition
 import grails.plugin.springsecurity.SpringSecurityUtils
 import grails.plugins.Plugin
 
+@Slf4j
 @CompileStatic
 class SpringSecurityCasGrailsPlugin extends Plugin {
```
matrei, line 48: While we are here:
- Should we bump `grailsVersion`?
- Jasig -> Apereo?
- organization metadata?
- issueManagement metadata?
- scm metadata?

.github/workflows/gradle.yml
```diff
@@ -567,6 +567,46 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        # The CAS test app authenticates against a real Apereo CAS server started via Testcontainers.
+        # 'cas' leaves the proxy settings unset and is the default, so it is also covered by the main
+        # security job; 'casProxy' exercises the proxy-granting-ticket callback; 'casNoSingleSignout'
+        # asserts the shipped default, where single signout is not enabled.
+        # TESTCONFIG is applied at app startup, so every config must be its own run.
+        # Carried over from the standalone grails-spring-security CI convention - add, don't remove.
+        test-config: [ 'cas', 'casProxy', 'casNoSingleSignout' ]
```
matrei, line 584: Will the 'cas' test-config will be run twice? Once in `security` job and then again here?

matrei (changes requested):

grails-test-examples/spring-security/cas/README.md
sbglasius: Good catch — fixed in e4a3e8e. The rewrite extracted the license header by line count and stopped two lines short, so the opening of the old text ("To run it successfully, a CAS server is required...") stayed above the new paragraph. Removed. Also corrected "two configurations" to three while I was in there, since `casNoSingleSignout` was added later.

.github/workflows/gradle.yml
sbglasius: Yes, and actually three times rather than twice. The specs defaulted to the cas configuration when `TESTCONFIG` was unset, so the security job picked it up in its Java 21 entry and one of its Java 25 shards, plus the run here.

Fixed in e4a3e8e by requiring a configuration to be selected, which is what the core functional-test-app specs already do — every one of them is gated on `TESTCONFIG`, so a run without it exercises none of them. The security job now skips all CAS specs (and no longer pulls the 882 MB CAS image at all), and each configuration runs exactly once here. `bootRun` still needs no `TESTCONFIG`.

One trade-off worth flagging: `:grails-test-examples-spring-security-cas-test1:check` without `-DTESTCONFIG` now reports no CAS tests. That matches the core app's behaviour and is called out in the `README`.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
sbglasius: * **Jasig → Apereo**: yes, clearly wrong — the plugin imports `org.apereo.cas.client.*`. Fixed the title and description, and repointed three dead `jasig.org` links in the guide at `apereo.github.io`.
* `grailsVersion` / `organization` / `issueManagement` / `scm`: updated to `'8.0.0-SNAPSHOT > *'`, `grails.apache.org/`, and `apache/grails-core`, matching what `MongodbGrailsPlugin`, `GormGraphqlGrailsPlugin` and `MarkupViewGrailsPlugin` use on this branch.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
```diff
@@ -35,25 +36,28 @@ import org.springframework.security.cas.authentication.CasAuthenticationProvider
...
 
-    String grailsVersion = '7.0.0 > *'
+    String grailsVersion = '8.0.0-SNAPSHOT > *'
     String author = 'Burt Beckwith'
```
jdaugherty: We should go ahead and remove the author here since we're updating the code and ASF policy is to no longer have that.

grails-doc/src/en/guide/security/securityPlugins/springSecurity/cas/usage.adoc
```diff
@@ -25,7 +25,7 @@ under the License.
 Configuring your CAS server is beyond the scope of this document. There are many different approaches and this will most likely be done by IT staff. It's assumed here that you already have a running CAS server.
 ====
 
-https://www.jasig.org/cas[CAS] is a popular single sign-on implementation. It's open source and has an Apache-like license, and is easy to get started with but is also highly configurable. In addition it has clients written in Java, .Net, PHP, Perl, and other languages.
+https://apereo.github.io/cas[CAS] is a popular single sign-on implementation. It's open source and has an Apache-like license, and is easy to get started with but is also highly configurable. In addition it has clients written in Java, .Net, PHP, Perl, and other languages.
```
jdaugherty, line 28: Not this line, but this file: the `=== Single Signout` section below (line 62) still says "Single signout is enabled by default" and only documents how to disable it. Since this PR flips `cas.useSingleSignout` to opt-in, that section needs updating to match the new default (off, enable explicitly).

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -2632,3 +2632,36 @@ used to apply to its own message source.
 
 Adding or removing a base name now needs a restart, because Spring Boot reads the configured base names
 once when it builds the message source.
+
+==== 46. CAS Single Sign-Out Is Opt-In
```
jdaugherty: Sections 46 (Startup Banner, line 2481) and 47 (Message Bundles, line 2517) already exist above, so this should be `==== 48.` to keep the numbering sequential.

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/AbstractCasSpec.groovy
```diff
@@ -0,0 +1,178 @@
...
+        URLEncoder.encode(value, 'UTF-8')
+    }
+
+    private static HttpClient newClient() {
+        CookieManager cookieManager = new CookieManager(null, CookiePolicy.ACCEPT_ALL)
+        HttpClient.newBuilder()
+                .cookieHandler(cookieManager)
+                .followRedirects(HttpClient.Redirect.NEVER)
+                .connectTimeout(Duration.ofSeconds(30))
+                .build()
```
jdaugherty, line 183: Only a connect timeout is set here — `client.send()` has no request timeout and none of the specs use `@Timeout`, so if the CAS container accepts the TCP connection but never responds (mid-startup stall, OOM), `get()`/`postForm()` block indefinitely and the CI matrix job hangs until the runner's job-level kill. Consider adding `.timeout(Duration.ofSeconds(...))` to the `HttpRequest` builders (or a Spock `@Timeout`) so a wedged container fails within minutes instead.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
```diff
@@ -87,10 +91,29 @@ class SpringSecurityCasGrailsPlugin extends Plugin {
...
+                // Session fixation prevention breaks single signout because the service ticket is
+                // mapped to the session id, which changes when the session is replaced on login.
+                // Disabling it is a security trade-off the application has opted into, so say so.
+                String message = '''
+    WARNING: cas.useSingleSignout is enabled, so session fixation prevention has been disabled.
+    CAS maps the service ticket to the HTTP session id, and a logout request cannot be matched to a
+    session that was replaced when the user authenticated. Set cas.useSingleSignout to false to keep
+    session fixation prevention and handle logout in the application instead.
+    '''
+                println message
```
jdaugherty, line 101: With `log.warn` removed, the `@Slf4j` annotation added above is now unused. Two options: drop the annotation, or keep both outputs like the core plugin does (`SpringSecurityCoreGrailsPlugin` pairs `println` with `log.warn` for its warnings). The docs added in this PR say the plugin "logs a warning at startup", and an app that captures logs but not stdout will never see a `println`, so pairing them seems preferable. Also, the line just below is whitespace-only and should be trimmed.

grails-test-examples/spring-security/cas/README.md
```diff
@@ -14,21 +14,77 @@ See the License for the specific language governing permissions and
...
+## Running the application
+
+```
+./gradlew :grails-test-examples-spring-security-cas-test1:bootRun
+```
 
 The test application URLs are:
 * [http://localhost:8081/secure/admins](http://localhost:8081/secure/admins)
 * [http://localhost:8081/secure/users](http://localhost:8081/secure/users)
+* [http://localhost:8081/secure/proxyStatus](http://localhost:8081/secure/proxyStatus) — asks CAS for a proxy ticket
```
jdaugherty, line 66: Under `bootRun`, `TESTCONFIG` is unset, so `CasTestEnvironmentPostProcessor` never configures `proxyReceptorUrl`/`proxyCallbackUrl` and this endpoint can only ever return `NO_PROXY_TICKET`. A developer following this section will conclude the proxy support is broken. Worth either noting that limitation here or documenting how to start `bootRun` with the proxy configuration so the endpoint can actually be exercised.

grails-test-examples/spring-security/cas/test1/grails-app/controllers/grails/plugin/springsecurity/cas/test/SecureController.groovy
```diff
@@ -0,0 +1,58 @@
...
+package grails.plugin.springsecurity.cas.test
+
+import grails.plugin.springsecurity.annotation.Secured
+import org.apereo.cas.client.authentication.AttributePrincipal
+import org.springframework.security.cas.authentication.CasAuthenticationToken
+import org.springframework.security.core.Authentication
+import org.springframework.security.core.context.SecurityContextHolder
+
+class SecureController {
+
```
jdaugherty, line 29: The body of this file is indented with tabs; `.editorconfig` sets `indent_style = space` (4 spaces) for Groovy sources. Since git treats this as a new file (the old controller path became `resources.groovy` in rename detection), it should come in with the standard indentation.

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/CasSingleSignOutSpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+package grails.plugin.springsecurity.cas.test
+
+import spock.lang.IgnoreIf
+
+import java.net.http.HttpResponse
+
+/**
+ * Covers single sign-out, which the plugin enables by default via {@code cas.useSingleSignout} by
```
jdaugherty: "which the plugin enables by default" is stale — this PR makes `cas.useSingleSignout` opt-in (default `false`), which is exactly what `CasNoSingleSignOutSpec` asserts. Suggest rewording to something like "which the app opts into via `cas.useSingleSignout`".

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/CasLoginSpec.groovy
```diff
@@ -0,0 +1,92 @@
...
+        response.statusCode() == 403 || !response.body().contains('Logged in with ROLE_ADMIN')
+    }
+
+    void 'bad credentials do not authenticate'() {
+        when:
+        HttpResponse<String> challenge = get(appClient, "${appBaseUrl}/secure/users")
+        String loginUrl = location(challenge)
+        HttpResponse<String> form = get(casClient, loginUrl)
+        String execution = form.body().find(/name="execution"\s+value="([^"]+)"/) { full, token -> token }
+        HttpResponse<String> submitted = postForm(casClient, loginUrl,
```
jdaugherty, line 86: This re-implements the execution-token extraction inline, duplicating `AbstractCasSpec.extractExecution` but dropping the null guard the other paths get. If a future `casContainerVersion` renders the login form with different attribute order or quoting, `execution` is null here and `postForm` fails with an opaque NPE from `URLEncoder.encode(null)` inside the when-block, instead of the clear missing-token assertion `authenticateAtCas` produces. Reuse the base-class extraction and guard.

matrei (approved):
One thing with the testconfigs is that tests will not run locally, if not explicitly configured. But that problem is also present with the other security testconfigs.

jdaugherty (approved):
AI Found a few issues, but my only concern is if the new library is pulling in transitive dependencies with CVEs.  Please run the vulnerability scanner and if it finds any, we'll need to bump versions to prevent a CVE exposure.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
sbglasius: Done in 2391490. Removed both `author` and `authorEmail` (the latter was an empty string and meaningless without the former).

grails-doc/src/en/guide/security/securityPlugins/springSecurity/cas/usage.adoc
sbglasius: Good catch, that section directly contradicted the change. Rewritten in 2391490: it now leads with the fact that it's opt-in and *why* (enabling it disables session fixation prevention, because CAS maps the service ticket to the session id), shows the `useSingleSignout: true` snippet, and mentions the startup warning. The former "if you don't want the filter registered, disable it" paragraph is now framed as returning to the default.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
sbglasius: Fixed in 2391490 — renumbered to 48.

Worth flagging that this was more than cosmetic: 45, 46 and 47 all landed upstream while this PR was open, so the file actually had *two* sections numbered 46 and would have produced duplicate asciidoctor anchors. I checked the whole file for other collisions and there are none.

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/AbstractCasSpec.groovy
sbglasius: Agreed, and this is the failure mode most likely to bite in CI. Fixed in 2391490: added a `REQUEST_TIMEOUT` of 60s applied to both `get()` and `postForm()` via `HttpRequest.timeout(...)`, so a container that accepts the connection but never answers fails the spec in a minute instead of hanging until the runner's job-level kill. Kept the 30s `connectTimeout` as-is.

I went with a per-request timeout rather than `@Timeout` so the failure points at the specific stalled call rather than the whole feature method.

grails-spring-security/cas/plugin/src/main/groovy/grails/plugin/springsecurity/cas/SpringSecurityCasGrailsPlugin.groovy
sbglasius: You're right that the annotation was dead, and right about the reasoning: the docs this PR adds say the plugin logs a warning, and an app capturing logs but not stdout would never see a `println`. Restored `log.warn` alongside the `println` in 2391490, matching `SpringSecurityCoreGrailsPlugin:423-424`. Whitespace-only line trimmed too.

grails-test-examples/spring-security/cas/README.md
sbglasius: Correct, and the root cause was worse than the docs: `bootRun` forks its own JVM, so `-DTESTCONFIG` never reached the app at all — there was no way to exercise the endpoint by hand.

Rather than just note the limitation, `build.gradle` now forwards `TESTCONFIG` and `casContainerVersion` to `bootRun`, and the README documents `bootRun -DTESTCONFIG=casProxy` (2391490). Verified end-to-end: drove a login against the container and `/secure/proxyStatus` returned `PROXY_TICKET:PT-2-1B26izig…` instead of `NO_PROXY_TICKET`.

grails-test-examples/spring-security/cas/test1/grails-app/controllers/grails/plugin/springsecurity/cas/test/SecureController.groovy
sbglasius: Fixed in 2391490, 4 spaces per `.editorconfig`. `git diff --ignore-all-space` on that file is empty, so it's purely whitespace.

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/CasSingleSignOutSpec.groovy
sbglasius: Reworded in 2391490 to "which the app opts into via `cas.useSingleSignout`", and added a pointer to `CasNoSingleSignOutSpec` as the spec covering the default.

grails-test-examples/spring-security/cas/test1/src/integration-test/groovy/grails/plugin/springsecurity/cas/test/CasLoginSpec.groovy
sbglasius: Done in 2391490 — it now calls `extractExecution` from the base class with the same `assert execution` guard the other paths get.

One thing your comment surfaced: `extractExecution` was `private static`, so the reuse you suggested wouldn't have compiled. Widened it to package-private to match the sibling helpers (`extractTicket`, `logoutRequest`).

PR #16178, opened 2026-08-19
build: disable Groovy invokedynamic for the Grails 8 compile

### LINKED ISSUE #15293: Groovy - invoke dynamic performance problems
### Issue description

We upgraded a large and data-intensive application from Grails 6.2.3 to Grails 7 but noticed a significant performance regression reaching `4x` on some operations.

We suspect this is related to Groovy 4's "indy" approach and GORM suffering from that more than expected.

It's not noticeable at first sight. I couldn't come up with a simple example where such a big regression would be noticeable so I ended up using an LLM to generate a small application with a method doing something on the data using GORM. 

It's a bunch of domains (Company, Department, Employee, Milestone, Project, Skill, Task) with associations and an initial dataset. Uses H2 dev database for simplicity.

I really hope I haven't missed anything obvious  (some optimisation or flag, hibernate cache is disabled in both) and this is indeed a Groovy 4 regression.

* With 50 iterations (and 5 warmup iterations) I observed the same `4x` regression.
* with `-Dgroovy.indy.optimize.threshold=0 -Dgroovy.indy.fallback.threshold=0` it's only ~2x as bad but it's not entirely clear how to use that, other combinations didn't yield results as good
* when looking at flame graphs unsurprisingly Grails 7 is full of `org.codehaus.groovy.vmplugin.v8.*` where Grails 6 is using `org.codehaus.groovy.runtime.*`
* @CompileStatic hasn't been explored as it requires code changes that I would really like to avoid
* benchmarks were done with `:bootWar`, embedded tomcat with `-Dgrails.env=development` but `:bootRun` is `8x` slower than Grails 6, and `3x` slower than Grails 7 bootWar - so development mode in IDE is also hit

At the moment we've had to stop and revert the migration as a 4x performance hit is hard to swallow.

From groovy4: https://groovy-lang.org/releasenotes/groovy-4.0.html#Groovy4.0-consolidation
```
Currently, the Groovy runtime still contains any necessary support for classes compiled using older versions of Groovy. Please use Groovy versions up to 3.x if you need to generate the older style bytecode.

This work was originally planned for Groovy 3.0, but there were numerous places where "indy" code was noticeably slower than "classic" bytecode. We have made numerous speed improvements (starting with [GROOVY-8298](https://issues.apache.org/jira/browse/GROOVY-8298)) and have some ability to tune internal thresholds (search the code base for groovy.indy.optimize.threshold and groovy.indy.fallback.threshold). That work gave us useful speed improvements, but we welcome further feedback to help improve overall performance of the indy bytecode.
```

Is there any chance this can be further optimised or could Grails 7 potentially work with Groovy 3?

Link to the benchmark applications: https://github.com/jglapa/grails7-performance-regression

jamesfredley:
## Summary

If Grails 9 on Groovy 6, this will flip the other direction:  https://github.com/apache/grails-core/pull/16165

Groovy 5's compiler default is invokedynamic on. In this repo, only modules that apply the Grails Gradle plugin inherited `grails.indy = false`. Everything else (`grails-gradle`, `grails-forge`, `build-logic`, and published framework modules that only apply `org.apache.grails.buildsrc.compile`) compiled with indy on. That left mixed bytecode and the #15293 performance regression in artifacts that never went through the Grails plugin.

This PR turns indy **off for the entire Grails 8 compile**, from one shared default, and applies it in all three independent builds.

## How it is centralized

- `CompilePlugin` (`org.apache.grails.buildsrc.compile`) now sets `optimizationOptions.indy = false`. That is the historical GroovyCompile convention plugin used by framework, `grails-gradle`, and `grails-forge` modules.
- `gradle/groovy-indy.gradle` is applied from `subprojects` in:
  - grails-core (`build.gradle`)
  - grails-gradle (`grails-gradle/build.gradle`)
  - grails-forge (`grails-forge/build.gradle`)
- `build-logic` cannot apply `CompilePlugin` to itself (it compiles that plugin). `plugins` and `docs-core` set the same default locally.
- CI can still opt in with `-PgrailsIndy=true` (same property as `grails-extension-gradle-config.gradle`).
- Applications are unchanged: `GrailsExtension.indy` still defaults to `false`, and `grails { indy = true }` / the GraalVM native convention still win in `afterEvaluate`.

Grails 9 / Groovy 6 can flip this default the other way when indy is ready.

## Verification

- `CompilePluginSpec` asserts `compileGroovy` / `compileTestGroovy` are indy=false by default and true with `-PgrailsIndy=true`
- `:build-logic:test` and `:grails-docs-core:test` pass
- `help` configures cleanly in grails-core, grails-gradle, and grails-forge
- `javap` on `grails-encoder` `CodecMetaClassSupport` shows classic `CallSiteArray`, not invokedynamic dispatch

Fixes #15293 (build-side: stop compiling the framework with Groovy's indy default)

See also #15431 (warning noise when indy is disabled; not changed here).


build-logic/docs-core/build.gradle
```diff
@@ -71,6 +71,13 @@ sourceSets {
     }
 }
 
+// docs-core does not apply org.apache.grails.buildsrc.compile. Keep the same
+// Grails 8 default (indy off) as CompilePlugin. See #15293.
+tasks.withType(GroovyCompile).configureEach {
+    groovyOptions.optimizationOptions.indy = project.hasProperty('grailsIndy') &&
+            Boolean.parseBoolean(project.property('grailsIndy') as String)
+}
```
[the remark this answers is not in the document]
jamesfredley: Done in 2a534f30e9. Every grailsIndy parse site now uses toString().trim().toBoolean(), matching GradleUtils.lookupProperty. build-logic applies the shared gradle/groovy-indy.gradle instead of a local Boolean.parseBoolean copy. CompilePluginSpec covers -PgrailsIndy with surrounding whitespace.

gradle/groovy-indy.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+// Grails 8 keeps invokedynamic off. Groovy 5's compiler default is indy=true,
+// which is a large runtime regression for dynamic Groovy (see #15293).
+// Applied from each independent build (grails-core, grails-gradle, grails-forge)
+// so modules that never apply the Grails Gradle plugin still inherit this.
+// Grails 9 / Groovy 6 can flip the default. CI can still opt in with -PgrailsIndy=true.
+boolean grailsIndyEnabled = false
+if (project.hasProperty('grailsIndy')) {
+    grailsIndyEnabled = Boolean.parseBoolean(project.property('grailsIndy') as String)
+}
```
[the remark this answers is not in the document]
jamesfredley: Done in 2a534f30e9. gradle/groovy-indy.gradle now uses toString().trim().toBoolean() so -PgrailsIndy=" true " matches CompilePlugin.

build-logic/plugins/build.gradle
```diff
@@ -57,6 +57,13 @@ tasks.named('test') {
     useJUnitPlatform()
 }
 
+// This project compiles CompilePlugin itself, so it cannot apply that plugin.
+// Keep the same Grails 8 default (indy off) as CompilePlugin. See #15293.
+tasks.withType(GroovyCompile).configureEach {
+    groovyOptions.optimizationOptions.indy = project.hasProperty('grailsIndy') &&
+            Boolean.parseBoolean(project.property('grailsIndy') as String)
+}
```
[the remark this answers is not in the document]
jamesfredley: Done in 2a534f30e9. This module now applies the shared gradle/groovy-indy.gradle instead of a local Boolean.parseBoolean copy, so trimming matches CompilePlugin.

gradle/grails-extension-gradle-config.gradle
```diff
@@ -35,6 +35,8 @@ grails {
     // Allow CI to toggle Groovy invokedynamic (indy) via -PgrailsIndy=true
     // This enables testing functional tests with both indy enabled and disabled.
     // See: https://github.com/apache/grails-core/issues/15321
+    // Framework modules that do not apply this plugin inherit the same default
+    // from org.apache.grails.buildsrc.compile (CompilePlugin).
     if (project.hasProperty('grailsIndy')) {
         indy = Boolean.parseBoolean(project.property('grailsIndy') as String)
```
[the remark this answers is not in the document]
jamesfredley: Done in 2a534f30e9. This site now uses toString().trim().toBoolean() as well.

build-logic/docs-core/build.gradle
```diff
@@ -71,6 +71,10 @@ sourceSets {
     }
 }
 
+// docs-core does not apply org.apache.grails.buildsrc.compile. Keep the same
+// Grails 8 default (indy off) as CompilePlugin. See #15293.
+apply from: layout.projectDirectory.file('../../gradle/groovy-indy.gradle')
```
jdaugherty: docs-core is not published... why bother turning off indy for it?

build-logic/plugins/build.gradle
```diff
@@ -57,6 +57,10 @@ tasks.named('test') {
     useJUnitPlatform()
 }
 
+// This project compiles CompilePlugin itself, so it cannot apply that plugin.
+// Keep the same Grails 8 default (indy off) as CompilePlugin. See #15293.
```
jdaugherty: this is not published... why bother turning off indy for it?

gradle/groovy-indy.gradle
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+// Grails 8 keeps invokedynamic off. Groovy 5's compiler default is indy=true,
```
jdaugherty: Why do we need this file?  The compile plugin should be used on any published artifact.

grails-forge/build.gradle
```diff
@@ -83,6 +83,8 @@ allprojects {
 }
 
 subprojects {
+    apply from: rootProject.layout.projectDirectory.file('../gradle/groovy-indy.gradle')
```
jdaugherty: The compile plugin should be applied to any project that's needed here so we're consistent.  We shouldn't maintain groovy-indy.gradle

grails-gradle/build.gradle
```diff
@@ -72,6 +72,8 @@ final class ActiveProcessorCountArgumentProvider implements CommandLineArgumentP
 }
 
 subprojects {
+    apply from: rootProject.layout.projectDirectory.file('../gradle/groovy-indy.gradle')
```
jdaugherty: We use composition, let's not break that rule.  Looking below it looks like someone already did this too.  Can we please fix this?

jdaugherty (changes requested):
1. we should be using composition and not using the subprojects { } block in gradle files
2. we should be exclusively applying this to the CompilePlugin since it's supposed to be applied to any project that's being published.
3. for gradle plugins that are not published, why do we need to even toggle indy?  We should use what Gradle's default is.

jamesfredley:
Addressed @jdaugherty review: indy-off is only in CompilePlugin (published artifacts). Removed groovy-indy.gradle and the subprojects applies. Unpublished build-logic/docs-core and plugins use Gradle's Groovy default. Rebased onto latest 8.0.x.

build-logic/docs-core/build.gradle
jamesfredley: Removed. docs-core is unpublished, so it now uses Gradle's Groovy default. Indy is only forced off via CompilePlugin on published artifacts.

build-logic/plugins/build.gradle
jamesfredley: Removed. This unpublished build-logic module now uses Gradle's Groovy default. Published artifacts still get indy-off from CompilePlugin.

gradle/groovy-indy.gradle
jamesfredley: Deleted gradle/groovy-indy.gradle. CompilePlugin is the only place that sets indy=false, and it is already applied to published artifacts.

grails-forge/build.gradle
jamesfredley: Dropped the groovy-indy.gradle apply. Published forge modules already apply CompilePlugin; unpublished ones keep Gradle's default.

grails-gradle/build.gradle
jamesfredley: Removed the groovy-indy.gradle apply from subprojects. Published grails-gradle modules already apply CompilePlugin via composition.

jdaugherty (approved):

PR #16183, opened 2026-08-20
feat: disable the hidden HTTP method filter by default

codeconsole:
## Summary

Grails currently rewrites hidden HTTP methods in a servlet filter that runs before the dispatcher and Spring Security. Reading `_method` there can force a multipart request body to be parsed—and temporary upload files written—before routing or authentication.

This PR disables the Grails hidden-method filter by default and resolves form method overrides inside the dispatcher instead. Existing browser forms continue to work without template changes.

## What changes

- `GrailsDispatcherServlet` resolves `_method` after multipart handling and the filter chain.
- Overrides are limited to `PUT`, `PATCH`, and `DELETE`; `X-HTTP-Method-Override` is no longer used in the new default mode.
- `UrlMappingsHandlerMapping` applies the same resolution when matching routes.
- `resources` mappings add `POST /$controller/$id -> update`, matching the `POST` method already accepted by `RestfulController.update`.
- `<g:form method="PUT">` and `PATCH` submit to that POST update route without `_method`.
- `<g:form method="DELETE">` still emits `_method`, because update and delete share the same URL.
- Grails backs off when Spring Boot's hidden-method filter is enabled, avoiding the `hiddenHttpMethodFilter` bean-name collision introduced in Grails 8.

## Compatibility and authorization

Servlet filters—including Spring Security—now see the real request method. For example, a browser delete form is seen by the filter chain as `POST /books/1`; it becomes `DELETE` only inside Grails dispatching.

Applications with method-specific security rules such as `DELETE /books/**` must review those rules before upgrading. Otherwise, the DELETE-specific rule will not match the form's POST request. This behavior and migration concern are documented in the Grails 8 upgrade guide.

The legacy filter behavior can be restored with:

```yaml
grails:
    web:
        hiddenmethod:
            filter:
                enabled: true
```

When enabled, the dispatcher-level override and additional POST update mapping are disabled.

## Why change the default?

The existing filter reads request parameters before authentication. For `multipart/form-data`, that can trigger early parsing of the entire upload. Spring Boot disabled its equivalent filter by default for the same early-consumption concern.

Grails 8 is the appropriate compatibility boundary for changing this default, while the configuration property provides an opt-out for applications that need time to migrate security rules.


codeconsole:
Note for anyone weighing the authorization risk called out above — it is concrete rather than theoretical, and it has a known remedy.

**Where it bites.** `AbstractFilterInvocationDefinition.findConfigAttributes` compares a rule's declared method to the request's actual method (`grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/access/intercept/AbstractFilterInvocationDefinition.groovy:129`):

```groovy
if (requestMethod && iu.httpMethod && iu.httpMethod != HttpMethod.valueOf(requestMethod)) {
    log.debug "Request '{} {}' doesn't match '{} {}'", requestMethod, url, iu.httpMethod, iu.pattern
    continue
}
```

With the filter enabled it runs at order `-170`, ahead of the security chain at `-100`, so `requestMethod` is already `DELETE` by the time this executes. With the filter disabled it is `POST`, and every Requestmap or `intercept.map` entry declared with `httpMethod: DELETE` is skipped for a browser form delete. The action still runs; the rule guarding it simply does not match, and nothing errors.

**A remedy already sits in the tree, unused.** `grails.plugin.springsecurity.web.filter.HttpMethodOverrideDetector` reads `_method` and returns the intended method. It is referenced by nothing but its own spec — it arrived with the Spring Security plugin merge and was never wired up, which makes sense, because while the filter is enabled it can only tell you what `request.getMethod()` already says. Once the filter is off it becomes the missing piece: it would let the security layer resolve the intended method for itself.

**One constraint if that is pursued.** Reading a request parameter inside the security chain re-creates, for `multipart/form-data`, the pre-authentication body parsing this PR exists to eliminate. That is avoidable by resolving `_method` only for `application/x-www-form-urlencoded` requests — a delete form is always form-encoded, and a multipart POST is a file upload, never a delete. The gap closes without any upload body being touched before authentication.

Deliberately not part of this PR: different module, security-sensitive, and it deserves its own review. Raising it here so the risk is not mistaken for an unsolvable one.

codeconsole:
Added `7cdb38535f` as a separate commit — a build fix that is independent of the feature and easy to review or split out.

**The bug.** The five isolated `Test` tasks in `grails-test-suite-uber` were registered without `testClassesDirs` or `classpath`. A manually registered `Test` task inherits neither from the test source set, so each resolved no candidate classes and reported `NO-SOURCE`. The main `test` task excludes exactly those patterns, so the isolated classes ran **nowhere**: `isolatedTestsOne`, `isolatedTestsTwo`, `isolatedRestRendererTests`, `isolatedPersonTests`, `isolatedRestfulControllerTests`.

Wiring both properties to the test source set brings **40 tests back into the build. 36 pass.**

**The other four, and a defect they expose.** They assert the validation-error branch of a controller action — an invalid domain instance should re-render the `create` or `edit` view. They fail because domain validation is not enforced in this unit-test harness. A minimal probe:

```
PROBE blankValid=true nullValid=true okValid=true
PROBE constrainedProperties=[title]
```

against

```groovy
@Entity
class Probe {
    String title
    static constraints = { title blank: false }
}
```

The constraint is registered and `constrainedProperties` reports it, but `validate()` returns `true` even for `title: null`, whose `nullable` constraint defaults to `false`. So the "invalid" instance is valid, the action takes its success path, and the error-view assertions never run.

That is pre-existing and unrelated to this PR — it was simply invisible while the tasks reported `NO-SOURCE`. I have marked those four `@PendingFeature` rather than adjust them to match the broken behaviour: the annotation fails the build if validation is fixed and they start passing, so it cannot outlive the bug. Happy to raise it as its own issue if that is preferred.

jdaugherty:
We removed the dispatch actions in Grails 8 already (deprecated in Grails 7).  Is _method just left over from that?  I had no idea you could even change the method for a form - why wouldn't you either set the form to the intended method or just set the [formmethod](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement/formMethod)?  It seems like _method is from an earlier implementation when there weren't built in constructs to change the method in html5.

codeconsole:
> We removed the dispatch actions in Grails 8 already (deprecated in Grails 7). Is _method just left over from that? I had no idea you could even change the method for a form - why wouldn't you either set the form to the intended method or just set the [formmethod](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement/formMethod)? It seems like _method is from an earlier implementation when there weren't built in constructs to change the method in html5.

No, it is for `RestfulServiceController` and `g:form` and has always been there.  It is used in the scaffolding plugin. 

```html
<g:form method="PUT" action="/books">
```

creates
```html
<form method="POST" action="/books">
<input type="hidden" name="_method" value="PUT">
```

then HiddenMethodFilter wraps the entire request so the entire filter stack (including Spring Security thinks the request  is `PUT /books` instead of `POST /books` (what it really is)
 

This PR keeps the actual request and uses `_method` just for `DELETE` and utilizes the existing mapping urls for the rest of the requests.  Alternatively, you could add 1 more rule for every mapping, BUT that creates O(n) mappings so if you have a lot of rules, it just overcomplicates.  

I a follow up PR would also be to not use `_method` and use actual javascript to send a `DELETE` but that also requires javascript working in the browser.  

If you only use POST and GET in your forms, this PR will only help you not forcing multipart resolution at the beginning of the filter stack

matrei:
The way I understand it, this is the current PR behavior:

```console
// Pseudo-code
if (grailsFilterEnabled) {
    if (springFilterEnabled) {
        use Spring HiddenHttpMethodFilter
    }
    else {
        use Grails HiddenHttpMethodFilter
    }

    disable Grails dispatcher-level override
}
else {
    if (springFilterEnabled) {
        use Spring HiddenHttpMethodFilter
        ALSO enable Grails dispatcher-level override
    }
    else {
        no servlet HiddenHttpMethodFilter
        enable Grails dispatcher-level override
    }
}
```

Wouldn't this be more appropriate?

```console
// Pseudo-code
if (springFilterEnabled) {
    use Spring HiddenHttpMethodFilter
    Grails compatibility/filter mode = true
}
else if (grailsFilterEnabled) {
    use Grails HiddenHttpMethodFilter
    Grails compatibility/filter mode = true
}
else {
    no servlet HiddenHttpMethodFilter
    use Grails dispatcher-level override
}
```

codeconsole:
Accurate description of the behaviour, and fixed in 482fa1c — your structure, with one addition.

The mode now comes from `HiddenHttpMethod.isServletFilterMode` (either property), used by the dispatcher, handler mapping, URL mapping evaluator and `g:form` alike so they cannot drift.

That is only sound if a filter really exists whenever either property is set, and it did not: Boot's filter comes from `WebMvcAutoConfiguration`, which backs off entirely under `@EnableWebMvc`. Asking for Boot's filter there produced **no** filter, while the Grails one had already backed off on the property. So the registration moved to its own auto-config ordered after Boot's, gating on `@ConditionalOnMissingBean` — detecting Boot's filter instead of assuming it — and contributing the Grails filter when Boot has not. Same pattern `GrailsFormContentFilterAutoConfiguration` already uses.

Worth noting the case you found was worse than redundant: with `spring=true`, `g:form` stopped emitting `_method` for PUT and the POST member route appeared, so form updates hit the security chain as a POST — the exact silent authorization change the upgrade note warns about, landing on whoever opted in to avoid it.

Covered by `GrailsHiddenHttpMethodFilterAutoConfigurationSpec`, including the `@EnableWebMvc` case.

codeconsole:

PR #16184, opened 2026-08-21
Run GSP in a Spring Boot application without the Grails plugin lifecycle

codeconsole:
A Spring Boot application that renders its views with GSP could not start, and behind the first failure were four more, each hidden by the one before it. The `gsp-spring-boot` example now starts, renders, decorates and packages, and its runtime test is enabled again.

## Using GSP from a Spring Boot application

Views render from a plain `SpringApplication`, with no Grails application class and no Grails plugins:

```java
@Configuration
@ComponentScan
@EnableAutoConfiguration
public class Application {
    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
    }
}
```

The whole of the example's `application.properties` is now the one property that says something about GSP:

```properties
sitemesh.decorator.default=main
```

`spring.main.allow-circular-references` and `spring.main.allow-bean-definition-overriding` are gone. Both were opted into because of how the GSP beans were wired, not because the application wanted either.

Views compiled by the `compileGroovyPages` build task are rendered from their compiled classes, so an application can ship without its templates:

```groovy
jar {
    processResources.exclude('**/*.gsp')
}

compileGroovyPages {
    source = project.file("${project.projectDir}/src/main/resources/templates")
    serverpath = '/'   // the path a standalone application looks a view up by
}
```

Templates on disk still win where they are there to edit, so `bootRun` re-renders a template as it changes.

## Styling with the asset pipeline

The example styles itself with Bootstrap, compiled by the asset pipeline at build time and served by the pipeline's own Spring Boot module. Bootstrap is an input to the build rather than a library the application ships - it goes on the `assets` configuration, and the artifact carries the compiled, digest-named stylesheet and script instead of the 1.8MB webjar:

```groovy
apply plugin: 'cloud.wondrify.asset-pipeline'

dependencies {
    assetDevelopmentRuntime platform(project(':grails-bom'))
    assetDevelopmentRuntime 'org.webjars.npm:bootstrap' // compiled from, not packaged
    implementation 'cloud.wondrify:asset-pipeline-spring-boot' // the filter that serves what was compiled
}
```

```java
@Import(AssetPipelineService.class) // the module's own configuration
public class Application { ... }
```

```html
<link rel="stylesheet" href="${request.contextPath}/assets/application.css"/>
```

Nothing Grails-side is declared for it: no tag library, no `GrailsApplication` wiring, and `asset-pipeline-grails` is not on the application's class path.

## What changed

**The Grails plugin lifecycle runs for a Grails application only** — one that `GrailsApp` launched, or one with a Grails application class among the sources. `GrailsPluginLifecycleInitializer` is registered for every Spring Boot application with grails-core on its class path, so an application using a Grails library was given a `GrailsApplication`, a plugin manager and the beans of every plugin found, over the top of what the libraries it did ask for auto-configure for themselves. It now gets that library's auto-configuration and nothing else. Documented in the 8.0 upgrade notes, with the one case inside a Grails application it changes: a `@SpringBootTest` that names configuration classes of its own now names the application class as well.

**Bean wiring.** Tag libraries are beans of the context, found by the lookup once they exist, rather than held inside it — they are autowired with the lookup, so a lookup that held them was a cycle. The GSP codec lookup is configured after the codecs module, so its `@ConditionalOnMissingBean` guard can do its work instead of being overridden. The core plugin's auto-configuration is contributed only where there is a `GrailsApplication` - the condition is on the plugin class, so the generated `CoreAutoConfiguration` is gated as a whole - and a context without one starts rather than failing on a bean of a type it never had.

**Standalone GSP.** The view registry `compileGroovyPages` writes is read into the page locator; `<g:applyLayout>` resolves its layout; the `grailsLayout` namespace is registered, so the capture tags the GSP compiler emits no longer reach the browser as markup; the JSP tag library resolver is optional, so GSP renders without JSP support on the class path; and the tag libraries a Grails plugin carries are found alongside the ones marked `@TagLib`.

**Auto-proxy creators.** The reflection patch that makes Spring recognise the Groovy aware auto-proxy creator had sat in a `GrailsAutoConfiguration` static initializer since 2015, while `CoreGrailsPlugin` is what registers the creator. It moved to `GroovyAwareAutoProxyCreators`, applied where the creator is registered, so it holds for an application `GrailsApp` launched from sources that do not load that class.

**An application that maps no URLs can still have a link generator.** `DefaultLinkGenerator` requires the URL mappings holder, as it should. `GspAutoConfiguration` contributes an empty one under `@ConditionalOnMissingBean(name = "grailsUrlMappingsHolder")`, ordered after `UrlMappingsAutoConfiguration` so a holder from there wins; only a standalone application has that module on its class path, so a Grails application never sees the bean.

**The view registry is written whole** rather than merged into what an earlier build left behind, which kept naming views since renamed or removed, against classes no longer there.

**The compiled pages are on the test class path of a build that is not a Grails build.** `GroovyPagePlugin` registered them with a path resolved against the project directory rather than the build directory, so an application tested from the build could load none of them. A Grails application's tests render GSP from source and gain nothing from them, so its `test` task is left as it was - the plugin records whether `GrailsGradlePlugin` is applied and decides when the task graph is built. They are kept off the main runtime class path, which a boot archive packages into its classes directory - where the archive copies them already.

**`sitemesh.decorator.default` is consulted for the default layout**, after both Grails keys. A Grails application that set only the SiteMesh key was decorated with the implicit `application` layout and now gets the layout the key names; in the upgrade notes.

**The JSP compiler is left out of the executable jar** — 3.3MB of Eclipse compiler that a jar cannot use, since it packages no JSP to compile. The war and `bootRun`, which do serve JSPs, keep it.

## Limitations

- **JSP cannot be served from an executable jar.** Jasper compiles a JSP from the servlet context, and a jar packages none, so the example offers its JSP rendering only where it can serve one. Jasper itself stays in every artifact: the GSP form uses Spring's form tag library, which is a JSP tag library, and renders nothing without the JSP API Jasper carries.
- **Precompiled views need `serverpath`** set to match the template root. The default registers each view below `/WEB-INF/grails-app/views/`, which is where a Grails application looks and a standalone application never does.
- **A logical asset URL is answered with a one-year `Cache-Control`.** The pipeline's servlet filter rewrites `/assets/application.css` to the digest-named file and then decides cacheability from the rewritten name, which is never a manifest key, so it answers `public, max-age=31536000` on a URL whose content changes on redeploy. Measured against the packaged example. The fix belongs in the asset pipeline, not here.
- Rendering GSP still puts `grails-core` on the class path for the `GrailsApplication` that the page locator, tag library lookup and JSP tag library resolver read, and with it `grails-datastore-core`, javassist, caffeine and `jakarta.persistence-api` — around 2.6MB of persistence machinery that `GrailsApplication.getMappingContext()` makes structural rather than incidental.


grails-test-examples/gsp-spring-boot/app/src/main/java/hello/AssetPipelineConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>A Grails application finds such a tag library by scanning the artefacts of the plugin that
+ * carries it. This application has no plugins, so it declares the tag library as a bean; GSP's tag
+ * library lookup registers every tag library bean of the context, and the tags are then usable in a
+ * page exactly as the ones GSP contributes itself.
+ *
+ * <p>The service the tag library reads asset URLs from is auto-configured by the asset pipeline, so
+ * only the tag library and the filter that serves the compiled assets are declared here.
+ */
+@Configuration(proxyBeanMethods = false)
+public class AssetPipelineConfiguration {
```
jdaugherty, line 55: This configuration shouldn't be in this PR.  We should be contributing this to the asset pipeline as a separate library that's included here.

codeconsole: @jdaugherty It will be removed after the asset pipeline is released

jdaugherty:
We shouldn't add this to 8.0.x since this will require another milestone before we can release RC1.  The associated asset pipeline changes are still not removed this PR either, so we should move this to a future release instead.

codeconsole:
@jdaugherty my only issue with it not being in 8.0 is that it currently **DOES NOT WORK AT ALL** and is **COMPLETELY BROKEN**

This is more of a bug fix.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
```diff
@@ -437,6 +436,22 @@ class GroovyPagePlugin implements Plugin<Project> {
...
+        //
+        // The main runtime class path is deliberately left alone. A boot archive packages every
+        // directory of it into its own classes directory, and the pages are copied there already, so
+        // each page would arrive twice - and an application run from the build renders its templates
+        // as they are edited, which is what a page compiled ahead of the edit would stand in the way
+        // of.
+        FileCollection compiledPages = project.files(destDir, webappDestDir).builtBy(compileGroovyPages)
+        SourceSet testSourceSet = SourceSets.findSourceSet(project, SourceSet.TEST_SOURCE_SET_NAME)
+        if (testSourceSet != null) {
+            testSourceSet.runtimeClasspath = testSourceSet.runtimeClasspath + compiledPages
```
jdaugherty, line 468: This puts every Grails project's `test` task behind `compileGroovyPages`. The classpath is `builtBy` the compile task, so `./gradlew test` now forks and compiles all of `grails-app/views` first, and any GSP edit invalidates `test` and reruns the whole unit suite.

A Grails project gets nothing for that cost. Unit tests render GSP from source through `GroovyPageUnitTestResourceLoader`, and integration tests run with the project directory available, so `GroovyPagesGrailsPlugin` never populates the precompiled map. The consumer is a standalone Spring Boot application such as the gsp-spring-boot example, whose tests live in `src/test` and whose build strips the templates. It is right that applying this plugin makes that work without the application wiring anything itself.

Please scope it to that case: add the compiled pages to the test runtime classpath only when `GrailsGradlePlugin` is not applied to the project, so a Grails application or plugin build is left as it was. `webappDestDir` also needs `builtBy(compileWebappGroovyPages)`; as written it can be stale on the classpath.

Detect the Grails build lazily, not with `hasPlugin` at configure time and not in `afterEvaluate`. Both plugins have historically been applied with `apply plugin:` in either order, and either can arrive from another plugin or a later callback, so an eager check is order-dependent. Record presence with `project.plugins.withType(GrailsGradlePlugin) { grailsBuild = true }`, which fires on apply however the plugin is applied, and make the decision inside a provider that Gradle evaluates when it builds the task graph, after every configuration-phase hook has run:

```groovy
FileCollection compiledPages = project.files(project.provider {
    grailsBuild ? [] : [project.files(destDir, webappDestDir).builtBy(compileGroovyPages, compileWebappGroovyPages)]
})
```

Gradle infers the `builtBy` dependencies from the provider's value, so `test` depends on the compile tasks only when the pages are actually contributed. The same works in the other direction if you prefer: `GrailsGradlePlugin` already has a `withPlugin('org.apache.grails.gradle.grails-gsp')` block and could flip a lazy `Property<Boolean>` on this plugin's extension from there, provided the extension is registered directly in `apply()` rather than inside the `withPlugin('groovy')` callback.

grails-core/src/main/groovy/org/grails/plugins/CoreGrailsPlugin.groovy
```diff
@@ -91,13 +93,19 @@ class CoreGrailsPlugin extends Plugin {
     private static final String SPRING_PROXY_TARGET_CLASS_CONFIG = 'spring.aop.proxy-target-class'
 
     def beans = {
-        bean(ClassLoader).primary() { GrailsApplication grailsApplication ->
+        // Both of these are the GrailsApplication read through another type, so both stand down for
+        // an application that has none: this configuration is contributed by every Spring Boot
+        // application with grails-core on its class path, where only a Grails application has the
+        // plugin lifecycle that builds one.
+        bean(ClassLoader).primary().annotate(ConditionalOnBean, value: GrailsApplication) { GrailsApplication grailsApplication ->
```
jdaugherty: This is Spring Boot conditional wiring inside the core plugin's `beans` block, and it makes core beans depend on Boot's condition ordering. `@ConditionalOnBean(GrailsApplication)` holds only because the early registration phase promotes `grailsApplication` before `ConfigurationClassPostProcessor` evaluates the generated auto-configuration. Any context where that ordering does not hold loses both beans, and `GroovyPagesGrailsPlugin` then fails on `ref('classLoader')`.

Conventionally these beans exist because there is a Grails application; the core plugin should not have to defend against contexts without one. If the intent is that a non-Grails Boot application gets nothing from core, gate the generated `CoreAutoConfiguration` as a whole rather than individual beans. As it stands the surface is mixed: these two back off while the rest still land in plain Boot applications, including `GrailsPlaceholderConfigurer` with `ignoreUnresolvablePlaceholders=true`, which `CoreAutoConfigurationSpec` now asserts as intended behaviour.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
```diff
@@ -76,7 +76,13 @@ class DefaultLinkGenerator implements LinkGenerator, PluginManagerAware {
...
     GrailsPluginManager pluginManager
 
-    @Autowired
+    /**
+     * The mappings a link to a controller and action is created from. An application that maps no
+     * URLs - one routing with Spring MVC, say - has none, and still generates links to a resource
+     * or to a path, so this is the one kind of link it cannot generate rather than a reason for it
+     * not to start.
+     */
+    @Autowired(required = false)
```
jdaugherty: This has to stay required. The URL mappings holder is core to the link generator and to the GSP tags built on it (`ApplicationTagLib` autowires both), and turning it into a nullable field with a runtime `IllegalStateException` softens a Grails invariant to accommodate a consumer outside this repo: the asset pipeline's auto-configuration contributes a `grailsLinkGenerator` to an application that maps no URLs.

The fix belongs on that side. Whatever contributes a `DefaultLinkGenerator` to a Spring MVC application should also contribute an empty `grailsUrlMappingsHolder` (`new DefaultUrlMappingsHolder([])`), or `GspAutoConfiguration` can do so under `@ConditionalOnMissingBean(name = "grailsUrlMappingsHolder")`. Please revert this and the null check below, along with the `LinkGeneratorSpec` cases that pin a generator without a holder.

grails-gsp/grails-web-gsp-taglib/src/main/groovy/org/grails/web/pages/StandaloneTagLibraryLookup.java
```diff
@@ -39,7 +42,12 @@
  * @author Lari Hotari
  * @since 2.4.0
  */
-public class StandaloneTagLibraryLookup extends TagLibraryLookup implements ApplicationListener<ContextRefreshedEvent> {
+public class StandaloneTagLibraryLookup extends TagLibraryLookup
+        implements SmartInitializingSingleton, ApplicationListener<ContextRefreshedEvent> {
+
+    /** What {@link Artefact} marks a tag library with, the way a Grails plugin declares one. */
+    private static final String TAG_LIB_ARTEFACT = "TagLib";
```
jdaugherty: Use `TagLibArtefactHandler.TYPE` rather than a local copy of the string. It lives in grails-taglib, which this class already depends on for `DefaultGrailsTagLibClass`.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
```diff
@@ -120,6 +127,20 @@ public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) t
...
+        // Two things make a context a Grails application: GrailsApp launched it, which it records by
+        // stashing the sources it was given, or one of its sources is a Grails application class.
+        // This initializer is registered for every Spring Boot application with grails-core on its
+        // class path, and the plugin lifecycle is not something the rest of them asked for: it would
+        // contribute a GrailsApplication, a plugin manager and the beans of every plugin found, over
+        // the top of whatever the libraries they did ask for auto-configure for themselves.
+        boolean launchedByGrails =
+                applicationContext.getBeanFactory().getSingleton(APPLICATION_SOURCE_CLASSES_BEAN_NAME) != null;
+        Class<?>[] applicationSources = resolveApplicationSourceClasses(registry);
+        if (!launchedByGrails && !containsApplicationClass(applicationSources)) {
```
jdaugherty, line 139: The class Javadoc above (line 87) still says that when the application was not started through `GrailsApp` "the phase proceeds without application classes". With this gate it stands down instead; please update it.

One case worth spelling out, here or in the upgrade note: a `@SpringBootTest(classes = SomeConfig)` inside a Grails application that does not name the `Application` class used to get the plugin lifecycle and now gets none. `@Integration` is unaffected because it always names the application class.

This is likely a critical regression.  We should either document this in the upgrade notes or restore the behavior for Grails apps to automatically configure SpringBootTest annotated tests.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -3025,3 +3025,21 @@ Grails 7 (`org.apache.grails.data:grails-datamapping-async`) coordinates in a pr
...
+views, say — was given a `GrailsApplication`, a plugin manager, and the beans of every plugin found on the
+class path, over the top of what the libraries it did ask for auto-configure for themselves.
+
+The lifecycle now runs only for a Grails application: one that `GrailsApp` launched, or one where a source of
+the context is a Grails application class, which is what `grails-app/init/Application.groovy` is. A Grails
+application is unaffected either way. A Spring Boot application using a Grails library gets that library's
+auto-configuration and nothing else.
+
+This matters if you deliberately put a Grails plugin on the class path of an application that is not a Grails
+application and relied on its `doWithSpring` or `beanRegistrar` beans being contributed. Those beans are no
```
jdaugherty, line 3545: Please also cover the in-application case: a `@SpringBootTest` in a Grails application that names a configuration class other than `Application` no longer gets the plugin lifecycle. That is the change a Grails user is most likely to hit.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
```diff
@@ -147,9 +147,13 @@ class Sitemesh3GrailsPlugin extends Plugin {
...
             // grails.views.layout.default key so existing apps keep their
-            // configured default layout when switching.
+            // configured default layout when switching, and finally to SiteMesh's own
+            // sitemesh.decorator.default - the key a Spring Boot application using GSP for views
+            // configures, and the one Sitemesh3EnvironmentPostProcessor derives from the Grails
+            // keys above, so it only decides when neither of them is set.
             String defaultLayout = config.getProperty('grails.sitemesh.default.layout') ?:
-                    config.getProperty('grails.views.layout.default')
+                    config.getProperty('grails.views.layout.default') ?:
+                    config.getProperty('sitemesh.decorator.default')
```
jdaugherty, line 156: This is a behaviour change for Grails applications, not only a standalone fallback. An application that set `sitemesh.decorator.default` without either Grails key previously got the finder's `application` default and now gets the configured layout. Probably what people want, but it should be stated in the upgrade notes alongside the layouts.adoc change, which currently reads as a description of existing behaviour.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
```diff
@@ -151,16 +151,14 @@ class GroovyPageCompiler {
...
                 viewregistryFile.parentFile.mkdirs()
                 // Use SortedProperties to ensure a consistent order of entries for reproducible builds
                 Properties views = CollectionFactory.createSortedProperties(false)
-                if (viewregistryFile.exists()) {
-                    // only changed files are added to the mapping, read the existing mapping file
-                    viewregistryFile.withInputStream { stream ->
-                        views.load(new InputStreamReader(stream, 'UTF-8'))
-                    }
-                }
                 views.putAll(compileGSPRegistry)
```
jdaugherty, line 167: Writing the registry whole is the right fix, but it now relies on every caller handing the compiler the full source set on each run; `GroovyPageForkCompileTask` and `GroovyPageCompilerTask` both do today. Worth saying so in this comment, since the old merge existed for callers that passed only changed files.

jdaugherty (dismissed):
Two themes across the framework changes:

First, Grails code changed to serve a non-Grails consumer: the @ConditionalOnBean conditions in CoreGrailsPlugin, the optional URL mappings holder in DefaultLinkGenerator, and the test classpath wiring in GroovyPagePlugin. Those need to stay conventional Grails; details inline.

Second, behaviour changes for Grails applications that the upgrade notes should state: the lifecycle gate's effect on @SpringBootTest inside a Grails application, and the SiteMesh default-layout fallback.

The standalone GSP module changes themselves look right. We need to remove the Asset configuration to merge this as well.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/views/gsp/GroovyPagePlugin.groovy
codeconsole: Scoped to a non-Grails build in 1ca0145550, in the provider form above: `plugins.withType(GrailsGradlePlugin)` records the answer however the plugin is applied, and `project.provider` decides when Gradle builds the task graph. `webappDestDir` now declares `builtBy(compileWebappGroovyPages)` rather than relying on `compileGroovyPages` depending on it.

New test project `gsp-compile-classpath-grails` asserts a Grails build gets no pages on the test runtime classpath and that `test` has no `compileGroovyPages` dependency; the existing non-Grails case still asserts it does.

grails-core/src/main/groovy/org/grails/plugins/CoreGrailsPlugin.groovy
codeconsole: Gated as a whole in 3af263cb69. `@ConditionalOnBean(GrailsApplication)` moved to the plugin class, which `@GrailsBeans` carries onto the generated `CoreAutoConfiguration` — `@Conditional` is in its sibling-annotation set — and both beans are declared plainly again.

The surface is no longer mixed: a plain Boot application gets none of the configuration, `grailsResourceLocator` and `GrailsPlaceholderConfigurer` included, and places holders the way Spring Boot does. `CoreAutoConfigurationSpec` asserts that instead. A Grails application is unaffected — `grails-core` and the `gsp-sitemesh3` `@Integration` suite pass.

grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
codeconsole: Reverted in d9c95450a5: `@Autowired` is required again, the null check and the `LinkGeneratorSpec` cases are gone, and `grails-web-url-mappings` is identical to 8.0.x.

`GspAutoConfiguration` contributes `new DefaultUrlMappingsHolder([])` under `@ConditionalOnMissingBean(name = "grailsUrlMappingsHolder")`, ordered after `UrlMappingsAutoConfiguration` so a holder from there wins. `grails-gsp-spring-boot` is only on a standalone application's classpath, so no Grails application sees the bean. Covered by `UrlMappingsHolderConfigurationSpec`, and the example application no longer declares the holder itself.

grails-gsp/grails-web-gsp-taglib/src/main/groovy/org/grails/web/pages/StandaloneTagLibraryLookup.java
codeconsole: Done in 3c8fe2eac2.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
codeconsole: Javadoc updated in 99cf46fb57.

The `@SpringBootTest` case is documented rather than restored. The only build-time marker separating a Grails application from a Boot application is `META-INF/grails.build.info`, and `GrailsGradlePlugin` writes it for plugin projects too — keying the gate off it would let any application with a Grails plugin on its classpath back into the lifecycle, which is what the gate exists to prevent.

The upgrade note now covers the case with its migration: name the application class alongside the configuration, `@SpringBootTest(classes = [Application, SomeConfig])`. `resolveApplicationSourceClasses` recovers a `GrailsApplicationClass` from the registry, so that context is a Grails application again; `EarlyPluginRegistrationOrderingSpec` covers that path. If you would rather restore the behaviour instead, say so and I will look for another signal.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Covered in 99cf46fb57, with a before/after example and the note that `@Integration` and the `grails-testing-support` unit test mixins are unaffected.

grails-gsp/grails-sitemesh3/src/main/groovy/org/grails/plugins/sitemesh3/Sitemesh3GrailsPlugin.groovy
codeconsole: Documented in 688436e00c as its own upgrade note. It states the change for a Grails application — one that set `sitemesh.decorator.default` and neither Grails key was decorated with the implicit `application` layout and is now decorated with the layout that property names — and gives both ways out: set `grails.sitemesh.default.layout`, or remove the SiteMesh key. The layouts.adoc wording stays as the description of the resolution order.

grails-gsp/core/src/main/groovy/org/grails/gsp/compiler/GroovyPageCompiler.groovy
codeconsole: Stated in the comment in 688436e00c: writing whole relies on a caller handing the compiler the entire source set on every run, which `GroovyPageForkCompileTask` and `GroovyPageCompilerTask` both do — the fork task passes its full `srcDir`, the Ant task scans `srcdir` for `**/*.gsp` — and a caller passing only changed files would now write a registry naming those alone.

grails-test-examples/gsp-spring-boot/app/src/main/java/hello/AssetPipelineConfiguration.java
codeconsole: Removed in 48175cf3fd. The example serves what the pipeline compiles through `asset-pipeline-spring-boot`'s own `AssetPipelineService`, imported on `Application`, and the layout links the compiled files by name. No tag library and nothing Grails-side is declared by the application; `asset-pipeline-grails` is off its class path.

matrei:
# AI Review

Verified the framework changes against the modules they touch and ran the example. The lifecycle gate, the whole-configuration gate on `CoreAutoConfiguration`, the auto-proxy creator move (Spring 7.0.9 still keeps `APC_PRIORITY_LIST` as a mutable `ArrayList`, and the order of the two creators is preserved), the write-whole view registry (up-to-date pages are re-registered on every run, and the two compile tasks write to separate directories) and the SiteMesh alias handling all check out. `@Integration` is unaffected by the gate because `GrailsApplicationContextLoader` launches through `GrailsApp`. One defect, a few test and doc gaps, and two design notes below.

### Bug: `Sitemesh3LayoutTagLib` renders attribute values unencoded in a standalone application

`TagLibraryLookupRegistrar.createTagLibBeanDefinition` registers the tag libraries with no autowire mode, relying on each class's own `@Autowired` declarations. `RenderTagLib` and `RenderSitemeshTagLib` declare them; `Sitemesh3LayoutTagLib` does not - its `codecLookup` is a plain property that a Grails application fills by name through `TagLibBeanDefinitionsPostProcessor` (`AUTOWIRE_BY_NAME`). In the gsp-spring-boot example the field is null (probed with a `@SpringBootTest`: `renderTagLib.codecLookup` is a `DefaultCodecLookup`, `layoutTagLib.codecLookup` is `null`), so `captureTagContent` takes its `htmlEncoder == null` branch and writes every attribute of the captured `<head>`, `<body>`, `<title>` and friends verbatim. A layout or page with `<body class="${...}">` puts the raw value into the response.

Fix: `@Autowired` on `Sitemesh3LayoutTagLib.codecLookup` (as `RenderSitemeshTagLib` does), and a test that a captured attribute value is HTML-encoded in the standalone context - either in `TagLibraryLookupRegistrarSpec` or as a rendered-page assertion in the example.

### Tests

- `CoreGrailsPluginRegistrarSpec` ("nothing here loads GrailsAutoConfiguration"): `APC_PRIORITY_LIST` is JVM-global and `test-config.gradle` sets `forkEvery = 100`, so any spec that touched `GrailsAutoConfiguration` earlier in the fork (`EarlyPluginRegistrationOrderingSpec`, `GrailsAutoConfigurationArtefactsSpec`, ...) has already patched the list. Reverting the `registerWithAopConfigUtils()` call in `CoreGrailsPlugin` does not reliably fail this test.
- `GroovyPageCompilerSpec` never sets `generatedGroovyPagesDirectory`, so the compiler falls back to `${java.io.tmpdir}/gspcompile`, shared across parallel forks and never cleaned. Point it at the `@TempDir`.
- `GspAutoConfigurationSpec` exercises only the protected `resolvePrecompiledViews`. The public surface - that the `groovyPageLocator` bean receives the map, and the new `paths.add(cleanUri(uri))` search path - is covered only by the example's `PrecompiledViewTest`; nothing in the module fails if either is reverted.
- `GroovyPagePluginFunctionalSpec`: the non-Grails project asserts the pages are on the test runtime class path but not that `test` depends on `compileGroovyPages`; the Grails project asserts the negative. The `builtBy` wiring is the whole point of the change and is unverified.
- No test asserts that `CoreAutoConfiguration` stays out of the standalone context once `GspAutoConfiguration` contributes its own `GrailsApplication`. That holds today only because Core is `@AutoConfigureOrder(HIGHEST_PRECEDENCE)` and is evaluated before the standalone `grailsApplication` bean definition exists; if that ordering ever changed, core beans would land silently. A `context.containsBean('pluginManager') == false` in the example would pin it.

### Docs

- `upgrading80x.adoc` (SiteMesh note) names `Sitemesh3EnvironmentPostProcessor`, an internal class (rule 8). The sentence works without it.
- Nothing user-facing documents the standalone additions: precompiled views read from `classpath:gsp/views.properties` and skipped when a `file:` template root is in use, the `serverpath = '/'` requirement, the empty `grailsUrlMappingsHolder`, and `@Artefact("TagLib")` beans being picked up. The module README is unchanged. Rule 7 asks for doc coverage with the feature.

### Design notes

- `@Bean(name = {"gspViewResolver", "jspViewResolver"})`: a Spring Boot application that also serves JSP may well have its own `jspViewResolver`. User configuration is processed before auto-configuration, and `SimpleAliasRegistry.registerAlias` does not check bean definition names, so the alias would silently take over `getBean("jspViewResolver")` rather than fail. Registering the alias only when no bean of that name exists (the `ReplaceViewResolverRegistrar` already does this kind of thing) would be safer.
- Example: `JspViewController`'s Javadoc says it "is mapped only where a JSP can be served"; it is always mapped and the guard is the runtime `canServeJsp` check. The `private static boolean jsp` in `WebController` is application-wide state that `JspViewTest` has to reset; session state would be the idiomatic example.

### Process

The thread still has the open question of whether this lands on 8.0.x before RC1; the lifecycle gate changes what a Grails application's `@SpringBootTest(classes = SomeConfig)` gets, which is the kind of change the "review then commit" policy is for.

codeconsole:
@matrei All addressed, pushed as 3ac93a3aa9..16eb980bb8.

**Bug — unencoded attribute values** (3ac93a3aa9). Confirmed: with the registrar's wiring `codecLookup` was null and `captureTagContent` wrote `<body class=""><script>x</script>"></body>`. `@Autowired` on the field, as `RenderSitemeshTagLib` has it. Two tests in `Sitemesh3LayoutTagLibSpec` pin the encoded output and what an unwired lookup costs, and `StandaloneContextTest` in the example asserts the context wires the lookup on both tag libraries — that one fails with the declaration removed.

**Tests** (7dc87ac976, 3ceb572b79)

- `CoreGrailsPluginRegistrarSpec` now clears the Grails entries from `APC_PRIORITY_LIST` before the feature and restores them after. Reverting `registerWithAopConfigUtils()` in `CoreGrailsPlugin` now fails it; it did not before.
- `GroovyPageCompilerSpec` points `generatedGroovyPagesDirectory` at its `@TempDir`.
- `GspAutoConfigurationSpec` drives the contributed locator through `findPage` and gets a `GroovyPageCompiledScriptSource` back. The registry fixture names classes that exist, so reverting either `setPrecompiledGspMap` or `paths.add(cleanUri(uri))` fails it in-module.
- `GroovyPagePluginFunctionalSpec`: the non-Grails project asserts `test` depends on both `compileGroovyPages` and `compileWebappGroovyPages`; the Grails project asserts neither.
- `StandaloneContextTest` asserts `pluginManager`, `grailsConfigProperties` and `classLoader` are absent from the example's context.

**Docs** (16eb980bb8). The SiteMesh note no longer names `Sitemesh3EnvironmentPostProcessor`. New guide page *The Web Layer → Groovy Server Pages → GSP in a Spring Boot Application* covering the precompiled views and when they are skipped, the `serverpath = '/'` requirement, the empty `grailsUrlMappingsHolder`, `@Artefact("TagLib")` beans, and the `spring.gsp.*` properties. The module README described sample applications that no longer exist and is replaced.

**Design notes**

- `jspViewResolver` (3d23083278): moved out of `@Bean(name = {...})` into `ReplaceViewResolverRegistrar`, which registers the alias only when no bean definition holds the name. Two cases in `TagLibraryLookupRegistrarSpec` cover both branches.
- Example (6ca3141d55): the selected rendering is a session attribute; `JspViewTest` keeps its cookies rather than resetting application state. The stale sentence was on `WebController.selectJsp` and is corrected — the controller is always mapped, and what it checks is the servlet context at request time.

**Process.** The 8.0.x-before-RC1 question is still open and is @codeconsole's call, not something these commits settle.

matrei:
**Second round**, on 16eb980bb8 against 8.0.x (merge base a5758d657f is the current tip of 8.0.x, so the branch is clean). Reviewed the six commits since the first round, 3ac93a3aa9..16eb980bb8, and ran `:grails-gsp-spring-boot:test`, `:grails-sitemesh3:test`, `GroovyPageCompilerSpec`, `CoreGrailsPluginRegistrarSpec`, the gsp-spring-boot example's tests, `codeStyle` for the touched modules and `GroovyPagePluginFunctionalSpec` in grails-gradle. All green.

Every point from the first round is addressed: the encoding defect, the four test gaps, the two doc gaps and both design notes. What is left is in the new guide page, plus one test-hygiene detail.

### Docs: the tag library example in `gspInSpringBoot.adoc` does not compile and is not what the example does

`grails-doc/src/en/guide/theWebLayer/gsp/gspInSpringBoot.adoc:80-81`:

```java
public AssetsTagLib assetsTagLib(AssetProcessorService assetProcessorService) {
    return new AssetsTagLib(assetProcessorService);
}
```

`AssetsTagLib` (asset-pipeline-grails 5.2.0-M3, the version the BOM manages) has only a no-arg constructor; `assetProcessorService` is a plain `Object` property. And `AssetProcessorService` is a Grails service artefact of that plugin, so in an application with no plugin lifecycle there is no bean of it to inject either - the snippet needs the reader to declare that too, and the tag library also expects `grailsApplication` and a `tagLibraryLookup` from the `TagLibraryInvoker` trait. None of this is exercised: per your own comment on 48175cf3fd, `asset-pipeline-grails` is off the example's class path and the layout links the compiled files by name. The TIP at line 117 ("using all of this - precompiled views, the asset pipeline, ...") therefore points the reader at an application that does not demonstrate the section it closes.

Suggest either a neutral example that is actually covered - a `@grails.gsp.TagLib` class the application declares as a `@Bean`, which is what `StandaloneTagLibraryLookupSpec` and `TagLibraryLookupRegistrarSpec` test - or wiring the asset tag library in the example for real and documenting what that took. Keep the `@Artefact("TagLib")` sentence at line 85; that is covered.

### Docs: two rendering details in the same page

- `gspInSpringBoot.adoc:26,35`: `{version}` inside a `[source,groovy]` block is not substituted; listing blocks only get verbatim subs. `pluginSupport.adoc` uses `[source,groovy,subs="attributes"]` for the same line. Use `subs="attributes+"` here so the `// <1>` callouts keep working.
- `gspInSpringBoot.adoc:31`: `id 'org.apache.grails.gradle.grails-gsp'` with no version does not resolve in a plain Spring Boot build. Either `version '{version}'` on the line, or say that the version comes from `pluginManagement` / the gradle BOM.

### Test: `CoreGrailsPluginRegistrarSpec` leaves duplicates behind

`grails-core/src/test/groovy/org/grails/plugins/CoreGrailsPluginRegistrarSpec.groovy:221-222`: the feature removes the two Grails creators, the registrar under test puts them back (`GroovyAwareAutoProxyCreators.registerWithAopConfigUtils` adds anything the list does not contain), and then cleanup does `priorityList.addAll(removed)` - so every run appends a second copy of both classes to `APC_PRIORITY_LIST`. Harmless for `findPriorityForClass`, which uses `indexOf`, but the Javadoc at line 216 says the fork is left as it was found, and it is not. Restore only what is missing:

```groovy
cleanupActions << { removed.each { if (!priorityList.contains(it)) priorityList.add(it) } }
```

### Nit

`Sitemesh3LayoutTagLibSpec.groovy:80-89` pins the unencoded output of a tag library built with no `codecLookup`. It makes the wiring load-bearing, as the comment says, but it also turns the absence of a guard into specified behaviour. If the intent is "the lookup is required", a `@Autowired(required = true)` failure at context start is the stronger contract, and this case could go.

### Verified as correct

- `Sitemesh3LayoutTagLib`: `@Autowired` on the property lands on the field, as it does in `RenderSitemeshTagLib`; a Grails application is unaffected because the codecs plugin always contributes `codecLookup`. `StandaloneContextTest` fails with the annotation removed.
- `ReplaceViewResolverRegistrar` (`GspAutoConfiguration.java:428`): the registrar runs after user configuration classes are loaded, so a user `jspViewResolver` is seen and left alone; `TagLibraryLookupRegistrarSpec` covers both branches. A bean of that name registered by a *later* auto-configuration now hits `BeanDefinitionOverrideException` under Boot's default rather than being silently aliased over - same timing as the old `@Bean(name = {...})` form, and loud, which is the right failure. The alias is registered regardless of `spring.gsp.replaceViewResolverBean`, as before.
- `GspAutoConfigurationSpec`: the `views.properties` fixture names real `GroovyPage` subclasses with the metadata constants the compiler emits, so `findPage` goes through `GroovyPageCompiledScriptSource`; reverting either `setPrecompiledGspMap` or the `cleanUri` search path fails it in-module.
- `GroovyPageCompilerSpec` now writes generated sources under its `@TempDir`.
- `GroovyPagePluginFunctionalSpec`: the non-Grails project asserts `test` depends on both `compileGroovyPages` and `compileWebappGroovyPages`; the Grails project asserts neither. Passes against the branch.
- Example: session-scoped selection, cookie-keeping client in `JspViewTest`, and `StandaloneContextTest` pins `pluginManager`, `grailsConfigProperties` and `classLoader` (`GrailsApplication.CLASS_LOADER_BEAN`) absent. The stale Javadoc sentence is gone.
- `upgrading80x.adoc` no longer names `Sitemesh3EnvironmentPostProcessor`. The new page names only public API and configuration keys; the property defaults it lists match `GspAutoConfiguration`. The `toc.yml` entry sits under the GSP section next to `resources` and `makingChangesToADeployedApplication`, and the `====` heading level matches those siblings.
- README rewrite is accurate to the module.

### Process

Unchanged from round one: whether this lands on 8.0.x before RC1 is still open in the thread.

codeconsole:
@matrei Addressed in 16eb980bb8..76bfb3bd08.

**Tag library example** (76bfb3bd08). Replaced with a `@grails.gsp.TagLib` class declared as a `@Bean`, the shape `StandaloneTagLibraryLookupSpec` tests. The `@Artefact("TagLib")` sentence stays, followed by one noting that a plugin's tag library may need its collaborators declared as beans. The TIP now names what the example demonstrates - precompiled views, SiteMesh layouts, assets compiled by the asset pipeline, and a JSP alongside GSP - and no longer implies the tag library section.

**Rendering** (76bfb3bd08). The `build.gradle` listing is `[source,groovy,subs="attributes+"]` and the plugin id carries `version '{version}'`. In the built guide both lines render `8.0.0-SNAPSHOT` and the callouts are intact.

**`CoreGrailsPluginRegistrarSpec`** (0327078ca1). Cleanup restores only a creator that is missing, as suggested.

**Nit** (ec9a9c4928). The unencoded-output case is removed. The encoded-output case stays, and `StandaloneContextTest` still fails if the field is not wired.

matrei:
Follow-up on 76bfb3bd08, found while running the example with `bootRun`: the error page of a standalone application does not render.

### Bug: any error or 404 in a standalone application ends in `Circular view path [error]`

Open `http://localhost:8080/error` or any unmapped path in a browser and the response is Tomcat's own 500/404 page, with this in the log:

```
jakarta.servlet.ServletException: Circular view path [error]: would dispatch back to the current handler URL [/error] again. Check your ViewResolver setup!
```

Reproduced in the example with a `@SpringBootTest` on a random port sending `Accept: text/html` (a client sending no `Accept` gets Boot's JSON error body, which is why the example's tests do not see it). The view resolvers in the example's context, in the order the dispatcher consults them:

| bean | class | order |
|---|---|---|
| `gspViewResolver` | `GrailsSiteMeshViewResolver` | `LOWEST_PRECEDENCE - 20` |
| `beanNameViewResolver` | `BeanNameViewResolver` | `LOWEST_PRECEDENCE - 10` |
| `mvcViewResolver`, `defaultViewResolver` | `ViewResolverComposite`, `InternalResourceViewResolver` | `LOWEST_PRECEDENCE` |

`BasicErrorController.errorHtml` falls back to the view name `error`, which Boot serves from its whitelabel `error` bean through `BeanNameViewResolver`. The GSP resolver is asked first. There is no `error.gsp`, and `spring.gsp.jspEnabled` defaults to `true` (`GspAutoConfiguration.java:108`, applied at line 290), so `GroovyPageViewResolver.createFallbackView` (`GroovyPageViewResolver.java:258-273`) returns a JSTL view for `/error` without checking that any such JSP exists. That view forwards to `/error`, the URL being handled, and the servlet container rejects the loop. Boot's whitelabel page is never reached.

`GroovyPageViewResolver` itself is untouched by this PR (`grails-web-gsp` has no diff against 8.0.x); what the PR changes is that the standalone module now starts, so this is the first time the path is reachable. A Grails application never hits it because its errors are routed by `UrlMappings`, not by `BasicErrorController`.

Two things I checked before suggesting a fix:

- `spring.gsp.jspEnabled=false` makes `/error` and 404s render the whitelabel page, and `form.jsp` still renders through Boot's `defaultViewResolver` - but undecorated: the `<meta name="layout" content="main"/>` in `form.jsp` is honoured only when the JSP is served through the GSP resolver's fallback, which the SiteMesh view resolver wraps. So flipping the default is not free for the example's JSP demonstration.
- Adding `templates/error.gsp` to the example would make the page render, but only hides the module defect: every other standalone application without one gets the container's raw error page instead of Boot's.

Suggested fix, in the module: have the JSP fallback resolve only a JSP that exists. `createJstlView` can ask `servletContext.getResource(url)` (the example's `JspSupport.canServeJsp` does exactly this) and return `null` otherwise, which lets resolution continue to `BeanNameViewResolver` and the whitelabel `error` bean, or to Boot's `defaultViewResolver` for a real JSP. That keeps JSP decoration in the example and fixes the error page for every consumer. A test in the example - `GET /does-not-exist` with `Accept: text/html` is a 404 whose body is the whitelabel page (or an `error.gsp` if you add one to show it off) - would pin it.

A longer-term item, not for this PR: Boot resolves `error/404`, `error/5xx` templates through `TemplateAvailabilityProvider`s, and there is none for `.gsp`, so a standalone application cannot use per-status GSP error pages the way a Thymeleaf one can.

codeconsole:
@matrei Fixed in 5f11919ad4.

Reproduced with `Accept: text/html` on an unmapped path: Tomcat's 404 page, and `Circular view path [error]` in the log.

**Fix.** `GspAutoConfiguration` now contributes a `StandaloneGroovyPageViewResolver` whose `createJstlView` returns `null` unless `servletContext.getResource` finds the JSP, so `error` falls through to `BeanNameViewResolver` and the whitelabel page. The check is kept out of the shared `GroovyPageViewResolver`: `GroovyPagesGrailsPlugin` turns the same fallback on for every Grails application with JSTL present, and a Grails application never reaches this path, so its fallback is left as it was. A JSP that exists resolves as before, so `form.jsp` is still decorated by its layout (`JspViewTest` passes).

**Tests.** `ErrorPageTest` in the example: `GET /does-not-exist` with `Accept: text/html` is a 404 whose body is the whitelabel page. `StandaloneGroovyPageViewResolverSpec` covers both branches through `resolveViewName`. Both fail with the check reverted.

**Docs.** The guide page's configuration table lists `spring.gsp.jspEnabled` and what it now does.

Per-status GSP error pages through a `TemplateAvailabilityProvider` are left out of this PR, as suggested.

matrei (approved):
Fifth round, on 2033f0eaf7 against 8.0.x (merge base 3067d0a855 is the current tip of 8.0.x, so the branch is clean). One commit since round four, 5f11919ad4, which is the fix for the error-page defect from that round. CI on the head is green: 89 checks passed, 4 skipped (publish, docs, forge, wrapper).

Ran `:grails-gsp-spring-boot:test`, `:grails-gsp-spring-boot:codeStyle` and the gsp-spring-boot example's tests. All pass. Approving.

### Verified

- **The fix** (`StandaloneGroovyPageViewResolver`): the JSP fallback is kept in a subclass the standalone module alone constructs; `GroovyPageViewResolver` and `grails-web-gsp` are untouched, so a Grails application keeps the fallback it had. The override calls the base `createJstlView`, asks the servlet context for the view's URL (leading slash added where missing, `MalformedURLException` answered with "no"), and returns `null` when nothing is there. With `setCache(false)` on the base resolver, a `null` from `loadView` is "unresolved" to the dispatcher and the next resolver is asked - `BeanNameViewResolver` and Boot's whitelabel `error` bean, as the round-four table showed.
- **Both tests are load-bearing.** With the existence check disabled, `StandaloneGroovyPageViewResolverSpec` fails on the `error` case (a view comes back where `null` is expected) and `ErrorPageTest` fails with Tomcat's own `HTTP Status 404 – Not Found` body in place of the whitelabel page. Restored afterwards; the tree is clean.
- **JSP decoration survives**: `JspViewTest` still gets the `form.jsp` rendering through the GSP resolver's fallback, wrapped by the SiteMesh layout, because `/form.jsp` is a real servlet-context resource in the example.
- **The spec tests through the public surface**: `resolveViewName` on a resolver built from the auto-configuration's public `groovyPageLocator` bean method, with a `MockServletContext` over a `@TempDir` as the document root and the context closed in `cleanup`. Same construction pattern as `GspAutoConfigurationSpec`.
- **Docs**: the new `spring.gsp.jspEnabled` row in the configuration table matches the default in `GspAutoConfiguration` and describes the new behaviour without naming an internal class. The module README needs no change; the resolver's behaviour is described on the guide page.
- **Style**: `codeStyle` on the module is clean; the new Java class carries the license header and explicit imports.

### Nit, not blocking

`StandaloneGroovyPageViewResolver.servletContext()` re-derives what `WebApplicationObjectSupport.getServletContext()` already provides, and the base resolver already relies on that accessor to build every GSP view (`GroovyPageViewResolver.java:221`). Its two `null` branches are unreachable in practice, since the resolver cannot render anything without a web context. `getServletContext()` in place of the helper would drop a dozen lines.

### Observation, no change needed

A miss is cached like a hit: `CacheEntry.setValue(null)` marks the entry initialised, so a name with neither a template nor a JSP is remembered for `spring.gsp.view.cacheTimeout` while reloading is on, and for the life of the context when it is off. That is the same treatment a missing GSP always had, and it keeps the `error` lookup cheap, so it is the right trade; noting it only so the next reader does not expect a JSP dropped into a running production deployment to be picked up.

### Process

Unchanged: whether this lands on 8.0.x before RC1 is a release-management call still open in the thread, not something in the diff.

grails-core/src/main/groovy/grails/boot/config/GrailsEarlyPluginRegistrationPostProcessor.java
jdaugherty: @matrei  Are you really ok with the behavior that Application class is now required for SpringBootTest usage?

matrei: @jdaugherty 

Yes, and I have now run it rather than read it, because the question deserved that. I added four `@SpringBootTest` specs to the app3 example (which has the `loadafter` plugin registering `earlyPluginProbe` in `doWithSpring`, so the lifecycle is observable) and ran them on this branch and on the 8.0.x merge base 3067d0a855.

| Spec | 8.0.x (before) | this PR |
|---|---|---|
| `@SpringBootTest` | context fails: `BeanDefinitionOverrideException` on `grailsDomainClassMappingContext` | same failure, unchanged |
| `@SpringBootTest(properties = [allow-bean-definition-overriding=true, allow-circular-references=true])` | passes, full lifecycle | passes, full lifecycle: `grailsApplication`, `pluginManager`, plugin `doWithSpring` bean, `beanRegistrar` bean, application artefacts |
| `@SpringBootTest(classes = SbtLifecycleProbeConfig)` | context fails: circular reference creating `hibernateDatastore` from `openSessionInViewInterceptor` | passes, no Grails lifecycle |
| `@SpringBootTest(classes = [Application, SbtLifecycleProbeConfig])` | not run | fails like the bare form (same override exception) |

So, on what the gate does:

- The application class is not required for `@SpringBootTest` as such. The compiler puts `@SpringBootConfiguration` on `Application` (`ApplicationClassInjector`, same in 7.x), Boot finds it, and `resolveApplicationSourceClasses` recovers it from the registry. The second row shows the lifecycle then runs in full. `@Integration` names it outright.
- Only `@SpringBootTest(classes = SomeConfig)` is gated out. That form never had the lifecycle in 7.x: the lifecycle was the `grailsApplicationPostProcessor` `@Bean` on the application class, and 7.0.x registers no `ApplicationContextInitializer` for it. And in the M3 to M6 window where the early phase ran for every context, the third row shows what that form actually got in an application with GORM: a context that could not start, because plugin beans were registered into a plain `SpringApplication` that disallows circular references. This PR is what makes that test load.

Two things the runs turned up that I would ask for before this is settled:

1. The upgrade note is wrong where it says "Grails 7: the plugin lifecycle ran, and this context had a GrailsApplication and every plugin's beans" and "the test was given the whole plugin runtime". 7.x gave that form nothing Grails, and the milestones gave it a context that failed to start. The note should say so, and drop the "Grails 7" label.
2. Separate from this PR, and already on 8.0.x: a bare `@SpringBootTest` in a Grails application with Hibernate does not start unless the test sets `spring.main.allow-bean-definition-overriding=true` and `spring.main.allow-circular-references=true`, the two defaults `GrailsApp` applies and a plain `SpringApplication` does not. `@Integration` avoids it via `useMainMethod = ALWAYS`. The override is the Hibernate plugin's `grailsDomainClassMappingContext` (`hibernateDatastore.getMappingContext`) against the generated `DomainClassAutoConfiguration` `@Bean` of the same name. That is the real "SpringBootTest usage" gap in 8.0, and it deserves its own issue: either the generated auto-configuration guards that bean with `@ConditionalOnMissingBean`, or `GrailsPluginLifecycleInitializer` applies the two `GrailsApp` defaults when it decides the context is a Grails application.

A classpath scan for a `GrailsApplicationClass` would be the only way to give `@SpringBootTest(classes = SomeConfig)` the lifecycle anyway, since `grails.build.info` is written for plugin projects too. That would be new behaviour relative to 7.x, added to every Boot context with grails-core on the class path, for a form Boot itself treats as "use these sources only". I would rather not.

jdaugherty: @codeconsole if you can address the upgrade note, I'm good to merge this.

codeconsole:
@matrei @jdaugherty Addressed in bf09ab4b6ceda93eea7343277a08f53249ae5af6.

The upgrade note now distinguishes Grails 7 from the earlier Grails 8 milestones: a custom `@SpringBootTest(classes = SomeConfig)` without the application class did not receive the Grails plugin lifecycle in Grails 7. It also explains automatic application-class discovery, the existing Hibernate startup limitation, and how to apply the startup defaults using `useMainMethod = ALWAYS` or the two Spring Boot properties.

Also addressed the non-blocking nit by replacing the duplicate servlet-context helper with Spring's inherited `getServletContext()`. Expanded coverage for JSP paths with and without a leading slash, missing views, disabled JSP fallback, and JSP/error-page rendering under a non-root context path.

jdaugherty:
@codeconsole Thank you, I'm good to merge this when you think it's ready.  Iv'e released AssetPipeline 5.2.0-RC1 as well if you want to pull that into this PR.

codeconsole:
@jdaugherty Pulled in 5.2.0-RC1 (e07a60eee3..b40c318616), which replaces the example's hand-written asset wiring from your first comment with the library itself.

- **BOM** (e07a60eee3): `asset-pipeline-gradle.version` and `asset-pipeline-bom.version` to 5.2.0-RC1.
- **Example** (490f5e1e3f): `asset-pipeline-gsp` and the self-configuring `asset-pipeline-spring-boot` replace the `@Import(AssetPipelineService)`, and the layout uses `<asset:stylesheet>`/`<asset:javascript>`. `grails-web-url-mappings` is no longer excluded, since the asset pipeline's link generator needs it; the generator runs against the empty `grailsUrlMappingsHolder` from `GspAutoConfiguration`, which the example now exercises. `WebControllerTest` follows the rendered links, and `AssetLinkTest` checks they carry the context path under `/gsp`.
- **Docs** (b40c318616): the guide page covers the two modules and keeping `grails-web-url-mappings`.

Example suite, `:grails-gsp-spring-boot:test`, `validateDependencyVersions` and the style gate pass locally.

jdaugherty (approved):
Should be good to merge

PR #16185, opened 2026-08-21
Import grails-quartz

jdaugherty:
Per vote [1], this merges grails-quartz into grails-core.  I also have fixed *some* vulnerabilities as part of this merge since I had to scan the repo to merge. 

1. https://lists.apache.org/thread/6781tod7s48kscdbcmd3yr4m18t10kg2


jdaugherty:
The hsitory was incorrectly imported, redoing this.

jdaugherty:
The commits appear to be here now so this should be ready to review once tests pass.

jdaugherty:
FYI: Given that we voted already to merge this, I intend to merge by Sunday so it can be in the latest 7.x releases.  @matrei @jamesfredley  please let me know if you disagree with this approach.

grails-quartz/src/main/groovy/quartz/QuartzGrailsPlugin.groovy
```diff
@@ -0,0 +1,363 @@
...
+    def scm = [url: 'https://github.com/apache/grails-quartz']
+
+    Closure doWithSpring() {
+        { ->
+            Properties properties = loadQuartzProperties()
+
+            boolean hasHibernate = hasHibernate(manager)
+            def hasJdbcStore = properties['org.quartz.jdbcStore']?.toBoolean()
+            if (hasJdbcStore == null) {
+                hasJdbcStore = true
```
sbglasius: The jdbcStore fallbacks default to true when unset (`?.toBoolean()` yields null, then the guard assigns true) in doWithSpring, configureScheduler and scheduleJob, contradicting the documented default of false.
An app that overrides the quartz config block without a jdbcStore key (so plugin.yml's false is not in effect) gets a JDBC job store wired to the `dataSource` and `transactionManager` beans, and startup fails against a database with no QRTZ_* tables.

grails-quartz/src/main/groovy/quartz/QuartzGrailsPlugin.groovy
```diff
@@ -0,0 +1,363 @@
...
+
+            jobFactory = quartzJobFactory
+
+            // Global listeners on each job.
+            globalJobListeners = [ref(ExceptionPrinterJobListener.NAME)]
+        }
+    }
+
+    void onChange(Map<String, Object> event) {
+        def pluginEnabled = properties['org.quartz.pluginEnabled']?.toBoolean()
```
sbglasius: Missing loading properties?

```
Properties properties = loadQuartzProperties()
```

grails-quartz/src/main/groovy/quartz/QuartzGrailsPlugin.groovy
```diff
@@ -0,0 +1,363 @@
...
+            boolean hasHibernate = hasHibernate(manager)
+            def hasJdbcStore = properties['org.quartz.jdbcStore']?.toBoolean()
+            if (hasJdbcStore == null) {
+                hasJdbcStore = true
+            }
+
+            def pluginEnabled = properties['org.quartz.pluginEnabled']?.toBoolean()
+            if (pluginEnabled == null) {
+                pluginEnabled = true
+            }
```
sbglasius: Used several places, perhaps a method
```
boolean getPluginEnabled() {
	grailsApplication.config.getProperty('quartz.pluginEnabled', Boolean, true)
}

File: grails-quartz/src/main/groovy/quartz/QuartzGrailsPlugin.groovy [hunk 4]
```diff
@@ -0,0 +1,363 @@
...
+
+    Closure doWithSpring() {
+        { ->
+            Properties properties = loadQuartzProperties()
+
+            boolean hasHibernate = hasHibernate(manager)
+            def hasJdbcStore = properties['org.quartz.jdbcStore']?.toBoolean()
+            if (hasJdbcStore == null) {
+                hasJdbcStore = true
+            }
```
sbglasius: use `properties.getProperty('org.quartz.jdbcStor', true).toBoolean()` for default value

sbglasius (commented):
I know cleaning up was not part of the merge, but `QuartzGrailsPlugin` property access could really use some love.

sbglasius (approved):

jdaugherty:
@sbglasius based on your feedback, I addressed the plugin descriptor & other issues.  To be consistent, and because we've previously voted on listing author/developers, I've gone ahead and cleaned up a wider set of plugins to be consistent across all of the ones in grails-core

PR #16190, opened 2026-08-21
Stop the embedded MongoDB once

codeconsole:
Every run that starts a real mongod ends with this on a shutdown thread, against a build whose
tests all passed:

```
Exception in thread "Thread-77" de.flapdoodle.reverse.TearDownException: tearDown errors
	at de.flapdoodle.reverse.TransitionWalker.tearDown(TransitionWalker.java:298)
	at org.grails.datastore.gorm.mongodb.embedded.FlapdoodleMongoBackend$RunningMongod.stop
	Caused by: java.nio.file.NoSuchFileException: /tmp/temp--.../workingDir.../mongod.pid
```

The server is stopped twice. `EmbeddedMongoLifecycle` stops it when the application context closes,
and the JVM shutdown hook - which is there for a context that never closes - stops it again when the
process exits. Flapdoodle removes the process directory as it tears the server down, so the second
teardown fails on the files the first one deleted.

Stopping is idempotent now, and so is starting a server that is already running. The in-memory
backend already tolerated both; a spec covers each.

### DIFF
--- a/grails-data-mongodb/embedded/src/main/java/org/grails/datastore/gorm/mongodb/embedded/FlapdoodleMongoBackend.java
+++ b/grails-data-mongodb/embedded/src/main/java/org/grails/datastore/gorm/mongodb/embedded/FlapdoodleMongoBackend.java
@@ -142,9 +142,20 @@ public class FlapdoodleMongoBackend implements EmbeddedMongoBackend {
             return this.address.getPort();
         }
 
+        /**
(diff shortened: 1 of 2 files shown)


codeconsole:
Folded into #16192, which already carries this commit — one review rather than two. The stop-once fix belongs with that change anyway: the four specifications it moves onto an embedded server start and stop one routinely, so a teardown that fails the second time would be noise on every run.

PR #16192, opened 2026-08-21
Serve the transaction specifications from an embedded replica set

codeconsole:
Fixes the `MongoTransactionSpec` failures that take out "Build Grails with Groovy snapshot (shard 1)"
about half the time. The MongoDB container those specifications talk to dies partway through, and a
transaction needs a replica set, which is the only reason they reached for a container.

## An embedded replica set

```properties
embedded.mongodb.replica-set=rs0
```

Flapdoodle takes the name on the command line; what was missing is `replSetInitiate` and the wait for
the node to elect itself, which turns a server that refuses every write into one that answers. An
application that asks GORM for transactions gets a replica set without naming one:

```properties
grails.mongodb.transactional=true
```

`MongoTransactionSpec`, `MongoTransactionDisabledSpec`, `UnifiedMongoTransactionSpec` and
`GormSpringDataSessionSupportSpec` start that server through `EmbeddedReplicaSetSpec` instead of a
container, on a port asked for rather than fixed. No Docker, and no container to lose halfway
through. The `in-memory` backend refuses a replica set rather than starting a standalone server that
fails at the first transaction. Both are documented in the embedded MongoDB guide.

## The version the matrix asks for

`-PmongodbContainerVersion` reached the Gradle JVM and stopped there, so `mongodb-version:
['7.0','8.0']` ran 7.0.19 twice. It is forwarded as a system property from the MongoDB and
functional test configurations, and the embedded server takes the same version. CI also caches
`~/.embedmongo`, so mongod is downloaded once per branch rather than once per job.

## Two flaky tests

`LoginPage.login()` returned before the login had landed, so the page it read was sometimes still the
login form. It waits for the url to leave `/login/`.

A callback attached to a promise that completed while it was being attached asked the container to
start a second asynchronous cycle on a request already delivering a result. The container refuses -
`Calling [asyncStart()] is not valid for a request with Async state [DISPATCHING]` - and the refusal
failed the response. Such a request is not decorated. Measured against the async endpoints of
`gsp-sitemesh3`: 3-5 failures per 2,400 requests before, none in 9,600 after.

## Server lifecycle

Stopping is idempotent, since the application context stops the server and the JVM shutdown hook
follows it. A server reused across a devtools reload is restarted where it is reused rather than left
to its lifecycle bean, which Spring starts only after the context has refreshed - a datastore builds
its indexes before that, over the network. A reload carrying different settings gets a server that
matches them. One shutdown hook per JVM rather than one per server.

## What is not changed

The Mongo specifications that do not need a replica set stay on Testcontainers. This does not answer
why the container died; it removes the dependency on one.


sbglasius:
My AI reviewed this and found two things worth a follow-up. Rather than leave them as prose I pushed a single commit on top of this branch — cherry-pick it if you agree with the reasoning:

**Branch:** [`fix/embedded-mongo-replica-set-review-fixes`](https://github.com/apache/grails-core/tree/fix/embedded-mongo-replica-set-review-fixes) · **commit:** [`bd099a8`](https://github.com/apache/grails-core/commit/bd099a892f7f1ab341b449016765d96055bfee2a)

### 1. `RunningMongod.restart()` returns silently when the server is up

```java
public synchronized void restart() {
    if (this.running != null) {
        return;
    }
    ...
```

Via `EmbeddedMongoLifecycle` this guard is unreachable — the bean's `started` flag means `stop()`, which nulls `running`, always precedes `start()`. It is reachable through the public `RunningEmbeddedMongo` interface, and there it turns a restart into a silent nothing; the pre-guard code would at least have failed loudly on the port. The commit tears the old process down and starts the replacement instead, so the guard still covers the double-stop case it was added for without swallowing a real restart.

### 2. `EmbeddedReplicaSetSpec.freePort()` sets `reuseAddress` after the bind

`ServerSocket(0)` binds in the constructor, so the assignment on the next line does nothing. Setting it before the bind would not help either: it says nothing about two forks being handed the same ephemeral port, and the probe socket has no connections to leave in `TIME_WAIT`. The real exposure is that a port lost between the offer and mongod binding it fails the whole specification, because only a server the asking JVM started is ever reused from `STARTED`. The commit drops the dead line and retries the start on a fresh port up to three times.

### Verification

- `EmbeddedMongoLifecycleSpec` gains *a running server is replaced by a restart rather than fought with for its port*, which compares mongod's `serverStatus.pid` across the call. It fails against the guard it replaces and passes with the fix.
- `:grails-data-mongodb-embedded:test` — 41 tests, 0 failures.
- The specs extending `EmbeddedReplicaSetSpec` all pass: `MongoTransactionSpec` (10), `MongoTransactionDisabledSpec` (2), `UnifiedMongoTransactionSpec` (12), `GormSpringDataSessionSupportSpec` (1).
- `codeStyle` clean on both touched modules.

The retry path itself has no direct test — forcing a port collision would mean stubbing a private static in a base class, which did not seem worth reshaping the class for.

Nothing else in the diff looked wrong to me: the replica-set initiation, the `WebUtils.clearGrailsWebRequest` guard and the async decorator catch all read correctly.

grails-data-mongodb/embedded/src/main/java/org/grails/datastore/gorm/mongodb/embedded/ReplicaSetInitiator.java
```diff
@@ -0,0 +1,116 @@
...
+ * <p>A replica set of one is what a single machine can offer, and it is enough for everything a
+ * standalone server refuses: a multi-document transaction, a change stream, a causally consistent
+ * read. MongoDB does not configure it from the command line - the set is initiated by a command
+ * sent to the server it names - so this is the step between starting the process and handing the
+ * server to an application.
+ *
+ * <p>Every reference to the MongoDB driver lives here, so a standalone embedded server needs
+ * nothing of it on the class path.
+ *
+ * @author Grails
```
matrei: Remove `@author`?

build.gradle
```diff
@@ -112,6 +112,14 @@ subprojects {
...
                 Runtime.runtime.availableProcessors(), gradle.startParameter.maxWorkerCount))
+
+        // The server a test starts is chosen by the build, and the test reads it as a system
+        // property: a project property named on the command line reaches the Gradle JVM and stops
+        // there, so a matrix asking for one version and a test defaulting to another agree only by
+        // accident. CI passes -PmongodbContainerVersion for exactly this.
+        if (rootProject.hasProperty('mongodbContainerVersion')) {
+            testTask.systemProperty('mongodbContainerVersion', rootProject.property('mongodbContainerVersion'))
+        }
     }
```
matrei: Should this be moved to the mongodb specific test-config files in the `gradle` directory?

matrei (approved):

grails-data-mongodb/embedded/src/main/java/org/grails/datastore/gorm/mongodb/embedded/ReplicaSetInitiator.java
codeconsole: Removed. Its seven neighbours in that module carried the same `@author Grails` placeholder from #16095, so those are gone too, as an isolated commit (e6fa4eb) - tag lines only, nothing else touched.

One is left in the repository, in `grails-core/.../CompileStaticArtefactInjector.groovy`, which is outside this PR.

build.gradle
codeconsole: Moved, though to three files rather than two.

`mongodb-test-config.gradle` and `mongodb-forked-test-config.gradle` cover the `grails-data-mongodb` modules. They do not cover the example applications: `grails-test-examples/mongodb/**` apply `functional-test-config.gradle`, and they read `mongodbContainerVersion` as well - `SpringBootStartMongoExtension` and the test-data-service `Application` both do. Putting it only in the two mongodb configs would have sent those back to their hardcoded `7.0.19` while the matrix asked for 8.0, which is the thing that commit exists to fix. So it is in the functional config too, behind the `isMongo` flag that file already works out for the `onlyMongodbTests` gating.

Checked per task with `-PmongodbContainerVersion=8.0`: every `grails-data-mongodb-*` and `grails-test-examples-mongodb-*` test task now sees `8.0`, while `grails-core` and the non-mongo example applications see nothing at all - which was your point.

PR #16200, opened 2026-08-23
i18n: let a plugin that cannot rename downgrade the namespace check to a warning

codeconsole:
Follow-up to #16102.

## The gap

`generateI18nDescriptor` fails a plugin build when a base name sits outside the plugin's namespace. That default is right, and this PR does not change it — Spring resolves a base name to the first match on the classpath rather than merging, so two artifacts shipping the same base name shadow one another with nothing at runtime to say which won. Failing the plugin's own build puts the error in front of the only person who can rename the bundle.

What it lacks is an escape hatch, and `grails { i18n { basenames } }` is not one: declared base names are folded into the index *before* `validatePluginNamespace` runs, so a plugin cannot declare its way past the check.

That leaves a plugin that genuinely cannot rename — a base name that is part of a published contract, or a bundle vendored from somewhere else — with two options:

- `-x generateI18nDescriptor`, which ships no descriptor at all and therefore drops the plugin's messages entirely, or
- a fork.

## The change

```groovy
grails {
    i18n {
        enforceNamespace = false
    }
}
```

Downgrades the same diagnosis to a build warning and writes the descriptor as it stands. Default unchanged (`true`).

The wording is identical whether it throws or warns — what is wrong and how to fix it do not depend on the severity — so `validatePluginNamespace` becomes `describeNamespaceViolation`, returning the message or `null`, and the caller decides. Both the message and the upgrade guide say plainly that the opt-out silences the build, not the collision.

## Tests

- `GenerateI18nDescriptorTaskSpec` — the check is on by convention; `enforceNamespace = false` records the bundle and does not throw; the message names the opt-out and its caveat.
- `I18nDescriptorFunctionalSpec` + a new `i18n-descriptor-plugin-namespace-off` test project — the same colliding bundle as `i18n-descriptor-plugin-collision`, plus the `i18n` block, exercised end to end through `processResources`: the build succeeds, still reports the collision, and the descriptor records `basenames=messages`.

`:grails-gradle-plugins:check` passes (including `validatePlugins`), and `rat` passes.

## Context

Found while upgrading an app to 8.0.0-M6. Its Spring Security fork shipped the base name it inherited from the upstream plugin — a real collision, and the check caught it, so this is not a complaint about the rule. Renaming was the right fix there. The opt-out is for the cases where it isn't available.

### DIFF
--- a/grails-doc/src/en/guide/upgrading/upgrading80x.adoc
+++ b/grails-doc/src/en/guide/upgrading/upgrading80x.adoc
@@ -2577,6 +2577,22 @@ application's. Rename it into the plugin's namespace.
 
 Applications are unaffected — there is only one application, so its base names cannot collide.
 
+A plugin that genuinely cannot rename — a base name that is part of a published contract, or a bundle
+vendored from somewhere else — can downgrade the failure to a build warning:
+
+[source,groovy]
+----
+grails {
+    i18n {
+        enforceNamespace = false
+    }
+}
+----
+
+This silences the build, not the collision. Spring still resolves a base name to the first matching
+resource on the classpath, so the bundle will shadow, or be shadowed by, any other artifact shipping
+that name, and nothing at runtime reports which won. Rename where you can.
+
 A namespaced bundle still works under Grails 7, so a plugin does not need a separate source branch to
 support both. Grails 7 discovers a plugin's bundles by scanning the plugin jar's root for
 `*.properties` and matching locale suffixes; it never inspects the base name, so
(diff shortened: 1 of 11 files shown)


codeconsole:
Withdrawing this — the premise doesn't hold up.

I justified the opt-out with plugins that "cannot rename": a base name that is part of a published contract, or a vendored third-party bundle. Neither survives scrutiny. Message codes are the contract and a rename doesn't touch them; nothing consumes a plugin's base name (Grails 7 discovered bundles by scanning and never read the name, and applications override messages by code in their own bundle). A vendored bundle is vendored by contents, not by file name, so renaming it costs nothing.

The case that prompted this was a Spring Security fork still shipping the upstream plugin's base name. The check was right to catch it — a genuine collision — and the fix was one `git mv` loop with codes untouched, backward compatible with M5. When the remedy is that cheap, an opt-out mostly serves postponement, and postponement means shipping the collision.

It also cuts against the reason the check fails hard in the first place: a warning in a plugin's own build is read by nobody, and the damage lands in someone else's application later.

Not worth adding speculative configuration surface to a feature this new. Happy to revisit if someone turns up with a bundle they genuinely cannot rename.

PR #16209, opened 2026-08-24
Binding empties a raw collection in applications that never opted in to deny-by-default

codeconsole:
### Impact

**Silent data loss, in applications that did not opt in to the change that causes it.**

- A domain's raw `List` / `Set` / `Collection` has **every** element it binds replaced by an empty `java.lang.Object`. The values are gone before validation or persistence can see them.
- **Regression:** correct in 8.0.0-M5, broken in 8.0.0-M6. Not mentioned in `upgrading80x.adoc`.
- **Not gated by the opt-in.** #15947 states the mode is opt-in and that "unconfigured applications continue binding permissively." An application that never set `grails.databinding.legacyBindableDefault=false` loses the data anyway, because the instantiation happens in the `try` while `isDenyByDefaultEnabled()` guards only the `catch`. There is a test for both settings.
- **It surfaces nowhere near its cause.** On GORM for MongoDB it appears as `Can't find a codec for … java.lang.Object` from the BSON encoder, three modules from the data binder that destroyed the value. Tracing it took eliminating the MongoDB driver version, the embedded-MongoDB module, the codec registry and the datastore bean wiring first.

8.0.0-M6 is staged but not yet on Maven Central, so this is still fixable before it becomes immutable.

### Problem

A domain property declared as a **raw** collection loses its data during binding on 8.0.0-M6. Given:

```groovy
class TopicType {
    List statusFilters = []          // no type argument
}

new TopicType(statusFilters: [[label: 'Answered', param: 'status=resolved']])
```

every map element is replaced by an empty `new java.lang.Object()`. With GORM for MongoDB the write then fails:

```
org.bson.codecs.configuration.CodecConfigurationException:
    Can't find a codec for CodecCacheKey{clazz=class java.lang.Object, types=null}
  at CodecExtensions$ListCodec.encode(CodecExtensions.groovy:362)
  at BasicCollectionTypeEncoder.encode(BasicCollectionTypeEncoder.groovy:64)
  at BsonPersistentEntityCodec.encode(BsonPersistentEntityCodec.groovy:229)
```

The exception is incidental — it is the first thing to notice that the elements are no longer maps. Without a codec in the path the empty objects would simply be persisted, so the data is lost either way. This worked on 8.0.0-M5.

### Cause

#15947 changed collection binding so a `Map` element is instantiated as the component type and bound through the allowlist, rather than passed to a map constructor:

```diff
- itemsWhichNeedBinding << item
+ def instance = instantiateAndBindNestedOrUseMapConstructor(referencedType, item, itemBindingSource, ...)
+ if (instance != null) { itemsWhichNeedBinding << instance }
```

That is right for a real nested type, but the component type is not always one. `Basic#componentType` falls back to `Object.class` when a property carries no generic signature, so a raw collection arrives with `referencedType == Object`. `Object` has a public no-arg constructor, so `getDeclaredConstructor().newInstance()` succeeds, and `Object` declares no properties, so `bindNested` has nowhere to put the map's contents.

It is not gated by the opt-in, either: the instantiation is in the `try`, while `isDenyByDefaultEnabled()` guards only the map-constructor fallback in the `catch`. The data is destroyed with the hardening on and with it off — there is a test for both.

### Fix

Three places bind an element into a collection, and the other two already keep the item when the element type is assignable from it:

- the array branch — `if (item == null || componentType.isAssignableFrom(item.getClass()))`
- the `Map` branch — `if (item == null || referencedType.isAssignableFrom(item.getClass()))`
- `SimpleDataBinder` also asks, at `genericType.isAssignableFrom(val?.getClass())`

Only the collection branch went straight to instantiating. This asks the same question there:

```groovy
if (item == null || referencedType.isAssignableFrom(item.getClass())) {
    itemsWhichNeedBinding << item
} else if (item instanceof Map || item instanceof DataBindingSource) {
    ...
}
```

`Object` is assignable from everything, so a raw collection's elements are kept. No special case for `Object`, no change to `instantiateAndBindNestedOrUseMapConstructor`, and no change for any other component type — a `Map` element whose target is a real nested type is not assignable from it and still instantiates and binds through the allowlist. `bindable: false` is unaffected.

That is the whole production change: one branch, in one file.

### Scope

Only GORM domain classes, and only raw `List` / `Set` / `Collection` properties whose elements are maps.

A non-domain target (command object, `Validateable`) is unaffected: `SimpleDataBinder` resolves a component type only from a `ParameterizedType`, so a raw field yields `null`, and `GrailsWebDataBinder` reaches `Basic#componentType` only through a `PersistentEntity`. With `referencedType` null the branch is skipped by its own guard. A raw `Map` property is likewise unaffected, because its branch already asks the assignability question.

### Tests

Five tests in `GrailsWebDataBinderSpec`, against a `RawCollectionContainer` domain. **Every one of them fails on an unmodified `8.0.x` binder and passes with the fix**, each covering a distinct way into the branch:

| test | what it covers |
|---|---|
| raw `List` of maps | the reported case |
| raw `Set` of maps | the same branch via a different collection type |
| raw `Collection`-typed property | a third declared type |
| `DataBindingSource` elements | the other item shape the branch instantiates |
| deny-by-default enabled, property allowlisted | the loss is not gated by the opt-in |

The failures report exactly the production symptom, e.g.

```
obj.rawList[0] instanceof Map
|   |      |   false
|   |      <java.lang.Object@32a074ed> (java.lang.Object)
```

In the deny-by-default case `size() == 1` holds first, so the property does bind and the failure is the data being destroyed rather than the binding being denied.

`:grails-databinding-core:test`, `:grails-web-databinding:test` (including `DenyByDefaultConfigSpec`) and the `grails.web.databinding.*` suite in `grails-test-suite-persistence` are all green.

Also verified end to end against the application that hit this: with this jar built at `8.0.0-M6` and substituted into an otherwise stock M6 app, the raw-collection domain that previously failed boots and persists its maps intact.

### DIFF
--- a/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
+++ b/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
@@ -505,7 +505,7 @@ class GrailsWebDataBinder extends SimpleDataBinder {
             } else if (Collection.isAssignableFrom(metaProperty.type)) {
                 def referencedType = getReferencedTypeForCollection(propName, obj)
                 if (referencedType) {
-                    def listValue
+                    List listValue
                     if (val instanceof List) {
                         listValue = (List) val
                     } else if (val instanceof GPathResultMap && ((GPathResultMap) val).size() == 1) {
--- a/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
+++ b/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
@@ -634,6 +634,16 @@ class GrailsWebDataBinder extends SimpleDataBinder {
 
     private Object instantiateAndBindNestedOrUseMapConstructor(Class referencedType, Object value,
             DataBindingSource source, List includeList, DataBindingListener listener) {
+        if (referencedType == null || referencedType == Object) {
+            // A raw collection -- List/Set/Map written without a type argument -- reports Object as
+            // its component type: Basic#componentType falls back to Object.class when a property
+            // carries no generic signature. Object declares no properties, so instantiating one and
+            // binding into it has nowhere to put the element's data and the element silently becomes
+            // an empty Object. A value that is never used as a property source cannot mass-assign
+            // anything, so it is kept as it stands, which is also how these collections bound before
+            // deny-by-default.
+            return value
+        }
         def instance
         try {
             instance = referencedType.getDeclaredConstructor().newInstance()
--- a/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
+++ b/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
@@ -545,7 +545,17 @@ class GrailsWebDataBinder extends SimpleDataBinder {
                                 }
                             }
                             if (persistentInstance == null) {
-                                if (item instanceof Map || item instanceof DataBindingSource) {
+                                if (item == null || referencedType.isAssignableFrom(item.getClass())) {
+                                    // Already of the element type, so there is nothing to instantiate
+                                    // and nothing to bind into. A raw collection always lands here:
+                                    // Basic#componentType falls back to Object.class when a property
+                                    // carries no generic signature, and Object is assignable from
+                                    // everything. The array branch above and the Map branch below ask
+                                    // the same question before instantiating; only this one did not,
+                                    // so a map element was replaced by an empty Object and its
+                                    // contents were dropped.
+                                    itemsWhichNeedBinding << item
+                                } else if (item instanceof Map || item instanceof DataBindingSource) {
                                     DataBindingSource itemBindingSource = item instanceof DataBindingSource ?
                                             (DataBindingSource) item : new SimpleMapDataBindingSource((Map) item)
                                     def instance = instantiateAndBindNestedOrUseMapConstructor(
@@ -634,16 +644,6 @@ class GrailsWebDataBinder extends SimpleDataBinder {
 
     private Object instantiateAndBindNestedOrUseMapConstructor(Class referencedType, Object value,
             DataBindingSource source, List includeList, DataBindingListener listener) {
-        if (referencedType == null || referencedType == Object) {
-            // A raw collection -- List/Set/Map written without a type argument -- reports Object as
-            // its component type: Basic#componentType falls back to Object.class when a property
-            // carries no generic signature. Object declares no properties, so instantiating one and
-            // binding into it has nowhere to put the element's data and the element silently becomes
-            // an empty Object. A value that is never used as a property source cannot mass-assign
-            // anything, so it is kept as it stands, which is also how these collections bound before
-            // deny-by-default.
-            return value
-        }
         def instance
         try {
             instance = referencedType.getDeclaredConstructor().newInstance()
--- a/grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy
+++ b/grails-databinding-core/src/main/groovy/grails/databinding/SimpleDataBinder.groovy
@@ -426,6 +426,16 @@ class SimpleDataBinder implements DataBinder {
     }
 
     protected Object instantiateAndBindOrUseMapConstructor(Class referencedType, Map values, DataBindingListener listener) {
+        if (referencedType == null || referencedType == Object) {
+            // A raw collection -- List/Set/Map written without a type argument -- reports Object as
+            // its component type: Basic#componentType falls back to Object.class when a property
+            // carries no generic signature. Object declares no properties, so instantiating one and
+            // binding into it has nowhere to put the element's data and the element silently becomes
+            // an empty Object. A value that is never used as a property source cannot mass-assign
+            // anything, so it is kept as it stands, which is also how these collections bound before
+            // deny-by-default.
+            return values
+        }
         def instance
         try {
             instance = referencedType.getDeclaredConstructor().newInstance()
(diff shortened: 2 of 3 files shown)


grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
```diff
@@ -545,7 +545,17 @@ class GrailsWebDataBinder extends SimpleDataBinder {
                                 }
                             }
                             if (persistentInstance == null) {
-                                if (item instanceof Map || item instanceof DataBindingSource) {
+                                if (item == null || referencedType.isAssignableFrom(item.getClass())) {
```
matrei, line 548: IntelliJ complains: "Cannot resolve symbol 'getClass'"
`inferred type: ?`

```suggestion
                                if (item == null || referencedType.isAssignableFrom((Object) item.getClass())) {
```

matrei (approved):

grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
codeconsole:
```
BEFORE 548:  if (item == null || referencedType.isAssignableFrom(item.getClass())) {
AFTER  548:  if (item == null || referencedType.isAssignableFrom((Object) item.getClass())) {

GrailsWebDataBinder.groovy: 548: [Static type checking] -
    Cannot find matching method java.lang.Class#isAssignableFrom(java.lang.Object).
1 error
BUILD FAILED
```

matrei: Sorry I think the cast must be in parentheses.

codeconsole:
Exception downstream 
```
org.bson.codecs.configuration.CodecConfigurationException: Can't find a codec for CodecCacheKey{clazz=class java.lang.Object, types=null}.
        at org.bson.internal.ProvidersCodecRegistry.lambda$get$0(ProvidersCodecRegistry.java:86)
        at org.bson.internal.ProvidersCodecRegistry.get(ProvidersCodecRegistry.java:79)
        at org.bson.internal.ChildCodecRegistry.get(ChildCodecRegistry.java:68)
        at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
        at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
        at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
        at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
```

matrei:
> Exception downstream 
> ```
> org.bson.codecs.configuration.CodecConfigurationException: Can't find a codec for CodecCacheKey{clazz=class java.lang.Object, types=null}.
>         at org.bson.internal.ProvidersCodecRegistry.lambda$get$0(ProvidersCodecRegistry.java:86)
>         at org.bson.internal.ProvidersCodecRegistry.get(ProvidersCodecRegistry.java:79)
>         at org.bson.internal.ChildCodecRegistry.get(ChildCodecRegistry.java:68)
>         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
>         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
>         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
>         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
> ```

Ok, just ignore my comment, sorry for not testing it. Must be a false positive in IntelliJ.

codeconsole:
> > Exception downstream
> > ```
> > org.bson.codecs.configuration.CodecConfigurationException: Can't find a codec for CodecCacheKey{clazz=class java.lang.Object, types=null}.
> >         at org.bson.internal.ProvidersCodecRegistry.lambda$get$0(ProvidersCodecRegistry.java:86)
> >         at org.bson.internal.ProvidersCodecRegistry.get(ProvidersCodecRegistry.java:79)
> >         at org.bson.internal.ChildCodecRegistry.get(ChildCodecRegistry.java:68)
> >         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
> >         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
> >         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:63)
> >         at org.bson.codecs.BsonDocumentWrapperCodec.encode(BsonDocumentWrapperCodec.java:29)
> > ```
> 
> Ok, just ignore my comment, sorry for not testing it. Must be a false positive in IntelliJ.

@matrei that wasn't based on your suggestion. I was just adding the more complete stack trace from before the PR

What is going on with IntelliJ? why does it report things that are not issues? 

I see these unrelated 3 errors in the same file.
```
Expected '?', found 'grails.databinding.TypedStructuredBindingEditor'
Expected '?', found 'grails.databinding.converters.ValueConverter'
Expected '?', found 'grails.databinding.converters.FormattedValueConverter'
```

Are we wanting to code to IntelliJ complaints even though gradle compiles everything fine?

codeconsole:
@matrei if I change line 508
```groovy
def listValue
```
to
```groovy
List listValue
```

it get's rid of
```
Cannot resolve symbol 'getClass'
```

but I still have the existing
```
Expected '?', found 'grails.databinding.TypedStructuredBindingEditor'
Expected '?', found 'grails.databinding.converters.ValueConverter'
Expected '?', found 'grails.databinding.converters.FormattedValueConverter'
```

So I am just trying to understand why IntellJ complains about things gradle or the groovy compiler does not and whether or not we should care or fix everything IntelliJ complains about or if we need to fix the Groovy Grails plugin

grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy
codeconsole: Pushed as `f408864` — `def listValue` → `List listValue` on line 508, where the value is introduced. That leaves `item.getClass()` uncast, matching the array branch above and the Map branch below.

Thanks for catching it.

PR #16213, opened 2026-08-24
ci: give each pull request its own release drafter run

codeconsole:
Every pull request against a release branch computes the same release-drafter concurrency group, because the group is keyed on the branch it targets:

```yaml
group: release-drafter-${{ github.event.pull_request.base.ref || github.ref_name }}
cancel-in-progress: true
```

Two pull requests pushed within seconds of each other therefore share one slot, and the later one cancels the earlier:

```
05:30:24  feat/hiddenmethod-handler-mapping        cancelled   (#16183)
05:30:36  perf/urlmapping-name-resolution-8.0.x    success
```

The evicted pull request shows a cancelled check with nothing wrong with it, and never receives the labels the run exists to apply. Anything that raises several pull requests at once — rebasing a stack, merging a branch into each of them — makes it routine.

Queueing instead of cancelling would not fix it: GitHub keeps one pending run per group and cancels the one pending before it, so a sweep of ten pull requests would still show eight cancellations, of queued runs rather than running ones.

## The group is the pull request

```yaml
group: release-drafter-${{ github.event.pull_request.number || github.ref_name }}
```

Pull requests no longer meet. `cancel-in-progress` still does the right thing within one subject: pushing twice to the same pull request supersedes its own earlier run, and pushes to a release branch still share a group by branch name.

## A pull request run labels; it does not draft

```yaml
disable-releaser: ${{ github.event_name == 'pull_request' }}
```

They shared a slot for a reason — two runs creating the same draft at once is how the duplicate drafts described in that file's header appear — so this removes the reason rather than working around it.

Release notes are built from **merged** pull requests. An open one contributes nothing to the draft, so the releaser was rewriting the draft it had already written. Checked against the live `v8.0.0` draft: it lists merged pull requests only, which also corrects the claim in the workflow comment that the draft "picks up new PRs as soon as they are opened".

Labels still apply on every pull request event. The draft is written by pushes to the release branch, which is when a merge can actually change it.

`./gradlew validateActions` passes.

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -38,16 +38,19 @@ on:
   push:
     branches:
       - '[0-9]+.[0-9]+.x'
-  # Runs on PRs whose BASE branch is a release branch so the autolabeler can
-  # apply labels (bug/feature/docs/...). The draft itself is not written here:
-  # release notes are built from MERGED pull requests, so a run for an open one
-  # can only rewrite the draft it already wrote. Feature-to-feature PRs (e.g.
-  # fix/foo -> feat/bar) are intentionally excluded - they cannot affect any
-  # release.
-  pull_request:
-    types: [opened, reopened, synchronize, labeled]
-    branches:
-      - '[0-9]+.[0-9]+.x'
+  # There is deliberately no `pull_request` trigger. It was here so the autolabeler could
+  # label pull requests, and the action pinned below cannot do that: release-drafter 7 split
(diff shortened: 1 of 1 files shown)


jdaugherty:
We intentionally changed the release drafter to not overlap b/c if a release gets posted or changed while doing a release, it causes the release to fail.

But we probably could go away from canceling the in progress if it's from a PR.  That seems reasonable?

jdaugherty:
@codeconsole I think we should either close this PR or we should address the original concern with it conflicting with releases - we cannot run the drafter while a release is pending for that version.

PR #16215, opened 2026-08-24
fix: fill the login form until the values stay filled (does not fix #16217)

### LINKED ISSUE #16217: Flaky: UserControllerSpec > User list in grails-test-examples/scaffolding submits the login form without a session cookie
`grails-test-examples/scaffolding` fails intermittently in CI, on four occasions between 24 and 25 August 2026, on four unrelated branches. The root cause is not yet known; this issue records what is established so far.

## The test

`com.example.UserControllerSpec > User list` — `grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/UserControllerSpec.groovy`

```groovy
void "User list"() {
    when: 'an unauthenticated user requests the user list and signs in when prompted'
    via(UserListPage)
    at(LoginPage).login()

    then: 'the saved request redirects to the user list'
    at(UserListPage)
    scaffoldTable
}
```

It fails inside `LoginPage.login()`, on the wait that follows the click:

```groovy
void login(String username = 'test@grails.org', String password = 'letmein') {
    this.username = username
    this.password = password
    loginButton.click()
    waitFor { title != pageTitle && $('input', name: 'username').empty }   // times out here
}
```

```
geb.waiting.WaitTimeoutException: condition did not pass in 30 seconds (failed with exception)
    at com.example.pages.LoginPage.login(LoginPage.groovy:39)
    at com.example.UserControllerSpec.User list(UserControllerSpec.groovy:48)

    Caused by: Assertion failed:

    title != pageTitle && $('input', name: 'username').empty
    |     |  |         |
    |     |  |         false
    |     |  'Please sign in'
    |     false
    'Please sign in'
```

That is the trace from #16150, #16184 and #16209, which are identical to each other: the same
condition, the same thirty seconds, `LoginPage.groovy:39` in `login_closure1`, and
`UserControllerSpec.groovy:48`.

The fourth, #16215, differs in one respect: `LoginPage.groovy:43`, in `login_closure2`. That branch
carries a proposed change to this page object which adds lines above the wait, so the same wait on
the same condition reports a different line and closure. It is the same failure, not a different
one - and worth noting because it shows the wait that fails is the one after the click, on a branch
where the form is verified as filled before clicking.

## Occurrences

| Date (UTC) | PR | Job | Duration |
|---|---|---|---|
| 2026-08-24 05:34 | #16150 | [Build Grails with Groovy snapshot (shard 2)](https://github.com/apache/grails-core/actions/runs/32693786704/job/97332626174) | 55m |
| 2026-08-24 05:34 | #16184 | [Functional Tests (Java 25, indy=false)](https://github.com/apache/grails-core/actions/runs/32693980319/job/97332597354) | 59m |
| 2026-08-24 21:35 | #16209 | [Functional Tests (Java 25, indy=false)](https://github.com/apache/grails-core/actions/runs/32780264043/job/97600561629) | 51m |
| 2026-08-24 23:29 | #16215 | [Functional Tests (Java 21, indy=true)](https://github.com/apache/grails-core/actions/runs/32789589183/job/97628349227) | 43m |

None of those four branches touches this application, Spring Security, or Geb. It has been seen on Java 21 and Java 25, with indy on and off, and in two different workflows. In each run the other two specifications of the same application - `BookControllerSpec` and `UserCommunityControllerSpec`, which sign in the same way - passed.

## What the browser was looking at

`@ContainerGebConfiguration(reporting = true)` means Geb captures the page at the failure. Three of
the four runs upload it - #16184, #16209 and #16215 - and those three are the same to the byte. The
#16150 run is in a workflow that uploads no Geb report, so there is no capture for it:

- 1089 bytes, `<title>Please sign in</title>` - Spring Security's generated login page
- both fields **empty**, showing their placeholders
- the focus ring still on the username field, where `autofocus` puts it on load
- **no error**: not `/login?error`, no "Bad credentials" banner
- a `_csrf` hidden field present

Two things follow. The page was loaded *after* the credentials were typed, because a page that had been typed into would hold the values and would have the focus on the second field. And the request that produced it was not a rejected authentication attempt, because that renders `?error` with a banner.

## What the failure is not

Determined by injecting faults locally into the same test and comparing the captured page:

| Injected fault | Page produced | Matches CI |
|---|---|---|
| Session cookie deleted between filling and submitting | 1089 bytes, login page, no error, autofocus | **yes, byte for byte** |
| Valid session, `_csrf` field overwritten with a stale value | 283 bytes, `Whitelabel Error Page` (403) | no |

So the form reaches the server **without a session cookie**, rather than with a stale or mismatched CSRF token. Missing session and wrong token are answered differently: with no session Spring Security treats the caller as anonymous and returns the login page, while a wrong token on a live session is denied with a 403.

It is also not the credentials or the user data: `UserService.loadUserByUsername` looks the user up by email, and a lookup failure would produce `?error`, which is absent.

## What is still unknown

Why the browser has no session cookie when the form is submitted. The specification clears cookies in `setup()`:

```groovy
void setup() {
    clearCookiesQuietly()
}
```

and `ContainerGebSpec` holds `@Shared static GebTestManager testManager`, so the browser is shared between specifications in a worker. Whether the cookie is cleared, never set, scoped to a different host than the one the form posts to, or lost some other way is not established.

A consequence worth noting for anyone attempting a fix: if the session is genuinely gone, Spring Security's saved request goes with it, so simply submitting the form again is not enough - the login then succeeds but lands on `/` rather than the user list, and the specification fails on `at(UserListPage)` instead. Confirmed locally.

## Reproducing

Not reproduced on demand. Four consecutive local runs of `:grails-test-examples-scaffolding:integrationTest` passed. The failure signature can be produced deliberately by deleting the browser's cookies between filling the form and clicking, which is how the comparison above was made.

## Suggested next step

Capture the browser's cookies and `currentUrl` into the Geb report when that wait fails. One further occurrence would then distinguish "no cookie at all" from "a cookie scoped to another host", which is what the fix depends on.

## Other CI flakes tracked separately

- #16218 - the whole-repository build stops after its last task and holds a runner until the six hour kill.
- #16219 - `AsyncPromiseSpec` times out waiting for a response from an endpoint that is bounded well below the client timeout.

Different tests and different mechanisms; listed so that anyone working through CI flakiness can see what else is open.

codeconsole:
> **This does not fix the flake it was opened for, and the mechanism it originally described was wrong.** The failure is tracked in #16217. Draft, pending a decision on whether any of this is worth keeping on its own merits.

## What this claimed, and why it was wrong

It claimed `UserControllerSpec > User list` fails because the login page is replaced under the driver while the form is being filled, leaving two empty `required` fields that the browser refuses to submit.

That was asserted from the captured page rather than tested, and this branch's own CI disproved it: the same test failed here, at `login_closure2` - the wait *after* the click. The fill retry added by this branch had already passed, so the form was filled and the click did submit. The failure is downstream of anything this changes.

## What is actually known

Established afterwards by injecting faults locally and comparing the captured page with CI's, byte for byte (see #16217):

| Injected fault | Page produced | Matches CI |
|---|---|---|
| Session cookie deleted between filling and submitting | 1089 bytes, login page, no error, autofocus | yes, byte for byte |
| Valid session, `_csrf` overwritten with a stale value | 283 bytes, Whitelabel Error Page (403) | no |

The form reaches the server **without a session cookie**. Why is not known. Resubmitting is not a fix either: without the session the saved request is gone too, so a successful retry lands on `/` and the test then fails on `at(UserListPage)`.

## What is still in this branch

Two things that stand or fall on their own merits, neither of which addresses the failure above:

- filling the form until the values stay, which guards against a document replaced mid-fill. The mechanism is real - an empty `required` form is silently unsubmittable - but it has not been observed happening in this application.
- the misc functional test application's page objects setting the fields rather than appending to them, so a second attempt cannot double the text.

Both are hardening against something unobserved. On the evidence available they should not be merged as a fix for #16217, and arguably not at all.

### DIFF
--- a/grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LoginPage.groovy
+++ b/grails-test-examples/scaffolding/src/integrationTest/groovy/com/example/pages/LoginPage.groovy
@@ -31,11 +31,25 @@ class LoginPage extends NavigationPage {
     }
 
     void login(String username = 'test@grails.org', String password = 'letmein') {
-        this.username = username
-        this.password = password
+        // Filled again until the values stay. This page is rendered because a protected page
+        // asked for it, and the document can be replaced under the driver after it is first
+        // parsed - taking the typed values with it. What is left is a form whose two required
+        // fields are empty, which the browser will not submit, so nothing reaches the server
+        // and the wait below waits for a page that will never change.
+        waitFor { fillCredentials(username, password) }
         loginButton.click()
         // Wait for a definitive authenticated signal: the login page must be fully replaced
         // (title changed AND the login form is gone), not merely a transient title change.
         waitFor { title != pageTitle && $('input', name: 'username').empty }
+    }
+
+    /**
+     * @return whether the credentials are in the form now, which is false while a replaced
+     *         document is still being filled in again
+     */
+    private boolean fillCredentials(String username, String password) {
+        this.username = username
+        this.password = password
+        this.username.value() == username && this.password.value() == password
     }
 }
--- a/grails-test-examples/spring-security/core/misc-functional-test-app/group/src/integration-test/groovy/demo/LoginPage.groovy
+++ b/grails-test-examples/spring-security/core/misc-functional-test-app/group/src/integration-test/groovy/demo/LoginPage.groovy
@@ -33,8 +33,11 @@ class LoginPage extends Page {
     }
 
     void login(String username, String password) {
-        usernameInputField << username
-        passwordInputField << password
+        // Set rather than appended, and set again until the values stay: this page is rendered
+        // because a secured page asked for it, and a document replaced under the driver after it
+        // was first parsed takes the typed values with it, leaving a form that submits nothing
+        // useful. Appending would also double the text on a second attempt.
+        waitFor { fillCredentials(username, password) }
         loginButton.click()
         // Submitting the form is a navigation, and the browser goes on showing this page until the
         // response lands. Returning before that leaves whatever the specification asserts next
@@ -42,6 +45,16 @@ class LoginPage extends Page {
         // left, where the title would only say the page reads differently - and a page reads
         // differently in another language.
         waitFor { !browser.currentUrl.contains('/login/') }
+    }
+
+    /**
+     * @return whether the credentials are in the form now, which is false while a replaced
+     *         document is still being filled in again
+     */
+    private boolean fillCredentials(String username, String password) {
+        usernameInputField.value(username)
+        passwordInputField.value(password)
+        usernameInputField.value() == username && passwordInputField.value() == password
     }
 
     @Override
--- a/grails-test-examples/spring-security/core/misc-functional-test-app/roles/src/integration-test/groovy/demo/LoginPage.groovy
+++ b/grails-test-examples/spring-security/core/misc-functional-test-app/roles/src/integration-test/groovy/demo/LoginPage.groovy
@@ -33,8 +33,11 @@ class LoginPage extends Page {
     }
 
     void login(String username, String password) {
-        usernameInputField << username
-        passwordInputField << password
+        // Set rather than appended, and set again until the values stay: this page is rendered
+        // because a secured page asked for it, and a document replaced under the driver after it
+        // was first parsed takes the typed values with it, leaving a form that submits nothing
+        // useful. Appending would also double the text on a second attempt.
+        waitFor { fillCredentials(username, password) }
         loginButton.click()
         // Submitting the form is a navigation, and the browser goes on showing this page until the
         // response lands. Returning before that leaves whatever the specification asserts next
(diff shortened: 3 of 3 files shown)


codeconsole:
Closing. This does not fix #16217, and the mechanism it was opened on turned out to be wrong - the failure reproduced on this very branch, at the wait *after* the click, with the form verified as filled.

What is left here guards against a document being replaced while the form is filled. That can happen in principle, and an empty `required` form is silently unsubmittable, but it has not been observed in this application - so it is hardening against something nobody has seen, which is not worth carrying.

The evidence is recorded in #16217: the form reaches the server without a session cookie, established by comparing the captured page against injected faults. Anyone picking that up should start there rather than from this branch.

The one piece worth doing separately is the diagnostic that issue asks for - capturing the browser's cookies and url into the Geb report when the wait fails - which would settle the open question on the next occurrence.

Issue #16219, opened 2026-08-25
Flaky: AsyncPromiseSpec times out waiting for a response the endpoint cannot have taken that long to produce


codeconsole:
`functionaltests.async.AsyncPromiseSpec` fails intermittently, and not always on the same feature, when the HTTP client gives up waiting for a response:

```
java.net.http.HttpTimeoutException: request timed out
    at java.net.http/jdk.internal.net.http.HttpClientImpl.send(HttpClientImpl.java:953)
    at org.apache.grails.testing.http.client.HttpClientSupport$Trait$Helper.send(HttpClientSupport.groovy:1133)
    at org.apache.grails.testing.http.client.HttpClientSupport$Trait$Helper.http(HttpClientSupport.groovy:203)
    at functionaltests.async.AsyncPromiseSpec.<feature>(AsyncPromiseSpec.groovy:<line>)
```

The client timeout is 120 seconds (`grails.http.client.timeout`, set in `gradle/functional-test-config.gradle`).

## Occurrences

| Date (UTC) | Branch | Job | Feature that timed out |
|---|---|---|---|
| 2026-07-29 20:32 | `feat/8.0.x-legacy-command-compat` | [Hibernate7 Functional Tests (Java 21, indy=true)](https://github.com/apache/grails-core/actions/runs/30488829586/job/90701275362) | conditional async uses sync mode when requested |
| 2026-07-29 21:01 | **`8.0.x`** | [Functional Tests (Java 21, indy=false)](https://github.com/apache/grails-core/actions/runs/30490747146/job/90707733744) | multi-stage process reports all stages |
| 2026-08-22 17:23 | `fix/embedded-mongo-replica-set` | [Build Grails with Groovy snapshot (shard 2)](https://github.com/apache/grails-core/actions/runs/32587420540/job/97066257811) | conditional async uses sync mode when requested |
| 2026-08-24 23:53 | `ci/…` on #16214 | [Build Grails with Groovy snapshot (shard 0)](https://github.com/apache/grails-core/actions/runs/32791117146/job/97633380610) | async service processes string input |

Found by scanning the failed job logs of every `CI` and `CI - Groovy Joint Validation Build` run between 2026-07-20 and 2026-08-24 - 308 logs, none unavailable - for that spec together with that exception. These four are all of them in that period.

[...]

Issue #16218, opened 2026-08-25
CI: the whole-repository build hangs after its last task, holding a runner until the 6 hour kill


codeconsole:
The job named `Neo4j Functional Tests (Java 25, indy=false)` stops producing output after its last task and never finishes, holding a runner until GitHub kills it at the six hour ceiling. Seen twice. All tests in the job have passed by the time it stops.

Despite the name, nothing about this is Neo4j. That job runs

```
./gradlew bootJar check --continue --rerun-tasks -PgrailsIndy=... -PonlyNeo4jTests -PskipCodeStyle
```

which is a whole-repository build; `-PonlyNeo4jTests` gates only the functional example applications, so the module tests of the whole repository run in it as well.

## Occurrences

| Date (UTC) | Branch | Job | Outcome |
|---|---|---|---|
| 2026-08-22 18:34 | `fix/embedded-mongo-replica-set` | [attempt 1](https://github.com/apache/grails-core/actions/runs/32591084544/job/97074994493) | ran 18:34:00 → 00:34:16, **exactly 6h00m**, cancelled by the platform |
| 2026-08-22 (retry) | same | attempt 2 | hung again, cancelled manually after 2h18m |
| 2026-08-24 22:49 | `ci/release-drafter-pr-concurrency` (#16213) | [job](https://github.com/apache/grails-core/actions/runs/32786454527/job/97619273551) | hung 4h44m, cancelled manually |

## Where it stops

Identical in both, to the line. The last thing in the log is the AOT example application starting and shutting down, then its `check` task, and then nothing:

```
INFO 50196 --- [main] o.s.boot.tomcat.TomcatWebServer : Tomcat started on port 46097 (http)
INFO 50196 --- [main] aot.Application                 : Started Application in 7.911 seconds
Grails application running at http://localhost:46097 in environment: production
INFO 50196 --- [main] o.s.boot.tomcat.GracefulShutdown: Commencing graceful shutdown
INFO 50196 --- [tomcat-shutdown] o.s.boot.tomcat.GracefulShutdown: Graceful shutdown complete

> Task :grails-test-examples-aot:check
   ← 219 minutes of silence →
##[error]The operation was canceled.
```

In a run that succeeds, `:grails-test-examples-aot:check` is also the last task, and the next lines follow immediately:

```
> Task :grails-test-examples-aot:check
gradle/actions: Writing build results to /home/runner/work/_temp/.gradle-actions/build-results/...json
[Incubating] Problems report is available at: ...
[...]

PR #16220, opened 2026-08-25
ci: make the three untraceable CI flakes say what is happening when they next occur

codeconsole:
Instrumentation for the three CI flakes that are currently untraceable: #16217, #16218 and #16219. No production code, no behaviour change, and each piece removable when its issue closes.

They are together rather than in three pull requests because each run of this branch exercises all three, and these failures are rare enough that coverage per run is what decides whether the next one is diagnosable.

## Why instrumentation rather than a fix

Each of the three has occurrences but no evidence beyond a client-side stack trace, and each has already had a fix proposed on a mechanism that turned out to be wrong. What follows is designed to make the *next* occurrence say which layer is responsible - not to guess at it now.

## #16219 - a request that never answers

`AsyncPromiseSpec` fails with `HttpTimeoutException` after 120 seconds against `/asyncTest/useAsyncService`, an action that waits on a future for five seconds and answers. Four occurrences, three different features, two applications, one of them on `8.0.x`; all say only that the client gave up.

An integration test runs the application in the test's own JVM, so the threads that would serve the request are available to be asked. A request that passes 90 seconds now prints every thread in the JVM while it is *still outstanding*; after the client times out, those threads have moved on. Every request also carries `X-Grails-Test-Correlation-Id`.

That distinguishes: never served, queued and starved, blocked in the controller, or answered but not read.

- `-Dgrails.http.client.diagnostics=false` turns it off; `-Dgrails.http.client.diagnostics.seconds` moves the threshold.
- The timer thread is a daemon deliberately - a build has already hung on a test JVM that could not exit (#16218).

**Checked:** with the threshold at zero, 17 reports, 36 threads each including Tomcat's acceptor and every `http-nio-auto-1-exec-*`. With it at one second, none, because no request was that slow. 800 functional tests pass unchanged either way.

## #16217 - a login form submitted without a session

The page captured at that failure is a fresh login form: empty fields, no error. Three explanations fit it and the page cannot tell them apart - no cookie was issued, one was issued for another host, or one was issued and then lost.

The browser is now reported at three moments - login page loaded, about to submit, and whatever follows - with cookie names, hashes, domain and path, plus the url. Hashed rather than printed: enough to see a session replaced or lost, without putting a session id in a public log. Spring Security logs at debug in this application for the server half.

**Checked**, on a passing run:

```
[16217] login page loaded: url=.../login cookies=[JSESSIONID#522aef1b@host.testcontainers.internal/]
[16217] about to submit:   url=.../login cookies=[JSESSIONID#522aef1b@host.testcontainers.internal/]
[16217] signed in:         url=.../book/index?continue cookies=[JSESSIONID#5289051f@...]
```

The same session through the submit, a new one after it - session fixation protection working. A failure will differ in exactly one of those places.

## #16218 - a build that stops after its last task

Twice this job has stopped producing output after `:grails-test-examples-aot:check` and held a runner to the six hour ceiling, leaving only the process list the runner prints while cleaning up.

A step placed after the build cannot help: while the build is hung, a later step never arrives. The watchdog is therefore armed *before* it, sleeps past the point where the job should have finished, and reports every JVM twice a minute apart - command line, heap, threads with locks - plus the process tree and the tail of the Gradle daemon logs. One dump says what a thread is doing; two say whether it is doing anything at all.

The job is also capped at `timeout-minutes: 120`. It normally finishes in 40 to 65 minutes, so a hang costs a fifth of what it did, and the watchdog has reported long before. Its output is printed by a step marked `always()`, which is reached when the cap cancels the job.

**Checked:** the script runs locally, finds the JVMs through `jps` and dumps them; `./gradlew validateActions` passes and the workflow parses with the steps in the intended order.

## Removing it

Each commit is self-contained and revertible on its own. When an issue closes, its commit goes.


codeconsole:
## The failing run on this branch is #16217

[SiteMesh 2 Functional Tests (Java 21, indy=false)](https://github.com/apache/grails-core/actions/runs/32830492513/job/97747841675) failed on `:grails-test-examples-scaffolding:integrationTest` - `UserControllerSpec > User list`, `WaitTimeoutException` after 30 seconds at `LoginPage.groovy:48`. Three tests in that task, one failed; the build otherwise ran to completion under `--continue`.

That is not a regression from this branch. It is #16217 occurring on the branch that instruments it, and the instrumentation caught it. What follows is the first evidence on this issue beyond a client-side stack trace.

## What the browser held

Three logins ran back to back in the same test JVM against the same application. The two that passed:

```
[16217] login page loaded: JSESSIONID#24f8c0a2@host.testcontainers.internal/
[16217] about to submit:   JSESSIONID#24f8c0a2@host.testcontainers.internal/
[16217] signed in:         JSESSIONID#a5eefb0a@host.testcontainers.internal/   url=.../book/index?continue

[16217] login page loaded: JSESSIONID#3802ec7f@host.testcontainers.internal/
[16217] about to submit:   JSESSIONID#3802ec7f@host.testcontainers.internal/
[16217] signed in:         JSESSIONID#60614cd@host.testcontainers.internal/    url=.../community/user/index?continue
```

The one that failed:

```
[16217] login page loaded:     JSESSIONID#94d58a76@host.testcontainers.internal/
[16217] about to submit:       JSESSIONID#94d58a76@host.testcontainers.internal/
[16217] login page never left: JSESSIONID#94d58a76@host.testcontainers.internal/  url=.../login
```

## What that settles

The description offered three explanations: no cookie was issued, one was issued for another host, or one was issued and then lost. **None of them holds.** A cookie was issued, it carried the right domain and path, and it was still held - unchanged, same hash - for the full thirty seconds after the submit.

What is absent is the change both successful logins show. `#24f8c0a2 -> #a5eefb0a` and `#3802ec7f -> #60614cd` are session fixation protection rotating the session on successful authentication. The failing login's session never rotated, so Spring Security never authenticated anyone.

The url is consistent with that. This application has no `SecurityFilterChain` of its own, so Spring Boot's default form login applies and a rejected credential redirects to `/login?error`. The browser sat on `/login` with no query string at all.

So the fault is upstream of the cookie, and two possibilities remain:

1. the POST never left the browser - `loginButton.click()` returned without navigating;
2. the POST arrived and something answered with the login page again.

The report as it stood cannot separate those two.

## Why it could not go further

Two pieces of evidence existed on the runner and neither reached us.

**The server half never ran.** The `org.springframework.security` DEBUG logger this branch added sits inside `<springProfile name="development">`. Integration tests run under the `test` profile - `GrailsApp` does `environment.addActiveProfile(env.name)`, and the banner in this same log reads `in environment: test` - so the logger was never applied and `<root level="ERROR">` silenced it. Across 49,134 lines of job log there is not one `org.springframework.security` line, against 4,284 DEBUG lines from other applications. The question that would have settled it, *did the POST reach the filter chain*, went unasked.

**The page was captured and then discarded.** `UserControllerSpec` is annotated `@ContainerGebConfiguration(reporting = true)`, and `GebOnFailureReporter` calls `testManager.reportFailure()`, so Geb wrote the page source and a screenshot of the failing browser into `build/geb-reports`. `gradle.yml` uploads that directory on failure; `sitemesh2.yml` had no upload step at all, so on this workflow it dies with the runner.

## Changes

**`logback-spring.xml`** - the security logger moves out of the `development` profile block, so it applies under `test`. `UserControllerSpec` run locally now produces the server half:

```
DEBUG o.s.security.web.FilterChainProxy         : Securing POST /login
DEBUG o.s.s.a.dao.DaoAuthenticationProvider     : Authenticated user
DEBUG .s.ChangeSessionIdAuthenticationStrategy  : Changed session id from 7E63940B...
DEBUG o.s.s.web.DefaultRedirectStrategy         : Redirecting to .../user/index?continue
```

`Securing POST /login` is the line that settles it. Present at the next occurrence, the request reached the filter chain and the answer is on the server; absent, the submit never left the browser.

**`sitemesh2.yml`** - uploads `**/build/geb-reports/` on failure, matching what `gradle.yml` already does for the same suites. The page and screenshot from a failure on this workflow now survive the job.

**`LoginPage`** - each report carries `form=`: `retained` when the username field still holds what was typed, `empty` when the field is there but blank, `gone` when it is absent. On a passing login that reads `empty -> retained -> gone`. A `retained` at the failure means the document was never replaced and the POST did not leave the browser; an `empty` means a new login page was served, so it did. That separates the two remaining possibilities without waiting on the server logs.

One asymmetry worth naming: the client-side report hashes session ids deliberately, and Spring Security at DEBUG prints them in full. These are sessions of a throwaway test application with hardcoded credentials, running in a container for the length of one test, so there is nothing there to protect - but it is not the same standard on both sides, and it is another reason this logger comes out when #16217 closes.

Each change remains revertible on its own.

PR #16222, opened 2026-08-25
GORM Native Datastore Id Support 

codeconsole:
A domain class that declares no `id` has always been given a `Long` one. That is right for Hibernate and wrong for MongoDB, where a `String id` holding a generated `ObjectId` needs no sequence collection and shards cleanly. Declaring `String id` gets that, but ties the source to MongoDB — the same class compiled against Hibernate would need `Long id` instead.

This adds a build setting that lets each domain class take the identity type of the GORM implementation it is actually mapped with:

```groovy
// build.gradle
grails {
    gorm {
        defaultIdType = 'native'
    }
}
```

```groovy
// grails-app/domain/example/Person.groovy — names no store
class Person {
    String name
}
```

Compiled with GORM for MongoDB, `Person` is given a `String id`. Compiled with Hibernate, it keeps `Long`. In an application using both, each domain class gets the right type from the single setting, resolved from its `mapWith` property. A domain class that declares an `id` keeps the type it declares.

The default is `defaultIdType = 'long'`, which is the behaviour of every earlier release.

### How it resolves

`GormEntityTransformation` already worked out which GORM implementation an entity belongs to, at compile time, in `pickGormEntityTrait` — from `mapWith` plus the `GormEntityTraitProvider`s on the compilation classpath. That resolution now also supplies the identity type, through a new default method on the SPI:

```groovy
interface GormEntityTraitProvider {
    Class getEntityTrait()
    boolean isAvailable()
    default Class getDefaultIdentityType() { Long }
}
```

`MongoEntityTraitProvider` returns `String`. Hibernate and Neo4j inherit `Long`. Since the trait and the identity type now come from a single resolution, the two cannot disagree.

The setting reaches the compiler as a system property published by the Gradle plugin, the same way `grails { compileStatic { } }` already does — a `CommandLineArgumentProvider` on `GroovyCompile.groovyOptions.forkOptions.jvmArgumentProviders`, with the effective value exposed as an `@Input` so that changing it invalidates the compile task. A stale class file would otherwise keep the type the previous setting asked for.

`EntityASTTransformation` now runs the discovered domain injectors before `DefaultGrailsDomainClassInjector`. It ran that injector first, and it unconditionally adds a `Long` id, so GORM never got to decide the type on the `@grails.persistence.Entity` path. Each of the default injector's injections is guarded on the property not already being present, so it still fills in everything GORM did not.

### Limitations

- The identity type is compiled into the class. The setting reaches the compiler through the Gradle plugin, so an IDE configured to compile without Gradle produces `Long` ids.
- Turning this on changes the type of a column or field that already holds data. It is a setting to choose when an application is written, not one to switch on over an existing database.
- A domain class compiled into a published plugin jar has its identity type fixed at plugin build time, not at consuming-application build time.
- RX entities are unchanged and keep `Long`.

## Commit 2 — bind a `Serializable` action parameter

An action parameter typed `Serializable` was treated as a command object type. Being an interface it could not be one, so `ControllerActionTransformer` warned *"Interface types and abstract class types are not supported as command objects. This parameter will be ignored"* and bound `null`.

`Serializable` is the type a domain class identifier is declared as when the action does not know the type itself, and it is what `GormEntity.get(Serializable)` accepts. It is now bound the way a `String` parameter is — the raw request value, which is a `String` and so a `Serializable`. GORM converts it to the identity type on the way into `get()`, returning `null` for a value that cannot be converted, so a non-numeric id against a `Long`-id domain still gives a 404 rather than an error.

The dispatch compares the declared type **exactly** rather than testing assignability, so a command object that implements `Serializable` — as many do — is still data bound as a command object. That case is covered by a test.

## Commit 3 — generate scaffolded controllers with a `Serializable` id

The generated controllers declared `show`, `edit`, `update` and `delete` as taking a `Long id`. A MongoDB domain class declaring `String id` — the form the GORM for MongoDB guide recommends — bound `null` instead, and every one of those actions returned `notFound()`. That is a pre-existing bug, independent of the setting above.

They now declare `Serializable id`, which works for `Long`, `String` and `ObjectId` domain classes alike. The generated services already declared `get(Serializable id)` and `delete(Serializable id)`, so this makes the controller agree with the service it calls. Applies to both the `grails-scaffolding` templates and the `rest-api` profile.

### Documentation

- `grails-data-mongodb` — new *Native Identity Types* section under *Identity Generation*
- `grails-doc` — What's New in 8.0


grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
```diff
@@ -0,0 +1,171 @@
...
+
+    private serializableIdController() {
+        gcl.parseClass('''
+            @grails.artefact.Artefact('Controller')
+            class SerializableIdController {
+                def show(Serializable id) {
+                    [value: id]
+                }
+            }
+        ''').newInstance()
```
matrei: `newInstance` is deprecated

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
```diff
@@ -0,0 +1,171 @@
...
+        def cls = gcl.parseClass('''
+            @grails.artefact.Artefact('Controller')
+            @groovy.transform.CompileStatic
+            class StaticIdController {
+                Map show(Serializable id) {
+                    [value: id]
+                }
+            }
+        ''')
+        def controller = cls.newInstance()
```
matrei: `newInstance` is deprecated

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
```diff
@@ -0,0 +1,171 @@
...
+                def $newCommand() {
+                    new WidgetCommand(name: 'Sally')
+                }
+            }
+
+            class WidgetCommand implements Serializable {
+                String name
+            }
+        ''')
+        def controller = cls.newInstance()
```
matrei: `newInstance` is deprecated

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
```diff
@@ -0,0 +1,171 @@
...
+            import grails.web.RequestParameter
+
+            @grails.artefact.Artefact('Controller')
+            class RenamedIdController {
+                def show(@RequestParameter('personId') Serializable id) {
+                    [value: id]
+                }
+            }
+        ''')
+        def controller = cls.newInstance()
```
matrei: `newInstance` is deprecated

borinquenkid (dismissed):
Thanks for this — commits 2 and 3 (binding a `Serializable` action parameter, and generating scaffolded controllers with `Serializable id`) look like solid, independent bug fixes. Both hit any domain class that already follows the current GORM for MongoDB guide (i.e. one that already declares `String id`), so I'd support merging those regardless of what happens below.

My concern is with commit 1 (`defaultIdType`): the type of an id already has a knob, and it's the declaration in the domain class itself. `String id` is one line, per-entity, visible in source, works under IDE compilation (which this system-property mechanism doesn't, per the PR's own limitations), and is what the GORM for MongoDB guide has recommended all along. The implicit-`Long` path this compensates for only works because GORM for MongoDB fakes auto-increment through an internal sequence/counter collection — the same shard-hostile bottleneck this PR's description calls out. As far as I can tell, the only concrete behavior change here is that an undeclared id on a MongoDB-mapped class gets `String` instead of `Long`; Hibernate's and Neo4j's trait providers are untouched.

So the effect of the new SPI method (`getDefaultIdentityType()`), the Gradle plugin class, and the system property threaded through the compiler is to let the build config supply, invisibly and at a distance, a fact that belongs in the source of truth — at the cost that reading the domain class no longer tells you its id type (you need the build config plus the store the class resolves to), plus the IDE/Gradle divergence and the plugin-jar freezing noted in the limitations. Unlike `id generator:`, which is a real degree of freedom the type can't express, the id type has an existing, better home.

I'd rather see a compile-time warning when a MongoDB-mapped domain class has no declared id — nudging the developer to write `String id` (or whatever they intend) explicitly — than machinery that makes the omission work.

Requesting changes on commit 1 pending discussion of the above; happy to be convinced there's a use case that needs automatic inference rather than a warning-and-explicit-declare approach.

grails-core/src/main/groovy/org/grails/compiler/injection/EntityASTTransformation.java
```diff
@@ -94,6 +98,9 @@ public void applyTransformation(SourceUnit sourceUnit, ClassNode classNode) {
             }
         }
 
+        GrailsDomainClassInjector domainInjector = new DefaultGrailsDomainClassInjector();
+        domainInjector.performInjectionOnAnnotatedEntity(classNode);
```
matrei:
```suggestion
        new DefaultGrailsDomainClassInjector().performInjectionOnAnnotatedEntity(classNode);
```

grails-core/src/main/groovy/org/grails/compiler/injection/EntityASTTransformation.java
```diff
@@ -74,13 +74,17 @@ public void applyTransformation(SourceUnit sourceUnit, ClassNode classNode) {
...
 
         final List<ClassInjector> domainInjectors = ArtefactTypeAstTransformation.findInjectors(DomainClassArtefactHandler.TYPE, classInjectors);
 
+        // The discovered injectors run first so that GORM's own transformation, where it is on the
+        // classpath, decides the type of the injected id: it knows which GORM implementation the
+        // entity is mapped with, and this one does not. DefaultGrailsDomainClassInjector then fills in
+        // whatever is still missing - which is everything, for an entity GORM did not process, and
+        // nothing for one it did, since each of its injections is guarded on the property already
+        // being there. The same order holds for entities under grails-app/domain, where
+        // GrailsAwareInjectionOperation sorts GormTransformer ahead of this injector by name.
```
matrei: The correct ordering described here is purely incidental which feels kind of dangerous as something that could blow up in the future.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGormIdTypeProvider.groovy
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * The configured identity type, validated. Stating a value that GORM does not recognise fails the
+     * build here, where the setting is, rather than silently compiling every domain class with a
+     * {@code Long} id.
+     */
+    @Input
+    String getDefaultIdType() {
+        String value = gorm.defaultIdType.getOrElse(BuildSettings.GORM_DEFAULT_ID_TYPE_LONG)
+        if (!(value in [BuildSettings.GORM_DEFAULT_ID_TYPE_LONG, BuildSettings.GORM_DEFAULT_ID_TYPE_NATIVE])) {
```
matrei, line 59: These are compared case-sensitively, whereas GormEntityTransformation (lines 537-544) compares case-insensitively.

Thus `defaultIdType = 'Native'` is accepted by the compiler when invoked directly, but rejected by the Gradle plugin. Either the compiler or provider should normalize/validate consistently. Should we make both paths enforce the same contract?

matrei:
> Thanks for this — commits 2 and 3 (binding a `Serializable` action parameter, and generating scaffolded controllers with `Serializable id`) look like solid, independent bug fixes. Both hit any domain class that already follows the current GORM for MongoDB guide (i.e. one that already declares `String id`), so I'd support merging those regardless of what happens below.
> 
> My concern is with commit 1 (`defaultIdType`): the type of an id already has a knob, and it's the declaration in the domain class itself. `String id` is one line, per-entity, visible in source, works under IDE compilation (which this system-property mechanism doesn't, per the PR's own limitations), and is what the GORM for MongoDB guide has recommended all along. The implicit-`Long` path this compensates for only works because GORM for MongoDB fakes auto-increment through an internal sequence/counter collection — the same shard-hostile bottleneck this PR's description calls out. As far as I can tell, the only concrete behavior change here is that an undeclared id on a MongoDB-mapped class gets `String` instead of `Long`; Hibernate's and Neo4j's trait providers are untouched.
> 
> So the effect of the new SPI method (`getDefaultIdentityType()`), the Gradle plugin class, and the system property threaded through the compiler is to let the build config supply, invisibly and at a distance, a fact that belongs in the source of truth — at the cost that reading the domain class no longer tells you its id type (you need the build config plus the store the class resolves to), plus the IDE/Gradle divergence and the plugin-jar freezing noted in the limitations. Unlike `id generator:`, which is a real degree of freedom the type can't express, the id type has an existing, better home.
> 
> I'd rather see a compile-time warning when a MongoDB-mapped domain class has no declared id — nudging the developer to write `String id` (or whatever they intend) explicitly — than machinery that makes the omission work.
> 
> Requesting changes on commit 1 pending discussion of the above; happy to be convinced there's a use case that needs automatic inference rather than a warning-and-explicit-declare approach.

@borinquenkid I think the use case is that a domain class in a plugin can be used by both hibernate and mongodb.

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
codeconsole: Fixed in `129c92fdb4`: replaced this deprecated call with `getDeclaredConstructor().newInstance()`.

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
codeconsole: Fixed in `129c92fdb4`: replaced this deprecated call with `getDeclaredConstructor().newInstance()`.

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
codeconsole: Fixed in `129c92fdb4`: replaced this deprecated call with `getDeclaredConstructor().newInstance()`.

grails-controllers/src/test/groovy/org/grails/compiler/web/ControllerActionTransformerSerializableParameterSpec.groovy
codeconsole: Fixed in `129c92fdb4`: replaced this deprecated call with `getDeclaredConstructor().newInstance()`.

grails-core/src/main/groovy/org/grails/compiler/injection/EntityASTTransformation.java
codeconsole: Fixed in `129c92fdb4`: `GormTransformer` now declares `HIGHEST_PRECEDENCE`, and injector discovery applies Spring’s `OrderComparator`, making the ordering explicit.

grails-gradle/plugins/src/main/groovy/org/grails/gradle/plugin/core/GrailsGormIdTypeProvider.groovy
codeconsole: Fixed in `129c92fdb4`: the provider normalizes with `toLowerCase(Locale.ROOT)`, matching the compiler. Tests cover mixed-case `Long` and `Native`.

codeconsole:
@borinquenkid The key use case is reusable domain classes supplied by a plugin. Such a class cannot declare `String id` without coupling the plugin to MongoDB, or `Long id` without coupling it to Hibernate. With `defaultIdType = "native"`, the same undeclared-id domain model can adopt the datastore selected by the consuming application: `String` for MongoDB and `Long` for Hibernate. An explicitly declared `id` still takes precedence.

codeconsole:
@matrei @borinquenkid I guess the only weakness in this design is that it does not work for binary plugins, but this wasn't my intention.  The plugins I use this for are built with the application.    I am going to explore the options for binary plugins.

I am looking at... bundling source, building variants, serializable id support, app side ast transformations, etc.

borinquenkid (approved):
One naming nit before merge: the new runtime property grails.gorm.default.idType sits a little too close to the existing compile-time DSL grails { gorm { defaultIdType } } (which is itself published to the compiler as grails.compile.gorm.default.id.type) — three similarly-shaped names for two different mechanisms invites someone transposing them from memory, and both default to 'long' so the mistake would fail silently. Since the code already has a name for this concept internally (portableIdentityType, resolvePortableIdentityType, "portable identity" in the docs), I'd rename the property to grails.gorm.default.portableIdType and let that vocabulary do the disambiguating work instead of "default id type" meaning two different things.

Approving with that suggestion — not blocking.

codeconsole:
@borinquenkid I agree on the confusion.  I just collapsed all 3 into the same settings and they are applied in the context in which they are used.     gradle: compile,  runtime: runtime.

matrei:
>  I guess the only weakness in this design is that it does not work for binary plugins

@codeconsole If this does not work with binary plugins, I think the feature becomes of limited use. It would be great if your investigation uncovers a solution for that.

Otherwise, I'm hesitant to add the feature, since I don't think we should introduce different behavior depending on how a plugin is distributed.

codeconsole:
> > I guess the only weakness in this design is that it does not work for binary plugins
> 
> @codeconsole If this does not work with binary plugins, I think the feature becomes of limited use. It would be great if your investigation uncovers a solution for that.
> 
> Otherwise, I'm hesitant to add the feature, since I don't think we should introduce different behavior depending on how a plugin is distributed.

@matrei you are reading an old message without seeing the updates after.  It works in binary plugins now with the commit "Support native IDs for compiled plugin domains"

PR #16223, opened 2026-08-25
Revert 16192 async changes, fix mongo cache, and fix tests that weren't running

jdaugherty:
Per https://lists.apache.org/thread/y8f8oc3nvdjs2d67ytbs4717l4rk5lmw I'm reverting the async changes.  While doing so, I also discovered that tests weren't running in the uber suite (and were broken due to the nullable default changes).  Finally, I've fixed the mongo cache to be separate / based on the mongo version.


jdaugherty:
Here are my notes on this change: 

### Why the async changes are a -1

`AsyncWebRequestPromiseDecoratorLookupStrategy` caught the `IllegalStateException` the container
throws when asked to start a second async cycle, returned no decorators, and logged at debug.
`WebUtils.clearGrailsWebRequest` swallowed the same exception on the teardown side. Both are the
same symptom: an async task operating on a request the container has already recycled — i.e. data
that was not cleaned up correctly.

The result is a changed contract. A promise may now run undecorated with nothing in the response and
nothing at any visible log level to say so. The tests pass because the failure is quiet, not because
the cause is gone. Reverted: the two classes and the spec added with them, restored byte-for-byte to
their pre-merge state.

### Two CI problems in the same PR

**The `~/.embedmongo` cache never worked.** It was added to all 17 dependency-jar cache steps across
both workflows, but nothing that can *write* that cache ever downloads a mongod. `gradle.yml`'s
`build` job is the only `actions/cache/save` writer, and it runs `-PonlyCoreTests`, which is in the
`onlyIf` of `mongodb-test-config.gradle` — no MongoDB task runs there. `mongodbFunctional` and the
joint workflow do download a mongod, but they use plain `actions/cache@v4`, which only saves on a
primary-key miss; the key already exists by then, so the binary was discarded every run. Meanwhile
13 unrelated jobs restored up to 250MB they have no use for, and the `hashFiles(dependencies.gradle)`
key threw the download away on every dependency bump.

Now it has its own step in only the two jobs that run a server, keyed `embedmongo-<os>-<version>`,
no branch and no hash. The version in the key also settles the matrix — under one key the entry that
missed first won and the other restored a binary it couldn't use.

**The joint workflow was testing three MongoDB releases at once.** It served the example apps from a
`mongo:8` service container but named no version to the build, so testcontainers used the `7.0.19`
default and the embedded server used `V7_0`. The service container is now gone rather than aligned:
`StartMongoGrailsIntegrationExtension` probes `localhost:27017` and starts its own container when
nothing answers, so it was only ever an optimisation — and `mongodbFunctional` already runs the whole
MongoDB suite with no service container at all, as a required upstream of `publish`.
`-PmongodbContainerVersion=8.0` now pins testcontainers and the embedded server to one release.

### 90 tests were running in no task at all

Reviving this turned up something bigger. Classes that pollute static state get a `Test` task of
their own and are excluded from `test`. A `Test` task that is *registered* has neither
`testClassesDirs` nor `classpath` — the `java` plugin wires those onto its own `test` task only — so
it is `NO-SOURCE`: **it reports success having run nothing.** Combined with the `test` exclusion, the
classes ran nowhere.

Affected: the five `isolatedTestPatterns` tasks in `grails-test-suite-uber` (40 tests) and
`execIsolatedTests` in `grails-test-suite-web` (50 tests).

This is older than it looks. `b156214b99` (Nov 2023) rewrote `task x(type: Test)` as
`tasks.register(x, Test)`, but **both forms are `NO-SOURCE`** — I verified them side by side in a
throwaway project — and Gradle was 7.6.3 on both sides of that commit. It inherited the problem
rather than causing it. `TestTaskShardingPlugin` collects these through `tasks.withType(Test)`, so
CI has been distributing empty tasks across shards.

Three things kept it quiet, all fixed:

1. the tasks ran nothing;
2. `grails-test-report` matches phases by exact task name, so results were invisible to the aggregate
   reports even when they did run;
3. in the web suite, nothing declared `execIsolatedTests` as work — `build` → `check` → `test` only,
   so it was reachable just by naming it. `check` now depends on it.

The fix is additive in both modules: two properties per task, one `check` dependency, and the task
names registered in the report. No task is removed.

### What the revived tests found

34 of 40 in the uber suite passed. Four failed, in `RestfulControllerSpec` and
`ResourceAnnotationRestfulControllerSpec`: both assert that `save`/`update`/`patch` re-render the
`create` or `edit` view for an instance with errors, and both build that instance with an empty
title.

This is fallout from our intentional Grails 8 change making persistent properties nullable by
default (`DefaultConstraintEvaluator.applyDefaultNullableConstraint`). Binding an empty string
stores `null`, `blank` never fires on a null, so the instance was valid and the controller
redirected. The domains said `title blank: false`, which meant required under the old default; they
now say `nullable: false` as well, which is what they always meant. **These tests were never updated
when we changed the default because they were dark — nothing flagged them.**

### Dead entries: only the ones that were deleted, not the ones that moved

Several isolated-test patterns name classes this repo no longer has. They fall into two groups, and
only the first is touched.

**Deleted outright.** `JSONBindingTests` and `AutoParams*MarshallingTests` are dropped from the web
suite's list. `1e38f1b5ea` (June 2013) removed all three of those classes — 335 deletions, zero
additions — because the params auto-marshalling feature went away. There is no successor, so naming
them selects nothing and excludes nothing.

**Moved, and rewritten under new names.** Everything else stays exactly as it is, patterns and tasks
alike:

- `GSPResponseWriterSpec` and `**/pages/ext/jsp/*` moved out in `6b0cf3817a` (Apr 2017). The JSP
  tests are back in the monorepo as `org.grails.gsp.jsp.*` (8 classes) in `grails-gsp/plugin`.
- `GroovyPageAttributesTests` now lives in `grails-gsp/grails-taglib`.
- `GrailsDomainBinderTests`, `ComponentValidationTests` and `HibernateMappingUniqueConstraintTests`
  moved out in `ac96f5d5a0` (Mar 2013) and were rewritten as the
  `org.grails.orm.hibernate.cfg.domainbinding.**` `*BinderSpec` suite plus the `UniqueConstraint*` /
  `EmbeddedWithValidationException` specs in `grails-data-hibernate5/7`.
- `DefaultGrailsControllerClassSpec` lives in `grails-core`, where it runs and passes today.

`grails-test-suite-persistence` is therefore untouched, including `testGrailsDomainBinder` and
`testIsolatedPersistentOne`.

### Deliberately left alone: GSP and Hibernate

Those moved tests came back into the monorepo when their repos were merged in, but **without the
isolation they had here** — `grails-gsp/plugin` and `grails-data-hibernate5/7` have no isolated test
tasks at all.

**I have not touched either.** Whether the original pollution concern still applies is a separate
question — the Hibernate binder has been decomposed since, and those specs pass today in a shared
fork — and it shouldn't ride along on a revert. Raising it here so it isn't lost.

jdaugherty:
Note: there are ~90 tests that haven't run since 2023 now as a result of these changes.  Also, one o fthe reasons the groovy jointworkflow was flaky is because it was running tests that no other project was.  @jamesfredley @matrei  if we can get this merged, we can restage the M6

sbglasius (approved):

codeconsole (changes requested):
### WebUtils.clearGrailsWebRequest: fundamentally correct

  Before #16192, this method did:

  1. Remove an attribute from the servlet request.
  2. Reset the thread-local RequestContextHolder.

  If step 1 threw because the container had recycled the request, step 2 never ran. That left a completed request attached to a pooled thread.

  Putting `RequestContextHolder.resetRequestAttributes()` in a finally block is correct defensive cleanup. Even if an earlier lifecycle bug caused the timing, cleanup of a thread-local must not depend on successful access to a potentially expired servlet request.

  Catching the failure from removeAttribute is also reasonable: once the request is recycled, there is no meaningful attribute left for Grails to remove. This change prevents real thread-local contamination rather than merely hiding an error.

jdaugherty:
> Putting `RequestContextHolder.resetRequestAttributes()` in a finally block is correct defensive cleanup. Even if an earlier lifecycle bug caused the timing, cleanup of a thread-local must not depend on successful access to a potentially expired servlet request.

Agreed, and you've convinced me — that half is now kept. 49b0bc97a8 puts the reset back in a `finally`. You're right that it's an unconditional postcondition and must not depend on the servlet request still being reachable.

> Catching the failure from removeAttribute is also reasonable: once the request is recycled, there is no meaningful attribute left for Grails to remove.

This is the half I still disagree with, for two reasons.

**It generalises from one call site.** "Once the request is recycled" describes exactly one of this method's four callers:

| Caller | State of the request when it is called |
|---|---|
| `GrailsAsyncContext:71` | completed — the case the change was written for |
| `UrlMappingUtils:398` | **alive** — clears, then immediately `dispatcher.include(request, …)` |
| `GrailsWebRequestFilter:92` | alive — inside its own `finally`, before the container recycles |
| spring-security `AnnotationFilterInvocationDefinition:141` | alive — mid filter chain |

`UrlMappingUtils` is the clearest: it clears the web request and hands *that same request* to `dispatcher.include` on the very next line. An `IllegalStateException` there isn't a recycled request, it's a bug — and the catch swallows it immediately before the request gets used again. `clearGrailsWebRequest` is public API, so this extends to callers outside the repo as well.

**It puts state inference in a method with no authority over that state.** `clearGrailsWebRequest` is a static utility told to clear two locations. With the catch it reads one exception type as proof of a particular container lifecycle state, then decides that state is unremarkable. `IllegalStateException` out of `getRequest()`/`removeAttribute` isn't a unique signal for "recycled" — anything else throwing it in that path gets reclassified as expected. The `finally` makes no inference at all, which is exactly why it's the right half and the catch isn't.

So the method is now `try`/`finally` with no `catch`: the thread is unbound either way, and a request that can't be reached is reported to the caller that owns the lifecycle instead of being absorbed here. Two tests in `WebUtilsTests` cover both halves — the thread is unbound when the request can't be reached, and the failure still propagates. Removing the `finally` fails the first.

On the promise decorator, I'd still like it reverted. Worth noting your own analysis in #16219 rules it out as the cause of the `AsyncPromiseSpec` timeouts — two of the four occurrences predate it by three and a half weeks, one of them on `8.0.x`. So the catch there isn't fixing that flake; it's making a promise run undecorated with nothing at any visible log level to say so. If there's a case it does fix, I'd rather see it fixed at the lifecycle that leaves a request in that state.

jdaugherty:
Update: I've taken the decorator fix from #16226 into this PR rather than leaving the plain revert — it addresses the cause (a callback attached during dispatch reuses the active `AsyncGrailsWebRequest` instead of asking for a second cycle), which is what the revert alone couldn't do. 2be110a4ed, credited to @codeconsole, including both of the new `WebPromisesSpec` features and the revised lookup-strategy spec asserting the refusal is reported rather than swallowed.

One adjustment on top, in bc41fa9274. The fix changed the guard on running a decorated task from `request.isAsyncStarted()` to `!asyncRequest.isAsyncComplete()` — for every task. Those two differ in a real window: `isAsyncStarted()` goes false the moment the cycle is dispatched or completed, while `isAsyncComplete()` only flips when the container's `onComplete` listener fires. The loosening is exactly what the reuse path needs — during delivery, `isAsyncStarted()` is already false although the request is still live — but applied to the path that starts its own cycle, it widened when tasks run: between `dispatch()`/`complete()` and `onComplete`, a task that would previously have refused now bound a request the container may be recycling.

So the guard is now split by how the decorator got its cycle, recorded at construction:

- **started here** → `request.isAsyncStarted()` required — byte-for-byte the pre-#16192 semantics for the only path that existed then;
- **joined in flight** → `!isAsyncComplete()` only — the exemption the delivery window exists for.

The completion check also moved ahead of `asyncContext.request`: a completed `AsyncContext` throws its own message from `getRequest()`, which would have preempted the intended one.

`AsyncWebRequestPromiseDecoratorSpec` pins all four corners — started-and-running runs, started-and-dispatched refuses, joined-in-delivery runs, joined-then-completed refuses. The started-and-dispatched feature fails against the single loosened guard (verified by removing the split and re-running), so the restored semantics are load-bearing, not decorative. Both #16226 `WebPromisesSpec` features still pass, so the reuse behaviour is intact.

One honest caveat: a joined cycle can still complete while the task body is executing. That race predates #16192 — the original code had the same check-then-run shape — and can't be closed from the decorator; the guard's job here is only not to widen it.

With this, the only remaining difference from #16226 is `WebUtils.clearGrailsWebRequest` — `try`/`finally` with the failure propagating, versus WARN-and-continue — per the discussion above.

matrei (approved):
Did some optional test cleanup in #16228

codeconsole (approved):

PR #16224, opened 2026-08-25
Register a generated auto-configuration where it is generated

codeconsole:
The class a `beans` closure compiles to is created during compilation and is not a source file anyone can open. Registering it in `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` therefore required knowing both that it exists and what the compiler decided to call it — the name follows from the descriptor's own name and package, so it is not something you can read off anything. Not knowing either produced a plugin whose beans were silently never registered, which is a poor thing to ask an author to notice.

The name is settled in `createAutoConfigurationSibling` and nowhere else, so that is where the entry is now written: into the compilation's target directory, beside the `grails-plugin.xml` and `grails.factories` already generated there.

## A module that keeps the file by hand keeps it

Nothing is generated beside it. Two copies of one resource cannot both go into the jar — Gradle fails the `jar` task outright — and folding the hand-authored entries into a copy under the build directory would lose them the moment anyone deleted the file they were, until then, written in.

Such a module is instead **warned at compile time** when the generated class is missing from its file, so the silent case is gone either way:

```
com.example.GreetingAutoConfiguration is generated from a beans closure but is not listed in
src/main/resources/META-INF/spring/...AutoConfiguration.imports, so Spring Boot will not read it.
Add it there, or delete that file once it holds nothing that is not generated and it will be
written for you.
```

Deleting the file is what opts in. Hand-authored entries have to stay possible, which is why they are never assumed to be redundant: a module may register a class from another jar, one annotated with a composed annotation, or one carrying no annotation at all — the imports file *is* the registration, and `@AutoConfiguration` only supplies ordering.

**Nothing that builds today builds differently.** All 21 modules in this repo that keep the file by hand are untouched, `grails-core` among them.

## Scope

Only the sibling generated for a plugin descriptor is registered. A standalone `@GrailsBeans` class is a source file its author can see, so it stays registered by hand like any other auto-configuration — the reason for generating the entry does not apply to it.

## Two modules stop generating it twice

`grails-databinding` and the `beans-dsl-plugin` example applied the `autoconfiguration-imports` convention plugin, which produced the same entries a second time. Verified the jars are unchanged from before:

| module | jar entry |
|---|---|
| `grails-databinding` | `org.grails.plugins.databinding.DataBindingAutoConfiguration` |
| `beans-dsl-plugin` | `beandsl.example.plugin.{Farewell,Greeting}AutoConfiguration` |
| `grails-core` (hand-authored, untouched) | `org.grails.plugins.CoreAutoConfiguration` |

The `beans-dsl` example keeps the convention plugin: what it registers is a standalone class, which is out of scope above.

## Tests

Five cases in a new `AutoConfigurationImportsWriterSpec`, each driving a real compilation with a target directory, since that is the only thing that makes the generated file observable: the sibling registers itself; `autoConfigurationName` registers under the name actually generated; siblings from separate source units accumulate rather than replacing one another; a hand-authored file is left alone and its unique entries untouched; a hand-authored file missing the generated class produces the warning.

`grails-beans-dsl` is green at 170 specs, as are `grails-core`, `grails-databinding`, `grails-i18n`, `grails-domain-class`, `grails-sitemesh3`, `grails-controllers`, `grails-url-mappings` and both `beans-dsl` examples. `./gradlew codeStyle` is clean.

Independent of #16221 — either can merge first.


codeconsole:
Two review findings addressed in c0fba04.

**Stale entries** — real, and confirmed before fixing: compiling `GreetingGrailsPlugin`, deleting its class file as an incremental build would, then compiling a renamed `FarewellGrailsPlugin` left `[com.example.FarewellAutoConfiguration, com.example.GreetingAutoConfiguration]`. Spring Boot does not start on an auto-configuration it cannot load.

An entry is now kept only while the class it names is still generated into the module's own output — which is what the scanning task this replaces recalculated every time. Anything the current compilation registered is kept regardless, because class generation runs in a later phase than this does; without that, two descriptors rebuilt together would each prune the other, neither class file having been written yet. The set is weakly keyed on the compilation, so it is per-build rather than per-JVM in a reused daemon.

Three cases added: a renamed descriptor drops its old entry; a descriptor left untouched by an incremental build keeps its entry; two descriptors recompiling together do not prune one another.

Fixing it also exposed that the specification was compiling only to `CLASS_GENERATION`, which writes no class files — so the output directory it asserted against was empty and could not have shown any of this. It now runs through to `OUTPUT`.

**Hand-authored file found only at the conventional path** — accurate, and now stated rather than glossed. A source set's resource directories are a build-tool notion and are not among the things a compiler is told, so one fixed location is all this can look in; `FactoriesFileWriter` reads `META-INF/grails.factories` from the same fixed path for the same reason. The javadoc and the guide now say so, and say that handling a relocated one needs the build to supply the location. I have not widened the behaviour, since doing it properly means passing resource directories in from the Gradle plugin — happy to follow up with that if you would like it in scope.

codeconsole:
Valid, and fixed in f197fa4. Reproduced first: compile `GreetingGrailsPlugin`, delete its class file as an incremental build does, then recompile a module with nothing that generates a sibling — `[com.example.GreetingAutoConfiguration]` survived. You were right that the rename case only passed because the class replacing it did the writing.

Reconciling is now something the compilation does rather than something generating a class does. The global Grails transform runs for every source unit of a project and already writes generated metadata to the same directory, so it drops entries whose classes are no longer generated there — which happens whether or not anything was generated. Names registered by the compilation in hand are kept regardless, their class files being written in a later phase than any of this runs in.

The file is now deleted rather than left empty when the last entry goes; an imports file with nothing in it is a resource that says nothing.

Verified end to end on a real module rather than only in the specification: deleting `FarewellGrailsPlugin.groovy` from the `beans-dsl-plugin` example and rebuilding takes `beandsl.example.plugin.FarewellAutoConfiguration` out of the file and leaves `GreetingAutoConfiguration` in it.

Four cases added: a deleted descriptor with nothing to replace it takes its entry with it; reconciling leaves an entry whose class is still generated; reconciling creates no file for a module that generates nothing; reconciling leaves a hand-authored file alone.

One case this still does not reach, which seems right to leave: a module whose only source was the descriptor now has no sources at all, so nothing compiles and nothing runs. There is no plugin left at that point either.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -194,6 +195,12 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
             // create or update grails-plugin.xml
             generatePluginXml(pluginClassNode, pluginVersion, transformedClassNames, pluginXmlFile)
         }
+
+        // The generated auto-configurations register themselves as they are created, but a descriptor
+        // that was deleted, or that no longer has a beans closure, creates nothing and so says nothing
+        // about the entry it used to leave behind. This runs for every source unit of a Grails
+        // project, which is what makes the entry go when the class it names does.
+        AutoConfigurationImportsWriter.reconcile(compilationTargetDirectory, compilationUnit)
```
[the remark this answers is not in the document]
codeconsole: Both parts are addressed in 8b6c82b.

The hand-authored resource is now an input to compileGroovy. `AutoConfigurationImportsCompileInput` registers `src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` on every `GroovyCompile` task, so adding or deleting it reruns the transform rather than leaving the task up to date.

The deletion-only case is not reproducible here. Deleting `FarewellGrailsPlugin.groovy` from the `beans-dsl-plugin` example and rebuilding removes `beandsl.example.plugin.FarewellAutoConfiguration` from the generated file and leaves `GreetingAutoConfiguration`: Gradle recompiles on source deletion, so the reconciliation runs. A build presenting zero source units would still not reconcile, and I have not been able to construct one.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
```diff
@@ -324,9 +326,18 @@ private ClassNode createAutoConfigurationSibling(ClassNode pluginClass, Annotati
...
+        // The name is settled here and nowhere else, so this is where it can be registered.
+        AutoConfigurationImportsWriter.register(siblingName, targetDirectory(source), source, compilationUnit);
+
         return sibling;
     }
 
+    /** The compiler's output directory, which is where generated metadata belongs. */
+    private static File targetDirectory(SourceUnit source) {
+        CompilerConfiguration configuration = source == null ? null : source.getConfiguration();
+        return configuration == null ? null : configuration.getTargetDirectory();
```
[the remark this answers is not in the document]
codeconsole: Correct. `resolveCompilationTargetDirectory` branches on `EclipseSourceUnit` at GlobalGrailsClassInjectorTransformation.groovy:469, and this read `CompilerConfiguration.targetDirectory` directly.

Addressed in 8b6c82b, by passing the resolved target in as suggested rather than duplicating the resolution. `createAutoConfigurationSibling` stores the generated name as node metadata (`GENERATED_AUTO_CONFIGURATION_NAME_METADATA`), the global transform reads it and registers using its own Eclipse-aware `compilationTargetDirectory`, and the write here is skipped for an `EclipseSourceUnit`. No dependency from grails-beans-dsl on grails-core.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/AutoConfigurationImportsWriter.java
```diff
@@ -0,0 +1,263 @@
...
+            Files.createDirectories(importsFile.toPath().getParent());
+            // Sorted and newline-terminated, so recompiling the same sources rewrites the same bytes.
+            Files.write(importsFile.toPath(), (String.join("\n", entries) + "\n")
+                    .getBytes(StandardCharsets.UTF_8));
+            return true;
+        }
+        catch (IOException notWritable) {
+            // The class is still generated and still registerable by hand, so failing compilation
+            // over the convenience of not having to would be the worse trade.
+            return false;
```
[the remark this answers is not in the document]
codeconsole: Addressed in 8b6c82b. The `IOException` is reported rather than discarded: `source.addErrorAndContinue(new SyntaxException(message, 1, 1))` where a `SourceUnit` is available, and `IllegalStateException` where one is not.

grails-beans-dsl/src/test/groovy/org/grails/compiler/beans/AutoConfigurationImportsWriterSpec.groovy
```diff
@@ -0,0 +1,268 @@
...
+     * would leave the output directory empty, and what is still generated there is exactly what
+     * decides whether an entry is kept.
+     */
+    private CompilerConfiguration run(CompilationUnit unit) {
+        try {
+            unit.compile(Phases.OUTPUT)
+        }
+        catch (Exception ignored) {
+            // the generated file is what is under test, not the class
+        }
```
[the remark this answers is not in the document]
codeconsole: Addressed in 8b6c82b. `run(unit)` now calls `unit.compile(Phases.OUTPUT)` unguarded, so a sibling that fails to compile or to be emitted fails the specification. No test in the file expects compilation to fail.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
```diff
@@ -324,9 +330,26 @@ private ClassNode createAutoConfigurationSibling(ClassNode pluginClass, Annotati
         sibling.addAnnotations(siblingAnnotations);
         pluginClass.getAnnotations().removeAll(siblingAnnotations);
 
+        // The name is settled here and nowhere else. The global transform consumes this metadata
+        // and registers it using its Eclipse-aware compilation target resolution.
+        pluginClass.putNodeMetaData(GENERATED_AUTO_CONFIGURATION_NAME_METADATA, siblingName);
+        if (!isEclipseSourceUnit(source)) {
+            AutoConfigurationImportsWriter.register(siblingName, targetDirectory(source), source, compilationUnit);
```
jdaugherty: In the standard-compiler path this call and the global transform's metadata-driven `register` both run for the same descriptor in the same compilation: the implicit path reaches here through `compileBeansDsl`, and the global transform then registers again from the node metadata. The double write is absorbed by the short-circuit in `register`, but the hand-authored branch has no such guard, so a module whose hand-authored file is missing the generated class gets the warning emitted twice per compile. Reproduced by compiling a descriptor with an implicit DSL `beans` closure through a `CompilationUnit` with `targetDirectory` set and a hand-authored file listing only another class: `errorCollector.warnings` holds the identical message twice.

Consider deduplicating per compilation, e.g. in the hand-authored branch of `register`, gate the warning on `registeredBy(compilation).add(className)` so the second call stays silent.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/AutoConfigurationImportsWriter.java
```diff
@@ -0,0 +1,278 @@
...
+
+        // A descriptor that was renamed, deleted, or given a different autoConfigurationName leaves
+        // an entry naming a class that is no longer generated, and Spring Boot fails to start on an
+        // auto-configuration it cannot load. Anything this compilation registered is kept regardless:
+        // its class file is written in a later phase than this one runs in.
+        entries.removeIf(entry -> !registeredHere.contains(entry) && !isGeneratedHere(targetDirectory, entry));
+
+        Set<String> written = new TreeSet<>(entries);
+        written.addAll(registeredHere);
+        if (written.equals(entries) && importsFile.isFile() && entries.contains(className)) {
```
jdaugherty: This compares `written` against the post-prune set, so a prune that adds nothing new is never persisted. With the file holding `com.example.GreetingAutoConfiguration` and `com.example.StaleAutoConfiguration`, and neither class file present, `register('com.example.GreetingAutoConfiguration', ...)` drops the stale entry from `entries`, then finds `written.equals(entries)`, returns `false`, and leaves the stale entry on disk - the startup failure the comment above says this pruning prevents. In a Gradle build `reconcile` repairs it moments later, but callers that only go through the local transform (this class's own spec exercises exactly that path) keep the stale entry.

Snapshot the on-disk contents before the `removeIf` and compare `written` against that instead; the `entries.contains(className)` clause then becomes redundant.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
```diff
@@ -148,6 +150,10 @@ class GlobalGrailsClassInjectorTransformation implements ASTTransformation, Comp
                 pluginVersion = resolvePluginVersion(classNode, projectVersion?.toString())
                 addPluginVersionProperty(classNode, pluginVersion)
                 compileBeansDsl(classNode, source)
+                String generatedAutoConfigurationName = classNode.getNodeMetaData(
+                        GrailsBeansASTTransformation.GENERATED_AUTO_CONFIGURATION_NAME_METADATA)
+                AutoConfigurationImportsWriter.register(
```
jdaugherty: `compileBeansDsl` just above deliberately loads `GrailsBeansASTTransformation` reflectively and documents the `ClassNotFoundException` branch, but this hunk adds direct references - `GENERATED_AUTO_CONFIGURATION_NAME_METADATA` is built via `getName()`, so it is not inlined as a compile-time constant and the field read loads the class, as do the `AutoConfigurationImportsWriter` calls (including `reconcile`, which runs for every source unit). If grails-beans-dsl is ever absent from the compile classpath, every project source now fails with `NoClassDefFoundError`, where previously the beans property was just left alone. Either that situation is impossible, in which case the reflective load and its guard comment in `compileBeansDsl` are dead ceremony that could go, or it is possible, and these calls should take the same reflective route. Worth making the file tell one story either way.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -336,7 +336,7 @@ class GlobalGrailsClassInjectorTransformationSpec extends Specification {
                     sourceFile,
                     '''
                         @org.springframework.boot.autoconfigure.AutoConfiguration
-                        class DslBeansGrailsPlugin {
+                        class DslBeansGrailsPlugin extends grails.plugins.Plugin {
```
jdaugherty, line 339: Adding `extends grails.plugins.Plugin` retargets this test at the sibling path, and the previous fixture's behavior - a descriptor not extending `Plugin` whose DSL-shaped `beans` closure is claimed onto the class itself - loses its only coverage here. Consider keeping the old fixture as its own test alongside this one; the non-Plugin case could also assert that no imports file is generated, which pins the 'only the sibling generated for a plugin descriptor is registered' scope.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/AutoConfigurationImportsWriter.java
codeconsole: Confirmed and fixed in 1bce8fd. With the file holding `com.example.GreetingAutoConfiguration` and `com.example.StaleAutoConfiguration` and neither class present, `register` pruned the stale entry, found the pruned set equal to what it was about to write, returned false, and left it on disk.

`register` now snapshots the file before the `removeIf` and compares against that, and the `entries.contains(className)` clause is gone as redundant. Covered by a case that fails against the previous code: the surviving class registers again, adding nothing already listed, and the stale entry does not survive the write it triggered.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
codeconsole: Confirmed and fixed in 1bce8fd. The warning is gated on `registeredBy(compilation).add(className)`, so the second registration of the same descriptor in one compilation is silent.

The regression lives in `GlobalGrailsClassInjectorTransformationSpec` rather than this class own specification: only one of the two registrations happens in grails-beans-dsl, since the global transform is not on that module test classpath, so the duplicate cannot be reproduced there. Reverting the gate makes it report two.

grails-core/src/main/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformation.groovy
codeconsole: Taken as the first of the two, in 4fdc4b2, kept as its own commit so it can be dropped on its own if you would rather have the second.

grails-core/build.gradle:59 declares `api project(:grails-beans-dsl)`, which is what the guard own comment gave as the reason its branch could never be taken. `compileBeansDsl` now constructs `GrailsBeansASTTransformation` directly, and the `ClassNotFoundException` branch is gone: it caught a condition the next statement would have failed on anyway, once the registering above it referenced the class directly.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
codeconsole: Fixed in 1bce8fd. The previous fixture is back as its own test: a descriptor named `*GrailsPlugin` that does not extend `Plugin`, asserting the closure is still consumed onto the class itself, plus the assertion you suggested that no imports file is generated for it. That second one is what pins the scope to the sibling generated for a plugin descriptor.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
codeconsole: Follow-up on this one. Seeding the resolved directory the other way round - the global transform putting it on the class before anything is transformed, so the local transform can write to it - made the global registration redundant: it reads metadata that either was just written by a call which already registered the same name, or, for a descriptor carrying the annotation, is not set yet because the local transform runs afterwards. Removed in d014214.

Verified rather than reasoned: with it gone, a clean build produces byte-identical imports files for grails-core, grails-databinding and both beans-dsl examples, the deletion case still drops the entry end to end, and the ten modules that exercise this are green.

That removes what the duplicate came from, so the gate you suggested is now defence rather than the thing standing between one missing entry and two reports of it - and the specification covering it no longer fails if the gate is taken out. Happy to drop both if you would rather not keep a guard for something that can no longer happen; I have left them in as the cheaper of the two mistakes.

codeconsole: Correcting the previous reply: I said the gate could stay as defence, and that was the wrong call. Removed in d178826, along with the counting claim in its specification.

Checked rather than assumed this time. There is one call site for `register` now, so the same name cannot reach the hand-authored branch twice in a compilation - the gate reported on nothing. Worse, its specification passed with the gate taken out, so it was a test that could not fail sitting in front of a branch nobody was holding.

The case itself is worth keeping, so it stays with the counting dropped: a hand-authored file missing the generated class is told which name to add, and no second file is generated beside it. That one does fail when the warning is removed - verified by suppressing it.

Net effect for the original point: the duplicate is gone structurally rather than suppressed.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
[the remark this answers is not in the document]
codeconsole: Correcting my earlier reply on this thread, which was wrong in two ways.

It claimed the point was addressed. It was addressed for a descriptor whose `beans` closure is claimed by the convention, and not for one carrying `@GrailsBeans` explicitly: the local transform skipped its write, and the global transform registers from node metadata that is not set yet when it runs, because for an annotated descriptor the local transform runs afterwards. So under Groovy-Eclipse that descriptor was registered by neither path. I found that after replying, not before.

It also described a mechanism that no longer exists. Rather than the name being carried out to a directory the global transform resolves, the directory is now carried in: the global transform seeds what it resolved onto every class of the source unit before anything is transformed, and the local transform prefers it over the compiler configuration. With a directory it can trust the local transform no longer skips, so `isEclipseSourceUnit` and its guard are gone, and the registration that read the metadata went with them - one call site, correct under both compilers. 512cc6f and d014214.

Still not verified against a real Groovy-Eclipse build: the specification proves the seeded directory is preferred over the configured one, which is the mechanism, not the IDE. That helper has no test in this repository either.

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
```diff
@@ -324,9 +339,27 @@ private ClassNode createAutoConfigurationSibling(ClassNode pluginClass, Annotati
         sibling.addAnnotations(siblingAnnotations);
         pluginClass.getAnnotations().removeAll(siblingAnnotations);
 
+        // The name is settled here and nowhere else.
+        pluginClass.putNodeMetaData(GENERATED_AUTO_CONFIGURATION_NAME_METADATA, siblingName);
```
jdaugherty: With the global transform's registration gone, nothing reads `GENERATED_AUTO_CONFIGURATION_NAME_METADATA` any more - this write and the constant are the only references left in the repository, and the constant's javadoc still describes the removed mechanism ("carrying the generated sibling's binary name to the global transform"). Either drop the constant and this write, or keep it as a deliberate hook for tooling and reword the javadoc to say so.

grails-beans-dsl/src/test/groovy/org/grails/compiler/beans/AutoConfigurationImportsWriterSpec.groovy
```diff
@@ -0,0 +1,356 @@
...
+ */
+package org.grails.compiler.beans
+
+import org.codehaus.groovy.ast.ClassNode
+import org.codehaus.groovy.classgen.GeneratorContext
+import org.codehaus.groovy.control.CompilationUnit
+import org.codehaus.groovy.control.CompilerConfiguration
+import org.codehaus.groovy.control.Phases
+import org.codehaus.groovy.control.SourceUnit
+import org.codehaus.groovy.control.SourceUnit
```
jdaugherty: Duplicate import - `org.codehaus.groovy.control.SourceUnit` is already imported on the line above. (`codeStyle` does not catch it, I checked.)

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
```diff
@@ -349,6 +349,71 @@ class GlobalGrailsClassInjectorTransformationSpec extends Specification {
...
         then: "the property is consumed by the transform, unlike the two cases above"
             classNode.getProperty('beans') == null
+
+        and: "the name settled by the local transform is registered at the global transform's target"
+            new File(targetDir,
+                    'META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports').text.trim() ==
+                    'DslBeansAutoConfiguration'
+    }
+
+    void "the implicit beans convention claims the closure of a descriptor that is not a Plugin"() {
```
jdaugherty, line 389: Every end-to-end case here goes through the implicit convention; the explicit `@GrailsBeans` entry path - where the local transform runs after this transform and picks up the seeded directory - has no coverage at this level. A descriptor carrying both `@grails.compiler.beans.GrailsBeans` and `@AutoConfiguration`, compiled through `compileToFile`, ends up with the imports file holding the sibling's name at the target directory (I ran exactly that and it passes), so it is a cheap add that pins the second of the two entry paths through the real pipeline. It cannot discriminate the seeded directory from the compiler-configuration fallback - only Groovy-Eclipse diverges there, as the writer spec's mechanism test already notes - but it would catch the explicit path silently registering nothing, which is the regression d014214's earlier shape actually had.

jdaugherty (approved):

grails-beans-dsl/src/main/java/org/grails/compiler/beans/GrailsBeansASTTransformation.java
codeconsole: Dropped in 4c0bd13, both the write and the constant. There was no tooling hook intended - it existed only to carry the name to the global transform, which stopped reading it.

grails-beans-dsl/src/test/groovy/org/grails/compiler/beans/AutoConfigurationImportsWriterSpec.groovy
codeconsole: Removed in 4c0bd13. Mine, added with the Groovy-Eclipse mechanism test without checking what was already imported.

grails-core/src/test/groovy/org/grails/compiler/injection/GlobalGrailsClassInjectorTransformationSpec.groovy
codeconsole: Added in 4c0bd13, in the shape you describe: a descriptor carrying both annotations through `compileToFile`, asserting the closure was consumed and the imports file at the target directory holds `AnnotatedBeansAutoConfiguration`.

Checked it can fail rather than trusting that it passes: suppressing the registration for that one source makes it fail on both assertions. So it does pin the entry path, and agreed it cannot discriminate the seeded directory from the configuration fallback - only Groovy-Eclipse diverges there, and the mechanism test in the writer specification is what covers that.

PR #16226, opened 2026-08-25
Reuse the active web request during async dispatch

codeconsole:
A callback attached while the container is dispatching an async result must not start a second async cycle. Reuse the active AsyncGrailsWebRequest instead, preserving request decoration and leaving completed or inconsistent requests as visible failures.

Always clear the thread-bound GrailsWebRequest when a container has recycled the servlet request, and log the failed attribute cleanup at warning level.

Cover both paths through WebPromises and WebUtils public APIs.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -44,7 +44,6 @@ jobs:
           path: |
             ~/.gradle/caches/modules-2
             ~/.gradle/wrapper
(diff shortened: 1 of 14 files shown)


jdaugherty:
@codeconsole since the original PR is what is referenced on the mailing list, I'd rather keep it in case people look for it later.  Can we work off the original one?

PR #15960, opened 2026-07-10
Add VirtualThreadPromiseFactory seed

jamesfredley:
### What was found

| Problem | Impact |
| --- | --- |
| Async still centers on older factory options including GPars path | Misses Java 21 virtual threads |
| Codebase 3 called for VirtualThreadPromiseFactory | No implementation yet |
| GPars should be deprecation-track for 9.0 | Needs docs + modern factory first |

### What changed

| Area | Change |
| --- | --- |
| Factory | Add `VirtualThreadPromiseFactory` in grails-async-core |
| Selection | Wire through `PromiseFactoryBuilder` with opt-in config |
| Docs | Async guide notes GPars deprecation direction |
| Tests | Virtual-thread factory coverage |

### Out of scope / follow-up

| Topic | Status |
| --- | --- |
| Removing grails-async-gpars | 9.0 |
| Full ThreadLocal audit | Separate epic (#15828 partial) |
| Default factory flip | Decision needed |

### Related MD topics

| Source | Topic |
| --- | --- |
| Codebase 3 | GPars deprecation / VirtualThreadPromiseFactory |
| Google Doc 1.3 | Virtual-thread readiness |

### Issue and Scope
- [x] Starter modern async factory.
- [x] Opt-in, non-breaking default path preserved.
- [x] Single focused seed.
- [x] Targets `8.0.x`.

### Code Quality
- [x] Tests added.
- [x] Focused async-core tests run.
- [x] No mass reformatting.
- [x] AI starting point labeled.

### Licensing and Attribution
- [x] Apache License 2.0.
- [x] Contributor rights confirmed.
- [x] `ai-generated-starting-point` label applied.

### Documentation
- [x] Async docs updated.
- [x] Scope explained.

Assisted-by: Sisyphus:xai/grok-4.5 [gpt-coding]

PR #16230, opened 2026-08-26
Modernize Grails async integration for Spring Boot 4.1

codeconsole:
## Summary

- Replace the default FutureTask-based promise implementation with a Groovy-friendly `CompletableFuture` implementation while retaining the existing `Promise` and `PromiseFactory` APIs.
- Execute Grails promises on Spring Boot's managed `applicationTaskExecutor`, including Boot lifecycle management, graceful shutdown, pool configuration, virtual-thread support, and composed task decorators. Resolve the executor by name; use a Spring-managed eight-core-thread fallback when Boot does not provide one.
- Preserve executor affinity across derived `CompletionStage` instances, use `CompletableFuture.supplyAsync` for submission, and expose Java 21's `Future` state and immediate-result APIs. Use Groovy 5 lambdas, pattern variables, and unnamed closure parameters where appropriate.
- Make the explicit `virtual-thread` promise factory a thin CompletableFuture-based implementation with its own decorated Java 21 virtual-thread executor, the same promise contracts, asynchronous event delivery, and lifecycle cleanup. This opt-in does not use Boot's executor pool or shutdown configuration.
- Propagate `GrailsWebRequest` through a Spring `TaskDecorator` around the entire executor task, including chained callbacks and standalone `WebPromises`. Retain compatibility decoration for third-party factories.
- Delegate controller async lifecycle, completion, dispatch, timeout, and exception processing to Spring's `StandardServletAsyncWebRequest`, `DeferredResult`, and `WebAsyncManager`.
- Keep the default event bus asynchronous by sharing the promise factory's executor without requiring the factory to implement `ExecutorService`.
- Preserve original checked and unchecked exception causes in tasks, chained and aggregate callbacks, and third-party promise bridges. Keep consistent `ExecutionException` semantics for direct and chained `get()` calls.
- Honor and cache `spring.mvc.async.request-timeout` for eager and returned web promises. Resolve async timeouts as HTTP 503 without an ERROR-level stack trace, support 503 URL mappings, and reject promises returned or submitted after request completion.
- Preserve explicit user-supplied promise decorators and third-party `PromiseFactory` providers discovered through `ServiceLoader`.
- Deprecate the retained FutureTask, raw Servlet async, legacy request decorator, and Spring factory-bean compatibility APIs for future removal. Neither default nor virtual-thread execution depends on the deprecated FutureTask implementation.
- Document configuration, exception contracts, virtual-thread execution, timeout behavior, and migration guidance in the async, events, and upgrade guides.

## LOC

| Area | Added | Removed | Net | Deprecated | Net after deprecated removal |
|---|---:|---:|---:|---:|---:|
| Production | 653 | 195 | +458 | 1,018 | -560 |
| Tests | 1,085 | 49 | +1,036 | 0 | +1,036 |
| Documentation | 66 | 12 | +54 | 0 | +54 |
| **Total** | **1,804** | **256** | **+1,548** | **1,018** | **+530** |

Physical line counts from `git diff --numstat e749cc79a9..f3624cb8c3`, excluding upstream changes merged into the branch. Test counts include functional-test application fixtures. The final column is added minus removed minus deprecated; it projects removal of retained compatibility code, not an additional deletion already made.

The deprecated total includes 863 physical lines in public APIs marked `@Deprecated(forRemoval = true)` and 155 lines in the internal `FutureTaskChildPromise`, required only by the deprecated FutureTask implementation.

## Verification

- Java 21: all module suites in `grails-async-core`, `grails-async`, `grails-web-mvc`, `grails-events-core`, and `grails-web-url-mappings` pass (427 tests discovered, one existing skip).
- Java 25: all suites in the three changed async/MVC modules pass with the build cache disabled (156 tests discovered, one existing skip).
- Java 21: all 18 integration tests in `grails-test-examples-async-events-pubsub-demo` pass, including the new HTTP 503 timeout regression.
- Affected-module code style and dependency-version validation pass; `git diff --check` is clean.
- Verification is scoped to avoid the memory and disk pressure of the repository-wide clean/aggregate sweep: one Gradle worker, one test fork, and a 2 GiB Gradle heap.


codeconsole:
Follow-up on the legacy async-library integrations: the current PR does **not** remove them. The branch still builds, publishes, documents, and discovers these optional PromiseFactory implementations through `ServiceLoader`:

- GPars 1.2.1 (`grails-async-gpars`)
- RxJava 1.3.8 (`grails-async-rxjava`)
- RxJava 2.2.21 (`grails-async-rxjava2`)
- RxJava 3 (`grails-async-rxjava3`)

None of these libraries is required for `grails-async` functionality anymore. The modern baseline can be implemented entirely with JDK `CompletableFuture`, Spring Boot's managed `applicationTaskExecutor`, Spring MVC async processing, and the existing Groovy `Promise` facade.

Recommended disposition:

- Remove RxJava 1 support because it is end-of-life.
- Remove RxJava 2 support because it has been superseded by RxJava 3.
- Remove GPars support from `grails-async`; it provides no required capability over the new JDK/Spring implementation.
- Prefer removing RxJava 3 as well unless Grails intentionally wants to maintain first-party RxJava interoperability.

Removing all four adapters would leave a single supported JDK/Spring implementation while retaining the `PromiseFactory` SPI for independently maintained integrations. It would also eliminate the current behavior where adding a provider library to the classpath can silently replace the global promise engine through `ServiceLoader`.

This is a user-visible breaking change because these modules are published artifacts, so it should be documented in the Grails 8 release and migration notes. It also requires a broader repository review: `grails-events-gpars` remains a separate optional event bus, `grails-cache` exposes GPars as an API dependency, and Neo4j uses GPars in tests.

jdaugherty:
We've discussed the gpars removal in the past.  I agree we should drop the older rxjava, etc versions but gpars I've objected to.  

I think this PR is a major shift, and not necessarily a bad one, but I don't think it should go into 8.0.

codeconsole:
> We've discussed the gpars removal in the past. I agree we should drop the older rxjava, etc versions but gpars I've objected to.
> 
> I think this PR is a major shift, and not necessarily a bad one, but I don't think it should go into 8.0.

@jdaugherty I am fine with retargeting? Which do you prefer 8.1 or 9? I am fine with either

matrei:
## Review

**Head:** `8dd2b0a1dd` (`feature/modernize-grails-async-8.0.x`) · **Base:** `8.0.x` @ `a5758d657f` · merge-base is clean (branch is a fast-forward candidate; `origin/8.0.x` has since moved 91 commits, none touching async or URL mappings).

**What I ran locally**

| Check | Result |
|---|---|
| `:grails-async-core:test` (fresh, `--no-build-cache`) | 54 tests, 0 failures, 1 skipped |
| `:grails-async:test` (fresh) | 18 tests, 0 failures |
| `:grails-web-url-mappings:test` (`--rerun`) | 259 tests, 0 failures |
| `:grails-async-core:codeStyle :grails-async:codeStyle :grails-web-url-mappings:codeStyle` | pass |
| Throwaway probe specs (deleted afterwards) for the findings below | see "Probe" notes per finding |
| Boot 4.1.1 bytecode (`OnBeanCondition`, `TaskExecutorConfigurations`, `WebMvcAutoConfiguration`) | inspected with `javap` |

**Verdict: request changes.** The direction is right and the Spring MVC `DeferredResult` integration is a clear improvement over the hand-rolled `AsyncContext` handling, but the PR currently regresses four things that ship enabled in every generated Grails app: the events bus silently becomes synchronous, exceptions with a cause are misreported, the promise factory bean fails to resolve when a second `AsyncTaskExecutor` exists (e.g. `@EnableScheduling`), and the fallback executor is single-threaded.

---

### High

**H1. Every Grails app's `EventBus` silently becomes synchronous.**
`grails-events/core/src/main/groovy/org/grails/events/bus/spring/EventBusFactoryBean.groovy:73` only reuses the promise factory when it `instanceof ExecutorService`. `CachedThreadPoolPromiseFactory` was one (via `ExecutorPromiseFactory`); `CompletableFuturePromiseFactory` is not. The factory bean therefore falls through to `EventBusBuilder.createDefaultEventBus()`, which logs `No event bus implementations found on classpath, using synchronous implementation.` and builds `ExecutorEventBus(new SyncTaskExecutor())`. `grails-events` is added to every forge-generated app (`grails-forge-core .../feature/grails/GrailsBase.java:102`), so `sendEvent`/`@Subscriber` listeners now run on the caller's thread with a spurious WARN at startup. The updated `grails-doc/src/en/guide/async/events.adoc:32` claims the bus "executes work on Spring Boot's `applicationTaskExecutor`", which is now false.
*Probe:* `EventBusFactoryBean` with a non-`ExecutorService` `grailsPromiseFactory` singleton → `ExecutorEventBus.executor` is `SyncTaskExecutor`.
*Fix:* `ExecutorEventBus` already accepts a plain `Executor`. Have `EventBusFactoryBean` use `CompletableFuturePromiseFactory.executor` (or introduce a small `ExecutorAware`-style interface on the factory) so the bus shares the Boot executor as the docs say. Add a test in `grails-events` for it.

**H2. `unwrap` strips one cause level from *any* exception, so `onError` and the error page see the wrong exception.**
`CompletableFuturePromise.groovy:144` (`failure.cause ?: failure`) is applied in `onError` (line 75) to the raw exception the promise itself failed with, not to a `CompletionException` wrapper. For `createPromise { throw new IllegalStateException('outer', new IOException('inner')) }`, the `onError` callback receives the `IOException`. `AsyncActionResultTransformer.groovy:72` then applies its own `unwrap` (line 79) on top, so the exception handed to `DeferredResult.setErrorResult` and ultimately to `GrailsExceptionResolver` / UrlMappings exception mappings is the cause, not the thrown exception. Any exception mapping keyed on the thrown type (`"500"(controller: 'errors', exception: MyServiceException)`) will stop matching when that exception has a cause.
*Probe:* both `Promise.onError` and `WebAsyncManager.concurrentResult` observed `java.io.IOException: inner`.
*Fix:* only unwrap `CompletionException`/`ExecutionException` (and only when `cause != null`), and drop the second unwrap in the transformer. Add tests with a caused exception to `CompletableFuturePromiseFactorySpec` and `AsyncActionResultTransformerSpec`.

**H3. `context.bean(AsyncTaskExecutor)` fails with `NoUniqueBeanDefinitionException` whenever a second `AsyncTaskExecutor` exists.**
`ControllersAsyncGrailsPlugin.groovy:72` resolves by type. `ThreadPoolTaskScheduler` implements `AsyncTaskExecutor`, so any app with `@EnableScheduling` (Boot auto-configures `taskScheduler`) or with its own second executor bean fails on context startup with `expected single matching bean but found 3: grailsPromiseExecutor,applicationTaskExecutor,taskScheduler`. The fallback flag does not help here because there are two non-fallback candidates. Boot itself avoids this: `WebMvcAutoConfiguration$WebMvcAutoConfigurationAdapter.configureAsyncSupport` looks up `applicationTaskExecutor` **by name** and does nothing otherwise.
*Probe:* registrar + `applicationTaskExecutor` + an initialized `ThreadPoolTaskScheduler` singleton → `BeanCreationException` wrapping the above.
*Fix:* `context.beanFactory.containsBean(TaskExecutionAutoConfiguration.APPLICATION_TASK_EXECUTOR_BEAN_NAME)` → use it, else the `grailsPromiseExecutor` bean by name. Cover it with a test that registers two executors.

**H4. The fallback executor is single-threaded.**
`ControllersAsyncGrailsPlugin.groovy:62` uses a default `ThreadPoolTaskExecutor` (core 1, max `Integer.MAX_VALUE`, queue `Integer.MAX_VALUE`). With an unbounded queue the pool never grows past the core size, so every promise in a non-Boot context (the Micronaut path the PR description mentions, plain Spring contexts, tests) is serialised on one thread, and a promise that waits on another promise deadlocks. The previous default was a cached thread pool.
*Probe:* promise A awaiting a latch that promise B releases → A times out; executor reports `core=1 max=2147483647 queueCapacity=2147483647`.
*Fix:* mirror Boot's defaults (core 8 / unbounded queue), or `SimpleAsyncTaskExecutor` (optionally virtual threads on Java 21), or simply fall back to `new CompletableFuturePromiseFactory()` and wrap its executor with the decorators. Add a two-promise concurrency test.

### Medium

**M1. `WebPromises` no longer propagates the request unless the plugin bean has been instantiated.**
Previously `WebPromises.getPromiseFactory()` / `setPromiseFactory()` always attached `AsyncWebRequestPromiseDecoratorLookupStrategy`, so `task { render ... }` worked in controller unit tests and in any context where `grailsPromiseFactory` had not been created yet. Now propagation lives only on the Spring-registered executor; `WebPromises.groovy:50` builds a bare `CompletableFuturePromiseFactory` with its own cached pool and no `GrailsWebRequestTaskDecorator`. `grails-testing-support-web` does not load `ControllersAsyncGrailsPlugin`, so `ControllerUnitTest` specs of async actions lose `render`/`params`/`request` inside `task {}`. The rewritten `WebPromisesSpec` masks this by wiring the decorator manually in `setup()`.
*Probe:* `WebPromises.promiseFactory = null`, bind a `GrailsWebRequest`, `WebPromises.task { GrailsWebRequest.lookup() }.get()` → `null` (was a `GrailsWebRequest`).
*Fix:* when `WebPromises` builds its default factory, wrap the executor with `GrailsWebRequestTaskDecorator` (or keep a lookup strategy that applies it). Restore a spec that does not pre-wire the decorator.

**M2. `get()` exception type differs between a promise and its `then {}` child.**
`CompletableFuturePromiseFactory.groovy:81` passes `unwrapFailureOnGet = false`, while `fromStage`'s default at `CompletableFuturePromise.groovy:127` is `true`, and `onComplete`/`then` use the default. So `createPromise { throw X }.get()` throws `ExecutionException(X)` (matching `FutureTaskPromiseFactorySpec` and `PromiseListSpec`), but `createPromise { throw X }.then { it }.get()` throws `X` directly. Lines 105/118 also sneaky-throw a checked `Throwable` from a method declared `throws ExecutionException`, which Java callers cannot catch by declaration. Pick one behaviour (the legacy `ExecutionException` one keeps compatibility) and cover it.
*Probe:* `direct=java.util.concurrent.ExecutionException chained=java.lang.IllegalStateException`.

**M3. Docs still present the deprecated Servlet API as current, and the upgrade guide is silent.**
`grails-doc/src/en/guide/async/asyncServletApi.adoc:25` and `traits/traitsprovided.adoc:35` document `AsyncController`/`startAsync()` with no deprecation notice, and `upgrading/upgrading80x.adoc` has nothing on: the executor switch (pool size now governed by `spring.task.execution.*`, 8 core threads instead of unbounded), the deprecations, the `ExecutorPromiseFactory` removal from the default factory, or the events-bus change. Per the repo rule that user-facing changes ship with docs, please add an 8.0 upgrade note and mark the Servlet section deprecated.

### Low

**L1. `onError(List, Closure)` hands the callback a freshly built `ExecutionException(cause)`** (`CompletableFuturePromiseFactory.groovy:128`) while single-promise `onError` hands over the unwrapped cause. Callers of `Promises.onError(list) { e -> }` and `promise.onError { e -> }` get different types for the same failure. Align them.

**L2. Timeout inconsistency between the two entry points.** `WebPromises.prepareAsyncRequest()` (`WebPromises.groovy:192`) sets an infinite timeout, whereas an action returning a `Promises.task {}` / GORM `Domain.async` promise goes through the transformer with the container default (Tomcat: 30 s). This mirrors the old split, so not a regression, but now that the flow is Spring-managed it would be natural to honour `spring.mvc.async.request-timeout` via `DeferredResult(timeout)` in both paths.

**L3. The "task after async completion" guard is gone.** The old decorator threw `Cannot start a task once asynchronous request processing has completed`; now such a task captures and later touches a recycled servlet request. Consider checking `asyncManager.asyncWebRequest?.isAsyncComplete()` in `prepareAsyncRequest()`.

**L4. `grailsWebRequestTaskDecorator` is a global `TaskDecorator` bean**, so Boot also applies it to `@Async` methods and (Boot ≥ 3.5) the task scheduler. Harmless when no request is bound, but it is a behaviour change for `@Async` methods called from controllers (they now see a `GrailsWebRequest` copy). Worth a sentence in the docs.

**L5. Commit history.** `479f5e0e2e Restore URL mapping state on async dispatch` and `c982feb634 Reuse matched URL mappings on async dispatch` cancel each other out (`UrlMappingsHandlerMapping` ends up unchanged). Please squash before merge per the contribution guide.

---

### Verified as correct

- **Fallback bean does not suppress Boot's executor.** `OnBeanCondition` in Boot 4.1.1 filters fallback definitions (`getNonFallbackBeans`/`isNotFallback` → `BeanDefinition.isFallback()`), and `TaskExecutorConfigurations$OnExecutorCondition$ExecutorBeanCondition` is `@ConditionalOnMissingBean(Executor.class)`, so `grailsPromiseExecutor` does not stop `applicationTaskExecutor` from being created.
- **Boot composes multiple `TaskDecorator` beans**: `ThreadPoolTaskExecutorBuilderConfiguration` takes `ObjectProvider<TaskDecorator>`, uses `orderedStream()` and `getTaskDecorator(...)`, so `GrailsWebRequestTaskDecorator` is applied alongside user decorators on `applicationTaskExecutor`.
- **`AsyncActionResultTransformer` + `UrlMappingsInfoHandlerAdapter`** now follow the same protocol as Spring's `RequestMappingHandlerAdapter` (`hasConcurrentResult()` → read → `clearConcurrentResult()` → rethrow `Exception` / return `ModelAndView`), and a `null` concurrent result correctly falls through to "no view" for actions that `render` inside the task (the `AsyncFunctionalSpec` path).
- **Joining an eagerly started async cycle**: `StandardServletAsyncWebRequest.startAsync()` is a no-op when already started, so `WebPromises.task {}` followed by returning the promise works (covered by the new transformer spec).
- **Executor affinity**: `defaultExecutor()` + `newIncompleteFuture()` keep `thenApplyAsync` etc. on the configured executor; the decorator therefore also covers derived stages.
- **Explicit decorators are honoured** by `WebPromises.createPromise(c, decorators)` — the old code silently replaced them with the lookup result, so this is a bug fix.
- **`@PreDestroy close()`** only shuts down an executor the factory owns, so the Boot-managed executor is left to Boot's lifecycle.
- All three affected module suites and their code-style checks pass on a fresh run.

codeconsole:
Thanks for the detailed review and probes, @matrei. Addressed the behavioral and documentation findings in five focused commits:

- H1: the default event bus now uses the promise factory's exposed executor, while retaining support for factories that implement `Executor`. Spring bean introspection keeps events-core independent of the optional async module. A regression test verifies asynchronous delivery on the shared worker.
- H2, M2, L1: only completion/execution wrappers are unwrapped; caused application exceptions retain their identity. Single and aggregate error callbacks agree, direct/chained/timed `get()` calls retain `ExecutionException`, and MVC receives the original failure.
- H3, H4: resolve `applicationTaskExecutor` by name, otherwise use the named Grails fallback with eight core threads. Tests cover additional executors/schedulers and a task waiting on another task.
- M1: default standalone `WebPromises` adapts the same Spring request decorator without requiring plugin initialization. Tests no longer pre-wire propagation and cover request parameters and response writing on a worker.
- L2, L3: both request entry points honor `spring.mvc.async.request-timeout`, falling back to the container default. Completion is also recorded on the request because Spring removes its async manager on completion; subsequent task submission is rejected before executing work.
- M3, L4: added the 8.0 migration guidance, Servlet API/trait deprecation notices, and documentation of the global decorator's effect on other Boot-managed executors.
- L5: left history unsquashed per the contributor's explicit instruction to retain focused commits. The cancelling historical commits remain unchanged.

Validation: Java 21 async-core, async web, events-core, and URL-mapping suites: 358 tests, one existing skip, no failures. Java 25 changed-module suites: 99 tests, one existing skip, no failures. The Java 21 async HTTP error-handling functional test also passes, as do affected-module style and dependency-version checks. No full-repository clean/aggregate sweep was run; verification used bounded workers and test parallelism.

Commits: f269620071, c8c9a0b1b2, b8796c25fd, f7da7dee19, a29efe7561. The PR summary and LOC table have been refreshed.

matrei:
## Review round 2

**Head:** `b65a64ed69` (`feature/modernize-grails-async-8.0.x`, 18 first-parent commits) · **Base:** `8.0.x` @ `a5758d657f`. The head merges upstream `8.0.x` up to the #15960 merge (`VirtualThreadPromiseFactory`); the two `8.0.x` merges after that (#16322, #16184) are not included and `upgrading80x.adoc` conflicts with them (N9).

Six commits since round 1: five fix commits (`f269620071` … `a29efe7561`) reviewed in detail below, plus the upstream merge `b65a64ed69`, covered in the post-merge addendum. The "Verified as correct" list from round 1 still holds.

**What I ran locally** (on the merged head `b65a64ed69`; the N1/N2 probes were run on `a29efe7561`, and none of the files they exercise changed in the merge)

| Check | Result |
|---|---|
| `:grails-async-core:cleanTest test` (`--no-build-cache`) | 69 tests, 0 failures, 1 skipped |
| `:grails-async:cleanTest test` (`--no-build-cache`) | 27 tests, 0 failures |
| `:grails-events-core:cleanTest test` (`--no-build-cache`) | 12 tests, 0 failures |
| `:grails-async-core:codeStyle :grails-async:codeStyle :grails-events-core:codeStyle` | pass |
| `git merge-tree origin/8.0.x pr/16230` | conflict in `upgrading80x.adoc` only |
| Throwaway probe spec (deleted afterwards) for N1 and N2 below | output quoted per finding |

**Verdict: request changes, but close.** All four High findings from round 1 are fixed and covered by tests (see "Round 1 status" at the end). The rework of the exception path exposed one new regression that the new tests miss because they only throw unchecked exceptions, and the standalone `WebPromises` fix restored propagation for the task closure but not for `then {}` callbacks, which the 7.x implementation did propagate.

---

### High

**N1. Checked exceptions thrown from a task reach `get()`/`onError` wrapped in `UndeclaredThrowableException`.**
`CompletableFuturePromiseFactory.groovy:78` submits `{ decorated.call() } as Supplier<T>`. Under `@CompileStatic` the `as` coercion still produces a `java.lang.reflect.Proxy` over `ConvertedClosure`; `Supplier.get()` declares no checked exceptions, so the proxy wraps any checked exception in `UndeclaredThrowableException` before `CompletableFuture` ever sees it. `unwrap()` (`CompletableFuturePromise.groovy:109`) only strips `CompletionException`/`ExecutionException`, so the wrapper survives. The same coercion sits in `asCompletableFuture` (`CompletableFuturePromiseFactory.groovy:144`, where `promise.get()` itself throws checked `ExecutionException`) and in the new standalone web decorator (`WebPromises.groovy:71`, `{ … } as Runnable`).
The old `FutureTask` implementation ran a `Callable`, so `task { throw new IOException() }.get()` used to throw `ExecutionException(IOException)` and `onError` received the `IOException`. Groovy code throws checked exceptions without declaring them all the time, so this is the common case, not an edge case, and it breaks `exception:` URL-mapping matches and `onError` type checks for every `IOException`/`SQLException`/`TimeoutException` coming out of a task. The new `CompletableFuturePromiseFactorySpec` cases all throw `IllegalStateException`, which is why they pass.
*Probe (standalone factory, `Promises.task { throw new IOException('io') }`):*
```
direct get():  ExecutionException -> UndeclaredThrowableException -> IOException
onError sees:  UndeclaredThrowableException (cause IOException)
WebPromises:   UndeclaredThrowableException (cause IOException)
```
*Fix:* never let a closure cross a proxy boundary. Either catch `Throwable` inside the coerced closure and complete the future yourself (`new CompletableFuturePromise(executor)` + `executor.execute { try { promise.complete(decorated.call()) } catch (Throwable t) { promise.completeExceptionally(t) } }`), or give `supplyAsync` a real `Supplier` (a Groovy lambda `() -> decorated.call()` compiles to invokedynamic, not a proxy). Apply the same to `asCompletableFuture` and to the `Runnable` in `WebPromises` (catch inside, stash, rethrow after `run()` returns). Add `IOException` variants of the three new `CompletableFuturePromiseFactorySpec` cases and one in `WebPromisesSpec`; they fail today.

### Medium

**N2. Standalone `WebPromises` propagates the request into `task {}` but not into `then {}`.**
The M1 fix decorates the *closure* (`WebPromises.groovy:53-76`), so the request is bound only for the duration of `work.call()` and reset in `GrailsWebRequestTaskDecorator`'s `finally` before `Supplier.get()` returns. `CompletableFuture.AsyncSupply.run()` then calls `postComplete()`, which is where `then {}` dependents (`CompletableFuturePromise.groovy:63`, plain `thenApply`) execute, so they run without a `GrailsWebRequest`. In 7.x `FutureTaskChildPromise.groovy:57` applied the factory decorators to every `then` callback, so `task { load() }.then { render it }` worked in `ControllerUnitTest` specs; it now silently loses `render`/`params` in the callback. The Boot path is fine because the `TaskDecorator` wraps the whole `run()`, including `postComplete()`.
*Probe (`WebPromises.promiseFactory = null`, latch so `then` is attached before the task completes):*
```
standalone: task sees request: true; then sees request: false
Boot-style executor + GrailsWebRequestTaskDecorator: then sees request: true
```
*Fix:* decorate the *executor*, not the closure, exactly as the plugin does: when `PromiseFactoryBuilder` yields a `CompletableFuturePromiseFactory`, build it over `Executors.newCachedThreadPool()` wrapped so each `Runnable` goes through `GrailsWebRequestTaskDecorator`; keep the lookup-strategy path only for third-party factories from `ServiceLoader`. That also removes the extra proxy boundary from N1. Add a `then {}` case to `WebPromisesSpec` that attaches the callback before the task completes (the current parameters test attaches nothing).

**N3. `spring.mvc.async.request-timeout` now fires by default for `WebPromises`, and a timeout renders as a logged 500, not a 503.**
Honouring the property is right, but the effect is now reachable in every app: Tomcat's 30 s default applies to `WebPromises.task {}` where it used to be infinite. When it fires, Spring's `TimeoutDeferredResultProcessingInterceptor` sets `AsyncRequestTimeoutException` as the concurrent result, `UrlMappingsInfoHandlerAdapter.groovy:94` rethrows it, and `GrailsExceptionResolver.java:166` unconditionally sets status 500 and logs the full stack trace at ERROR. Spring MVC's own resolvers map that exception to 503 without a stack trace. The docs (`asyncRequests.adoc:32`) say only that the timeout "ends the response lifecycle". Please either map `AsyncRequestTimeoutException` (it is an `ErrorResponse`) to its status code in `GrailsExceptionResolver`, or document that the app should add `"503"(controller: …, exception: AsyncRequestTimeoutException)` and that the 30 s Tomcat default now applies to eager web promises. A functional test that times out a `WebPromises.task {}` and asserts the status would pin this.

### Low

**N4. Silently dropping a promise returned after the async request completed.** `AsyncActionResultTransformer.groovy:51` returns `null` in that case, so the action's result is discarded, the task keeps running against a recycled request, and nothing is logged. `WebPromises.prepareAsyncRequest` throws `IllegalStateException` for the same situation; the transformer should be consistent (throw, or at least log at WARN) rather than swallow.

**N5. `GrailsAsyncWebRequest` naming and small nits.** The class is a static helper, not a web request, and it sits next to the deprecated `AsyncGrailsWebRequest`, which is going to confuse readers; `AsyncRequestSupport` or folding the two static methods into `WebPromises`/the transformer would be clearer. `GrailsAsyncWebRequest.groovy:41` uses `def` under `@CompileStatic`; and the timeout property is re-read and re-parsed with `DurationStyle` on every async request (`:44`). Reading the `WebMvcProperties` bean once, or caching the parsed value, would avoid that.

**N6. Fallback for a mistyped `applicationTaskExecutor` is silent.** `ControllersAsyncGrailsPlugin.groovy:76-79` catches `NoSuchBeanDefinitionException` with the same bean name, which also matches `BeanNotOfRequiredTypeException` when an app defines `applicationTaskExecutor` as a plain `Executor`. That mirrors Boot's MVC behaviour, so it is acceptable, but a DEBUG/INFO line saying which executor the promise factory ended up with would make the fallback diagnosable.

**N7. Squash still pending (round 1 L5).** The branch is now 18 first-parent commits including two upstream merges and the two URL-mapping commits that cancel each other out.

---

### Post-merge addendum (`b65a64ed69`, merge of upstream `8.0.x`)

The merge itself is clean for this PR's own code: none of the files behind N1–N7 changed, the `PromiseFactoryBuilder` conflict was resolved keeping `build(Executor)`, the merged `VirtualThreadPromiseFactorySpec` was extended to assert the builder ignores a supplied executor under the opt-in, and all three module suites pass on the merged head. Two things do change.

**N8 (Medium). The `grails.async.promiseFactory=virtual-thread` opt-in now bypasses everything this PR builds on.**
`PromiseFactoryBuilder.groovy:49-55` returns `new VirtualThreadPromiseFactory()` before looking at the executor, and the merged spec pins that. Under the opt-in, in a Boot app, `grailsPromiseFactory` is therefore a `VirtualThreadPromiseFactory`, which:
- is built on `FutureTaskPromise`, which this PR marks `@Deprecated(since = '8.0', forRemoval = true)` (`FutureTaskPromise.groovy:43`). Either the deprecation is premature, or the new factory needs to move off it before it ships in the same release.
- has no readable `executor` property (`private final ExecutorService executorService`), so `EventBusFactoryBean` falls through to the synchronous bus with the startup WARN. Round-1 H1 comes back for this mode, and `events.adoc:32` is wrong for it.
- gets neither Boot's `TaskDecorator` (it owns its executor) nor a `PromiseDecoratorLookupStrategy`, because `WebPromises.setPromiseFactory` no longer attaches one (`WebPromises.groovy:60`). On `8.0.x` today the old `setPromiseFactory` did attach `AsyncWebRequestPromiseDecoratorLookupStrategy`, so `WebPromises.task { render … }` works with the virtual-thread factory there and stops working with this PR. Same for `Promises.task {}` in a controller.
- keeps the old aggregate `onError` contract (returns an empty list instead of failing), so the "consistent `ExecutionException` semantics" section of the docs is only true for the default factory.
*Fix:* the simplest coherent option is to make `VirtualThreadPromiseFactory` a thin `CompletableFuturePromiseFactory` over `Executors.newVirtualThreadPerTaskExecutor()` (it then has the `executor` property, the same `get()`/`onError` semantics, `CompletionStage`, and no dependency on the deprecated classes), and have `ControllersAsyncGrailsPlugin` wrap whatever executor the built factory owns with the `TaskDecorator` beans. If that is out of scope for this PR, then at least restore the lookup strategy in `WebPromises.setPromiseFactory` for factories that are not `CompletableFuturePromiseFactory`, and say in `asyncPromises.adoc` that the opt-in does not share Boot's executor, event bus or request propagation. Add a plugin spec that sets the system property and asserts `grailsPromiseFactory` still propagates the request.

**N9 (Low). Needs another merge or rebase before it can land.** `git merge-tree` against the current `origin/8.0.x` conflicts in `upgrading80x.adoc`: sections 62 and 63 were added upstream after this branch merged, so "64. Promise Execution and Async Request Migration" will need renumbering again. Nothing else conflicts.

---

### Round 1 status

| Round 1 | Status | Notes |
|---|---|---|
| H1 event bus went synchronous | Fixed | `EventBusFactoryBean` reads the factory's `executor` property reflectively; `ExecutorEventBus` handles a plain `Executor` via `execute()`, no shutdown ownership. Covered by `TaskExecuterEventBusSpec`. |
| H2 `unwrap` stripped one cause level | Fixed | Only `CompletionException`/`ExecutionException` with a non-null cause are unwrapped; transformer passes the failure straight to `setErrorResult`. Covered with a caused exception in both specs. See N1 for the remaining wrapper. |
| H3 `NoUniqueBeanDefinitionException` with a second executor | Fixed | Resolved by name; spec registers `taskScheduler` + `otherExecutor`. |
| H4 single-threaded fallback | Fixed | `corePoolSize = 8`; two-promise latch test. |
| M1 standalone `WebPromises` lost the request | Partly fixed | Task closure OK; `then {}` not (N2). |
| M2 `get()` type differed between promise and child | Fixed | Both now `ExecutionException(original)`, direct and timed; the sneaky-throw overrides are gone. |
| M3 docs / upgrade note | Fixed | Upgrade section added (numbered 64 after the merge, see N9), Servlet API marked deprecated. |
| L1 aggregate `onError` type | Fixed | Both callbacks get the original. |
| L2 timeout consistency | Fixed | Both paths honour `spring.mvc.async.request-timeout` (see N3 for the consequence). |
| L3 task-after-completion guard | Fixed | `COMPLETED` request attribute set from a completion handler, checked in both entry points. |
| L4 decorator applies to `@Async` | Documented | |
| L5 squash | Open | N7 |

### Verified as correct in this round

- `CompletableFuture` semantics for the fixed `get()` path: a `supplyAsync` failure reaches `whenComplete` as `CompletionException`, `fromStage` completes the promise with the unwrapped original, and both the direct promise and its `thenApply` child report `ExecutionException(original)` from `get()` and `get(timeout)`. Spec matches observed behaviour.
- `WebAsyncManager` removes itself on completion (Spring 6.2+), so the `COMPLETED` attribute is needed; the completion handler runs in `AsyncListener.onComplete`, before Tomcat recycles the request, and the attribute is cleared with the request. Dispatch reuses the same request object with no `onComplete` in between, so the guard does not trip mid-dispatch.
- `DurationStyle.detectAndParse` defaults to milliseconds for a bare number, matching Boot's binding of `spring.mvc.async.request-timeout`; `spring-boot` is on the plugin's compile classpath transitively via `grails-core`'s `api` dependency.
- `ThreadPoolTaskExecutor` with `corePoolSize = 8` and the default unbounded queue matches Boot's defaults as documented.
- `EventBusFactoryBean`'s reflective lookup: `CompletableFuturePromiseFactory.executor` is a public final property; GPars/RxJava factories have no such property and keep the previous synchronous fallback.
- The `TaskDecorator`-on-executor design does propagate the request into `then {}` dependents in Boot mode (probe P3 above), because `AsyncSupply.run()` calls `postComplete()` inside the decorated `Runnable`.
- All three module suites and their code-style checks pass on a fresh, cache-bypassed run.

codeconsole:
Addressed the round-2 review in `69b234e3a5`, `683342d225`, `b698b3f540`, and `f3624cb8c3`, following upstream merge `b09d755655`.

- **N1:** Replaced the relevant proxy-based closure coercions with real Groovy lambdas. Added checked-exception coverage for direct/timed/chained results, completion and error callbacks, aggregate callbacks, non-CompletableFuture promise bridges, web tasks, and controller error handling. `IOException` now remains the original cause rather than an `UndeclaredThrowableException`.
- **N2:** Standalone `WebPromises` now decorates the owned executor, keeping request context bound through CompletableFuture completion and attached `then` callbacks. Third-party providers retain the compatibility lookup-strategy path. Latched tests cover default, virtual-thread, and legacy factories. Also fixed list task creation to use the receiving factory rather than the global factory, preserving standalone web-context propagation.
- **N3:** `GrailsExceptionResolver` now resolves `AsyncRequestTimeoutException` as 503, including status-specific URL mappings, without logging an ERROR stack trace. Added resolver coverage and a real HTTP timeout integration test. Documented the container default (normally 30 seconds on Tomcat), configuration, and that response timeout does not cancel the task.
- **N4:** Returning a promise after async request completion now throws `IllegalStateException`, consistent with submitting through `WebPromises`.
- **N5:** Renamed the static helper to `AsyncRequestSupport`, made the request variable explicitly typed, and cached the parsed timeout per application context using weak keys.
- **N6:** Added debug logging for executor selection and a regression test for a mistyped named executor. One clarification: `BeanNotOfRequiredTypeException` extends `BeansException`, not `NoSuchBeanDefinitionException`, so a mistyped `applicationTaskExecutor` already fails startup rather than silently selecting the fallback; the new test pins that behavior.
- **N8:** Converted `VirtualThreadPromiseFactory` to a thin CompletableFuture factory over its owned virtual-thread executor. It now shares the same failure/aggregate contracts, composes task decorators, propagates web request context, exposes its executor for asynchronous event delivery, and closes the owned executor on context shutdown. Plugin tests exercise request propagation, custom decoration, virtual-thread event delivery, and lifecycle cleanup. The changed aggregate contracts and distinction from Boot-managed virtual threads are documented.
- **N9:** Merged upstream `8.0.x` through `e749cc79a9`, retained both upstream upgrade sections, and moved the async migration section to 66. GitHub currently reports the PR as mergeable.
- **N7:** Kept the commits focused and unsquashed, as explicitly requested by the PR author. No AI trailers were added.

Verification: Java 21 module suites cover 427 tests (one existing skip); Java 25 async/MVC suites cover 156 (one existing skip, build cache disabled); the Java 21 sample integration suite passes all 18 tests, including the new timeout case. Affected-module style and dependency-version checks pass. These were resource-limited runs, not a repository-wide clean/aggregate sweep.

Updated the PR's combined summary and LOC table. Relative to the merged upstream base: 1,804 lines added, 256 removed, net +1,548; 1,018 retained compatibility lines are slated for future removal, leaving a projected net +530 overall, or -560 for production code alone.

matrei (approved):
## Review round 3

**Head:** `f3624cb8c3` (`feature/modernize-grails-async-8.0.x`, 23 first-parent commits) · **Base:** `8.0.x` @ `e749cc79a9`. The branch now contains everything on `origin/8.0.x` (0 commits behind) and `git merge-tree origin/8.0.x HEAD` is clean.

Five commits since round 2: the upstream merge `b09d755655` and four fix commits (`69b234e3a5`, `683342d225`, `b698b3f540`, `f3624cb8c3`).

**What I ran locally** (on `f3624cb8c3`, `--no-build-cache` + `cleanTest`, result XML timestamps checked)

| Check | Result |
|---|---|
| `:grails-async-core:test` | 81 tests, 0 failures, 1 skipped |
| `:grails-async:test` | 38 tests, 0 failures |
| `:grails-events-core:test` | 12 tests, 0 failures |
| `:grails-web-mvc:test` | 37 tests, 0 failures |
| `:grails-async-core:codeStyle :grails-async:codeStyle :grails-web-mvc:codeStyle` | pass |
| Throwaway probe spec (deleted afterwards) re-running the round-2 N1/N2 probes plus R1 below | output quoted below |
| `TimeoutControllerSpec` (new integration test) | not run locally; CI on this head is green (80 checks pass, 13 skipped) |

**Verdict: approve, with three optional low-severity notes.** Every round-2 finding is fixed and covered by a test that would have failed before. Nothing below needs to block the merge.

---

### Low

**R1. A user-supplied `CompletableFuturePromiseFactory(executor)` gets no request propagation from `WebPromises`.**
`WebPromises.groovy:66` skips the `WebRequestPromiseDecorator` lookup strategy for every `CompletableFuturePromiseFactory`, on the assumption that its executor is already decorated. That holds for the two factories Grails builds itself (plugin bean, standalone default), but not for `WebPromises.promiseFactory = new CompletableFuturePromiseFactory(myPool)`, which is a public constructor introduced by this PR.
*Probe:* `WebPromises.promiseFactory = new CompletableFuturePromiseFactory(Executors.newFixedThreadPool(2))`, bind a request, `WebPromises.task { GrailsWebRequest.lookup() != null }.get()` → `false`.
Either document that an executor passed to that constructor must already apply `GrailsWebRequestTaskDecorator`, or add a small marker (e.g. the factory remembers that it was built with an executor decorator) so that `setPromiseFactory` only skips the strategy when propagation is really in place.

**R2. `AsyncRequestSupport.timeoutFor` takes a JVM-wide lock on every async request** (`AsyncRequestSupport.groovy:54`, `synchronized (TIMEOUTS)` around a `WeakHashMap`). The critical section is tiny so this is not a throughput problem, but it is more machinery than the saving warrants (the original cost was one `Environment.getProperty` plus a `Duration` parse). A one-line alternative is to read Boot's already bound `WebMvcProperties` bean (`async.requestTimeout`) when present. Fine to leave as is.

**R3. Squash (round 1 L5 / round 2 N7).** Noted that the commits are kept separate on purpose. `479f5e0e2e` and `c982feb634` still cancel each other out, so please use "Squash and merge" (or drop those two) when landing, per the contribution guide.

### Correction to my round-2 N6

You are right: `BeanNotOfRequiredTypeException` extends `BeansException`, not `NoSuchBeanDefinitionException`, so a mistyped `applicationTaskExecutor` fails startup rather than silently falling back. My description was wrong; the new spec pins the correct behaviour and the DEBUG line covers the diagnosability part.

---

### Round 2 status

| Round 2 | Status | Notes |
|---|---|---|
| N1 checked exceptions wrapped in `UndeclaredThrowableException` | Fixed | All `as Supplier`/`as Runnable` coercions on the execution path are now lambdas (`CompletableFuturePromiseFactory.groovy:84,162`, `CompletableFuturePromise.groovy:63`, `GrailsWebRequestTaskDecorator.groovy:48`, `WebPromises.groovy:83`). Probe, default and virtual-thread factory, `throw new IOException()`: `get()` → `ExecutionException -> IOException`; `onError` → `IOException`; `then` child `get()` → `ExecutionException -> IOException`; aggregate `get()`/`onError` the same. `IOException` cases added to `CompletableFuturePromiseFactorySpec`, `WebPromisesSpec` and `AsyncActionResultTransformerSpec`. |
| N2 standalone `then {}` lost the request | Fixed | Standalone factory now owns a cached pool wrapped by `GrailsWebRequestTaskDecorator` (`WebPromises.groovy:55-60`), so `postComplete()` dependents run inside the decorated task. Latched spec covers default, virtual-thread and legacy factories. Probe: standalone `onError` callback and `WebPromises.tasks([...]).then {}` also see the request. The legacy decorator now captures the request at decoration time, which is correct for `FutureTaskChildPromise`. |
| N3 timeout was a logged 500 | Fixed | `GrailsExceptionResolver` uses the exception's own status (503) for the response, the error-status attribute and URL-mapping lookup, and logs at DEBUG only. With no `503` mapping the default `/error` view still renders (from the `java.lang.Exception` mapping in `ControllersAutoConfiguration`), now with status 503. Resolver spec plus a real HTTP `TimeoutControllerSpec`; docs state the Tomcat 30 s default. |
| N4 promise after completion silently dropped | Fixed | Throws `IllegalStateException`, spec added. |
| N5 helper naming / `def` / re-parsing | Fixed | Renamed to `AsyncRequestSupport`; see R2 for the cache. |
| N6 silent fallback | Fixed / my error | See correction above. |
| N7 squash | Open by choice | R3 |
| N8 virtual-thread opt-in bypassed the PR | Fixed | `VirtualThreadPromiseFactory` is now a 3-line `CompletableFuturePromiseFactory` over an owned virtual-thread executor: no dependency on the deprecated `FutureTask*` classes, exposes `executor` for the event bus, composes all `TaskDecorator` beans through the plugin, same failure contracts, closed on context shutdown. Plugin spec asserts virtual thread + request for `Promises.task`, `WebPromises.task` and event delivery, custom decorator invoked, and `RejectedExecutionException` after `context.close()`. The changed aggregate `onComplete`/`onError` contract for the opt-in is called out in the upgrade notes. |
| N9 conflict with `8.0.x` | Fixed | Merged through `e749cc79a9`; async section renumbered to 66. |

### Verified as correct in this round

- `CompletableFuturePromiseFactory` ownership: only the protected `(ExecutorService, UnaryOperator)` constructor records an owned executor, and `close()` shuts down the undecorated `ExecutorService`, not the decorating lambda. A Boot-supplied executor is still never closed by the factory.
- `PromiseFactoryBuilder.build(executor, decorator)`: the decorator is applied only to executors the factory owns (standalone default and virtual-thread opt-in); Boot's `applicationTaskExecutor` and the `grailsPromiseExecutor` fallback are decorated by Boot / the fallback bean itself, so nothing is decorated twice.
- The new `createPromise(List<Closure>, List<PromiseDecorator>)` override routes list tasks through the receiving factory, so standalone `WebPromises.tasks([...])` uses the request-aware executor (spec + probe).
- `whenComplete`/`addCompletionHandler` closures that remain coerced cannot leak a checked exception: the `onError` one catches `Throwable`, the others do not throw.
- Docs: `asyncPromises.adoc`, `asyncRequests.adoc`, `events.adoc` and the upgrade section now match the behaviour I observed, including the "default and virtual-thread factories only" qualifier on the failure contract.
- All four module suites and the code-style checks pass on a fresh, cache-bypassed run; CI is green on the same SHA.

grails-async/plugin/src/main/groovy/org/grails/plugins/web/async/ControllersAsyncGrailsPlugin.groovy
```diff
@@ -19,14 +19,27 @@
...
+import grails.async.Promises
 import grails.plugins.Plugin
+import grails.async.web.WebPromises
+import org.grails.async.factory.PromiseFactoryBuilder
+import org.grails.async.factory.future.VirtualThreadPromiseFactory
 import org.grails.plugins.web.async.mvc.AsyncActionResultTransformer
-import org.grails.plugins.web.async.spring.PromiseFactoryBean
 
 /**
  * Async support for the Grails 2.0. Doesn't do much right now, most logic handled
```
jdaugherty: Lets update the comment?

jdaugherty (approved):
I discussed this with Mattias and agree we can go ahead and merge.

PR #16231, opened 2026-08-26
Fix document publishing

jdaugherty:
See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency for reference.

The queue size for our concurrency group was set to 1 instead of a larger value, so this means the doc publish jobs get cancelled if another exist. This should fix our publishing problem.


matrei:
Won't we still have a problem as we are pushing from multiple repositories?

PR #16235, opened 2026-08-27
Upgrade Groovy to 5.1.1

sbglasius:
## Summary
- Bumps `groovy.version` from `5.1.0` to `5.1.1` in `dependencies.gradle` (both the primary BOM pin and the duplicate literal used by `grails-micronaut-bom`/`grails-hibernate5-micronaut-bom`)
- Fixes stale "5.0.x" Groovy version references in `AGENTS.md`/`CLAUDE.md` and the `groovy-developer`, `gradle-developer`, and `grails-developer` skill docs, which had drifted from the prior 5.1.0 bump

Groovy 5.1.1 is a maintenance release on the 5.1.x line, fixing:
- GROOVY-12142: Metaspace leak on parallel Tomcat deployments
- GROOVY-12166: static type checking regression causing `ClassCastException` on `instanceof`-narrowed fields
- GROOVY-12257: Groovy 4/5 `@Immutable` bytecode incompatibility
- GROOVY-12286: bumps its own Jackson dependency to 2.22.2 (already independently pinned in this repo's BOM, so no further change needed here)

See the [changelog](http://groovy-lang.org/changelogs/changelog-5.1.1.html).

## Test plan
- [x] `./gradlew clean build -PskipTests` succeeds
- [x] `./gradlew validateDependencyVersions` — one pre-existing, unrelated failure (`org.graalvm.sdk:nativeimage` version drift in the micronaut example projects) confirmed to also occur on unmodified 5.1.0, so not introduced by this change
- [ ] Full test suite / style / violation checks left to CI per project convention


jdaugherty (approved):

matrei (approved):

sbglasius:
I cannot merge this as long as there is an OutOfMemory Exception being thrown. On slack @paulk-asert  mentioned:

We have had another user report increased memory requirements for 5.1.1 over 5.1.0, the suggestion for them while we try to find a better fix is to bump up their mem spec to `-Xmx512m`.

You could also potentially test against:
https://github.com/apache/groovy/pull/2840

And @jamesfredley has created a potential fix for it here https://github.com/apache/grails-core/pull/16167 

So perhaps 5.1.1 is not suitable as it is?

@matrei / @jdaugherty since you approved this, what are your thoughts?

matrei:
> since you approved this, what are your thoughts?

@sbglasius I think we hold off 5.1.1 until there is a fix for the memory problem.

sbglasius:
I have changed this PR to draft for the time being, until the issue with Groovy 5.1.1 is resolved.

PR #16269, opened 2026-08-29
Register a late ConfigurationClassPostProcessor for resources.groovy/xml

### LINKED ISSUE #16029: 8.0.0-M3 - `@Configuration` classes registered in `resources.groovy` are not processed by a `ConfigurationClassPostProcessor` anymore
### Expected Behavior

`@Configuration` classes registered in `resources.groovy` are processed by a `ConfigurationClassPostProcessor`.

### Actual Behaviour

`@Configuration` classes registered in `resources.groovy` are not processed by a `ConfigurationClassPostProcessor` anymore.

* from taking a look at the grails-8 plugin bean registration changes (which are great btw.), i think that the `ConfigurationClassPostProcessor`registered in `CoreGrailsPlugin` now runs before an apps' `resources.groovy` beans are registered?  
  * ref. https://github.com/apache/grails-core/blob/v8.0.0-M3/grails-core/src/main/groovy/org/grails/plugins/CoreGrailsPlugin.groovy#L75-L76

### Steps To Reproduce

* run grails-8 sample app
  * no "sample bean registered" console out
* run grails-7 sample app
  * "sample bean registered" console out

### Environment Information

* grails-8.0.0-M3 (vs. grails-7.2.1)
* ubuntu-24.04
* jdk-21

### Example Application

* grails-8.0.0-M3: https://github.com/zyro23/grails-core-16029-grails-8
* grails-7.2.1: https://github.com/zyro23/grails-core-16029-grails-7

### Version

8.0.0-M3

borinquenkid:
## Summary
- Fixes #16029: `@Configuration` classes registered in an application's `resources.groovy` (and `resources.xml`, and the application's own `doWithSpring`/`beanRegistrar()`) were no longer processed by a `ConfigurationClassPostProcessor` in 8.0.0-M3.
- Root cause: `GrailsApplicationPostProcessor` (which drains those late-arriving beans) is discovered as an ordinary registry bean, so Spring always invokes it *after* its own `ConfigurationClassPostProcessor` has already completed its one and only pass over the registry. `CoreGrailsPlugin` used to unconditionally register a second processor that happened to catch this late content, but a later change made it stand down whenever Spring's own processor is already present — correct for the early plugin-registration phase, but wrong for this late phase, since Spring's processor has always already run there.
- Fix registers a second `ConfigurationClassPostProcessor` at the end of `GrailsApplicationPostProcessor.postProcessBeanDefinitionRegistry`, after all late-phase beans are drained. Spring marks every bean definition a processor has already parsed, so this second pass only expands what's genuinely new; it's skipped under AOT, matching `CoreGrailsPlugin`'s existing guard.

## Test plan
- [x] Added `LateConfigurationClassPostProcessorSpec`, which reproduces the exact bug shape and was confirmed to fail without the fix and pass with it.
- [x] Full `grails-core` test suite green (1102 suites, 0 failures/errors).
- [x] `./gradlew :grails-core:codeStyle` clean.

### DIFF
--- a/grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
+++ b/grails-core/src/main/groovy/grails/boot/config/GrailsApplicationPostProcessor.groovy
@@ -30,6 +30,7 @@ import org.springframework.beans.factory.config.ConfigurableListableBeanFactory
 import org.springframework.beans.factory.support.BeanDefinitionRegistry
 import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor
 import org.springframework.context.annotation.AnnotationConfigUtils
+import org.springframework.context.annotation.ConfigurationClassPostProcessor
 import org.springframework.aot.AotDetector
 import org.springframework.beans.factory.support.RootBeanDefinition
 import org.springframework.beans.factory.config.BeanDefinition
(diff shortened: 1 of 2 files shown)


jdaugherty:
I don't think we should fix this. Resources.groovy was never meant to be anything other than beans and longer term we want to remove it.

PR #16272, opened 2026-08-29
Resolve resource links to the controller exposing the domain class

codeconsole:
A `resource` link derives its controller from the domain class name, so a controller not named after its domain class is never found:

```groovy
class PeopleController extends RestfulController<Person> {
    PeopleController() { super(Person) }
}
```

```gsp
<g:link resource="${person}">          <%-- targeted "person", which does not exist --%>
```

`DefaultLinkGenerator` now keeps a domain-class-to-controller index alongside the existing namespace index, built lazily and guarded by the same artefact-array identity check, so it costs nothing on the common path and rebuilds when the set of controllers changes.

**A controller named after the domain class always wins.** The index is consulted only when no such controller is registered, so an application following the naming convention is unaffected — including one that has both a plain `BookController` and a `BookRestfulController extends RestfulController<Book>`.

A controller's domain class is found by walking its supertypes for a generic type argument the mapping context recognises as a persistent entity. Superclasses and interfaces are both walked, so a base class or a Groovy trait at any depth works, without this class depending on the REST controller hierarchy:

```groovy
class PeopleController extends RestfulController<Person> {}            // resolved
class WidgetsController extends WidgetControllerBase {}                // resolved through the base
class TagsController implements ResourceHolder<Tag> {}                 // resolved through a trait
```

A supertype declaring more than one persistent entity is ambiguous and is skipped rather than guessed at, with the walk continuing upwards. If no controller declares the domain class, or more than one does, the link falls back to the domain class name.

Applies only when `resource` is given a domain instance and GORM is configured; passing a controller name directly (`resource: "book"`) is unchanged.

An association still held as a lazy proxy is rendered from the domain class rather than the instance, to avoid loading it. A domain class passed to the `resource` attribute resolves through the same rule, so a proxied and a loaded association produce the same href.


grails-web-url-mappings/src/main/groovy/org/grails/web/mapping/DefaultLinkGenerator.groovy
```diff
@@ -93,6 +94,9 @@ class DefaultLinkGenerator implements LinkGenerator, PluginManagerAware {
     private volatile Map<String, Set<String>> controllerNamespacesByName
     private volatile GrailsClass[] cachedControllers
 
+    private volatile Map<String, Set<String>> controllerNamesByDomainClass
```
jdaugherty, line 97: We should discuss this further on the weekly or mailing list.  This assumes all applications use the same design here and it's formalizing something that may not be true for all applications.  This is also a major shift, not necessarily a bad one, but it will require another milestone.  Thus this should go into 8.1 or later.

codeconsole: Agreed — worth the weekly or the mailing list before this goes anywhere. No objection to holding it for a later milestone.

matrei (commented):
Reviewed at head `c88e535bf0` (`fix/resource-link-controller-lookup`, 26 commits) against `8.0.x`. Merge base `b49a48e2ec`; `8.0.x` has moved 15 commits since, and `git merge-tree` reports no conflicts. CI is green on the head.

Run locally at the head with `--no-build-cache` after `cleanTest`:

| Module | Tests | Result |
|---|---|---|
| `grails-web-url-mappings` | 310 | pass |
| `grails-controllers` | 91 (1 skipped) | pass |
| `grails-scaffolding` | 39 | pass |
| `grails-views-gson` | 186 | pass |
| `grails-fields` | 685 (8 skipped) | pass |
| `grails-rest-transforms` | 47 (1 skipped) | pass |
| `grails-test-suite-uber` (the four new specs plus `RedirectMethodTests`) | 35 | pass |

I also ran a throwaway spec against the head to probe the edge cases behind the findings below; it is not part of the PR.

The design reads well: one scope chain for both `controller` names and `resource` classes, the cache keyed on the resolved target, and the redirect resolved from the issuing namespace instead of into it. The findings are small.

### 1. A resource link that names its `controller` still logs the ambiguity warning

`resolveLinkTarget` resolves the resource target before it looks at the `controller` attribute (`DefaultLinkGenerator.groovy:273` and `:283`), and only at `:293` does the explicit controller replace the result. So for a domain class two controllers serve, with neither named after it, this

```groovy
generator.link(resource: new Gadget(id: 3), controller: 'gadgets', action: 'show')
```

produces the right link and also logs, once per class, that the link "names no controller ... pass a controller attribute to choose one", although the caller did exactly that. Verified with the `DomainLinkFixtures` controllers on the head. It is also wasted work on every call, including every cache lookup, since `resolvedLinkKey` goes through the same path. Skipping `resolveResourceTarget` when `truthy(controllerAttribute)` (falling through to the decapitalized name as the class-without-entity branch does) fixes both, and a test asserting no warning with an explicit `controller` would pin it.

Related nit at `:325`: the namespace pin checks `controllerAttribute == null` while `:293` checks `truthy(controllerAttribute)`, so `controller: ''` takes the resolved controller but infers the namespace by name rather than using the pinned one. The two happen to agree in the request-controller case; using the same test in both places keeps them agreeing everywhere.

### 2. The "reported once" sets outlive the index they describe

`getControllerIndex()` (`:443`) rebuilds the index whenever the artefact array or mapping context changes, but `ambiguousControllersReported` and `ambiguousResourcesReported` are only cleared by `resetControllerNamespaceCache()` (`:498`), which only `CachingLinkGenerator.clearCache()` calls. A plain `DefaultLinkGenerator` in development mode therefore never re-reports: after a reload that introduces a different ambiguity for a name or class reported earlier, nothing is logged. Holding the two sets inside `ControllerIndex` so they are born and discarded with it would make the warning track the controllers it describes. Nit.

### 3. Behaviour change worth an explicit decision: a `controller` name nothing defines now stays in the request namespace

`getDefaultNamespace` returns the request namespace when no registered controller has the name (`:409`–`:412`), where 8.0.x returned `null`. The PR needs this so a redirect to an undefined controller keeps producing the URL it produced when the issuing namespace was forced onto it. The cost is on the link side: from a namespaced controller, `<g:link controller="nowhere">` moves from `/nowhere` to `/admin/nowhere`, and so does `g.createLink` in a `ControllerUnitTest` for a namespaced controller, where the other controller is usually not registered. Existing application unit tests asserting such URLs will change. The docs and upgrade notes state the rule, so this is not a defect, but given the discussion on the PR about timing it is the part of the change I would call out on the list explicitly. An alternative that keeps links as they were is to have `Controller.redirect` fall back to the old explicit namespace only when the name resolves to nothing, and leave `getDefaultNamespace` returning `null` for an unknown name.

### 4. Nit: `issuingNamespace` is stored on the request as an `Object`

`Controller.groovy:259` and `:273` put whatever `resolveNamespace(Class)` returned into `CONTROLLER_NAMESPACE_ATTRIBUTE`; for an unregistered controller that is the raw static field value. `GrailsWebRequest.getControllerNamespace()` casts it to `String`, so a `GString` `namespace` on a controller constructed directly in a test now throws where the old `argMap` path called `toString()`. `issuingNamespace?.toString()` closes it.

### 5. Nit: the warning assertions capture `System.err` where `LogCapture` would do

`LinkGeneratorNamespaceInferenceSpec` and `LinkGeneratorResourceControllerSpec` assert on the log by swapping `System.err`, which relies on `slf4j-simple` writing there at the module's `info` default level. `grails-core` publishes `org.apache.grails.core.testing.support.LogCapture` in its test fixtures for exactly this: a Logback `ListAppender` on one logger, restored on `close()`, already used by `grails-web-mvc` and by the `grails-core` specs. Wiring it here is what `grails-web-mvc` does:

- `testImplementation(testFixtures(project(':grails-core')))` in `grails-web-url-mappings/build.gradle` (no cycle: `grails-core` does not depend on this module, and `logback-classic` comes with the fixtures)
- replace `testImplementation 'org.slf4j:slf4j-simple'` and `src/test/resources/simplelogger.properties` with a `logback-test.xml` holding `<root level="WARN"/>`, so two bindings are not on the test classpath

The specs then read `logCapture.events` for `DefaultLinkGenerator` instead of parsing stderr, and the "reported once" assertions stop depending on which binding is active or on what else writes to stderr in the fork.

### Verified as correct

- `DefaultUrlMappingsHolder.java:320`: the any-method retry was computed and discarded on 8.0.x, so a method-specific mapping needing other parameters pushed the link to the wildcard. Real fix, now pinned by `UrlMappingsWithHttpMethodSpec`.
- `CachingLinkGenerator`: keying on `makeKey` plus the resolved target plus encoding covers every input resolution reads (request controller, namespace, HTTP method); the `makeKey`-override test covers the asset pipeline subclass. The same-attributes-across-methods and across-encodings bugs on 8.0.x are real and both are now tested.
- `ControllerIndex` published as one volatile reference with the array and mapping context it was built from; the identity check rebuilds on late registration, reload, or a mapping context arriving after first use.
- `domainClassNameFor` walks superclasses and interfaces via `ResolvableType`, resolves type variables through intermediate bases, skips a level declaring two entities, and matches on the mapping context rather than on `RestfulController`, so `@Scaffold` (which parameterises the superclass at compile time), `RestfulServiceController<T>` and user traits all qualify.
- `servingControllers` filters by the action the link targets, so a report controller declaring the class is not sent `show`; when nothing serves the action the old derived name is used, which also keeps runtime-scaffolded controllers whose actions are not in `GrailsControllerClass.actions` on the old path.
- `truthy` matches Groovy truth for the value types that reach it (null, CharSequence, Boolean, Number, Collection, Map, else `castToBoolean`).
- `Controller.redirect` restores the previous request attribute in `finally`; an explicit `namespace` still bypasses the swap.
- `FormFieldsTagLib`, `AbstractLinkingRenderer`, `DefaultHalViewHelper`: passing the domain class instead of its decapitalized name is equivalent when no mapping context knows the class, and the scaffolding views and the fields `_table.gsp` already pass `controller` explicitly, so the templates needed no change.
- `hasNamespacedControllers()` shipped only in `v8.0.0-M3` to `M6`, as the description says.
- Generated-controller templates: `redirect action: "show", id: ..., method: "GET"` reaches the same reverse mapping `redirect instance` did (`GET` + `show`).
- Docs: the guide sections, the scaffolding note and the upgrade entry match the code; the CSV table syntax is used elsewhere in the guide.

codeconsole:
Thanks @matrei, all five are addressed in the push up to `e81621a6c2`.

1. **A resource link naming its `controller`** (`031440d091`): such a link no longer resolves the controllers serving the domain class, so it reports no ambiguity and pays for no resolution. The namespace a resolved link carries now applies whenever the link was resolved, so `controller: ''` resolves as an absent attribute does throughout. Tests cover no warning with an explicit `controller`, and a blank-versus-absent case where the name alone is ambiguous (`print` and `archive` `PamphletsController`, only one of which has `show`), which failed before.

2. **The reported-once sets** (`bf9c544d5b`): both now live in `ControllerIndex`, so they are discarded with it whenever the controllers change; `resetControllerNamespaceCache()` just drops the index. The tests reload a controller on a plain `DefaultLinkGenerator` and assert the warning is logged again.

3. **An unknown controller name keeps the request namespace**: I've kept this. It is what makes a link and a redirect to the same unqualified name agree, and a redirect to an unknown name has always stayed in the issuing namespace (7.0.x put the issuing controller's `namespace` on every redirect naming none). Falling back only in `redirect` would bring back the link/redirect split for exactly those names. `2edef2b164` spells out where it shows in the upgrade notes: a namespaced controller's unit test usually registers no other controller, so `g.createLink(controller: 'other', ...)` there now includes the namespace, and the fix is to register that controller or pass `namespace`. Agreed that this is the part to call out explicitly in the list discussion.

4. **`issuingNamespace` stored as an `Object`** (`cfbaf5c52a`): `resolveNamespace(Class)` now returns a `String`, with `?.toString()` on the static-field fallback. `ControllerRedirectSpec` reads the namespace through `GrailsWebRequest.controllerNamespace`, as link generation does, and a controller declaring its namespace as a `GString` threw the `ClassCastException` there before the fix.

5. **`LogCapture`** (`e87bbf388f`): `testFixtures(project(':grails-core'))` in `grails-web-url-mappings`, with `slf4j-simple` and `simplelogger.properties` replaced by a `logback-test.xml` holding `<root level="WARN"/>`. The specs read `logCapture.events` for `DefaultLinkGenerator`.

I also renamed the private `getControllerIndex()` to `currentControllerIndex()` (`d142808b20`): it shared a name with the `controllerIndex` field it rebuilds, so rewriting a call as property access would have skipped the rebuild.

PR #16275, opened 2026-08-30
Add optional grails-openapi module for springdoc integration

codeconsole:
Adds an optional `grails-openapi` module so a Grails application can publish an OpenAPI description of its REST endpoints and browse it with Swagger UI, plus an `openapi` forge feature that selects it.

springdoc builds its document by scanning Spring MVC handler methods. Grails dispatches through `UrlMappingsHandlerMapping` rather than `@RequestMapping` handler methods, so springdoc on its own serves `/v3/api-docs` with no paths. This module supplies them through springdoc's `OpenApiCustomizer` SPI, and describes domain classes and command objects through swagger-core.

Everything is derived. The annotations below are available to correct or enrich what is derived, and none of them are required.

## Usage

Select `openapi` when generating an application, or add it to an existing one:

```groovy
dependencies {
    // the OpenAPI document, served at /v3/api-docs
    implementation 'org.apache.grails:grails-openapi'

    // optional: Swagger UI, served at /swagger-ui/index.html
    implementation 'org.springdoc:springdoc-openapi-starter-webmvc-ui'
}
```

The module produces the document; Swagger UI is a separate choice. The document is equally useful to a code generator or a contract test, and Swagger UI is not the only viewer that reads it, so the module does not pre-decide one. The forge feature selects both.

No further configuration is required. Given:

```groovy
class Book {
    String title
    String genre

    static constraints = {
        title blank: false, nullable: false, maxSize: 255
        genre nullable: true, inList: ['scifi', 'fantasy', 'history']
    }
}

class UrlMappings {
    static mappings = {
        "/books"(resources: 'book')
    }
}
```

`/v3/api-docs` describes the operations the mapping generates, and:

```json
"Book": {
  "type": "object",
  "properties": {
    "id":      { "type": "integer", "format": "int64", "readOnly": true },
    "title":   { "type": "string", "maxLength": 255 },
    "genre":   { "type": "string", "enum": ["scifi", "fantasy", "history"] },
    "version": { "type": "integer", "format": "int64", "readOnly": true }
  },
  "required": ["title"]
}
```

The identifier and version are marked `readOnly` because the server assigns them, so one schema describes both directions.

## Constraint mapping

| Constraint | OpenAPI |
|---|---|
| `nullable: false` | `required` |
| `maxSize` / `size` | `maxLength`, `minLength` |
| `min` / `max` / `range` | `minimum`, `maximum` |
| `inList` | `enum` |
| `matches` | `pattern` |
| `email` / `url` | `format` |

## Responses

A `RestfulController` action is described from what the controller does, rather than as a uniform `200`:

| Action | Status | Body |
|---|---|---|
| `index` | 200 | a collection of the resource |
| `show`, `edit` | 200, 404 | the resource |
| `create` | 200 | the resource |
| `save` | 201, 422 | the resource |
| `update`, `patch` | 200, 404, 422 | the resource |
| `delete` | 204, 404 | none |

## Request bodies

An operation that accepts a body describes the type the action binds. Where the action takes a command object that is the command, in preference to the resource the controller is named for:

```groovy
class OrdersController {
    def submit(OrderCommand cmd) { }
}

class OrderCommand implements Validateable {
    String customerEmail
    Integer quantity

    static constraints = {
        customerEmail email: true, nullable: false
        quantity min: 1, nullable: true
    }
}
```

`POST /orders/submit` is described as accepting an `OrderCommand`, with the constraints the command declares carried across the same way a domain class's are.

## Controllers reached through the default mapping

A `RestfulController` is also served by the default `"/$controller/$action?/$id?"` mapping, so those routes are documented too, including for controllers no mapping names, which covers scaffolded controllers:

```
GET    /book/index          POST   /book/save
GET    /book/show/{id}      PUT    /book/update/{id}
GET    /book/create         PATCH  /book/patch/{id}
GET    /book/edit/{id}      DELETE /book/delete/{id}
```

The method comes from the controller's `allowedMethods`. A controller a `resources` mapping also names appears at both `/books/{id}` and `/book/show/{id}`, because both answer; an application wanting only the first form removes the default mapping from `UrlMappings.groovy`.

Only `RestfulController` subclasses are described this way, so controllers rendering GSP views stay out of the document.

## Describing a domain class

`@Schema` on the class or on a property supplies what the constraints cannot:

```groovy
import io.swagger.v3.oas.annotations.media.Schema

@Schema(description = 'A book in the catalogue')
class Book {

    @Schema(description = 'Full title as printed', example = 'Dune')
    String title

    static constraints = {
        title blank: false, nullable: false, maxSize: 255
    }
}
```

The two combine on the same property:

```json
"title": {
  "type": "string",
  "description": "Full title as printed",
  "example": "Dune",
  "maxLength": 255
}
```

## Describing an action

```groovy
import io.swagger.v3.oas.annotations.Operation

@Operation(summary = 'List the catalogue',
        description = 'Every book, most recently added first.',
        operationId = 'listBooks',
        tags = ['Catalogue'])
def index() { }
```

`@Operation` supplies the summary, description, operation id, tags, and whether the operation is deprecated. A value it does not set is left as derived, so an action can be given a summary without restating anything else.

## Adding a response

```groovy
import io.swagger.v3.oas.annotations.responses.ApiResponse

@ApiResponse(responseCode = '403', description = 'Not your book')
def show() { }
```

The declared response is added alongside the derived `200` and `404` rather than replacing them.

## Naming a response type

An action's return type is not declared, so a controller responding with something other than the resource it is named for can say so:

```groovy
import io.swagger.v3.oas.annotations.media.Content
import io.swagger.v3.oas.annotations.media.Schema
import io.swagger.v3.oas.annotations.responses.ApiResponse

@ApiResponse(responseCode = '200',
        content = @Content(schema = @Schema(implementation = BookSummary)))
def summary() { }
```

The declared type replaces the convention and its schema is described alongside the others.

## Withholding an endpoint

```groovy
import io.swagger.v3.oas.annotations.Hidden

@Hidden
def internalAudit() { }
```

`@Hidden` on the controller withholds all of it, and `@Operation(hidden = true)` does the same as on an action. This is how an endpoint that is reachable but not part of the published API is kept out.

## Limitations

- A mapping that accepts any HTTP method is documented as `GET`, because OpenAPI requires a concrete operation.
- A controller that is not a `RestfulController` is described as responding with the resource it is named for. That is a convention rather than something that can be determined; name the type with `@ApiResponse` where it differs.
- The mapping context is injected optionally, so an application without GORM still gets a document describing its paths, without schemas.
- An application whose URL mappings include a catch-all resolving to a view or URI needs `static excludes = ['/swagger-ui/**', '/v3/api-docs/**']`, which would otherwise match the springdoc paths.


zyro23:
looks promising! will it be possible to customize paths/schemas by annotating actions/domain classes/command objects/response objects with swagger annotations?

codeconsole:
> looks promising! will it be possible to customize paths/schemas by annotating actions/domain classes/command objects/response objects with swagger annotations?

Thanks @zyro23  - please take a look at the updates in the description. LMK if there is anything else you think might be beneficial.

zyro23:
> > looks promising! will it be possible to customize paths/schemas by annotating actions/domain classes/command objects/response objects with swagger annotations?
> 
> Thanks @zyro23 - please take a look at the updates in the description. LMK if there is anything else you think might be beneficial.

in fact, i think there is. but first: hats off for the changes. awesome.

springdoc allows defining [`GroupedOpenApi`](https://github.com/springdoc/springdoc-openapi/blob/v3.1.0/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/models/GroupedOpenApi.java) beans to define one or multiple "grouped" openapi definitions (e.g. for separate apis):

https://springdoc.org/#how-can-i-define-multiple-openapi-definitions-in-one-spring-boot-project

`GroupedOpenApi` (or rather `GroupedOpenApi.Builder`) supports filtering (at least for spring(-mvc) apps) by:
* `pathsToMatch`
* `packagesToScan`
* `packagesToExclude`
* `pathsToExclude`
* `producesToMatch`
* `headersToMatch`
* `consumesToMatch`
* `methodFilters`

to what extend is/could that be supported by the grails impl.?

thanks & regards.

matrei:
Reviewed at head `11b3d985bf` (`feat/openapi-springdoc-8.0.x`, 19 commits) against `8.0.x`. Merge base `3dc36325dc`; `8.0.x` has taken 33 merges since, and `git merge-tree` reports no conflicts. CI is green on the head.

Run locally at the head:

| Module | Tests | Result |
|---|---|---|
| `grails-openapi` (`check`: tests, CodeNarc, Checkstyle) | 76 | pass, no violations |
| `grails-test-examples-openapi` (`integrationTest`) | 10 | pass |
| `grails-forge-core` (`OpenApiSpec`) | 3 | pass |
| `validateDependencyVersions` for both new projects | | pass |

I also ran the functional app with the default `"/$controller/$action?/$id?(.$format)?"` mapping added and dumped `/v3/api-docs` to check what an application actually gets, since the shipped example does not exercise that path (see 4). The document was clean: `Book` carried only `title`, `genre`, `id` and `version`, the expanded `/book/...` routes appeared alongside `/books`, every reference resolved, and every operation id was distinct. The `@Autowired(required = false)` property injection of `grailsApplication` works in a real context. I also checked the command object rule in `ActionAnnotations.isCommandObject` against `ControllerActionTransformer` and it mirrors it exactly, including the `Serializable` exclusion.

The design is sound: derive everything from the mappings and the mapping context, let swagger-core resolve types, overlay the GORM constraints, and let the standard annotations correct the result. The findings are mostly leftovers from refactoring plus two behavioural gaps.

### 1. Dead code in `PersistentEntitySchemaBuilder`

`retainReferenced` (`:208`), `referencedNames` (`:222`) and `collectReference` (`:233`) are private and nothing calls them; `buildCommand` walks declared fields instead. CodeNarc does not flag them because the repo does not enable `UnusedPrivateMethod`, but they should go, along with the Javadoc at `:204` that describes them.

### 2. Orphaned and stacked Javadoc from earlier refactors

Groovydoc is published for this module, so these surface in the API docs. Each is a Javadoc block sitting on the wrong member or stacked on top of another block:

- `UrlMappingsOpenApiCustomizer.groovy:145` describes `addExpandedMappings` but is attached to `addMappedOperation`; the real doc follows at `:151`.
- `UrlMappingsOpenApiCustomizer.groovy:423` and `:427` describe a schema registration and a tag description method that no longer exist at that spot; both are stacked over the `disambiguateOperationIds` doc at `:431`.
- `UrlMappingsOpenApiCustomizer.groovy:586` describes `asStaticName` but sits on `describe`.
- `UrlMappingsOpenApiCustomizer.groovy:64` (class Javadoc) says mappings whose controller is only known per request "are skipped", which `addExpandedMappings` contradicts.
- `PersistentEntitySchemaBuilder.groovy:143` describes `declaredPropertyTypes` but is stacked over the `collectTypeArguments` doc at `:147`.
- `RestfulControllerActions.groovy:40` describes `DEFAULT_METHODS` but is attached to `DEFAULT_SUCCESS_CODE`.

### 3. An application with more than one datastore gets no schemas, silently

`setMappingContextProvider` uses `getIfUnique()`, so with Hibernate and MongoDB together (two `MappingContext` beans) `mappingContext` is `null` and every operation is described without a shape. The Javadoc at `:93` acknowledges the "exactly one" condition, but the user gets no hint why the schemas vanished. Two options, in order of preference:

- Iterate `provider.orderedStream()` and index the entities of every context, resolving the validator from the context that owns each entity. `indexEntitiesByController` and `constraintsFor` are the only two places that need the owning context.
- Or keep the single-context rule and log at INFO when the provider returned nothing because it was ambiguous, and say so in the guide's limitations list.

### 4. The functional example does not cover the expansion path

`grails-test-examples/openapi/.../UrlMappings.groovy` has only the `resources` mapping, so `addExpandedMappings`, and with it the injection of `grailsApplication`, is only exercised by unit specs that wire the customizer by hand. The PR body and guide advertise the default mapping routes as a headline feature. Adding `"/$controller/$action?/$id?(.$format)?" {}` to the example and one assertion on `/book/show/{id}` would make the functional spec cover it. I confirmed it passes.

### 5. A mapping that names a controller but no action is described as an action-less operation

For `"/catalogue"(controller: 'book')` Grails dispatches to the controller's default action (`index`). `addMappedOperation` passes a `null` action through, so the operation id becomes `book_get`, a `RestfulController` falls into the generic 200 branch at `:736` with no paging parameters and a single-resource schema rather than the collection, and the action annotations are not consulted. Defaulting `mappedAction` to the controller's default action (`GrailsControllerClass.defaultAction`, `index` when unset) when the mapping names a controller only would make that case consistent with the rest.

### 6. Guide: nothing says the document is public

`openApi.adoc` never mentions that `/v3/api-docs` and Swagger UI are served to anyone and enumerate every REST route, domain class and constraint. One sentence pointing at securing the two paths with whatever the application uses, or disabling them per environment (`springdoc.api-docs.enabled: false`), belongs under Getting Started or Customizing.

### 7. Drop the `@author` tags

All five new classes under `grails-openapi/src/main` carry `@author Scott Murphy Heiberg` in their class Javadoc. We should not add `@author` tags to new files; git history records authorship, and the tag goes stale as soon as someone else touches the class. Please remove them.

### 8. Nits

- `settings.gradle:573`: `grails-test-examples-openapi` is inserted between `latency` and `namespaces`, breaking the alphabetical order of that list.
- `UrlMappingsOpenApiCustomizer.groovy:83`, `:353`: `java.util.regex.Pattern` and `Matcher` are fully qualified inline; `:386` does the same with `grails.core.GrailsControllerClass`, and `ActionAnnotations` qualifies `io.swagger.v3.oas.annotations.Operation` and friends throughout. Aliased imports (as the file already does for `Parameter as ParameterModel`) would read better.
- `PersistentEntitySchemaBuilder.entityNamed` scans every persistent entity per resolved schema. Fine at these sizes, but a name-keyed map built once in `build` is a one-liner.
- The two `MediaType` fallbacks (`:738`, `:804`) emit `"application/json": {}` when the entity is unknown. Valid, but an operation with no content at all reads better than a content entry with no schema. Preference only.

### Verified as correct

- `resources` really does generate `POST /books/{id}` for `update` (`DefaultUrlMappingEvaluator.java:1025`), so the extra `post` on `/books/{id}` is the mapping, not the customizer.
- The generated `META-INF/grails-plugin.xml` carries `dependsOn urlMappings='version'` literally; every core plugin's descriptor does the same and the runtime reads the instance property, so this is not new.
- springdoc 3.1.0 resolves against Spring Boot 4.1.1 and Framework 7.0.9 on the module classpath, and the functional app serves both the document and Swagger UI with it.
- `disambiguateOperationIds` runs after the `_byAction` suffix is applied at expansion time, so the path-qualified suffix is only reached for two declared mappings hitting the same action, which the `RestApiMappingSpec` covers.

jdaugherty:
@codeconsole if you're willing to accept the changes I pushed to this PR, I'm good to merge this.  It's a bit of a design change. I originally meant to push this to a separate PR, but accidentally pushed to this one.  Can you please take a look?

jdaugherty (commented):
I tried this against a production Grails API we document with OpenAPI today, and pushed the result as a branch on top of this PR's head (`11b3d985bf`): [`feat/openapi-build-time-generation`](https://github.com/apache/grails-core/tree/feat/openapi-build-time-generation). It is one commit, [`c5797b3d7b`](https://github.com/apache/grails-core/commit/c5797b3d7b), meant to be taken into this PR if you agree with the direction.

## What prompted it

Modelling that API's patterns in a spec against this head, the document came out wrong in ways that are hard for an application to work around:

- **The schema is chosen by controller name, not by the resource.** `OrderController extends RestfulController<OrderCommand>` is documented with the `Order` *domain class* when one exists - publishing the persistence model, internal columns included, instead of the DTO the controller renders. In our API 18 of 35 controllers share a name with a domain class.
- **Namespaced controllers collide.** With `v1` and `v2` controllers of the same name, `/api/v1/...` is described with the `v2` controller's annotations.
- **Much of what swagger-core annotations express is dropped.** Controller-level `@ApiResponse` and `@SecurityRequirement`, `@Parameter(in = QUERY)` for a parameter not already derived, `@RequestBody`, and non-JSON content (a PDF download is described as JSON of the entity).
- **Renamed command properties disappear.** Keeping only declared field names drops a property renamed with `@Schema(name = ...)` or `@JsonProperty`.
- **springdoc groups get nothing** (which answers @zyro23's question): springdoc only gives a `GroupedOpenApi` the customizers it carries plus `GlobalOpenApiCustomizer`s (`AbstractMultipleOpenApiResource` 3.1.0, line 129), and group criteria are only applied while scanning Spring MVC handler methods, never to paths a customizer adds.
- **Catch-alls and optional variables:** `"/**"(controller: 'index', action: 'notFound')` is documented as `GET /**`, and `"/photos/$id?"` only as `/photos/{id}`.
- **Runtime only.** We publish our description as a file generated at build time - packaged and served statically, diffed in review, handed to client generators - and nothing is served from introspection at runtime.

## How it works on the branch

```
          grails.openapi.* config  (+ springdoc.* filters, GroupedOpenApi beans)
                              │
   UrlMappings ─┐             ▼
   Controllers ─┼──►  GrailsOpenApiGenerator  ──►  OpenAPI model
   GORM ctx(s) ─┘     (no springdoc dependency)         │
                                          ┌──────────────┴───────────────┐
                                          ▼                              ▼
                             generate-open-api command         springdoc (optional)
                             build/openapi/openapi.yaml        /v3/api-docs
                             build/openapi/openapi-<g>.yaml    /v3/api-docs/<g>
```

`grails-openapi` alone gives the build-time command; adding springdoc also serves the same description at runtime. The derivation is this PR's, reorganised around a generator both front-ends share:

1. **Paths** from the URL mappings, as here, plus: a namespaced mapping resolves to that namespace's controller; an optional variable yields both paths; a wildcard that captures nothing is skipped; a mapping naming only a controller is its default action.
2. **Selection**: `@Hidden`; `grails.openapi.annotated-only` (only actions with `@Operation`, or controllers with `@Tag`); path and package include/exclude for the default document and for each group; `create`/`edit` left out unless `include-form-actions`.
3. **Operations**: a `RestfulController<T>` serves `T` - domain class or not - with the statuses from this PR; a controller that is not a `RestfulController` gets no guessed body. Simple-typed action arguments are query parameters. A `422` refers to a shared `ValidationErrors` schema matching what Grails renders; a `PATCH` body is a `<Name>Patch` copy that requires nothing.
4. **Annotations** are converted by swagger-core itself (`AnnotationsUtils`, `ParameterProcessor`), controller-level first then action-level, so they mean what they mean on any other endpoint.
5. **Types** resolve through swagger-core with a registered `ModelConverter` in place of the post-processing: it drops `metaClass` and `errors`, applies the constraints, marks the identifier, version and read-only command properties `readOnly`, and describes an association by its identifier - `{"id": ...}`, which is what Grails renders and binds for it - while an embedded one is described in full. Because it is a converter, a type named by an annotation gets the same treatment.
6. **Assembly** starts from an optional base document (info, servers, security, tag descriptions, `x-` extensions, hand-written paths and components) and adds what is derived.

The front-ends:

- **Build time**: `./gradlew generateOpenApi` (or `runCommand -Pargs="generate-open-api --format=json --output-directory=..."`) starts the context without a web server and writes the default document and one file per group. The command ships in a `grails-openapi-cli` companion, like the other framework commands, so it never reaches the runtime classpath.
- **Runtime**: a customizer contributes to springdoc's default document, and every group - configured under `grails.openapi.groups`, which the plugin registers as `GroupedOpenApi` beans, or declared by the application - is given its own customizer applying that group's criteria. The springdoc adapter is only loaded when springdoc is present.

| Setting | Purpose |
|---|---|
| `grails.openapi.enabled` | turn it off |
| `grails.openapi.base-document` | e.g. `classpath:openapi-base.yml` |
| `grails.openapi.annotated-only` | describe only what is annotated |
| `grails.openapi.include-form-actions` | describe `create`/`edit` |
| `grails.openapi.paths-to-match` / `-exclude`, `packages-to-scan` / `-exclude` | default document; the `springdoc.*` equivalents are honored too |
| `grails.openapi.groups.<name>.{display-name, paths-*, packages-*}` | groups |
| `grails.openapi.output-directory` / `output-format` | where and how the command writes |
| `springdoc.api-docs.version` | OpenAPI 3.1 (default) or 3.0, for both front-ends |

Groups are configured under `grails.openapi.groups` rather than `springdoc.group-configs` because the latter does not work in a Grails application at all: springdoc's `SpecPropertiesCondition` casts the `NavigableMap` Grails loads from `application.yml` to `GroupConfig` and startup fails, in the list form and the indexed `group-configs[0]` form alike. That is worth its own issue.

## Changes relative to this PR

- Public API: `UrlMappingsOpenApiCustomizer` is replaced by `grails.openapi.GrailsOpenApiGenerator`, `OpenApiSettings` and `OpenApiSelection`. The annotation reader, the RESTful action table, the converter and the springdoc adapter are internal, under `org.grails.openapi`.
- springdoc becomes an optional dependency of the module; the forge feature still selects the webmvc-ui starter.
- The BOM manages `swagger-core-jakarta`, `swagger-models-jakarta` and `swagger-annotations-jakarta` at the version springdoc 3.1.0 resolves (`swagger.version`).
- Behaviour: `create`/`edit` are not described by default; a non-`RestfulController` has no guessed body; associations are identifiers; `@author` tags are gone.
- The guide page is rewritten for both front-ends, including securing or disabling the runtime endpoints.

It also covers @matrei's review: the dead code and the misplaced Javadoc are gone with the rewrite, every mapping context is used, a controller-only mapping is described as its default action, the functional app now has the default mapping, and the guide says the runtime description is public.

## Tests

- `grails-openapi`: 126 unit specs, including the existing ones updated to the new API, and 4 command specs in `testCli`.
- `grails-test-examples-openapi`: 19 integration specs. Besides the existing checks, the app now has an association, a base document, a group and the default mapping, and generates its description at build time before the tests run. The specs check that the generated files describe the same paths, schemas and info springdoc serves for the default document and the group, that a book created through the documented reference renders its author as that reference, and that a failed save answers with the described `ValidationErrors` shape.

jdaugherty:
actually, it looks like i didn't push it! i'll open a separate PR

jdaugherty:
@codeconsole https://github.com/codeconsole/grails-core/pull/77 is the review with the new changes to merge into this one.

PR #16282, opened 2026-08-31
Keep collection properties dirty-checked after reassignment; track iterator-based removals

codeconsole:
### Mongo Collection Property Dirty Tracking

On interception-based stores (MongoDB), two everyday mutation patterns silently escape dirty checking — `save()` reports success and persists nothing:

```groovy
// 1. The defensive re-init — true for an EMPTY tracked list, because empty collections are falsy
if (!schedule.shares) {
    schedule.shares = []              // replaces the tracked wrapper with a plain ArrayList
}                                     // and [] == [] means the assignment isn't even flagged
schedule.shares.add(newShare)         // invisible: plain list, nothing marks the entity dirty
schedule.save(flush: true)            // writes nothing
```

```groovy
// 2. Groovy's closure-based removal — removes via iterator().remove()
schedule.shares.removeAll { it.userId == userId }   // DirtyCheckingCollection doesn't override iterator()
schedule.save(flush: true)                          // writes nothing
```

Hibernate is unaffected — its flush-time snapshot comparison catches everything. Mongo relies exclusively on the `DirtyChecking*` wrappers, so anything that escapes them is lost. Hit in production: a schedule-sharing feature showed "shared" while the document kept `shares: []`.

### The fix (interception only — no snapshots, no flush-time diffing)

**1. Wrappers track every mutation path.** `iterator()`/`listIterator()` now return dirty-marking iterators (covers `removeAll(Closure)`, `retainAll(Closure)`, `removeIf`), plus the missing direct overrides: `retainAll(Collection)`, `List.sort`, `List.replaceAll`. Same approach as Hibernate's `PersistentCollection`.

**2. Generated setters keep tracking across reassignment.** Collection/List/Set/Map-typed properties assign through `DirtyCheckingSupport.rewrap`:

```groovy
// generated setter, before:
void setShares(List shares) { markDirty("shares", shares); this.shares = shares }
// after:
void setShares(List shares) { markDirty("shares", shares); this.shares = (List) DirtyCheckingSupport.rewrap(this, "shares", this.shares, shares) }
```

`rewrap` wraps the new value **only when the value being replaced was itself a tracked wrapper** — otherwise it returns the raw value after one `instanceof`. Never-tracked properties (transient instances, Hibernate entities) behave exactly as before, and non-collection properties compile to identical bytecode.

**3. Replacement wrappers are flagged `isAssigned()`** (default method on `DirtyCheckableCollection`, so binary-compatible). `PersistentEntityCodec` then takes the full-rewrite path instead of per-element diffing — a replacement's layout need not match the stored array. Without the flag, a same-size replacement holding clean elements emitted no update at all.

### Tests

Each escape is reproduced by a spec that fails without the fix:

- `DirtyCheckingCollectionSpec` — 8 wrapper mutation paths that bypassed tracking
- `DirtyCheckCollectionReassignmentSpec` — reassignment loses tracking (List/Set/Map); never-tracked values stay untouched
- `EmbeddedCollectionDirtyTrackingSpec` — end-to-end against MongoDB, replicating the production shape (an auto-timestamped entity: the `lastUpdated` write during flush resets the explicit-save dirty marker, so persistence depends entirely on the wrappers)


grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy
```diff
@@ -130,4 +130,53 @@ class DirtyCheckingSupport {
...
+     * own snapshot-based dirty checking and never installs these wrappers — the new value is
+     * returned untouched, keeping this a no-op for those cases.
+     *
+     * @param parent The dirty-checkable owner
+     * @param property The property being assigned
+     * @param oldValue The value being replaced
+     * @param newValue The value being assigned
+     * @return The value to store: {@code newValue}, wrapped if it replaces a tracked value
+     */
+    static Object rewrap(DirtyCheckable parent, String property, Object oldValue, Object newValue) {
```
jdaugherty, line 167: `PersistentCollection` and the Neo4j store's own wrappers (`Neo4jList`/`Neo4jSet`/`Neo4jSortedSet`, `Neo4jPersistentList`/`Neo4jPersistentSet`) all implement `DirtyCheckableCollection`, so this also fires when one of *those* is the value being replaced, and it installs a plain `DirtyChecking*` wrapper in their place. `Neo4jEntityPersister.createDirtyCheckableAwareCollection` then takes its "already dirty-checkable" branch and never re-wraps into a Neo4j collection. Before this change the raw replacement went through the other branch and came back as a `Neo4jList`/`Neo4jSet` on the first save.

Net effect: once a `hasMany` property has been reassigned, in-place removals on that instance stop deleting relationships for the rest of its life (inserts still happen because `hasChanged()` is true, and they are MERGEs, so adds look fine). Orphan removal in `GraphAdapter.adaptGraphUponRemove` is lost the same way.

Repro against the embedded harness, unidirectional `List songs` / `hasMany = [songs: Song]`:

```groovy
p = Playlist.get(id); p.trackChanges()      // songs: Neo4jPersistentList
p.songs = [a, b, c]; p.save(flush: true)    // songs: DirtyCheckingList, and still is after save
p.songs.remove(a); p.save(flush: true)      // hasChanged('songs') == true, but no RelationshipPendingDelete
```

| | relationships after reassign+save | after remove+save | reload |
|---|---|---|---|
| this branch | 3 | **3** | `[a, b, c]` |
| merge-base | 3 | 2 | `[b, c]` |

Two ways out, either works for me: (a) have the Neo4j persister treat an `isAssigned()` wrapper like a raw collection, which is what the flag means, by wrapping it (or its `target`) in `Neo4jList`/`Neo4jSet` and registering the inserts; or (b) only rewrap when the old value is one of the generic `DirtyChecking*` classes and leave store-specific collections to the store. Either way this needs a Neo4j spec covering reassign, save, remove, save.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingMap.groovy
```diff
@@ -34,12 +34,23 @@ class DirtyCheckingMap implements Map, DirtyCheckableCollection {
     final DirtyCheckable parent
     final String property
     final int originalSize
+    final boolean assigned
```
jdaugherty, line 40: The Map wrapper still has the same class of escape this PR closes for collections. Checked the compiled class: `@Delegate` generates `putIfAbsent`, `merge`, `compute`/`computeIfAbsent`/`computeIfPresent`, `replaceAll`, `replace` and `remove(key, value)` as straight calls on `target` with no `markDirty`, and `entrySet()`/`keySet()`/`values()` hand out the raw views. So Groovy's `Map.removeAll(Closure)` / `retainAll(Closure)` (they iterate `entrySet()`), `keySet().remove(k)` and `values().removeIf { }` are all invisible to tracking, exactly like `List.removeAll(Closure)` was. `DirtyCheckingList.subList()` is a raw view too.

Since the description says the wrappers now track every mutation path, either cover these (override the default methods, return tracking views for the three collection views) or call them out as known gaps in the docs update.

grails-datamapping-core/src/main/groovy/org/grails/compiler/gorm/DirtyCheckingTransformer.groovy
```diff
@@ -94,6 +97,13 @@ class DirtyCheckingTransformer implements CompilationUnitAware {
     public static final ClassNode DIRTY_CHECKED_PROPERTY_CLASS_NODE = ClassHelper.make(DirtyCheckedProperty)
     public static final ClassNode DIRTY_CHECK_CLASS_NODE = ClassHelper.make(DirtyCheck)
     public static final AnnotationNode DIRTY_CHECKED_PROPERTY_ANNOTATION_NODE = new AnnotationNode(DIRTY_CHECKED_PROPERTY_CLASS_NODE)
+    private static final ClassNode DIRTY_CHECKING_SUPPORT_CLASS_NODE = ClassHelper.make(DirtyCheckingSupport)
+    // Interface-typed collection properties whose generated setter re-establishes change
+    // tracking via DirtyCheckingSupport.rewrap. Restricted to the exact interfaces the
+    // DirtyChecking* wrappers implement so the cast in the generated setter is always valid.
+    private static final Set<String> REWRAPPABLE_TYPE_NAMES = [
```
jdaugherty, line 104: `SortedSet` is a supported GORM collection type and `DirtyCheckingSortedSet` exists for it, but it isn't in this set, so a `SortedSet`-typed property reassigned over a tracked value loses tracking exactly the way `List`/`Set` did. Adding `SortedSet.name` here plus a `newValue instanceof SortedSet` branch in `rewrap` ahead of the `Set` check keeps the generated cast valid and stops a `TreeSet` assigned to a `Set` property from losing its `SortedSet` API behind the wrapper. `wrap()` has the same blind spot, so worth fixing both together.

grails-datamapping-core/src/main/groovy/org/grails/compiler/gorm/DirtyCheckingTransformer.groovy
```diff
@@ -364,7 +374,21 @@ class DirtyCheckingTransformer implements CompilationUnitAware {
...
             setterBody.addStatement(stmt(markDirtyMethodCall))
-            setterBody.addStatement(assignS(propX(varX('this'), fieldName), varX(setterParameter)))
+            // Collection/Map-typed properties assign through DirtyCheckingSupport.rewrap so a
+            // value that replaces a tracked wrapper (installed by an interception-based store's
+            // decoder) stays tracked. Without this, `entity.items = []` over a tracked list
+            // stored a plain untracked collection and every later in-place mutation was
+            // invisible to change tracking. rewrap is a no-op when the old value was untracked,
+            // so stores with their own dirty checking (Hibernate) are unaffected.
+            Expression assignedValue
+            if (REWRAPPABLE_TYPE_NAMES.contains(returnType.name)) {
```
jdaugherty, line 384: Only generated setters get the rewrap. A domain class with its own `void setShares(List shares)` goes through `weaveIntoExistingSetter`, which only prepends `markDirty`, so the original bug is still there for hand-written setters. Fine to leave for a follow-up, but it should be stated in the docs update so nobody assumes it's covered.

jdaugherty (dismissed):
Ran the two unit specs locally (both green) and read through the wrapper, transformer and codec changes. The approach is sound and the Mongo fix is real. Three things before this goes in:

1. **Neo4j regression** (inline on `rewrap`): reassigning a `hasMany` now permanently replaces the Neo4j-aware collection with a generic wrapper, so later in-place removals stop deleting relationships. Reproduced on this branch against the embedded harness; the same spec passes on the merge-base.
2. **Map wrapper still leaks** (inline on `DirtyCheckingMap`): the `@Delegate`-generated default methods and the three views bypass tracking.
3. **Docs**: `grails-data-mongodb/docs/src/docs/asciidoc/objectMapping/dirtyChecking.adoc` documents the exact caveat this PR removes ("if you override the collection with a non-dirty checking aware type it can disable dirty checking and prevent the property from being updated"). That paragraph needs to describe the new behaviour, plus whatever gaps remain after the above (Map views, `subList`, hand-written setters).

codeconsole:
All four addressed, in `8581d5604f` (branch also synced with `8.0.x`):

**1. Neo4j regression** — took option (b): `rewrap` now gates on the replaced value being one of the plugin's own generic wrapper classes by **exact class**, not `instanceof` (`Neo4jList extends DirtyCheckingList`, so `instanceof` was the trap). A store-specific wrapper or `PersistentCollection` being replaced stores the raw value, and the store's persister re-wraps it on save exactly as at merge-base. Added `HasManyReassignDirtyCheckingSpec` in grails-data-neo4j (embedded harness, your exact reassign → save → remove → save → reload shape) — verified it fails with the previous `instanceof` gate and passes with the exact-class one. There's also a unit spec pinning that a `DirtyCheckingList` *subclass* as the old value is left alone.

**2. Map wrapper** — covered rather than documented away: overrides for the `@Delegate`-generated default methods (`putIfAbsent`, `merge`, `compute*`, `replace`, `replace(k,old,new)`, `replaceAll`, `remove(k,v)`), and `entrySet()`/`keySet()`/`values()` now return tracking wrappers, which catches Groovy's `Map.removeAll(Closure)`/`retainAll(Closure)` (entrySet iteration), `keySet().remove(k)` and `values().removeIf { }`. `DirtyCheckingList.subList()` returns a tracking view too. The one Map path left untracked is `Map.Entry.setValue` during iteration — called out in the docs.

**3. SortedSet** — added to the transformer's rewrappable set, and both `wrap()` and `rewrap()` now produce `DirtyCheckingSortedSet` ahead of the `Set` check, so a `TreeSet` keeps its `SortedSet` API behind the wrapper (including on `Set`-declared properties via `wrap()`).

**4. Docs** — rewrote the caveat paragraph in `objectMapping/dirtyChecking.adoc`: reassignment now stays tracked, with the remaining gaps listed explicitly — hand-written setters (weave-only, unchanged in this PR), `Map.Entry.setValue`, and store-specific wrappers deliberately left to their store's persister.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy
codeconsole: Took option (b), with one adjustment your repro made necessary: `Neo4jList extends DirtyCheckingList`, so an `instanceof` check against the generic classes would still have caught the Neo4j wrappers. `rewrap` now gates on the **exact class** of the replaced value being one of the five generic wrappers (`isGenericWrapper`). A store-specific wrapper or `PersistentCollection` being replaced stores the raw value, and the store's persister re-wraps it on save exactly as at merge-base.

Added `HasManyReassignDirtyCheckingSpec` in grails-data-neo4j (embedded harness) running your exact reassign → save → remove → save → reload sequence — verified it fails with the previous `instanceof` gate and passes with the exact-class one. There's also a unit spec pinning that a `DirtyCheckingList` subclass as the old value is left alone. In `8581d5604f`.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingMap.groovy
codeconsole: Covered rather than documented away, in `8581d5604f`: overrides for the `@Delegate`-generated default methods (`putIfAbsent`, `merge`, `compute`/`computeIfAbsent`/`computeIfPresent`, `replace`, `replace(k,old,new)`, `replaceAll`, `remove(k,v)`), and `entrySet()`/`keySet()`/`values()` now return tracking wrappers — which is what catches Groovy's `Map.removeAll(Closure)`/`retainAll(Closure)` (they iterate `entrySet()`), `keySet().remove(k)` and `values().removeIf { }`. `DirtyCheckingList.subList()` returns a tracking view as well, with a spec per path.

The one Map path left untracked is `Map.Entry.setValue` during iteration; that's now called out explicitly in the docs update.

grails-datamapping-core/src/main/groovy/org/grails/compiler/gorm/DirtyCheckingTransformer.groovy
codeconsole: Done in `8581d5604f`: `SortedSet.name` added here, and both `wrap()` and `rewrap()` now produce `DirtyCheckingSortedSet` ahead of the `Set` check — so a `TreeSet` keeps its `SortedSet` API behind the wrapper on `Set`-declared properties too (the `wrap()` blind spot you pointed out). Specs cover the SortedSet-typed reassignment and `wrap(TreeSet)` returning the SortedSet wrapper.

grails-datamapping-core/src/main/groovy/org/grails/compiler/gorm/DirtyCheckingTransformer.groovy
codeconsole: Left for a follow-up as you suggested, and now stated explicitly in the docs update: the rewritten caveat in `objectMapping/dirtyChecking.adoc` lists hand-written setters first among the paths that still require an explicit `markDirty(propertyName)` — "a hand-written setter … stores the value it is given without re-wrapping, so a collection assigned through it loses tracking until the next load."

codeconsole:
The Mongo CI failures (BasicCollectionsSpec Currency/BigDecimal maps, MapOfDomainsSpec) were caused by the new Map views: Groovy's `Map == Map` extension compares `self.keySet().equals(other.keySet())`, and the tracking view fell back to identity equality. Fixed in `3ba7fb58d2` by delegating `equals`/`hashCode` on the wrappers to their target — the contract `AbstractPersistentCollection` already implements — with a spec pinning content equality both ways plus the three views.

sbglasius:
One edge in `DirtyCheckingSupport.rewrap` worth noting. It is not a regression from this PR, so feel free to defer it.

`rewrap` returns early for a value that is already a `DirtyCheckableCollection`:

```groovy
if (newValue instanceof DirtyCheckableCollection) {
    return newValue
}
```

So assigning one entity's tracked collection onto another entity's property stores the *source* wrapper, whose `parent`/`property` still point at the source entity:

```groovy
a.shares = b.shares    // a.shares is now b's DirtyCheckingList
a.shares.add(share)    // marks b dirty for 'shares' — a stays clean
```

The wrong-parent binding pre-dates this PR (the old generated setter stored that wrapper verbatim), so nothing regresses here. What is new is the `assigned` consequence: a value taking this early return keeps `assigned = false`, so `PersistentEntityCodec.encodeEmbeddedCollectionUpdate` takes the per-element update path for what is in fact a wholesale replacement — while the same assignment written as `a.shares = b.shares.toList()` correctly gets `assigned = true` and falls through to the full re-encode.

Re-wrapping in that branch when the existing wrapper's `parent`/`property` do not match the target (instead of returning it as-is) would cover both.

codeconsole:
Thanks — taken rather than deferred, in `f11e85549b`.

`rewrap` now re-binds a borrowed wrapper: it re-wraps that wrapper's **raw target** against the assigning entity with `assigned = true`. Three constraints shaped it:

- **Store-specific wrappers stay untouched.** Your wording taken literally (re-wrap on parent/property mismatch) would convert a `Neo4jList` belonging to another entity into a generic wrapper — the regression @jdaugherty caught earlier, since `Neo4jList extends DirtyCheckingList`. The exact-class check now guards the incoming value as well as the replaced one.
- **Unwrap, never nest.** Wrapping the foreign wrapper itself would leave its `parent.markDirty` firing underneath, marking *both* entities on every mutation. It unwraps in a loop, because `BasicCollectionTypeEncoder:78` builds a `DirtyCheckingMap` with no already-wrapped guard, so a wrapper can arrive nested inside another.
- **Owner compared by identity**, not `equals` — two rows equal by business key are not the same owner, and on an association `equals` can initialise a proxy.

Two things worth flagging back:

**Your repro passes as-is today.** The codec condition is a three-way AND, and the `add` grows the list, so `hasChangedSize()` is true and the full re-encode runs anyway. I wrote that spec first and it was green pre-fix, so I replaced it. The shape that does fail is the one where the two collections are **equal in content**: the assignment is then equality-suppressed, so A is never flagged, the update carries only `lastUpdated`, and the addition is silently lost while B absorbs it. That is the new Mongo spec, verified red before the change and green after.

**It also fires when the property was untracked.** `new Entity(shares: other.shares)` had the same defect and could not self-heal — the insert's write-back assigns the field directly and never reaches `rewrap` — so the re-binding no longer depends on what the property held before. Assigning a plain collection to an untracked property still introduces no tracking.

Not addressed here, deliberately: `wrap()` keeps the same owner-blind early return, so a foreign wrapper reaching it through another path is still handed back as-is. It is a public method that has returned the same instance since 2.0 and is on the encode path, so it felt like its own change rather than a rider on this one — happy to do it here if you would rather it were one PR.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingMap.groovy
codeconsole: Following up on this thread, since it is where the "tracks every mutation path" claim was litigated: reviewing the compiled classes for the latest round turned up two more `@Delegate`-generated escapes of exactly this kind, now closed in `a2ac5178cf`.

- The Java 21 `SequencedCollection` mutators — `addFirst`, `addLast`, `removeFirst`, `removeLast` — were generated straight through to the target on `DirtyCheckingList` and `DirtyCheckingSortedSet`, with no `markDirty`.
- `DirtyCheckingSortedSet` handed out `headSet`, `tailSet` and `subSet` as raw live views, the same hole `subList` was overridden for.

Those views (plus `reversed`) now come back as tracking wrappers, and are constructed as assigned, because a view stored back onto a property is a wholesale replacement rather than something to diff element by element. The docs sentence is narrowed to `List` and `SortedSet` properties, since a plain `Set`/`Map` wrapper does not expose the sequenced API at all.

`Map.Entry.setValue` during iteration is still the one Map path left untracked, and remains listed as a gap.

matrei:
## AI Review Findings

The three findings from the first round and the borrowed-wrapper edge from the second are all resolved, and the fixes hold up: I ran `DirtyCheckingCollectionSpec`, `DirtyCheckCollectionReassignmentSpec`, `EmbeddedCollectionDirtyTrackingSpec` (Mongo) and `HasManyReassignDirtyCheckingSpec` (Neo4j) locally, all green, and checked the compiled wrappers with `javap` to confirm no `@Delegate`-generated mutator is left calling straight through on any of the five classes.

One finding below I would like either fixed or stated in the docs before this goes in, because it is the PR's own bug on a shape one mapping away from the one in the description. The rest are non-blocking.

## Resolved Findings

- Neo4j regression: `rewrap` gates on the exact generic wrapper classes, so `Neo4jList`/`Neo4jSet` and the `PersistentCollection` types are left for the store to re-wrap. `HasManyReassignDirtyCheckingSpec` runs the reassign, save, remove, save, reload sequence and passes.
- Map wrapper: the `@Delegate`-generated default methods are overridden and `entrySet()`/`keySet()`/`values()` come back as tracking views. `javap` on `DirtyCheckingMap` shows only `forEach` still delegated, which is read-only. The same check on `DirtyCheckingList`/`DirtyCheckingSortedSet` shows the `SequencedCollection` mutators and range views covered as of `a2ac5178cf`.
- `SortedSet` is in `REWRAPPABLE_TYPE_NAMES`, and both `wrap()` and `rewrap()` produce `DirtyCheckingSortedSet` ahead of the `Set` check.
- Hand-written setters are called out in the docs as untracked.
- Borrowed wrappers are re-bound to the assigning entity by wrapping the raw target, with owner compared by identity and the store-specific types excluded on the incoming side too. The Mongo spec uses the equal-content shape, which is the one that actually fails without the fix.
- The `Map == Map` CI failures are fixed by content `equals`/`hashCode` on the wrappers, matching `AbstractPersistentCollection`.

## Findings

### Medium: A loaded `hasMany` on Mongo still loses the falsy empty re-init

References:

- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy:192` (the `isGenericWrapper(oldValue)` gate)
- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy:268`
- `grails-data-mongodb/core/src/main/groovy/org/grails/datastore/mapping/mongo/engine/codecs/PersistentEntityCodec.groovy:551-565` (to-many decode)
- `grails-data-mongodb/docs/src/docs/asciidoc/objectMapping/dirtyChecking.adoc:39`

The description's first escape is reproduced with `shares` as an embedded collection. Declare the same property as a unidirectional `hasMany` instead and the exact same code still loses the add on this branch. Mongo decodes a stored `members: []` into an empty `PersistentList`, whose class is not one of the five generic wrappers, so `rewrap` stores the raw `[]`; `markDirty` suppresses the assignment because `ArrayList.equals(emptyPersistentList)` is true; the `add` then goes to a plain list nobody tracks.

Probe against the Mongo harness, `List<ProbeMember> members; static hasMany = [members: ProbeMember]`, entity saved with `members: []`:

```groovy
board = ProbeBoard.get(id); board.trackChanges()   // members: PersistentList (empty)
if (!board.members) board.members = []             // members: java.util.ArrayList
board.members.add(new ProbeMember(userId: 'u1'))
board.save(flush: true)
```

| | class held after re-init | document after save | reload |
|---|---|---|---|
| this branch | `java.util.ArrayList` | `members: []`, `version: 1` | `0` members |

This is not a regression: the generated setter stored the raw list before this PR as well. But it is the bug in the title, on the mapping most applications would reach for before `embedded`, and the `instanceof DirtyCheckableCollection` gate from the first push did cover it; the exact-class gate that fixed Neo4j dropped it. The docs sentence at line 39 reads as if reassignment is covered generally, and the one-to-many caveat at line 53 only appears inside the borrowed-collection paragraph.

Two ways out:

1. Accept the three datastore-core classes `PersistentList`, `PersistentSet` and `PersistentSortedSet` as exact classes in the old-value gate alongside the generic wrappers. `Neo4jPersistentList`/`Neo4jPersistentSet`/`Neo4jPersistentSortedSet` are subclasses, so the Neo4j path is unchanged. The resulting state, a `hasMany` property holding a generic `DirtyCheckingList`, is already what `MongoCodecEntityPersister` writes back after any save, and the probe's second case (reassign after an in-session save, then add) round-trips through exactly that state today. It would need a Mongo spec on the `hasMany` shape next to the embedded one.
2. Leave the behaviour and add a bullet to the gap list at line 56: a one-to-many or many-to-many loaded from the database is a `PersistentCollection`, is not re-wrapped on reassignment, and needs `markDirty` after an in-place mutation that follows a reassignment.

I would take the first, since it is the production shape with a different mapping keyword.

### Low: Wrapper `equals` is not reflexive when the target has identity equality

Reference:

- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingCollection.groovy:82`

`target.equals(other)` is fine when the target defines content equality (`ArrayList`, `HashSet`, `TreeSet`, every `Map`), but the new `values()` view and any `DirtyCheckingCollection` produced by `wrap()` for a non-List, non-Set target wrap an `AbstractCollection`, which inherits `Object.equals`. The wrapper then fails to equal itself. Checked against the compiled classes:

```text
map.values().equals(map.values())      = false
[map.values()].contains(map.values())  = false
new DirtyCheckingCollection(new ArrayDeque(['x']), o, 'p').equals(self) = false
```

Groovy `==` still returns true because it checks identity first, so nothing in the specs sees it, but `List.contains`, `indexOf` and `HashSet` membership go through `equals`. Short-circuiting on identity keeps the delegation and restores the contract:

```groovy
boolean equals(Object other) {
    other.is(this) || target.equals(other)
}
```

### Low: `BasicCollectionTypeEncoder` nests a new `DirtyCheckingMap` on every save (follow-up)

References:

- `grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/BasicCollectionTypeEncoder.groovy:76-78`
- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy:255`

The write-back for a map property wraps whatever the property currently holds, with no already-wrapped guard, so an entity saved N times in a session holds a map N wrappers deep. The unwrap loop in `genericWrapperTarget` works around this for `rewrap`, and the comment there names the cause. Pre-existing, so fine to defer, but worth a follow-up now that each level also hands out a fresh view wrapper per `entrySet()` call, since the cost of the nesting is no longer just an extra `markDirty`. The collection branch a few lines up already goes through `DirtyCheckingSupport.wrap`, which has the guard; the map branch only needs the same `instanceof` check.

### Nit: The explicit `iterator()` overrides in the three subclasses are redundant

References:

- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingList.groovy:83`
- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSet.groovy:45`
- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSortedSet.groovy:45`

`@Delegate` skips methods the owner already has, including inherited ones. `javap` on `DirtyCheckingSet` shows no `removeIf`, `retainAll` or `removeAll` of its own, so the ones inherited from `DirtyCheckingCollection` are what run, and the same holds for `iterator()`. The three `super.iterator()` overrides and their "route through" comments can go. Harmless if kept.

### Nit: The docs list `addFirst`/`addLast` for `SortedSet` properties

Reference:

- `grails-data-mongodb/docs/src/docs/asciidoc/objectMapping/dirtyChecking.adoc:43-44`

`SortedSet.addFirst`/`addLast` throw `UnsupportedOperationException` by contract, which is why `DirtyCheckingSortedSet` does not override them (its own comment at line 50 says so). The sentence should attach those two to `List` properties only; the removals and views apply to both.

codeconsole:
Thanks — all five taken, one commit each.

**Medium, `hasMany` re-init (`0ae70cef89`).** Took option 1. Reproducing it first was worth it: my initial probe passed, because saving the board without ever setting `members` stores no field at all and it decodes as `null`, not an empty `PersistentList`. With `members: []` actually stored it reproduces exactly as you described — decoded `PersistentList` → `java.util.ArrayList` after the re-init, assignment equality-suppressed, add untracked, `0` members on reload.

`PersistentList`/`PersistentSet`/`PersistentSortedSet` now count as tracked in the old-value gate, by exact class, so the Neo4j subclasses still fail it. The gate reads only `getClass()`, never the contents, so an uninitialised lazy association is not forced. `HasManyReassignDirtyTrackingSpec` covers it next to the embedded one — verified red without the change. The docs now say the reassignment behaviour covers associations too, rather than leaving the caveat to the borrowed-collection paragraph.

**Low, reflexive `equals` (`f799282454`).** Identity short-circuit before delegating, exactly as you wrote it. Spec pins `values()` and a `DirtyCheckingCollection` over an `ArrayDeque` — both fail without it, and `[wrapper].contains(wrapper)` was the assertion that actually caught it, since Groovy `==` does not.

**Low, encoder nesting (`a7f7edb0d6`).** Fixed rather than deferred — it is one missing overload. `DirtyCheckingSupport.wrap` now has a `Map` form with the same already-wrapped guard as the `Collection` form, and the map branch routes through it. Kept as its own commit so it can be dropped if you would rather it were a separate PR. The unwrap loop in `genericWrapperTarget` stays, since it is what makes `rewrap` correct against any encoder.

**Nit, redundant `iterator()` overrides (`2dade74084`).** Removed on all three subclasses. Confirmed the way you described: the iterator-removal and `removeAll(Closure)` specs on `List`, `Set` and `SortedSet` still pass without them, so the inherited one was doing the work.

**Nit, docs `addFirst`/`addLast`.** Attached to `List` properties only; the removals and views stay on both.

matrei:
## AI Review Findings (round 2)

Head `2dade74084`, base `8.0.x` (PR base `0980623481`; `origin/8.0.x` has moved on by 33 commits since, none touching the PR's files, and a trial merge is clean).

All five items from the previous round are taken, one commit each, and each fix does what the author's comment says. What I ran on the head commit:

| Suite | Result |
|---|---|
| `grails-datastore-core` `dirty.checking.*` (29 tests) | green |
| `grails-datamapping-core` `DirtyCheckCollectionReassignmentSpec` (17) | green |
| Mongo, Docker: `HasManyReassignDirtyTrackingSpec`, `EmbeddedCollectionDirtyTrackingSpec`, `OneToManySpec`, `ManyToManySpec`, `EmbeddedCollectionSpec` (8 classes, 18 tests) | green |
| Neo4j, Docker: `HasManyReassignDirtyCheckingSpec` | green |
| Simple-map TCK in `grails-datamapping-core-test`: `OneToManySpec`, `CircularOneToManySpec`, `CircularManyToManySpec`, `ManyToManySpec`, `InheritanceWithOneToManySpec`, `CircularCascadeSpec` (13 tests) | green |
| `checkstyleMain` + `codenarcMain` on `grails-datastore-core` | clean |
| `checkstyleMain` + `codenarcMain` on `grails-data-mongodb-bson` | **1 CodeNarc violation** |

One blocking item, and it is the one CI already flags: the "Code Style / Core Projects" job fails on this head. The rest is non-blocking.

## Resolved Findings

- `hasMany` re-init on Mongo (`0ae70cef89`): `wasTracked` accepts `PersistentList`/`PersistentSet`/`PersistentSortedSet` by exact class, so the Neo4j and Rx subclasses still fall through to their stores. `HasManyReassignDirtyTrackingSpec` reproduces the shape from the last round with `members: []` actually stored, and passes. The gate reads only `getClass()`, so an uninitialised lazy association is not forced.
- Reflexive `equals` (`f799282454`): identity short-circuit in both wrappers, pinned by `[wrapper].contains(wrapper)` on a `values()` view and on a `DirtyCheckingCollection` over an `ArrayDeque`.
- Map nesting on save (`a7f7edb0d6`): `DirtyCheckingSupport.wrap(Map)` carries the same already-wrapped guard as the collection form, and the encoder's map branch goes through it. The other two `wrap` call sites pass a `List` or a `Collection`, one of them under `@CompileDynamic`, so the new overload is unambiguous at every caller.
- Redundant `iterator()` overrides (`2dade74084`): removed on all three subclasses; the iterator-removal and `removeAll(Closure)` cases in `DirtyCheckingCollectionSpec` still pass, so the inherited override is the one doing the work.
- Docs: `addFirst`/`addLast` are attached to `List` properties only.

## Findings

### Blocking: Unused import fails the code-style job

Reference:

- `grails-data-mongodb/bson/src/main/groovy/org/grails/datastore/bson/codecs/encoders/BasicCollectionTypeEncoder.groovy:32`

`a7f7edb0d6` replaced the only use of `DirtyCheckingMap` in this file with `DirtyCheckingSupport.wrap`, but left the import behind. CodeNarc's `UnusedImport` rule rejects it:

```text
BasicCollectionTypeEncoder.groovy  UnusedImport  line 32
The [org.grails.datastore.mapping.dirty.checking.DirtyCheckingMap] import is never referenced
```

That is the "Code Style / Core Projects" failure TestLens reports on `2dade74084`. Reproduces locally with:

```bash
./gradlew :grails-data-mongodb-bson:codenarcMain
```

Dropping the import line is the whole fix.

### Low: The exact-class rule for `PersistentList` has no unit-level pin

References:

- `grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/dirty/checking/DirtyCheckingSupport.groovy:292-305`
- `grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/dirty/checking/DirtyCheckCollectionReassignmentSpec.groovy:188`

The new gate is covered only by the Mongo container spec, and only for the `List` shape. The reassignment spec already has the mirror case for the wrapper side, a `StoreSpecificList` subclass of `DirtyCheckingList` that must not be rewrapped. The same pair for the association side would pin all three exact classes and the subclass exclusion without a container, and would catch a future refactor that relaxes the check to `instanceof PersistentCollection` (which is exactly the change that regressed Neo4j in the first push):

```groovy
def 'a plain collection replacing a datastore PersistentList is rewrapped'() {
    given:
    def entity = entityClass.newInstance()
    entity.shares = new PersistentList([], String, Mock(Session))
    entity.trackChanges()

    when:
    entity.shares = []
    entity.shares.add('a')

    then:
    entity.shares instanceof DirtyCheckingList
    ((DirtyCheckableCollection) entity.shares).isAssigned()
    entity.hasChanged('shares')
}

def 'a store-specific PersistentList subclass is left to its store'() {
    given:
    def entity = entityClass.newInstance()
    entity.shares = new StoreSpecificPersistentList([], String, Mock(Session))   // extends PersistentList
    entity.trackChanges()

    when:
    entity.shares = ['a']

    then:
    !(entity.shares instanceof DirtyCheckableCollection)
}
```

The setter's `markDirty` equality check does initialise the `PersistentList`, but `loadInverseChildKeys` returns before consulting the session when the key set is empty, so a Spock mock is enough for these shapes. Same two cases for `PersistentSet` and `PersistentSortedSet` if you want the full matrix; the `where:` block form keeps it short.

### Nit: The borrowed-collection paragraph now reads as contradicting the one above it

Reference:

- `grails-data-mongodb/docs/src/docs/asciidoc/objectMapping/dirtyChecking.adoc:54-55`

Line 41 now says the reassignment behaviour "holds for ... one-to-many and many-to-many associations alike". Thirteen lines later the borrowed-collection paragraph says "A one-to-many or many-to-many is held in a `PersistentCollection` instead and is left to the store". Both are true, but only because the second sentence is scoped to the borrowing case, and the reader has to work that out. Leading the sentence with the scope fixes it: "When *borrowed* from another entity, a one-to-many or many-to-many is held in a `PersistentCollection` and is stored as-is, ...".

## Verified as correct

- **Only the generated setter reaches `rewrap`.** `FieldEntityAccess.setProperty` writes through `FieldWriter.write`, which is `field.set`, so the key-value persister's post-persist `entityAccess.setProperty(name, associatedObjects)` in `NativeEntryEntityPersister` never runs the setter, and the newly accepted `PersistentList` old value has no effect on that path. The simple-map TCK one-to-many and many-to-many specs confirm it.
- **Mongo update path for a rewrapped `hasMany`.** `DirtyCheckingCollection.hasChanged()` is `parent.hasChanged(property) || hasChangedElements()`. After the re-init the `add` marks the property, `MongoCodecEntityPersister` persists the children and records the ids, and `OneToManyEncoder` writes them. A same-size replacement with different members is flagged by the setter's `markDirty` equality check. An equal-content replacement writes nothing, which is the correct outcome.
- **Borrowed `PersistentList`.** `boardA.members = boardB.members` still stores the `PersistentCollection` verbatim, since `genericWrapperTarget` returns null for it. That is the pre-PR behaviour and what the docs paragraph describes.
- **`wrap(Map)` and `@CompileStatic`.** In `BasicCollectionTypeEncoder` the map branch is guarded by `value instanceof Map`, so static overload resolution picks the `Map` form; the `Collection` branch picks the collection form. No call site passes an untyped or possibly-null value.
- **Style on the changed datastore-core files.** Checkstyle and CodeNarc are clean on `grails-datastore-core`.

codeconsole:
All three taken.

**Blocking, unused import (`534956d3f5`).** Dropped. Confirmed from the job log that CodeNarc in `grails-data-mongodb/bson` was the only style failure, and `:grails-data-mongodb-bson:codenarcMain` + `checkstyleMain` are clean locally, as are the other two changed modules.

**Low, unit pin for the exact-class rule (`bc0b7950e6`).** Added, as the `where:` form over all three classes plus the subclass exclusion. `Mock(Session)` is enough exactly as you said — the empty key set short-circuits before the session is consulted.

Verified both directions rather than just that they pass: the three parameterised cases go red if the gate stops accepting `PersistentCollection`, and the subclass case goes red if the exact-class check is relaxed to `instanceof PersistentCollection` — which is the change that regressed Neo4j in the first push, so it now has a unit-level guard that does not need a container.

**Nit, contradicting paragraphs (`bc0b7950e6`).** Led the sentence with the scope, and added the other half explicitly, since the asymmetry is the part worth stating: re-initialising a to-many on its own entity is tracked; only one borrowed from another entity is stored as-is.

matrei (approved):
## AI Review Findings (round 3)

Head `bc0b7950e6`, base `8.0.x` (PR base `0980623481`; `origin/8.0.x` is 33 commits ahead, none touching the PR's files, and a trial merge is clean).

Two commits since the last round, one per open item, and both do what they say. Nothing blocking. What I ran on the head commit:

| Check | Result |
|---|---|
| `grails-datamapping-core` `DirtyCheckCollectionReassignmentSpec` (21 tests, 4 new) | green |
| Mutation: `isPersistentCollection` relaxed to `instanceof` | the new subclass case fails, the other 20 pass |
| Mutation: `wasTracked` stops accepting `PersistentCollection` | exactly the three new parameterised cases fail |
| `checkstyleMain` + `codenarcMain` on `grails-datastore-core` and `grails-data-mongodb-bson` | clean (Checkstyle 0 violations on 207 and 7 files) |
| GitHub check runs on `bc0b7950e6` | 91 runs, all success or skipped; both "Core Projects" jobs green, Mongo 7.0/8.0 and Neo4j functional jobs green |

The source was restored after each mutation; the working tree is unchanged.

## Resolved Findings

- Unused `DirtyCheckingMap` import (`534956d3f5`): removed. CodeNarc is clean on `grails-data-mongodb-bson` and the "Core Projects" job that failed on `2dade74084` passes on this head.
- Unit-level pin for the exact-class rule (`bc0b7950e6`): the parameterised case covers `PersistentList`, `PersistentSet` and `PersistentSortedSet` by exact class, with the replacement mutated in place and `isAssigned()` asserted, and the `StoreSpecificPersistentList` case pins the subclass exclusion. The commit message's claim that each mutation fails the intended cases holds; see the table above. A Spock mock session is enough: the setter's equality check initialises the empty `PersistentList`, and `loadInverseChildKeys` returns before consulting the session when the key set is empty.
- Docs contradiction (`bc0b7950e6`): the borrowed-collection sentence now leads with its scope and ends by pointing back at the re-init paragraph, so the two statements no longer read as contradicting each other.

## Findings

### Nit: New imports in the spec are out of group order

Reference:

- `grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/dirty/checking/DirtyCheckCollectionReassignmentSpec.groovy:29-32`

The four `org.grails.datastore.mapping.collection.*` / `core.Session` imports were appended as a separate block after the `dirty.checking.*` imports. Every other `org.grails` block in the file is one alphabetical group, and `collection` sorts before `dirty`. Neither `checkstyleTest` nor `codenarcTest` runs in this build (both are `SKIPPED`), so nothing enforces it; purely for consistency with the rest of the file. Move the four lines up into the existing `org.grails` block:

```groovy
import org.grails.datastore.mapping.collection.PersistentList
import org.grails.datastore.mapping.collection.PersistentSet
import org.grails.datastore.mapping.collection.PersistentSortedSet
import org.grails.datastore.mapping.core.Session
import org.grails.datastore.mapping.dirty.checking.DirtyCheckableCollection
...
```

### Nit: The subclass case could also assert the assignment is flagged

Reference:

- `grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/dirty/checking/DirtyCheckCollectionReassignmentSpec.groovy:378`

The mirror case for the wrapper side (`'a store-specific wrapper subclass is never replaced by a generic rewrap'`) asserts both that the value is left raw and that `hasChanged('shares')` is true, since the store persister relies on the assignment being flagged to re-wrap on save. The new `PersistentCollection` subclass case asserts only the first half. Adding `entity.hasChanged('shares')` to the `then:` block makes the two cases symmetric and pins that the setter's `markDirty` still fires when the old value is a `PersistentCollection` (it does today: `['a'] == emptyPersistentList` is false).

## Verified as correct

- **`Mock(Session)` inside the `where:` closures.** The mock is created in the `given:` block of each iteration and passed into the data-table closure, not created at data-collection time, so Spock's mock lifecycle is respected. All three iterations pass.
- **The parameterised `SortedSet` row exercises `DirtyCheckingSortedSet`.** `ranked` is declared `SortedSet<String>` on the test entity, and the `then:` block's `DirtyCheckableCollection` check is satisfied by the `SortedSet` wrapper; the earlier `'reassigning a plain SortedSet over a tracked SortedSet'` case already pins the exact wrapper class for that property.
- **No production code changed this round.** Only the encoder import line, the spec and the docs moved, so the Mongo, Neo4j and TCK results from the previous round still describe this head, and CI confirms it.

codeconsole:
Thanks for the approval — both nits taken in `40fe646ca6`.

**Import order.** The four `collection.*` / `core.Session` imports now sit inside the existing `org.grails` block, alphabetically ahead of `dirty.checking.*`.

**Subclass case symmetry.** Added `entity.hasChanged('shares')` to its `then:`, so it mirrors the wrapper-side subclass case and pins that the setter still flags the assignment when the old value is a `PersistentCollection` — the thing the store persister relies on to re-wrap on save.

`DirtyCheckCollectionReassignmentSpec` is 21/21 green on this head. Test-only change; no production code moved.

sbglasius (approved):

PR #16291, opened 2026-09-01
M6 FIX: Require a username and password on the spring-boot-starter-security User

codeconsole:
Two defects in the `spring-boot-starter-security` `User` template, both of which only show up once you generate an app and run it.

## 1. The credentials were not actually required

Grails 8 makes an unconstrained persistent property nullable by default (documented in `upgrading80x.adoc`, "GORM Properties Are Nullable by Default"). The template was never updated for that flip, so its `blank: false` constraints no longer make the credentials required — a `User` with a null username and a null password validates clean, and the scaffolded user admin will save one.

Every other domain template in the forge already declares `nullable: false`, which is what makes this look like a miss rather than a choice:

| Template | Constraint |
|---|---|
| `role.rocker.raw` | `authority unique: true, nullable: false` |
| `userClassic.rocker.raw` | `username unique: true, nullable: false` / `password nullable: false` |
| `userRole.rocker.raw` | `user nullable: false` / `role nullable: false, unique: 'user'` |
| `user.rocker.raw` | `username blank: false, unique: true` ← missing |

```groovy
static constraints = {
    username nullable: false, blank: false, unique: true
    password nullable: false, blank: false, password: true
}
```

### The generated spec failed out of the box

Creating an app with this feature produced a red build on the very first `./gradlew test`:

```
UserSpec > username and password are required FAILED
    Condition not satisfied:
    !new User(username: '', password: '').validate()
    ||                                    |
    |example.User : (unsaved)             true
    false
```

The assertion exercised neither constraint it appeared to test. The domain map constructor binds through data binding with `convertEmptyStringsToNull`, so `username: ''` arrives as `null` — the `blank` constraint never sees an empty string — and nullable-by-default then let the nulls through.

The spec now covers the two constraints as separate features, assigning the blank values directly since that is the only way a blank string actually reaches the property:

```groovy
void 'a user whose username and password are blank is rejected'() {
    given: 'the values are assigned directly, since the map constructor binds an empty string to null'
    User user = new User()
    user.username = ''
    user.password = ''

    expect:
    !user.validate()

    and:
    with(user.errors) {
        getFieldError('username').code == 'blank'
        getFieldError('password').code == 'blank'
    }
}
```

## 2. The mapping directive did not match the data store

The template emitted `table 'users'` for every data implementation, but `table` is a Hibernate directive. On MongoDB it is silently ignored, so the reason it is there — don't name the persistence unit `user`, a reserved word in PostgreSQL, H2 and SQL Server — was lost exactly where nothing warns about it. Reading the embedded MongoDB of a generated app, before and after:

| Directive | Collections created | docs in `users` | docs in `user` |
|---|---|---|---|
| `table 'users'` | `["user.next_id", "user"]` | 0 | 2 |
| `collection 'users'` | `["users.next_id", "users"]` | 2 | 0 |

The directive is now chosen from the selected features, the way `GrailsBase` and `DatabaseMigrationPlugin` already do, rather than from the gorm option — so it holds whether the implementation was picked with `--data` or by naming the feature:

```
@if (features.contains("gorm-mongodb")) {
    static mapping = {
        collection 'users'
    }

} else if (features.contains("gorm-hibernate5") || features.contains("gorm-hibernate7")) {
    static mapping = {
        table 'users'
    }

}
```

Hibernate 5 and 7 keep `table 'users'`, so the default app is unchanged. Neo4j gets no mapping block rather than a directive it does not understand.

## Verification

Generated a real app (`create-app --data mongodb --features spring-boot-starter-security --jdk 25`) and ran it, rather than only asserting on template text:

- `:grails-forge-core:test` — 347 tests, 0 failures
- `grails-forge` `codeStyle` — clean
- Generated app's `UserSpec` — 4/4 pass
- Reverting `nullable: false` and re-running makes the nullable feature fail, so the new spec is not vacuous. The blank feature passes either way — `blank: false` was always correct; the nullable gap was the whole defect.
- Each rendered variant is parsed to `Phases.CONVERSION` to confirm it is valid Groovy; that check caught a swallowed blank line before `static constraints` that the `contains` assertions did not.
- New tests unroll across HIBERNATE5 / HIBERNATE7 / MONGODB, asserting both the expected directive and that the other one does not leak in, plus a Neo4j case asserting no mapping block.

The app was also exercised end to end on embedded MongoDB: `/` anonymous 200, `/user` anonymous redirects to login, form login succeeds, `/user` as `ROLE_ADMIN` 200, `/user` as `ROLE_USER` 403.

### DIFF
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/security/SecurityFeature.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/security/SecurityFeature.java
@@ -76,7 +76,8 @@ public abstract class SecurityFeature implements Feature {
     protected void applyClassicDomainModel(GeneratorContext generatorContext) {
         final Project project = generatorContext.getProject();
         generatorContext.addTemplate("securityUser",
-                new RockerTemplate("grails-app/domain/{packagePath}/User.groovy", userClassic.template(project)));
+                new RockerTemplate("grails-app/domain/{packagePath}/User.groovy",
+                        userClassic.template(project, generatorContext.getFeatures())));
         generatorContext.addTemplate("securityRole",
                 new RockerTemplate("grails-app/domain/{packagePath}/Role.groovy", role.template(project)));
         generatorContext.addTemplate("securityUserRole",
--- a/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/security/SpringBootStarterSecurity.java
+++ b/grails-forge/grails-forge-core/src/main/java/org/grails/forge/feature/security/SpringBootStarterSecurity.java
@@ -76,7 +76,8 @@ public class SpringBootStarterSecurity extends SecurityFeature implements Primar
 
         final Project project = generatorContext.getProject();
         generatorContext.addTemplate("securityUser",
-                new RockerTemplate("grails-app/domain/{packagePath}/User.groovy", user.template(project)));
+                new RockerTemplate("grails-app/domain/{packagePath}/User.groovy",
+                        user.template(project, generatorContext.getFeatures())));
         generatorContext.addTemplate("securityUserController",
                 new RockerTemplate("grails-app/controllers/{packagePath}/UserController.groovy", userController.template(project)));
         generatorContext.addTemplate("securityUserService",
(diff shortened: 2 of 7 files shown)


borinquenkid (approved):
Reviewed the nullable/blank constraint fix and the Hibernate/MongoDB mapping-directive fix for `spring-boot-starter-security`'s `User` template.

Also pushed a follow-up commit extending the same mapping-directive fix to the classic `User`/`Role`/`UserRole` model shared by `grails-spring-security` and `grails-spring-security-ui` (`userClassic.rocker.raw`), which had the identical Hibernate-only `table` directive and no validator restricting it from MongoDB. Added matching test coverage in `GrailsSpringSecuritySpec` for Hibernate5/7, MongoDB, and Neo4j. All targeted tests pass and `codeStyle` is clean.

Issue #16289, opened 2026-09-01
Jupiter engine is not on the test runtime classpath by convention, so a module can silently run zero tests


sbglasius:
Split out from review discussion on https://github.com/apache/grails-core/pull/16281.

## Problem

`gradle/test-config.gradle` adds `testRuntimeOnly 'org.junit.platform:junit-platform-launcher'` and calls `useJUnitPlatform()`, but never adds `org.junit.jupiter:junit-jupiter-engine`.

Without the engine on the test runtime classpath, JUnit 5 tests are not discovered — and the build does not fail. The test task reports `BUILD SUCCESSFUL` while executing zero tests.

This was found in `grails-gsp/grails-taglib`, whose only JUnit 5 test (`GroovyPageAttributesTests`) had never executed. Reverting just the one-line engine addition in that module reproduces it: `:grails-taglib:test` succeeds and discovers nothing.

## Why it usually works

Every other module with JUnit 5 tests receives the engine only *transitively*, via `runtimeOnly 'org.junit.jupiter:junit-jupiter-engine'` in `grails-testing-support-core`. The six modules that declare `junit-jupiter-api` without an explicit engine all resolve it that way today. If that transitive edge is ever narrowed, the same silent-skip returns in those modules with a green build.

## Suggested fix

Add the Jupiter engine to the `gradle/test-config.gradle` convention alongside the platform launcher, so no module can silently skip its JUnit 5 tests, and drop the now-redundant module-local declaration in `grails-gsp/grails-taglib/build.gradle`.

Worth considering alongside this: a guard that fails the build when a test task discovers zero tests in a module that has test sources, so this class of problem cannot be green again.

The module-local fix landed in #16281 to unblock that PR; this issue tracks the convention-level fix.

jdaugherty:
@matrei  given that the behavior is a silent failure, why aren't we exporting this with an api dependency in our grails-testing-support-core module?

matrei:
> @matrei  given that the behavior is a silent failure, why aren't we exporting this with an api dependency in our grails-testing-support-core module? 

Because we should be using Spock? 😄Also, it's a runtime dependency so I don't think API would be the right scope.

jdaugherty:
@matrei have we eliminated all junit 5 tests in core yet?  I thought we still had some.

matrei:
> @matrei have we eliminated all junit 5 tests in core yet?  I thought we still had some.

[...]

PR #16293, opened 2026-09-02
Fix params and attrs entries shadowed by bean accessors on Groovy 5

### LINKED ISSUE #16280: GrailsParameterMap: params['identifier'] and params['request'] resolve to bean properties instead of map entries on Groovy 5

sbglasius:
Fixes #16280
Supersedes #16281

## Root cause

The issue and #16281 both attributed this to a change in `DefaultGroovyMethods` overload selection. That is not what happened — the `getAt`/`putAt` signatures are byte-identical in Groovy 4.0.33 and 5.1.0, and selection among them did not change.

The change is inside `MetaClassImpl`, for classes that implement `Map`:

| | Groovy 4.0.33 | Groovy 5.1.0 |
|---|---|---|
| `getProperty` | `if (isMap && !isStatic) return ((Map) object).get(name);` — unconditionally, before any getter lookup | getter lookup runs first; the map `get` is only reached when there is no visible public getter |
| `setProperty` | setter if one exists, **else** `((Map) object).put(name, newValue)` | the map-`put` branch is guarded by `(mp == null \|\| !mp.isPublic() \|\| isSpecialProperty(name))`, so a public getter-only bean property falls through to `throw new ReadOnlyPropertyException` |

So the trigger is simply: **a public JavaBean accessor on a `Map` implementation now shadows the map entry of the same name.** The subscript form reaches the same code via `DefaultGroovyMethods.putAt(Object, String, Object)` → `InvokerHelper.setProperty`.

`GrailsParameterMap` declared `getRequest()` and `getIdentifier()`, which is why request parameters named `request` and `identifier` became unreadable and threw on assignment.

## Fix

Remove the colliding accessors. With no bean accessor of that name, Groovy 5 reaches the map on every path:

- dynamic — `getProperty` → `method == null` → `isMap` → `Map.get`; `setProperty` → `mp == null` → `Map.put`
- `@CompileStatic` — `StaticTypeCheckingVisitor.getTypeForMapPropertyExpression` types it as a map entry and `StaticTypesCallSiteWriter` emits `Map.get` / `Map.put`
- subscript, both modes — DGM `getAt`/`putAt` land on the same `get`/`put`

The whole production change is two methods renamed, one deleted, and five call sites. No `getProperty`, `setProperty`, `getAt` or `putAt` override.

### Why not the metaclass overrides in #16281

#16281 kept the accessors and overrode `getProperty`/`setProperty`/`getAt`/`putAt` on `AbstractTypeConvertingMap`. That cannot reach statically compiled call sites. In `StaticTypesCallSiteWriter`, `makeGetPropertyWithGetter(...)` is tried **before** `writeMapDotProperty(...)`, and the mutator loop runs before the `Map.put` branch — so `@CompileStatic` binds to the declared accessor and never consults the metaclass hooks. #16281 had to document that as a limitation. It also regressed `attrs.gspTagSyntaxCall = false` from "invoke the setter" (Grails 7) to "write a map entry".

The same reasoning rules out an annotation plus AST transform: a local transform on the map class can only generate those same runtime hooks, and the static compiler at the *call site* still binds to the declared getter. It would add public API without fixing the static half.

Thanks to @jdaugherty, whose review on #16281 established both of those points and redirected this to restoring the behaviour rather than documenting around it.

## API changes

| Grails 7 | Grails 8 |
|---|---|
| `params.getRequest()` | `params.request()` |
| `params.getIdentifier()` | `params.id` — it only ever read the `id` parameter |
| `GroovyPageAttributes.isGspTagSyntaxCall()` | `gspTagSyntaxCall()` (internal class) |

`getRequest()` gained a non-bean replacement rather than being deleted, because there is a real caller: `AbstractRequestBodyDataBindingSourceCreator` reads the request body from a `GrailsParameterMap` binding source. `getIdentifier()` had exactly one caller in the monorepo (`Controller.groovy`), now reading `GormProperties.IDENTITY` directly, so the `id`-only semantics are unchanged.

`setGspTagSyntaxCall(boolean)` is deliberately retained. A write-only property reproduces Grails 7 exactly — the read addresses the map, the assignment invokes the setter — which avoids the regression #16281 introduced. `TagOutput` and `GroovyPage` rely on that setter.

`AbstractTypeConvertingMap` gains a javadoc statement of the invariant so this does not recur: a subclass must not declare a no-argument `getX()`/`isX()`, because such an accessor cannot be worked around at runtime.

## Tests

- `GrailsParameterMapTests` — 7 new cases: `identifier` and `request` via subscript and dotted syntax, values arriving from the request, absent keys reading `null`, `remove`/`clone`, `request()` still reachable, and a `@CompileStatic` helper pinning statically compiled read *and* write in both forms.
- `GroovyPageAttributesTests` — 4 new cases covering the `attrs` side, including the static read and the setter-parity case.
- `TypeConvertingMapTests` — one case at the base-class level.
- `CommandObjectInstantiationSpec` — a request submitting **only** `identifier` (no `id`), so `SimpleMapDataBindingSource.getIdentifierValue()` returns null and the `Controller` fallback actually runs. This is the shape @jdaugherty asked for; the earlier version passed regardless of `getIdentifier()`.
- `AbstractRequestBodyDataBindingSourceCreatorSpec` — covers the `GrailsParameterMap` branch that now calls `request()`.

Every new test was mutation-checked. Reinstating the accessors makes `GrailsParameterMapTests` fail to **compile** with `Cannot set read-only property: identifier` — the exact static error #16281 documented — and with that one case removed so the file compiles, all 7 dynamic cases fail. Reinstating `isGspTagSyntaxCall()` fails the two `attrs` read cases. Pointing the `Controller` fallback at `identifier` fails the command-object spec on `model.commandObject == null`, the resolution itself.

### Build fix

`grails-taglib` had no `junit-jupiter-engine` on its test runtime classpath. Its Spock specs ran (spock-core brings its own engine), but `GroovyPageAttributesTests` — the module's only plain JUnit 5 class — was never discovered. Added, module-local; the convention-level fix is tracked in #16289.

## Documentation

`upgrading80x.adoc` §28.3 — the Groovy 5 change, the restored rule, and the two method renames. No caveat section is needed any more, because `params` is uniformly a map again in both compilation modes.

### DIFF
--- a/grails-controllers/src/main/groovy/grails/artefact/Controller.groovy
+++ b/grails-controllers/src/main/groovy/grails/artefact/Controller.groovy
@@ -374,7 +374,7 @@ trait Controller implements ResponseRenderer, ResponseRedirector, RequestForward
                 if (entityIdentifierValue == null) {
                     final GrailsWebRequest webRequest = GrailsWebRequest
                             .lookup(request)
-                    entityIdentifierValue = webRequest?.getParams().getIdentifier()
+                    entityIdentifierValue = webRequest?.getParams()?.get(GormProperties.IDENTITY)
                 }
             }
             if (entityIdentifierValue instanceof String) {
--- a/grails-core/src/main/groovy/grails/util/AbstractTypeConvertingMap.java
+++ b/grails-core/src/main/groovy/grails/util/AbstractTypeConvertingMap.java
@@ -38,6 +38,23 @@ import org.apache.grails.core.internal.util.TypeConverters;
  * Type converting maps have no inherent ordering. Two maps with identical entries
  * but arranged in a different order internally are considered equal.
  *
+ * <h2>Subclasses must not declare JavaBean accessors</h2>
+ *
+ * A subclass must not declare a no-argument {@code getX()} or {@code isX()} method, because this
+ * class implements {@link Map} and Groovy resolves a JavaBean accessor ahead of the map entry of
+ * the same name. Such an accessor makes the entry {@code x} unreadable through
+ * {@code map.x} and {@code map['x']}, and makes assignment to it fail with
+ * {@code ReadOnlyPropertyException}. It applies to statically compiled callers too: the static
+ * compiler binds to the declared accessor and never reaches {@code getProperty}/{@code setProperty},
+ * so the collision cannot be worked around at runtime.
+ *
+ * Expose such a value under a name that is not a JavaBean accessor - for example
+ * {@code GrailsParameterMap.request()} - or under a method that takes an argument.
+ *
+ * A setter is a weaker case and is allowed: it leaves reads addressing the map, but assignment to
+ * that one name invokes the setter instead of storing an entry, so such an entry must be written
+ * with {@link Map#put}. {@code GroovyPageAttributes.setGspTagSyntaxCall(boolean)} is the only one.
+ *
  * @author Graeme Rocher
  * @author Lari Hotari
  * @since 1.2
--- a/grails-core/src/test/groovy/org/grails/util/TypeConvertingMapTests.groovy
+++ b/grails-core/src/test/groovy/org/grails/util/TypeConvertingMapTests.groovy
@@ -156,6 +156,21 @@ class TypeConvertingMapTests {
         assert toTypeConverting(a: 1, b: 2).hashCode() != ["b": 2, a: 1].hashCode()
     }
 
+    // https://github.com/apache/grails-core/issues/16280
+    // The base class declares no no-argument getX()/isX() accessor, so no key name is shadowed.
+    // Keys named after methods that do exist on the hierarchy are ordinary entries.
+    @Test
+    void testKeysNamedAfterMethodsOnTheHierarchyAreOrdinaryEntries() {
+        def map = toTypeConverting([:])
+
+        ['identifier', 'request', 'gspTagSyntaxCall', 'byte', 'int', 'wrappedMap'].each { String key ->
+            map[key] = "value of $key".toString()
+            assert map[key] == "value of $key".toString()
+            assert map.get(key) == "value of $key".toString()
+            assert map.containsKey(key)
+        }
+    }
+
     protected toTypeConverting(map) {
         new TypeConvertingMap(map)
     }
--- a/grails-doc/src/en/guide/upgrading/upgrading80x.adoc
+++ b/grails-doc/src/en/guide/upgrading/upgrading80x.adoc
@@ -1488,6 +1488,39 @@ def value = configObject.containsKey(key) ? configObject.get(key) : null
 
 Grails applies this guard internally when resolving Spring profile activation keys; you only need it in application code that navigates a raw `ConfigObject` by subscript.
 
+===== 28.3 A getter on a map no longer hides behind a map entry
+
+In Groovy 4, reading or writing a property on a class that implements `Map` always addressed a map entry.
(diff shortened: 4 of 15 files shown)


sbglasius:
Local verification on this branch:

- `./gradlew clean aggregateViolations :grails-test-report:check --continue` — **BUILD SUCCESSFUL**, 2407 tasks, 28m42s. Aggregate reports: **18,493 tests, 0 failures, 0 errors**, 448 skipped (14,708 `test` + 3,785 `integrationTest`). Note this is one local configuration; CI multiplies it across the Java 21/25 and indy matrix.
- Checkstyle, CodeNarc, PMD and SpotBugs all report no violations.
- `#16030` (`EndToEndSpec > async multiple levels of layouts`) did not fire locally. It is a known ~1% flake and failed on the superseded PR's CI run for unrelated reasons, so if it appears here it is not a regression from this change.

Every new test was mutation-checked against reinstated accessors — details in the Tests section above.

jdaugherty (approved):

matrei (approved):

PR #16296, opened 2026-09-02
Fix marshalling of non-public classes and Serializable beans

### LINKED ISSUE #16294: as JSON: IllegalAccessException marshalling a non-public class (anonymous/inner) that implements a public interface
### Expected Behavior

`render(someObject as JSON)` should serialize an object whose class is not `public` — an anonymous
class, a local class, or a package-private class — as long as the property getters themselves are
`public`. This is a very common shape in Spring Security code, where you hand out an anonymous
implementation of a public interface such as `UserDetails`.

### Actual Behaviour

The request fails with HTTP 500:

```
org.grails.web.converters.exceptions.ConverterException:
    Error converting Bean with class com.example.DemoController$1
	at org.grails.web.converters.marshaller.json.GroovyBeanMarshaller.marshalObject(GroovyBeanMarshaller.java:89)
	at grails.converters.JSON.value(JSON.java:209)
	at org.grails.web.converters.marshaller.json.MapMarshaller.marshalObject(MapMarshaller.java:47)
	...
Caused by: java.lang.IllegalAccessException:
    class org.grails.web.converters.marshaller.json.GroovyBeanMarshaller
    cannot access a member of class com.example.DemoController$1 with modifiers "public"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:400)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:670)
	at java.base/java.lang.reflect.Method.invoke(Method.java:556)
	at org.grails.web.converters.marshaller.json.GroovyBeanMarshaller.marshalObject(GroovyBeanMarshaller.java:69)
```

**Root cause.** `GroovyBeanMarshaller.java:69` invokes the read method with no accessibility
handling at all:

```java
Object value = readMethod.invoke(o, (Object[]) null);
```

`BeanUtils.getPropertyDescriptors(clazz)` returns, for the *overridden* properties, read methods
whose declaring class is the anonymous class itself. An anonymous class is effectively
package-private, so a caller in another package cannot invoke those methods even though the methods
carry the `public` modifier. `setAccessible`/`makeAccessible` is never called — grepping
`grails-converters` for `makeAccessible|setAccessible` returns nothing.

Note the asymmetry this produces. Reproduced standalone against the reproducer app's own classpath,
the `UserDetails` **default** methods keep the interface as their declaring class and succeed, while
the three overridden ones fail:

```
clazz=com.example.probe.Factory$1 public=false
  accountNonExpired:      declared on ...UserDetails  -> OK (true)
  accountNonLocked:       declared on ...UserDetails  -> OK (true)
  authorities:            declared on ...Factory$1    -> IllegalAccessException
  credentialsNonExpired:  declared on ...UserDetails  -> OK (true)
  enabled:                declared on ...UserDetails  -> OK (true)
  password:               declared on ...Factory$1    -> IllegalAccessException
  username:               declared on ...Factory$1    -> IllegalAccessException
      getInterfaceMethodIfPossible -> UserDetails => OK (user)
      makeAccessible               -> OK (user)
```

### Suggested fix

Resolve the read method to a publicly accessible interface method, and make it accessible:

```java
Method invokable = ClassUtils.getInterfaceMethodIfPossible(readMethod, clazz);
ReflectionUtils.makeAccessible(invokable);
Object value = invokable.invoke(o, (Object[]) null);
```

All four marshallers call `readMethod.invoke` bare and need the same treatment (line numbers on
`8.0.x`):

- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GroovyBeanMarshaller.java:69`
- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GenericJavaBeanMarshaller.java:67`
- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/xml/GroovyBeanMarshaller.java:69`
- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/xml/GenericJavaBeanMarshaller.java:49`

I verified the fix by shadowing the marshaller classes in the reproducer's `src/main/groovy` with
patched copies: the request then returns HTTP 200 and the integration test passes.

Fixing this bug alone is **not** sufficient to make the reproducer pass — it then hits a second,
independent bug in `GenericJavaBeanMarshaller`, filed separately as the companion issue.

### Affected versions

This is not a regression. The bare `readMethod.invoke(...)` call is present on every branch that
carries the `grails-converters` module — `7.0.x`, `7.1.x`, `7.2.x`, `8.0.x`, `8.1.x` and `9.0.x`.
I reproduced it on 8.0.0-M5 only; the other branches are identified by code inspection.

### Steps To Reproduce

```bash
git clone https://github.com/sbglasius/asjson-bug
cd asjson-bug
./gradlew integrationTest
```

`DemoControllerTest > simple GET on show` fails. The controller action is simply:

```groovy
def show() {
    def user = new UserDetails() {
        @Override Collection<? extends GrantedAuthority> getAuthorities() { [new SimpleGrantedAuthority('ROLE_ADMIN')] }
        @Override String getPassword() { null }
        @Override String getUsername() { "user" }
    }
    render([user: user] as JSON)
}
```

`DemoControllerTest > simple GET on index`, which renders a plain `Map`, passes — so the failure is
specific to marshalling the non-public class.

### Environment Information

- Operating System: Linux (kernel 6.17)
- JDK Version: 25.0.4 (BellSoft Liberica)
- Groovy 5.0.8, Spring Boot 4.1.0, Spring Framework 7.0.8, Spring Security 7.1.0
- Gradle 9.6

### Example Application

https://github.com/sbglasius/asjson-bug

### Version

8.0.0-M5

### LINKED ISSUE #16295: as JSON: GenericJavaBeanMarshaller calls field.canAccess(o) before the static check, throwing IllegalArgumentException on any Serializable bean
### Expected Behavior

`render(someObject as JSON)` should serialize an ordinary `Serializable` Java bean. Practically every
Java bean declares `private static final long serialVersionUID`, so this covers a very large share of
all objects users pass to `as JSON`.

### Actual Behaviour

The request fails with HTTP 500 as soon as the object graph reaches a `Serializable` Java bean — here
Spring Security's `SimpleGrantedAuthority`:

```
org.grails.web.converters.exceptions.ConverterException:
    Error converting Bean with class org.springframework.security.core.authority.SimpleGrantedAuthority
	at org.grails.web.converters.marshaller.json.GenericJavaBeanMarshaller.marshalObject(GenericJavaBeanMarshaller.java:87)
	at org.grails.web.converters.marshaller.json.CollectionMarshaller.marshalObject(CollectionMarshaller.java:43)
	...
Caused by: java.lang.IllegalArgumentException: non-null object for
    private static final long org.springframework.security.core.authority.SimpleGrantedAuthority.serialVersionUID
	at org.grails.web.converters.marshaller.json.GenericJavaBeanMarshaller.marshalObject(GenericJavaBeanMarshaller.java:74)
```

**Root cause.** `GenericJavaBeanMarshaller.java:74` has its `&&` operands in the wrong order:

```java
for (Field field : o.getClass().getDeclaredFields()) {
    int modifiers = field.getModifiers();
    if (field.canAccess(o) && Modifier.isPublic(modifiers) && !(Modifier.isStatic(modifiers) || Modifier.isTransient(modifiers))) {
```

Per its javadoc, `AccessibleObject.canAccess(obj)` throws `IllegalArgumentException` when the member
is `static` and `obj` is non-null. Because `canAccess(o)` is evaluated **first**, the
`Modifier.isStatic(...)` guard never gets the chance to short-circuit it, and the loop blows up on
the first static field it meets — typically `serialVersionUID`.

### Suggested fix

Swap the operands so the cheap modifier checks run first:

```java
if (Modifier.isPublic(modifiers) && !(Modifier.isStatic(modifiers) || Modifier.isTransient(modifiers)) && field.canAccess(o)) {
```

Two call sites (line numbers on `8.0.x`):

- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GenericJavaBeanMarshaller.java:74`
- `grails-converters/src/main/groovy/org/grails/web/converters/marshaller/xml/GenericJavaBeanMarshaller.java:57`

Verified by shadowing the class in the reproducer's `src/main/groovy` with a patched copy.

### Regression

This is a regression introduced in the 8.x line, in commit
[`9e60b8a4de`](https://github.com/apache/grails-core/commit/9e60b8a4de10822ecb22d404a38e21840f5ab821)
("refactor: replace usages of deprecated external APIs"). `7.2.x` and earlier read:

```java
if (field.isAccessible() && Modifier.isPublic(modifiers) && !(Modifier.isStatic(modifiers) || Modifier.isTransient(modifiers))) {
```

`isAccessible()` takes no argument and never throws, so the operand order was harmless there. The
mechanical `isAccessible()` → `canAccess(o)` swap made the first operand throwing, which is what
turns this into a hard failure.

Worth noting while fixing: on `7.2.x` this branch was effectively dead code, because
`field.isAccessible()` returns `false` for any field that was never `setAccessible(true)`-ed. So
`canAccess(o)` is the right *intent* — it makes public fields on public classes actually serialize —
it just has to be evaluated last.

Present on `8.0.x`, `8.1.x` and `9.0.x`; the commit is contained in tags `v8.0.0-M2` through
`v8.0.0-M6`. Not present on `7.x`.

### Steps To Reproduce

```bash
git clone https://github.com/sbglasius/asjson-bug
cd asjson-bug
./gradlew integrationTest
```

`DemoControllerTest > simple GET on show` fails.

Note this bug is **masked** by the companion issue: out of the box the request dies earlier, in
`GroovyBeanMarshaller`, on the anonymous `UserDetails` class. Patch that first (or hand `as JSON` any
`Serializable` bean with a public class) and you land on this one. The reproducer's README documents
both and shows how to shadow the marshallers to step through them one at a time.

### Environment Information

- Operating System: Linux (kernel 6.17)
- JDK Version: 25.0.4 (BellSoft Liberica)
- Groovy 5.0.8, Spring Boot 4.1.0, Spring Framework 7.0.8, Spring Security 7.1.0
- Gradle 9.6

### Example Application

https://github.com/sbglasius/asjson-bug

### Version

8.0.0-M5

### Companion issue

Masked by https://github.com/apache/grails-core/issues/16294 — fix that one first to reach this bug.

sbglasius:
Fixes #16294
Fixes #16295

`someObject as JSON` (and `as XML`) failed for two very common shapes of object. Both are fixed here, since fixing the first one only exposes the second.

## #16294 — `IllegalAccessException` on a non-public class

A class that is not public — anonymous, local or package-private — cannot have its read methods invoked reflectively from another package, even though the methods themselves carry the `public` modifier. All four bean marshallers called `readMethod.invoke(...)` with no accessibility handling, so handing `as JSON` an anonymous implementation of a public interface (the usual shape of a Spring Security `UserDetails`) failed with:

```
java.lang.IllegalAccessException: class org.grails.web.converters.marshaller.json.GroovyBeanMarshaller
    cannot access a member of class com.example.DemoController$1 with modifiers "public"
```

The read method is now resolved to the interface method where one exists, and made accessible otherwise:

```java
Method invokableMethod = ClassUtils.getInterfaceMethodIfPossible(readMethod, clazz);
ReflectionUtils.makeAccessible(invokableMethod);
Object value = invokableMethod.invoke(o, (Object[]) null);
```

Resolving to the interface is not sufficient on its own — a package-private class with a public getter and no interface has nowhere to resolve to — so `makeAccessible` carries that case.

The public *field* loops of both `GroovyBeanMarshaller`s failed identically on `field.get(o)` and are now made accessible too.

## #16295 — `IllegalArgumentException` on any `Serializable` bean

`GenericJavaBeanMarshaller` evaluated `field.canAccess(o)` before the static check, and per its javadoc `canAccess` throws for a static member when the object is non-null. Any bean declaring `private static final long serialVersionUID` — nearly every `Serializable` bean — therefore failed:

```
java.lang.IllegalArgumentException: non-null object for
    private static final long org.springframework.security.core.authority.SimpleGrantedAuthority.serialVersionUID
```

The modifier checks now run first so they short-circuit before `canAccess`. This is a regression in the 8.x line from `9e60b8a4de`, which mechanically swapped the non-throwing `isAccessible()` for `canAccess(o)`.

## One change beyond the two issues

Groovy compiles the variables captured by an anonymous class into `ACC_PUBLIC | ACC_SYNTHETIC` `groovy.lang.Reference` fields. Once the field loops could actually read them, they were emitted as duplicate keys with empty-object values (`{"name":{},"age":{}}`) — the new tests caught exactly that. Synthetic fields are compiler artifacts and never part of a bean's state, so they are now skipped in all four marshallers.

## Files changed

- `grails-converters/.../marshaller/json/GroovyBeanMarshaller.java`
- `grails-converters/.../marshaller/json/GenericJavaBeanMarshaller.java`
- `grails-converters/.../marshaller/xml/GroovyBeanMarshaller.java`
- `grails-converters/.../marshaller/xml/GenericJavaBeanMarshaller.java`

## Tests

New fixtures under `org.grails.web.converters.beans`, deliberately in a different package from the marshallers — in the same package the JVM access check passes and neither bug reproduces. They cover a public interface with both abstract and `default` read methods (mirroring `UserDetails`), Groovy and Java anonymous / package-private / no-interface implementations, an anonymous class with a declared public field, and a `Serializable` bean carrying `private static final long serialVersionUID` alongside a public constant and a public instance field.

Two new specs of 9 features each, driven through the public `new JSON(x)` / `new XML(x)` API: `json.NonPublicClassMarshallingSpec` and `xml.NonPublicClassMarshallingSpec`. Every non-public case first asserts `!Modifier.isPublic(person.getClass().modifiers)`, so the tests fail loudly rather than silently passing if a future compiler stops producing a non-public class.

With the production change reverted, **18 of 18 fail** with the two reported exceptions verbatim; with it, 18/18 pass. Also green locally: full `:grails-converters:test`, `:grails-rest-transforms:test`, `:grails-test-suite-web:test`, `:grails-test-suite-uber:test`, `:grails-web-common:test` and `:grails-converters:codeStyle`.

No documentation change: this restores the documented behaviour of `as JSON` / `as XML` and adds no public API.

### DIFF
--- a/grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
+++ b/grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
@@ -78,20 +78,20 @@ public final class ReflectionUtils {
     /**
      * Resolves a read method that can actually be invoked on {@code target}.
      *
-     * <p>Where the property is declared by an interface, the interface method is returned: it is
-     * declared by an accessible type, and virtual dispatch still reaches the implementation. Where
-     * it is not -- a non-public class with no interface declaring the getter -- a private copy of
-     * the method is widened, so that the accessibility flag cannot leak into a {@code Method}
-     * instance shared through a descriptor cache.
+     * <p>Where an equivalent method is declared by a public type anywhere in the hierarchy -- an
+     * interface or a superclass -- that one is returned: it is declared by an accessible type, and
+     * virtual dispatch still reaches the override. Only where there is no such type does the method
+     * need help, and then a private copy is widened, so that the accessibility flag cannot leak
+     * into a {@code Method} instance shared through a descriptor cache.
      *
      * @param readMethod  the property's read method, typically from a {@code PropertyDescriptor}
-     * @param targetClass the class being read, used to resolve the interface method
+     * @param targetClass the class being read, used to resolve the publicly accessible method
      * @param target      the instance being read, or {@code null} for a static read method
      * @return a method that may be invoked on {@code target}
      */
     public static Method resolveInvokableReadMethod(Method readMethod, Class<?> targetClass, @Nullable Object target)
             throws NoSuchMethodException {
-        Method invokable = ClassUtils.getInterfaceMethodIfPossible(readMethod, targetClass);
+        Method invokable = ClassUtils.getPubliclyAccessibleMethodIfPossible(readMethod, targetClass);
         if (canAccess(invokable, target)) {
             return invokable;
         }
@@ -151,10 +151,10 @@ public final class ReflectionUtils {
             return false;
         }
         if (LOG.isWarnEnabled()) {
-            LOG.warn("Class [{}] is not public, so its properties can only be read by widening access reflectively. " +
-                            "Declare the class public - as a named class rather than an anonymous one where " +
-                            "necessary - so that it reads as a standard JavaBean. This handling may be withdrawn in " +
-                            "a future major release. To silence this, set the log level of [{}] above WARN. " +
+            LOG.warn("Class [{}] is not public. Grails reads its properties through compatibility handling that may " +
+                            "be withdrawn in a future major release. Declare it as a named public class so that it " +
+                            "reads as a standard JavaBean, or register an ObjectMarshaller for it if the class is " +
+                            "not yours. To silence this, set the log level of [{}] above WARN. " +
                             "(warned once per class)",
                     clazz.getName(), LOG.getName());
         }
--- a/grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
+++ b/grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
@@ -1,0 +1,196 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.grails.common.reflect;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InaccessibleObjectException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.jspecify.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.util.ClassUtils;
+
+/**
+ * Reads the properties of a bean whose class is not {@code public}.
+ *
+ * <p>A class that is not public -- anonymous, local, or package-private -- cannot have its read
+ * methods invoked from another package even when the methods themselves are public, so reflection
+ * over such a bean needs help. Groovy 4 stamped {@code ACC_PUBLIC} on anonymous inner classes and
+ * Groovy 5 does not, which is why beans of that shape reach the framework at all.
+ *
+ * <p>This compatibility handling is deliberately visible: {@link #warnOnNonPublicClass} reports the
+ * class once so an application can be corrected, and so the handling can be withdrawn if the
+ * compiler stops producing non-public classes for this shape.
+ *
+ * @since 8.0.0
+ */
+public final class ReflectionUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ReflectionUtils.class);
+
+    /** Best-effort upper bound, so that generated class names cannot grow the set without limit. */
+    private static final int MAX_WARNED_CLASSES = 1024;
+
+    /**
+     * Keyed by class name rather than by {@link Class}, so that reporting a class never retains its
+     * class loader. A reload therefore does not report the same class again, which is intended: a
+     * reload loop would otherwise repeat every warning.
+     */
+    private static final Set<String> WARNED_NON_PUBLIC_CLASSES = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Packages whose classes an application cannot declare public itself, so reporting them would
+     * only be noise. {@code java.util.KeyValueHolder}, handed out by {@code Map.entry}, is the one
+     * that turns up in practice.
+     *
+     * <p>The {@code grails} packages are deliberately absent: a framework class reaching this code
+     * is worth seeing, and the tests declare their fixtures under {@code org.grails}.
+     */
+    private static final String[] NON_APPLICATION_PACKAGES = {
+        "java.", "javax.", "jakarta.", "groovy.", "org.apache.groovy.", "org.codehaus.groovy.",
+        "org.springframework."
+    };
+
+    private ReflectionUtils() {
+    }
+
+    /**
+     * Resolves a read method that can actually be invoked on {@code target}.
+     *
+     * <p>Where the property is declared by an interface, the interface method is returned: it is
+     * declared by an accessible type, and virtual dispatch still reaches the implementation. Where
+     * it is not -- a non-public class with no interface declaring the getter -- a private copy of
+     * the method is widened, so that the accessibility flag cannot leak into a {@code Method}
+     * instance shared through a descriptor cache.
+     *
+     * @param readMethod  the property's read method, typically from a {@code PropertyDescriptor}
+     * @param targetClass the class being read, used to resolve the interface method
+     * @param target      the instance being read, or {@code null} for a static read method
+     * @return a method that may be invoked on {@code target}
+     */
+    public static Method resolveInvokableReadMethod(Method readMethod, Class<?> targetClass, @Nullable Object target)
+            throws NoSuchMethodException {
+        Method invokable = ClassUtils.getInterfaceMethodIfPossible(readMethod, targetClass);
+        if (canAccess(invokable, target)) {
+            return invokable;
+        }
+        // getDeclaredMethod cannot fail here: invokable was resolved from this very class's methods.
+        Method widened = invokable.getDeclaringClass()
+                .getDeclaredMethod(invokable.getName(), invokable.getParameterTypes());
+        widened.setAccessible(true);
+        return widened;
+    }
+
+    /**
+     * Makes an instance field readable, widening it only when it is not readable already.
+     *
+     * <p>The field is expected to have come from {@link Class#getDeclaredFields}, which hands out a
+     * fresh copy on every call, so widening it is confined to the caller's own copy.
+     *
+     * @param field an instance field
+     * @param target the instance being read
+     * @return {@code false} when the field cannot be read, in which case the caller should skip it
+     *         rather than fail the whole read
+     */
+    public static boolean tryMakeReadable(Field field, @Nullable Object target) {
+        if (canAccess(field, target)) {
+            return true;
+        }
+        try {
+            field.setAccessible(true);
+            return true;
+        }
+        catch (InaccessibleObjectException | SecurityException e) {
+            // The declaring class is in a named module that does not open its package to us
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Cannot read field [{}] of [{}]: {}",
+                        field.getName(), field.getDeclaringClass().getName(), e.getMessage());
+            }
+            return false;
+        }
+    }
+
+    /**
+     * Reports {@code clazz} if it is not public, at most once per class.
+     *
+     * <p>Synthetic classes and classes from the JDK, Groovy and Spring are not reported: they are
+     * not written by the application whose log this is, so nobody reading it can act on them.
+     *
+     * <p>The bookkeeping deliberately runs before the log level is consulted, so that "once" does
+     * not depend on how logging happens to be configured.
+     *
+     * @return whether this call was the one that reported the class
+     */
+    public static boolean warnOnNonPublicClass(@Nullable Class<?> clazz) {
+        if (clazz == null || Modifier.isPublic(clazz.getModifiers()) || !isApplicationClass(clazz)) {
+            return false;
+        }
+        if (WARNED_NON_PUBLIC_CLASSES.size() >= MAX_WARNED_CLASSES ||
+                !WARNED_NON_PUBLIC_CLASSES.add(clazz.getName())) {
+            return false;
+        }
+        if (LOG.isWarnEnabled()) {
+            LOG.warn("Class [{}] is not public, so its properties can only be read by widening access reflectively. " +
+                            "Declare the class public - as a named class rather than an anonymous one where " +
+                            "necessary - so that it reads as a standard JavaBean. This handling may be withdrawn in " +
+                            "a future major release. To silence this, set the log level of [{}] above WARN. " +
+                            "(warned once per class)",
+                    clazz.getName(), LOG.getName());
+        }
+        return true;
+    }
+
+    /**
+     * Forgets which classes have been reported. Intended for tests, which would otherwise depend on
+     * whether another test in the same JVM reported the same class first.
+     */
+    public static void resetWarnedClasses() {
+        WARNED_NON_PUBLIC_CLASSES.clear();
+    }
+
+    /**
+     * {@code canAccess} rejects a non-null instance for a static member, which a caller iterating
+     * property descriptors or declared fields should not have to know.
+     */
+    private static boolean canAccess(Method method, @Nullable Object target) {
+        return method.canAccess(Modifier.isStatic(method.getModifiers()) ? null : target);
+    }
+
+    private static boolean canAccess(Field field, @Nullable Object target) {
+        return field.canAccess(Modifier.isStatic(field.getModifiers()) ? null : target);
+    }
+
+    private static boolean isApplicationClass(Class<?> clazz) {
+        if (clazz.isSynthetic()) {
+            return false;
+        }
+        String name = clazz.getName();
+        for (String prefix : NON_APPLICATION_PACKAGES) {
+            if (name.startsWith(prefix)) {
+                return false;
+            }
+        }
+        return true;
+    }
+}
--- a/grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
+++ b/grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
@@ -1,0 +1,137 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.grails.common.reflect
+
+import java.beans.Introspector
+import java.beans.PropertyDescriptor
+import java.lang.reflect.Field
+import java.lang.reflect.Method
+import java.lang.reflect.Modifier
+
+import org.apache.grails.common.reflect.beans.PublicFieldBean
+import org.apache.grails.common.reflect.beans.PublicThing
+import org.apache.grails.common.reflect.beans.ThingFactory
+
+import spock.lang.Specification
+
+/**
+ * Covers reading a bean whose class is not public. The fixtures deliberately live in another
+ * package: in this one the JVM access check passes and nothing would need widening.
+ */
+class ReflectionUtilsSpec extends Specification {
+
+    void setup() {
+        ReflectionUtils.resetWarnedClasses()
+    }
+
+    void cleanup() {
+        ReflectionUtils.resetWarnedClasses()
+    }
+
+    private static Method readMethodOf(Object bean, String property) {
+        PropertyDescriptor descriptor = Introspector.getBeanInfo(bean.getClass()).propertyDescriptors
+                .find { it.name == property }
+        descriptor.readMethod
+    }
+
+    void 'a read method declared by a public interface needs no widening'() {
+        given:
+        def thing = ThingFactory.anonymousThing('thing')
+        Method readMethod = readMethodOf(thing, 'name')
+
+        when:
+        Method invokable = ReflectionUtils.resolveInvokableReadMethod(readMethod, thing.getClass(), thing)
+
+        then: 'the interface declares it, so it is invokable as it stands'
+            invokable.declaringClass == PublicThing
+            invokable.invoke(thing) == 'thing'
+    }
+
+    void 'a read method with no accessible declaring type is widened on a copy'() {
+        given: 'a package-private class implementing nothing, so there is nowhere to resolve to'
+        def thing = ThingFactory.standaloneThing('thing')
+        Method readMethod = readMethodOf(thing, 'name')
+
+        expect: 'it cannot be invoked from this package to begin with'
+            !readMethod.canAccess(thing)
+
+        when:
+        Method invokable = ReflectionUtils.resolveInvokableReadMethod(readMethod, thing.getClass(), thing)
+
+        then:
+            invokable.invoke(thing) == 'thing'
+
+        and: 'the method handed in was left alone, so a shared descriptor cache keeps its own flag'
+            !invokable.is(readMethod)
+            !readMethod.canAccess(thing)
+    }
+
+    void 'a covariant read method resolves to the override rather than the bridge'() {
+        given:
+        def thing = ThingFactory.covariantThing('value')
+        Method readMethod = readMethodOf(thing, 'value')
+
+        when:
+        Method invokable = ReflectionUtils.resolveInvokableReadMethod(readMethod, thing.getClass(), thing)
+
+        then:
+            invokable.invoke(thing) == 'value'
+            !invokable.bridge
+    }
+
+    void 'a field that is already readable is not widened'() {
+        given:
+        def bean = new PublicFieldBean('label')
+        Field field = PublicFieldBean.getDeclaredField('label')
+
+        expect:
+            field.canAccess(bean)
+            ReflectionUtils.tryMakeReadable(field, bean)
+            field.get(bean) == 'label'
+    }
+
+    void 'a class that is not public is reported exactly once'() {
+        given:
+        Class<?> beanClass = ThingFactory.standaloneThing('thing').getClass()
+
+        expect: 'the first call reports it and the second finds it already reported'
+            ReflectionUtils.warnOnNonPublicClass(beanClass)
+            !ReflectionUtils.warnOnNonPublicClass(beanClass)
+
+        when:
+        ReflectionUtils.resetWarnedClasses()
+
+        then: 'the reset is what tests rely on to be independent of each other'
+            ReflectionUtils.warnOnNonPublicClass(beanClass)
+    }
+
+    void 'a public class is not reported'() {
+        expect:
+            !ReflectionUtils.warnOnNonPublicClass(PublicFieldBean)
+    }
+
+    void 'a class the application could not have declared public is not reported'() {
+        expect: 'java.util.KeyValueHolder is not public, but nobody reading the log can act on it'
+            !Modifier.isPublic(Map.entry('a', 'b').getClass().modifiers)
+            !ReflectionUtils.warnOnNonPublicClass(Map.entry('a', 'b').getClass())
+
+        and:
+            !ReflectionUtils.warnOnNonPublicClass(null)
+    }
+}
--- a/grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
+++ b/grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
@@ -18,12 +18,14 @@
  */
 package org.apache.grails.common.reflect
 
-import java.beans.Introspector
 import java.beans.PropertyDescriptor
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
 
+import org.springframework.beans.BeanUtils
+
+import org.apache.grails.common.reflect.beans.PublicCovariantBase
 import org.apache.grails.common.reflect.beans.PublicFieldBean
 import org.apache.grails.common.reflect.beans.PublicThing
 import org.apache.grails.common.reflect.beans.ThingFactory
@@ -44,8 +46,13 @@ class ReflectionUtilsSpec extends Specification {
         ReflectionUtils.resetWarnedClasses()
     }
 
+    /**
+     * Takes the read method the way the marshallers do. {@code java.beans.Introspector} pre-resolves
+     * a getter to its publicly accessible declaration, which would leave these features asserting
+     * nothing, so this asks Spring for the declaring-class method instead.
+     */
     private static Method readMethodOf(Object bean, String property) {
-        PropertyDescriptor descriptor = Introspector.getBeanInfo(bean.getClass()).propertyDescriptors
+        PropertyDescriptor descriptor = BeanUtils.getPropertyDescriptors(bean.getClass())
                 .find { it.name == property }
         descriptor.readMethod
     }
@@ -55,11 +62,15 @@ class ReflectionUtilsSpec extends Specification {
         def thing = ThingFactory.anonymousThing('thing')
         Method readMethod = readMethodOf(thing, 'name')
 
+        expect: 'the method as declared by the anonymous class cannot be invoked from here'
+            !readMethod.canAccess(thing)
+
         when:
         Method invokable = ReflectionUtils.resolveInvokableReadMethod(readMethod, thing.getClass(), thing)
 
-        then: 'the interface declares it, so it is invokable as it stands'
+        then: 'the interface declares it, so nothing has to be widened'
             invokable.declaringClass == PublicThing
+            invokable.canAccess(thing)
             invokable.invoke(thing) == 'thing'
     }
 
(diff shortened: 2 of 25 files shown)


grails-converters/src/test/groovy/org/grails/web/converters/beans/GroovyPersonFactory.groovy
```diff
@@ -0,0 +1,81 @@
...
+ */
+package org.grails.web.converters.beans
+
+import groovy.transform.CompileStatic
+
+/**
+ * Hands out instances of Groovy classes that are not public, but whose read methods are. This is the
+ * shape produced by an anonymous implementation of a public interface inside a Grails controller.
+ */
+@CompileStatic
```
bkoehm, line 27: Question: Any need to test a Groovy bean that isn't `@CompileStatic`?  Probably not, but I thought I'd ask the question anyway.

grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GenericJavaBeanMarshaller.java
```diff
@@ -64,14 +66,17 @@ public void marshalObject(Object o, JSON json) throws ConverterException {
                     if (Modifier.isStatic(readMethod.getModifiers())) continue;
                     if (readMethod.getAnnotation(PersistenceMethod.class) != null) continue;
                     if (readMethod.getAnnotation(ControllerMethod.class) != null) continue;
-                    Object value = readMethod.invoke(o, (Object[]) null);
+                    Method invokableMethod = ClassUtils.getInterfaceMethodIfPossible(readMethod, clazz);
```
jdaugherty, line 69: Doesn't this make it permanently accessible and not just for this code?

jdaugherty:
@sbglasius is this a result of Groovy now honoring the modifiers where previously it would treat package private / protected as public?

sbglasius:
> @sbglasius is this a result of Groovy now honoring the modifiers where previously it would treat package private / protected as public?

Yes, it is because Groovy 4 stamped ACC_PUBLIC on anonymous inner classes, Groovy 5 keeps them package-private. It would actually show in Groovy 4, if a class was marked `@PackageScope`.

grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GenericJavaBeanMarshaller.java
sbglasius: Yes, it does, but only for non-interface classes. The full answer goes like this:
```
== beans.PkgWithInterface  (class public: false)
  raw read method canAccess from this package: false
  getInterfaceMethodIfPossible -> beans.Public (same object as raw: false)
  resolved canAccess BEFORE makeAccessible: true      <- already accessible
  resolved canAccess AFTER  makeAccessible: true
  the raw class-declared method still canAccess: false <- untouched

== beans.PkgNoInterface  (class public: false)
  getInterfaceMethodIfPossible -> beans.PkgNoInterface (same object as raw: true)
  resolved canAccess BEFORE makeAccessible: false
  resolved canAccess AFTER  makeAccessible: true       <- flag flipped
  a later BeanUtils+resolve sees it accessible: true   (same object: true)
  a FRESH getDeclaredMethod copy canAccess: false      <- not globally opened
```

* Interface case (the reported bug — anonymous `UserDetails`): `getInterfaceMethodIfPossible` hands back a different Method, the interface's, which is public-on-public. `ReflectionUtils.makeAccessible` short-circuits and *nothing is mutated.* The class's own read method stays inaccessible.
* No-interface case (package-private class, public getter, nothing to resolve to): yes, `setAccessible(true)` fires and it persists. Two bounds on how far:
  * The flag lives on the Method instance, not on class metadata. `BeanUtils.getPropertyDescriptors` is backed by the static `CachedIntrospectionResults` cache and returns the same instance every call (confirmed above), so anything else in the JVM that asks Spring for that class's descriptors gets an already-invokable method, for the lifetime of that cache.
  * A fresh `getDeclaredMethod`/`getMethod` copy is still false. The member is not globally opened, and this grants nothing a caller couldn't get itself — any classpath code in the unnamed module can `setAccessible` a public method of a non-public class. It's also the pattern Spring itself uses on cached members (`AutowiredAnnotationBeanPostProcessor`, `AbstractNestablePropertyAccessor`).

If you'd rather not mutate shared cached state at all, the clean alternative is to work on our own copy — `clazz.getDeclaredMethod(name)` returns a new instance per call (also confirmed above), so `setAccessible` on that leaks nowhere. Cost is one extra reflective lookup per property per marshal unless we cache it ourselves.

grails-converters/src/test/groovy/org/grails/web/converters/beans/GroovyPersonFactory.groovy
sbglasius: It doesn't make a difference if it's `@CompileStatic` or not. Reflection wise the class is very similar looking.

sbglasius: A `DynamicGroovyPersonFactory` was added, just to be on the safe side.

grails-converters/src/main/groovy/org/grails/web/converters/marshaller/json/GenericJavaBeanMarshaller.java
jdaugherty: I'm honestly not sure in this case.  I'm hoping we can discuss in the weekly - instead of fixing this, why not force people to make those inner classes public?

sbglasius: @jdaugherty I don't feel we reached an agreement on the weekly? 

And with regards to a case, where access is widened, is that even a problem. Every developer can do so them self, using the exact same techniques. 

We could also easily go for the "not mutated" case.

IMO Grails does not have to become the framework where everything has to be according to standards.

bkoehm: I don't see much downside to going the "not mutated" case but I acknowledge the concern may be performance, as you put it: "Cost is one extra reflective lookup per property per marshal."  I think this is an acceptable cost, but it's easy for me to say that as I don't have visibility into that potential cost for other people.  I will accept that low cost.  I'm a thumbs up on the "not mutated" case suggestion.

jdaugherty:
I question if we should instead just force people to update their code to make those inner classes public.  @sbglasius I think you said the plugin itself didn't do this, can't we fix that instead?  Otherwise, we're encouraging bad habbits as newer versions of groovy are released.  (I am ultimately ok with this change, but I'm playing devil's advocate here to ensure we're not introducing something that should just be a note in the upgrade guide).

I'm also curious what @matrei thinks on this.

bkoehm:
> I question if we should instead just force people to update their code to make those inner classes public. @sbglasius I think you said the plugin itself didn't do this, can't we fix that instead? Otherwise, we're encouraging bad habbits as newer versions of groovy are released. (I am ultimately ok with this change, but I'm playing devil's advocate here to ensure we're not introducing something that should just be a note in the upgrade guide).

I was under the impression that more than just inner classes are in play, such as anonymous classes?

sbglasius:
@jdaugherty What @bkoehm is a good point. It's not just non-public classes that's the issue, also anonymous classes. Does that change your mind?

matrei (approved):
## AI Review

### Overview

Fixes two production-visible bugs in the four JSON/XML bean marshallers under `grails-converters` (#16294, #16295). The problem shapes are extremely common — Spring Security–style anonymous `UserDetails` implementations, any `Serializable` bean with `serialVersionUID` — so this is a real quality-of-life fix.

Three concerns are addressed:

1. **Non-public read methods** — `readMethod.invoke(o, …)` failed with `IllegalAccessException` when the declaring class wasn't accessible from another package, even if the method itself was public. Resolved by walking to the interface method via `ClassUtils.getInterfaceMethodIfPossible`, and falling back to reflectively widening a **fresh copy** of the method when no interface match exists.
2. **`Field.canAccess(o)` throws on static fields with a non-null `obj`** — the JDK contract explicitly throws `IllegalArgumentException` in that case, so any bean declaring `private static final long serialVersionUID` blew up. Fixed by short-circuiting the modifier checks before `canAccess`.
3. **Groovy synthetic capture fields** — anonymous inner classes compile enclosing variables into public synthetic `Reference` fields. Once the field loops can read them, they emitted as duplicate keys with empty-object values. All four marshallers now filter `field.isSynthetic()`.

All four marshallers get symmetric treatment. New specs (`NonPublicClassMarshallingSpec` for JSON and XML) plus a dynamic-Groovy fixture cover the paths, and each spec has a dedicated assertion that the shared descriptor cache is not mutated.

### Correctness

- **Interface-method resolution**: `ClassUtils.getInterfaceMethodIfPossible(readMethod, clazz)` never returns null (falls back to the original), and virtual dispatch on `invoke(o, …)` still calls the overridden concrete method. This is exactly how Spring resolves inaccessible-class calls internally.
- **Static-field short-circuit**: `!Modifier.isStatic(modifiers)` is evaluated before `field.canAccess(o)` in each marshaller, so the `canAccess` exception on static fields is avoided.
- **Synthetic filter**: necessary now that field access is widened in the Groovy marshallers, otherwise Groovy's `Reference` capture fields would leak. The XML spec explicitly asserts `xml.count('<name>') == 1`, which locks this in.
- **Cache non-mutation** (the important one for this codebase, given how many callers hit `BeanUtils.getPropertyDescriptors`):
  ```java
  Method invokableMethod = ClassUtils.getInterfaceMethodIfPossible(readMethod, clazz);
  if (!invokableMethod.canAccess(o)) {
      // Widen a private copy, so the flag never leaks into the shared descriptor cache
      invokableMethod = invokableMethod.getDeclaringClass().getDeclaredMethod(invokableMethod.getName());
      invokableMethod.setAccessible(true);
  }
  ```
  `Class.getDeclaredMethod` allocates a new `Method` per call (via `Method.copy()` in the JDK), so `setAccessible(true)` on the copy cannot flow back to Spring's `CachedIntrospectionResults`. The dedicated spec method that reads the cached descriptor back after marshalling and asserts the flag is still down is the right way to lock this in.
- **Fields don't need the same treatment**: `Class.getDeclaredFields()` already returns fresh copies each call, so the `ReflectionUtils.makeAccessible(field)` calls in the two `GroovyBeanMarshaller`s can't leak. Correctly identified and called out in the commit message.
- **Existing safeguards preserved**: `PersistenceMethod` / `ControllerMethod` skips, `metaClass` / `class` filtering, and GORM `attached`/`errors` filtering are untouched.

### Suggestions

1. **Java vs Groovy field-access asymmetry** — worth explicitly deciding on:
   - `GroovyBeanMarshaller` (both): drops the `field.canAccess(o)` guard and unconditionally calls `ReflectionUtils.makeAccessible(field)` before `field.get(o)`.
   - `GenericJavaBeanMarshaller` (both): keeps `field.canAccess(o)` and does **not** widen.

   Consequence: a Java anonymous class with a public field on a non-public declaring class will have that field silently skipped, while its Groovy equivalent (`anonymousPersonWithPublicField`) is emitted. If that asymmetry is intentional, fine — but the tests don't cover the Java-side variant, so it's easy to regress. Consider either aligning behavior (widen on both) or adding a test that asserts the current Java behavior on purpose.

2. **Duplicated widening block across four files** — the same five lines and the same comment appear in all four marshallers. Consider a small helper to keep the intent centralized:

   ```java
   private static Method resolveInvokable(Method readMethod, Class<?> targetClass, Object target) throws NoSuchMethodException {
       Method invokable = ClassUtils.getInterfaceMethodIfPossible(readMethod, targetClass);
       if (invokable.canAccess(target)) return invokable;
       Method copy = invokable.getDeclaringClass().getDeclaredMethod(invokable.getName());
       copy.setAccessible(true);
       return copy;
   }
   ```

   Even placed as a private static in each file, it eliminates four copies of the comment. Optional.

3. **Prefer `readMethod.getParameterTypes()` over the implicit empty array** — `getDeclaredMethod(invokable.getName())` works today because JavaBean getters are no-arg, but the call is subtly ambiguous to a reader who doesn't know that invariant. `getDeclaredMethod(invokable.getName(), invokable.getParameterTypes())` reads as "get the exact same method" and is trivially more defensive.

4. **Leak-detection test only covers the standalone-Java path** — `JavaPersonFactory.standalonePerson('user')` is the only shape exercised. The Groovy standalone path (`PackagePrivateStandaloneGroovyBean`) goes through the same widening but its spec doesn't assert the same invariant. Both marshallers copy the same pattern, so this is defensible, but a matching Groovy-side assertion would prove both marshallers preserve the invariant symmetrically.

5. **`NoSuchMethodException` propagation** — the new `getDeclaredMethod` call adds a checked exception into the try body. Fine because the outer `catch (Exception e)` wraps it as `ConverterException`. A one-liner comment noting "cannot fail: `invokable` came from this class's declared methods" would help future readers who wonder whether this is a real failure surface.

6. **Nit — variable name** — `invokableMethod` reads oddly; `invokable` or just `method` would be cleaner.

### Test coverage

Strong. Coverage matrix:

| Shape                                        | Groovy static | Groovy dynamic | Java |
|----------------------------------------------|:-:|:-:|:-:|
| Anonymous impl of public interface           | ✅ | ✅ | ✅ |
| Anonymous with public field                  | ✅ | — | — |
| Package-private impl of public interface     | ✅ | — | ✅ |
| Package-private standalone (no interface)    | ✅ | — | ✅ |
| Serializable + static fields                 | — | — | ✅ |
| Composite map (reproducer)                   | ✅ | — | — |
| Shared descriptor cache not mutated          | — | — | ✅ |

Optional additions: a Java anonymous class with a public field (to lock in the intentional asymmetry, if kept), a Groovy-side cache-non-mutation assertion, and a `Serializable` Groovy bean.

### Risk

Low. Changes are localized to four small marshallers. On the happy path (public class, public read method) the only added cost is one `ClassUtils.getInterfaceMethodIfPossible` cache lookup and one `canAccess` check — both cheap. The widening branch runs only when access is actually denied, and pays exactly one `getDeclaredMethod` per property in that case. The synthetic-field filter is defensive and can only remove keys that were themselves produced by the widened field access.

### Verdict

**LGTM.** Well-scoped fix with a thorough commit message and matching test coverage. The Java-vs-Groovy field-access asymmetry (point 1) is the only thing I'd want a decision on before merge — either accept it as intentional and add a Java-side test to lock it in, or align by widening in the Java `GenericJavaBeanMarshaller` too. The other suggestions are polish.

sbglasius:
@jdaugherty and @bkoehm do you want to weigh in on @matrei's bullet 1?

jdaugherty:
@sbglasius My original concern was we should force these to be made public (as in change your code, not the framework) and not do this PR.  Isn't that what bullet 1 is effectively identifying - that this is already the behavior of java based code?

bkoehm:
> @jdaugherty and @bkoehm do you want to weigh in on @matrei's bullet 1?

If we're talking about the AI-generated "Interface-method resolution" bullet, I can't make much of it.

But I'll say, I'm nearly ready to approve this, but with a couple points.  One, by approving, I'm not looking to circumvent any opinions of @jdaugherty .  Two, I would like to know if you (@sbglasius) had any opinion on my response in this comment:
https://github.com/apache/grails-core/pull/16296#discussion_r3925544370
My only outstanding question on this PR is whether the "not mutated" case is the way to go and I was waiting for a further discussion on that before approving.

sbglasius:
@jdaugherty Pushed `e9932f371f` with what we agreed on Slack:

- **The warning.** Every non-public bean class is reported once, through its own logger (`org.apache.grails.common.reflect.ReflectionUtils`), so one logging line silences it. The bookkeeping runs *before* the log level is consulted, so "once" doesn't depend on how logging is configured, and it holds class names rather than `Class` references so no class loader is pinned. JDK/Groovy/Spring classes aren't reported — nobody reading the log can declare those public. `warnOnNonPublicClass` is the seam to delete if Groovy stops compiling anonymous classes as non-public.
- **One shared utility, not a fifth.** The widening block was copy-pasted in all four marshallers; it now lives in `org.apache.grails.common.reflect.ReflectionUtils` in grails-common, which is already on the compile classpath of every module that could need it. I left the GORM one alone — `org.grails.datastore.mapping.reflect.ReflectionUtils` is published, public since 1.0, with 8 call sites — so the full move you suggested is #16326 rather than growing this PR into a cross-module API change.
- **Java and Groovy aligned** (Mattias's open point): the Java marshallers gated their field loop on `canAccess`, so a public field on a non-public class was silently dropped where the Groovy ones emitted it. Both now widen a field copy, and skip rather than fail when a field genuinely can't be widened (a class in a named module that doesn't open its package).
- **Upgrade guide §54**, since that field change alters response payloads and the warning needs a documented off switch.

On forcing the fix instead of handling it: for the shape in the original report there's no modifier to add — an anonymous class can't be declared public — which is what the warning is for, pointing people at the named-class fix.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -3097,3 +3097,70 @@ grails {
 imported whether or not they are present — a star import of an absent package contributes no classes and is
 not an error in Groovy, so the probe changed nothing it could observe. An application that relied on the
 import being *omitted* when the package was absent sees no difference in compiled output.
+
+==== 54. Non-Public Bean Classes Are Marshalled, and Reported Once
+
+`someObject as JSON` (and `as XML`) previously failed when the object's class was not `public` — an
```
jdaugherty: This paragraph reads like it was broken since upgrading to Grails 7 - but it's only broken as of upgrading to Groovy 5.  We should call out the Groovy 5 change that made this error, and explain that Grails adds this workaround (with a warning to update your code).

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
```diff
@@ -0,0 +1,196 @@
...
+    public static boolean warnOnNonPublicClass(@Nullable Class<?> clazz) {
+        if (clazz == null || Modifier.isPublic(clazz.getModifiers()) || !isApplicationClass(clazz)) {
+            return false;
+        }
+        if (WARNED_NON_PUBLIC_CLASSES.size() >= MAX_WARNED_CLASSES ||
+                !WARNED_NON_PUBLIC_CLASSES.add(clazz.getName())) {
+            return false;
+        }
+        if (LOG.isWarnEnabled()) {
+            LOG.warn("Class [{}] is not public, so its properties can only be read by widening access reflectively. " +
```
jdaugherty: The point of this warning is to get people to change their code so the class is public, so the text has to be accurate about why. As written it says the properties "can only be read by widening access reflectively", which is not true for the shape that opened #16294: the anonymous `UserDetails` resolves to the interface method in `resolveInvokableReadMethod` and is invoked like any other public method, nothing is widened. Since the trigger is the class shape rather than the mechanism, describe the shape:

```
Class [{}] is not public. Grails reads its properties through compatibility handling that may be withdrawn in a future major release. Declare it as a named public class so that it reads as a standard JavaBean, or register an ObjectMarshaller for it if the class is not yours. To silence this, set the log level of [{}] above WARN. (warned once per class)
```

The `ObjectMarshaller` clause matters because `NON_APPLICATION_PACKAGES` is a short list; a non-public bean from any other library gets reported to someone who cannot declare it public. The upgrade guide quotes this message verbatim in 54.2, so it needs the same edit.

grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+    void setup() {
+        ReflectionUtils.resetWarnedClasses()
+    }
+
+    void cleanup() {
+        ReflectionUtils.resetWarnedClasses()
+    }
+
+    private static Method readMethodOf(Object bean, String property) {
+        PropertyDescriptor descriptor = Introspector.getBeanInfo(bean.getClass()).propertyDescriptors
```
jdaugherty: `java.beans.Introspector` already resolves a read method to its publicly accessible declaration, so two of these features never reach the branch they are named for. I compiled the fixtures and compared what each introspector hands back:

| Fixture | `Introspector` read method | `BeanUtils.getPropertyDescriptors` read method |
|---|---|---|
| `anonymousThing` | `PublicThing.getName`, accessible | the anonymous class, not accessible |
| `covariantThing` | `PublicCovariantBase.getValue`, accessible | `CovariantThing.getValue`, not accessible |
| `standaloneThing` | `StandaloneThing.getName`, not accessible | same |

So in "declared by a public interface needs no widening" and "covariant read method resolves to the override rather than the bridge", `resolveInvokableReadMethod` returns its argument unchanged and both would pass with `return readMethod`. Only the standalone feature exercises widening, and nothing exercises the `getDeclaredMethod` override-vs-bridge selection at all, because the covariant shape is not in the converters specs either.

Please take the read methods from `BeanUtils.getPropertyDescriptors` here, which is what the marshallers use and which returns the declaring-class method (spring-beans is already on this module's classpath through spring-context), and add the covariant shape to one of the marshaller specs so the bridge selection is covered end to end. Codecov's 72% on this class lines up with the two vacuous features.

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
```diff
@@ -0,0 +1,196 @@
...
+     * instance shared through a descriptor cache.
+     *
+     * @param readMethod  the property's read method, typically from a {@code PropertyDescriptor}
+     * @param targetClass the class being read, used to resolve the interface method
+     * @param target      the instance being read, or {@code null} for a static read method
+     * @return a method that may be invoked on {@code target}
+     */
+    public static Method resolveInvokableReadMethod(Method readMethod, Class<?> targetClass, @Nullable Object target)
+            throws NoSuchMethodException {
+        Method invokable = ClassUtils.getInterfaceMethodIfPossible(readMethod, targetClass);
```
jdaugherty: `ClassUtils.getPubliclyAccessibleMethodIfPossible(readMethod, targetClass)` is in the spring-core we pin and does this plus the superclass walk: it returns the first equivalent method declared on a public type, interface or class. For `CovariantThing` (package-private, extends the public `PublicCovariantBase`) the interface-only resolver finds nothing and this method widens a copy; the publicly-accessible one returns `PublicCovariantBase.getValue`, `canAccess` is true, and `invoke` still dispatches to the override. I checked this with a probe against spring-core 7.0.9.

That confines the `setAccessible` path to getters with no public declaring type anywhere in the hierarchy, which is where the non-standard handling belongs, and it makes the override-vs-bridge `getDeclaredMethod` below rarer.

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
```diff
@@ -0,0 +1,196 @@
...
+        if (canAccess(invokable, target)) {
+            return invokable;
+        }
+        // getDeclaredMethod cannot fail here: invokable was resolved from this very class's methods.
+        Method widened = invokable.getDeclaringClass()
+                .getDeclaredMethod(invokable.getName(), invokable.getParameterTypes());
+        widened.setAccessible(true);
```
jdaugherty, line 101: `tryMakeReadable` below catches `InaccessibleObjectException` and skips the field, but this `setAccessible` is left to throw, so a getter on a non-public class in a named module that does not open its package still fails the whole conversion (the marshallers wrap it as `ConverterException`, the same as before this PR). Section 54.1 of the upgrade guide says fields in that situation are skipped rather than failing the conversion, which reads as if the conversion survives; with a getter in the same module it will not. Either skip the property here the same way, or narrow that sentence in the guide. Low priority since it is not a regression.

jdaugherty (commented):
I like the direction and I think the warning is an acceptable solution.  There's some minor improvements then we can merge ...

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
sbglasius: Done in `60598510cf`. I probed it against spring-core 7.0.9 with the built fixtures first, and it behaves as you said:

| fixture | `getInterfaceMethodIfPossible` | `getPubliclyAccessibleMethodIfPossible` |
|---|---|---|
| anonymous impl of public interface | `PublicThing.getName` ✅ | `PublicThing.getName` ✅ |
| package-private covariant override of a public superclass | `CovariantThing.getValue` ❌ | `PublicCovariantBase.getValue` ✅ |
| package-private, no interface | ❌ | ❌ |

So the covariant shape no longer calls `setAccessible` at all, and `invoke` still dispatches to the override. Widening is now confined to a getter with no public type anywhere in its hierarchy, which is the only shape that is genuinely non-standard.

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
sbglasius: You're right that the old text was inaccurate, and for exactly the reason you give — the anonymous `UserDetails` resolves to the interface method and nothing is widened. I've taken your wording as-is, and `54.2` now quotes it verbatim.

Since the guide and the code can drift apart silently, I also added a check to the verification step that extracts the `LOG.warn` string from the class, normalises the two `{}` placeholders, and diffs it against the quoted block in the guide. It matches.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
sbglasius: Rewritten in `60598510cf`. The section now opens on the Groovy codegen change — Groovy 4 stamped `ACC_PUBLIC` on anonymous inner classes, Groovy 5 does not — says explicitly that nothing in the application changed, only the modifier the compiler emits, and then says Grails adds compatibility handling and reports each class so the code can be corrected.

I also added a NOTE that ordinary Groovy code is unaffected: property access and method calls through the metaclass still work on such an object from another package, as does the JDK `Introspector`; only raw reflection against the method as the non-public class declares it fails, which is what bean marshalling was doing. Verified both, see my top-level comment.

The `serialVersionUID` fix (#16295) kept its own paragraph, since that one is unrelated to Groovy 5 and hits practically every `Serializable` bean.

grails-common/src/test/groovy/org/apache/grails/common/reflect/ReflectionUtilsSpec.groovy
sbglasius: Confirmed — I reproduced your table before changing anything, and it is exactly right. `Introspector` hands back `PublicThing.getName` and `PublicCovariantBase.getValue`, both already accessible, so those two features asserted nothing about the code they were named for.

Fixed in `60598510cf`:

- `readMethodOf` now uses `BeanUtils.getPropertyDescriptors`, with a comment saying why, so the next person does not helpfully switch it back to `Introspector`.
- The covariant feature is re-pointed: with `getPubliclyAccessibleMethodIfPossible` that shape resolves to the public superclass, so it now asserts *that* rather than implying widening.
- New fixture `HiddenCovariantThing` — a package-private covariant override of a package-private base, so there is no public type anywhere. That is the only remaining shape that reaches the `getDeclaredMethod` override-versus-bridge choice, and it was uncovered.
- Both marshaller specs now marshal that shape end to end; the XML one asserts `xml.count('<tag>') == 1` so a bridge cannot quietly add a second node.

The check that these are no longer vacuous: stubbing `resolveInvokableReadMethod` to `return readMethod` now fails **four** features. Before this commit it failed two.

grails-common/src/main/groovy/org/apache/grails/common/reflect/ReflectionUtils.java
sbglasius: Took your second option — narrowed the guide rather than changing the behaviour, in `60598510cf`. 54.1 now reads:

> A `public` *field* that cannot be read at all — because its class is in a named module that does not open its package — is skipped, and the rest of the object still converts. A *getter* in that same situation is not skipped: the conversion fails, as it did before Grails 8.

The reasoning for keeping them different: a field in that situation was already being skipped by the old `canAccess` gate, so skipping preserves its behaviour, whereas a getter has always failed — and silently dropping a declared property from a response body seems worse than an error, since the client gets a payload missing a field with no signal. Happy to make it fail-soft instead if you'd rather have consistency.

sbglasius:
@bkoehm @jdaugherty — `60598510cf` addresses James's review (five inline replies above). Two things for Brian, and then the data behind the "is this Grails-only behaviour" question, since you are both circling it.

**Brian, on your r3925544370 — the "not mutated" case is already in.** It landed in `2e08c1fc15`, before your comment; I think it just got lost in the noise. Both specs carry a guard feature (*marshalling does not widen access on the shared property descriptor cache*) that reads the cached descriptor back after marshalling and fails if the flag was flipped — I verified it bites by temporarily restoring `makeAccessible`. So nothing is pending from you there.

The cost is also lower than you generously accepted. It is not one extra lookup per property per marshal — only the path that would otherwise mutate pays it, and after `60598510cf` that path is narrower again: a getter now has to have no public type *anywhere* in its hierarchy before anything is widened.

**Matrei's bullet 1, in plain terms:** to read `getUsername()` on an anonymous `UserDetails`, the marshaller walks to the same method as declared by the public `UserDetails` interface and invokes that. Java dispatches it to the anonymous class's override anyway, so the value is right and nothing is modified. Only when there is no publicly reachable declaration does it take a throwaway copy of the method, widen the copy, and invoke that — leaving the shared one untouched. That is the whole mechanism.

**On whether this is Grails inventing behaviour.** I probed it rather than argue it. Same source, two compilers:

| | Groovy 4.0.33 | Groovy 5.1.1 |
|---|---|---|
| anonymous class is public | **yes** | **no** |
| `thing.name` via the metaclass | works | **works** |
| `thing.getName()` via the metaclass | works | **works** |
| raw `Method.invoke` | works | **IllegalAccessException** |
| JDK `Introspector` | works | **works** — resolves to the interface |

So ordinary Groovy is not broken by the change; only raw `Method.invoke` against the declaring class is, which is what our marshallers were doing. Groovy's own MOP, the JDK's `Introspector`, and Spring's `BeanWrapperImpl$BeanPropertyHandler` (which calls `ReflectionUtils.makeAccessible` on the cached read method — confirmed in the 7.0.9 bytecode) all handle this case. This PR does the same thing, more conservatively than Spring does, since we widen a copy rather than the cached member.

James, that is also the answer to your question about bullet 1: **Java-based code is not already working.** Pre-fix, the plain-Java fixtures in this PR — an anonymous implementation, a package-private class, and a package-private class with no interface — all threw `IllegalAccessException` from `GenericJavaBeanMarshaller`, because javac has always compiled anonymous classes as non-public. Bullet 1 was only ever about the *field* loop, where Java skipped a public field that Groovy emitted; this PR aligned that. So Groovy 5 moved *towards* javac here rather than away from it, and the getter path was latently broken for Java beans on every branch.

The warning stays as the removal seam you asked for.

bkoehm (approved):
Thank you!

PR #16299, opened 2026-09-02
Keep GORM entities on the DevTools restart classloader

### LINKED ISSUE #16287: spring devtools & grails 8 do not work
### Expected Behavior

On Bootstrap phase, the application will create an user with admin permissions and then, start successfully.

```
class BootStrap {
    def springSecurityService

    def init = { servletContext ->
        environments {
            development {
                initData()
            }
        }
    }

    void initData() {
        User.withTransaction {
            if (Role.count() == 0) {
                def password = UUID.randomUUID().toString()
                def adminRole = new Role(authority: 'ROLE_ADMIN').save(failOnError: true)

                def admin = new User(
                        username: 'admin',
                        password: springSecurityService.encodePassword(password)).save(failOnError: true)

                UserRole.create(admin, adminRole, true)
                log.warn "Generated admin credentials: admin / $password"
            }
        }
    }

    def destroy = {
    }

}
```

### Actual Behaviour

On bootstrap phase, the application raises this error:

`java.lang.IllegalArgumentException: Not an entity: class com.example.Role`

Disabling spring boot dev tools the application starts successfully.

Claude Code summary:

java.lang.IllegalArgumentException: Not an entity: class com.example.Role

  The IllegalArgumentException: Not an entity: class com.example.Role wasn't a problem in your Role.groovy/User.groovy/BootStrap.groovy — those are all correctly written. 
It's a compatibility bug between Spring Boot  DevTools' restart classloader and GORM/Hibernate5 on this exact stack: Grails 8.0.0-M5 + Hibernate ORM 5.6.15 + Spring Boot 4.1.0.

  How I confirmed it:
  - Reproduced the crash reliably via bootRun against your dev MySQL.
  - Proved it wasn't Role-specific — swapping in User.count() first hit the identical "Not an entity" error, and it also failed from a live HTTP request long after startup completed, ruling out a boot-order race.
  - Isolated it to Criteria-based GORM calls specifically: Role.count()/Role.list() failed, but Role.get() and raw HQL worked — and .save() threw a field/getter reflection IllegalArgumentException, the textbook
    signature of the same class name being loaded by two different ClassLoaders.
  - Removing spring-boot-devtools from the classpath made count(), list(), and save() all work immediately (thread name also switched from restartedMain to plain main, confirming DevTools' restart classloader was the
    split).
  - Re-verified end-to-end with a clean DB: BootStrap now logs Generated admin credentials: admin / <uuid> with zero errors.

  What I changed:
  - build.gradle — commented out developmentOnly "org.springframework.boot:spring-boot-devtools" with an explanatory comment.
  - Everything else (BootStrap, application.groovy, domain classes) is back to exactly how you had it — my diagnostic edits (temp DiagController, debug logging, a scoped /diag security rule) were all reverted.
  - Cleaned up the one leftover test row I'd accidentally committed to the role table during diagnosis.

  Tradeoff: you lose DevTools' auto-restart/LiveReload during development until this Grails milestone/Hibernate5 combo is fixed upstream (worth watching Grails' issue tracker, since 8.0.0-M5 + Spring Boot 4 is very
  fresh). 

### Steps To Reproduce

Fresh Grails 8.0.0-M5 project created from start.grails.org.

Features:
- spring security core
- database migration
- spring boot devtools as development relaoding

### Environment Information

- Operating System: Windows
- Java JDK: 21

### Example Application

_No response_

### Version

Grails 8.0.0-M5

jamesfredley:
## Description

Fixes https://github.com/apache/grails-core/issues/16287.

This is a real compatibility bug, not user error. With Spring Boot DevTools restart active, Grails 8 loads application domain classes in DevTools' `RestartClassLoader` while GORM and Hibernate remain on the base loader. Hibernate's JPA metamodel keys entities by `Class` identity, so GORM Criteria calls such as `count()` / `list()` and `save()` fail with `IllegalArgumentException: Not an entity` while `get()` and HQL still work.

### What was actually wrong on Hibernate 5

The Hibernate 5 connection-source factory looked up a Spring bean named `default` (`ConnectionSource.DEFAULT`). The DataSource bean is `dataSource`, so `setApplicationContext` never ran and `CLASSLOADERS` stayed on `connectionSource.getClass().getClassLoader()` (the base loader). Hibernate 7 already computed `dataSource` / `dataSource_<name>`.

Hibernate then re-resolves entity `Class` objects by name during SessionFactory construction (`ReflectHelper.classForName` uses the thread context class loader). Setting `AvailableSettings.CLASSLOADERS` is not enough if TCCL is still the base loader at bootstrap.

### What this PR does

- Hibernate 5 factory looks up `dataSource` / `dataSource_<name>` and sets `dataSourceName` before `setApplicationContext`, so the application-context class loader is used. Named sources inject `dataSource_<name>`, not the default DataSource.
- Both Hibernate 5 and 7 call `DevToolsClassLoaders.preferRestartClassLoader(...)` at SessionFactory build time and temporarily set TCCL around `super.buildSessionFactory`.
- `CLASSLOADERS` is left unset when the application context loader is null and DevTools is not active, preserving the old Hibernate 7 fallback.
- The helper matches DevTools' `RestartClassLoader` by FQCN first (simple-name fallback for tests/shaded copies), keeps a descendant-loader guard, and deprecates `resolve()` in favor of `preferRestartClassLoader()`.
- Framework jars stay on the base loader; moving GORM/Hibernate onto the restart loader would split Grails types such as `MappingContext` across loaders.

## Contributor Checklist

Please review the following checklist before submitting your pull request. Pull requests that do not meet these requirements may be closed without review.

### Issue and Scope

- [x] This PR is linked to an existing issue that has been **acknowledged or approved** by the project team. If no approved issue exists, please give background on why this change is necessary.  Tickets are preferred for release change log history.
- [x] This PR contains a **single, focused change**. Unrelated changes should be submitted as separate pull requests.

### Code Quality

### Licensing and Attribution

### Documentation

- [ ] If this PR adds a new feature, I have updated the **What's New** section of the Grails Guide.
- [ ] If this PR introduces breaking changes or changes that require user action during an upgrade, I have updated the **Upgrade Notes** for the corresponding version in the Grails Guide.

## Generative AI attribution

Generative AI tooling (Cursor Grok 4.6 with GPT review) was used to draft the implementation, tests, and documentation. The change was reviewed, tested in the affected modules (`grails-datastore-core`, `grails-data-hibernate5-core`, `grails-data-hibernate7-core`), and edited before submission.


grails-doc/src/en/guide/gettingStarted/developmentReloading.adoc
```diff
@@ -25,6 +25,8 @@ Spring Boot Developer Tools is a feature of Spring Boot that provides automatic
 
 For larger applications, you may need to adjust the default settings for optimal performance.  This works well until your application becomes very large, at which point restarts may take longer or fail.
 
+When Spring Boot Developer Tools is on the classpath, Grails bootstraps Hibernate with DevTools' restart class loader. That keeps domain class identity consistent with Hibernate's metamodel without moving GORM or Grails jars onto the restart loader. Without that alignment, GORM calls such as `Role.count()` and `save()` can fail with `IllegalArgumentException: Not an entity` while `get()` and HQL still work.
```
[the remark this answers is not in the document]
jamesfredley: Valid catch. The Hibernate bootstrap only prefers the restart class loader when restart is actually active (not merely because DevTools is on the classpath). The docs now say that.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/DevToolsClassLoaders.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private DevToolsClassLoaders() {
+    }
+
+    /**
+     * @param classLoader the loader to inspect, possibly {@code null}
+     * @return {@code true} when {@code classLoader} is Spring Boot DevTools'
+     * {@code RestartClassLoader}
+     */
+    public static boolean isRestartClassLoader(ClassLoader classLoader) {
```
jdaugherty, line 48: This heuristic was tolerable as a private inline check, but promoting it to a public static method in a published module makes it API surface that has to keep working.

devtools' loader is always `org.springframework.boot.devtools.restart.classloader.RestartClassLoader`, so matching the FQCN first (with the simple name as a fallback for shaded/relocated cases) would be both tighter and self-documenting. As written, a user class coincidentally named `RestartClassLoader` — or `restartclassloader`, given the `equalsIgnoreCase` — matches.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/DevToolsClassLoaders.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Prefer the thread context class loader when it is DevTools'
+     * {@code RestartClassLoader}; otherwise return {@code fallback}, or this
+     * class's loader when {@code fallback} is {@code null}.
+     *
+     * @param fallback the loader to use when DevTools is not active
+     * @return a non-null class loader
+     */
+    @SuppressWarnings("PMD.UseProperClassLoader")
+    public static ClassLoader resolve(ClassLoader fallback) {
```
jdaugherty: This prefers the TCCL unconditionally, which discards a strictly better loader when `fallback` is a *descendant* of the restart loader — the child can see everything the restart loader sees plus its own classes, and returning the parent loses those.

A cheap guard is to keep `fallback` when the restart loader is already in its parent chain. That doesn't weaken the intended fix: in the devtools scenario the base loader is the restart loader's *parent*, not a descendant, so the guard still returns the restart loader there.

Also worth a one-line comment justifying the `@SuppressWarnings("PMD.UseProperClassLoader")` — the reason (`DevToolsClassLoaders.class.getClassLoader()` is the deliberate last-resort fallback, not a TCCL oversight) isn't obvious from the code.

Nit: `resolve(x)` doesn't read well at the call sites. `preferRestartClassLoader(x)` says what it actually does.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
```diff
@@ -153,10 +154,8 @@ public void setApplicationContext(@Nullable ApplicationContext applicationContex
             properties.put("hibernate.enhancer.enableLazyInitialization", FALSE_LITERAL);
             properties.put("hibernate.enhancer.enableDirtyTracking", FALSE_LITERAL);
             properties.put("hibernate.enhancer.enableAssociationManagement", FALSE_LITERAL);
-            ClassLoader classLoader = applicationContext.getClassLoader();
-            if (classLoader != null) {
-                properties.put(AvailableSettings.CLASSLOADERS, classLoader);
-            }
+            properties.put(AvailableSettings.CLASSLOADERS,
+                    DevToolsClassLoaders.resolve(applicationContext.getClassLoader()));
```
jdaugherty: Behavior change worth confirming is intentional: the old code only set `CLASSLOADERS` when `applicationContext.getClassLoader()` was non-null. `resolve` never returns null, so when the app context loader is null this now writes `DevToolsClassLoaders.class.getClassLoader()` (grails-datastore-core's loader) instead of leaving the key absent and letting `buildSessionFactory` fall back to `HibernateMappingContextConfiguration.class.getClassLoader()` (hibernate7-core's loader).

Same result on a flat classpath, but it's an unintended semantic change. Preserving the null check, or having `resolve` be null-preserving, would keep the old fallback.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
```diff
@@ -292,13 +283,9 @@ public SessionFactory buildSessionFactory() throws HibernateException {
...
-        ClassLoader appClassLoader;
-
-        if (classLoaderObject instanceof ClassLoader) {
-            appClassLoader = (ClassLoader) classLoaderObject;
-        } else {
-            appClassLoader = getClass().getClassLoader();
-        }
+        ClassLoader storedClassLoader = classLoaderObject instanceof ClassLoader ?
+                (ClassLoader) classLoaderObject : getClass().getClassLoader();
+        ClassLoader appClassLoader = DevToolsClassLoaders.resolve(storedClassLoader);
```
jdaugherty: Two things on this third `resolve` call:

1. **It silently overrides an explicit setting.** Both setters already resolve, and they run on the same thread immediately before this, so this can only differ when `CLASSLOADERS` was populated by something else — a `configClass` subclass, `configuration.addProperties(...)`, or `HibernateMappingContextSessionFactoryBean.newConfiguration()`. In those cases a deliberate loader choice gets replaced by whatever the TCCL happens to be. Either drop it or document why re-deriving is intended.

2. **It's untested.** This is the one hunk that could plausibly change the resolved loader relative to the setters, and no spec exercises it. The existing `"buildSessionFactory handles classloader object when it is a ClassLoader"` test never calls `buildSessionFactory` — it puts a value into the properties and asserts it's still there. Per CLAUDE.md rules 11/13 this branch needs real coverage.

grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
```diff
@@ -239,14 +236,9 @@ public SessionFactory buildSessionFactory() throws HibernateException {
...
-
-        if (classLoaderObject instanceof ClassLoader) {
-            appClassLoader = (ClassLoader) classLoaderObject;
-        }
-        else {
-            appClassLoader = getClass().getClassLoader();
-        }
+        ClassLoader storedClassLoader = classLoaderObject instanceof ClassLoader ?
+                (ClassLoader) classLoaderObject : getClass().getClassLoader();
+        ClassLoader appClassLoader = DevToolsClassLoaders.resolve(storedClassLoader);
```
jdaugherty: Same as the hibernate7 counterpart: this re-resolve can only differ from what the setter stored when `CLASSLOADERS` came from somewhere other than the two setters, in which case it discards an explicit choice — and it has no test coverage.

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfigurationSpec.groovy
```diff
@@ -0,0 +1,114 @@
...
+            getName() >> "secondary"
+            getSource() >> ds
+        }
+
+        when:
+        config.setDataSourceConnectionSource(connSrc)
+
+        then:
+        config.dataSourceName == "secondary"
+        config.getProperties().containsKey(AvailableSettings.CLASSLOADERS)
```
jdaugherty: This assertion doesn't test what the feature name claims. `containsKey` plus `!= null` passes regardless of which loader was stored, including a wrong one — `resolve` could be changed to always return `DevToolsClassLoaders.class.getClassLoader()` and this would still be green.

Assert the actual loader: `config.getProperties().get(AvailableSettings.CLASSLOADERS).is(connSrc.getClass().getClassLoader())`.

grails-doc/src/en/guide/gettingStarted/developmentReloading.adoc
```diff
@@ -25,6 +25,8 @@ Spring Boot Developer Tools is a feature of Spring Boot that provides automatic
 
 For larger applications, you may need to adjust the default settings for optimal performance.  This works well until your application becomes very large, at which point restarts may take longer or fail.
 
+When Spring Boot Developer Tools restart is active, Grails bootstraps Hibernate with the restart class loader. That keeps domain class identity consistent with Hibernate's metamodel without moving GORM or Grails jars onto the restart loader. Without that alignment, GORM calls such as `Role.count()` and `save()` can fail with `IllegalArgumentException: Not an entity` while `get()` and HQL still work.
```
jdaugherty: The second sentence documents the symptom of a bug that is now fixed, and uses `Role.count()` — a class the reader has no context for in this section. That reads as release-note material rather than guide content.

Suggest keeping only the first sentence, or dropping the paragraph and putting it in the changelog.

jdaugherty (commented):
AI did the below review and seems to think the hibernate 5 is still broken.  Did you test to confirm these changes fix it?  See below for it's review.

Clean consolidation of the duplicated `RestartClassLoader` sniffing, and the new helper is well covered. But tracing it against the reported reproduction, I don't think it changes behavior on the path that actually fails.

## The change appears to be a no-op on the reporter's path

The issue is Grails 8.0.0-M5 from start.grails.org with Hibernate 5.6.15. Forge's default is `GormImpl.DEFAULT_OPTION = HIBERNATE5` (`GormImpl.java:31`), so the failing path is hibernate5:

1. `grails-data-hibernate5/core/.../connections/HibernateConnectionSourceFactory.java:142` branches on `applicationContext.containsBean(dataSourceConnectionSource.getName())`. `ConnectionSource.DEFAULT` is `"default"` (`ConnectionSource.java:34`) and there is no bean named `default` — the DataSource bean is `dataSource`. So this always takes the `else` branch, `setDataSourceConnectionSource(...)`. (hibernate7 computes the real bean name `"dataSource"` and therefore takes the `setApplicationContext` branch — the two modules genuinely diverge here.)

2. `setDataSourceConnectionSource` has preferred a restart TCCL since `56df99d587` ("Support DevTools RestartClassLoader. Fixes grails/grails-core#11159"). That is exactly the code this PR replaces with an equivalent `DevToolsClassLoaders.resolve(...)` call.

3. `create()` (`HibernateConnectionSourceFactory.java:107-109`) calls `buildConfiguration(...)` and then `configuration.buildSessionFactory()` back to back **on the same thread**. So the new `resolve(storedClassLoader)` in `buildSessionFactory` reads the same `Thread.currentThread().getContextClassLoader()` the setter just read, and can only return the same loader.

So on the hibernate5 default path the resolved loader is identical before and after this PR. Either the failure has a root cause other than `AvailableSettings.CLASSLOADERS`, or the TCCL was not a `RestartClassLoader` at connection-source creation — and in that second case this PR doesn't help either, for the same-thread reason above.

Earlier investigation of this issue confirmed via `identityHashCode`/loader diagnostics that the JPA metamodel held `AppClassLoader` copies while GORM held restart-loader copies. Since the restart-loader preference was already active on that path, something else is re-resolving those classes by name.

The one genuine behavior change here is hibernate7's `setApplicationContext`, which now prefers a restart TCCL over `applicationContext.getClassLoader()`. Under devtools those normally agree, since `AbstractApplicationContext` extends `DefaultResourceLoader` and captures the TCCL at construction on `restartedMain`.

**Ask:** how was the fix verified? Ideally a generated app with `gorm-hibernate5` + spring-security-core + devtools that failed before and boots after. The checklist also has "verified that all existing tests pass" unchecked. If verification was against hibernate7, the "Fixes #16287" claim should probably be re-scoped, since the reporter's stack is hibernate5.

CI is green apart from the known-flaky `UserControllerSpec > User list` (#16030).

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/DevToolsClassLoaders.java
jamesfredley: Agreed. `isRestartClassLoader` now walks the class hierarchy matching `org.springframework.boot.devtools.restart.classloader.RestartClassLoader` first. The exact simple-name `RestartClassLoader` match is only a fallback for tests and shaded/relocated copies; case-insensitive matching is gone.

grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/DevToolsClassLoaders.java
jamesfredley: Done. `preferRestartClassLoader` keeps `fallback` when it is the restart loader or a descendant of it. The PMD suppression has a one-line reason, and `resolve()` is deprecated and delegates to `preferRestartClassLoader()`.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
jamesfredley: Valid. `setApplicationContext` now writes `CLASSLOADERS` only when the application context loader is non-null or the TCCL is a `RestartClassLoader`. That preserves the old Hibernate 7 fallback to `HibernateMappingContextConfiguration.class.getClassLoader()` in `buildSessionFactory`. Covered by tests for both the null-loader and restart-TCCL cases.

grails-data-hibernate7/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
jamesfredley: Kept on purpose. `addProperties()` / a custom `configClass` can overwrite `CLASSLOADERS` after the setters, and Hibernate re-resolves entity `Class` objects via `ReflectHelper.classForName` (TCCL) during SessionFactory construction. `resolveSessionFactoryClassLoader()` is now tested, and `buildSessionFactory` sets TCCL around `super.buildSessionFactory` and restores it in `finally`.

grails-data-hibernate5/core/src/main/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfiguration.java
jamesfredley: Same as the Hibernate 7 counterpart: the re-resolve is intentional because `addProperties()` can overwrite `CLASSLOADERS` and Hibernate re-resolves entity classes from TCCL. `resolveSessionFactoryClassLoader()` is now tested, and `buildSessionFactory` wraps TCCL around `super.buildSessionFactory`.

grails-data-hibernate5/core/src/test/groovy/org/grails/orm/hibernate/cfg/HibernateMappingContextConfigurationSpec.groovy
jamesfredley: Fixed. The spec now asserts identity: `config.getProperties().get(AvailableSettings.CLASSLOADERS).is(connSrc.getClass().getClassLoader())`.

grails-doc/src/en/guide/gettingStarted/developmentReloading.adoc
jamesfredley: Dropped the `Role.count()` symptom sentence. The guide now only says that when DevTools restart is active, Grails bootstraps Hibernate with the restart class loader so domain class identity stays consistent with the metamodel.

jamesfredley:
Independent check of the review feedback (not taking the review text as given):

The Hibernate 5 factory **was** looking up a bean named `default` (`ConnectionSource.DEFAULT`). The DataSource bean is `dataSource`, so `setApplicationContext` never ran on the reporter's Hibernate 5 path. That claim was valid. `setDataSourceConnectionSource` already preferred a restart TCCL, so swapping `CLASSLOADERS` alone on that else-branch was equivalent. The real Hibernate 5 hole was taking the wrong setter, which left `CLASSLOADERS` on `connectionSource.getClass().getClassLoader()` (the base loader) whenever TCCL was not already a `RestartClassLoader`.

Hibernate also re-resolves entity `Class` objects by name during SessionFactory construction (`ReflectHelper.classForName` uses TCCL). That is why this follow-up sets TCCL around `super.buildSessionFactory`, not only `AvailableSettings.CLASSLOADERS`.

Updates in c8945fd6c3:

- Hibernate 5 factory looks up `dataSource` / `dataSource_<name>` and sets `dataSourceName` **before** `setApplicationContext`, so named sources inject `dataSource_secondary` rather than the default DataSource.
- Hibernate 5 and 7 use `preferRestartClassLoader`, re-resolve at SessionFactory build, and wrap TCCL around `super.buildSessionFactory` (restored in `finally`).
- `CLASSLOADERS` is left unset when the application context loader is null and DevTools is not active.
- Helper: FQCN match first, case-sensitive simple-name fallback, descendant-loader guard, `resolve()` deprecated.
- Tests for factory bean wiring, named DataSource injection, loader identity, null-loader fallback, and `resolveSessionFactoryClassLoader`.
- Docs: dropped the `Role.count()` symptom sentence.

Targeted tests passed in `grails-datastore-core`, `grails-data-hibernate5-core`, and `grails-data-hibernate7-core`. Module `codeStyle` is clean. Review threads from that pass are resolved.

I did not run a generated Forge app with `gorm-hibernate5` + spring-security-core + DevTools end-to-end in this follow-up; the factory bean-name miss and TCCL wrap are covered by unit tests against the actual bootstrap path.

jdaugherty:
@jamesfredley it looks like the tests aren't running on this PR b/c of the infrastructure unapproving the actions.

jdaugherty:
I opened https://github.com/apache/grails-core/pull/16324 to address all of my feedback on this review.

PR #16308, opened 2026-09-03
ci: pin every action to an ASF-approved SHA

sbglasius:
## Problem

Every workflow in this repo is currently failing **at startup**, before any job runs:

> The action `gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e` is not allowed in `apache/grails-core` because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: `1Password/load-secrets-action/configure@3a12b0ab…`, …

The `1Password/…` entries at the front of that message are just the alphabetically-first patterns in the ASF allowlist — they are not the failing action. The failing action is **`gradle/actions/setup-gradle` pinned to `50e97c2` (v6.1.0)**, which is not in [`apache/infrastructure-actions/actions.yml`](https://github.com/apache/infrastructure-actions/blob/main/actions.yml).

Affected: CI, Code Style, Code Analysis, Coverage, End to End, Licensing (RAT), Vulnerability Scan, Groovy Snapshot Canary, and every release workflow.

## Changes

| Action | Before | After |
|---|---|---|
| `gradle/actions/setup-gradle` (29×) | `50e97c2` v6.1.0 — **not approved** | `9c971963` v6.3.0 — approved, no expiry |
| `aws-actions/configure-aws-credentials` | `e7f100cf` v6.2.0 — approved, **expires 2026-09-20** | `e6de0542` v6.2.3 — approved, no expiry |
| `actions/cache` (12×, `gradle.yml`) | `@v4` floating | `0057852b` v4.3.0 |
| `actions/checkout` (3×, `vulnerability-scan.yml`) | `@v6` floating | `de0fac2e` v6.0.2 |
| `actions/setup-java` (2×, `vulnerability-scan.yml`) | `@v4` floating | `be666c2f` v5.2.0 |

`gradle/actions/setup-gradle@9c971963` is v6.3.0, the newest approved entry and the only v6 entry with no `expires_at`. v6.1.1 (`5e2ebd06`) is also approved but expires 2026-09-05, so pinning to it would have re-broken CI in two days.

The `actions/*` refs were allowed by namespace and were not breaking anything; they are pinned here for the supply-chain consistency `CLAUDE.md` asks for, using the same versions already in use elsewhere in the repo. `apache/grails-github-actions/*@asf` is deliberately left on the branch ref — that floating pin is intentional and ASF-owned.

## Verification

Every `uses:` in `.github/workflows/` was checked against `actions.yml`; all third-party actions now resolve to an approved SHA with no expiry date, and each SHA was confirmed against its upstream tag via the GitHub API. All workflow YAML parses.

## Note

`8.0.x` carries the same unapproved `50e97c2` pin and is failing identically; it needs the same fix.

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -39,7 +39,7 @@ jobs:
           distribution: liberica
           java-version: 17
       - name: "🐘 Setup Gradle"
-        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        uses: gradle/actions/setup-gradle@9c971963bec38e04b3d30dcc455b5382be2fdbfb # v6.3.0
         with:
           cache-provider: basic # 'basic' uses the MIT-licensed, open-source cache provider; the default 'enhanced' provider (v6+) is proprietary (Gradle commercial Terms of Use)
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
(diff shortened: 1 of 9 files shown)


jdaugherty (approved):
historically we have pinned the shas for github owned actions but itherwise if it runs i am good

matrei (approved):

PR #13478, opened 2024-04-21
chore: update url to HTTPS

puneetbehl:
Update to HTTPS url in the Apache License and other places.

PR #16310, opened 2026-09-03
Fix SAX parser feature URIs and reject DOCTYPE declarations

jamesfredley:
## Summary

ASF security review finding **f002**: SAX XXE hardening in `SpringIOUtils.createParserFactory` set parser features with `https://xml.org/...` and `https://apache.org/...`. The registered JAXP / Xerces identifiers are the `http://` forms, so the factory rejected the features and the empty catch hid the failure. `XmlDataBindingSourceCreator` still uses `SpringIOUtils.createXmlSlurper()` for `application/xml` request bodies. The HTTP test-client `XmlUtils` had the same identifiers and left `disallow-doctype-decl` false.

This is hardening, not a HIGH CVE against the current threat model (THREAT_MODEL.md §9 disclaims parser configuration / XXE). No threat-model change in this PR.

## Changes

- Use the registered `http://xml.org/...` and `http://apache.org/...` SAX feature URIs.
- Set `disallow-doctype-decl` to `true` and disable XInclude.
- Apply the same defaults in the HTTP test-client XML slurper.
- Cover DOCTYPE rejection in `SpringIOUtilsSpec`, `XmlUtilsSpec`, and `TestHttpResponseSpec`.

## Testing

- `:grails-gradle:grails-gradle-model:test`
- `:grails-testing-support-http-client:test`
- `:grails-gradle:grails-gradle-model:codeStyle`
- `:grails-testing-support-http-client:codeStyle`

### DIFF
--- a/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
+++ b/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
@@ -423,7 +423,11 @@ public class SpringIOUtils {
             saxParserFactory = FactorySupport.createSaxParserFactory();
             saxParserFactory.setNamespaceAware(true);
             saxParserFactory.setValidating(false);
-            saxParserFactory.setXIncludeAware(false);
+            try {
+                saxParserFactory.setXIncludeAware(false);
+            } catch (UnsupportedOperationException e) {
+                // ignore, parser doesn't support
+            }
 
             try {
                 saxParserFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
--- a/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
+++ b/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
@@ -423,19 +423,20 @@ public class SpringIOUtils {
             saxParserFactory = FactorySupport.createSaxParserFactory();
             saxParserFactory.setNamespaceAware(true);
             saxParserFactory.setValidating(false);
+            saxParserFactory.setXIncludeAware(false);
 
             try {
-                saxParserFactory.setFeature("https://apache.org/xml/features/disallow-doctype-decl", false);
+                saxParserFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
             } catch (Exception pce) {
                 // ignore, parser doesn't support
             }
             try {
-                saxParserFactory.setFeature("https://xml.org/sax/features/external-general-entities", false);
+                saxParserFactory.setFeature("http://xml.org/sax/features/external-general-entities", false);
             } catch (Exception pce) {
                 // ignore, parser doesn't support
             }
             try {
-                saxParserFactory.setFeature("https://xml.org/sax/features/external-parameter-entities", false);
+                saxParserFactory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
             } catch (Exception pce) {
                 // ignore, parser doesn't support
             }
@@ -445,12 +446,12 @@ public class SpringIOUtils {
                 // ignore, parser doesn't support
             }
             try {
-                saxParserFactory.setFeature("https://apache.org/xml/features/nonvalidating/load-dtd-grammar", false);
+                saxParserFactory.setFeature("http://apache.org/xml/features/nonvalidating/load-dtd-grammar", false);
             } catch (Exception e) {
                 // ignore, parser doesn't support
             }
             try {
-                saxParserFactory.setFeature("https://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+                saxParserFactory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
             } catch (Exception e) {
                 // ignore, parser doesn't support
             }
--- a/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
+++ b/grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
@@ -49,6 +49,9 @@ import groovy.xml.XmlSlurper;
 
 import org.xml.sax.SAXException;
 
+import grails.util.Metadata;
+import org.apache.grails.gradle.common.XmlParserFeature;
+
 /**
  * Simple utility methods for file and stream copying.
  * All copy methods use a block size of 4096 bytes,
@@ -416,50 +419,86 @@ public class SpringIOUtils {
         return factory.newSAXParser();
     }
 
-    private static SAXParserFactory saxParserFactory = null;
+    /**
+     * Configuration key permitting {@code DOCTYPE} declarations in documents parsed by this class.
+     *
+     * <p>Parsers handed out here reject a {@code DOCTYPE} by default. Set
+     * {@code grails.xml.allowDocTypeDeclaration} to {@code true} in {@code application.yml}, or as
+     * a system property, to accept one.
+     *
+     * <p>Opting in does not reopen the XXE vector. External general entities, external parameter
+     * entities and external DTDs stay refused whichever way this is set, so an entity pointing at
+     * a file on disk still contributes nothing. What opting in changes is only whether a document
+     * carrying a declaration is refused outright.
+     *
+     * <p>It exists because these parsers also read trusted descriptors from the classpath, and
+     * some of those carry a {@code DOCTYPE}. JSP tag library descriptors are the common case:
+     * {@code jakarta.servlet.jsp.jstl} ships several, among them {@code c-1_0-rt.tld}, which the
+     * default {@code grails.gsp.tldScanPattern} scans.
+     */
+    public static final String ALLOW_DOCTYPE_DECLARATION = "grails.xml.allowDocTypeDeclaration";
+
+    /**
+     * Parser features switched off for every parser this class hands out.
+     *
+     * <p>{@link XmlParserFeature#DISALLOW_DOCTYPE_DECL} is handled separately because it is the
+     * one feature an application may turn off; see {@link #ALLOW_DOCTYPE_DECLARATION}.
+     */
+    private static final XmlParserFeature[] DISABLED_PARSER_FEATURES = {
+        XmlParserFeature.EXTERNAL_GENERAL_ENTITIES,
+        XmlParserFeature.EXTERNAL_PARAMETER_ENTITIES,
+        XmlParserFeature.LOAD_DTD_GRAMMAR,
+        XmlParserFeature.LOAD_EXTERNAL_DTD
+    };
+
+    private static SAXParserFactory strictParserFactory = null;
+
+    private static SAXParserFactory docTypeParserFactory = null;
 
     private static SAXParserFactory createParserFactory() throws ParserConfigurationException {
-        if (saxParserFactory == null) {
-            saxParserFactory = FactorySupport.createSaxParserFactory();
-            saxParserFactory.setNamespaceAware(true);
-            saxParserFactory.setValidating(false);
-            try {
-                saxParserFactory.setXIncludeAware(false);
-            } catch (UnsupportedOperationException e) {
-                // ignore, parser doesn't support
+        if (isDocTypeDeclarationAllowed()) {
+            if (docTypeParserFactory == null) {
+                docTypeParserFactory = buildParserFactory(true);
             }
+            return docTypeParserFactory;
+        }
+        if (strictParserFactory == null) {
+            strictParserFactory = buildParserFactory(false);
+        }
+        return strictParserFactory;
+    }
 
+    private static boolean isDocTypeDeclarationAllowed() {
+        return Boolean.TRUE.equals(
+                Metadata.getCurrent().getProperty(ALLOW_DOCTYPE_DECLARATION, Boolean.class, Boolean.FALSE));
+    }
+
+    private static SAXParserFactory buildParserFactory(boolean allowDocTypeDeclaration) throws ParserConfigurationException {
+        SAXParserFactory factory = FactorySupport.createSaxParserFactory();
+        factory.setNamespaceAware(true);
+        factory.setValidating(false);
+        try {
+            factory.setXIncludeAware(false);
+        } catch (UnsupportedOperationException e) {
+            // ignore, parser doesn't support
+        }
+        try {
+            factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        } catch (Exception e) {
+            // ignore, parser doesn't support
+        }
+        try {
+            factory.setFeature(XmlParserFeature.DISALLOW_DOCTYPE_DECL.getFeatureName(), !allowDocTypeDeclaration);
+        } catch (Exception e) {
+            // ignore, parser doesn't support
+        }
+        for (XmlParserFeature feature : DISABLED_PARSER_FEATURES) {
             try {
-                saxParserFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
-            } catch (Exception pce) {
-                // ignore, parser doesn't support
-            }
-            try {
-                saxParserFactory.setFeature("http://xml.org/sax/features/external-general-entities", false);
-            } catch (Exception pce) {
-                // ignore, parser doesn't support
-            }
-            try {
-                saxParserFactory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
-            } catch (Exception pce) {
-                // ignore, parser doesn't support
-            }
-            try {
-                saxParserFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
-            } catch (Exception e) {
-                // ignore, parser doesn't support
-            }
-            try {
-                saxParserFactory.setFeature("http://apache.org/xml/features/nonvalidating/load-dtd-grammar", false);
-            } catch (Exception e) {
-                // ignore, parser doesn't support
-            }
-            try {
-                saxParserFactory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+                factory.setFeature(feature.getFeatureName(), false);
             } catch (Exception e) {
                 // ignore, parser doesn't support
             }
         }
-        return saxParserFactory;
+        return factory;
     }
 }
--- a/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
+++ b/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
@@ -41,4 +41,15 @@ class SpringIOUtilsSpec extends Specification {
         then:
         thrown(SAXParseException)
     }
+
+    void 'createXmlSlurper rejects doctype declarations with internal entities'() {
+        when:
+        SpringIOUtils.createXmlSlurper().parseText('''<!DOCTYPE root [
+<!ENTITY msg "safe">
+]>
+<root>&msg;</root>''')
+
+        then:
+        thrown(SAXParseException)
+    }
 }
--- a/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
+++ b/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
@@ -1,0 +1,44 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.grails.io.support
+
+import org.xml.sax.SAXParseException
+import spock.lang.Specification
+
+class SpringIOUtilsSpec extends Specification {
+
+    void 'createXmlSlurper parses documents without a doctype'() {
+        when:
+        def xml = SpringIOUtils.createXmlSlurper().parseText('<root><child>ok</child></root>')
+
+        then:
+        xml.child.text() == 'ok'
+    }
+
+    void 'createXmlSlurper rejects doctype declarations with external entities'() {
+        when:
+        SpringIOUtils.createXmlSlurper().parseText('''<!DOCTYPE root [
+<!ENTITY ext SYSTEM "file:///not-resolved">
+]>
+<root>&ext;</root>''')
+
+        then:
+        thrown(SAXParseException)
+    }
+}
--- a/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
+++ b/grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
@@ -18,12 +18,50 @@
  */
 package org.grails.io.support
 
-import org.xml.sax.SAXParseException
-import spock.lang.Specification
+import java.nio.file.Files
+import java.nio.file.Path
 
+import grails.util.Metadata
+
+import org.xml.sax.SAXParseException
+
+import spock.lang.Specification
+import spock.lang.TempDir
+
+/**
+ * Asserts the parser hardening applied by {@link SpringIOUtils} through observable parsing
+ * behaviour rather than by reading feature flags back off the factory.
+ *
+ * <p>This is deliberate. Reading the flags back would require this spec to hold its own copy of
+ * the feature identifiers, so a search-and-replace over those identifiers would rewrite the
+ * production code and this spec together and the suite would still pass. Driving real documents
+ * through the parser keeps the assertions independent of how the hardening is spelled.
+ */
 class SpringIOUtilsSpec extends Specification {
 
-    void 'createXmlSlurper parses documents without a doctype'() {
+    /** Shape of a JSP 1.2 tag library descriptor, as shipped inside jakarta jstl. */
+    private static final String TLD = '''<!DOCTYPE taglib
+  PUBLIC "-//Sun Microsystems, Inc.//DTD JSP Tag Library 1.2//EN"
+  "http://java.sun.com/dtd/web-jsptaglibrary_1_2.dtd">
+<taglib>
+  <uri>jakarta.tags.core</uri>
+  <tag><name>out</name><tag-class>org.example.OutTag</tag-class></tag>
+</taglib>'''
+
+    @TempDir
+    Path tempDir
+
+    void cleanup() {
+        System.clearProperty(SpringIOUtils.ALLOW_DOCTYPE_DECLARATION)
+        Metadata.reset()
+    }
+
+    private static void allowDocTypeDeclarations() {
+        System.setProperty(SpringIOUtils.ALLOW_DOCTYPE_DECLARATION, 'true')
+        Metadata.reset()
+    }
+
+    void 'createXmlSlurper parses a document without a doctype'() {
         when:
         def xml = SpringIOUtils.createXmlSlurper().parseText('<root><child>ok</child></root>')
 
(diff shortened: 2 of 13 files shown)


grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
```diff
@@ -0,0 +1,44 @@
...
+    void 'createXmlSlurper rejects doctype declarations with external entities'() {
+        when:
+        SpringIOUtils.createXmlSlurper().parseText('''<!DOCTYPE root [
+<!ENTITY ext SYSTEM "file:///not-resolved">
+]>
+<root>&ext;</root>''')
+
+        then:
+        thrown(SAXParseException)
+    }
```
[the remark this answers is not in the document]
jamesfredley: Valid. Added coverage confirming SpringIOUtils rejects an internal-subset DOCTYPE with an internal entity.

grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
```diff
@@ -423,19 +423,20 @@ private static SAXParserFactory createParserFactory() throws ParserConfiguration
             saxParserFactory = FactorySupport.createSaxParserFactory();
             saxParserFactory.setNamespaceAware(true);
             saxParserFactory.setValidating(false);
+            saxParserFactory.setXIncludeAware(false);
```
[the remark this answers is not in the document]
jamesfredley: Valid. Wrapped setXIncludeAware(false) in UnsupportedOperationException handling so unsupported parsers continue securely.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy
```diff
@@ -224,6 +223,7 @@ class XmlUtils {
         def saxParserFactory = FactorySupport.createSaxParserFactory().tap {
             it.namespaceAware = true
             it.validating = false
+            it.XIncludeAware = false
         }
```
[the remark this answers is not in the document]
jamesfredley: Valid. Moved XIncludeAware configuration outside the factory tap and catch UnsupportedOperationException so parsing continues when unsupported.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/utils/XmlUtilsSpec.groovy
```diff
@@ -278,37 +276,15 @@ class XmlUtilsSpec extends Specification {
...
 
-    void 'newXmlSlurper allows inline doctype declarations with internal entities'() {
+    void 'newXmlSlurper rejects doctype declarations with external entities'() {
         when:
         def parsed = XmlUtils.newXmlSlurper().parseText('''<!DOCTYPE root [
-<!ENTITY msg "safe">
-]>
-<root>&msg;</root>''')
+ <!ENTITY ext SYSTEM "file:///not-resolved">
+ ]>
```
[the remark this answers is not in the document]
jamesfredley: Valid. Added XmlUtils.newXmlSlurper() coverage confirming an internal-subset DOCTYPE with an internal entity is rejected.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/TestHttpResponseSpec.groovy
```diff
@@ -199,36 +198,18 @@ class TestHttpResponseSpec extends Specification {
...
-        def secretFile = Files.createTempFile('test-http-response-xml', '.txt')
-        Files.writeString(secretFile, 'top-secret-token')
-        def uri = secretFile.toUri().toASCIIString()
-        def response = mockResponse(200, """<!DOCTYPE root [
-<!ENTITY ext SYSTEM '${uri}'>
-]>
-<root>&ext;</root>""")
+        def response = mockResponse(200, '''<!DOCTYPE root [
+ <!ENTITY ext SYSTEM "file:///not-resolved">
+ ]>
```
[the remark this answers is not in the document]
jamesfredley: Valid. Added response.xml() coverage confirming an internal-subset DOCTYPE with an internal entity is rejected.

jamesfredley:
@copilot re-review please. Follow-up 6fdc50880aa1e4f0bf33131cf343006ec7924ad7 guards XInclude setXIncludeAware and adds internal-subset DOCTYPE rejection tests.

grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
```diff
@@ -423,19 +423,24 @@ private static SAXParserFactory createParserFactory() throws ParserConfiguration
...
             saxParserFactory.setValidating(false);
+            try {
+                saxParserFactory.setXIncludeAware(false);
+            } catch (UnsupportedOperationException e) {
+                // ignore, parser doesn't support
+            }
 
             try {
-                saxParserFactory.setFeature("https://apache.org/xml/features/disallow-doctype-decl", false);
+                saxParserFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```
jdaugherty: This factory is shared with `TldReader` and `WebXmlTagLibraryReader`, which read TLDs and `web.xml` from the classpath. Those routinely carry a DOCTYPE, so rejecting it here stops JSP tag library resolution.

Concretely: `GspAutoConfiguration` sets the default `grails.gsp.tldScanPattern` to include `classpath*:/META-INF/c-1_0-rt.tld`, and that file in `jakarta.servlet.jsp.jstl:3.0.1` begins

```xml
<!DOCTYPE taglib
  PUBLIC "-//Sun Microsystems, Inc.//DTD JSP Tag Library 1.2//EN"
  "http://java.sun.com/dtd/web-jsptaglibrary_1_2.dtd">
```

Parsing it with this feature set yields `SAXParseException: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.` `TagLibraryResolverImpl.initialize()` doesn't catch, so it surfaces from the first `resolveTagLibrary(uri)` call.

A corroborating hint in our own code: `TldReader` falls back to `tag.'tagclass'.text()`, the JSP 1.1 element name, which only appears in DTD-era TLDs — the reader is written to handle exactly the documents this now rejects.

Setting this to `false` keeps the descriptors parsing. XXE is still closed by the `external-*-entities` and `load-external-dtd` settings below; I checked that a `file://` external entity resolves to empty rather than the file contents with DOCTYPE permitted.

grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
```diff
@@ -423,19 +423,24 @@ private static SAXParserFactory createParserFactory() throws ParserConfiguration
             saxParserFactory = FactorySupport.createSaxParserFactory();
             saxParserFactory.setNamespaceAware(true);
             saxParserFactory.setValidating(false);
+            try {
```
jdaugherty: Suggestion rather than a defect: these five identifiers are also declared in `XmlUtils` in the HTTP test client, as bare literals in both places. That duplication is how the `https` spelling got in and stayed in.

Worth collecting them in one enum in `grails-gradle-common` — same build as this module, and `grails-common` already re-exports it via `api 'org.apache.grails.gradle:grails-gradle-common'`, so the root build and the test client can both see it. It gives the "these are registered names, not URLs, don't rewrite the scheme" comment a single home.

Worth knowing for context: `io.spring.nohttp`, which forge already applies, would *not* have flagged these. Its default allowlist carries `^http://xml\.org/.*` and `^http://apache\.org/xml/features/.*`. So a linter isn't what needs guarding against here.

grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
```diff
@@ -0,0 +1,55 @@
...
+        when:
+        SpringIOUtils.createXmlSlurper().parseText('''<!DOCTYPE root [
+<!ENTITY msg "safe">
+]>
+<root>&msg;</root>''')
+
+        then:
+        thrown(SAXParseException)
+    }
+}
```
jdaugherty: These assert DOCTYPE rejection, which leaves `external-general-entities`, `external-parameter-entities`, `load-dtd-grammar` and `load-external-dtd` with no coverage at all — once DOCTYPE is refused, no document can reach them. The tests removed from `XmlUtilsSpec` and `TestHttpResponseSpec` were the only thing pinning them. The same wrong-identifier mistake in those four lines would still pass CI.

One thing to watch when adding coverage: a test that reads the flags back with `getFeature(...)` has to restate the identifiers, so a search-and-replace rewrites the test and the production code together and the suite still passes — it just errors with `SAXNotRecognizedException` pointing at the test. Driving a document with an external entity through the parser and asserting the file contents don't appear keeps the assertion free of any identifier, so it can't be rewritten in step.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/utils/XmlUtilsSpec.groovy
```diff
@@ -278,37 +276,26 @@ class XmlUtilsSpec extends Specification {
         xml == "<?xml version='1.1' encoding='UTF-16'?><product />"
     }
 
-    void 'newXmlSlurper allows inline doctype declarations with internal entities'() {
+    void 'newXmlSlurper rejects doctype declarations with external entities'() {
```
jdaugherty: Small one: `parsed` is assigned but the block only asserts `thrown(SAXParseException)`, so the variable is dead. Same in the internal-entity test below.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/TestHttpResponseSpec.groovy
```diff
@@ -199,36 +198,32 @@ class TestHttpResponseSpec extends Specification {
         xmlResponse.xml().item.text() == 'value'
     }
 
-    void 'xml uses a secure default slurper that does not resolve external entities'() {
+    void 'xml rejects doctype declarations with external entities'() {
```
jdaugherty: Dropping `e.message.contains('External Entity')` leaves both tests asserting a bare `SAXParseException`, which a merely malformed document would also satisfy. Asserting on `'DOCTYPE is disallowed'` would keep them pinned to the behaviour they're named for.

Also a formatting nit: the fixture here indents `<!ENTITY` and `]>` by one space while the internal-entity fixture below doesn't.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy
```diff
@@ -118,8 +118,7 @@ class XmlUtils {
     /**
      * Creates an {@link XmlSlurper} with secure defaults.
      * <p>
-     * The default parser is namespace aware, non-validating, permits inline DOCTYPE declarations,
-     * and disables external entity expansion plus external DTD loading.
+     * The default parser is namespace aware, non-validating, and rejects DOCTYPE declarations.
```
jdaugherty: This javadoc drops "disables external entity expansion plus external DTD loading", but those features are still being set. The README and `integrationTesting.adoc` both kept the sentence, so it's only the javadoc that loses it.

grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
jdaugherty, line 433: This default is right, but nothing lets an application past it, and one common setup needs to get past it.

This factory is shared with `TldReader` and `WebXmlTagLibraryReader`, which read TLDs and `web.xml` from the classpath. `GspAutoConfiguration` sets the default `grails.gsp.tldScanPattern` to include `classpath*:/META-INF/c-1_0-rt.tld`, and that file in `jakarta.servlet.jsp.jstl:3.0.1` begins

```xml
<!DOCTYPE taglib
  PUBLIC "-//Sun Microsystems, Inc.//DTD JSP Tag Library 1.2//EN"
  "http://java.sun.com/dtd/web-jsptaglibrary_1_2.dtd">
```

Parsing it with this feature set gives `SAXParseException: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.` `TagLibraryResolverImpl.initialize()` doesn't catch, so it surfaces from the first `resolveTagLibrary(uri)` call.

A hint from our own code: `TldReader` falls back to `tag.'tagclass'.text()`, the JSP 1.1 element name, which only appears in DTD-era TLDs — the reader is written to handle exactly the documents this now rejects.

jamesfredley/grails-core#4 keeps this `true` and adds `grails.xml.allowDocTypeDeclaration`, read via `Metadata` so it comes from `application.yml` or a system property, off by default. Opting in doesn't reopen XXE: with declarations permitted, a `file://` external entity still resolves to empty rather than the file contents, which the tests there assert.

grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java
jdaugherty, line 426: These five are also declared in `XmlUtils` in the HTTP test client, as bare literals in both places. That duplication is how a single sweep became two broken call sites.

`eed8df3594` (#13478) is where the `https` spelling came from — a 79-file pass done as preparation for adopting `io.spring.nohttp`. Since we still intend to adopt it, this can happen again, which is the argument for putting the values somewhere with the explanation attached rather than leaving them as bare literals in two files.

For what it's worth, nohttp itself would not have flagged these: its default allowlist already carries `^http://xml\.org/.*` and `^http://apache\.org/xml/features/.*`. Adopting the tool needs no exception for them, and its allowlist is a decent reference for which `http://` strings here are identifiers rather than links.

jamesfredley/grails-core#4 collects them in an enum in `grails-gradle-common` — same build as this module, and already visible to the root build through `grails-common`.

grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy
jdaugherty, line 55: These assert DOCTYPE rejection, which leaves `external-general-entities`, `external-parameter-entities`, `load-dtd-grammar` and `load-external-dtd` with no coverage at all — once DOCTYPE is refused, no document can reach them. The tests removed from `XmlUtilsSpec` and `TestHttpResponseSpec` were the only thing pinning them, so the same wrong-identifier mistake in those four lines would still pass CI.

One thing worth knowing when adding coverage: a test that reads the flags back with `getFeature(...)` has to restate the identifiers, so a sweep rewrites the test and the production code together. It does fail, but at the test, with a name-lookup error whose tempting fix is to make the test tolerate it. Driving a document with an external entity through the parser and asserting the file contents don't appear keeps the assertion free of any identifier.

jamesfredley/grails-core#4 does that, and adds a spec deriving the list from `values()` that asserts each identifier is one a parser actually registers — so an unrecognised name fails by name instead of silently. I checked both failure modes against it: rewriting the enum to `https` fails 4 of 7 behavioural cases plus all 5 of those.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/utils/XmlUtilsSpec.groovy
jdaugherty, line 279: Small one: `parsed` is assigned but the block only asserts `thrown(SAXParseException)`, so the variable is dead. Same in the internal-entity test below.

grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/TestHttpResponseSpec.groovy
jdaugherty, line 201: Dropping `e.message.contains('External Entity')` leaves both tests asserting a bare `SAXParseException`, which a merely malformed document would also satisfy. Asserting on `'DOCTYPE is disallowed'` would keep them pinned to the behaviour they're named for.

Formatting nit: the fixture here indents `<!ENTITY` and `]>` by one space while the internal-entity fixture below doesn't.

grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy
jdaugherty, line 121: This javadoc drops "disables external entity expansion plus external DTD loading", but those features are still being set. The README and `integrationTesting.adoc` both kept the sentence, so it's only the javadoc that loses it.

No change needed to the strict behaviour here — a test client refusing DOCTYPE is the right call, and #4 leaves it exactly as you have it.

jdaugherty (changes requested):
Confirmed the premise before anything else: against the JDK parser all six `http://` identifiers are accepted and every `https://` form is answered with `SAXNotRecognizedException`, so the previous values were setting nothing at all. The correction is real, and the strict DOCTYPE default you picked is the right one to ship. Everything below is about keeping it that way.

I've opened jamesfredley/grails-core#4 against this branch with a suggested implementation of the three points below, so they're easy to take or drop.

**Where the `https` spelling came from.** `eed8df3594` ("chore: update url to HTTPS", #13478) rewrote them across 79 files in April 2024. That sweep was preparatory work toward adopting `io.spring.nohttp` — the tool itself has never run over these modules and would not have flagged these: its default allowlist carries `^http://xml\.org/.*` and `^http://apache\.org/xml/features/.*`, so adopting it needs no exception for them. The hazard was the manual preparation, not the tool. Worth calling out because if we make another run at adopting nohttp the same well-meant rewrite can land again, and the empty catches around `setFeature` will hide it again exactly as they did for two years. For this reason, I'm suggesting we centralize the config into an enum constant that we document this risk explicitly.

**DOCTYPE rejection stops JSP tag library resolution, with nothing configurable to get past it.** `createParserFactory()` returns one cached factory shared with `TldReader` and `WebXmlTagLibraryReader`, which read classpath descriptors that legitimately carry a DOCTYPE. `GspAutoConfiguration` scans `classpath*:/META-INF/c-1_0-rt.tld` by default, and that file in `org.glassfish.web:jakarta.servlet.jsp.jstl:3.0.1` opens with a JSP 1.2 DOCTYPE — eight of the twenty TLDs in that jar have one. `TagLibraryResolverImpl.initialize()` doesn't catch, so it surfaces from the first `resolveTagLibrary(uri)` call. No existing test covers it (the GSP tests use a fixture TLD without a DOCTYPE), so CI stays green.

Rather than weaken the default, #4 adds `grails.xml.allowDocTypeDeclaration`, read through `Metadata` so it works from `application.yml` or as a system property, defaulting to off. Applications opt in knowing what they're accepting, and it relaxes only whether a declaration is permitted — external entities and external DTDs stay refused either way. I'm really not sure this is a "true" vulnerability since we use this utility to parse developer code, but the config seems like a sensible default if we're going to continue to support JSP.

**The external-entity features end up untested.** The other four identifiers can't be reached by any test once DOCTYPE is rejected, and the deleted tests were their only coverage. Given this PR exists because a wrong identifier went unnoticed, worth not leaving those four in the same position.

**The identifiers are declared twice**, here and in the HTTP test client, as bare literals in both — which is how one rewrite became two. #4 collects them in an enum in `grails-gradle-common` (same build as `SpringIOUtils`; `grails-common` already re-exposes it via `api 'org.apache.grails.gradle:grails-gradle-common'`, so the test client sees it too), with the reason the scheme is fixed written next to the values.

jdaugherty:
All of my review feedback is implemented in https://github.com/jamesfredley/grails-core/pull/4 - I made this configurable given that it can break JSPs

matrei:
# AI Review Findings

Head `6fdc50880a` on `8.0.x` base `55d5076aa5`. I reproduced the premise independently against the JDK 21 parser (`com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl`): every `http://` identifier is accepted and every `https://` spelling is answered with `SAXNotRecognizedException`. The sweep that introduced the `https` forms is `eed8df3594` (#13478, 79 files), so since April 2024 `createParserFactory()` has configured nothing beyond `FEATURE_SECURE_PROCESSING`, and the `catch (Exception)` around each call kept that invisible. The correction is real and the direction is right.

There is one blocking problem with the default this PR ships, and it is not covered by any test in the repository. The follow-up in jamesfredley/grails-core#4 addresses it with a global opt-in; I would take that PR with one change in shape (see the review on jamesfredley/grails-core#4).

## [P1] Rejecting DOCTYPE in the shared factory breaks JSP tag library resolution for every application that uses JSTL

**Files:**

- `grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java:433`
- `grails-gsp/grails-web-jsp/src/main/groovy/org/grails/gsp/jsp/TldReader.groovy:49`
- `grails-gsp/grails-web-jsp/src/main/groovy/org/grails/gsp/jsp/TagLibraryResolverImpl.groovy:71-90`
- `grails-gsp/spring-boot/src/main/java/grails/gsp/boot/GspAutoConfiguration.java:323`

`createParserFactory()` returns one cached factory for every caller of `createXmlSlurper()` and `newSAXParser()`. Those callers fall into two trust levels:

| Caller | Input |
|---|---|
| `XmlDataBindingSourceCreator`, `HalXmlDataBindingSourceCreator`, `grails.converters.XML.parse` | HTTP request bodies (untrusted) |
| `TldReader`, `WebXmlTagLibraryReader`, `PluginUtils` | Descriptors on the application classpath (trusted) |

The descriptors are where the strict default bites. `GspAutoConfiguration` defaults `grails.gsp.tldScanPattern` to a list that ends with `classpath*:/META-INF/c-1_0-rt.tld`. In `org.glassfish.web:jakarta.servlet.jsp.jstl:3.0.1` that file opens with a JSP 1.2 DOCTYPE, and so do 8 of the 22 descriptors in the jar (`c-1_0*.tld`, `fmt-1_0*.tld`, `sql-1_0*.tld`, `x-1_0*.tld`). `TagLibraryResolverImpl.initialize()` scans every pattern in one loop and catches nothing, so the first `resolveTagLibrary(uri)` call throws `SAXParseException: DOCTYPE is disallowed ...` and no JSP tag library resolves, including `jakarta.tags.core` from the DOCTYPE-free `c.tld` that was scanned earlier in the same loop.

I confirmed this with a spec in `grails-gsp/plugin` (JSTL is already on that module's test runtime classpath) that scans `c-1_0-rt.tld` plus `c.tld` and resolves `jakarta.tags.core`. On this head it fails with `DOCTYPE is disallowed`. The existing GSP tests stay green only because `GroovyPageWithJSPTagsTests`, `AbstractGrailsTagTests` and `TagLibraryResolverTests` scan `c.tld`, `fmt.tld`, `core.tld` and `spring*.tld`, none of which carries a DOCTYPE; `TldReaderTests` uses a fixture without one. The two test examples that put JSTL on the runtime classpath (`gsp-layout`, `gsp-sitemesh3`) contain no `<%@ taglib %>` directive, so the resolver is never initialised there either.

The documented custom pattern `classpath*:/META-INF/*.tld` in `usingJSPTagLibraries.adoc` hits all eight.

The user-visible symptom is the worst kind: a 500 on the first page that uses a JSP tag, with a parser error that names a feature URI nobody set in the application.

**What I would change.** The two trust levels want two parsers, and the readers of trusted descriptors know they are reading trusted descriptors. Keep the strict default for request bodies and let the descriptor readers ask for DOCTYPE tolerance explicitly:

```java
// SpringIOUtils
public static XmlSlurper createXmlSlurper() throws ... {            // strict, for request bodies
    return createXmlSlurper(false);
}

public static XmlSlurper createXmlSlurper(boolean allowDocTypeDeclaration) throws ... {
    return new XmlSlurper(createParserFactory(allowDocTypeDeclaration).newSAXParser());
}
```

with `TldReader`, `WebXmlTagLibraryReader` and `PluginUtils` passing `true`. External general and parameter entities, DTD grammar loading and external DTD retrieval stay off on both factories, so the tolerant parser still resolves a `file://` entity to nothing and skips the `web-jsptaglibrary_1_2.dtd` reference instead of fetching it. That needs no configuration key, no upgrade note for JSP users, and does not relax request-body parsing as a side effect of using JSTL.

jamesfredley/grails-core#4 instead adds `grails.xml.allowDocTypeDeclaration`, read through `Metadata`, that switches the single shared factory. It works, and I ran its tests, but it makes "I use JSTL" and "accept DOCTYPE in request bodies" the same switch. Details in that review.

## [P2] A rejected hardening feature is still swallowed silently

**Files:**

- `grails-gradle/model/src/main/groovy/org/grails/io/support/SpringIOUtils.java:431-461`
- `grails-testing-support-http-client/src/main/groovy/org/apache/grails/testing/http/client/utils/XmlUtils.groovy:235-243`

The PR exists because `catch (Exception) { /* ignore */ }` hid a `SAXNotRecognizedException` for two years, and the catch blocks are unchanged. The tolerance for parsers lacking a feature is reasonable, but it should leave a trace. `grails-gradle-model` already has `slf4j-api` as an `api` dependency; a `LOG.warn("Parser {} does not support feature {}", factory.getClass(), name)` inside each catch costs nothing and would have surfaced this in every application log since 2024. Copilot's suggestion to make `disallow-doctype-decl` mandatory and propagate is too strong for a shared factory, but a warning is not.

jamesfredley/grails-core#4 adds `XmlParserFeatureSpec`, which asserts each identifier is recognised by the JDK parser. That pins the identifiers at test time; the warning covers the runtime case where a different SAX provider is first on the classpath.

## [P2] The four external-entity features lose their only behavioural coverage

**Files:**

- `grails-gradle/model/src/test/groovy/org/grails/io/support/SpringIOUtilsSpec.groovy`
- `grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/utils/XmlUtilsSpec.groovy:279-300`
- `grails-testing-support-http-client/src/test/groovy/org/apache/grails/testing/http/client/TestHttpResponseSpec.groovy:201-227`

Every new test asserts DOCTYPE rejection, and once a DOCTYPE is refused no document reaches `external-general-entities`, `external-parameter-entities`, `load-dtd-grammar` or `load-external-dtd`. The deleted `newXmlSlurper blocks external entities` and `xml uses a secure default slurper that does not resolve external entities` were the only tests that drove an entity through the parser and asserted the file contents did not appear. After this PR the same wrong identifier in any of those four lines passes CI again. jamesfredley/grails-core#4 restores this coverage for `SpringIOUtils` by testing with DOCTYPE permitted; the split-factory shape above makes that natural, because the tolerant factory is the one that can be tested for entity blocking.

## [P2] `THREAT_MODEL.md` §9 now states something the code no longer does

**Files:**

- `THREAT_MODEL.md:346`
- `THREAT_MODEL.md:470`

The description says "No threat-model change in this PR", but §9 currently reads "XXE in XML data binding. XML parsing is delegated to the underlying parser; the framework does not impose a parser configuration. *(inferred)*", and §14 question 13 proposes confirming exactly that. After this PR the framework does impose a configuration on every XML request body it binds, and the strict DOCTYPE default is a behaviour change users will hit. The sentence in §9 should move to §8 or be rewritten to describe what is now guaranteed (external entities and external DTDs refused, DOCTYPE refused unless opted in), and Q13 should be answered rather than left open. Leaving the disclaimer in place means the next security review will re-report f002 against a document that says the fix does not exist.

Related: the upgrade notes need to say that `application/xml`, `text/xml` and HAL XML request bodies carrying a DOCTYPE are now refused. `integrationTesting.adoc` only covers the test client. jamesfredley/grails-core#4 adds an "XML Parsing Defaults" section to `upgrading.adoc` that does this.

## [P3] Small things in the test client, still present on jamesfredley/grails-core#4's head

- `XmlUtils.groovy:116` javadoc drops "disables external entity expansion plus external DTD loading" although those features are still set; the README and `integrationTesting.adoc` kept the sentence.
- `XmlUtilsSpec.groovy:281,292`: `def parsed =` is assigned and never read.
- `TestHttpResponseSpec.groovy:201-227` and `XmlUtilsSpec.groovy:279-300`: both tests now assert a bare `thrown(SAXParseException)`, which a malformed document also satisfies. `e.message.contains('DOCTYPE is disallowed')` pins them to the behaviour their names claim.
- The external-entity fixtures indent `<!ENTITY` and `]>` by one space; the internal-entity fixtures next to them do not.

## Verified

- Feature recognition probe against JDK 21: 5/5 `http://` accepted, 5/5 `https://` rejected with `SAXNotRecognizedException`.
- JSTL 3.0.1 jar: 8 of 22 `.tld` files contain `<!DOCTYPE`; `c-1_0-rt.tld` (in the default scan pattern) is one of them, uri `http://java.sun.com/jstl/core_rt`.
- No `https://apache.org/xml/features` or `https://xml.org/sax/features` strings remain anywhere in the tree after this PR. The other `http://` sites (`GrailsUpdater`, `GrailsViolationAggregationPlugin`, `grails-test-report/build.gradle`) were already correct.
- `:grails-gsp:test` with a two-case spec scanning `c-1_0-rt.tld` plus `c.tld` (run on jamesfredley/grails-core#4's head, whose strict default is identical to this PR's): `resolveTagLibrary('jakarta.tags.core')` throws `SAXParseException` containing `DOCTYPE is disallowed`; with jamesfredley/grails-core#4's opt-in both `jakarta.tags.core` and `http://java.sun.com/jstl/core_rt` resolve.
- `:grails-gradle-model:test`, `:grails-gradle-common:test`, `:grails-testing-support-http-client:test` and the matching `codeStyle` tasks on jamesfredley/grails-core#4's head: 173 cases, all green.
- Commit-level checks on `6fdc50880a` show only CodeQL and the release drafter; the PR workflow runs are what CI reports, and none of them scans a DOCTYPE-bearing TLD.

jamesfredley:
Review follow-up that was mistakenly opened on the archive fork is now https://github.com/apache/grails-core/pull/16328

jamesfredley:
This PR now includes jdaugherty's review follow-up (previously jamesfredley/grails-core#4 / apache/grails-core#16328).

jamesfredley:
Superseded by https://github.com/apache/grails-core/pull/16331 so the head branch is on apache/grails-core (not the archive fork).

PR #16311, opened 2026-09-03
Canonicalize interceptor and security matcher request paths

jamesfredley:
## Summary

ASF security review finding **f003**, with sibling matchers **f017** and **f020**.

Interceptor URI matching used the raw servlet `requestURI` (not decoded). `UrlMappingMatcher` only stripped `;` then ran `AntPathMatcher`. Grails dispatch uses Spring `UrlPathHelper.getPathWithinApplication`, which decodes and strips matrix parameters. Encoded segments (`/%61dmin/deleteUser`) or matrix parameters (`/admin;x=1/deleteUser`) could therefore reach a controller while `match(uri: '/admin/**')` did not fire.

The same requestURI/dispatch desync existed in:

- Spring Security compat `AntPathRequestMatcher` (f017)
- `IpAddressFilter` (f020), including forwarded URIs

This is a real matcher/dispatch inconsistency. It is **not** a framework-wide unauthenticated auth bypass: the shipped auth example matches controller/action from `UrlMappingInfo`, application-level auth is out of scope (threat model §3), and URI-pattern interceptors are a documented optional matcher. Treat as a bug / ASF MODERATE at most, not a 9.8 CVE.

## Changes

- Interceptor matching uses `UrlPathHelper.getPathWithinApplication`.
- `UrlMappingMatcher` canonicalizes the path (decode + strip matrix parameters) and still accepts context-prefixed patterns from issue 10857.
- Compat `AntPathRequestMatcher` and `IpAddressFilter` match the same canonical application path.

## Testing

- `:grails-interceptors:test`
- `:grails-spring-security-compat:test`
- `:grails-spring-security:test`
- codeStyle on those modules

### DIFF
--- a/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
+++ b/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
@@ -29,6 +29,7 @@ import jakarta.servlet.http.HttpServletResponse
 
 import org.springframework.core.Ordered
 import org.springframework.web.servlet.ModelAndView
+import org.springframework.web.util.UrlPathHelper
 
 import grails.artefact.controller.support.RequestForwarder
 import grails.artefact.controller.support.ResponseRedirector
@@ -89,25 +90,17 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
             allMatchers << matcher
         }
 
-        HttpServletRequest req = request
-        String ctxPath = req.contextPath
-        String uri = req.requestURI
-        String noCtxUri = uri - ctxPath
-        boolean checkNoCtxUri = ctxPath && uri.startsWith(ctxPath)
+        String uri = UrlPathHelper.defaultInstance.getPathWithinApplication(request)
 
         def matchedInfo = request.getAttribute(UrlMappingsHandlerMapping.MATCHED_REQUEST)
 
         UrlMappingInfo grailsMappingInfo = (UrlMappingInfo) matchedInfo
 
         for (Matcher matcher in allMatchers) {
-            boolean matchUri = matcher.doesMatch(uri, grailsMappingInfo, req.method)
-            boolean matchNoCtxUri = matcher.doesMatch(noCtxUri, grailsMappingInfo, req.method)
-
-            if (matcher.isExclude() && matchUri && matchNoCtxUri) {
-                // Exclude interceptors are special because with only one of the conditions being false the interceptor
-                // won't be applied to the request
-                return true
-            } else if (!matcher.isExclude() && (matchUri || (checkNoCtxUri && matchNoCtxUri))) {
+            boolean matches = matcher instanceof UrlMappingMatcher
+                    ? ((UrlMappingMatcher) matcher).doesMatch(uri, grailsMappingInfo, request.method, request.contextPath)
+                    : matcher.doesMatch(uri, grailsMappingInfo, request.method)
+            if (matches) {
                 return true
             }
         }
--- a/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
+++ b/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
@@ -78,6 +78,15 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
         doesMatch(request)
     }
     /**
+     * Whether this interceptor matches the given request.
+     *
+     * <p>URI matchers see the same path that URL mappings route on: decoded, with matrix parameters removed and the
+     * context path stripped. During a {@code RequestDispatcher} include this is the included path, which is also
+     * the path the include is dispatched on. The path is canonicalized once per call and passed to every
+     * {@link Matcher}. A request whose URI contains a malformed percent escape is matched undecoded rather than
+     * failing.
+     *
+     * @param request The request to match
      * @return Whether the current interceptor does match
      */
     @Generated
@@ -90,17 +99,21 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
             allMatchers << matcher
         }
 
-        String uri = UrlPathHelper.defaultInstance.getPathWithinApplication(request)
-
-        def matchedInfo = request.getAttribute(UrlMappingsHandlerMapping.MATCHED_REQUEST)
-
-        UrlMappingInfo grailsMappingInfo = (UrlMappingInfo) matchedInfo
+        String uri
+        try {
+            uri = UrlPathHelper.defaultInstance.getPathWithinApplication(request)
+        } catch (IllegalArgumentException ignored) {
+            // the URI has a malformed percent escape and cannot be decoded: match it undecoded rather than fail
+            UrlPathHelper rawPathHelper = new UrlPathHelper()
+            rawPathHelper.urlDecode = false
+            uri = rawPathHelper.getPathWithinApplication(request)
+        }
+        String contextPath = request.contextPath
+        String method = request.method
+        UrlMappingInfo grailsMappingInfo = (UrlMappingInfo) request.getAttribute(UrlMappingsHandlerMapping.MATCHED_REQUEST)
 
         for (Matcher matcher in allMatchers) {
-            boolean matches = matcher instanceof UrlMappingMatcher
-                    ? ((UrlMappingMatcher) matcher).doesMatch(uri, grailsMappingInfo, request.method, request.contextPath)
-                    : matcher.doesMatch(uri, grailsMappingInfo, request.method)
-            if (matches) {
+            if (matcher.doesMatch(uri, grailsMappingInfo, method, contextPath)) {
                 return true
             }
         }
--- a/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
+++ b/grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
@@ -80,11 +80,19 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
     /**
      * Whether this interceptor matches the given request.
      *
-     * <p>URI matchers see the same path that URL mappings route on: decoded, with matrix parameters removed and the
-     * context path stripped. During a {@code RequestDispatcher} include this is the included path, which is also
-     * the path the include is dispatched on. The path is canonicalized once per call and passed to every
-     * {@link Matcher}. A request whose URI contains a malformed percent escape is matched undecoded rather than
-     * failing.
+     * <p>URI matchers see the same path that URL mappings route on, resolved with the same {@link UrlPathHelper}
+     * call: path parameters ({@code ;name=value}) are removed per segment before percent-decoding, as Jakarta
+     * Servlet 6.0 section 3.5.2 requires of containers, so an encoded semicolon ({@code %3B}) stays a literal
+     * character (RFC 3986 section 2.2); the path is decoded exactly once (RFC 3986 section 2.4); and the context
+     * path is stripped. Two points follow Spring, and therefore dispatch, rather than the URI specification: the
+     * context path is compared case-insensitively although RFC 3986 section 6.2.2.1 defines paths as case-sensitive,
+     * and during a {@code RequestDispatcher} include the included path is matched, which is also the path the
+     * include is dispatched on. A matcher that disagreed with dispatch on either point would let a request reach a
+     * controller without its interceptors, so dispatch wins.
+     *
+     * <p>The path is canonicalized once per call and passed to every {@link Matcher}. A request whose URI contains
+     * an illegal percent escape, which a Servlet 6.0 container rejects with 400 before dispatch, is matched
+     * undecoded rather than failing.
      *
      * @param request The request to match
      * @return Whether the current interceptor does match
@@ -103,7 +111,7 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
         try {
             uri = UrlPathHelper.defaultInstance.getPathWithinApplication(request)
         } catch (IllegalArgumentException ignored) {
-            // the URI has a malformed percent escape and cannot be decoded: match it undecoded rather than fail
+            // illegal percent escape: match the undecoded path rather than fail
             UrlPathHelper rawPathHelper = new UrlPathHelper()
             rawPathHelper.urlDecode = false
             uri = rawPathHelper.getPathWithinApplication(request)
--- a/grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
+++ b/grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
@@ -18,12 +18,15 @@
  */
 package org.grails.plugins.web.interceptors
 
+import java.nio.charset.StandardCharsets
 import java.util.regex.Pattern
 
 import groovy.transform.CompileStatic
 import org.codehaus.groovy.util.HashCodeHelper
 
 import org.springframework.util.AntPathMatcher
+import org.springframework.web.util.UriUtils
+import org.springframework.web.util.UrlPathHelper
 
 import grails.artefact.Interceptor
 import grails.interceptors.Matcher
@@ -61,16 +64,20 @@ class UrlMappingMatcher implements Matcher {
     }
 
     boolean doesMatch(String uri, UrlMappingInfo info, String method) {
-        boolean hasUriPatterns = !uriPatterns.isEmpty()
+        doesMatch(uri, info, method, null)
+    }
 
-        boolean isExcluded = this.isExcluded(uri, info)
+    boolean doesMatch(String uri, UrlMappingInfo info, String method, String contextPath) {
+        boolean hasUriPatterns = !uriPatterns.isEmpty()
+        String path = canonicalizePath(uri)
+
+        boolean isExcluded = this.isExcluded(path, info, contextPath)
         if (matchAll && !isExcluded) return true
 
         if (!isExcluded) {
             if (hasUriPatterns) {
-                uri = uri.replace(';', '')
                 for (pattern in uriPatterns) {
-                    if (pathMatcher.match(pattern, uri)) {
+                    if (matchesPattern(pattern, path, contextPath)) {
                         return true
                     }
                 }
@@ -84,8 +91,12 @@ class UrlMappingMatcher implements Matcher {
     }
 
     protected boolean isExcluded(String uri, UrlMappingInfo info) {
+        isExcluded(uri, info, null)
+    }
+
+    protected boolean isExcluded(String uri, UrlMappingInfo info, String contextPath) {
         for (pattern in uriExcludePatterns) {
-            if (pathMatcher.match(pattern, uri)) {
+            if (matchesPattern(pattern, uri, contextPath)) {
                 return true
             }
         }
@@ -97,6 +108,30 @@ class UrlMappingMatcher implements Matcher {
             }
         }
         return false
+    }
+
+    private String canonicalizePath(String uri) {
+        if (uri == null || uri.isEmpty()) {
+            return '/'
+        }
+        String path = UrlPathHelper.defaultInstance.removeSemicolonContent(uri)
+        try {
+            path = UriUtils.decode(path, StandardCharsets.UTF_8)
+        } catch (IllegalArgumentException ignored) {
+            // keep the semicolon-stripped form when the URI is malformed
+        }
+        path = UrlPathHelper.defaultInstance.removeSemicolonContent(path)
+        path ?: '/'
+    }
+
+    private boolean matchesPattern(String pattern, String path, String contextPath) {
+        if (pathMatcher.match(pattern, path)) {
+            return true
+        }
+        if (contextPath && contextPath != '/' && (pattern == contextPath || pattern.startsWith(contextPath + '/'))) {
+            return pathMatcher.match(pattern.substring(contextPath.length()) ?: '/', path)
+        }
+        false
     }
 
     protected boolean doesMatchInternal(UrlMappingInfo info, String method) {
--- a/grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
+++ b/grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
@@ -18,15 +18,12 @@
  */
 package org.grails.plugins.web.interceptors
 
-import java.nio.charset.StandardCharsets
 import java.util.regex.Pattern
 
 import groovy.transform.CompileStatic
 import org.codehaus.groovy.util.HashCodeHelper
 
 import org.springframework.util.AntPathMatcher
-import org.springframework.web.util.UriUtils
-import org.springframework.web.util.UrlPathHelper
 
 import grails.artefact.Interceptor
 import grails.interceptors.Matcher
@@ -34,6 +31,12 @@ import grails.web.mapping.UrlMappingInfo
 
 /**
  * Used to match {@link UrlMappingInfo} instance by {@link grails.artefact.Interceptor} instances
+ *
+ * <p>URI patterns are matched against the path passed to {@link #doesMatch(String, UrlMappingInfo, String, String)}
+ * as-is. {@link grails.artefact.Interceptor#doesMatch(jakarta.servlet.http.HttpServletRequest)} passes the decoded,
+ * application-relative path that URL mappings route on, so this class must not decode or remove matrix parameters
+ * again: that would match a different path than the one dispatched ({@code /health%3Bx} is dispatched as
+ * {@code /health;x}, not as {@code /health}).
  *
  * @author Graeme Rocher
  * @since 3.0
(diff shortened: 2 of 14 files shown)


jamesfredley:
Copilot's approval is valid. No further code changes are required: `interceptor`, `UrlMappingMatcher`, `AntPathRequestMatcher`, and `IpAddressFilter` already share the dispatch canonical path.

grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
```diff
@@ -89,25 +90,17 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
             allMatchers << matcher
         }
 
-        HttpServletRequest req = request
-        String ctxPath = req.contextPath
-        String uri = req.requestURI
-        String noCtxUri = uri - ctxPath
-        boolean checkNoCtxUri = ctxPath && uri.startsWith(ctxPath)
+        String uri = UrlPathHelper.defaultInstance.getPathWithinApplication(request)
```
jdaugherty, line 93: `UrlPathHelper.getRequestUri()` consults `jakarta.servlet.include.request_uri` before it falls back to `getRequestURI()`, so this changes which URI is matched during an include sub-dispatch. Grails re-runs the interceptor chain on those (that is why `UrlMappingUtils.includeForUrl` clears `MATCHED_REQUEST` and the once-per-request filter marker before `dispatcher.include`), so an interceptor that used to see the outer page URI now sees the included one.

Measured with `match(uri: '/admin/**')` on a request for `/admin/dashboard`:

| | 8.0.x | this branch |
| --- | --- | --- |
| normal request | true | true |
| during `<g:include>` of `/stats/index` | true | **false** |

The effect is that interceptors decorating a page stop running for anything pulled in through `<g:include>`. If matching the included URI is the intent it should be deliberate and pinned by a test; otherwise this could fall back to `request.requestURI` when `WebUtils.isInclude(request)`.

jdaugherty, line 93: Separate point on the same line: `UriUtils.decode` throws `IllegalArgumentException` on a malformed escape, and `UrlPathHelper.decodeInternal` only catches `UnsupportedCharsetException`, so that exception escapes here. `canonicalizePath` in `UrlMappingMatcher` guards it, but none of the three new `getPathWithinApplication(request)` call sites do.

Request URI `/foo%`:

| | 8.0.x | this branch |
| --- | --- | --- |
| `Interceptor.doesMatch()` | false | throws `IllegalArgumentException` |
| `IpAddressFilter.doFilter` | request allowed | throws |
| compat `AntPathRequestMatcher.matches` | false | throws |

Most containers reject these before dispatch, but `doesMatch(HttpServletRequest)` is public API and is driven directly from application interceptor unit tests via `withRequest(uri: ...)`, and a security filter throwing mid-chain is a worse failure mode than not matching. The try/catch already used in `canonicalizePath` would cover all three.

grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy
```diff
@@ -89,25 +90,17 @@ trait Interceptor implements ResponseRenderer, ResponseRedirector, RequestForwar
...
-
-            if (matcher.isExclude() && matchUri && matchNoCtxUri) {
-                // Exclude interceptors are special because with only one of the conditions being false the interceptor
-                // won't be applied to the request
-                return true
-            } else if (!matcher.isExclude() && (matchUri || (checkNoCtxUri && matchNoCtxUri))) {
+            boolean matches = matcher instanceof UrlMappingMatcher
+                    ? ((UrlMappingMatcher) matcher).doesMatch(uri, grailsMappingInfo, request.method, request.contextPath)
+                    : matcher.doesMatch(uri, grailsMappingInfo, request.method)
+            if (matches) {
```
jdaugherty, line 103: Two things about this dispatch.

`grails.interceptors.Matcher` is public API, but the context-path-aware overload only exists on the internal `UrlMappingMatcher`, so a third-party `Matcher` gets no context path — and, less visibly, a different `uri` argument than before (decoded and application-relative, where it used to be the raw context-prefixed request URI). A default method on `Matcher` would keep that contract explicit instead of leaving it to an `instanceof`.

Second, this is a per-request hot path: `GrailsInterceptorHandlerInterceptorAdapter.preHandle` calls `doesMatch(request)` once per interceptor, each call does a full `getPathWithinApplication` (semicolon scan, decode, sanitize), and then every matcher does another `canonicalizePath` on top. That is N + N*M decode passes per request where there were previously none. Canonicalizing once here and passing the result down would fix both that and the double decode in `UrlMappingMatcher`. `grails-benchmarks/src/jmh/java/org/apache/grails/benchmarks/interceptors/UrlMappingMatcherBenchmark.java` is worth a run either way.

grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
```diff
@@ -99,6 +110,30 @@ class UrlMappingMatcher implements Matcher {
...
+            return '/'
+        }
+        String path = UrlPathHelper.defaultInstance.removeSemicolonContent(uri)
+        try {
+            path = UriUtils.decode(path, StandardCharsets.UTF_8)
+        } catch (IllegalArgumentException ignored) {
+            // keep the semicolon-stripped form when the URI is malformed
+        }
+        path = UrlPathHelper.defaultInstance.removeSemicolonContent(path)
+        path ?: '/'
```
jdaugherty, line 124: This decodes and strips semicolons a second time. `Interceptor.doesMatch` already hands in the output of `UrlPathHelper.getPathWithinApplication`, which has done both. So the matcher's path is not the path dispatch routes on, which is the desync this PR is closing, just moved one step along:

| request | path `UrlMappingsHandlerMapping` routes on | path the matcher sees | 8.0.x | this branch |
| --- | --- | --- | --- | --- |
| `/%2561dmin/x` vs `match(uri: '/admin/**')` | `/%61dmin/x` | `/admin/x` | false | **true** |
| `/health%3Bx` vs `matchAll().excludes(uri: '/health')` | `/health;x` | `/health` | interceptor runs | **interceptor skipped** |

The second row is the one that matters: over-normalizing on the exclude side is fail-open, and `%3B` is a literal semicolon inside a path segment rather than a matrix parameter, so stripping it after decoding is wrong independently of the double decode.

The decode here is also hardcoded to UTF-8 while `UrlPathHelper` uses `request.getCharacterEncoding()` with an ISO-8859-1 fallback, so the two passes can disagree on charset as well.

Suggest canonicalizing exactly once: either skip this for the 4-arg overload, or have `Interceptor` pass `request.requestURI` and let this method own the single decode.

grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy
```diff
@@ -99,6 +110,30 @@ class UrlMappingMatcher implements Matcher {
...
+
+    private boolean matchesPattern(String pattern, String path, String contextPath) {
+        if (pathMatcher.match(pattern, path)) {
+            return true
+        }
+        if (contextPath && contextPath != '/' && (pattern == contextPath || pattern.startsWith(contextPath + '/'))) {
+            return pathMatcher.match(pattern.substring(contextPath.length()) ?: '/', path)
+        }
+        false
+    }
```
jdaugherty, line 135: Stripping the context path off the *pattern* covers the 10857 cases, but it is not equivalent to the old "match against contextPath + path" rule for patterns that do not literally start with the context path. With `server.servlet.context-path=/app`, `match(uri: '/*/*')` matched `/app/save` before and no longer does.

That old match was accidental so I am not arguing to keep it, but it is a silent semantic change and nothing in the suite pins it either way.

grails-spring-security/compat/src/main/groovy/org/springframework/security/web/util/matcher/AntPathRequestMatcher.groovy
```diff
@@ -47,11 +48,7 @@ class AntPathRequestMatcher implements RequestMatcher {
         if (httpMethod && !httpMethod.equalsIgnoreCase(request.method)) {
             return false
         }
-        def path = request.requestURI ?: '/'
-        def contextPath = request.contextPath
-        if (contextPath && path.startsWith(contextPath)) {
-            path = path.substring(contextPath.length())
-        }
+        def path = UrlPathHelper.defaultInstance.removeSemicolonContent(UrlPathHelper.defaultInstance.getPathWithinApplication(request)) ?: '/'
```
jdaugherty, line 51: Worth noting this class is on the core security path rather than only a shim for application code: `GrailsSecurityFilterChain` and `ChannelFilterInvocationSecurityMetadataSourceFactoryBean` both resolve to it, so this changes which filter chain a request selects.

Three deltas beyond the intended fix, measured against 8.0.x with `contextPath = '/app'` and pattern `/admin/**`:

| request URI | 8.0.x | this branch |
| --- | --- | --- |
| `/app/foo%` | false | throws `IllegalArgumentException` |
| `/APP/admin/x` | false | true — `getRemainingPath` compares the context path with `ignoreCase = true` |
| `/app/public` with `include.request_uri = /app/admin/x` | false | true |

And the mirror of the last row: pattern `/public/**` against that same request goes true to false. The throw is the one I would guard first here, since it aborts chain selection rather than declining to match.

grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilter.groovy
```diff
@@ -138,6 +134,19 @@ class IpAddressFilter extends GenericFilterBean {
...
+            return urlPathHelper.removeSemicolonContent(urlPathHelper.getPathWithinApplication(new HttpServletRequestWrapper(request) {
+                @Override
+                String getRequestURI() {
+                    forwardUri
+                }
+            }))
+        }
+        urlPathHelper.removeSemicolonContent(urlPathHelper.getPathWithinApplication(request))
+    }
+
```
jdaugherty, line 149: The wrapper does not take effect when an include attribute is present: `UrlPathHelper.getRequestUri()` reads `jakarta.servlet.include.request_uri` before it calls `getRequestURI()`, so the override is never consulted and the restricted forward URI is skipped entirely.

With restriction `/admin/**`, `forward.request_uri = /admin/x`, `include.request_uri = /public` and request URI `/public`:

| | 8.0.x | this branch |
| --- | --- | --- |
| result | 404, chain not invoked | **200, chain invoked** |

Stripping the context path off `forwardUri` and running `removeSemicolonContent` plus the decode on the string directly would avoid both this and the per-request wrapper allocation. The decode needs the same `IllegalArgumentException` guard as `canonicalizePath` too: `/foo%` currently throws out of `doFilter`, where before it simply did not match.

grails-spring-security/compat/build.gradle
```diff
@@ -52,8 +52,15 @@ dependencies {
...
+    testImplementation 'jakarta.servlet:jakarta.servlet-api'
+    testImplementation 'org.springframework:spring-test'
+    testImplementation 'org.spockframework:spock-core'
+
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 }
 
 apply {
     from rootProject.layout.projectDirectory.file('gradle/docs-config.gradle')
+    from rootProject.layout.projectDirectory.file('gradle/test-config.gradle')
```
jdaugherty, line 65: `gradle/test-config.gradle` disables `Test` tasks when `onlySpringSecurityTests` is set, and that is the flag the "Spring Security Tests" workflow job runs with, so this module's new spec is skipped in that job:

```
./gradlew :grails-spring-security-compat:test -PonlySpringSecurityTests
> Task :grails-spring-security-compat:test SKIPPED
```

It still runs in the core job so the coverage is not lost, but the sibling `grails-spring-security/plugin` applies `gradle/spring-security-test-config.gradle`, which gates on `skipSpringSecurityTests` / `onlyCoreTests` instead. Using that file here would put the spec in the job that owns this code.

grails-interceptors/src/test/groovy/grails/artefact/InterceptorSpec.groovy
```diff
@@ -267,6 +267,19 @@ class InterceptorSpec extends Specification {
...
+        def webRequest = GrailsWebMockUtil.bindMockWebRequest(new MockServletContext(), new MockHttpServletRequest('', requestUri), new MockHttpServletResponse())
+        webRequest.request.setAttribute(UrlMappingsHandlerMapping.MATCHED_REQUEST, new ForwardUrlMappingInfo(controllerName: 'admin', action: 'deleteUser'))
+
+        expect:
+        interceptor.doesMatch()
+
+        where:
+        requestUri << ['/admin;x=1/deleteUser', '/%61dmin/deleteUser']
+    }
+
```
jdaugherty, line 282: The case this changes most for existing applications is not covered here: an interceptor combining `match(uri:)` with `excludes(uri:)` could never fire under a non-root context path before, because the old `isExclude() && matchUri && matchNoCtxUri` gate also required the pattern to match the context-prefixed URI.

```groovy
match(uri: '/api/**').excludes(uri: '/api/health')   // contextPath '/app', request '/app/api/orders'
// 8.0.x: false, this branch: true
```

That is the correct behaviour, but it means applications deployed at a context path start executing interceptors that have silently never run, so it deserves a case here and a release note.

Also uncovered, and both currently changing behaviour: matching during an include sub-dispatch, and a request URI with a malformed percent escape.

grails-spring-security/plugin/src/test/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilterSpec.groovy
```diff
@@ -189,4 +190,42 @@ class IpAddressFilterSpec extends AbstractUnitSpec {
...
+        request.requestURI = '/public'
+        request.setAttribute(WebUtils.FORWARD_REQUEST_URI_ATTRIBUTE, '/admin;x=1/deleteUser')
+
+        when:
+        filter.doFilter(request, response, chain)
+
+        then:
+        response.status == 404
+        0 * chain.doFilter(_, _)
+    }
```
jdaugherty, line 230: Could these grow two more cases? Both regress as things stand:

- `forward.request_uri` set together with `include.request_uri` — the forward URI is not evaluated at all, so a restricted path is allowed through.
- a request URI with a malformed escape such as `/foo%` — `doFilter` now throws instead of falling through.

jdaugherty (changes requested):
Ran this branch and its parent side by side to check the behaviour deltas. The premise holds up: `UrlMappingsHandlerMapping.getHandlerInternal()` does route on `urlHelper.getPathWithinApplication(request)` with a default `UrlPathHelper`, so aligning the matchers on that is the right target. Most of the inline notes are places where the implementation lands somewhere slightly different from that target, plus a few behaviour changes that look unintended.

The four I would want addressed before merge:

1. The path ends up canonicalized twice (`Interceptor` calls `getPathWithinApplication`, then `UrlMappingMatcher.canonicalizePath` decodes and strips semicolons again), so the matcher and dispatch still disagree. This time the matcher over-normalizes, which is fail-open on `excludes(uri:)`.
2. `UrlPathHelper` prefers `jakarta.servlet.include.request_uri`, which changes what is matched during `<g:include>` sub-dispatches for both interceptors and the compat `AntPathRequestMatcher`.
3. A malformed percent escape now throws `IllegalArgumentException` out of `Interceptor.doesMatch`, `IpAddressFilter.doFilter` and `AntPathRequestMatcher.matches`, where all three previously just failed to match.
4. In `IpAddressFilter`, the wrapper that supplies the forward URI is bypassed whenever an include attribute is present, so a restricted forwarded path can be let through.

Scope is otherwise well bounded: `matchAll()` and the `match(controller:)` convention go through `UrlMappingInfo`, so only `uri:`-based matchers are affected.

Two non-code items:

- `grails-doc/src/en/guide/theWebLayer/interceptors/interceptorMatching.adoc` still describes `uri` as just "The URI of the request". This change makes the semantics decoded, matrix-parameter-stripped and application-relative, with context-prefixed patterns accepted for backwards compatibility. That belongs in the guide.
- Interceptors that combine `match(uri:)` with `excludes(uri:)` were inert under a context path and now run, which is worth a release note.

`codeStyle` is clean on the three modules, and `:grails-interceptors:test`, `:grails-spring-security:test`, `:grails-spring-security-compat:test` plus the downstream `UriMatchingInterceptorSpec` and `BookInterceptorSpec` all pass. The suite simply does not reach any of the four items above.

jdaugherty:
All of my suggested feedback is implemented in a PR for this branch: https://github.com/jamesfredley/grails-core/pull/6 


While implementing the feedback I checked what the specifications actually say, and it changed two details:

First, Jakarta Servlet 6.0 §3.5.2 removes path parameters per segment before percent-decoding, RFC 3986 §2.2 says a reserved character and its percent-encoded form "are not equivalent", and §2.4 says implementations "must not percent-encode or decode the same string more than once". So %3B is a literal semicolon: /admin%3Bx=1/deleteUser is dispatched as /admin;x=1/deleteUser (controller admin;x=1) and never reaches AdminController. The second semicolon strip after decoding in this PR's canonicalizePath matched it against /admin/** anyway, which over-matches on match and is fail-open on excludes (/health%3Bx skipping matchAll().excludes(uri: '/health')). The follow-up removes that strip and flips the two test expectations that encoded it. The raw ;x=1 and %61 cases are unchanged and still match, which is the finding this PR set out to fix.

Second, in my review I listed the case-insensitive context path (/APP/admin/x) and the include attribute as deltas against 8.0.x, and my first cut of the follow-up had the two security matchers doing their own case-sensitive strip on requestURI to stay close to 8.0.x. I reversed that. RFC 3986 §6.2.2.1 does say paths are case-sensitive, but UrlMappingsHandlerMapping has resolved the path with Spring's UrlPathHelper.getPathWithinApplication since Grails 3, and that compares the context path case-insensitively and honors include.request_uri (Spring's ServletRequestPathUtils, which Spring Security 7's PathPatternRequestMatcher builds on, does the same). If a container ever delivers /APP/admin/x under /app, dispatch routes it to AdminController, so a matcher that strips case-sensitively selects the wrong chain, which is exactly the desync this PR is about. Consistency with dispatch is the security property; RFC parity on its own isn't. All three matchers now call getPathWithinApplication, and the two deviations from the RFC are documented as deliberate, with the reason, in the Interceptor Javadoc and the upgrade note.

One more correction to my own review: the "interceptors that silently never ran under a context path" case is narrower than I described. Only a matcher pairing a uri pattern with an excludes(...) was inert; matchAll() with excludes, name-based matchers and match(uri:) without excludes always ran. The release note says so and both variants are pinned.

matrei:
## AI Review Findings

Reviewed together with the follow-up at https://github.com/jamesfredley/grails-core/pull/6 (three commits on top of `f0108c28ed`). The premise of this PR is right: `UrlMappingsHandlerMapping.getHandlerInternal()` routes on a default `UrlPathHelper.getPathWithinApplication(request)`, so URI matchers must see that same path. The PR head approximates that path instead of reusing it, and the approximation reopens the same class of inconsistency. The follow-up fixes every item below and should be merged into this branch before this PR merges.

Verified against the Spring 7.0.9 source: `UrlPathHelper.decodeAndCleanUriString` runs `removeSemicolonContent`, then `decodeRequestString`, then `getSanitizedPath`, and `getRequestUri` reads `jakarta.servlet.include.request_uri` before falling back to `getRequestURI()`.

### High: The path is canonicalized twice, which is fail-open on `excludes(uri:)`

References:

- `grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy:93`
- `grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy:113-124`

`Interceptor.doesMatch` passes the output of `getPathWithinApplication` to `UrlMappingMatcher`, and `canonicalizePath` then decodes it again and strips semicolons again. Dispatch removes path parameters before decoding, so `/health%3Bx` is dispatched as `/health;x`, but the matcher sees `/health` and `matchAll().excludes(uri: '/health')` skips the interceptor for a path that is not `/health`.

| request | dispatched path | matcher path (PR head) |
|---|---|---|
| `/health%3Bx` | `/health;x` | `/health` |
| `/%2561dmin/x` | `/%61dmin/x` | `/admin/x` |

Fixed in the follow-up by canonicalizing once in `Interceptor.doesMatch` and passing the result to every matcher through a new default method `Matcher.doesMatch(uri, info, method, contextPath)`. `UrlMappingMatcher` no longer decodes or strips anything.

### High: An encoded semicolon is treated as a matrix-parameter delimiter

References:

- `grails-interceptors/src/main/groovy/org/grails/plugins/web/interceptors/UrlMappingMatcher.groovy:121`
- `grails-spring-security/compat/src/test/groovy/org/springframework/security/web/util/matcher/AntPathRequestMatcherSpec.groovy:38`
- `grails-spring-security/plugin/src/test/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilterSpec.groovy:212`

Two test rows pin `/admin%3Bx=1/deleteUser` as matching `/admin/**`. Jakarta Servlet 6.0 §3.5.2 removes path parameters per segment before percent-decoding and RFC 3986 §2.2 says a reserved character and its escape are not equivalent, so that request is dispatched to controller `admin;x=1`, which does not exist. Matching it against `/admin/**` over-matches on `match` and is fail-open on `excludes`. The follow-up flips both expectations. The raw `;x=1` and `%61` rows, which are the finding this PR set out to fix, are unchanged.

### High: A malformed percent escape throws out of all three matchers

References:

- `grails-interceptors/src/main/groovy/grails/artefact/Interceptor.groovy:93`
- `grails-spring-security/compat/src/main/groovy/org/springframework/security/web/util/matcher/AntPathRequestMatcher.groovy:51`
- `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilter.groovy:140-149`

`UriUtils.decode` throws `IllegalArgumentException` on `/foo%`, and `UrlPathHelper.decodeInternal` only catches `UnsupportedCharsetException`. None of the three new `getPathWithinApplication` call sites guard it, so `Interceptor.doesMatch`, `IpAddressFilter.doFilter` and `AntPathRequestMatcher.matches` all throw where 8.0.x declined to match. `doesMatch(HttpServletRequest)` is public API driven from application unit tests, and a security filter throwing mid-chain is a worse failure mode than not matching. The follow-up falls back to a `UrlPathHelper` with `urlDecode = false` at all three sites.

### High: `IpAddressFilter` forward wrapper is bypassed by an include attribute

References:

- `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilter.groovy:137-149`

The wrapper overrides `getRequestURI()`, but `UrlPathHelper.getRequestUri` reads `jakarta.servlet.include.request_uri` first, so when an include attribute is present the forward URI is never consulted and a restricted forwarded path is let through. With restriction `/admin/**`, `forward.request_uri = /admin/x`, `include.request_uri = /public`: 8.0.x returns 404, this branch invokes the chain. The follow-up canonicalizes the forward URI string directly (semicolon content removed, decoded once, forward context path stripped case-insensitively) with no per-request wrapper.

### Medium: Compat spec is skipped in the Spring Security CI job

References:

- `grails-spring-security/compat/build.gradle:65`

`gradle/test-config.gradle` disables `Test` tasks under `-PonlySpringSecurityTests`, which is what the Spring Security workflow job runs with, so the new `AntPathRequestMatcherSpec` only runs in the core job. The follow-up switches the module to `gradle/spring-security-test-config.gradle` and guards that script's `integrationTest` block with `pluginManager.withPlugin(...)` so a library module can apply it.

### Medium: Documentation

References:

- `grails-doc/src/en/guide/theWebLayer/interceptors/interceptorMatching.adoc:67`
- `grails-doc/src/en/guide/upgrading/upgrading80x.adoc`

The guide still describes `uri` as "The URI of the request". The semantics are now decoded, matrix-parameter-stripped and application-relative, with context-prefixed patterns accepted for compatibility. Two behaviour changes need an upgrade note: interceptors pairing `match(uri:)` with any `excludes(...)` were inert under a non-root context path and now run, and `match(uri: '/*/*')` no longer matches `/app/save` under context path `/app`. The follow-up adds both, plus `ip.adoc` coverage.

## Follow-up (jamesfredley#6) findings

All three module suites pass on the follow-up head:

| Module | Tests | Failures |
|---|---|---|
| grails-interceptors | 89 | 0 |
| grails-spring-security-compat | 16 | 0 |
| grails-spring-security (plugin) | 215 | 0 |

### Medium: Upgrade note number collides after rebase

References:

- `grails-doc/src/en/guide/upgrading/upgrading80x.adoc:3101`

The PR base is 116 commits behind 8.0.x, which now has sections through 57. A trial merge is conflict-free but leaves two sections numbered 54. Renumber to 58 when rebasing.

### Low: Forwarded `ipRestrictions` under a context path should be in the upgrade note

References:

- `grails-doc/src/en/guide/security/securityPlugins/springSecurity/core/ip.adoc:51`
- `grails-spring-security/plugin/src/main/groovy/grails/plugin/springsecurity/web/filter/IpAddressFilter.groovy:146-166`

8.0.x never stripped the context path from `forward.request_uri`, so a forwarded restricted path under a context path was fail-open. The follow-up enforces it, which is right, but it is a behaviour change on a security filter and neither doc mentions it.

### Low: Nits

- `Interceptor.doesMatch` allocates a new `UrlPathHelper` on every malformed-escape fallback; the compat matcher and the IP filter hold a raw helper once. Do the same here.
- The forward branch in `IpAddressFilter.getPathWithinApplication` skips `getSanitizedPath`, so `//` is not collapsed for forwarded URIs. Harmless because `AntPathMatcher` ignores empty segments, but it is the one place the forward path deviates from the direct path.
- `bindRequest` sits between feature methods in `InterceptorSpec`; move it to the bottom with the other helpers.
- The new compat spec pins that `AntPathRequestMatcher` is case-insensitive by default. That is pre-existing, but Spring Security's original defaulted to case-sensitive, and this class selects filter chains. Worth a deliberate decision.

## Questions for maintainers

- This PR targets 8.0.x only. Is a backport to the 7.x lines intended?
- Includes are now matched on the included path, so `match(uri: '/admin/**')` no longer fires for `<g:include>` sub-dispatches inside an admin page. The upgrade note covers it and it is consistent with `match(controller:)`, but it is the one change that could surprise an application relying on the old double-fire.

jamesfredley:
Review follow-up that was mistakenly opened on the archive fork is now https://github.com/apache/grails-core/pull/16330

jamesfredley:
This PR now includes jdaugherty's review follow-up (previously jamesfredley/grails-core#6 / apache/grails-core#16330).

jamesfredley:
Superseded by https://github.com/apache/grails-core/pull/16332 so the head branch is on apache/grails-core (not the archive fork).

PR #16316, opened 2026-09-04
test(grails-datamapping-core): restore the transaction service specs lost in the 7.2.x merge

sbglasius:
## Problem

The 7.2.x → 8.0.x merge hit an `add/add` conflict on `DefaultTransactionServiceSpec` — both branches had independently created the file. The resolution kept 8.0.x's version, which had **1** test, and dropped 7.2.x's **14**.

**The drop was accidental, not a scoping decision.** The resolution was half-applied: it took 7.2.x's four Spring transaction imports but not the test methods that use them, leaving `PlatformTransactionManager`, `TransactionDefinition`, `TransactionStatus` and `TransactionSystemException` imported and unused on `8.0.x`. Pre-merge `8.0.x` had neither the imports nor the tests, so they arrived together and only half survived.

## Changes

Restores 7.2.x's suite and keeps 8.0.x's accessor test — 15 tests:

- the full `withTransaction` / `withRollback` / `withNewTransaction` matrix across the `Closure`, `TransactionDefinition` and `Map` overloads, each with its "datastore does not support transactions" counterpart
- `getDatastore/setDatastore round-trip`, covering the explicit `datastore` field `8.0.x` added to `DefaultTransactionService`

Also drops the class-level comment claiming the overloads were out of scope — untrue once the suite is restored, and it referenced "this PR" and "item 14" of a plan that is not in the repository.

## Why the suite ports unchanged

`DefaultTransactionService` differs between the two branches *only* by that added field and its accessors. Every method the restored tests exercise is byte-identical between 7.2.x and 8.0.x, and 7.2.x's property-style `transactionService.datastore = ...` resolves to the new setter.

## Verification

```
./gradlew :grails-datamapping-core:test --tests "org.grails.datastore.gorm.services.DefaultTransactionServiceSpec"
```

All 15 pass. `:grails-datamapping-core:codeStyle` is clean and every import is used.

`DefaultTenantServiceSpec`, the other `add/add` in that merge, was resolved cleanly (only a `{ }` → `{}` whitespace change) and is left alone — its class genuinely diverged on 8.0.x with the `RESOLVING` reentrancy guard, so 8.0.x's version is the correct one there.

### DIFF
--- a/grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/services/DefaultTransactionServiceSpec.groovy
+++ b/grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/services/DefaultTransactionServiceSpec.groovy
@@ -24,27 +24,174 @@ import org.springframework.transaction.TransactionStatus
 import org.springframework.transaction.TransactionSystemException
 
 import org.grails.datastore.mapping.core.Datastore
+import org.grails.datastore.mapping.transactions.CustomizableRollbackTransactionAttribute
+import org.grails.datastore.mapping.transactions.TransactionCapableDatastore
 import spock.lang.Specification
 
-/**
- * The diff only adds the explicit {@code datastore}/{@code getDatastore()}/{@code setDatastore()}
- * {@code Service} contract (previously implicit/inherited) - the rest of this class (the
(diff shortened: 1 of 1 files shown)


jdaugherty (approved):

jdaugherty:
the failure are related to jboss / logger

PR #16322, opened 2026-09-07
M6 Bug Fix generated GSP views: scaffolded count model field type and welcome page static compilation

codeconsole:
## Description

Two defects that stop a freshly generated app from working. Both are in the generated GSP views, and both surface the moment you use the app the generator produces.

### 1. Scaffolded `index.gsp` declares the count model field as `Integer`

`grails generate-views` produces an `index.gsp` whose typed model directive declares:

```
@{ model="List<website.Sample> sampleList; Integer sampleCount" }
```

but the `Service.groovy` template declares `Long count()`, and `Controller.groovy` passes that value straight through as the `${propertyName}Count` model entry. Model fields are populated by reflective `Field.set`, which does no numeric conversion, so **every scaffolded index view throws on first render**:

```
java.lang.IllegalArgumentException: Can not set java.lang.Integer field
  ..._views_sample_index_gsp.sampleCount to java.lang.Long
    at org.grails.gsp.GroovyPage.applyModelFieldsFromBinding(GroovyPage.java:185)
```

The template now declares the count as `Number`. The generated service supplies a `Long`, but `static scaffold = X` goes through `RestfulController.countResources()`, which supplies an `Integer`, so no single concrete type fits both.

While tracking this down, the binding itself turned out to be stricter than the rest of the page. `<g:set type="int" .../>` converts, and so does a plain Groovy assignment, but a model field declared `Integer` failed outright when given a `Long`, with a bare JDK reflection message naming the generated class. Model values are now converted to the declared type the way a Groovy assignment converts them, so existing views that declare `Integer bookCount` keep working.

`DefaultTypeTransformation.castToType` on its own wraps `3_000_000_000L` into `-1294967296` and turns `42.9` into `42`, silently. So a conversion that would change a numeric value is refused, as is a value that cannot be converted at all, and both name the field and the two types:

```
GroovyPagesException: Model field 'bookCount' is declared as java.lang.Integer, which cannot
  hold the java.lang.Long 3000000000 the model supplied without changing it.
```

### 2. Welcome page does not compile under GSP static compilation

Adding the following to a generated app's `build.gradle` fails `compileGroovyPages`:

```groovy
grails {
    compileStatic {
        all = true
        gsp = true
    }
}
```

```
[Static type checking] - Cannot find matching method java.lang.Object#toLowerCase()
[Static type checking] - No such property: name for class: java.lang.Object
```

Three `sort` closures in the welcome page take untyped parameters, so static type checking infers `Object` for the element and rejects the calls on it. Each loses the element type for a different reason:

| Site | Why the element is `Object` |
| --- | --- |
| plugin list | `.collect { p, i -> [plugin: p, order: ...] }` — map literal values are `Object` |
| `domainsByPlugin` | the `groupBy` closure returns a `def` local, so the key is `Object` |
| `mimeTypes` | `applicationContext.getBean('mimeTypes')` returns `Object` |

Fixed by typing the closure parameters and converting the values whose static type is `Object`. The `appListeners` sort is typed the same way for consistency; its map values are all `String`, so it compiled without that change. The plugin and mime type comparisons need the concrete element type, so those are cast.

One wrinkle worth recording: `grails.plugins.GrailsPlugin` cannot be imported into a GSP, because the compiler already auto-imports the unrelated `grails.plugins.metadata.GrailsPlugin` annotation and the collision fails the build with `The name GrailsPlugin is already declared`. The cast uses the qualified name instead.

`grails-forge-core`'s resource copy and the web profile skeleton copy of this page were byte-identical, so both are updated and remain identical.

### Keeping both from coming back

CI was green on `8.0.x` with both defects present. `gsp-compile-static` now compiles the web profile's welcome page statically alongside its own pages (the forge copy is pinned to the profile copy by `GrailsGspSpec`), and `ScaffoldedIndexViewModelSpec` expands the scaffolded `index.gsp` and renders its model declaration with an `Integer`, a `Long`, and a count past `Integer.MAX_VALUE`.

### DIFF
--- a/.agents/skills/gradle-developer/SKILL.md
+++ b/.agents/skills/gradle-developer/SKILL.md
@@ -288,7 +288,7 @@ For POM property generation, **map key must be the dependency name prefix**:
 
 ```groovy
 bomDependencyVersions = [
-    'groovy.version': '5.0.7',
+    'groovy.version': '5.1.2',
 ]
 bomDependencies = [
     'groovy': "org.apache.groovy:groovy:${bomDependencyVersions['groovy.version']}",
--- a/.agents/skills/grails-developer/SKILL.md
+++ b/.agents/skills/grails-developer/SKILL.md
@@ -33,7 +33,7 @@ Activate this skill when developing with current Grails, including:
 Current Grails is built on:
 - **Spring Boot**: 4.1.x
 - **Spring Framework**: 7.0.x
-- **Groovy**: 5.0.x
+- **Groovy**: 5.1.x
 - **Gradle**: 9.6.x
 - **Spock**: 2.4-groovy-5.0
 - **Jakarta EE**: 10 (migrated from javax.*)
@@ -865,6 +865,6 @@ grails.plugin.springsecurity.controllerAnnotations.staticRules = [
 - **Grails User Guide**: https://grails.apache.org/docs/latest/guide/single.html
 - **GORM Documentation**: https://grails.apache.org/docs/latest/grails-data/
 - **Grails Plugins**: https://grails.apache.org/plugins.html
-- **Groovy 5 Documentation**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.0.x)
+- **Groovy 5 Documentation**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.1.x)
 - **Spock Framework**: https://spockframework.org/spock/docs/2.4/all_in_one.html
 - **Geb Manual**: https://groovy.apache.org/geb/manual/current/
--- a/.agents/skills/groovy-developer/SKILL.md
+++ b/.agents/skills/groovy-developer/SKILL.md
@@ -11,7 +11,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more contributor l
 
 ## What I Do
 
-- Provide deep expertise in Groovy 5.0, including core language features: syntax, closures, traits, GStrings, operators, and Groovy Truth.
+- Provide deep expertise in Groovy 5.1, including core language features: syntax, closures, traits, GStrings, operators, and Groovy Truth.
 - Guide metaprogramming techniques: runtime (categories, ExpandoMetaClass, extension modules) and compile-time (AST transformations like @Immutable, @Builder, @Delegate, @Singleton, @Log, @Memoized).
 - Assist with DSL creation and usage (MarkupBuilder, JsonBuilder, builders for configuration).
 - Support testing with Spock Framework (BDD, data-driven tests, mocking) and built-in power assertions.
@@ -27,7 +27,7 @@ Activate this skill for any Groovy-related task, including:
 - Building or enhancing Gradle build scripts (build.gradle).
 - Creating DSLs for configuration, testing, or business rules.
 - Writing Spock specifications for unit and integration tests.
-- Migrating from older Groovy versions (2.x/3.x/4.x) to 5.0.
+- Migrating from older Groovy versions (2.x/3.x/4.x) to 5.1.
 - Performance optimization with static compilation.
 - Working with JSON/XML data, GORM criteria queries, or builders.
 
@@ -509,7 +509,7 @@ sql.execute("INSERT INTO books (title) VALUES ($title)")
 
 ## Resources
 
-- **Groovy 5 Documentation**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.0.x)
+- **Groovy 5 Documentation**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.1.x)
 - **Groovy Style Guide**: https://groovy-lang.org/style-guide.html
 - **GORM Documentation**: https://grails.apache.org/docs/latest/grails-data/
 - **Spock Framework**: https://spockframework.org/spock/docs/2.4/all_in_one.html
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -88,7 +88,7 @@ export GRADLE_OPTS="-Xms2G -Xmx5G"
 | Component | Version |
 |-----------|---------|
 | JDK | 21+ (baseline 21) |
-| Groovy | 5.0.x |
+| Groovy | 5.1.x |
 | Spring Boot | 4.1.x |
 | Spring Framework | 7.0.x |
 | Spock | 2.4-groovy-5.0 |
@@ -309,7 +309,7 @@ and known non-findings — before reporting issues.
 ## Resources
 
 - **Grails Guide**: https://grails.apache.org/docs/latest/guide/single.html
-- **Groovy 5 Docs**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.0.x)
+- **Groovy 5 Docs**: https://groovy-lang.org/documentation.html#all-versions (select latest 5.1.x)
 - **Spock 2.4 Docs**: https://spockframework.org/spock/docs/2.4/all_in_one.html
 - **GORM Docs**: https://grails.apache.org/docs/latest/grails-data/
 - **Issues**: https://github.com/apache/grails-core/issues
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -84,7 +84,7 @@ ext {
             // extended-scalars compatibility. See https://github.com/apache/grails-core/issues/15674
             'graphql-java.version'          : '25.0',
             'graphql-java-extended-scalars.version': '24.0',
-            'groovy.version'                : '5.1.0',
+            'groovy.version'                : '5.1.2',
             'guava.version'                 : '33.7.1-jre',
             // Groovy YAML 5.1.0 requires a minimum of Jackson 2.22.1, and 2.22.2 also overrides the
             // 2.21.5 Spring Boot 4.1 manages: it adds the StreamReadConstraints length limit to
@@ -468,7 +468,7 @@ ext {
                 'liquibase-hibernate.version': '4.27.0',
                 'liquibase.version'          : '4.27.0',
                 'hibernate.version'          : '5.6.15.Final',
-                'groovy.version'  : '5.1.0',
+                'groovy.version'  : '5.1.2',
                 'spock.version'   : '2.4-groovy-5.0',
                 'protobuf.version': '4.30.2',
                 'asm.version'     : '9.10.1',
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
@@ -1,4 +1,5 @@
 <%@ page import="grails.util.Environment"%>
+<%@ page import="grails.web.mime.MimeType"%>
 <%@ page import="org.springframework.boot.SpringBootVersion"%>
 <%@ page import="org.springframework.core.SpringVersion"%>
 <%@ page import="org.springframework.util.ClassUtils"%>
@@ -1341,7 +1342,7 @@
                 <g:set var="mimeTypes"
                        value="${applicationContext.containsBean('mimeTypes') ?
                                applicationContext.getBean('mimeTypes').toList()
-                                       .sort { grails.web.mime.MimeType a, grails.web.mime.MimeType b -> ((a.extension ?: '').toLowerCase() <=> (b.extension ?: '').toLowerCase()) ?: (a.name <=> b.name) } : []}"/>
+                                       .sort { MimeType a, MimeType b -> ((a.extension ?: '').toLowerCase() <=> (b.extension ?: '').toLowerCase()) ?: (a.name <=> b.name) } : []}"/>
                 <div class="card border-1 shadow-sm mt-4">
                     <div class="card-body">
                         <div class="d-flex align-items-center justify-content-between mb-3">
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
@@ -9,7 +9,7 @@
        value="${pluginManager.allPlugins.toList()
                .withIndex()
                .collect { p, i -> [plugin: p, order: ((int) i) + 1] }
-               .sort { Map a, Map b -> ((grails.plugins.GrailsPlugin) a.plugin).name.toLowerCase() <=> ((grails.plugins.GrailsPlugin) b.plugin).name.toLowerCase() }}"
+               .sort { Map row -> ((grails.plugins.GrailsPlugin) row.plugin).name.toLowerCase() }}"
 />
 <g:def type="int" var="numControllers" value="${grailsApplication.controllerClasses.size()}"/>
 <g:def type="int" var="numDomains" value="${grailsApplication.domainClasses.size()}"/>
@@ -456,11 +456,11 @@
                         <g:set var="domainsByPlugin"
                                value="${grailsApplication.domainClasses.toList()
                                        .groupBy { dc ->
-                                           def plugin = null
-                                           try { plugin = pluginManager.getPluginForClass(dc.clazz) } catch (Throwable ignored) { }
-                                           plugin?.name ?: ''
+                                           String pluginName = ''
+                                           try { pluginName = pluginManager.getPluginForClass(dc.clazz)?.name ?: '' } catch (Throwable ignored) { }
+                                           pluginName
                                        }
-                                       .sort { Map.Entry a, Map.Entry b -> a.key.toString().toLowerCase() <=> b.key.toString().toLowerCase() }}"/>
+                                       .sort { it.key.toLowerCase() }}"/>
                         <div id="domains-list">
                             <g:each var="pEntry" in="${domainsByPlugin}" status="pIndex">
                                 <div class="${pIndex > 0 ? 'mt-4' : ''}" data-filter-group>
@@ -567,7 +567,7 @@
                                .collect { l -> [name: (l.getClass().simpleName ?: l.getClass().name.tokenize('.').last()),
                                                 packageName: (l.getClass().package?.name ?: ''),
                                                 detail: l.toString()] }
-                               .sort { Map a, Map b -> (a.name.toString().toLowerCase() <=> b.name.toString().toLowerCase()) ?: (a.detail.toString() <=> b.detail.toString()) }}"/>
+                               .sort { Map<String, String> a, Map<String, String> b -> (a.name.toLowerCase() <=> b.name.toLowerCase()) ?: (a.detail <=> b.detail) }}"/>
                 <g:set var="bindingGroups"
                        value="${[[code: 'welcome.binding.value', beans: applicationContext.getBeansOfType(grails.databinding.converters.ValueConverter)],
                                  [code: 'welcome.binding.formatted', beans: applicationContext.getBeansOfType(grails.databinding.converters.FormattedValueConverter)],
--- a/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
+++ b/grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp
@@ -8,7 +8,7 @@
        value="${pluginManager.allPlugins.toList()
                .withIndex()
                .collect { p, i -> [plugin: p, order: ((int) i) + 1] }
-               .sort { a, b -> a.plugin.name.toLowerCase() <=> b.plugin.name.toLowerCase() }}"
+               .sort { Map a, Map b -> ((grails.plugins.GrailsPlugin) a.plugin).name.toLowerCase() <=> ((grails.plugins.GrailsPlugin) b.plugin).name.toLowerCase() }}"
 />
 <g:def type="int" var="numControllers" value="${grailsApplication.controllerClasses.size()}"/>
 <g:def type="int" var="numDomains" value="${grailsApplication.domainClasses.size()}"/>
@@ -459,7 +459,7 @@
                                            try { plugin = pluginManager.getPluginForClass(dc.clazz) } catch (Throwable ignored) { }
                                            plugin?.name ?: ''
                                        }
-                                       .sort { a, b -> a.key.toLowerCase() <=> b.key.toLowerCase() }}"/>
+                                       .sort { Map.Entry a, Map.Entry b -> a.key.toString().toLowerCase() <=> b.key.toString().toLowerCase() }}"/>
                         <div id="domains-list">
                             <g:each var="pEntry" in="${domainsByPlugin}" status="pIndex">
                                 <div class="${pIndex > 0 ? 'mt-4' : ''}" data-filter-group>
@@ -566,7 +566,7 @@
                                .collect { l -> [name: (l.getClass().simpleName ?: l.getClass().name.tokenize('.').last()),
                                                 packageName: (l.getClass().package?.name ?: ''),
                                                 detail: l.toString()] }
-                               .sort { a, b -> (a.name.toLowerCase() <=> b.name.toLowerCase()) ?: (a.detail <=> b.detail) }}"/>
+                               .sort { Map a, Map b -> (a.name.toString().toLowerCase() <=> b.name.toString().toLowerCase()) ?: (a.detail.toString() <=> b.detail.toString()) }}"/>
                 <g:set var="bindingGroups"
                        value="${[[code: 'welcome.binding.value', beans: applicationContext.getBeansOfType(grails.databinding.converters.ValueConverter)],
                                  [code: 'welcome.binding.formatted', beans: applicationContext.getBeansOfType(grails.databinding.converters.FormattedValueConverter)],
@@ -1341,7 +1341,7 @@
                 <g:set var="mimeTypes"
                        value="${applicationContext.containsBean('mimeTypes') ?
                                applicationContext.getBean('mimeTypes').toList()
-                                       .sort { a, b -> ((a.extension ?: '').toLowerCase() <=> (b.extension ?: '').toLowerCase()) ?: (a.name <=> b.name) } : []}"/>
+                                       .sort { grails.web.mime.MimeType a, grails.web.mime.MimeType b -> ((a.extension ?: '').toLowerCase() <=> (b.extension ?: '').toLowerCase()) ?: (a.name <=> b.name) } : []}"/>
                 <div class="card border-1 shadow-sm mt-4">
                     <div class="card-body">
                         <div class="d-flex align-items-center justify-content-between mb-3">
--- a/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java
+++ b/grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java
@@ -179,11 +179,17 @@ public abstract class GroovyPage extends Script {
 
     private void applyModelFieldsFromBinding(Iterable<Field> modelFields) {
         for (Field field : modelFields) {
+            Object value = getProperty(field.getName());
+            if (value == null) {
+                continue;
+            }
             try {
-                Object value = getProperty(field.getName());
-                if (value != null) {
-                    field.set(this, value);
-                }
+                field.set(this, value);
+            } catch (IllegalArgumentException e) {
+                throw new GroovyPagesException("Model field '" + field.getName() + "' is declared as " +
+                        field.getType().getName() + " but the model supplied an instance of " +
+                        value.getClass().getName() + ". Declare the field with a type the model value is " +
+                        "assignable to; model values are not coerced.", e, -1, getGroovyPageFileName());
             } catch (IllegalAccessException e) {
                 throw new GroovyPagesException("Error setting model field '" + field.getName() + "'", e, -1, getGroovyPageFileName());
             }
--- a/grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy
+++ b/grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy
@@ -238,6 +238,27 @@ Date d4=new Date(123L)
         rendered == '123-456-789-123'
     }
 
+    def "model field is applied when the model supplies the declared type"() {
+        given:
+        def template = '''@{ model="Long sampleCount"}${sampleCount}'''
+        when:
+        def rendered = renderTemplate(template, [sampleCount: 42L], true)
+        then:
+        rendered == '42'
+    }
+
+    def "a model value of the wrong type names the field and both types"() {
+        given:
+        def template = '''@{ model="Integer sampleCount"}${sampleCount}'''
+        when:
+        renderTemplate(template, [sampleCount: 42L], true)
+        then:
+        GroovyPagesException e = thrown()
+        e.message.contains("Model field 'sampleCount'")
+        e.message.contains('java.lang.Integer')
+        e.message.contains('java.lang.Long')
+    }
+
     def renderTemplate(templateSource, model, expectedCompileStaticMode, printSource = false) {
         def t = gpte.createTemplate(templateSource, "template${templateSource.hashCode()}")
         assert t.metaInfo.compilationException == null
--- a/grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy
+++ b/grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy
@@ -247,6 +247,19 @@ Date d4=new Date(123L)
         rendered == '42'
     }
 
+    def "a Number model field accepts every numeric type a controller may supply"() {
+        given:
+        def template = '''@{ model="Number sampleCount"}${sampleCount}'''
+        when:
+        def rendered = renderTemplate(template, [sampleCount: supplied], true)
+        then:
+        rendered == '42'
+        where:
+        // scaffolded controllers supply Long via the generated service and Integer
+        // via RestfulController.countResources()
+        supplied << [42 as Integer, 42L, 42 as Short, 42 as BigInteger]
+    }
+
     def "a model value of the wrong type names the field and both types"() {
         given:
         def template = '''@{ model="Integer sampleCount"}${sampleCount}'''
--- a/grails-profiles/web/skeleton/grails-app/views/index.gsp
+++ b/grails-profiles/web/skeleton/grails-app/views/index.gsp
@@ -1,4 +1,5 @@
 <%@ page import="grails.util.Environment"%>
+<%@ page import="grails.web.mime.MimeType"%>
 <%@ page import="org.springframework.boot.SpringBootVersion"%>
 <%@ page import="org.springframework.core.SpringVersion"%>
 <%@ page import="org.springframework.util.ClassUtils"%>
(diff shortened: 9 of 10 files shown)


jdaugherty (approved):

matrei:
## AI Review Findings

Head `8855b88340`, base `8.0.x` at `0980623481`. The head is 103 commits behind the base but merges clean (`git merge-tree` reports no conflicts). Both defects in the description are real and both fixes work:

- Compiling the base-branch welcome page statically (copied into `grails-test-examples/gsp-compile-static`) fails with 4 type-checking errors at three sites, the plugins sort, the domains-by-plugin sort and the mime-types sort. The PR's page compiles, and the compiled class extends `CompileStaticGroovyPage`, so it really was compiled statically rather than falling back.
- The scaffolded `index.gsp` expanded for a domain class compiles statically with `Number` and the `${count} > params.int('max')` comparison type-checks.
- `GspCompileStaticSpec`: 30 tests pass. Checkstyle on `grails-gsp-core` main: 0 violations.

The findings below are about the choice the `GroovyPage` change makes, and about what would keep either bug from coming back. None of them changes the welcome page hunks.

### [P2] Assign model fields the way Groovy assigns, instead of declaring that model values are not coerced

References:

- `grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java:180-197` (`applyModelFieldsFromBinding`)
- `grails-scaffolding/src/main/templates/scaffolding/index.gsp:1`

The new message states a rule, "model values are not coerced", that is the opposite of what the rest of the page does. `<g:set type="int" var="total" value="${books.size()}"/>` converts (the static compilation guide says so explicitly), and plain Groovy converts too:

```groovy
Integer i = 42L      // 42, java.lang.Integer
int p = 42L          // 42
String s = "${40+2}" // GString assigned to String
```

A model field declared `Integer bookCount` is the one place in a page where the same assignment throws. That declaration is also the natural one to write: GORM's `count()` returns `Integer` (`GormEntity.groovy:732`), the generated service returns `Long`, and `RestfulController.countResources()` returns `Integer`, so any page author who declares the count with the type of whichever source they looked at has a coin-flip chance of a render failure. `String` fields fed a GString from a controller fail the same way today.

`DefaultTypeTransformation.castToType` is Groovy's own assignment conversion, so it gives exactly the semantics a `def`-free Groovy assignment would:

```java
try {
    field.set(this, DefaultTypeTransformation.castToType(value, field.getType()));
} catch (IllegalArgumentException | GroovyCastException e) {
    throw new GroovyPagesException("Model field '" + field.getName() + "' is declared as " +
            field.getType().getName() + " but the model supplied an instance of " +
            value.getClass().getName() + '.', e, -1, getGroovyPageFileName());
}
```

I ran the PR's spec with that in place plus a data-driven case:

| declared | supplied | result |
|---|---|---|
| `Integer` | `42L` | renders `42` |
| `int` | `42L` | renders `42` |
| `Long` | `42` | renders `42` |
| `long` | `42 as Short` | renders `42` |
| `String` | `"${40 + 2}"` (GString) | renders `42` |
| `Number` | `42 as BigInteger` | renders `42` |
| `Integer` | `new Date()` | `GroovyPagesException` naming the field, cause `GroovyCastException` |

The only existing test that changes is *a model value of the wrong type names the field and both types*, which would flip from `Integer`/`Long` to a pair that genuinely cannot be converted, `Integer`/`Date` say. Everything else in the spec passes unchanged. The prototype is not in the working tree.

The `Number` change in the scaffold template is still the honest type for a value that is `Long` from one controller and `Integer` from the other, so keep it either way. With coercion it stops being load-bearing, and every already-generated `Integer` view in existing 8.0.x applications starts working as well, which the template change alone does not give them.

### [P2] Document what a declared model field does with a value of another type

References:

- `grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc:28-56` (Declaring the Model)
- `grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc:208`

The guide says nothing about what happens when the supplied value's type differs from the declaration. The only nearby sentence, at line 208, covers the framework-supplied names and says they fail with a `GroovyCastException`. Whichever way the finding above is decided, that section needs one sentence: either "a model value is converted the way a Groovy assignment converts it, so a `Long` count satisfies an `Integer` field" or "a model value must be assignable to the declared type; a mismatch fails at render naming the field". The behaviour is user-facing and is exactly what this PR changes, so this is the doc coverage the contributing rules ask for.

### [P2] Neither failure had a test that could catch it, and this PR adds coverage for the engine but not for the two pages

References:

- `grails-test-examples/gsp-compile-static/build.gradle`
- `grails-test-examples/scaffolding-fields/grails-app/controllers/scaffoldingfields/EmployeeController.groovy:32`
- `grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy:241-274`

CI was green on `8.0.x` with both bugs present, and stays green if either regresses:

- Nothing compiles the welcome page statically. `gsp-compile-static` has its own small pages only. A `Copy` step that drops `grails-profiles/web/skeleton/grails-app/views/index.gsp` into that app's views ahead of `compileGroovyPages` is the check I ran by hand above, and it fails on the base branch with the four errors. The forge copy is already pinned to the profile copy by *test the profile skeleton mirrors the forge welcome templates* in `GrailsGspSpec`, so one copy is enough.
- Nothing renders a service-backed scaffolded index. `scaffolding-fields` and `hyphenated` use `static scaffold = Domain`, which goes through `RestfulServiceController.countResources()` and supplies an `Integer`, so the generated-service `Long` path is never rendered. The new spec cases prove the engine accepts a `Long` for a `Number` field; they do not prove the template declares `Number`. A test that expands `scaffolding/index.gsp` for a domain class and renders it with `[bookList: [], bookCount: 3L]` would pin the template itself.

### Nit: the listener sort hunk is not needed

References:

- `grails-profiles/web/skeleton/grails-app/views/index.gsp:570`
- `grails-forge/grails-forge-core/src/main/resources/gsp/index.gsp:570`

`Map<String, String> a, Map<String, String> b` on the listener sort compiles either way: the `collect` above it builds maps whose values are all `String`, so `a.name` already infers as `String`. I reverted that one hunk on the PR's page and it compiled. The description says four closures fail; three do. Harmless, and fine to keep for symmetry with the other three.

### Nit: the scaffolding guide describes model names that the templates no longer use

References:

- `grails-doc/src/en/guide/scaffolding.adoc:191`

Pre-existing and out of scope, noting it because it is the doc a reader would go to for the count field: it says the standard views expect `<propertyName>InstanceList` and `<propertyName>Instance`, while the templates bind `<propertyName>List` and `<propertyName>Count`.

### Confirmed

- The `grails.plugins.GrailsPlugin` fully-qualified cast is needed: `GroovyPageParser.DEFAULT_IMPORTS` imports `grails.plugins.metadata.GrailsPlugin` into every page, so a page import of the plugin interface would clash.
- The rewritten sorts keep their ordering. The one-argument `sort` on the plugin rows and on the `domainsByPlugin` entries orders by the same lower-cased key the comparators used, and `List.sort(Closure)` still sorts in place as before.
- The `groupBy` closure now returns a typed `String`, which is what lets `it.key.toLowerCase()` type-check on the entry.
- Reading the model value before the `try` changes nothing: only `IllegalAccessException` was caught before, so a failing `getProperty` propagated then and propagates now.

## Verification

- `./gradlew :grails-gsp-core:test --tests org.grails.gsp.GspCompileStaticSpec`: 30 tests, 0 failures on the PR head.
- `./gradlew :grails-gsp-core:checkstyleMain`: 0 violations.
- `./gradlew :grails-test-examples-gsp-compile-static:compileGroovyPages` with the PR's welcome page and an expanded scaffold `index.gsp` (`List<gspstatic.Book> bookList; Number bookCount`) copied into the app: success, both classes extend `CompileStaticGroovyPage`.
- Same task with the merge-base welcome page: fails, `Cannot find matching method java.lang.Object#toLowerCase()` at generated lines 42, 413 and 1340 plus `No such property: name for class: java.lang.Object` at 1340.
- Same task with the PR's page and only the listener sort hunk reverted: success.
- Coercion prototype in `GroovyPage.java` plus 7 temporary spec cases: 37 tests, the single failure being the PR's `Integer`/`Long` mismatch case, as expected. Both files restored afterwards; `git status` shows no tracked changes.
- The copied pages were removed from `grails-test-examples/gsp-compile-static` afterwards.

codeconsole:
@matrei 

Thanks. All addressed in `0a93f8d`, `07afd38` and `0542112`.

**Model binding: converts now, but refuses a conversion that changes the value.** I took the coercion, with one guard. `castToType` on its own reproduces the truncation #16323 exists to remove. On the Groovy 5.1 jar, `Integer ← 3_000_000_000L` gives `-1294967296`, `Integer ← Long.MAX_VALUE` gives `-1`, and `Integer ← 42.9G` gives `42`, all silently. So `GroovyPage` converts with `castToType` and throws, naming the field and both types, when a numeric value doesn't survive the conversion (checked as `BigInteger` for whole numbers, `BigDecimal` otherwise). `Integer ← 42L`, `Integer ← 42.0G` and GString → String convert. A `Date` still fails, with the `GroovyCastException` as cause. The mismatch spec now uses `Integer`/`Date`, plus data-driven conversion and lossy rows. Without the guard, exactly the four lossy rows fail.

**Docs:** one paragraph under *Declaring the Model* in `gspStaticCompilation.adoc` states the conversion and the lossy failure.

**Tests:**
- `gsp-compile-static` stages the profile welcome page beside its own pages for `compileGroovyPages`. It compiles, and the page class extends `CompileStaticGroovyPage`.
- `ScaffoldedIndexViewModelSpec` expands `scaffolding/index.gsp` with `ModelBuilder`, the way dynamic scaffolding does, and renders its model declaration with `3`, `3L` and `3_000_000_000L`. The last row is a count the old `Integer` declaration can no longer hold.

**Nits:** the description now says three closures fail, with the listener sort kept for consistency. The `scaffolding.adoc` model names are fixed too, since it's the page a reader would check for the count field.

Verified: `GspCompileStaticSpec` (40 tests), `ScaffoldedIndexViewModelSpec`, `gsp-compile-static:compileGroovyPages` and `grails-gsp-core:checkstyleMain`.

matrei:
## Review Findings, round 2

Head `22163f7f11`, base `8.0.x` at `3067d0a855`. The branch now contains the base head, so the merge-base is the base itself and there is nothing to merge. All three P2s from the first round are addressed: model values are converted with `DefaultTypeTransformation.castToType`, the static compilation guide says what a mismatched value does, the welcome page is compiled statically by `gsp-compile-static`, and the scaffolded `index.gsp` count declaration has a spec. The scaffolding guide nit is fixed too. The listener sort hunk is unchanged, which is fine.

What I ran on the head, all green:

- `:grails-gsp-core:test --tests GspCompileStaticSpec`: 40 tests, 0 failures.
- `:grails-scaffolding:test --tests ScaffoldedIndexViewModelSpec`: 3 tests, 0 failures.
- `:grails-test-examples-gsp-compile-static:integrationTest`: 7 tests, 0 failures. The staged directory holds the app's five pages plus the welcome page, `views.properties` lists `/WEB-INF/grails-app/views/index.gsp`, and the compiled class extends `CompileStaticGroovyPage`.
- `:grails-gsp-core:codeStyle` and `:grails-scaffolding:codeStyle`: no violations. The modules have no PMD or SpotBugs tasks.

One new finding, in the value-change guard the PR adds on top of the coercion, and one small one next to it.

### [P2] The value-change guard rejects a `Float` widened to a `Double` field

References:

- `grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java:198` (the guard)
- `grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java:211-219` (`sameNumericValue`)
- `grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc:56`

`sameNumericValue` compares the two numbers through `new BigDecimal(n.toString())`. `Float.toString` prints the shortest decimal that rounds back to the float, `Double.toString` prints the shortest decimal that rounds back to the double, and those differ for the same binary value: `0.1f` prints `0.1`, and the same value as a `double` prints `0.10000000149011612`. The widening itself is exact, every `float` is representable as a `double`, but the decimal forms are not equal, so the guard throws.

I rendered `@{ model="Double v"}${v}` with `[v: 0.1f]` through `GroovyPagesTemplateEngine` on the head:

```
GroovyPagesException: Model field 'v' is declared as java.lang.Double, which cannot hold the java.lang.Float 0.1 the model supplied without changing it.
```

Same for `double v` with `1.1f`. Plain Groovy assigns `double d = 0.1f` without complaint, so the guide's sentence at line 56, "converted to the declared type the way a Groovy assignment converts it", is not true for the one conversion that loses nothing. The same run showed the reverse direction, `Float v` with `0.1d`, rendering `0.1`: the narrowing that does drop bits passes because both sides print `0.1`. The guard is backwards for the float family.

A `Double` model field fed a `Float` is not exotic: a domain property declared `Float` rendered through a page whose author wrote `Double`, or `BigDecimal` for that matter, which works, so the `Double` failure will read as arbitrary.

The rule the doc states, "a conversion that would change the value fails", holds up if the comparison is done at the precision of the value that came in rather than through decimal strings. One shape that gives that: convert `converted` back to the original's class and compare it to the original, using `compareTo` when the original is a `BigDecimal` so that scale does not matter.

```java
Number back = (Number) DefaultTypeTransformation.castToType(converted, original.getClass());
return original instanceof BigDecimal
        ? ((BigDecimal) original).compareTo((BigDecimal) back) == 0
        : original.equals(back);
```

Against the cases in the spec plus the ones above: `0.1f` to `Double` comes back as `0.1f` and passes; `0.1d` to `Float` comes back as `0.10000000149011612d` and fails, which is the lossy one; `3_000_000_000L` to `Integer` comes back as `-1294967296L` and fails; `42.0G` to `Integer` comes back as `42` and compares equal; `42.9G` fails; `NaN` to `Integer` comes back as `0.0` and fails; `Long.MAX_VALUE` to `Double` comes back as `Long.MAX_VALUE` because `Double.longValue` saturates, so it passes the way Java's own `long` to `double` widening does. The one thing to guard is a `Number` subclass Groovy cannot cast back to, `AtomicLong` say, where `castToType` throws; falling back to the current comparison for that is enough. Whichever rule is chosen, the spec's "converted the way a Groovy assignment converts it" table at `GspCompileStaticSpec.groovy:264` should carry a `'Double' | 0.1f` row, since that is the row that fails today.

For the record, two other rejections from the same probe are consistent with the doc sentence and I do not count them as defects: `Long.MAX_VALUE` into a `Double` field and `16_777_217` into a `Float` field both throw, and both really do lose precision.

### [P3] `NaN` into a `BigDecimal` or `BigInteger` field escapes as a raw `NumberFormatException`

References:

- `grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java:192-196`

`castToType(Double.NaN, BigDecimal.class)` throws `NumberFormatException`, not `GroovyCastException`, so it passes the `catch` at line 193 and surfaces without the field name:

```
NumberFormatException: Character N is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
```

Every other failure in this method is a `GroovyPagesException` naming the field. Catching `GroovyCastException | NumberFormatException` (or `RuntimeException`, since `castToType` has no checked failure mode) at line 193 keeps that promise for the infinities and `NaN` too. The `Double.compare` fallback at line 217 already anticipates these values for the guard; this is the same case one step earlier.

### Nit: mirror the plugin's duplicate handling in the staging task

References:

- `grails-test-examples/gsp-compile-static/build.gradle:56-60`

The Gradle plugin's own view staging (`GroovyPagePlugin.stageGroovyPages`) sets `duplicatesStrategy = DuplicatesStrategy.EXCLUDE` with the application's views listed first, so the app's page wins. The PR's `Sync` lists the app's views first too but leaves the strategy at the default, which in Gradle 9 fails the task on a duplicate path. Nothing collides today, because the app has no root `index.gsp`; the day someone adds one, the message will be about `Sync` rather than about the page. One line keeps it consistent with the plugin, and the comment above the task already explains why the welcome page is there.

### Confirmed

- Setting `source` on `compileGroovyPages` to a `Directory` also sets `srcDir` (`GroovyPageForkCompileTask.setSource`), so the staged pages keep their relative names: the welcome page is registered as `/WEB-INF/grails-app/views/index.gsp`, not under `build/generated`. The plugin sets `source` inside the task's registration action, so the build script's `tasks.named` block runs after it and is not overwritten.
- The example app maps `/` to `demo/index`, so the precompiled welcome page sits in `views.properties` unused at runtime. Compiling it is the whole point, and it changes nothing the integration tests request.
- `ScaffoldedIndexViewModelSpec` expands the template with `GStringTemplateEngine` the way `ScaffoldingViewResolver` does, and a model directive sets `compileStaticModeSetting` in `GroovyPageParser`, so the declaration really becomes a field. The `3_000_000_000L` row is the one that would fail if the template went back to `Integer`; the `3` and `3L` rows pass under either declaration now that values are coerced, which the `where:` label says.
- `GroovyPagesTemplateEngine.createTemplate(String, String)` caches by name, and the spec names the `3` and `3L` iterations both `scaffoldedIndex3`, but each iteration builds a fresh engine and the source is identical, so nothing is reused across rows.
- `Boolean showAll` fed the `String` `"false"` now renders `true`. That is Groovy truth, `Boolean b = "false"` gives the same, and it matches the doc sentence. Before the PR it threw an `IllegalArgumentException` from `Field.set`. Noting it so nobody reads it as a regression later.
- The `42.0G` to `Integer` row passes through the guard because `sameNumericValue` compares `BigDecimal` values with `compareTo`, so scale is ignored. Good.
- `List` declared, `Set` supplied throws a `GroovyCastException`, same as a plain Groovy assignment, and is reported through the new message with both types.
- Import order in the spec (`org.codehaus.groovy` between `org.grails.core.gsp` and `org.grails.taglib`) passes CodeNarc.
- The stale `book/index.gsp` classes from my round-one hand test were still in the example app's `build/gsp-classes` and were listed in the regenerated `views.properties`; I removed them. The compile task does not clear stale output, but that is pre-existing and not touched here.

## Verification

- `./gradlew :grails-gsp-core:test --tests org.grails.gsp.GspCompileStaticSpec :grails-scaffolding:test --tests grails.plugin.scaffolding.ScaffoldedIndexViewModelSpec :grails-test-examples-gsp-compile-static:integrationTest :grails-gsp-core:checkstyleMain --continue`: BUILD SUCCESSFUL, 40 + 3 + 7 tests, 0 failures.
- `./gradlew :grails-gsp-core:codeStyle :grails-scaffolding:codeStyle`: BUILD SUCCESSFUL.
- `DefaultTypeTransformation.castToType` probed directly with Groovy 5.1.2 for 23 declared/supplied pairs, and 10 of them rendered through `GroovyPagesTemplateEngine` on the head via a throwaway spec (deleted afterwards). The `Double`/`Float`, `NaN`/`BigDecimal` and `Boolean`/`"false"` results above come from that run.
- `git status` shows no tracked changes.

codeconsole:
Addressed in fc1941a96e.

- Exact numeric comparisons now accept finite `Float` → `Double` and reject lossy narrowing. I avoided round-tripping so `Long.MAX_VALUE` → `Double` stays rejected.
- Conversion exceptions, including `NaN`/infinity → `BigDecimal`/`BigInteger`, now name the model field and retain the original cause.
- Staging uses `DuplicatesStrategy.EXCLUDE` with the application views first.

Added regression cases and clarified the guide. All 227 GSP core tests and the module's style checks passed; a staging check with a duplicate `index.gsp` also passed.

matrei:
## Review Findings, round 3

Head `7a31802892`, base `8.0.x` at `37f8ca6377`. The branch contains the base head, so there is nothing to merge. The round-2 findings are all addressed in `fc1941a96e`: `Float` widened to `Double` now passes, every `castToType` failure is wrapped with the field name (the `NaN`/`BigDecimal` case has its own spec), and the staging `Sync` mirrors the plugin's `EXCLUDE` strategy.

What I ran on the head, all green, with `cleanTest` and `--no-build-cache` so the result XML is from this run:

- `:grails-gsp-core:test`, the whole module: 227 tests, 0 failures. `GspCompileStaticSpec` is 66 of them.
- `:grails-scaffolding:test`, the whole module: 31 tests, 0 failures.
- `:grails-test-examples-gsp-compile-static:integrationTest`: 7 tests, 0 failures. The welcome page is compiled and listed in `views.properties` as before.
- `:grails-gsp-core:codeStyle` and `:grails-scaffolding:codeStyle`: no violations.
- Duplicate check: with a throwaway root `index.gsp` in the example app, `stageViewsWithWelcomePage` succeeds and the staged file is the app's, not the profile's. Removed afterwards.

One finding, in the rule the new commit chose for the numeric guard.

### [P2] The exact-binary comparison rejects almost every real value crossing between `BigDecimal` and `Double`/`Float`

References:

- `grails-gsp/core/src/main/groovy/org/grails/gsp/GroovyPage.java:209-228` (`sameNumericValue`, `exactDecimalValue`)
- `grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy:313-314` (the `'Double' | 0.1G` and `'BigDecimal' | 0.1d` rows)
- `grails-doc/src/en/guide/theWebLayer/gsp/gspStaticCompilation.adoc:58`

Round 2 asked for the `Float` to `Double` widening to be compared at binary precision, which the new floating-point branch does. The commit also switched the cross-family comparison, `BigDecimal` against `Double` or `Float`, from decimal strings to the exact binary expansion, and that goes the other way: a `double` is only ever equal to a decimal fraction when the fraction is dyadic, so `0.5` and `100.00` pass and everything else fails. I rendered `@{ model="<declared> v"}${v}` on the head:

| declared | supplied | head |
|---|---|---|
| `Double` | `19.99G` | fails, "cannot hold the java.math.BigDecimal 19.99 ... without changing it" |
| `double` | `19.99G` | fails |
| `Double` | `1234.56G` | fails |
| `Float` | `19.99G` | fails |
| `Double` | `0.1G` | fails, as the doc now says |
| `BigDecimal` | `19.99d` | fails |
| `BigDecimal` | `0.1f` | fails |
| `BigDecimal` | `19.99f` | fails |
| `Double` | `1.5G` | renders `1.5` |
| `Double` | `100.00G` | renders `100.0` |

These are the two families a Grails application actually mixes: GORM money properties are `BigDecimal`, measurements are `Double` or `Float`, and a `Double price` page fed a `BigDecimal` domain property, or the reverse, now fails at render on almost every row of real data. Neither direction loses anything a user can observe. Groovy's own conversion agrees: `castToType(0.1d, BigDecimal)` is `0.1`, `castToType(0.1G, Double)` is `0.1`, `castToType(19.99f, BigDecimal)` is `19.99`, and plain `BigDecimal b = 0.1d` and `double d = 19.99G` assign without complaint. So the guide's first sentence, "converted the way a Groovy assignment converts it", is contradicted by the paragraph two lines below it.

The rule that gives the guide's promise is: a conversion is accepted when the converted value converts back to the original. For the floating-point family that is the `Double.compare` branch the commit added. For everything else it is what round 2 already had, the decimal strings, because a `double` prints as the shortest decimal that reads back to it and that is also the decimal Groovy converts it to. Concretely, `sameNumericValue` keeps the new first branch and goes back to `new BigDecimal(original.toString()).compareTo(new BigDecimal(converted.toString())) == 0` for the rest, with the existing `NumberFormatException` catch returning `false`; `exactDecimalValue` goes away. I ran the PR's spec and the table above with that in place:

- `GspCompileStaticSpec`: 66 tests, 2 failures, and they are exactly the `'Double' | 0.1G` and `'BigDecimal' | 0.1d` rows in "a conversion that would change the value", which encode the exact-binary rule. Moving them to the "converted the way a Groovy assignment converts it" table with expected `0.1` is the whole spec change.
- Every other rejection stays rejected: `3_000_000_000L` for `Integer`, `42.9G` for `Integer`, `70_000` for `Short`, `0.1d` for `Float`, `Long.MAX_VALUE` for `Double`, `16_777_217` for `Float`, `9_007_199_254_740_993G` for `Double`, `123456789012345678L` for `Double`, `NaN` for `Integer`, `1E400` for `Double`, `42.5d` for `Integer`, `1.0E+23G` for `Long`.
- Every row in the table above renders, `0.1f` for `Double` still renders `0.10000000149011612`, and `0.30000000000000004d` for `BigDecimal` renders `0.30000000000000004`.
- The one visible consequence of the rule: the `BigInteger` `10^23` becomes a `Double` and renders `1.0E23`, where the head rejects it because the nearest `double` is `99999999999999991611392`. That is the round-trip semantics, the same one that lets `19.99G` through, and it prints what the user supplied.

The doc paragraph at line 58 then reads along the lines of: "A `BigDecimal` and a `Double` or `Float` are compared through the decimal the floating-point value prints as, so `19.99G` satisfies a `Double` declaration and `19.99d` a `BigDecimal` one, while a value with more digits than the type can carry, `9007199254740993G` for a `Double`, is rejected. Widening a finite `Float` to `Double` is accepted; narrowing a `Double` to `Float` is rejected if it loses precision." The `NaN`/infinity sentence stays. The prototype is not in the working tree.

### Nit: two rows in the "would change the value" table are Groovy refusals, not guard rejections

References:

- `grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy:317-318`

`'Double' | new BigDecimal('1E400')` and `'Double' | Float.NEGATIVE_INFINITY` never reach `sameNumericValue`: `castToType` throws `GroovyRuntimeException` ("Automatic coercion of java.lang.Float value -Infinity to double failed. Value is out of range.") and the page reports the "cannot be converted" message with that cause. The feature's assertions check only the field name and the two type names, which both messages contain, so the rows pass under the wrong heading. They belong next to the `Date` case in "a model value that cannot be converted", or that feature's `e.cause` assertion should become the discriminator. Cosmetic.

### Confirmed

- The guide's "widening a finite `Float` to `Double`" is precise: Groovy itself refuses an infinite `Float` for a `Double` (plain `Double d = Float.NEGATIVE_INFINITY` throws the same `GroovyRuntimeException`), while `Float.NaN` for `Double` and either infinity for `Float` or `float` pass. The page inherits Groovy's asymmetry rather than adding one.
- `catch (RuntimeException e)` around `castToType` is right: `GroovyCastException`, `GroovyRuntimeException` and `NumberFormatException` are all what it throws, none checked, and the new "non-finite value" feature pins the `NumberFormatException` cause with the field name in the message.
- `Double.compare` in the floating-point branch handles `-0.0d` for `float` (spec row 9) and `NaN` for `Double` (row 10) as intended, and the `AtomicLong` row shows a `Number` subclass with no Groovy cast of its own still goes through the decimal comparison.
- `duplicatesStrategy = DuplicatesStrategy.EXCLUDE` resolves without an import in a Groovy build script (Gradle's default imports cover `org.gradle.api.file`), and the app's views being listed first is what makes the app's page win, verified above.
- The scaffold template, welcome pages, `ScaffoldedIndexViewModelSpec` and the staging task are unchanged since round 2, and the full `grails-scaffolding` and `grails-gsp-core` test tasks pass on the head.

## Verification

- `./gradlew :grails-gsp-core:cleanTest :grails-gsp-core:test :grails-scaffolding:cleanTest :grails-scaffolding:test :grails-test-examples-gsp-compile-static:cleanIntegrationTest :grails-test-examples-gsp-compile-static:integrationTest :grails-gsp-core:codeStyle :grails-scaffolding:codeStyle --no-build-cache --continue`: BUILD SUCCESSFUL; 227 + 31 + 7 tests, 0 failures; result XML timestamps from this run.
- `stageViewsWithWelcomePage --rerun-tasks` twice, once with a throwaway app-level `index.gsp` (staged copy is the app's) and once without (staged copy is the profile's). The throwaway page was removed.
- A throwaway `ZzProbeSpec` in `grails-gsp-core` rendered the 33 declared/supplied pairs above through `GroovyPagesTemplateEngine`, once on the head and once with the round-trip prototype in `GroovyPage.java`; the prototype run also executed `GspCompileStaticSpec` (66 tests, the 2 expected failures). Both files restored, `git status` shows no tracked changes.
- Stale `book/index.gsp` classes from the round-one hand test were still in the example app's `build/gsp-classes` and `views.properties`; removed. Build output only.

codeconsole:
Agreed, and fixed in `762cffe848`. I reproduced your table on the Groovy 5.1.2 jar first: under the exact-binary rule every `BigDecimal`↔`Double`/`Float` row except the dyadic ones is rejected, and under the round-trip rule all of them convert while every rejection you listed stays rejected.

**Guard:** `sameNumericValue` keeps the `Double.compare` branch for the floating-point family and compares everything else through the decimal strings again. `exactDecimalValue` is gone. The one visible change is the one you named: `10^23` as a `BigInteger` becomes a `Double` and renders `1.0E23`.

**Spec:** the `'Double' | 0.1G` and `'BigDecimal' | 0.1d` rows moved to the conversion table with `0.1` expected, alongside `19.99G`→`double`/`Float`, `19.99d`/`0.1f`/`0.30000000000000004d`→`BigDecimal`, and the `BigInteger`→`Double` case. The rejection table gains `42.5d`→`Integer`, `123456789012345678L`→`Double` and `1.0E+23G`→`Long`.

**Nit:** the `1E400` and `-Infinity` rows sit next to the `Date` case now, and the two features discriminate on what produced the failure: the guard's message ends "without changing it" with no cause, Groovy's refusals say "cannot be converted" with the cause named per row (`GroovyCastException` for the `Date`, `GroovyRuntimeException` for the other two; on 5.1.2 the former is not a subtype of the latter, which is why the rows are separate).

**Doc:** line 58 rewritten to your wording.

Verified: `GspCompileStaticSpec` 75 tests, 0 failures with `--no-build-cache`; with the exact-binary rule put back, exactly the eight cross-family conversion rows fail and nothing else; `:grails-gsp-core:codeStyle` clean.

matrei (approved):
## Review Findings, round 4

Head `762cffe848`, base `8.0.x` at `bb0ee3c611`. The merge-base is `37f8ca6377`; the branch is 72 commits behind the base, `git merge-tree` reports no conflicts, and none of those 72 commits touch `grails-gsp/core`, `grails-scaffolding`, `gsp-compile-static`, the welcome pages or the two guide chapters, so the merged result is the head as tested. The round-3 P2 is addressed in `762cffe848` exactly as proposed: `sameNumericValue` keeps the binary comparison for the `Float`/`Double` family and compares everything else through the decimal each value prints as, `exactDecimalValue` is gone, the `'Double' | 0.1G` and `'BigDecimal' | 0.1d` rows moved to the "converted the way a Groovy assignment converts it" table with six more cross-family rows, and the round-3 nit is fixed too: the `1E400` and `-Infinity` rows now sit in "cannot be converted" with the `GroovyRuntimeException` cause pinned, and the guard rows assert `without changing it` with a `null` cause, so the two features now discriminate.

What I ran on the head, all green, with `cleanTest` and `--no-build-cache`, result XML timestamps from this run:

- `:grails-gsp-core:test`, the whole module: 236 tests, 0 failures. `GspCompileStaticSpec` is 75 of them.
- `:grails-scaffolding:test`, the whole module: 31 tests, 0 failures.
- `:grails-test-examples-gsp-compile-static:integrationTest`: 7 tests, 0 failures.
- `:grails-gsp-core:codeStyle` and `:grails-scaffolding:codeStyle`: no violations.

Approve. One nit.

### Nit: `groovy.lang.GroovyRuntimeException` does not need importing

References:

- `grails-gsp/core/src/test/groovy/org/grails/gsp/GspCompileStaticSpec.groovy:26`

`groovy.lang.*` is a default import in every Groovy compilation unit, and CodeNarc's `UnnecessaryGroovyImport` rule, which is in the project ruleset, flags this line:

```
Violation: Rule=UnnecessaryGroovyImport P=3 Line=26 Src=[import groovy.lang.GroovyRuntimeException]
```

`codeStyle` does not report it because the code-style plugin disables every CodeNarc task whose name contains `test`, so test sources are never scanned. Drop the line; the class resolves without it. Cosmetic, and I would not hold the PR for it.

### Confirmed

- The new rule is the round-trip rule the guide describes. I ran `DefaultTypeTransformation.castToType` plus a copy of `sameNumericValue` over 44 declared/supplied pairs on Groovy 5.1.2, beyond the 24 accept and 13 reject rows in the spec. Everything the guide says is accepted is accepted (`19.99G` for `Double`, `Float` and `double`; `19.99d`, `19.99f`, `0.1d`, `0.1f` and `0.30000000000000004d` for `BigDecimal`; `1E+2G` and `100.00G` for `Double`; `42.0d` for `Long` and `BigInteger`; `1.0E23d` for `BigInteger`, which converts to `10^23` and back), and every conversion that loses something is rejected: `42.5f` and `42.5d` for `Long` and `Integer`, `42.5G` for `BigInteger`, `1E+20G` for `Long` (Groovy wraps to `7766279631452241920`), `1.0E23d` and `Infinity` for `Long` (saturate to `Long.MAX_VALUE`), `200` for `Byte`, `9007199254740993G` for `Double`, `1.0E-400G` for `Double` and `1.0E-50G` for `Float` (both underflow to `0.0`), and `1.0E-45G` for `Float` (rounds to the subnormal `1.4E-45`).
- The one value the decimal comparison cannot distinguish is the sign of zero: `-0.0d` for a `BigDecimal` field becomes `0.0` and passes, because `BigDecimal` has no negative zero and `compareTo` says they are equal. Nothing observable is lost, and the `Float`/`Double` branch still keeps `-0.0` (spec row `'float' | -0.0d`).
- `Double.toString` and `Float.toString` always produce a form `new BigDecimal(String)` parses (`1.0E-5`, `1.0E23`, `4.9E-324`), so the only `NumberFormatException` the `catch` sees is `NaN` or an infinity, which is what the comment says.
- The doc paragraph at `gspStaticCompilation.adoc:58` now matches the implementation and the round-3 table: `19.99G`/`19.99d` accepted, `9007199254740993G` rejected, `Float` widening accepted, lossy `Double` to `Float` rejected, non-finite values named. The sentence "which is the decimal Groovy converts it to" is true: `castToType(19.99d, BigDecimal)` is `19.99`, not the binary expansion.
- `'Double' | new BigInteger('1' + '0' * 23) | '1.0E23'` documents the one consequence round 3 called out, and the `'Long' | 1.0E+23G` row next to it shows the same magnitude is still refused where it genuinely does not fit.
- `GroovyCastException` extends `GroovyRuntimeException`, so `cause.isInstance(e.cause)` in "cannot be converted" would also pass with the broader class on the `Date` row; using the more specific class there is the better pin.
- `catch (RuntimeException e)` around `castToType`, the `EXCLUDE` staging strategy, the scaffold template, the welcome pages and `ScaffoldedIndexViewModelSpec` are unchanged since round 3.

## Verification

- `./gradlew :grails-gsp-core:cleanTest :grails-gsp-core:test :grails-scaffolding:cleanTest :grails-scaffolding:test :grails-test-examples-gsp-compile-static:cleanIntegrationTest :grails-test-examples-gsp-compile-static:integrationTest :grails-gsp-core:codeStyle :grails-scaffolding:codeStyle --no-build-cache --continue`: BUILD SUCCESSFUL in 3m 17s; 236 + 31 + 7 tests, 0 failures, 0 skipped; result XML written 21:06 to 21:08 on 2026-09-16.
- A throwaway Groovy script (Groovy 5.1.2 from the Gradle cache, outside the build) ran `castToType` and a copy of the head's `sameNumericValue` over the 44 pairs above.
- CodeNarc 4.0.0 run directly on a copy of `GspCompileStaticSpec.groovy` with `UnnecessaryGroovyImport` and `UnusedImport`: the one P3 above, nothing else.
- `git status` shows no tracked changes.

codeconsole:
Thanks for the approval. Import dropped in `ee8c4c2db8`; `GspCompileStaticSpec` still 75/0.

One correction to the *Confirmed* list, for the record: on 5.1.2 `GroovyCastException` extends `java.lang.ClassCastException`, not `GroovyRuntimeException` (chain: `GroovyCastException → ClassCastException → RuntimeException`). So the broader class would not pass on the `Date` row, which is why the cause is pinned per row.

PR #16323, opened 2026-09-08
Return Long from count() instead of silently narrowing to Integer

codeconsole:
## Description

`count()` returns `Long` instead of `Integer`, matching what the datastore actually produces.

### The problem

`GormStaticOperations` declares:

```groovy
/**
 * Counts the number of persisted entities
 * @return The number of persisted entities
 */
Integer count()
```

How wide the value underneath actually is depends on the datastore:

| Datastore | Underlying count value |
| --- | --- |
| Hibernate 5 / 7 | SQL `COUNT(*)` — 64-bit |
| Neo4j | Cypher `count(*)` — 64-bit |
| MongoDB | `{$sum: 1}` — Int32, promoting to Int64 once the total exceeds it |
| Simple map (in-memory) | `entityList.size()` — `int` |

`GormStaticApi` normalises all of them with `longValue()`, and the declared return type then narrowed the result again. Every site did that narrowing a different way:

| Site | Narrowing |
| --- | --- |
| `GormStaticApi` | `((Number)res).intValue()` |
| Hibernate 5 | builds a `CriteriaQuery<Long>`, then casts `(Integer)` |
| Hibernate 7 | `select count(*) ... as Integer` |
| GraphQL `CountEntityDataFetcher` | declared `Integer`, so the returned `Long` is narrowed on return |
| GraphQL schema field | typed `Int`, which is 32-bit by GraphQL specification |

MongoDB and Neo4j need no change of their own: neither overrides `count()`, so both inherit
`GormStaticApi` and are fixed by the core change. MongoDB is the sharpest case for declaring `Long`, since
the width it returns depends on how many documents are in the collection — a collection that grows past
2³¹ starts returning Int64, which is the correct answer, and `intValue()` then wrapped it.

`intValue()` truncates silently, so a table with more than `Integer.MAX_VALUE` rows reported a wrong and possibly negative count with no error at all. `Long` is the only type that holds every value a datastore can produce here, so declaring it removes the narrowing rather than moving it around.

`RestfulServiceController` gets simpler as a direct consequence — it was calling `Math.toIntExact(getService().count(params))` on a service that already returned a `Long`, narrowing purely to fit the controller signature. That call is gone.

### Scope

`GormStaticOperations.count()` / `getCount()`, the `count()` / `count` members generated on every domain class via `GormEntity`, `TenantDelegatingGormOperations`, both Hibernate backends, and `RestfulController.countResources()`.

Plus the GraphQL count fetcher and its schema field, which changes the published schema from `personCount: Int` to `personCount: Long` (`graphql-java` maps `Long` to `ExtendedScalars.GraphQLLong`). Queries selecting the field and the JSON they return are unchanged, but generated clients and schema snapshots need regenerating.

`grails-datamapping-rx` needed no change: `RxGormStaticOperations.count()` already returns `Observable<Number>` and never committed to `Integer`.

### Compatibility

Dynamic Groovy is unaffected. `Book.count() == 5`, `int n = Book.count()`, and arithmetic on the result all continue to work, because Groovy converts between numeric types on assignment and compares them by value. What breaks is statically compiled code assigning the result to an `Integer`, Java callers relying on unboxing, and subclasses overriding `countResources()`. Documented as section 54 of the 8.0 upgrade notes with before/after examples.

### Typed views

Worth calling out, because it is the same defect twice. Typed view models are bound by reflective `Field.set` with **no coercion** — in GSP (`GroovyPage.applyModelFieldsFromBinding`) and in JSON views (`WritableScriptTemplate.FieldSetter`) alike. A view that names a concrete numeric type therefore fails outright when a controller supplies the other one, and the two scaffolding paths supply different types: `generate-views` pairs with a generated service declaring `Long count()`, while `static scaffold = X` is backed by `RestfulController.countResources()`.

So the scaffolded `index.gsp` and both `product/index.gson` views now declare `Number`, which accepts whatever a controller supplies. `HalViewHelper.paginate` accordingly takes a `Number total` (converted once at the internal `getPaginationLinks` boundary) so a count can reach it without being narrowed again on the way in.

The `paginate` signature change is the one piece here that is not strictly about `count()`. It is included because without it the in-tree JSON views cannot compile against a `Long` count. Happy to split it out if reviewers prefer.

## Relationship to #16322

#16322 changes the scaffolded `index.gsp` count field to `Number` for the same underlying reason. This branch is cut from `8.0.x`, where that file still declares `Integer`, and the `grails-test-examples-gorm` scaffolding functional tests fail without it — so the identical one-line change is included here to keep this PR green on its own. The two resolve to the same line and should merge cleanly in either order.


jdaugherty:
Under compilestatic code wouldn't this be considered a breaking change since there wouldn't be type coercion?  Would it then be a better change to return a type that extends Number, ie: 

         T <T extends Number> count()

codeconsole:
You're right that it's a breaking change for statically compiled callers — that's intentional, and `8.0.x` is the branch where it's allowed. But the generic form is worth testing rather than reasoning about, so I ran all three signatures against Groovy 5.1.2, each called from `@CompileStatic` code written against the old API (`Integer n = api.count()`), with an implementation returning `3_000_000_000L`:

| Signature | Result for `Integer n = api.count()` |
| --- | --- |
| `<T extends Number> T count()` | **compiles**, returns `-1294967296` |
| `Long count()` | compile error — `loss of precision from java.lang.Long to java.lang.Integer` |
| `Number count()` | compile error — `Cannot assign value of type java.lang.Number to variable of type java.lang.Integer` |

The generic signature does remove the compile error, but it doesn't add coercion — it erases to `Object` and lets the call site insert the cast. `T` is chosen by the caller, while the implementation can only ever return the one type the datastore produced, so the truncation survives and now shows up as a negative row count at runtime instead of a build failure. That is the bug this PR exists to remove, made invisible. Erasure also means the implementation can't inspect `T` and convert to it, so there's no version of that signature that could coerce.

`Number` breaks the same callers, so it doesn't avoid the migration either — it just hands callers a type they have to unwrap before using.

So there's no non-breaking option here: any type wide enough to hold what the datastores actually return is unassignable to `Integer` under STC. Given that, the compile error seems like the feature rather than the cost — it points at precisely the call sites that were silently truncating, and the fix at each is a one-word type change. §54 of the upgrade notes covers the migration, including the Java-caller and GraphQL-client cases.

Worth noting where this actually bites: MongoDB aggregates counts with `{$sum: 1}`, which returns Int32 and promotes to Int64 once the total exceeds it. So on a large collection GORM was receiving the correct 64-bit count and `intValue()` was wrapping it — no error, just a wrong number.

jdaugherty:
We discussed this in the weekly and there was agreement the breaking change is acceptable for 8.0.  Reviews can proceed now.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -3097,3 +3097,92 @@ grails {
 imported whether or not they are present — a star import of an absent package contributes no classes and is
 not an error in Groovy, so the probe changed nothing it could observe. An application that relied on the
 import being *omitted* when the package was absent sees no difference in compiled output.
+
+==== 54. `count()` Returns `Long`
```
jdaugherty: This file conflicts with `8.0.x` on merge, and `8.0.x` now numbers sections through 57, so after a rebase this becomes §58.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
```diff
@@ -3097,3 +3097,92 @@ grails {
...
+    protected Long countResources() { ... }
+}
+----
+
+Java callers assigning the result to an `Integer` or `int` need an explicit conversion, since `Long` does
+not unbox to `int`:
+
+[source,java]
+----
+long total = BookGormEntity.count();          // preferred
```
jdaugherty: `BookGormEntity` isn't a type. The trait's static methods land on the domain class itself, so a Java caller writes `Book.count()`, the same as the line below. `long total = Book.count();` also demonstrates the auto-unboxing point directly.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/GormStaticApiSpec.groovy
```diff
@@ -105,10 +106,20 @@ class GormStaticApiSpec extends Specification {
...
         when:
-        Integer n = api.count()
+        Long n = api.count()
 
         then:
-        n == 0
+        n == 0L
+    }
+
+    void "count() returns a Long so a large table is not truncated"() {
```
jdaugherty: The name promises the truncation case, but the body only checks declared types. Nothing pushes a value past `Integer.MAX_VALUE` through the `longValue()` branch this PR changed, which is the behaviour that was broken. Stubbing the session gets at it directly, and this passes on the branch (needs `org.grails.datastore.mapping.query.Query` imported):

```groovy
void "count() preserves a datastore count above Integer.MAX_VALUE"() {
    given:
    def ds = Stub(Datastore)
    def session = Stub(Session)
    def query = Mock(Query)
    ds.getMappingContext() >> datastore.mappingContext
    ds.connect() >> session
    session.getDatastore() >> ds
    session.createQuery(GormStaticApiThing) >> query
    query.projections() >> Mock(Query.ProjectionList)
    query.singleResult() >> 3_000_000_000L
    def api = new GormStaticApi(GormStaticApiThing, ds, [])

    expect:
    api.count() == 3_000_000_000L
}
```

Worth also asserting the trait path, since that is what domain classes actually expose: `GormStaticApiThing.count() instanceof Long` and `GormStaticApiThing.count instanceof Long` both hold against the `SimpleMapDatastore` this spec already has. The two `getMethod(...).returnType == Long` lines can go once the behaviour is covered; they restate the signature.

grails-data-graphql/core/src/main/groovy/org/grails/gorm/graphql/Schema.groovy
```diff
@@ -381,7 +381,7 @@ class Schema {
                     DataFetcher countFetcher = dataFetcherManager.getReadingFetcher(entity, COUNT).orElse(new CountEntityDataFetcher(entity))
 
                     final String countFieldName = namingConvention.getCount(entity)
-                    final GraphQLOutputType countOutputType = (GraphQLOutputType) typeManager.getType(Integer)
+                    final GraphQLOutputType countOutputType = (GraphQLOutputType) typeManager.getType(Long)
```
jdaugherty, line 384: Nothing asserts the type the count field is built with; `ReadOnlyOpSpec` only checks the field exists. A `type == ExtendedScalars.GraphQLLong` assertion next to that check (or in `SchemaSpec`) would pin this.

Also worth a sentence in the upgrade note: paginated list responses already type `totalCount` as `Long` through the same type manager (`DefaultGraphQLPaginationResponseHandler`), so the count field now matches an existing scalar in the schema rather than introducing a new one.

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/HalViewHelper.groovy
```diff
@@ -113,7 +113,7 @@ interface HalViewHelper {
      * @param object The object to create links for
      * @param total The total number of objects to be paginated
      */
-    void paginate(Object object, Integer total)
+    void paginate(Object object, Number total)
```
jdaugherty: Keeping this in the PR is fine. Callers passing an `Integer` still compile, and `DefaultHalViewHelper` is the only implementer in the tree. A line in the upgrade note that `paginate` now takes a `Number` would cover anyone with their own `HalViewHelper` implementation.

jdaugherty: The paginate options seem additional to this review, Number allows BigDecimal, why aren't we making this a Long too?

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/internal/DefaultHalViewHelper.groovy
```diff
@@ -258,7 +258,7 @@ class DefaultHalViewHelper extends DefaultJsonViewHelper implements HalViewHelpe
      * @param order The order in which the results are to be sorted eg: DESC or ASC
      */
     //TODO: Once GROOVY-9662 is fixed, remove explicit delegate call and typecast to StreamingJsonDelegate
-    void paginate(Object object, Integer total, Integer offset = null, Integer max = null,  String sort = null, String order = null) {
+    void paginate(Object object, Number total, Integer offset = null, Integer max = null,  String sort = null, String order = null) {
```
jdaugherty: Why aren't we changing this to Long?

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/internal/DefaultHalViewHelper.groovy
```diff
@@ -277,7 +277,7 @@ class DefaultHalViewHelper extends DefaultJsonViewHelper implements HalViewHelpe
                 ((StreamingJsonBuilder.StreamingJsonDelegate) delegate).call(HREFLANG_ATTRIBUTE, locale.toString())
                 ((StreamingJsonBuilder.StreamingJsonDelegate) delegate).call(TYPE_ATTRIBUTE, contentTypeMimeType ?: contentType)
             }
-            List<Link> links = getPaginationLinks(object, total, max, offset, sort, order) as List<Link>
+            List<Link> links = getPaginationLinks(object, total?.intValue(), max, offset, sort, order) as List<Link>
```
jdaugherty: Why are we forcing this to intValue() and not converting it to long too?

jdaugherty (commented):
I think we should keep this review focused on count() if we're only going to partially implement the Long in the pagination code.  If we fully implement Long in the pagination, then I'm fine with updating the paging logic

codeconsole:
Thanks — rebased onto current `8.0.x` and took the full-Long option on the pagination, so the paging logic is in scope rather than half-converted.

**Pagination is now `Long` end to end.** `paginate` (all five overloads), `getPaginationLinks`, and every offset derived from a total: `getNextOffset`, `getLastOffset`, `getPrevOffset`, `paramsWithOffset`, `buildPaginateParams`. The offsets mattered as much as the total — `getLastOffset` computes `(ceil(total / max) - 1) * max`, so an offset overflows the moment a total stops fitting in an `Int`. `Parameters` already exposed `long(name, default)`, so the request-read path needed nothing new. You're right that `Number` was the wrong choice; it admits `BigDecimal` and says less than the code knows.

One thing I did not change, and want to flag rather than bury. `links(Map, Object, Number total)` on `HalViewHelper` is pre-existing and was already narrowing at `DefaultHalViewHelper:229` via `total.intValue()` — a truncation that predates this PR. Changing that signature to `Long` would reject callers currently passing an `Integer`, so I left it as `Number` and made the call widen (`total?.longValue()`) instead of truncate. That does leave one `Number` in the pagination area, which is the shape you objected to on `paginate`. Happy to take it to `Long` as well if you'd rather have consistency than source compatibility there.

**On the other comments:**

- §54 → §58 after the rebase, as you called. Separately: `8.0.x` currently has `==== 54.` twice — *Request Processing Behaviour Changes* (line 3256) and *Non-Public Bean Classes Are Marshalled, and Reported Once* (line 3454). I left both alone since neither is mine, but you may want to renumber the second.
- `BookGormEntity` fixed — it's now `long total = Book.count();`, which shows the auto-unboxing point directly as you suggested.
- The count spec is replaced with your stubbed version pushing `3_000_000_000L`, plus `GormStaticApiThing.count()` / `.count` for the trait path, and the two `getMethod(...).returnType` lines are gone. I checked it isn't vacuous: reverting `longValue()` to `intValue()` makes it fail, restoring it makes it pass.
- GraphQL scalar is asserted, but in `SchemaSpec` rather than `ReadOnlyOpSpec` — the only test in the latter is `@Ignore`d, so an assertion there never executes. It checks every `*Count` query field is `ExtendedScalars.GraphQLLong`, with a non-empty guard so it can't pass by matching nothing.
- Added the `totalCount` sentence to the upgrade note; confirmed at `DefaultGraphQLPaginationResponseHandler:52`, which already builds that field with `typeManager.getType(Long)`.

**Verification after the rebase:** `compileGroovy compileTestGroovy` across every module, then `grails-datamapping-core`, `grails-data-graphql-core`, `grails-test-examples-gorm:integrationTest` and `grails-test-examples-views-functional-tests:integrationTest` with `--rerun-tasks` — including `ProductSpec`, which exercises the HAL pagination links over the wire. All green.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Rebased onto current `8.0.x` — it's §58 now.

Separately: `8.0.x` already has `==== 54.` twice, *Request Processing Behaviour Changes* (3256) and *Non-Public Bean Classes Are Marshalled, and Reported Once* (3454). Left both alone since neither is mine, but you may want to renumber the second.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
codeconsole: Fixed — it's `long total = Book.count();` now, which shows the auto-unboxing point directly as you suggested.

grails-datamapping-core/src/test/groovy/org/grails/datastore/gorm/GormStaticApiSpec.groovy
codeconsole: Taken as written, plus `GormStaticApiThing.count()` / `.count` for the trait path, and the two `getMethod(...).returnType` lines are gone.

Checked it isn't vacuous the way the old one was: reverting `longValue()` to `intValue()` makes it fail, restoring it makes it pass.

grails-data-graphql/core/src/main/groovy/org/grails/gorm/graphql/Schema.groovy
codeconsole: Added — in `SchemaSpec` rather than `ReadOnlyOpSpec`, because the only test in the latter is `@Ignore`d, so an assertion there would never execute. It asserts every `*Count` query field is `ExtendedScalars.GraphQLLong`, with a non-empty guard so it can't pass by matching nothing.

Added the `totalCount` sentence to the upgrade note too — confirmed at `DefaultGraphQLPaginationResponseHandler:52`, which already builds that field with `typeManager.getType(Long)`.

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/HalViewHelper.groovy
codeconsole: Agreed — `Number` admits `BigDecimal` and says less than the code knows. All five `paginate` overloads take `Long` now, and so does `links(Map, Object, Long total)`, which was pre-existing but truncating for the same reason.

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/internal/DefaultHalViewHelper.groovy
codeconsole: Done — `Long total`.

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/internal/DefaultHalViewHelper.groovy
codeconsole: Gone. `getPaginationLinks` takes `Long`, and so do the offsets derived from a total — `getNextOffset`, `getLastOffset`, `getPrevOffset`, `paramsWithOffset`, `buildPaginateParams`.

The offsets mattered as much as the total: `getLastOffset` computes `(ceil(total / max) - 1) * max`, so the offset overflows the moment the total stops fitting in an `Int`. `Parameters` already exposed `long(name, default)` for the request path.

grails-doc/src/en/guide/upgrading/upgrading80x.adoc
jdaugherty: Can you go ahead and fix the other numbering problem please? @codeconsole

codeconsole: Done in `af45fc2` — should have just fixed it rather than pointing at it, sorry.

"Non-Public Bean Classes Are Marshalled, and Reported Once" is §58 and `count()` moves to §59. Top-level sections now run 1–59 with no duplicates and no gaps.

grails-views-gson/src/main/groovy/grails/plugin/json/view/api/internal/DefaultJsonViewHelper.groovy
```diff
@@ -160,33 +160,33 @@ class DefaultJsonViewHelper extends DefaultGrailsViewHelper {
         ViewUtils.getBooleanFromMap(ASSOCIATIONS, arguments, true)
     }
 
-    protected List<Link> getPaginationLinks(Object object, Integer total, Parameters params) {
-        int offset = params.int(PAGINATION_OFFSET, 0)
+    protected List<Link> getPaginationLinks(Object object, Long total, Parameters params) {
+        long offset = params.long(PAGINATION_OFFSET, 0L)
         int max = params.int(PAGINATION_MAX, 10)
```
jdaugherty, line 165: Shouldn't max be a long too?

codeconsole: No — `max` is a page size, and GORM takes it as an `int` at the source (`Query.max(int)`, `Query.maxResults(int)`). `total` and `offset` are positions in the result set, which is what can exceed 2^31.

Every expression mixing them already widens through the `Long` operand — `offset + max`, `Math.max(offset - max, 0L)`, `laststep * max` — so nothing truncates.

sbglasius (approved):

matrei:
# AI Review

Thanks, the direction is right and the core change is complete: every `GormStaticOperations.count()` implementation in the repo is updated (Mongo and Neo4j go through `GormStaticApi.count()` and need nothing), and the tests for `grails-datamapping-core`, `grails-views-gson` and `grails-data-graphql-core` pass locally. The GSP template, the `RestfulServiceController` and the section-58 renumbering all look correct.

A few things to address before merge.

### 1. The JSON API pagination path still truncates

`DefaultJsonApiViewHelper` (around line 478) was not touched:

```groovy
Integer total = (Integer) paginationArgs.get(PAGINATION_TOTAL)
...
List<Link> links = getPaginationLinks(resource, total, params)
```

Under `@CompileStatic` that cast is Groovy's numeric cast, not a checkcast, so a `Long` coming from `Book.count()` is silently wrapped (I verified `(Integer) 3_000_000_000L` gives `-1294967296`). That is the exact bug this PR is fixing, still present in `jsonapi.render(list, [pagination: [total: Book.count(), resource: Book]])`. Suggest:

```groovy
Long total = ((Number) paginationArgs.get(PAGINATION_TOTAL)).longValue()
```

and a row in `IterableRenderSpec` with a total above `Integer.MAX_VALUE` that asserts the `last` link offset.

### 2. `PaginationSpec` was not updated for the changed signatures

`getPaginationLinks`, `getPrevOffset`, `getNextOffset` and `getLastOffset` all changed from `Integer` to `Long`, but `PaginationSpec` still declares the results as `Integer` and only exercises int-range values. It passes because dynamic Groovy converts on assignment, so it no longer pins the type it is testing. Please switch the declarations to `Long` and add at least one row above `Integer.MAX_VALUE`, for example `getLastOffset(3_000_000_000L, 10) == 2_999_999_990L`. That row is the behaviour the change exists to enable, and it also covers the `Math.round(Math.ceil(...))` to `(long) Math.ceil(...)` rewrite.

### 3. `offset` in `HalViewHelper.paginate` stayed `Integer`

`total` is now `Long`, and the protected helpers all take `Long offset`, but the public `paginate(Object, Long total, Integer offset, ...)` overloads still take `Integer offset`. A page past row 2^31 can be described by `total` but not requested through `paginate`. Low priority, but widening `offset` in the same PR keeps the surface consistent instead of doing it in a second breaking change later.

### 4. Upgrade note

- "GORM normalises whatever it gets with `longValue()`, then the declared return type narrowed it again with `intValue()`" describes the new code as if it were the old pipeline. The pre-change `GormStaticApi.count()` called `intValue()` directly. Suggest: "`GormStaticApi.count()` narrowed the datastore result with `intValue()`, which truncates silently, so ...".
- The PR had to change `Integer productCount` to `Long productCount` in two `.gson` `model {}` blocks, but the note only mentions the GSP `@{ model=... }` directive. Please add a sentence for JSON views: a `model { Integer fooCount }` declaration should become `Long` (or `Number`).
- Worth a line that `HalViewHelper.paginate()` and `links()` now declare `total` as `Long`. Callers are unaffected because Groovy widens `Integer` to `Long` in both dynamic and statically compiled code (verified), but anyone implementing `HalViewHelper` has to update.

### Minor / no action needed

- `GormStaticApiSpec`'s new large-count test drives everything through `Datastore`, `Session` and `Query`, so it stays on the public surface. Good.
- `SchemaSpec` pinning the `GraphQLLong` scalar is a good regression guard for the schema-level change.
- Custom GraphQL count fetchers registered through the data fetcher manager that still return `Integer` keep working, since the `Long` scalar coerces integers, so the type change is only breaking for generated clients, which the note already says.

jdaugherty:
@codeconsole looks like Mattias found some more ints for you to change =)  I'm good to merge once these are changed.

codeconsole:
All four addressed in `8f497f5`.

**1. JSON API truncation** — real miss, thanks. Fixed as you suggested, plus an `IterableRenderSpec` case rendering with `total: 3_000_000_000L` that asserts the `last` offset. Confirmed it fails with the old `(Integer)` cast and passes with `longValue()`.

**2. `PaginationSpec`** — declarations now `Long`, with rows past `Integer.MAX_VALUE` including your `getLastOffset(3_000_000_000L, 10) == 2_999_999_990L`. One row I added was wrong and the suite caught it: `getNextOffset` returns `null` on the last page, so it's now a penultimate-page row that advances past 2³¹ plus a `null` row.

**3. `offset`** — widened to `Long`. That needed `Parameters.long()` instead of `int()` in `DefaultHalViewHelper`, or the request path would have capped at 2³¹ regardless.

**4. Upgrade note** — wording corrected; JSON view `model { }` and the `paginate()`/`links()` signature change both noted.

One precision on your last point: widening holds for *arguments*, so callers are indeed fine, but the reverse — a `Long` result assigned into an `Integer` variable — is a compile error under STC, which is the migration this PR exists to force. The note says that in the `count()` section; I kept the pagination paragraph to callers-unaffected so the two don't read as contradicting.

Verified: full `compileGroovy compileTestGroovy`, `grails-views-gson` / `grails-datamapping-core` / `grails-data-graphql-core` tests, and `views-functional-tests:integrationTest --rerun-tasks`.

matrei:
# Round 2

Thanks, this round addresses everything from the first pass:

- The JSON API path now reads the total as a `Number` and widens with `longValue()`, and the new `IterableRenderSpec` case pins the 64-bit `last` offset.
- `PaginationSpec` declares `Long` results and has rows past `Integer.MAX_VALUE` for all three offset helpers.
- `offset` is `Long` on every `paginate()` overload and is read back with `Parameters.long()`, so the request path no longer caps at 2^31.
- The upgrade note now covers `.gson` `model {}` declarations and the `HalViewHelper` signature change, and the section renumbered cleanly to 60 after the 8.0.x merge.

Locally: `grails-views-gson` (184 tests) and the `views-functional-tests` integration suite (52 tests, including `ProductSpec` which drives the real `index.gson` through `hal.paginate`) both pass.

Three documentation nits left, nothing in code.

### 1. Editing leftover in the upgrade note

```
`GormStaticApi.count()` narrowed the datastore result with `intValue()`, which truncates silently. That second step truncates silently, so
a table with more than `Integer.MAX_VALUE` rows ...
```

"truncates silently" appears twice, and "That second step" refers to a step the rewritten sentence no longer describes. Suggest:

```
`GormStaticApi.count()` narrowed the datastore result with `intValue()`, which truncates silently, so
a table with more than `Integer.MAX_VALUE` rows ...
```

### 2. Comment wording in the `.gson` example

`Long bookCount     // or Number, to accept either controller` reads oddly. Perhaps `// or Number, to accept either width`.

### 3. The HAL guide still shows the old model type, and its argument order is wrong

`grails-doc/src/en/guide/theWebLayer/gson/gsonHal.adoc` (around line 255) is the user-facing reference for `hal.paginate()` and this PR changes that signature, so it should move with it:

```groovy
model {
    Iterable<Book> bookList
    Integer bookCount        // now contradicts the upgrade note
    ...
}
json {
    hal.paginate(Book, bookCount, max, offset, sort, order)
}
```

Two things: `Integer bookCount` should become `Long`, and the call passes `max, offset` while the signature is `(object, total, offset, max, sort, order)`. The swap predates this PR, but it still compiles today only because `Integer` widens into the new `Long offset`, and it produces links with `max` and `offset` exchanged. Since the guide is being touched for the type anyway, fixing the order in the same commit avoids shipping an example that silently paginates wrong.

codeconsole:
All three fixed in `c8f5c8f9`.

1. Duplicated clause removed.
2. Now `// or Number, to accept either width`.
3. `gsonHal.adoc` — `bookCount` is `Long` and the call passes `offset, max`. Confirmed against the signature: `paginate(Object, Long total, Long offset, Integer max, String sort, String order)`. Good catch that it only compiles today because `Integer` widens into the new `Long offset` — it would have kept producing links with the two exchanged.

One you didn't mention, same file: the `hal.links()` example at line ~308 also declared `Integer bookCount`, and `links(Map, Object, Long total)` changed in this PR too, so I moved that one to `Long` as well.

matrei (approved):
Thanks, this closes out round 2. Head `c8f5c8f9d4` is the round-2 head plus a merge of `8.0.x` (base now `a5758d657f`) and one doc commit, and `git merge-tree` against the current `8.0.x` is clean.

All three doc nits are fixed:

- The duplicated "truncates silently. That second step" clause in the upgrade note is gone.
- The `.gson` comment now reads "or Number, to accept either width".
- The HAL guide declares `Long bookCount` in both examples, and `hal.paginate(Book, bookCount, offset, max, sort, order)` now matches the `(object, total, offset, max, sort, order)` signature.

I also checked that the `8.0.x` merge did not bring in anything the PR needs to cover. The only merged file that touches a `count()` is `CriteriaMethodInvoker` in `grails-data-hibernate7`, which is a Hibernate criteria projection rather than the static `count()` this PR widens, so nothing to do there. The upgrade note numbering is contiguous after the merge, with the count section now at 62.

Locally on this head:

| Module | Result |
|--------|--------|
| `grails-datamapping-core` | 970 tests, 0 failures, 1 skipped |
| `grails-views-gson` | 184 tests, 0 failures, 0 skipped |
| `grails-data-graphql-core` | 258 tests, 0 failures, 2 skipped |

No further findings. LGTM.

sbglasius (approved):

PR #16325, opened 2026-09-08
Pin that the compiler config generator stays off the runtime classpath

matrei:
The task-based wiring that 9a529d1996 replaced with a doFirst declared runtimeClasspath as an input of the generator task, so any task whose output reached the runtime classpath became a transitive dependency of compileGroovy. A post-compile step that feeds the runtime classpath, such as asset-pipeline output or the exploded plugin directory, then closed a cycle. That is the concern raised on #16114.

The generator introduced there takes its script from grails { } state and declares no classpath input, so that edge no longer exists. This test recreates the scenario: a task that runs after classes and puts its output on the runtime classpath via runtimeOnly files(task). It asserts the generator does not depend on that task, and that compileGroovy and the producer schedule together without a cycle.

### DIFF
--- a/grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/core/GrailsGroovyCompilerConfigSpec.groovy
+++ b/grails-gradle/plugins/src/test/groovy/org/grails/gradle/plugin/core/GrailsGroovyCompilerConfigSpec.groovy
@@ -59,6 +59,29 @@ class GrailsGroovyCompilerConfigSpec extends GradleSpecification {
         result.output.contains('GENERATOR_HAS_CONTENT_INPUT=true')
     }
 
+    def "a task that adds to the runtime classpath after compiling does not cycle through the generator"() {
+        given: 'an application whose runtime classpath carries the output of a task that runs after classes'
+        setupTestResourceProject('compiler-config-runtime-classpath-producer')
+
+        when: 'the wiring is inspected'
+        def result = executeTask('inspectRuntimeProducer')
+
+        then: 'the producer is on the runtime classpath, and the generator does not depend on it'
+        result.output.contains('RUNTIME_CLASSPATH_BUILT_BY_PRODUCER=true')
+        result.output.contains('GENERATOR_DEPENDS_ON_RUNTIME_PRODUCER=false')
+        result.output.contains('COMBINED_CONTAINS_GRAILS_IMPORT=true')
+
+        when: 'the compile task and that producer are scheduled together'
+        def graph = executeTask('compileGroovy', ['packageRuntimeExtra', '--dry-run'])
+
+        then: 'the graph builds without a cycle, the generator ahead of the compile and the producer after'
+        def order = graph.output.readLines().findAll { it.startsWith(':') }
+        order.findIndexOf { it.startsWith(':generateCompileGroovyGrailsCompilerConfig') } <
+                order.findIndexOf { it.startsWith(':compileGroovy') }
+        order.findIndexOf { it.startsWith(':compileGroovy') } <
+                order.findIndexOf { it.startsWith(':packageRuntimeExtra') }
+    }
+
     def "a configurationScript assigned from a later callback is folded in, not clobbered"() {
         given: 'a user assigning configurationScript from projectsEvaluated'
         setupTestResourceProject('compiler-config-late-assignment')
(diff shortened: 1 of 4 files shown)


jdaugherty:
Thanks @matrei  I see now why it can't reoccur.  From discussing with AI, I thought it would be helpful to paste it's output here.  It seems the only regression is if a configuration script was being provided manually (see the end of the AI output). @codeconsole  or @matrei would you mind addressing this with an entry in the upgrade guide? 

# PR #16325 / compileGroovy config-generator regression check

**Bottom line:** the regression does not exist on 8.0.x, PR #16325's test really does guard it, and asset-pipeline was not the trigger. One related shape does still behave differently from the old `doFirst`, detailed at the end.

## What the old regression actually was

The pre-September-2025 wiring registered per-script tasks that declared `inputs.files(configurations.runtimeClasspath)`, with `compileGroovy` depending on them. PR #15076 names the mechanism: "any task that interacts with the runtime classpath now needs to declare a dependency on our customization task". Two concrete failures follow from that input:

- Any task whose output lands on the runtime classpath and itself depends on `classes` closes a cycle through the config task.
- The config task forces `runtimeClasspath` resolution just to write a script, so a build that cannot resolve at that point fails on the generator rather than the compile.

## Why it cannot recur with the merged design

The generator from #16114 takes a string input and, optionally, the build's own config script file. It declares no classpath input, and its only task dependency is the producer of that user script.

Confirmed on real builds: in `grails-test-examples-app1` (asset-pipeline, GSP, three plugin subprojects), dry runs of `bootJar`, `compileGroovyPages`, `integrationTestClasses` and `bootRun` all build their graphs with the generator scheduled directly before each project's `compileGroovy` and nothing else attached.

## PR #16325 verification

- The spec passes on the PR branch, 11 of 11, including the new cycle test.
- **Mutation check:** re-adding `t.inputs.files(project.configurations.named('runtimeClasspath'))` to the generator and rerunning the spec makes the new test fail with exactly the old cycle:

  ```
  Circular dependency between the following tasks:
  :classes
  \--- :compileGroovy
       \--- :generateCompileGroovyGrailsCompilerConfig
            \--- :packageRuntimeExtra
                 \--- :classes (*)
  ```

  Seven other tests also fail under the mutation because the generator then needs to resolve `runtimeClasspath` (`Cannot resolve external dependency org.apache.grails:grails-bom`), which is the second failure mode above. The suite catches either form of the regression.
- The one red CI check is the known flaky Geb login test tracked in #16030, on the Groovy snapshot canary job. Unrelated.

## Asset-pipeline was not the cause

From the 5.2.0-M3 sources, `assetCompile` reads the `runtimeClasspath` and `compileClasspath` configurations and is depended on by `bootJar` (or `processResources` when `developmentRuntime` is off). Nothing it produces reaches `compileGroovy`.

A fixture with the real plugin plus `grails-web`: with the current wiring a full `bootJar` succeeds, and even with the old wiring simulated on the generator the graph still builds, because in a single project nothing task-built is on `runtimeClasspath`. The cycle needs a task-built runtime classpath entry that depends on `classes`, which is the shape the PR's fixture uses and the shape custom packaging steps or the old exploded-plugin wiring produced.

## The remaining gap

A Grails 7-style build that writes its own compiler config script in a task and wires it the old way:

```groovy
tasks.named('compileGroovy') {
    dependsOn generateUserConfigScript
    groovyOptions.configurationScript = file('build/user-config.groovy')
}
```

fails on 8.0.x head (`e8d33a9318`) because the generator has no edge to that producer:

```
A problem was found with the configuration of task ':generateCompileGroovyGrailsCompilerConfig'
Input file does not exist ... property 'configurationScript' specifies file '.../build/user-config.groovy' which doesn't exist
```

Provider wiring works, and the combined script carries both the Grails and user imports:

```groovy
groovyOptions.configurationScriptFile.fileProvider(generateUserConfigScript.map { it.outputs.files.singleFile })
```

The #16114 author acknowledged this in that thread and promised a follow-up regression test plus upgrade-guide text. Neither is on 8.0.x yet: the upgrade guide's section 53 covers only `importJavaTime`, and #16325 does not cover this case. That follow-up should be the condition for calling this area closed. Options:

1. Document the provider wiring as a required change in the upgrade guide.
2. Have the generator inherit `compileGroovy`'s explicit `dependsOn` edges when the user script is a plain file, accepting the author's concern that this could reintroduce cycles.

jdaugherty (approved):
Approving, but I'd like to see us update the upgrade guide.

codeconsole (approved):
Thanks, @matrei. This test captures the original failure mode: a task contributes output to the runtime classpath while depending on `classes`, which would create a cycle if the compiler-config generator consumed that classpath.

Checking task-graph construction is appropriate here because the cycle would fail before execution.

The separate compatibility case—assigning a generated user script as a plain file and adding its producer only to `compileGroovy.dependsOn`—is covered by #16327, including execution-level tests and the upgrade-guide entry James requested.

These changes complement each other. Approving.

PR #16337, opened 2026-09-12
Convert Grails Forge to Grails 8.1.x and fold it into the root build

jamesfredley:
If 8.0 has no room, let's consider for Grails 8.1 or 9.0.

## Summary

Convert Grails Forge from its Micronaut runtime to a Grails 8 Spring application, then fold Forge into the root Gradle build so it is no longer a nested project.

- Replace Micronaut dependency injection with Spring components.
- Add `grails-forge-web` as the hosted Grails REST application.
- Preserve the existing Forge HTTP paths and CLI command names. No intended HTTP API, URL, or CLI command breaking changes.
- Package Elastic Beanstalk as `grails-forge-web-aws.zip` (`app.jar`, `Procfile`, `start.sh`).
- Remove the Micronaut Netty and PostgreSQL analytics applications. Analytics reporting is disabled and PostgreSQL is no longer required.
- Update the AWS Elastic Beanstalk workflow and documentation. Operators deploy a selected Forge slot through the `Forge - AWS Elastic Beanstalk Deploy` workflow using `AWS_FORGE_DEPLOY_ROLE_ARN`.
- Take Groovy, Spock, and related versions from `grails-bom` (Groovy 5.1.2). Isolate `grails-forge-test-core` groovyc from Gradle TestKit's Groovy 4.
- Include Forge as root subprojects (`grails-cli`, `grails-forge-api`, `grails-forge-cli`, `grails-forge-core`, `grails-forge-web`, `grails-forge-test-core`). Move Forge `buildSrc` into `build-logic`. Remove the nested Forge Gradle build, wrapper, and settings.
- Snapshot `publish` waits for `buildForge`, then publishes Forge once with grails-core. Core CI `build -PonlyCoreTests` disables Forge tasks so Forge assemble, shadow JAR, dist zip, and the test-case Maven publish chain do not run on core jobs.
- Groovy snapshot canary uses `-PskipForgeTests` (not `-PonlyCoreTests`) so Forge generate-app tests stay on dedicated Forge jobs. Nested generated apps search `build/local-maven` and cannot resolve the canary's Groovy snapshot from `~/.m2`. `-PonlyCoreTests` would also skip Hibernate, MongoDB, and functional suites that canary is meant to run.
- Local hosted generator: `./gradlew :grails-forge-web:bootRun`

## Build topology

Forge is part of the root Gradle build. It is no longer an included nested build.

## Verification

- Forge module tests (api, core, cli, web, test-core)
- `javadoc` / `checkstyleMain`
- Root `build -PonlyCoreTests --dry-run` has no Forge publication chain


jamesfredley:
At Community over Code we are going to focus on application generation in Profiles vs Forge.  I think we will always have a website application generator, so I went ahead and created this PR, so we can eat our own dog food.

jdaugherty:
I agree this is the right direction, but I disagree this should be done in 8.  We still haven't gotten the "required" 8.0.0-RC1 changes reviewed and this is yet another feature to review.  I'd say this is fine to target for 8.1 though.

grails-forge/grails-forge-core/src/main/java/org/grails/forge/ForgeContexts.java
```diff
@@ -0,0 +1,48 @@
...
+public final class ForgeContexts {
+
+    private ForgeContexts() {
+    }
+
+    public static AnnotationConfigApplicationContext create() {
+        return create(Map.of());
+    }
+
+    public static AnnotationConfigApplicationContext create(Map<String, Object> configuration) {
```
jdaugherty: Nothing on the main classpath consumes the `Environment` this overload populates: there is no `@Value` or `Environment#getProperty` in `grails-forge-core` or `grails-forge-cli`, and Boot's `@ConfigurationProperties` binder is not registered in this plain `AnnotationConfigApplicationContext`, so `GrailsForgeConfiguration` would not bind here either (it only lives in the hosted app, where auto-configuration handles it). The only caller is `ApplicationContextSpec`, which passes an empty map.

Suggest removing this overload and the `configuration` hook in the spec rather than keeping an API that implies configuration takes effect. If a consumer is planned, the property source is only half of it; the reader has to be written as well.

grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/Application.java
```diff
@@ -91,31 +93,36 @@ public static void main(String[] args) {
...
-        try (BeanContext beanContext = ApplicationContext.builder().deduceEnvironment(false).start()) {
+        try (AnnotationConfigApplicationContext beanContext = ForgeContexts.create()) {
             return createCommandLine(beanContext, noOpConsole);
         }
     }
 
     static int execute(String[] args) {
         boolean noOpConsole = args.length > 0 && args[0].startsWith("update-cli-config");
-        try (BeanContext beanContext = ApplicationContext.builder().deduceEnvironment(false).start()) {
+        try (AnnotationConfigApplicationContext beanContext = ForgeContexts.create()) {
```
jdaugherty: Every command still builds and tears down a full context here, and the interactive shell calls this once per line. With classpath scanning replacing Micronaut's compile-time DI that is noticeably more expensive. Measured locally on JDK 21, five runs each, against the `grails-forge-cli` runtime classpath (no shadow jar):

| Command | 8.0.x | this PR |
|---|---|---|
| `--version` | ~295 ms | ~390 ms |
| `create-app --help` | ~310 ms | ~415 ms |
| `create-app temp --list-features` | ~320 ms | ~430 ms |

Acceptable for one-shot `grails create-app`, but in the shell it is paid on every line. Suggest creating a single context in `main` for the interactive path and handing it to `execute`, so the scan runs once per session.

That would also remove the arrangement where `createCommandLine()` closes the context and then gives the shell a `CommandLine` whose `GrailsPicocliFactory` still references it. It works today because picocli instantiates subcommands, converters and the version provider eagerly during construction (completion, usage and version rendering after close all check out), but `GrailsPicocliFactory` only catches `NoSuchBeanDefinitionException`, so anything picocli ever creates lazily would surface as an `IllegalStateException` from the closed context.

grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/Application.java
```diff
@@ -46,28 +50,28 @@
...
-        DevelopmentReloadingCandidates.class,
-        DevelopmentReloadingConverter.class
-})
+@Component
+@Scope("prototype")
 public class Application extends BaseCommand implements Callable<Integer> {
 
     private static Boolean interactiveShell = false;
 
+    private static final List<Class<? extends CodeGenCommand>> CODE_GEN_COMMANDS = List.of(
```
jdaugherty, line 68: This replaces the previous `getBeanDefinitions(CodeGenCommand)` discovery with a static list, so a new code-gen command now has to be added here as well as written. Reasonable trade for a plain Spring context, since these commands take `CodeGenConfig` in their constructor and are not beans, but worth a one-line comment saying this list is the registration point. Otherwise the natural instinct is to annotate the new command and wonder why it never appears.

gradle/forge-test-config.gradle
```diff
@@ -20,23 +20,37 @@
...
+                  'onlyHibernate5Tests',
+                  'onlyHibernate7Tests',
+                  'onlyMongodbTests',
+                  'onlyNeo4jTests',
+                  'onlyRedisTests',
+                  'onlySpringSecurityTests'].any { project.hasProperty(it) })
+
+tasks.withType(Test).configureEach { Test testTask ->
+    if (forgeTestsEnabled) {
+        testTask.dependsOn(
```
jdaugherty: This wires every forge `Test` task, including `grails-forge-core`, `grails-forge-api` and `grails-forge-web`, to publishing all root and grails-gradle publications into `build/local-maven` first. Only the specs that spawn a generated application's Gradle build need that repository: `grails-forge-cli` (`CommandSpec#executeGradleCommand`) and `grails-forge-test-core`. The `GRAILS_REPO_URL` specs in `grails-forge-core` only exercise URL parsing and never resolve anything.

With the fold-in, a plain `./gradlew :grails-forge-core:test` now runs the full framework publish chain before a unit test can start, which the nested build did not do. Suggest scoping the `dependsOn` (and the `GRAILS_REPO_URL` environment) to the two modules that consume the repository, for example via a project `ext` flag those two build files set before applying this script.

grails-forge/grails-forge-web/src/test/groovy/org/grails/forge/web/GrailsForgeConfigurationSpec.groovy
```diff
@@ -17,25 +17,32 @@
...
 
-    @Inject
-    FeatureOperations featureOperations
+        then:
+        config.redirectUrl == 'https://start.grails.org/'
+        config.redirectUri().get().toString() == 'https://start.grails.org/'
+    }
 
-    void "only visible features are exposed"() {
+    void "cors origins remain overridable by environment variables"() {
```
jdaugherty: This asserts the text of `application.yml` rather than the behaviour it stands in for, so it needs editing whenever the yml changes and would still pass if the placeholders stopped being resolved. The two mechanisms behind it are untested: Boot binding `grails.forge.redirect-url` into `GrailsForgeConfiguration` inside the Grails context (the POJO test above bypasses binding), and `corsConfigurationPlaceholderResolver` in `Application` rewriting `${...}` in `GrailsCorsConfiguration#allowedOrigins`. `ForgeApiIntegrationSpec` checks CORS for the literal origins but never that a placeholder was resolved, and the `text/html` branch of `ForgeApplicationController#home` is never taken.

Suggest replacing this feature with integration coverage of those branches: `GET /` with `Accept: text/html` expecting a permanent redirect to the bound URL, and an assertion after startup that no `allowedOrigins` entry still contains `${`.

grails-forge/grails-forge-core/build.gradle
```diff
@@ -42,48 +47,38 @@ sourceSets {
...
 
     testImplementation "ch.qos.logback:logback-classic"
-    testRuntimeOnly "org.objenesis:objenesis:$objenesisVersion"
+    testRuntimeOnly 'org.objenesis:objenesis'
     testImplementation 'org.apache.groovy:groovy-test'
 }
 
 
 def grailsVersionsPath = layout.buildDirectory.dir('version-info')
 def grailsVersionInfoTask = tasks.register('grailsVersionInfo', WriteGrailsVersionInfoTask)
```
jdaugherty, line 78: Anchoring here because the task class itself moved unchanged from the forge `buildSrc` (a pure rename, so GitHub offers no lines in it). Now that `WriteGrailsVersionInfoTask` lives in shared `build-logic`: at line 88 of the task the `GradleException` is constructed without `throw`, so a parse failure falls through with `pom == null` and line 93 fails with a `NullPointerException` instead of the intended message. Add `throw`.

grails-forge/README.md
```diff
@@ -43,15 +43,13 @@ The user interface is [written in React](https://github.com/apache/grails-forge-
 
 ## API
 
-API documentation for the production instance can be found at:
+API usage for the production instance is available at:
 
-* [Swagger / OpenAPI Doc](https://latest.grails.org/swagger-ui/index.html)
-* [RAPI Doc](https://latest.grails.org/rapidoc/index.html)
+* [Production API](https://latest.grails.org/)
```
jdaugherty: With the Micronaut OpenAPI endpoints gone these now point at the plain-text root, which lists the paths but not parameters or response shapes. Is a replacement API description planned (a checked-in OpenAPI document or springdoc), or is the plain-text response the API documentation going forward? Worth stating in the PR either way, since the start.grails.org UI and external tooling consume this API.

grails-forge/README.md
```diff
@@ -43,15 +43,13 @@ The user interface is [written in React](https://github.com/apache/grails-forge-
 
 ## API
 
-API documentation for the production instance can be found at:
```
jdaugherty, line 46: I think this is important to preserve.  We could take @codeconsole 's PR to auto generate this documentation like it was before.

jdaugherty (commented):
Here is an initial AI review, I'll take a closer pass after these issues are looked at: 

Verified locally against the grails-forge-cli runtime classpath and the grails-forge-web module at this head:

    The CLI/shell boots a bare AnnotationConfigApplicationContext (ForgeContexts -> ForgeCoreConfiguration), not SpringApplication, so there is no web-application type in play and nothing to set. grails-forge-web is the only Grails application, and it is not on the CLI or grails-cli shadow-jar classpath.
    Interactive shell, tab completion, --version and help all work. Startup is roughly 100 ms slower per command; numbers are inline on Application.java.

Outside the diff, .agents/skills/gradle-developer/SKILL.md still describes the nested build: line 70 lists grails-forge/ as a separate build with its own wrapper and dependencies, and line 157 says the wrapper bootstrap copies into grails-forge. Both should follow the fold-in.

jamesfredley:
moved to 8.1.x

grails-forge/grails-forge-core/src/main/java/org/grails/forge/ForgeContexts.java
```diff
@@ -0,0 +1,48 @@
...
+        AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+        if (configuration != null && !configuration.isEmpty()) {
+            ConfigurableEnvironment environment = new StandardEnvironment();
+            environment.getPropertySources().addFirst(new MapPropertySource("forge-test", configuration));
+            context.setEnvironment(environment);
+        }
```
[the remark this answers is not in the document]
jamesfredley: Resolved in 932c1cbd81: dropped the unused ForgeContexts.create(Map) overload. GrailsForgeConfiguration binding is covered in the hosted Grails app via ForgeApiIntegrationSpec (HTML redirect to the bound URL).

grails-forge/grails-forge-core/src/main/java/org/grails/forge/ForgeContexts.java
jamesfredley: Resolved in 932c1cbd81: removed the Map overload and the empty configuration hook from ApplicationContextSpec / CommandSpec.

grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/Application.java
jamesfredley: Resolved in 932c1cbd81: the interactive shell now creates one AnnotationConfigApplicationContext for the session and reuses it for each line. One-shot execute still opens and closes its own context. Covered by ApplicationContextReuseSpec.

grails-forge/grails-forge-cli/src/main/java/org/grails/forge/cli/Application.java
jamesfredley: Resolved in 932c1cbd81: added a comment on CODE_GEN_COMMANDS stating it is the registration point because these commands take CodeGenConfig in their constructor and are not Spring beans.

gradle/forge-test-config.gradle
jamesfredley: Resolved in 932c1cbd81: local-Maven publish and GRAILS_REPO_URL are now gated on ext.forgeTestsNeedLocalMaven, set only by grails-forge-cli and grails-forge-test-core.

grails-forge/grails-forge-web/src/test/groovy/org/grails/forge/web/GrailsForgeConfigurationSpec.groovy
jamesfredley: Resolved in 932c1cbd81: dropped the application.yml text assertion. ForgeApiIntegrationSpec now covers GET / with Accept: text/html (301 to the bound UI URL) and asserts CORS allowedOrigins have no unresolved placeholders.

grails-forge/grails-forge-core/build.gradle
jamesfredley: Resolved in 932c1cbd81: WriteGrailsVersionInfoTask now throws the GradleException on parse failure. WriteGrailsVersionInfoTaskSpec covers that path.

grails-forge/README.md
jamesfredley: Resolved in 932c1cbd81: restored OpenAPI 3 for the hosted generator from grails-forge-web (not Micronaut OpenAPI) at /v3/api-docs, with Swagger UI and RapiDoc. README now points at those URLs. A general grails-openapi module is not required for these endpoints.

grails-forge/README.md
jamesfredley: Preserved in 932c1cbd81 with a checked-in OpenAPI document served by ForgeOpenApiController. A later auto-generated grails-openapi module can replace this Forge-specific description without blocking this PR.

PR #16338, opened 2026-09-13
Keep the embedded MongoDB tests from hanging CI jobs for six hours

codeconsole:
### Problem

Neo4j and MongoDB functional jobs on `8.0.x` were cancelled at GitHub's six-hour limit six times between 2026-08-26 and 2026-09-11, most recently [this Neo4j job](https://github.com/apache/grails-core/actions/runs/34655597581/job/103452387382). Each hung in `:grails-data-mongodb-embedded:test`: three after every spec had passed, and three while `stopping a server that is already stopped is what a shutdown hook does` was running.

An ordinary shutdown stops the embedded server twice. `EmbeddedMongoLifecycle` stops it when the application context closes, and the shutdown hook `EmbeddedMongoInitializer` registers stops it again as the JVM exits. The in-memory backend called `MongoServer.shutdownNow()` both times:

```java
@Override
public void stop() {
    this.server.shutdownNow();
    this.running = false;
}
```

A second `shutdownNow()` can wait forever. A client that connected just as the first shutdown began can be registered by Netty after its event loop has closed its channels, so the loop terminates with the client still in mongo-java-server's channel group, and `closeClients()` waits, with no timeout, on a close the terminated loop never completes. Gradle's test worker exits through `System.exit`, which waits for every shutdown hook, so the task never finished.

The Neo4j jobs ran these tests at all because `mongodb-test-config.gradle` did not skip its suites under `-PonlyNeo4jTests`.

### Fix

`RunningInMemoryMongo.stop()` returns when the server is already stopped, and `stop()` and `restart()` are synchronized, since the context's shutdown hook and the initializer's run at the same time on the way out:

```java
@Override
public synchronized void stop() {
    if (!this.running) {
        return;
    }
    this.server.shutdownNow();
    this.running = false;
}
```

The rest is the test and CI configuration this exposed:

- `EmbeddedMongoReplicaSetSpec` stops its servers through the lifecycle bean. Its cleanup now follows `EmbeddedMongoInitializerSpec`. It called `close()` on contexts that were never refreshed, which does nothing, so its four mongods were left for the shutdown hook.
- `mongodb-test-config.gradle`, `hibernate5-test-config.gradle`, `hibernate7-test-config.gradle`, `redis-test-config.gradle` and `spring-security-test-config.gradle` skip their suites under `-PonlyNeo4jTests`. The Neo4j jobs had been running the MongoDB suites, 19 Hibernate test tasks, and Spring Security suites that their dedicated jobs already run.
- `grails-redis` applies only `redis-test-config.gradle`, as the Hibernate and MongoDB modules apply only their own config. With `test-config.gradle` applied as well, its `test` and `integrationTest` tasks ran in no CI job, the Redis jobs included.
- The nine test jobs set `timeout-minutes` to twice the slowest run that finished on its own, passed or failed, over the 160 CI runs from 2026-09-03 to 2026-09-13, rounded up to the half hour: 90 minutes for the Spring Security config and CAS jobs, 120 to 210 for the rest. The build jobs keep GitHub's default, since the macOS build has passed after more than five hours.

The unbounded wait itself is fixed in mongo-java-server by [bwaldvogel/mongo-java-server#256](https://github.com/bwaldvogel/mongo-java-server/pull/256).

**Worth calling out:**
- `grails-redis` no longer takes the `--add-opens` JVM arguments, `byte-buddy` and `objenesis` test runtime dependencies, `forkEvery` / `maxParallelForks` / `maxHeapSize`, the `**/*$*.class` exclusion, or the CI build-cache opt-out from `test-config.gradle`. Its current tests do not mock classes and pass without these settings against Redis 7.4 and 8.0. Future class mocks may need those runtime dependencies added explicitly.
- The timeouts come from ten days of runs, and a job whose suite grows past half of its limit will need a higher one.

### Tests

`an application whose context stopped its in-memory server still exits` runs an application in a JVM of its own that lets a client connect, stops the server as its context closes, and returns from `main`. The hang is a timing-dependent race. On the author’s machine, the Java child application, `ContextStoppedServerApplication`, failed to exit in 8 of 8 runs without the fix. On the reviewer’s Linux machine, the unfixed child exited in all eight runs while the in-process double-stop feature reproduced the hang. Both tests exercise the race; neither is guaranteed to reproduce it on every machine. The child feature fails after 30 seconds if the JVM does not exit.

The in-process double-stop feature repeats the start / connect / stop / bounded-stop sequence ten times on separate ports. Each second stop has a 30-second bound, so a reproduced regression fails instead of hanging the build. The child JVM uses `slf4j-simple` so its captured log includes initializer diagnostics. `RunningEmbeddedMongo.stop()` now documents that stopping an already stopped server must do nothing.

`:grails-data-mongodb-embedded:test`, and `:grails-redis:test` with `:grails-redis:integrationTest`, are green.


matrei (approved):
# Review Findings

Head `38f32adee7` on `8.0.x` base `3067d0a855` (the current tip of `8.0.x`, so the merge is clean). Seven non-merge commits, each one a separate concern: the in-memory double-stop fix, the replica set spec cleanup, the `-PonlyNeo4jTests` gaps in three test configs, job timeouts, a corrected class comment, and `grails-redis` dropping `test-config.gradle`.

Verdict: approve. The backend fix is right, small and matches what the flapdoodle backend already does. One finding worth acting on before merge is about what the new regression test actually proves; the rest are small.

Verified on this head (JDK 21, Linux):

| Run | Result |
|---|---|
| `:grails-data-mongodb-embedded:test :grails-data-mongodb-embedded:codeStyle --rerun-tasks` | 51 tests, 0 failures, 0 skipped; codeStyle clean |
| `:grails-redis:check -PonlyRedisTests -PredisContainerVersion=8.0 --rerun-tasks` | `test` 4/0/0, `integrationTest` 33 run, 0 failed, 2 skipped (the two `@Ignore`d `RedisTagLibSpec` features) |
| Control: `InMemoryMongoBackend.java` restored to the base version, the two double-stop features re-run | see finding 1 |

The `grails-redis` figures match the commit message, and on the base both tasks were indeed skipped under `-PonlyRedisTests` because `test-config.gradle` lists that flag. I also confirmed that no Neo4j module applies `hibernate5-test-config`, `hibernate7-test-config`, `mongodb-test-config` or `redis-test-config`, so adding `onlyNeo4jTests` to those lists cannot skip anything the Neo4j jobs are meant to run.

## Findings

### 1. Medium: the out-of-process regression test does not reproduce the hang here, and the in-process one does

`ContextStoppedServerApplication` and the first commit message say the child JVM "never exited in 8 of 8 runs" without the fix, and that "started through Groovy and Spock, the same steps almost never hang even unfixed". On this machine it is the other way round. With `InMemoryMongoBackend.java` restored to the base version:

```
stopping a server that is already stopped is what a shutdown hook does   FAIL  30.6 s  (TimeoutException from the bounded second stop)
an application whose context stopped its in-memory server still exits     ok     1.6 s
```

and running the child application by hand against the unfixed classes, eight times in a row:

```
run 1..8  exit=0  1255-1273 ms
```

So the race that leaves a client channel tracked on a terminated event loop is timing-dependent, and which of the two features hits it depends on the machine. That is fine as far as the fix goes: the fix is correct regardless, and between them the two features did catch the regression here. But the class comment on `ContextStoppedServerApplication` and the commit message present the Java child process as the deterministic reproducer and the Spock feature as the one that "proves nothing", which a reader on a machine like this one will find contradicted.

Suggestions, none blocking:

- Soften the comment on `ContextStoppedServerApplication` (and the PR description) to say the hang is a race that reproduced reliably as a separate JVM on the author's machine, rather than that the in-process steps "almost never" hang.
- Consider widening the net in `stopping a server that is already stopped is what a shutdown hook does`: loop the start / connect / stop / bounded-stop sequence a handful of times (say 10, on ports that do not collide with the other features) so that a regression has more chances to hit the window on any machine. The bounded `CompletableFuture...get(30, SECONDS)` already turns a hit into a failure rather than a hung build, which is the important part.

References: `grails-data-mongodb/embedded/src/test/java/org/grails/datastore/gorm/mongodb/embedded/ContextStoppedServerApplication.java:28-38`, `EmbeddedMongoLifecycleSpec.groovy:280-311`.

### 2. Low: `spring-security-test-config.gradle` has the same `-PonlyNeo4jTests` gap

The PR adds `onlyNeo4jTests` to the Hibernate 5, Hibernate 7, MongoDB and Redis configs so their suites stay out of the Neo4j jobs. `gradle/spring-security-test-config.gradle:27-35` has the identical list without `onlyNeo4jTests`, and it is applied by twelve `grails-spring-security/*` modules, of which `plugin` (28 specs), `rest/core` (19), `ui/plugin` (4) and three others have tests. The three Neo4j jobs run `bootJar check -PonlyNeo4jTests`, so they still run those suites, which the `security` jobs already run. One more line in that list closes it; alternatively leave it for a follow-up, but then the commit message "every other datastore's -Ponly flag except -PonlyNeo4jTests" describes only four of the five configs.

### 3. Low: the idempotent `stop()` is now relied on, but the interface does not say so

`EmbeddedMongoInitializer.stopEverythingAtExit()` calls `stop()` on every started server without checking `isRunning()`, and `EmbeddedMongoLifecycle.stop()` checks it but races with that hook. Both backends now make a second `stop()` a no-op, and the whole PR hinges on that. `RunningEmbeddedMongo.stop()`'s javadoc (`RunningEmbeddedMongo.java:36-39`) only says "Called from a JVM shutdown hook, so it must not throw". Adding "and it may be called again after the server has stopped, which must do nothing" turns the fix into a contract a third backend would have to honour, rather than a property of the two shipped ones.

### 4. Low: what `grails-redis` loses with `test-config.gradle`

Dropping `test-config.gradle` is the right fix for the tasks never running, and the module passes without it. For the record, it also loses the `--add-opens` JVM arguments, the `byte-buddy` and `objenesis` test runtime dependencies (nothing in `grails-redis` mocks a class, so unused today), `forkEvery`/`maxParallelForks`/`maxHeapSize`, the `**/*$*.class` exclusion, and the CI build-cache opt-out. `mongodb-test-config.gradle` does without all of these too, so this is consistent with the other datastore configs; just worth knowing if a future Spock `Mock(SomeClass)` in this module fails with a missing byte-buddy.

### 5. Nits

- The commit "Stop the servers the replica set specification starts" cites `EmbeddedReplicaSetSpec` as prior art; no such class exists. It means `EmbeddedMongoInitializerSpec`, which has the identical `cleanup()`.
- `EmbeddedMongoLifecycleSpec.startApplication` redirects the child's output to a log and prints it on failure, but the test runtime binds `slf4j-nop`, so the log will contain only JVM and Netty stderr, never the initializer's "Embedded MongoDB started at" lines. If the log is meant to help diagnose a hang on CI, `slf4j-simple` in `testRuntimeOnly` would make it say something.
- The timeout values in `gradle.yml` are stated as twice the slowest observed run over 160 CI runs. I cannot verify those figures locally; the ratio is generous enough that a slow runner will not trip them, and the comment explaining the method is a good precedent.

## What I checked and found no problem with

- `RunningInMemoryMongo.stop()` and `restart()` are `synchronized`, `running` and `server` stay `volatile` for the unsynchronized `isRunning()` reader, and `running` is set only after `shutdownNow()` returns, so a first stop that throws leaves the server marked running for a retry, the same as flapdoodle's `finally`.
- The mechanism described in the javadoc matches mongo-java-server 1.47.0: `shutdownNow()` calls `closeClients()`, which is `channelGroup.close().syncUninterruptibly()` with no timeout, and `MongoDatabaseHandler.channelActive` adds the channel to the group.
- `EmbeddedMongoReplicaSetSpec`: each feature's `cleanup:` block (closing session and client) runs before the spec-level `cleanup()` stops the server, and a feature whose `initialize()` throws registers no lifecycle bean, which the `containsBean` guard handles.
- The argument file for the child JVM quotes and escapes per the JDK `@argfiles` rules, and port 27969 is not used anywhere else in the repository.
- No user-facing behaviour changes, so no `grails-doc` update is needed.

jdaugherty (approved):
Once Mattias feedback is addressed, I'm good to merge.

codeconsole:
Addressed the Review Findings in b4aa3c6622.

- The bounded in-process shutdown regression now runs ten times on separate ports. The Java fixture comment and PR description now describe both reproductions as timing-dependent.
- Added `onlyNeo4jTests` to the Spring Security test configuration.
- Documented that `RunningEmbeddedMongo.stop()` must do nothing when the server is already stopped, and strengthened the stopped-state assertions for both backends.
- Enabled `slf4j-simple` for the embedded module's tests so the child JVM log includes initializer diagnostics.
- Clarified the Redis settings and runtime dependencies removed with `test-config.gradle`, and corrected the cleanup reference to `EmbeddedMongoInitializerSpec` in the PR description and follow-up commit message.

PR #16351, opened 2026-09-17
build: migrate remaining Java sources to Groovy

borinquenkid:
## Summary

Converts the framework's remaining Java source files to Groovy, across ~90 commits touching essentially every module: grails-core, grails-bootstrap, grails-encoder (including the 2932-line `StreamCharBuffer.java`), grails-shell-cli, grails-beans-dsl, grails-i18n, grails-controllers, grails-databinding-core, grails-web-url-mappings, grails-web-common, grails-converters, grails-gsp (sitemesh3/web-gsp-taglib/plugin-main/spring-boot/web-gsp/web-taglib/layout/taglib/web-jsp), the grails-gradle model module, and more.

This idea was mine; the implementation and verification work below was Claude's, under my direct supervision throughout.

A cross-cutting correctness bug was found and fixed along the way: under `@CompileStatic`, `Closure.call(new Object[]{x})` resolves to the `call(Object)` overload instead of `call(Object[])`, double-wrapping arguments — this had silently corrupted single-param tag-closure calls in already-committed grails-taglib/grails-web-jsp/grails-core/grails-converters code before it was caught and fixed.

Every remaining `.java` file in the repository is either in a module deliberately excluded from this migration (grails-data-hibernate5 — since removed from `9.0.x` separately, grails-forge, grails-benchmarks, grails-test-examples, grails-wrapper, grails-undertow's vendored Spring Boot code) or one of several individually-verified deliberate exclusions (e.g. `package-info.java` files, standard practice for package-level javadoc).

Merged forward from `origin/9.0.x` throughout rather than rebased, per this branch's established practice, most recently a 45-conflict merge plus two Groovy 6 upgrade compatibility fixes.

## Test plan

- [x] Every batch of conversions was compiled and its module's full test suite run before moving to the next
- [x] A `javap`-based audit confirmed no accidental MOP (meta-object protocol) leaks or signature drift across converted classes
- [x] Downstream consumers of each converted module were independently verified to still compile
- [x] Checkstyle/CodeNarc clean across converted files


jdaugherty:
There are a lot of test failures and this review is very large. Can I suggest we split this by gradle project and stack the PRs so we can merge as one once all are reviewed?

PR #16356, opened 2026-09-20
8.0.0-SNAPSHOT  grails-converters needs compileOnlyApi for jakarta.servlet-api

codeconsole:
`ConvertersExtension` is a Groovy extension module whose `getJSON`/`getXML` signatures take `HttpServletRequest`. The Groovy compiler loads extension modules from any compile classpath they are on, so a module that depends on `grails-converters` without a web starter fails `compileGroovy`:

```
NoClassDefFoundError: Unable to configure org.grails.web.converters.ConvertersExtension due to missing dependency jakarta/servlet/http/HttpServletRequest
```

Declare the servlet API `compileOnlyApi`, as `grails-controllers`, `grails-web-gsp` and `grails-web-taglib` already do. Runtime classpath unchanged; the module metadata puts it on `apiElements` only.

Note: #16237 moves `getXML` into a new `grails-xml` with a `compileOnly` servlet declaration, which would need the same change.


matrei (approved):
Should we do this for `grails-web-core` and `grails-mimetypes` as well for completeness?

sbglasius (approved):

jdaugherty:
@matrei that's probably a good idea, i'd like to go ahead and merge this to get the RC1 milestone closer to getting done.  Will you open a separate PR?

matrei:
> @matrei that's probably a good idea, i'd like to go ahead and merge this to get the RC1 milestone closer to getting done. Will you open a separate PR?

@jdaugherty Opened #16360

PR #16359, opened 2026-09-21
docs: @Configuration classes in resources.groovy are no longer processed

matrei:
## Summary

Documents the Grails 8 breaking change reported in #16029. The code fix (#16269) was closed as won't-fix since `resources.groovy` is meant for bean definitions only, so the change is documented instead.

Adds a note to the Grails 8 upgrade guide, under _Plugin Beans Now Register Before Spring Boot Auto-Configuration_:

- `@Configuration` classes registered in the application's `resources.groovy` / `resources.xml`, or in the `Application` class's `doWithSpring` / `beanRegistrar()`, are no longer processed. They are created as plain beans and their `@Bean` methods, `@Import` and `@Enable...` annotations are silently ignored.
- `@Configuration` classes contributed by a plugin's `doWithSpring` / `beanRegistrar()` are unaffected.
- Replacements: `@Import` or `@ComponentScan` on the `Application` class, or declaring the beans directly (`resources.groovy` or a `beans` block on the `Application` class).
- Guidance for plugin authors who relied on users replacing a plugin-provided `@Configuration` class.

Also qualifies the sentence in _More Framework Beans Are Cleanly Overridable_ that said `resources.groovy` beans behave "exactly as in previous releases".

Relates to #16029


jdaugherty (approved):

sbglasius (approved):

matrei:
Closes #16029

